PR #16, opened 2019-09-24
[nitro-protocols] Expose createRefuteTransaction

cf19drofxots:


packages/nitro-protocol/src/transactions.ts
```diff
@@ -32,8 +32,20 @@ export function createRespondTransaction(
   );
 }
 
+export function createRefuteTransaction(channelStorage: ChannelStorage, refute: SignedState) {
+  if (!channelStorage.challengeState) {
+    throw new Error('No active challenge in challenge state');
+  }
```
snario: I'm not sure this is a helpful or necessary error to throw; probably better to let the normal destructuring of the object handle it.

cf19drofxots: I've copied this from how other wrappers work - I did not look at where error handling happens in the pipeline.

cf19drofxots:
`refute` is being removed [here](https://github.com/statechannels/nitro-protocol/pull/102) so there's nothing to expose.

Issue #29, opened 2019-09-24
[Build] Get jest-gas-reporter working with Nitro-Protocol


lalexgap:
Currently I've disabled the `jest-gas-reporter` as `jest` doesn't seem to resolve it to the correct location in node_modules.

It looks like it's looking in the correct location (based on a console.log added to `defaultResolver.js`:
```Users/alexgap/code/monorepo/packages/nitro-protocol/node_modules/@statechannels/jest-gas-reporter
/Users/alexgap/code/monorepo/packages/nitro-protocol/node_modules/@statechannels
/Users/alexgap/code/monorepo/packages/node_modules/@statechannels/jest-gas-reporter
/Users/alexgap/code/monorepo/packages/node_modules/@statechannels
/Users/alexgap/code/monorepo/node_modules/@statechannels/jest-gas-reporter
/Users/alexgap/code/monorepo/node_modules/@statechannels
/Users/alexgap/code/monorepo/node_modules/@statechannels/jest-gas-reporter/lib
/Users/alexgap/code/node_modules/@statechannels/jest-gas-reporter
/Users/alexgap/code/node_modules/@statechannels
/Users/alexgap/node_modules/@statechannels/jest-gas-reporter
/Users/alexgap/node_modules/@statechannels
/Users/node_modules/@statechannels/jest-gas-reporter
/Users/node_modules/@statechannels
/node_modules/@statechannels/jest-gas-reporter
/node_modules/@statechannels
```

However `jest` seems to think all those paths are invalid and never finds anything in any of them.

lalexgap:
Resolved by https://github.com/statechannels/monorepo/pull/30#discussion_r327880898

snario:
https://github.com/statechannels/monorepo/pull/30#discussion_r327912514

PR #30, opened 2019-09-24
[Build] - Get tests passing

lalexgap:
### Description

Gets everything compiling and the tests passing.
- Switches to the correct ganache env variables `GANACHE_HOST/GANACHE_PORT`
- Removes invalid `aysnc` in `description`
- Fixes all ts-lint errors in web3torrent
- Work around a typing error in the test (see 5fa2920)
- Comment out `jest-gas-reporter` for now (see #29)


packages/nitro-protocol/jest.config.js
```diff
@@ -1,7 +1,11 @@
 module.exports = {
   setupFilesAfterEnv: ['./jest.setup.js'],
   collectCoverageFrom: ['src/**/*.{js,jsx,ts,tsx}'],
-  reporters: ['default', '@statechannels/jest-gas-reporter'],
+  // TODO: It looks like jest is having some problems resolving to the jest-gas-reporter in node_modules
+  // We need to figure out what the exact issue is and a way to work around it
+  // reporters: ['default', '@statechannels/jest-gas-reporter'],
```
snario: I believe you saw this issue because you ran `yarn test` before running `yarn build`. You need to run `tsc -b` once in the `nitro-protocol` package such that the [references](https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/tsconfig.json#L19) (see [here](https://www.typescriptlang.org/docs/handbook/project-references.html)) built `@statechannels/jest-gas-reporter`.

snario: I think we can change `main` to export a `.ts` file: https://github.com/statechannels/monorepo/blob/master/packages/jest-gas-reporter/package.json#L6

lalexgap: Wouldn't `lerna` have run `yarn build` before `yarn test`? I was seeing this on circleCI (see https://circleci.com/gh/statechannels/monorepo/69)

lalexgap: This was fixed by changing `jest-gas-reporter` to export a `.ts`

lalexgap: I tried this but I'm still seeing the error so I've reverted to the commented out version for now.

lalexgap: Here's an example of the failure https://circleci.com/gh/statechannels/monorepo/104

snario: Ah I figured it out ... because I added `composite: true` to `tsconfig.json` of `@statechannels/jest-gas-reporter` the output directory structure was `lib/src/*` not `lib/*`. Fixed in 1c446b16.

PR #31, opened 2019-09-25
Remove refute :(

andrewgordstewart:
Since refute can return the value of channelStorageHashes to a previous value,
it always leaves Alice to a front-running attack by Eve. Here's a new
one, discovered with updated TLA+ specs linked below:

0. Say the turnNumRecord is 100, on chain.
1. Alice calls forceMove(s100) in transaction `tx`
2. Eve front-runs with `forceMove(s100)`. Eve is now the challenger.
3. `tx` reverts
4. Eve then calls `refute(s1000)` with a made-up state. Go to 0

There are solutions to this problem, such as storing a lookup table of
who has challenged at the current turnNumRecord. For the moment, it is
simpler to remove refute and later consider our options on how to use
refute.

Spec: https://github.com/magmo/specs/blob/b20081890c9b4892759934c2418fd7262afa0f90/ForceMove/ForceMove.tla


geoknee:
> For the moment, it is simpler to remove checkpoint and later consider our options on how to use
refute.

Presumably you mean remove `refute`?

PR #33, opened 2019-09-25
[documentation] NatSpec the ForceMove contract 

geoknee:
### Description

[NatSpec](https://solidity.readthedocs.io/en/develop/natspec-format.html) is the Ethereum Natural Language Specification Format. It calls for special comments in the solidity source code that describe ABI, and can be displayed to the user when they interact with the contract (e.g. in MetaMask).

Adding the NatSpec also allows for the `solidity-docgen` tool to generate some markdown in a customizable format. This PR explores using that before `docusaurus-start` in order to get autogenerated pages for our documentation website that should make (with relative ease) stay in sync with our code/implementation.

Downsides to `solidity-docgen`: 
- does not currently support contracts that [have external imports](https://github.com/OpenZeppelin/solidity-docgen/issues/24). I had to isolate a couple of contracts that don't import into their own folder to get this demo working. 
- isn't documented yet, so configuration and templating involves a fair bit of guesswork

We should probably do the NatSpec regardless of whether we think `solidity-docgen` is useful to us. 

Instructions:
1. `cd website`
2. `yarn start`
3.  `cd ..`
3.   Update the template file `/docs/implementation/contract.hbs` and/or the NatSpec comments in the Solidity source `/contracts/ForceMove/ForceMove.sol`
4. The website should live reload to show the changes

### Related issues
n/a


packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -1,7 +1,7 @@
 pragma solidity ^0.5.11;
 pragma experimental ABIEncoderV2;
 import './Outcome.sol';
-import 'openzeppelin-solidity/contracts/math/SafeMath.sol';
+import './openzeppelin/math/SafeMath.sol';
```
snario: Why make this local?

geoknee: This was one strategy to get around the limitations of `solidity-docgen`.

> Downsides to `solidity-docgen`: 
- does not currently support contracts that [have external imports](https://github.com/OpenZeppelin/solidity-docgen/issues/24)

It uses `solc`, which does not allow imports from contracts outside of the directory of the importing contract.

geoknee:
Closing in favor of #70

PR #38, opened 2019-09-25
Nitro-protocol: Switch from truffle to etherlime

kerzhner:
This is ground work for sharing contracts deployed to ganache between multiple packages.
- Switch from truffle to etherlime.
- Separate contract compilation and deployment.


packages/nitro-protocol/deployment/deploy-test.js
```diff
@@ -0,0 +1,74 @@
...
+const countingAppArtifact = require('../build/contracts/CountingApp.json');
+const singleAssetPaymentsArtifact = require('../build/contracts/SingleAssetPayments.json');
+
+const erc20AssetHolderArtifact = require('../build/contracts/ERC20AssetHolder.json');
+const ethAssetHolderArtifact = require('../build/contracts/ETHAssetHolder.json');
+const tokenArtifact = require('../build/contracts/Token.json');
+
+let contractsToAddresses = {};
+
+async function getNetworkMap() {
```
kerzhner: At some point, we probably want to factor out common functionality shared by deployers line `deploy.js` and `deploy-test.js`.

packages/nitro-protocol/deployment/deploy-test.js
```diff
@@ -0,0 +1,74 @@
...
+  try {
+    return await loadJsonFile(path.join(__dirname, 'network-map.json'));
+  } catch (err) {
+    if (!!err.message.match('ENOENT: no such file or directory')) {
```
kerzhner: What's the reason for the `!!` here?

packages/nitro-protocol/package.json
```diff
@@ -33,15 +33,18 @@
...
-    "deploy": "yarn clearContracts && npx truffle migrate",
+    "test:prepare": "yarn lint:check && yarn deploy:test",
+    "test": "yarn test:prepare && npx jest",
+    "test:ci": "yarn contract:compile && yarn test:prepare && npx jest --ci --bail",
+    "deploy": "npx etherlime deploy --compile=false",
+    "deploy:test": "yarn deploy && npx etherlime deploy --compile=false --file=./deployment/deploy-test.js",
+    "startGanache": "npx etherlime ganache --verbose > ganache.log",
     "build:typescript": "tslint --project . && tsc -b",
-    "build:truffle": "npx truffle compile",
+    "contract:compile": "npx etherlime compile --runs=1 --buildDirectory=./build/contracts",
```
kerzhner: The value of 1 for `--runs` is just something I tried at some point to turn on the optimizer. I am not actually sure if that is the correct value to use.

packages/nitro-protocol/deployment/deploy-test.js
andrewgordstewart: `!!anyExpression` is always a boolean value. So, if `anyExpression` is truthy, then `!!anyExpression == true`

PR #41, opened 2019-09-25
[wallet] [frontend] Implements first version of an Embedded Wallet, with basic data flow 

### LINKED ISSUE #40: [wallet] [frontend] Implement basic data flow
Any dapp should be able to include a `wallet.js` file via CDN that setups an embedded wallet in the dapp.

Furthermore, that dapp would be able to send a JSONRPC message via `window.postMessage()` and request action from the embedded wallet.

joelalejandro:
Resolves #40.


joelalejandro:
Will need to rewrite this using a provider/consumer paradigm to prevent React from complaining about side-effects when tests are running.

packages/embedded-wallet/src/json-rpc-router/JsonRpcRoute.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * ```
+ * <JsonRpcRoute method="jsonrpc_doSomething" component={DoSomethingUI} />
+ * ```
+ *
+ * @property method - The name of the JSON-RPC method.
+ * @property component - A React component that will be rendered as a response to the message.
+ */
+const JsonRpcRoute: React.FC<JsonRpcRouteProps> = ({ method, component }: JsonRpcRouteProps) => {
+  return (
+    <JsonRpcRouterConsumer>
```
spersico: should the responsibility of defining which component to render fall back to the routes, or to the router?.
Correct me if I'm wrong, but this logic check shouldn't be on the router instead of here?

joelalejandro: In the declaration, the Route defines which method goes with each component. The router would have to be inspecting each child element instead of delegating and letting them resolve whether to render or not on their own.

Take a look at [what `react-router` does](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Route.js#L38-L42). Granted, they have a `Switch` component that does the logic check, but they _also_ check for the match locally if it hasn't been computed somewhere else.

PR #43, opened 2019-09-25
Reorder jobs

andrewgordstewart:
It seems like circleci runs all "prepared" jobs in alphabetical order,
where a job is "prepared" if its requirements are satisfied.

The workflow is currently defined here https://github.com/statechannels/monorepo/blob/96e3ac8bf6eedc5b58fc704f25e2273774ee1fa2/.circleci/config.yml#L110-L125

By pre-pending the job titles with letters, we force tests to run first,
accomplishing the goal of b693452

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ jobs:
             - packages/*/build
             - packages/*/lib
 
-  run-tests:
+  a-run-tests:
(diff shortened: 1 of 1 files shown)


andrewgordstewart:
I think I hit a false positive -- it seems like jobs are run in some arbitrary order.

PR #46, opened 2019-09-26
Faster circle runs

andrewgordstewart:
This combines the circleci four jobs into one job.
- `ensure-updated-lockfiles` was redundant, since the only step it ran, `yarn --frozen-lockfile`, is included in other jobs
- restoring the cache, and running `yarn --frozen-lockfile` takes about 40 seconds. So, combing the three remaining jobs into one job saves 80 seconds.

I don't see much benefit in having separate jobs for building, linting, and testing. If there's no benefit, we might as well save ourselves at least 160s per pull request (once on the PR, and once on the merge into master).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ restore_dep: &restore_dep
     key: dependency-cache-{{ checksum "yarn.lock" }}
 
 jobs:
-  build:
+  test:
     <<: *defaults
     steps:
       - <<: *restore_code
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -38,7 +38,11 @@ jobs:
           name: "yarn --frozen-lockfile"
           command: yarn --frozen-lockfile
 
+      - run: yarn lint
       - run: yarn build:ci
+      - run:
+          name: Run State Channels tests
```
lalexgap: Do we want to remove the name?

PR #49, opened 2019-09-26
Monorepo adjustments to support adding magmo/wallet

snario:
This PR supports @lalexgap's `add-magmo-wallet` by fixing some of the issues preventing the tests from passing. There is still an outstanding issue which is that `yarn build` fails in `wallet` due to some `tslint` error (trailing commas).


.prettierrc
```diff
@@ -1,4 +0,0 @@
-{
-  "printWidth": 120,
```
andrewgordstewart: How is prettier configured if you remove this file?

packages/wallet/src/__stories__/dummy-wallet-states.ts
```diff
@@ -14,24 +14,26 @@ export const defaultParams = {
...
 // WALLET NOT INITIALIZED //
 ////////////////////////////
 
 export const dummyWaitForLogin: walletStates.WalletState = walletStates.waitForLogin();
-export const dummyWaitForMetaMask: walletStates.WalletState = walletStates.metaMaskError({
-  ...dummyWaitForLogin
-});
+export const dummyWaitForMetaMask: walletStates.WalletState = walletStates.metaMaskError(
+  {
+    ...dummyWaitForLogin,
```
andrewgordstewart: It becomes difficult to manage PRs that are merged _after_ a change to prettier configuration:
- I might have significantly changed the AST in `my-branch`
- You merge `your-branch`, which has a lot of LOC changed due to `prettier:write`
- I have to spend a lot of time resolving merge conflicts.

If you contain the changes to prettier configuration into its own PR, with no other changes, then the converse is easy:
- I might have significantly changed the AST in `my-branch`
- You change prettier config in `your-branch` in commit `C1`
- You run `yarn prettier:write` in commit `C2`
- You notice that `my-branch` was merged into master, and when you write `git rebase master`, you have merge conflicts in `your-branch`

Resolving your merge conflicts is easy: `git reset --hard C1 && yarn prettier:write && git add -a && git commit -, $MESSAGE`

Because of this, we use the convention that anytime a tool is used to automatically modify code,
1. those changes are committed in a single commit
2. it's contained in the smallest PR possible
3. once there are no open PRs, we resolve merge conflicts in the commit where the code was modified, and merge it.

packages/wallet/contracts/pre-built-artifacts/ERC20AssetHolder.json
```diff
@@ -0,0 +1,4676 @@
+{
```
andrewgordstewart: The majority of this file is not needed: IIRC, we only need `contractName`, `abi`, and `bytecode`. If you run 
```
$ jq '{abi: .abi, bytecode: .bytecode, contractName: .contractName}' build/contracts/$CONTRACT.json  > tmp && mv tmp contracts/pre-built-contracts/Migrations.json
```
then this diff is more manageable, and the git repository does not balloon

andrewgordstewart: (Assuming that you've got `jq` installed)

andrewgordstewart: Oh, we'll eventually need the addresses from the `networks` property, but
- it's not necessary to get the development & test environments working
- we are testing out deploys with `etherlime`, which makes the management of prebuilt-artifacts saner

packages/wallet/src/__stories__/dummy-wallet-states.ts
snario: In this case, I'm not intending to edit any prettier configurations at all. When I ran `yarn build` I hit `tslint` issues (specifically, `trailing-comma` was not respected). So, to fix that, I ran `tslint --fix`.

packages/wallet/contracts/pre-built-artifacts/ERC20AssetHolder.json
snario: Got it, thanks :)

snario:
@andrewgordstewart I updated this PR to simplify it dramatically, stripping out unnecessary bits.

packages/wallet/src/utils/nitro-converter.ts
```diff
@@ -1,13 +1,17 @@
...
+import {
+  ChannelStorage,
+  SignedState,
+  State,
+  encodeConsensusData,
+  Channel as NitroChannel,
+  Signatures,
+  AllocationItem,
+  Outcome
+} from "@statechannels/nitro-protocol";
```
andrewgordstewart: 👍

snario:
CircleCI tests do pass btw but hitting an out of memory issue...

PR #57, opened 2019-09-26
Parallel tests

andrewgordstewart:
Run nitro-adjudicator tests in parallel.

This required a change in how we notice events in tests. The officially recommended way causes issues. Pulling the events from the transaction receipt always works. Perhaps we could write some test utility to do some of this work for us.

While debugging some failing tests, I also added some logging to the circle config.


packages/nitro-protocol/test/contracts/ETHAssetHolder/deposit.test.ts
```diff
@@ -33,73 +30,66 @@ const description3 =
...
 
         // check for any partial refund
-        const gasCost = await tx.gasPrice.mul(receipt.cumulativeGasUsed);
-        await expect(
-          (await signer.getBalance()).eq(balanceBefore.sub(eventAmountDeposited).sub(gasCost)),
-        ).toBe(true);
+        // const gasCost = (await provider.getGasPrice()).mul(cumulativeGasUsed);
+        // await expect(await signer.getBalance()).toEqual(
+        //   balanceBefore.sub(event.amountDeposited).sub(gasCost),
+        // );
```
snario: Why comment out this check?

andrewgordstewart: When suites are run serially, each suite effectively has a lock on the signer. Now, this might be running along another test that is spending the signer's gas. The test should be rewritten to function independently of other tests.

geoknee:
Just wanted to confirm that on my machine I saw a 3-4x speedup with respect to serial tests. As far as I can tell, we're not yet seeing quite that dramatic an effect on Circle?

andrewgordstewart:
> Just wanted to confirm that on my machine I saw a 3-4x speedup on this branch, with respect to serial tests. As far as I can tell, we're not yet seeing quite that dramatic an effect on Circle?

In fact, the parallel tests seem to have _no_ effect on the speed of tests on circle. On top of that, jest seems to behave strange in lots of ways in a CI setting, since it can supposedly allocate resources according to the physical server's specs, rather than the virtual container's specs. Many people find running tests serially with `--runInband`

In spite of this, the change in the tests still seems to speed up the tests, by virtue of not having to wait ~2s (on average) for ethers to look for new events. `contract.on('someEventName', callback)` will cause ethers to poll new blocks every 4s, and look for new events since the last known block. The new method for testing events avoids this polling.

andrewgordstewart:
For comparison, [this run](https://circleci.com/gh/statechannels/monorepo/692) and [this run](https://circleci.com/gh/statechannels/monorepo/694) are running the same test script.

Run 692: `--runInBand` with old event testing
```
@statechannels/nitro-protocol: Test Suites: 23 passed, 23 total
@statechannels/nitro-protocol: Tests:       137 passed, 137 total
@statechannels/nitro-protocol: Snapshots:   0 total
@statechannels/nitro-protocol: Time:        331.452s
```

Run 694: `--runInBand` with new event testing
```
@statechannels/nitro-protocol: Test Suites: 23 passed, 23 total
@statechannels/nitro-protocol: Tests:       137 passed, 137 total
@statechannels/nitro-protocol: Snapshots:   0 total
@statechannels/nitro-protocol: Time:        160.716s
```

PR #66, opened 2019-09-27
Use tslint instead of prettier as the default check before tests run

snario:
Since we use [this plugin](https://github.com/prettier/tslint-plugin-prettier#tslint-plugin-prettier) `tslint` actually uses `prettier` under the hood, so these checks are more extensive. This also fixes an issue where the `wallet` was inheriting properties from a root `.prettierrc` which `web3torrent` and `embedded-wallet` are inheriting their linting rules from.

TODO: Standardizing our TypeScript code styles across all packages.


packages/nitro-protocol/package.json
```diff
@@ -35,7 +35,7 @@
     "build:ci": "yarn build",
     "test:ci": "yarn test",
     "test:typescript": "npx jest --testPathIgnorePatterns='test/contracts' --reporters='default'",
-    "test": "yarn prettier:check && yarn clearContracts && npx ganache-then-jest --runInBand",
+    "test": "yarn lint:check && yarn clearContracts && npx ganache-then-jest --runInBand",
```
cf19drofxots: Can we decouple linting from testing? They are separate concerns.

packages/wallet/package.json
```diff
@@ -15,12 +15,14 @@
     "build:typescript": "npx tsc",
     "build:webpack": "CI=false node scripts/build.js",
     "prepare": "npx tsc",
-    "test": "run-s prettier:check 'test:app --all'",
-    "test:ci": "GANACHE_PORT=8503 yarn run prettier:check && yarn run test:ci:contracts && yarn run test:ci:app && yarn build",
+    "test": "run-s lint:check 'test:app --all'",
```
cf19drofxots: Ditto earlier decoupling comment

packages/nitro-protocol/package.json
snario: Going to do this in a separate PR

packages/wallet/package.json
snario: Going to do this in a separate PR

PR #68, opened 2019-09-27
Use TypeScript references in wallet and nitro-protocol

snario:
Gives us some nice features inside VS Code such as Jump To Source:

### DIFF
--- a/packages/wallet/tsconfig.json
+++ b/packages/wallet/tsconfig.json
@@ -9,6 +9,7 @@
     "jsx": "react",
     "moduleResolution": "node",
     "declaration": true,
+    "declarationMap": true,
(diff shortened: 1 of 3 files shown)


packages/wallet/tsconfig.json
```diff
@@ -19,17 +21,15 @@
...
+  "include": [
+    "src/**/*",
+    "types/globals.d.ts",
+    "contracts/**/*.json"
+  ],
+  "references": [
+    { "path": "../nitro-protocol"}
+  ],
   "exclude": [
-    "node_modules",
```
cf19drofxots: All of these should still be excluded, no?

snario: They are excluded because they are not included except for `src/**/*`

PR #69, opened 2019-09-27
Update gitignore for logs

cf19drofxots:


snario:
I think there are already `.gitignore` files inside the sub packages

cf19drofxots:
There are, and I actually put this inside the `.gitignore` file of the `wallet` package, but seems like it's applicable for all packages unless for some reason we want to track local logs for some packages?

snario:
I would put it inside the local `.gitignore` since it is the only one that produces files like this as of now

cf19drofxots:
AFAICT the same script is used by the `devtools` package, so any package using `devtools` will end up producing these files as well, rendering the global coverage more useful, but I digress for now.

PR #70, opened 2019-09-27
[nitro-protocol] Introduce Interfaces with NatSpec

geoknee:
This PR introduces the following interfaces:

- `Adjudicator`
- `IAssetHolder`
- `IForceMove`*

*not actually an interface, but an abstract contract. See comment below.

The new inheritance graph looks like this:

image

image

The idea is to move toward a clearer separation between the specification of a contract and its implementation. 

A nice side effect is that the interfaces can be the main target of the NatSpec documentation, and this will get around some prior issues (see discussion in #33 ) with NatSpec-ing contracts with external imports (since interfaces are unlikely to rely on external contracts).


geoknee:
Note that `IForceMove` is an abstract contract and not an interface. This is because interfaces have fairly strict requirements which would require a fairly large refactor to satisfy. That refactor to be subject to another PR and separate discussion (it comes with downsides and may not be worth it).

PR #72, opened 2019-09-27
[documenation] Tweaks to the quick-start and state-format docs.

### LINKED ISSUE #71: [nitro-protocol] Docs Comments
Some comments that I am writing down after going through the Nitro Protocol docs a few times:

1. https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/docs/quick-start.md
"7. Either the dispute will be resolved and the channel continues (goto 5) or the channel is finalized. Otherwise if all participants agree to close the channel, it can be finalized more cheaply."
This sounds like: "Either x or y. Otherwise z?"
I think I now understand that `6` and `7` are tightly coupled as one paragraph. Maybe they should be bundled up into one list item? Maybe:
```
6. In the case of inactivity, participants may call forceMove on the adjudicator.
6.a Either the dispute will be resolved and the channel continues (goto 5) or the channel is finalized.
7. If forceMove is not called and all participants agree to close the channel, it can be finalized more cheaply than if forceMove had been called.
```

2. https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/docs/adjudicator/state-format.md#implementation-faqs
"Why include the ChannelId separately? We have to calculate the channelId for every single operation anyway. Given that we have it it's cheaper to hash in the hash, rather than the individual components again."
The second sentence is a bit confusing. I don't understand "cheaper to hash in the hash, rather than the individual components again"

geoknee:
Fixes #71.

### DIFF
--- a/packages/nitro-protocol/docs/adjudicator/state-format.md
+++ b/packages/nitro-protocol/docs/adjudicator/state-format.md
@@ -73,13 +73,13 @@ which contains fields which are allowed to change. These structs, along with rem
 
 ## Implementation FAQs
 
-**Why include the ChannelId separately?** We have to calculate the `channelId` for every single operation anyway. Given that we have it it's cheaper to hash in the hash, rather than the individual components again.
+**Why include the `channelId` in the State, rather than its pre-image?** It is necessary for the chain to calculate (and hence store in memory) the `channelId` for every state-changing method, in order to read the correct slot from storage. When computing the hash of a State struct (which often happens next), it is more efficient (in terms of gas consumption) to reuse `channelId`, rather than to (re)hash the pre-image of `channelId` along with the rest of the data. This is because `channelId` is generally shorter than its pre-image, and the gas costs of hashing increase with the length of the input.
 
-**Why not include the `AppDefinition` in the `ChannelId`?** The `AppDefinition` is fixed as part of the transition rules, so it seems like it would make sense to include it in the `ChannelId`. However, this would rule out participants being able to collaboratively upgrade the app without refunding the channel.
+**Why not include the `appDefinition` in the `channelId`?** The `appDefinition` is fixed as part of the transition rules, so it seems like it would make sense to include it in the `channelId`. However, this would rule out participants being able to collaboratively upgrade the app without refunding the channel.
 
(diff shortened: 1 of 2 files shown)


packages/nitro-protocol/docs/adjudicator/state-format.md
```diff
@@ -73,13 +73,13 @@ which contains fields which are allowed to change. These structs, along with rem
 
 ## Implementation FAQs
 
-**Why include the ChannelId separately?** We have to calculate the `channelId` for every single operation anyway. Given that we have it it's cheaper to hash in the hash, rather than the individual components again.
+**Why include the `channelId` in the State, rather than its pre-image?** It is necessary for the chain to calculate (and hence store in memory) the `channelId` for every state-changing method, in order to read the correct slot from storage. When computing the hash of a State struct (which often happens next), it is more efficient (in terms of gas consumption) to reuse `channelId`, rather than to (re)hash the pre-image of `channelId` along with the rest of the data. This is because `channelId` is generally shorter than its pre-image, and the gas costs of hashing increase with the length of the input.
```
Alonski: What do you mean by `pre-image` exactly? Is it the state from before? Maybe it is a term that I just don't know?
Other than that this looks much more readable :)

packages/nitro-protocol/docs/quick-start.md
```diff
@@ -14,9 +14,12 @@ How do I write a DApp (or convert an existing DApp) to run in a Nitro state chan
 4. Deposits (ETH and/or Tokens) are then made against the `channelId`, by interfacing with the relevant AssetHolder contracts.
 5. Participants exchange state updates and the default outcome is updated.
 6. In the case of inactivity, participants may call `forceMove` on the adjudicator
-7. Either the dispute will be resolved and the channel continues (goto 5) or the channel is finalized. Otherwise if all participants agree to close the channel, it can be finalized more cheaply.
-8. The outcome is pushed from the Adjudicator to the AssetHolders by calling `pushOutcome`.
-9. Funds are released from the AssetHolders.
+   - Either the dispute will be resolved and the channel continues (goto 5) or
+   - The challenge times out (goto 8).
+7. If all participants agree to close the channel, they may each sign an `isFinal = true` state and submit this via the `conclude` method.
```
Alonski: Much clearer! 💯

PR #74, opened 2019-09-27
[nitro-protocol] Convert IForceMove into an interface

### LINKED ISSUE #37: [nitro-protocol] Try to make forceMove methods external (rather than public)
These functions are only ever designed to be called externally.

There are also possible gas savings here, since the tx data can remain in `calldata` rather than being copied to `memory`.

However, you cannot store a `struct` in `calldata`.

A possible solution is to `abi.encode` the function parameters which are `struct`s. There will then be a gas cost associated with decoding the data -- I am not sure whether this will lead to an overall reduction or not.

geoknee:
In #70, the `ForceMove` contract extends an abstract contract `IForceMove`. This PR performs the necessary further work to make `IForceMove` an actual [`interface`](https://solidity.readthedocs.io/en/v0.5.11/contracts.html?highlight=interface#interfaces). 

The bulk of the work involved making all of the `ForceMove` public methods `external`. This seemed right, since these methods should probably only be called by external accounts and not by other methods in the same contract (the exception might be a future `batch` function that performs multiple actions in one transaction). Another benefit is that the function parameters remain in `calldata` (instead of being copied into `memory` as in a `public` method), hinting at reductions in gas consumption. 

However, a problem is that [you cannot use structs in `calldata`](https://github.com/ethereum/solidity/pull/5806).  I therefore had to encode much of the function parameters into a basic `bytes` type, and then decode them as the first step in the function implementation. There are at least two downsides to this: 

- There is now some *upwards* pressure on the gas consumption. 
- Debugging malformed input is likely more difficult.

The gas consumption appeared to go up overall: 

| Contract Name | Method Name              | Calls | Min Gas | Max Gas | Average Gas |
| ------------- | ------------------------ | ----- | ------- | ------- | ----------- |
| TESTForceMove | forceMove (currently)    | 12    | 64054   | 183871  | 112951      |
| TESTForceMove | forceMove (this branch)  | 12    | 65556   | 186221  | 114973      |
| TESTForceMove | conclude (currently)     | 11    | 56688   | 98347   | 79545       |
| TESTForceMove | conclude (this branch)   | 11    | 58420   | 100239  | 81363       |
| TESTForceMove | checkpoint (currently)   | 10    | 58890   | 107155  | 83740       |
| TESTForceMove | checkpoint (this branch) | 10    | 60340   | 108877  | 85337       |
| TESTForceMove | respond (currently)      | 5     | 61148   | 86421   | 70384       |
| TESTForceMove | respond (this branch)    | 5     | 63246   | 88576   | 72509       |

Fixes #37 .

On balance it seems not to be worth making this change, but I'm interested in what others think.

### DIFF
--- a/packages/nitro-protocol/.prettierignore
+++ b/packages/nitro-protocol/.prettierignore
@@ -1,3 +1,4 @@
 contracts/AssetHolder.sol
 contracts/NitroAdjudicator.sol
+contracts/ForceMove.sol
 contracts/examples/SingleAssetPayments.sol
--- a/packages/nitro-protocol/contracts/ForceMove.sol
+++ b/packages/nitro-protocol/contracts/ForceMove.sol
@@ -5,12 +5,50 @@ import './interfaces/IForceMove.sol';
 import './interfaces/ForceMoveApp.sol';
 
 contract ForceMove is IForceMove {
+    struct Signature {
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+    }
+
+    struct FixedPart {
+        uint256 chainId;
+        address[] participants;
+        uint256 channelNonce;
+        address appDefinition;
+        uint256 challengeDuration;
+    }
+
+    struct State {
+        // participants sign the hash of this
+        uint256 turnNum;
+        bool isFinal;
+        bytes32 channelId; // keccack(chainId,participants,channelNonce)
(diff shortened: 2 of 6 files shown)


tomclose:
It seems like the benefits don't outweigh the costs here. Going to close.

PR #81, opened 2019-09-27
Faster circle

andrewgordstewart:
### Description

Speeds up circleci by running tests only for packages developed on this feature branch.


.circleci/config.yml
```diff
@@ -90,6 +73,7 @@ jobs:
             - packages/*/build
             - packages/*/lib
             - packages/*/node_modules
+            - node_modules
```
andrewgordstewart: I'm not sure if this speeds up or slows down circle runs. Currently testing it.

andrewgordstewart: With this change: https://circleci.com/gh/statechannels/monorepo/789
Without: https://circleci.com/gh/statechannels/monorepo/790

I have no idea why restoring the cache takes two minutes. But it seems like it's not necessary to do it, so we're better off not doing it.

The downside to this change is that "Persisting to Workspace" takes longer with [this change](https://circleci.com/gh/statechannels/monorepo/786), vs [without this change](https://circleci.com/gh/statechannels/monorepo/787). This is on account of having to persist ~1gig of dependencies in `$PROJECT_ROOT/node_modules`. It seems like this tacks on 30s to one job, and cuts an unpredictable amount (up to two minutes?!?) from two jobs.

PR #82, opened 2019-09-27
Faster wallet tests

andrewgordstewart:
Speeds up wallet tests by using etherlime instead of truffle.


packages/wallet/deployment/deploy-test.js
```diff
@@ -0,0 +1,49 @@
...
+const trivialAppArtifact = require('../build/contracts/TrivialApp.json');
+
+let contractsToAddresses = {};
+
+async function getNetworkMap() {
+  try {
+    return await loadJsonFile(path.join(__dirname, 'network-map.json'));
+  } catch (err) {
+    if (!!err.message.match('ENOENT: no such file or directory')) {
+      return {};
```
snario: Maybe we should add a single `console.warn` to indicate that this file was not found?

andrewgordstewart:
```suggestion
      console.warn("network-map.json not found");
      return {};
```

andrewgordstewart:
> Does etherlime support writing those deploy scripts in typescript?

I could not find any support for typescript. Perhaps there's a way to get around this?

Issue #20, opened 2019-09-24
[wallet] AdjudicatorWatcher updates the adjudicator-state


tomclose:
Update the adjudicator watcher to handle the new events emitted by the nitro contracts, and update the adjudicator-state accordingly. 

### Notes

The adjudicator watcher is responsible for subscribing to events from the provider, and triggering actions that are consumed by the adjudicator-state-reducer. The adjudicator-state-reducer updates the adjudicator-state, a cached version of the adjudicator state, restricted to channels of relevance to the wallet. The wallet declares the channels it is interested in via the [channel subscriptions](https://github.com/magmo/apps/blob/master/packages/wallet/src/redux/state.ts#L68) and the adjudicator-watcher should only trigger events for channels that appear in this list.

The adjudicator-state will need to be updated to the following format:
```ts
export interface AdjudicatorChannelState {
  channelId: string;
  balance: string; // leave for time being
  finalized: boolean; // see note below
  challenge?: Challenge;
  finalizesAt: string; // or bignumber ??
  outcomeBytes: string;
  challengerAddress: Address;
}
```
Note: we'll keep the `finalized` property, even though it's redundant information, to keep the current workflow with the `ChallengeWatcher`: the `ChallengeWatcher` will observe open challenges, and trigger an event when they timeout, which will flip this property to `true`. That way reducers (a) don't have to do time comparisons themselves, and (b) receive an action to initiate the next steps.

### Testing

The [current testing strategy](https://github.com/magmo/apps/blob/master/packages/wallet/src/contract-tests/adjudicator-events.test.ts) seems pretty good. It would be nice to hook the reducer in (as in the full example [here](https://github.com/wix/redux-saga-tester#full-example)), so that we can do a full integration tests of the form:

1. Trigger action on the contract.
2. Check to see that `adjudicator-state` is as expected.

We should add tests for all of the methods on the nitro-adjudicator. You'll probably want to import the nitro-test-contracts, so you can use [this approach](https://github.com/statechannels/nitro-protocol/blob/master/test/contracts/ForceMove/checkpoint.test.ts#L144) for putting the on-chain state in the right state before the test (if necessary).

### Scope
* We'll limit the scope to the public methods on the force-move contract: `force-move`, `respond`, `refute`, `checkpoint`, and `conclude`
* The `deposit` event is out of scope for this work, but will be covered later as part of the `AssetHolderWatcher` work.
[...]

PR #83, opened 2019-09-27
[Wallet] - Handle Challenge Registered Event

lalexgap:
### Description
Updates the adjudicator event listener to handle the `ChallengeRegistered` event from the nitro-protocol `ForceMove` contract.

### Related issues
#20


packages/nitro-protocol/src/contract/challenge.ts
```diff
@@ -1,9 +1,50 @@
...
   );
 }
+
+export interface ChallengeRegisteredEvent {
+  challengerAddress: string;
+  finalizesAt: string;
+  challengeStates: State[];
+}
+export function getChallengeRegisteredEvent(eventArgs): ChallengeRegisteredEvent {
+  // TODO: There must be a better way of parsing out the event
```
andrewgordstewart: `event.args` is an array with properties. So, something like
```
const { turnNumRecord, finalizesAt } = eventArgs
```
should work, as long as `eventArgs === event.args` for some `event` fetched by ethers.

packages/nitro-protocol/src/index.ts
```diff
@@ -22,4 +23,14 @@ export interface ChannelStorage {
...
+  createDepositTransaction,
+  State,
+  encodeConsensusData,
+  Outcome,
+  AllocationItem,
+  Channel,
+  getChallengeRegisteredEvent,
+  isAllocationOutcome,
+  isGuaranteeOutcome,
+};
```
snario: It's worth grouping these exported symbols into something like (a) `utils` and (b) `types`

packages/nitro-protocol/src/contract/challenge.ts
snario: Based on the [docs](https://docs.ethers.io/ethers.js/html/api-contract.html#event-object) I think that `eventArgs[eventArgs.length - 1]` is an object with the key being the key name in the event.

e.g.,

```
const turnNumRecord = eventArgs[eventArgs.length - 1].turnNumRecord;
```

lalexgap: It looks like the last argument passed into the listener from `NitroAdjudicator.on() is the the actual event, so we could pull the `args` off from there.  Still slightly annoying that we have to manually parse `FixedPart`/`VariablePart`

packages/nitro-protocol/src/contract/challenge.ts
```diff
@@ -1,9 +1,57 @@
...
+    fixedPart,
+    variableParts: variablePartsUnstructured,
+  } = event.args;
+
+  // Fixed part
+  const chainId = bigNumberify(fixedPart[0]).toHexString();
+  const participants = fixedPart[1].map(p => bigNumberify(p).toHexString());
+  const channelNonce = bigNumberify(fixedPart[2]).toHexString();
+  const appDefinition = fixedPart[3];
+  const challengeDuration = bigNumberify(fixedPart[4]).toNumber();
```
snario: This decoding will certainly be getting re-used in the handlers for other events. I would recommend stripping it out into a short re-usable function.

packages/nitro-protocol/src/contract/state.ts
```diff
@@ -12,16 +12,14 @@ export interface State {
...
-export function getFixedPart(
-  state: State,
-): {
+export interface FixedPart {
   chainId: Uint256;
   participants: Address[];
   channelNonce: Uint256;
   appDefinition: Address;
   challengeDuration: Uint48;
-} {
```
snario: Could we try to organize types of this nature into logical buckets? Perhaps `FixedPart` should be within `@statechannels/nitro-protocol` because it is a contract-level structure that happens to bleed a little bit into the client code.

packages/wallet/package.json
```diff
@@ -110,6 +110,15 @@
...
     "@types/storybook__react": "4.0.0",
+    "@types/lodash": "4.14.129",
+    "@types/node": "^10.5.4",
+    "@types/prop-types": "^15.5.4",
+    "@types/react": "16.9.2",
+    "@types/react-dom": "^16.0.6",
+    "@types/react-redux": "^6.0.5",
+    "@types/react-sidebar": "3.0.0",
+    "@types/reactstrap": "^6.0.3",
+    "@types/web3": "^1.0.2",
```
snario: I think this is from #136?

packages/wallet/src/contract-tests/adjudicator-events.test.ts
```diff
@@ -1,5 +1,5 @@
-import {adjudicatorWatcher} from "../redux/sagas/adjudicator-watcher";
-import {ethers} from "ethers";
+import { adjudicatorWatcher } from "../redux/sagas/adjudicator-watcher";
+import { ethers } from "ethers";
```
snario: Careful of linting issues that are getting added to this changeset.

packages/wallet/src/contract-tests/adjudicator-events.test.ts
```diff
@@ -9,17 +9,19 @@ import {
...
   respondWithMove,
   getChannelId,
-  defaultDepositAmount
+  defaultDepositAmount,
 } from "./test-utils";
 import * as walletStates from "../redux/state";
-import {getGanacheProvider} from "@statechannels/devtools";
+import { getGanacheProvider } from "@statechannels/devtools";
+import { convertCommitmentToState } from "../utils/nitro-converter";
+// import { convertCommitmentToState } from "../utils/nitro-converter";
```
snario:
```suggestion
```

packages/wallet/src/contract-tests/transactions.test.ts
```diff
@@ -165,7 +165,7 @@ describe("transactions", () => {
       channel,
       allocation: ["0x05", "0x05"],
       destination: [participantA.address, participantB.address],
-      turnNum: 6,
+      turnNum: 8,
```
snario: I've noticed this in my work on #143 and #152 that these numbers need to be exactly right across these test files and utils. It would be worth spending a bit of time to make selecting them idiot-proof.

packages/wallet/src/redux/actions.ts
```diff
@@ -128,68 +130,68 @@ export interface ChannelUpdate {
...
 export const concludedEvent: ActionConstructor<ConcludedEvent> = p => ({
   ...p,
-  type: "WALLET.ADJUDICATOR.CONCLUDED_EVENT"
+  type: "WALLET.ADJUDICATOR.CONCLUDED_EVENT",
 });
 
 export const refutedEvent: ActionConstructor<RefutedEvent> = p => ({
   ...p,
-  type: "WALLET.ADJUDICATOR.REFUTED_EVENT"
+  type: "WALLET.ADJUDICATOR.REFUTED_EVENT",
```
snario: Similarly, linting is bleeding into this PR.

packages/wallet/src/redux/adjudicator-state/__tests__/reducer.test.ts
```diff
@@ -1,18 +1,34 @@
...
+    channelNonce: "0x01",
+    participants: [],
+  },
+  appDefinition: ethers.Wallet.createRandom().address,
+  appData: "0x0",
+  challengeDuration: 5,
+  outcome: [],
+};
+const gameCommitment1 = testScenarios.appCommitment({ turnNum: 19 }).commitment;
+// tslint:disable-next-line: no-shadowed-variable
```
snario: Might as well fix this issue

packages/wallet/src/redux/sagas/adjudicator-watcher.ts
```diff
@@ -41,123 +37,113 @@ export function* adjudicatorWatcher(provider) {
...
+    //     })
+    //   );
+    //   break;
+    // case AdjudicatorEventType.RespondWithMove:
+    //   const { v, r, s } = event.eventArgs;
+    //   const signature = ethers.utils.joinSignature({
+    //     v,
+    //     r,
+    //     s,
+    //   });
```
snario: Remove commented out code?

PR #90, opened 2019-09-28
[wallet] - Correctly set port

lalexgap:
### Description

Correctly sets the `GANACHE_HOST` env variable in the `test:ci` command. Prior to this it wasn't being set correctly, and defaulting to 8545. Which means the `wallet` package could end up using a `ganache` instance started by another package.


lalexgap:
Obsoleted by #82

PR #95, opened 2019-09-28
Parallel lerna tests

andrewgordstewart:
This PR refactors the deploy and test scripts to allow the wallet and nitro-adjudicator tests to allow them to run in parallel. As all circleCI containers have at least two vCPUs, not running them in parallel is wasting resources (= time)

The primary difficulty in running the tests in parallel is the interaction with a ganache server. If `packageA` and `packageB` are both using the same signer for the same ganache server, then ethers will most likely submit two transactions from that account to the one server with the same nonce. This PR solves the problem by starting a different ganache server for each package -- each of them manages their own port.

There is some duplication of code in the deploy scripts between two packages. This should be considered in the future.


packages/wallet/package.json
```diff
@@ -10,29 +10,30 @@
...
     "lint:check": "npx tslint --config tslint.json --project .",
     "lint:write": "npx tslint --config tslint.json --project . --fix",
+    "postinstall": "rm -f node_modules/web3/index.d.ts",
+    "prepare": "yarn build:typescript",
     "prettier:check": "npx prettier --check 'src/**/*.{ts,tsx}'",
     "prettier:write": "npx prettier --write 'src/**/*.{ts,tsx}'",
-    "startGanache": "npx etherlime ganache",
-    "postinstall": "rm -f node_modules/web3/index.d.ts",
+    "start": "npm run clearContracts && node scripts/start.js",
+    "startGanache": "npx etherlime ganache --port=8546 > ganache.log",
```
lalexgap: We should probably have a simple script that reads in `GANACHE_PORT` and starts `etherlime ganache` so `startGanache` will work when GANACHE_PORT is not 8546

PR #118, opened 2019-09-28
[Wallet] - Etherlime fixed network id

lalexgap:
Gets etherlime working with the fixed `CHAIN_NETWORK_ID` and change `getLibraryAddress` to be a sync method.

This allows us to treat contract addresses as constants as we were before.

### DIFF
--- a/packages/wallet/src/contract-tests/test-utils.ts
+++ b/packages/wallet/src/contract-tests/test-utils.ts
@@ -1,67 +1,38 @@
-import {ethers} from "ethers";
-import {CommitmentType, Commitment, signCommitment2} from "../domain";
+import { ethers } from "ethers";
+import { CommitmentType, Commitment, signCommitment2 } from "../domain";
 import {
   createDepositTransaction,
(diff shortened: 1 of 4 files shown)


packages/wallet/src/contract-tests/test-utils.ts
```diff
@@ -1,67 +1,38 @@
...
-      throw err;
-    }
-  }
-};
-
 export async function getChannelId(provider, channelNonce, participantA, participantB) {
   const network = await provider.getNetwork();
   const networkId = network.chainId;
-  const libraryAddress = await getLibraryAddress(networkId);
+  const libraryAddress = await getLibraryAddress(networkId, "TrivialApp");
```
andrewgordstewart: nit: I wonder if thous should be like `TrivialAppArtifact.contractName`

lalexgap: I think the advantage of this is that the typescript can be compiled without the artifacts being present.

PR #133, opened 2019-09-29
[wallet] Handle ChallengeCleared event from Nitro-Protocol's ForceMove contract

cf19drofxots:
Some cosmetic changes got included here.


snario:
Is this in accordance with the linter?

cf19drofxots:
> Is this in accordance with the linter?

The lint check passed so...yeah?

snario:
Could we split up the cosmetic changes into a separate PR?

snario:
I've made a new PR which consolidates linting: #139

packages/wallet/src/redux/actions.ts
```diff
@@ -79,6 +79,13 @@ export interface ChallengeCreatedEvent {
   challengeStates: State[];
 }
 
+export interface ChallengeClearedEvent {
+  type: "WALLET.ADJUDICATOR.CHALLENGE_CLEARED_EVENT";
+  channelId: string;
+
+   newTurnNumRecord: number;
+}
```
snario: Weird formatting  ^

packages/wallet/src/redux/sagas/adjudicator-watcher.ts
```diff
@@ -36,9 +36,9 @@ export function* adjudicatorWatcher(provider) {
 }
 
 function* dispatchEventAction(event: AdjudicatorEvent) {
+  const { channelId } = event;
```
snario: May as well destructure `eventType` here too

packages/wallet/src/redux/sagas/adjudicator-watcher.ts
```diff
@@ -52,6 +52,15 @@ function* dispatchEventAction(event: AdjudicatorEvent) {
         })
       );
       break;
+    case AdjudicatorEventType.ChallengeCleared:
+      const newTurnNumRecord = event.eventArgs[1].toNumber();
```
snario: You should be able to do `event.args.newTurnNumRecord`

packages/wallet/src/redux/sagas/adjudicator-watcher.ts
cf19drofxots: I tried but looks like `eventType` doesn't have a specific structure, it's just an array of values which is why elsewhere in the code I had to use indexing to access its elements.

packages/wallet/src/redux/actions.ts
cf19drofxots: Right, lack of linting.

packages/wallet/src/redux/sagas/adjudicator-watcher.ts
cf19drofxots: `args` is not a property of `event` though.

PR #135, opened 2019-09-29
Separate build process in circle

andrewgordstewart:
### Description

Since `packageA` is linked to `packageB` when `packageA` depends on `packageB`, some of the current build process is needed for tests to run in `packageA`. An example of this is the `wallet` package depending on `nitro-protocol`.

However, some of the build process is not needed to run tests, such as `build:webpack` in the `wallet` package. This script takes over 30s on my laptop. The goal of running `yarn build:webpack` in a circleci build is to make sure that the current changes don't break the app in a way not covered by our test suite.

This PR separates the build commands required by dependencies from commands building the project for production. The former go in `prepare`, and are run in the `prepare` job on circle. The latter go in `build:ci`, and go into the `build` job on circle. `build:ci` is only run on packages that have changed in the current branch.

It seems to save about 30s per build.

Open to alternative script names. One idea:
- `yarn build:ci` to prepare the circle workspace
- `yarn build:production` to test the production build sequence

### DIFF
--- a/packages/nitro-protocol/package.json
+++ b/packages/nitro-protocol/package.json
@@ -41,7 +41,7 @@
     "deploy": "npx etherlime deploy --compile=false",
     "lint:check": "npx tslint --config tslint.json --project .",
     "lint:write": "npx tslint --config tslint.json --project . --fix",
-    "prepare": "yarn build",
+    "prepare": "yarn build:typescript",
     "prettier:check": "npx prettier --check '{contracts,src,test}/**/*.{ts,tsx,sol}'",
     "prettier:write": "npx prettier --write '{contracts,src,test}/**/*.{ts,tsx,sol}'",
     "startGanache": "node bin/start-ganache.js",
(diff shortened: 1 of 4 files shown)


packages/nitro-protocol/package.json
```diff
@@ -32,7 +32,6 @@
   "homepage": "https://github.com/statechannels/nitro-protocol/",
   "scripts": {
     "build": "run-s clearContracts clearLib contract:compile build:typescript",
-    "build:ci": "yarn build",
```
snario: As a note, the reason why I think this can be skipped is that `wallet` uses `tsc -b` which uses TypeScript references to auto-run `tsc -b` in the `nitro-protocol` package.

andrewgordstewart: All packages that are dependents need to be built in the `prepare` step, so `nitro-protocol` isn't out of date once circle is running the `build` job. Is the `-b` needed considering this?

PR #136, opened 2019-09-29
Run typesync to add missing types

snario:
Runs [typesync](https://github.com/jeffijoe/typesync).


packages/wallet/src/contract-tests/adjudicator-events.test.ts
```diff
@@ -91,7 +91,7 @@ describe.skip("adjudicator listener", () => {
     await depositContract(provider, channelId);
     await sagaTester.waitFor("WALLET.ADJUDICATOR.FUNDING_RECEIVED_EVENT");
 
-    const action: actions.FundingReceivedEvent = sagaTester.getLatestCalledAction();
+    const action = sagaTester.getLatestCalledAction() as actions.FundingReceivedEvent;
```
andrewgordstewart: Previously, typescript could know that `getLatestCalledAction` would return something that is of type `FundingReceivedEvent`, and the compiler could check that. The change here is casting the return value to be of type `FundingReceivedEvent` regardless of what type `getLatestCalledAction` can return.

Is this intentional? It seems like you lose type safety by doing this.

snario: It seems as though the types package was upgraded to remove that feature? It is expecting that `sagaTester.getLatestCalledAction();` is of type  `AnyAction`.

tomclose:
Closing as this is old. Feel free to resubmit, when the tests pass!

PR #139, opened 2019-09-30
Consolidate all tslint configuration files into one

snario:
This standardizes our linting across all `.tsx?` files in the repo.


joelalejandro:
Also, what about using non-js files for this, since it isn't actually code. The `.prettierrc` syntax we have on `web3torrent` / `embedded-wallet` looks nicer to implement and the `extends` rule mimics that of `tsconfig.json`.

snario:
Okay @joelalejandro I have changed this branch so that now what it does it consolidates all the linting configurations but changes the absolute minimum source code. Thus, from here we can slowly remove all the special case rules each package has by simply removing the rule from the config file and running `yarn lint:write` on a per package basis.

packages/wallet/.prettierrc.js
```diff
@@ -0,0 +1,8 @@
+// NOTE: This package has inconsistent use of trailing commas and bracket spacing
+
+module.exports = {
+    ...require('../../.prettierrc.js'),
+    "printWidth": 120,
+    "singleQuote": false,
+    "bracketSpacing": false,
```
cf19drofxots: What's the reason for using these values?

I, for one, prefer a shorter width (80) and that we have `bracketSpacing` on.

cf19drofxots:
> Thus, from here we can slowly remove all the special case rules each package has by simply removing the rule from the config file and running yarn lint:write on a per package basis.

Ignore my previous comment.

packages/wallet/.prettierrc.js
snario: Me too; this is just how it is now.

PR #140, opened 2019-09-30
Remove linting from test scripts.

snario:
Seems unnecessary to be linting on the CI and every time we run a test script.

I think it might be wiser to make linting a pre-commit hook instead.

### DIFF
--- a/packages/nitro-protocol/package.json
+++ b/packages/nitro-protocol/package.json
@@ -31,7 +31,7 @@
   },
   "homepage": "https://github.com/statechannels/nitro-protocol/",
   "scripts": {
-    "precommit": "lint-staged",
+    "precommit": "lint-staged --quiet",
     "build": "run-s clearContracts clearLib contract:compile build:typescript",
     "build:typescript": "tslint --project . && tsc -b",
     "clearContracts": "rm -rf build/contracts",
--- a/packages/nitro-protocol/package.json
+++ b/packages/nitro-protocol/package.json
@@ -31,6 +31,7 @@
   },
   "homepage": "https://github.com/statechannels/nitro-protocol/",
   "scripts": {
+    "precommit": "lint-staged",
     "build": "run-s clearContracts clearLib contract:compile build:typescript",
     "build:typescript": "tslint --project . && tsc -b",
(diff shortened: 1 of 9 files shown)


Alonski:
We can use Husky + lint-staged to setup git commit hooks for the repo.
Saying that, I think that linting is something that should run on CI and not on every commit.
If during the course of coding I make 5 commits and linting takes 2 minute each time that's 10 minutes of dev time.
If CI does it it's 2 minutes of robot time...

packages/nitro-protocol/package.json
```diff
@@ -77,6 +79,9 @@
     "typescript": "3.6.3",
     "write-json-file": "4.2.0"
   },
+  "lint-staged": {
+    "{src,test}/**/*.tsx?": "yarn lint:check"
```
Alonski: Should this be `tsx` or `ts`?

andrewgordstewart:
In practice, linting doesn't add any time to circleci runs, as it's not the rate-limiting step. The whole job takes less than 30s After [this](https://github.com/statechannels/monorepo/pull/135) has been merged, we could move the `lint` check to the `build:ci` job, and we wouldn't have to spin up a container just to run `yarn lint`. Would you be opposed to keeping it in CI in that context?

andrewgordstewart:
I just tried this out locally, and I would personally find it more annoying to wait 5s on each local commit compared to waiting 5s on each circleci run. At the same time, having `lint:check` (or some other similar command) run in a pre-commit hook seems like a good idea, since it means the git history always conforms to some set of rules.

packages/nitro-protocol/package.json
andrewgordstewart: `tsx?` matches both `ts` and `tsx`. Are you asking if we shouldn't lint `tsx` files?

andrewgordstewart:
Oh, I completely misunderstood the intention of the PR until just now. I see that linting is still checked in circle:ci, it's just not checked _twice_.

I've made [another PR](https://github.com/statechannels/monorepo/pull/146) branched from this, which combines `yarn lint:ci` with `yarn build:ci`, and speeds up the pre-commit hook.

PR #148, opened 2019-09-30
Install solidity-docgen and hook into docs website

tomclose:
A second attempt at merging the work in https://github.com/statechannels/monorepo/pull/75. (That PR was onto a branch that had already been merged ..)


packages/nitro-protocol/docs/interfaces/Adjudicator.md
```diff
@@ -0,0 +1,33 @@
+---
```
andrewgordstewart: Since the docs are generated from the code on `yarn build`, we can ignore `packages/nitro-protocol/docs` (or maybe `packages/nitro-protocol/docs/interfaces`?)

PR #152, opened 2019-10-01
Support refute transactions

snario:
Adds calls to `checkpoint` and tests pass


snario:
I agree we ought to remove it, however phase 1 of our plan with @lalexgap is to get the contract tests passing first and then following that refactoring the protocols to have updated concepts and data structures

PR #153, opened 2019-10-01
Get conclude transaction tests passing

snario:


packages/wallet/src/contract-tests/test-utils.ts
```diff
@@ -97,22 +97,22 @@ export async function concludeGame(provider: ethers.providers.JsonRpcProvider, c
 
   const fromCommitment: Commitment = {
     channel,
-    allocation: fiveFive,
+    allocation: ["0x05", "0x05"],
```
cf19drofxots: I'd suggest declaring a const allocation the top, something like
```
const ALLOCATION = ["0x05", "0x05"]
```

and reusing that everywhere instead of having string literals all over.

PR #156, opened 2019-10-01
[wallet] Type action names

cf19drofxots:
This is the first of a series of PRs that replaces the usage of string literals to refer to action types of various kinds across the wallet.

### DIFF
--- a/packages/wallet/src/redux/actions.ts
+++ b/packages/wallet/src/redux/actions.ts
@@ -29,6 +29,8 @@ export enum WalletActionType {
   METAMASK_LOAD_ERROR = "METAMASK_LOAD_ERROR",
   WALLET_ADJUDICATOR_CHALLENGE_EXPIRY_TIME_SET = "WALLET.ADJUDICATOR.CHALLENGE_EXPIRY_TIME_SET",
   WALLET_ADJUDICATOR_CHALLENGE_CREATED_EVENT = "WALLET.ADJUDICATOR.CHALLENGE_CREATED_EVENT",
+  WALLET_ADJUDICATOR_CONCLUDED_EVENT = "WALLET.ADJUDICATOR.CONCLUDED_EVENT",
+  WALLET_ADJUDICATOR_REFUTED_EVENT = "WALLET.ADJUDICATOR.REFUTED_EVENT",
   WALLET_MULTIPLE_ACTIONS = "WALLET.MULTIPLE_ACTIONS",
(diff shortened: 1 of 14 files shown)


tomclose:
I see that you're replacing string literals with enums here. It isn't completely clear to me what the benefits are here, given that we already get typesafety on those string literals, and that the change adds to the verbosity of the code. Would you be able to explain the motivation behind making this change across the codebase?

cf19drofxots:
Can you explain how we already get type safety on those string literals?

Having the same string literal be used in multiple places also generally goes against the DRY principle, and reducing the redundancy is fairly trivial.

packages/wallet/src/redux/actions.ts
```diff
@@ -111,15 +129,15 @@ export interface FundingReceivedEvent {
...
+  type: WalletActionType.WALLET_ADJUDICATOR_CHALLENGE_EXPIRED;
   processId: string;
   protocolLocator: ProtocolLocator;
   channelId: string;
   timestamp: number;
 }
 
 export interface ChannelUpdate {
-  type: "WALLET.ADJUDICATOR.CHANNEL_UPDATE";
+  type: WalletActionType.WALLET_ADJUDICATOR_CHANNEL_UPDATE;
```
andrewgordstewart: The type of `type` in this interface is not `string`. It is `"WALLET.ADJUDICATOR.CHANNEL_UPDATE"`. Therefore, `"WALLET.ADJUDICATOR.CHANNEL_UPDATE" ` is the only allowable value of `o.type` for an object `o` of type `ChannelUpdate`.

As an example, Typescript uses this fact in a union type, and infers that when `o` is of type `ChannelUpdate | AdjudicatorKnown`, then `o.type` is of type `"WALLET.ADJUDICATOR.CHANNEL_UPDATE" | "WALLET.ADJUDICATOR_KNOWN"`. An editor like VSCode can use this type information to provide autocomplete, for instance, and to display type errors in such union types.

I'm not aware of any additional type safety introduced by replacing the type of `type` with an enum. Is there something we're unaware of?

cf19drofxots: The type of `type` is not a string but it is a using a string value to represent the type. The rationale for this change was to introduce a common reference for this string value that's being used to define the type of an object, such that those who don't get these inferences (and therefore, autocomplete) have a common reference and are less prone to running into build issues.

But all of this seems like it's coming at too large of an expense, so I digress and we can close this.

PR #165, opened 2019-10-02
[wallet] Handle conclude event

cf19drofxots:


packages/wallet/src/contract-tests/test-utils.ts
```diff
@@ -109,7 +109,7 @@ export async function concludeGame(provider: ethers.providers.JsonRpcProvider, c
     channel,
     allocation: fiveFive,
     destination: [participantA.address, participantB.address],
-    turnNum: 6,
+    turnNum: 7,
```
snario: Double check that `yarn test:ci:contracts` still passes with this change. I had to handle this in #161

packages/wallet/src/redux/sagas/adjudicator-watcher.ts
```diff
@@ -64,6 +64,9 @@ function* dispatchEventAction(event: AdjudicatorEvent) {
         })
       );
       break;
+    case AdjudicatorEventType.Concluded:
+      yield put(actions.concludedEvent({channelId}));
+      break;
```
snario: Shouldn't there be a `default:` here?

cf19drofxots: There should be - what's an appropriate default for this switch block?

packages/wallet/src/contract-tests/test-utils.ts
cf19drofxots: It does

PR #168, opened 2019-10-02
[wallet] Introduce the ETHAssetHolderWatcher saga

cf19drofxots:
Relevant issue: https://github.com/statechannels/monorepo/issues/22

This PR only focuses on watching for the `DEPOSITED` event (`ASSET_TRANSFERRED` watching will be in a follow up PR) while introducing specifically the `ETHAssetHolderWatcher` to keep the scope tight.

Following iterations will either update the code to have a generic `AssetHolderWatcher` or watchers specific to asset types (discussion to be had).


tomclose:
Looking good so far. My guess is that you'll be able to adapt this into a generic `AssetHolderWatcher` fairly easily, when the time comes!

cf19drofxots:
> My guess is that you'll be able to adapt this into a generic AssetHolderWatcher fairly easily

Correct

PR #175, opened 2019-10-03
[wallet] Watch AssetTransferred event

cf19drofxots:
Continuation of work done at https://github.com/statechannels/monorepo/pull/168


packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -13,6 +13,14 @@ contract AssetHolder is IAssetHolder {
 
     mapping(bytes32 => bytes32) public outcomeHashes;
 
+    function getHoldings(bytes32 channelId)
```
cf19drofxots: For debugging purposes.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -107,6 +115,8 @@ contract AssetHolder is IAssetHolder {
                 emit AssetTransferred(allocation[m].destination, payoutAmount);
             } else {
                 holdings[allocation[m].destination] += payoutAmount;
+                // TODO: Remove this as it's just for debugging purposes
```
cf19drofxots: For debugging purposes.

cf19drofxots:
@andrewgordstewart @tomclose can you double check my sanity here. The following is the log of the allocations used for transferring some assets.

```
    console.log src/contract-tests/eth-asset-holder-events.test.ts:168
      {
          "0": "0x0000000000000000000000005b04e633bc7dd59ec1df4f33fb2c6c4706a24065",
          "1": {
              "_hex": "0x05"
          },
          "destination": "0x0000000000000000000000005b04e633bc7dd59ec1df4f33fb2c6c4706a24065",
          "amount": {
              "_hex": "0x05"
          },
          "length": 2
      }
    console.log src/contract-tests/eth-asset-holder-events.test.ts:168
      {
          "0": "0x000000000000000000000000d2ba45142fcc0cc336a9dfc90d810c749a75e593",
          "1": {
              "_hex": "0x05"
          },
          "destination": "0x000000000000000000000000d2ba45142fcc0cc336a9dfc90d810c749a75e593",
          "amount": {
              "_hex": "0x05"
          },
          "length": 2
      }
```

I got this by placing an event emission on [this line](https://github.com/statechannels/monorepo/pull/175/files#diff-25022c179254001069801373d122f301R118), whereas the correct event emission is supposed to happen in the block right above it.

This seems to indicate that the `_isExternalAddress` call is returning false, making me confused because the `destination` values logged above seem like they're left-padded external addresses.

Any idea what I'm overlooking here?

cf19drofxots:
So based on further investigation, it looks like `bytes32(bytes20(destination)) == destination` is an incorrect check.

Below is the event emission of a `destination` before it's been casted and after it's been casted via `bytes32(bytes20())` and it looks like the wrong side is being chopped off during the `bytes20` cast, leading to the check of `_isExternalAddress` failing.
```
    console.log src/contract-tests/eth-asset-holder-events.test.ts:132
      {
          "0": "0x000000000000000000000000aee50d91e6eef94235608a96758b6bc2f2c1c5e5",
          "1": {
              "_hex": "0x05"
          },
          "destination": "0x000000000000000000000000aee50d91e6eef94235608a96758b6bc2f2c1c5e5",
          "amount": {
              "_hex": "0x05"
          },
          "length": 2
      }
    console.log src/contract-tests/eth-asset-holder-events.test.ts:132
      {
          "0": "0x000000000000000000000000aee50d91e6eef942000000000000000000000000",
          "1": {
              "_hex": "0x05"
          },
          "destination": "0x000000000000000000000000aee50d91e6eef942000000000000000000000000",
          "amount": {
              "_hex": "0x05"
          },
          "length": 2
      }
```

tomclose:
@cf19drofxots thanks for looking into this, and apologies for not responding sooner. I think the decision to use `bytes32(bytes20())` was a conscious one, and was meant to be coupled with changing the definition of an external address, so that it is right-padded with zeros, instead of left-padded. The tests for `_isExternalAddress` support this: https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/test/contracts/AssetHolder/isExternalAddress.test.ts#L26.

snario:
@tomclose and I noticed a discrepancy in the code. The helper function in `nitro-converter.ts` uses left-padding (I changed it to do this actually, not realizing the right-pad was deliberate by @geoknee). The test case in `isExternalAddress.test.ts` checks for right-padding though.

Also, the reason @geoknee chose to use right padding was because he thought it was easier to implement using type casting in Solidity. However, here is a counterpoint to that argument:

```diff
diff --git i/packages/nitro-protocol/contracts/AssetHolder.sol w/packages/nitro-protocol/contracts/AssetHolder.sol
index f828ba078..6a85ef771 100644
--- i/packages/nitro-protocol/contracts/AssetHolder.sol
+++ w/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -283,15 +283,15 @@ contract AssetHolder is IAssetHolder {
     function _transferAsset(address payable destination, uint256 amount) internal {}
 
     function _isExternalAddress(bytes32 destination) internal pure returns (bool) {
-        return (destination == bytes32(bytes20(destination)));
+        return (destination == bytes32(uint256(destination)));
     }
 
     function _addressToBytes32(address participant) internal pure returns (bytes32) {
-        return bytes32(bytes20(participant));
+        return bytes32(uint256(participant));
     }
 
     function _bytes32ToAddress(bytes32 destination) internal pure returns (address payable) {
-        return address(bytes20(destination));
+        return address(uint160(uint256(destination)));
     }
 
     function recoverSigner(bytes memory _d, uint8 _v, bytes32 _r, bytes32 _s)
```

Applying this diff shows that you can do left padding with type casting.

I think we should standardize around left-padding as this is the typical way Ethereum handles padding in similar situations to this one.

cf19drofxots:
The relevant tests for this feature now pass as of the latest commit.

Though it looks like some non-related tests are failing as of merging master into this branch. Waiting on an update from @andrewgordstewart for possible comments on this behavior

packages/wallet/src/redux/actions.ts
```diff
@@ -53,6 +53,15 @@ export interface DisplayMessageSent {
   type: "WALLET.DISPLAY_MESSAGE_SENT";
 }
 
+export interface AssetTransferredEvent {
+  type: "WALLET.ASSET_HOLDER.ASSET_TRANSFERRED";
+
+  // This is either the `channelId` (bytes32) or (the left-padded bytes20)
+  // recipient address
```
geoknee: Nit: this should *always* be the left-padded recipient address (now known as an external destination). The event is intended to be emitted whenever [`_transferAsset`](https://github.com/statechannels/monorepo/blob/85ea52a16ebb36d024465c024fb3ecf91011aec0/packages/nitro-protocol/contracts/ETHAssetHolder.sol#L46) is called, and that function only accepts an `address` (20 bytes) for the destination. 

The dev-comments hopefully make it clear: 
https://github.com/statechannels/monorepo/blob/95c18a3d7de10d1a7d092d147bde0035dce8000d/packages/nitro-protocol/contracts/interfaces/IAssetHolder.sol#L43

Thinking about it, however, I suggest we change the event to emit the `address` (20 bytes) itself rather than the padded `bytes32`. This will require an additional call to `_bytes32ToAddress`, but in fact we will always have called that when preparing the call to ` _transferAsset`, so we can we can just keep the output in memory. This would hopefully remove any remaining ambiguity. Thoughts?

andrewgordstewart:
> @cf19drofxots thanks for looking into this, and apologies for not responding sooner. I think the decision to use `bytes32(bytes20())` was a conscious one, and was meant to be coupled with changing the definition of an external address to be an address that is right-padded with zeros, instead of left-padded. The tests for `_isExternalAddress` support this: https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/test/contracts/AssetHolder/isExternalAddress.test.ts#L26.

To me, it would make more sense to do something like
- Always store the address in the 20 least-significant bytes
- Indicate whether something is internal/external in a flag stored in the 12 most-significant bytes

The link quoted is broken, but this conversation leads me to believe that there could be a bug in this logic: when passed an internal address that happens to end with enough zeros, `_isExternalAddress` would return true.

geoknee:
See also https://github.com/statechannels/monorepo/pull/196

packages/wallet/src/redux/actions.ts
cf19drofxots: Chatted offline about this: this will be a follow up PR.

cf19drofxots:
Note that a lot of the setup in the test file for this feature will be removed once https://github.com/statechannels/monorepo/pull/252 gets in.

packages/wallet/src/utils/transaction-generator.ts
```diff
@@ -126,3 +138,22 @@ export function createDepositTransaction(destination: string, depositAmount: str
...
+  const allocation: Allocation = [
+    {
+      destination: normalizeDestinationAddress(destination),
+      amount
+    } as AllocationItem
+  ];
+  return createNitroTransferAllTransaction(source, allocation);
+}
+
+function normalizeDestinationAddress(address: string): string {
```
snario: What does "normalize" mean? Maybe this should be more explicit and say that it makes any address or external address into an external address.

packages/wallet/src/utils/transaction-generator.ts
```diff
@@ -126,3 +138,22 @@ export function createDepositTransaction(destination: string, depositAmount: str
...
+  let normalizedDestinationAddress = address;
+  // If the address is not already left-padded to be of type byte32
+  if (normalizedDestinationAddress.length !== 66) {
+    normalizedDestinationAddress = convertAddressToBytes32(address);
+  }
+  return normalizedDestinationAddress;
```
snario:
```suggestion
  return address.length === 66 ? address : convertAddressToBytes32(address);
```

packages/wallet/src/utils/transaction-generator.ts
```diff
@@ -1,14 +1,26 @@
...
+
+interface GetDataResult {
+  turnNumRecord: number;
+  finalizesAt: number;
+  fingerprint: BigNumber;
+}
+
+export function nitroGetData(provider, channelId: string): Promise<GetDataResult> {
+  return nitroTrans.getData(provider, getAdjudicatorContractAddress(), channelId);
+}
```
snario: This seems like a helper function that does not need to exist

packages/wallet/src/utils/nitro-converter.ts
```diff
@@ -110,7 +110,8 @@ export function convertAddressToBytes32(address: string): string {
   // We pad to 66 = (32*2) + 2('0x')
   return `0x${normalizedAddress.substr(2).padStart(64, "0")}`;
 }
-function convertAllocationToOutcome({ allocation, destination }: { allocation: string[]; destination: string[] }): Outcome {
+
+export function convertAllocationToOutcome({ allocation, destination }: { allocation: string[]; destination: string[] }): Outcome {
```
snario: Super long line?

packages/wallet/src/redux/actions.ts
snario: I think we should use the term "external address" which is the official name from the docs

cf19drofxots: It is - will update

packages/wallet/src/utils/nitro-converter.ts
cf19drofxots: Yeah mainly because of the ad-hoc typing used.

Will update the typing here in a separate PR.

packages/wallet/src/utils/transaction-generator.ts
cf19drofxots: Yeah it does not  - will remove.

packages/wallet/src/utils/transaction-generator.ts
cf19drofxots: Yeah, perhaps `externalizeAddress`

packages/wallet/src/utils/transaction-generator.ts
snario: The suggested change feature only worked on one line, not all of them, so you'll have to accept this suggestion manually.

cf19drofxots: Yeah that's why I didn't commit the suggestion

cf19drofxots:
Addressed comments

Items for follow up PRs:
- https://github.com/statechannels/monorepo/pull/175/files#r339211973
- https://github.com/statechannels/monorepo/pull/175#discussion_r339207900

cf19drofxots:
@geoknee commit the suggested changes and updated the commands to use `--runInBand`.

Anything else before this is good to merge?

cf19drofxots:
I'll be merging this as soon as CI is green as the comments have been addressed and no further change requests have been made.

packages/wallet/package.json
```diff
@@ -28,11 +28,11 @@
     "prettier:write": "npx prettier --write 'src/**/*.{ts,tsx}'",
     "start": "npm run clearContracts && node scripts/start.js",
     "storybook": "start-storybook -p 9001 -c .storybook",
-    "test:app": "npx jest -c ./config/jest/jest.config.js",
-    "test:ci:app": "yarn test:app --all --ci --runInBand",
-    "test:ci:contracts": "yarn contracts:compile && yarn test:contracts --all --ci --runInBand --detectOpenHandles --bail",
+    "test:app": "npx jest --runInBand -c ./config/jest/jest.config.js",
```
lalexgap: @cf19drofxots Is there a reason we have to use `--runInBand` here? Slows app tests down quite a bit!

cf19drofxots: Only reason was [this comment](https://github.com/statechannels/monorepo/pull/175#pullrequestreview-307982556)

If it's no longer applicable and the tests run fine without that flag then we should remove it.

lalexgap: I think that issue only applies to the contract tests so I think we should be fine to remove it for `test:app`. (They ran fine for me locally without it). I'll update it in one of my PRs.

andrewgordstewart: FYI the issue might have been [resolved](https://github.com/ethers-io/ethers.js/issues/613#issuecomment-547895782) in version 4.0.39. @lalexgap I suggest trying to upgrade ethers-js and running all tests in parallel.

PR #176, opened 2019-10-04
Include signature in AdjudicatorState when challenge arises

snario:
Assuming you were genuinely offline, you will need this information to respond.

### DIFF
--- a/packages/nitro-protocol/src/contract/state.ts
+++ b/packages/nitro-protocol/src/contract/state.ts
@@ -4,7 +4,7 @@ import {encodeOutcome, hashOutcome, Outcome} from './outcome';
 import {Address, Bytes32, Uint256, Uint48} from './types';
 
 export interface State {
-  turnNum: number;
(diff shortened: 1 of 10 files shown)


packages/nitro-protocol/src/contract/state.ts
```diff
@@ -4,7 +4,7 @@ import {encodeOutcome, hashOutcome, Outcome} from './outcome';
 import {Address, Bytes32, Uint256, Uint48} from './types';
 
 export interface State {
-  turnNum: number;
+  turnNum: number; // TODO: This should maybe be a string b/c it is uint256 in solidity
```
lalexgap: While turnNum is currently a `uint256` it sounds like we may be limiting it to `uint48` per https://research.statechannels.org/t/how-large-does-the-largest-turnnum-need-to-be/35, which would allow us to keep `turnNum` as a `number`.  I think it makes sense to keep the TODO for now and address this in the future.

PR #192, opened 2019-10-06
Change padding of external addresses in AssetHolder to left-pad

snario:
Addressing an inconsistency noticed in #175 (see https://github.com/statechannels/monorepo/pull/175#issuecomment-538711824).


geoknee:
See also #196, which is based on this branch.

PR #195, opened 2019-10-12
[Web3Torrent] Initial Library Integration

spersico:
### Description
This PR, when merged, will **add functional download and upload screens**, which allows the user to start (using a magnet link), cancel and re-share the link of the torrent..
#### Demos
[Demo for the Download Screen](https://drive.google.com/open?id=1jyOqmBTM8L7bz1Bt8bYXIiABhn9jZ1xo)
[Demo for the Upload Screen](https://drive.google.com/open?id=1AnOdNCcgM6g7QMo7I0BNENvS8z39jaqO)
#### Screenshots
image
image

#### About the sharing link
Right now, the sharing url is not fully fledged.
There are two approaches to share links, one is using the infoHash (kind of... like an universal ID for the torrent), which is what [instant.io](instant.io) uses:
> /download/infoHash#b7bc5a10ae26784277d04b8955d1fbc48eea7f85

And the other is using the Torrent's Magnet link.
> /download/magnet#magnet:?xt=urn:btih:b7bc5a10ae26784277d04b8955d1fbc48eea7f85&dn=Gestion+de+Proyecto.pdf&xl=30000&cost=2&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com

Personally, I like the infoHash link more, 'cause it's shorter. But it's harder to implement (AFAIK, we need to have an active Tracker Server in order to this approach to work), and the magnet URL (which is implemented in this PR, as shown in the demo) allows me to easily extend the protocol in order to show the cost for each torrent (by adding an extra parameter to the magnet URL).


packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -0,0 +1,138 @@
...
+  } else {
+    return 'Downloading';
+  }
+};
+
+const getFormattedETA = (torrent: Torrent) => {
+  const {done, timeRemaining} = torrent;
+  if (done) {
+    return 'Done';
+  } else {
```
joelalejandro: No need for an `else` here, the rest of the code can be outside the `if`.

packages/web3torrent/src/components/torrent-info/TorrentInfo.scss
```diff
@@ -50,6 +50,48 @@ section {
...
       text-align: right;
+      color: $c-orange;
+      font-weight: 500;
+      margin-top: 0px;
+      font-size: 18px;
+      position: relative;
+      outline:0;
+
+      .tooltiptext {
+        font-family: "Nunito Sans", sans-serif;
```
joelalejandro: Is this font being loaded somewhere?

packages/web3torrent/src/components/torrent-info/download-info/DownloadInfo.tsx
```diff
@@ -1,30 +1,43 @@
...
 import {FormButton} from '../../form';
 import './DownloadInfo.scss';
 import {ProgressBar} from './progress-bar/ProgressBar';
-
 export type DownloadInfoProps = {torrent: Torrent};
 
 const DownloadInfo: React.FC<DownloadInfoProps> = ({torrent}: DownloadInfoProps) => {
+  const [fileURL, setURL] = useState('');
+  useEffect(() => {
+    if (torrent && torrent.done && torrent.files[0] && torrent.files[0].getBlobURL) {
```
joelalejandro: So many `&&` :P

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -96,7 +96,17 @@ export abstract class PaidStreamingExtension implements Extension {
...
 
   protected executeExtensionCommand(command: PaidStreamingExtensionNotices, data = {}) {
-    this.wire.extended(this.name, bencode.encode({msg_type: 0, command, data}));
+    if (!this.peerAccount) {
+      log(
+        'WARNING: Peer does not support Web3Torrent - This client will block all non-web3torrent leechers.'
+      );
+      this.messageBus.emit(
+        PaidStreamingExtensionEvents.WARNING,
+        new Error('!>Peer does not support Web3Torrent')
```
joelalejandro: Can we get rid of the `!>` prefixes?

packages/web3torrent/src/utils/copy-to-clipboard.ts
```diff
@@ -0,0 +1,28 @@
...
+    console.error('error', err);
+  }
+  if (selection) {
+    selection.removeAllRanges();
+  }
+  window.document.body.removeChild(span);
+
+  return success
+    ? Promise.resolve()
+    : Promise.reject(new DOMException('The request is not allowed', 'NotAllowedError'));
```
joelalejandro: What does `The request is not allowed` mean in this context? Can we use a better error message? AFAICT the concept of a request can be a bit weird to bring up during a copy/paste operation.

packages/web3torrent/src/components/torrent-info/download-info/progress-bar/ProgressBar.tsx
```diff
@@ -32,10 +26,10 @@ export const ProgressBar: React.FC<ProgressBarProps> = ({
       </div>
       <div className="negative">
         <span className="bar-progress">
-          {downloaded}Mb/{size}Mb
+          {prettier(downloaded)}/{prettier(length)}
         </span>
         <span className="bar-status">{status}</span>
-        <button type="button" className="bar-cancelButton">
+        <button type="button" className="bar-cancelButton" onClick={() => remove(infoHash || '')}>
```
joelalejandro: `remove()` could have `infoHash` default to `''` to prevent this `||` fallback.

packages/web3torrent/src/components/torrent-info/download-info/progress-bar/ProgressBar.scss
```diff
@@ -9,6 +9,7 @@
   top: 13px;
   left: 300px;
   text-align: center;
+  width:140px;
```
joelalejandro:
```suggestion
  width: 140px;
```

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -0,0 +1,138 @@
...
+      } else {
+        resolve(id);
+      }
+    })
+  );
+};
+
+const getStatus = (torrent: Torrent) => {
+  const {uploadSpeed, downloadSpeed} = torrent;
+  if (torrent.downloaded && torrent.downloaded === torrent.length && torrent.done) {
```
joelalejandro: No need to use `else if` in any of these. Just a sequence of `ifs`, the first one that resolves the condition will make the function exit.

packages/web3torrent/src/components/share-list/ShareList.tsx
```diff
@@ -5,7 +5,7 @@ import {Torrent} from '../../types';
 import {ShareFile} from './share-file/ShareFile';
 import './ShareList.scss';
 
-export type ShareListProps = {torrents: Torrent[]; history: History};
+export type ShareListProps = {torrents: Array<Partial<Torrent>>; history: History};
```
joelalejandro: Good use of `Partial<T>` :+1:

packages/web3torrent/src/components/torrent-info/TorrentInfo.tsx
```diff
@@ -1,28 +1,48 @@
...
-        <span className="fileCost">Est. cost ${Number(torrent.cost).toFixed(2)}</span>
-        {torrent.magnetURI ? (
-          <a href={torrent.magnetURI} className="fileLink">
-            Share Link
-          </a>
-        ) : (
-          false
-        )}
+        <span className="fileCost">
+          Est. cost ${!torrent.cost ? '?' : Number(torrent.cost).toFixed(2)}
```
joelalejandro: Can we say `Unknown` instead of `?`?

packages/web3torrent/src/components/torrent-info/TorrentInfo.scss
spersico: I thought that I did, but... now, seeing the code, I dont, I'll fix it

packages/web3torrent/src/library/paid-streaming-extension.ts
spersico: It follows the _format_ so to speak that we had before. It signals something that's on the peer, and not in _this client_.
I would remove it, but in another PR so that we can define a _format_ so to speak.

packages/web3torrent/src/utils/copy-to-clipboard.ts
spersico: ummm okay :-)

packages/web3torrent/src/library/paid-streaming-extension.ts
joelalejandro: We could have a `PeerError` typed error for this.

packages/web3torrent/src/components/torrent-info/TorrentInfo.scss
spersico: umm... yeah, I did, I forgot, but I import it in the global App.scss

packages/web3torrent/src/library/paid-streaming-extension.ts
spersico: Sure!

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export const download: (torrent: any) => Promise<Torrent> = torrentData =>
+  new Promise(resolve =>
+    web3torrent.add(torrentData, torrent => resolve({...torrent, status: Status.Connecting}))
+  );
+
+export const upload: (files: any) => Promise<Torrent> = files =>
+  new Promise(resolve =>
+    web3torrent.seed(files as FileList, torrent => resolve({...torrent, status: Status.Seeding}))
+  );
```
joelalejandro: These functions should be tagged as `async`, since they're used with `await` further in the code.

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -0,0 +1,124 @@
...
+  const magnetAsArray = magnetURI.split('&');
+
+  const magnetName = magnetAsArray.find(h => h.includes('dn=') || h.includes('name='));
+  const name = magnetName && magnetName.replace(/(dn=)|(name=)/g, '').replace(/\+/g, ' ');
+
+  const magnetLength = magnetAsArray.find(h => h.includes('xl='));
+  const length: number = Number(magnetLength && magnetLength.replace(/(xl=)/g, ''));
+
+  const magnetCost = magnetAsArray.find(h => h.includes('cost='));
+  const cost: number = Number(magnetCost && magnetCost.replace(/(cost=)/g, ''));
```
joelalejandro: This could be simplified using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#Examples):

```ts
const magnetParams = new URLSearchParams(rawMangetUrl.replace(/magnet:/g, ''));
```

As described in the [Magnet URI scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme#Design), you'd have the following parameters as a maximum possible set:

- `dn` (Display Name): a filename to display to the user, for convenience
- `xl` (eXact Length): size in bytes
- `xt` (eXact Topic): URN containing file hash
- `as` (Acceptable Source): Web link to the file online
- `xs` (eXact Source): P2P link identified by a content-hash
- `kt` (Keyword Topic): a more general search, specifying keywords, rather than a particular file
- `mt` (Manifest Topic): link to the metafile that contains a list of magneto (MAGMA – MAGnet MAnifest)
- `tr` (address TRacker): tracker URL for BitTorrent downloads

joelalejandro: Of course, we can leave it as a future enhancement, no need to do it right here.

packages/web3torrent/src/clients/web3torrent-client.ts
spersico: uh, yeah

PR #198, opened 2019-10-15
[Web3Torrent] improve magnetURI generation & parsing

spersico:
### Description

When merged, this PR will improve the magnet link parsing, and generation:

The sharing link will now have this structure (we assume that the trackers are always the same, so I hard coded them so that the sharing link is shorter):
http://localhost:3000/download/magnet#magnet:?xt=urn%3Abtih%3A148c62a7f7845c91e7d16ca9be85de6fbaed3a1f&dn=test.zip&xl=1398978&cost=0

Where 
- **xt** is the id of the torrent 
- **dn** is the name of the torrent/file
- **xl** is the length (size) of the torrent
and **cost**... well.. that's kind of obvious.

### DIFF
--- a/packages/web3torrent/src/clients/web3torrent-client.ts
+++ b/packages/web3torrent/src/clients/web3torrent-client.ts
@@ -1,7 +1,6 @@
 import React from 'react';
 import WebTorrentPaidStreamingClient, {ExtendedTorrent} from '../library/web3torrent-lib';
-import {EmptyTorrent, Status, Torrent} from '../types';
-
+import {defaultTrackers, EmptyTorrent, Status, Torrent} from '../types';
 export const web3torrent = new WebTorrentPaidStreamingClient();
 export const WebTorrentContext = React.createContext(web3torrent);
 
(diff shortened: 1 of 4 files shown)


joelalejandro:
Keep in mind that the meaning of `cost` might be obvious to you since you're working on it, but other people might be clueless about it :)

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -102,23 +101,28 @@ export const parseMagnetURL: (rawMagnetURL: string) => Torrent = (rawMagnetURL =
...
 
   return {
     ...emptyTorrentData,
-    name,
-    magnetURI,
-    length: isNaN(length) ? 0 : length,
-    cost: isNaN(cost) ? 0 : cost
+    name: magnetParams.get('name') || magnetParams.get('dn') || 'unknown',
+    magnetURI: `magnet:?${magnetParams.toString()}`,
+    length: Number(magnetParams.get('xl')) || 0,
```
joelalejandro: Is it worth it to store this info as a number, considering that it's going to be part of a URL which is a string?

spersico: I didn't give this much thought, 'cause I don't have clear how is this going to work. But... yeah, I could change this to be a string everywhere... I'll get to it.

spersico: though length should be a number just the same. In the library the length is always the amount of bytes, and it's a number there, and I believe that it should be a number everywhere, URL notwithstanding

Issue #199, opened 2019-10-15
[embedded-wallet] Unit-Test everything!


joelalejandro:
- [x] Generic UI components (#199)
- [x] JSON RPC router (#204)
- [x] Flow router (#205)
- [x] Message dispatchers (#206)
- [x] Onboarding flow (#208)
- [x] Budget Allocation dialog (#209)
- [x] No Hub dialog (#210)
- [x] Connect to Hub dialog (#211)

PR #201, opened 2019-10-15
[embedded-wallet] Adds unit tests to all generic UI components

joelalejandro:
This PR adds a basic unit test suite for all components under the `ui` directory.

Contributes to #199.

### DIFF
--- a/packages/embedded-wallet/src/ui/dialog/Dialog.tsx
+++ b/packages/embedded-wallet/src/ui/dialog/Dialog.tsx
@@ -55,9 +55,9 @@ const Dialog: React.FC<DialogProps> = ({
               <span className={css.title}>{title}</span>
             </h1>
           ) : (
-            {}
+            []
           )}
-          {closable ? <button onClick={onClose} className={css.close}></button> : {}}
+          {closable ? <button onClick={onClose} className={css.close}></button> : []}
         </header>
         {children ? (
           <section className={css.content}>
--- a/packages/embedded-wallet/src/ui/flow-process/FlowStep.test.tsx
+++ b/packages/embedded-wallet/src/ui/flow-process/FlowStep.test.tsx
@@ -1,0 +1,67 @@
+import Enzyme, {mount, ReactWrapper} from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import React from 'react';
+import {Icon, IconProps, Icons} from '../icon/Icon';
+import {Spinner} from '../spinner/Spinner';
+import {FlowStep, FlowStepProps, FlowStepStatus} from './FlowStep';
+import css from './FlowStep.module.css';
+
+Enzyme.configure({adapter: new Adapter()});
+
+type MockFlowStep = {
+  flowStepWrapper: ReactWrapper;
+  stepElement: ReactWrapper;
+  statusWrapper: ReactWrapper;
+  statusIconElement: ReactWrapper<IconProps>;
+  statusSpinnerElement: ReactWrapper;
+  titleElement: ReactWrapper;
+};
+
+const mockFlowStep = (props?: Partial<FlowStepProps>): MockFlowStep => {
+  const flowStepWrapper = mount(
+    <FlowStep title="Step 1" status={FlowStepStatus.Pending} {...props} />
+  );
+
+  return {
+    flowStepWrapper,
+    stepElement: flowStepWrapper.find('li'),
+    statusWrapper: flowStepWrapper.find(`span.${css.stepStatus}`),
+    statusIconElement: flowStepWrapper.find(Icon),
+    statusSpinnerElement: flowStepWrapper.find(Spinner),
+    titleElement: flowStepWrapper.find(`label.${css.stepTitle}`)
+  };
+};
+
+describe('UI - FlowStep', () => {
+  let component: MockFlowStep;
+
+  beforeEach(() => {
+    component = mockFlowStep();
+  });
+
+  it('can be instantiated', () => {
+    expect(component.stepElement.exists()).toEqual(true);
+    expect(component.titleElement.exists()).toEqual(true);
+    expect(component.statusWrapper.exists()).toEqual(true);
+    expect(component.statusIconElement.exists()).toEqual(true);
+    expect(component.statusSpinnerElement.exists()).toEqual(false);
+  });
+
+  it('can show an Hourglass icon when the status is Pending', () => {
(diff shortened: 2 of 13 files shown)


packages/embedded-wallet/src/ui/dialog/Dialog.tsx
```diff
@@ -55,9 +55,9 @@ const Dialog: React.FC<DialogProps> = ({
               <span className={css.title}>{title}</span>
             </h1>
           ) : (
-            {}
+            []
```
spersico: I recomend that, to the else values of if expressions, where you don't want to show anything, that you use false or null. Because of [this](https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored).

packages/embedded-wallet/src/ui/flow-process/FlowStep.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+import React from 'react';
+import {Icon, IconProps, Icons} from '../icon/Icon';
+import {Spinner} from '../spinner/Spinner';
+import {FlowStep, FlowStepProps, FlowStepStatus} from './FlowStep';
+import css from './FlowStep.module.css';
+
+Enzyme.configure({adapter: new Adapter()});
+
+type MockFlowStep = {
+  flowStepWrapper: ReactWrapper;
```
spersico: If all elements are of the same type, except one, to improve the readability of the test, wouldn't it be better to use something like 
```
type MockFlowStep = {
    [key: string]: ReactWrapper,
    statusIconElement: ReactWrapper<IconProps>
}
```
Or something like that?

packages/embedded-wallet/src/ui/slider/Slider.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const sliderWrapper = mount<SliderProps>(
+    <Slider min={0} max={100} step={1} unit="ETH" {...props} />
+  );
+
+  return refreshSliderFrom({sliderWrapper} as MockSlider);
+};
+
+const refreshSliderFrom = (slider: MockSlider) => {
+  const {sliderWrapper} = slider;
+
```
spersico: Umm... you destructured sliderWrapper, then you use it directly... TOC's!!! TOC's triggered!

packages/embedded-wallet/src/ui/slider/Slider.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+describe('UI - Slider', () => {
+  let component: MockSlider;
+
+  beforeEach(() => {
+    component = mockSlider();
+  });
+
+  it('can be instantiated', () => {
+    expect(component.minLabelElement.exists()).toEqual(true);
```
spersico: You could destructure here :-P

packages/embedded-wallet/src/ui/slider/Slider.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+      ['decreasing', 'ArrowDown', -1],
+      ['decreasing', 'ArrowLeft', -1],
+      ['decreasing', 'PageDown', -10]
+    ];
+
+    beforeEach(() => {
+      component = mockSlider({initialValue: 50});
+    });
+
+    it.each(keyboardCases)('%s it by %i when pressing %s', (_, key, numberChange) => {
```
spersico: niiice

packages/embedded-wallet/src/ui/slider/Slider.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+      expect(value).toEqual(50 + (numberChange as number));
+      expect(component.valueLabelElement.text()).toEqual(`${value}`);
+      expect(component.sliderElement.prop('aria-label')).toEqual(
+        `Current allocation: ${value} ETH`
+      );
+    });
+  });
+
+  it("can't surpass its maximum value", () => {
+    const changed = jest.fn();
```
spersico: I... don't understand what this does, 'cause according to [this](https://jestjs.io/docs/en/mock-functions.html#using-a-mock-function), you need to pass an argument to this function so that jest mocks it, so.... I don't get it

packages/embedded-wallet/src/ui/dialog/Dialog.tsx
joelalejandro: Thanks for suggesting this! I've been bothered by it for quite a while :P

packages/embedded-wallet/src/ui/flow-process/FlowStep.test.tsx
joelalejandro: Not really, I want the keys to be declared explicitly. `[key: string]` is great, but it won't help IntelliSense and makes the type more error-prone by allowing minor typos (i.e. `flowStepWrapper` vs `flowStepWraper` to happen).

packages/embedded-wallet/src/ui/slider/Slider.test.tsx
joelalejandro: Oh, evil me.

Maleficent looks angry.

packages/embedded-wallet/src/ui/slider/Slider.test.tsx
joelalejandro: Maleficent says...

I could. But do I want to?
.
.
.
(I will, I will.)

packages/embedded-wallet/src/ui/slider/Slider.test.tsx
joelalejandro: I actually wanted to use the [TST syntax](https://jestjs.io/docs/en/api#2-testeach-table-name-fn-timeout) but `prettier` kept destroying it.

packages/embedded-wallet/src/ui/slider/Slider.test.tsx
joelalejandro: The purpose of this test is to make sure that an event _is not_ triggered. Since I don't care what the callback does and I just want to know if it was called, I can just spawn an empty mock function and leverage [the `.mock` property](https://jestjs.io/docs/en/mock-functions.html#mock-property) available to all mock functions, and then use matchers such as `toHaveBeenCalled`.

packages/embedded-wallet/src/ui/dialog/Dialog.tsx
joelalejandro: I'll spin up this change in a separate PR to apply it across the entire package.

PR #227, opened 2019-10-21
[nitro-protocol] Split documentation into Docs and API pages

geoknee:
This PR is an attempt to make the documentation easier to digest, chiefly by separating out the more verbose and explanatory content from the technical content (which is more tightly bound to the contract code itself).

The former now has a home under a Docs page (with its own sidebar), and the later under an API page (with its own sidebar). 

The Docs page features some new text that aims to be more introductory and helpful to those uninitiated with state channels. There is still much room for improvement here, though (the subject of future PRs).

The API page features a diagram showing the inheritance relationships between all of the contracts in the codebase, as well as those imported (such as openzeppelin's SafeMath and ERC20 contracts). It also includes autogenerated API documentation from nearly all of these contracts (namely interfaces & implementations), separated into groups such as "Deployed Contracts" and "Test Contracts". A key step in achieving this was switching from `solidity-docgen` (which does not support contracts with external imports) to [a fork of](https://github.com/statechannels/solidoc) `solidoc` (which does, and produces richer docs besides). I forked the repo because it is not maintained and some tweaks were required. 

Preview of this PR: https://5dadb2b783f7b100082aed74--angry-villani-ad8c15.netlify.com/
Current site (for comparison): https://angry-villani-ad8c15.netlify.com


packages/nitro-protocol/docs/state-channels/overview.md
```diff
@@ -0,0 +1,74 @@
...
+id: overview
+title: Overview
+---
+Welcome to the nitro-protocol state channels documentation site. In this introductory pages, you will find a high level description of state channels; their use-cases and the challenges they represent; our solution to those challenges. You will also find a quick start guide to help you nitro-charge / channelize your DApp using our framework.
+
+Elsewhere on the site you will find API documentation for a reference implementation of nitro that is gas optimized and tested.
+
+## What is a State Channel?
+
+State channels are a technique improving the scalability of a blockchain, and for increasing throughput for some of its Decentralized Applications (DApps).
```
Alonski: This reads a bit odd to me. Maybe it should be something like:
"State channels are a technique `for` improving the..."

packages/nitro-protocol/docs/state-channels/overview.md
```diff
@@ -0,0 +1,74 @@
...
+The quickest way to gain an understanding of state channel and their capabilities is by looking at a simple example.
+Here we look at the example of a two-party, _direct_ state channel, between Alice and Bob.
+
+Let's suppose that Alice and Bob want to use a state channel for payments: for example, Alice might be streaming a video from Bob, and wants to pay Bob per megabyte she consumes.
+Executing each payment as a blockchain transaction would be too slow and costly.
+
+To execute the payments through a state channel, instead, Alice and Bob proceed as follows:
+
+1. Alice and Bob agree to open a state channel. By doing this, they agree to a `channelId`, an initial state, and a set of update rules. In this case, the initial balances might be `{A: 10, B: 0}` - 10 coins to Alice and 0 coins to Bob. The update rules might say that Alice can send coins to Bob, but Bob can't unilaterally take coins from Alice.
+2. Alice deposits the coins into an _asset holder_, where they are held specifically for the Alice and Bob's channel. In our example, Bob doesn't need to send any coins because his channel balance starts out at 0 coins.
```
Alonski: Grammer:
"... where they are held specifically for ~the~ Alice and Bob's channel"

packages/nitro-protocol/docs/forcemove-and-nitro/force-move.md
```diff
@@ -0,0 +1,558 @@
...
+
+## Transition Rules
+
+In ForceMove, every state has an associated 'mover' - the participant who had the unique ability to progress the channel at the point the state was created. The mover can be calculated from the `turnNum` and the `participants` as follows:
+
+```solidity
+moverAddress = participants[turnNum % participants.length]
+```
+
+The implication of this formula is that participants take it in turn to update the state of the channel. Furthermore, there are strict rules about whether a state update is valid, based on the previous state that has been announced. Beyond conforming to the state format, there are certain relationships that must hold between the state in question, and the previously announced state.
```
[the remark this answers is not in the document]
geoknee: Yes that is the meaning. I'll change to your wording since there is at least one person who finds that clearer 😄 .

Issue #215, opened 2019-10-17
[Web3Torrent] Unit-Testing!


spersico:
#### App (general)
- [x] App (#225)
#### Generic Components
- [x] Spinner (#225)
- [x] FormButton (#222)
- ~FormInput~ (unused component)
#### Layout Components
- [x] Footer (#225)
- [x] Header (#225)
#### Share Components
- [x] List Share (#225)
- [x] File Share (#225)
#### Info Components
- [x] Torrent Info (#228)
- [x] Progress Bar (#229)
- [x] Download Info (#230)
- [x] Upload Info (#234)
#### Pages
- [x] Welcome (#225)
- [x] Upload (#231)
- [x] Download  (#231)
#### Utils
- [x] Magnet (#232)
- [x] Torrent status checker (#235)
#### Client
- [x] Web3TorrentClient (#236)
- [x] EmbeddedWalletClient (#237)

PR #231, opened 2019-10-21
[Web3torrent] Add Download and Upload Pages tests

spersico:
### Description
Adds tests for the Upload and Download Pages, refactors the Embedded Wallet interaction functions  into a separate file, and refactors the download and upload pages code to be easier to read and test.

### Related issues
#215

### DIFF
--- a/packages/web3torrent/src/pages/download/Download.test.tsx
+++ b/packages/web3torrent/src/pages/download/Download.test.tsx
@@ -1,0 +1,94 @@
+import Enzyme, {mount} from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import {createMemoryHistory} from 'history';
+import React from 'react';
+import {MemoryRouter as Router, RouteComponentProps} from 'react-router-dom';
+import {JsonRPCResponse} from 'web3/providers';
+import {testSelector} from '../../utils/test-utils';
+import * as EmbeddedWalletClient from './../../clients/embedded-wallet-client';
+import Download from './Download';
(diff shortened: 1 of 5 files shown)


packages/web3torrent/src/pages/download/Download.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  beforeEach(() => {
+    const mock = setup();
+    component = mock.component;
+    askForFunds = mock.askForFunds;
+    torrentDownload = mock.torrentDownload;
+    jest.useFakeTimers();
+  });
+
+  it('should render an Download button', () => {
```
joelalejandro: We should also check for `.subtitle`. These tests should ensure no UI components are missing, IMO.

packages/web3torrent/src/pages/upload/Upload.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+  let torrentUpload: jest.SpyInstance<Promise<Torrent>, [WebTorrentSeedInput]>;
+
+  beforeEach(() => {
+    const mock = setup();
+    component = mock.component;
+    torrentUpload = mock.torrentUpload;
+    jest.useFakeTimers();
+  });
+
+  it('should render an Upload button', () => {
```
joelalejandro: We should check for the rest of the UI as well, using the `exists()` helper, at the very least. For this page, as a minimum, a check for having a `type="file"` input would be nice.

PR #252, opened 2019-10-24
[nitro-protocol] Streamline asset withdrawal

geoknee:
~A work in progress~
~Note base branch is not master~

Currently in the happy path, a participant must `conclude`, `pushOutcome` and `transferAll` to get assets out of a channel. We want to streamline that to (ideally) a single transaction with minimal gas expenditure and minimal `calldata` requirements. And without introducing any vulnerabilities into the contracts. 

This PR introdues a `pushOutcomeAndTransferAll` method and a `concludePushOutcomeAndTransferAll` methos (each with an accompanying passing test) which melds multiple transactions into one, shortcutting  writes/reads to storage in each asset holder contract.

TODO:

- [x] combine `pushOutcome` and `transferAll`  in to`pushOutcomeAndTransferAll`
- [x] passing test for this (2x Asset Holders, 57K gas without any *actual* transfers)
- [x] add additional expectations to the test: e.g. were the asset holder holdings updated correctly?

- [x] Make a new method that absorbs `conclude` into `pushOutcomeAndTransferAll` to make `concludePushTransfer`, shortcutting a ~write~/read from storage in the adjudicator contract.
- [x] passing test for this


packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -264,11 +283,45 @@ contract AssetHolder is IAssetHolder {
...
             );
         } else {
-            delete outcomeHashes[guarantee.targetChannelId];
+            delete assetOutcomeHashes[guarantee.targetChannelId];
         }
 
     }
 
+    struct FunctionCall {
+        bool isClaimAll; // otherwise transferAll implied
```
andrewgordstewart: This could be an enum `FunctionType`.

geoknee: I'm bouncing this generalised functionality to future work. This PR is now focussed only on convenience methods that manage a single channel.

Issue #254, opened 2019-10-24
[embedded-wallet] Wallet load feels slow


joelalejandro:
There's a long delay (as much as 4 seconds) between the click of the "Start Download" button on Web3Torrent and the UI of the Embedded Wallet showing up.

We should introduce some form of preloading to prevent this from happening.

PR #257, opened 2019-10-24
[web3torrent] Implements Spinner on "Start Download" button

### LINKED ISSUE #256: [web3torrent] Add a spinner state to the Button component
This proposal is a complement to #254.

Since the "Start Download" button might take a couple of seconds to load the Embedded Wallet app into the iframe, it might be useful to add a "loading" state to the button, which would render a Spinner component, similar to the one the Embedded Wallet has.

joelalejandro:
Contributes to #254. Resolves #256.

image

### DIFF
--- a/packages/web3torrent/src/pages/download/Download.tsx
+++ b/packages/web3torrent/src/pages/download/Download.tsx
@@ -12,6 +12,8 @@ import './Download.scss';
 
 const Download: React.FC<RouteComponentProps> = () => {
   const [torrent, setTorrent] = useState(parseMagnetURL(useLocation().hash));
+  const [loading, setLoading] = useState(false);
+  const [buttonLabel, setButtonLabel] = useState('Start Download');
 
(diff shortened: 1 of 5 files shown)


packages/web3torrent/src/pages/download/Download.tsx
```diff
@@ -25,12 +27,15 @@ const Download: React.FC<RouteComponentProps> = () => {
         <>
           <FormButton
             name="download"
+            spinner={loading}
```
Alonski: Does it make sense for the prop names to be `shouldShowSpinner`, `isShowingSpinner` , `isSpinning` or `isLoading`
This in addition to `loading` being changed to `isLoading`?

joelalejandro: We could definitely improve the names of the properties. I'll log this in a separate issue.

PR #281, opened 2019-10-26
[Web3Torrent] Routing improvements

### LINKED ISSUE #242: [web3torrent] Refactor Download/Upload routes for better file shareability
If a user starts seeding a file, it'd be nice to set the URL of the page to the same that "Share Link" produces, so we provide the user with yet another way of sharing the file's URL.

This would imply refactoring route names. It feels a bit awkward to be in `/upload`, start seeding a file and then be in `/download/magnet`.

We could refactor the routes as follows:

- `/file/new`: the Upload screen.
- `/file/magnet?...`: the Download screen.

spersico:
### Description
closes #242.
[DEMO](https://drive.google.com/open?id=1xRfjW9yvJIC9fbjVXlTlmoEIcOeOd1Lx)

This is done by unifying the Active Upload and Download pages, into a single File page, that checks if the web3torrent client has an active torrent with the infoHash when the page loads, and if so, shows the seeding info, and if not, shows the download screen.

This change also paves the way for a better transition into seeding, after a leecher downloads the file.


packages/web3torrent/src/routes.ts
```diff
@@ -1,5 +1,6 @@
 export enum RoutePath {
   Root = '/',
   Download = '/download/',
```
joelalejandro: This path is now obsolete, we can safely remove it.

packages/web3torrent/src/utils/useInterval.ts
```diff
@@ -1,6 +1,6 @@
 import {MutableRefObject, useEffect, useRef} from 'react';
 
-export function useInterval(callback, delay = 0) {
+export function useInterval(callback, delay: number | boolean = 0) {
```
joelalejandro: Why `number | boolean`?

spersico: 'cause it's the way to stop the setInterval. I thought that that could be a good way to stop/pause the interval. If I pass false, the setInterval pauses.

Issue #277, opened 2019-10-25
[embedded-wallet] [channel-provider] Create new package for injecting `embedded-wallet`


joelalejandro:
Derived from #258, I think we should move the [wallet.js](/statechannels/monorepo/tree/master/packages/embedded-wallet/public/wallet.js) loader to a package of its own, called `channel-provider`.

The benefits of this are non-trivial:

- Better dependency management
- Makes this piece of code testable
- Since it's not React code, it shouldn't be inside a React package

The build output from this package would be the aforementioned `wallet.js` file.

`embedded-wallet` will have a dependency on this package, and then copy `wallet.js` into `public/` after every `react-scripts build` execution.

So, these are the requirements:

- [x] Create the `channel-provider` package with TypeScript and a bundler (done in #278).
- [x] Refactor the `wallet.js` code into TypeScript, as it was started on #272 (done in #282).
- [x] Add unit tests for the code (done in #282).
- [x] Add bundling configuration (this is needed because EIP 1193 requires the use of `EventEmitter`; done in #278).
- [x] Install `@statechannels/channel-provider` as a dependency in `embedded-wallet` (done in #285).
- [x] Modify the build process so the output of `channel-provider` is exposed via the `embedded-wallet` `public` directory  (done in #285).

PR #282, opened 2019-10-28
[channel-provider] Rewrites the wallet loader in TypeScript

joelalejandro:
Contributes to #277.


packages/channel-provider/src/channel-provider.ts
```diff
@@ -1,15 +1,104 @@
-import EventEmitter, {ListenerFn} from 'eventemitter3';
+import debug from 'debug';
+import EventEmitter from 'eventemitter3';
+import {JsonRPCRequest} from 'web3/providers';
```
snario: Is it possible to a void using `web3`?

joelalejandro: I'm just borrowing type definitions, but sure. I could declare them on my own or use something else.

joelalejandro:
@snario requested to add a README (with good reason!), please do not merge this yet :pray:

packages/channel-provider/src/channel-provider.ts
joelalejandro: Done via 7376f32.

PR #283, opened 2019-10-28
feat: Add test for TrivialApp NitroProtocol Contract

Alonski:
### Description
Add test for TrivialApp NitroProtocol Contract
This is mostly to learn how the contract tests work :)

I know this exists:
https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/test/contracts/ForceMoveApp/validTransition.test.ts

It uses different syntax and doesn't test this contract directly.


packages/nitro-protocol/test/contracts/TrivalApp/validTransition.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  `('$description', async ({isValid}: {isValid: boolean}) => {
+    const variablePart: VariablePart = {
+      outcome: HashZero,
+      appData: HashZero,
+    };
+    const isValidFromCall = await trivialApp.validTransition(variablePart, variablePart, 0, 0);
+    if (isValid) {
+      expect(isValidFromCall).toBe(true);
+    } else {
+      expect(isValidFromCall).toBe(true);
```
geoknee: This is quite an unexpected expectation. I understand why you ended up here, since the app we are testing is trivial: but then perhaps we shouldn't ever assign `isValid = false` in this test file?

packages/nitro-protocol/test/contracts/TrivalApp/validTransition.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  it.each`
+    isValid  | description
+    ${true}  | ${'IsValid True = True'}
+    ${false} | ${'IsValid False = True'}
+  `('$description', async ({isValid}: {isValid: boolean}) => {
+    const variablePart: VariablePart = {
+      outcome: HashZero,
+      appData: HashZero,
+    };
+    const isValidFromCall = await trivialApp.validTransition(variablePart, variablePart, 0, 0);
```
geoknee: The existing test has named `from` and `to` `variablePart` structs, which was a bit more readable. Also, here you are testing a pretty special case of transitioning from some fiducial state to itself. That seems overly specific, to me (see other comment about possibility of generating some random data to test with, instead).

packages/nitro-protocol/test/contracts/TrivalApp/validTransition.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+beforeAll(async () => {
+  trivialApp = await setupContracts(provider, TrivialAppArtifact);
+});
+
+describe('validTransition', () => {
+  it.each`
+    isValid  | description
+    ${true}  | ${'IsValid True = True'}
+    ${false} | ${'IsValid False = True'}
```
geoknee: These tables are designed to be used for different test cases, for example each row might describe a different pair of `to` and `from` states to feed the `validTransition` function. As it is, only your *expectations* change (and even then not really) from row to row. 

The table format is really great but this test might not be the place where it really shines :-)

Alonski:
@geoknee I feel that this test is still useful:
https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/test/contracts/ForceMoveApp/validTransition.test.ts

It showscases a ForceMoveApp test in a different way than the new TrivialApp test no?

Alonski:
@geoknee Just pushed an update regarding what we discussed. Moved test from ForceMoveApp -> TrivialApp test file

packages/nitro-protocol/test/contracts/TrivalApp/validTransition.test.ts
```diff
@@ -0,0 +1,69 @@
...
+function computeSaltedHash(salt: string, num: number) {
+  return solidityKeccak256(['bytes32', 'uint256'], [salt, num]);
+}
+
+beforeAll(async () => {
+  trivialApp = await setupContracts(provider, TrivialAppArtifact);
+});
+
+describe('validTransition', () => {
+  it('Random inputs always return true', async () => {
```
geoknee:
```suggestion
  it('Transitions between random VariableParts are valid', async () => {
```

packages/nitro-protocol/test/contracts/TrivalApp/validTransition.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    expect.assertions(5);
+    for (let i = 0; i < 5; i++) {
+      const from: VariablePart = getRandomVariablePart();
+      const to: VariablePart = getRandomVariablePart();
+      const isValidFromCall = await trivialApp.validTransition(from, to, 0, 0);
+      expect(isValidFromCall).toBe(true);
+    }
+  });
+
+  it('Proper state transitions return true', async () => {
```
geoknee:
```suggestion
  it('Transitions between States with mocked-up data are valid', async () => {
```

packages/nitro-protocol/test/contracts/TrivalApp/validTransition.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      turnNum: 1,
+      isFinal: false,
+      challengeDuration: 0x0,
+      appDefinition: trivialApp.address,
+      appData: '0x0',
+    };
+    const toState: State = {...fromState, turnNum: 2};
+
+    expect(
+      // Passing valid transitions, as opposed to VariableParts, can properly validTransition
```
geoknee:
```suggestion
      // Use the helper function, which accepts States instead of VariableParts
```

PR #302, opened 2019-10-29
[nitro-protocol] Introduce website:deploy script

### LINKED ISSUE #259: [docs] Trigger documentation preview deploys on contract/docs changes
Currently the docs site is deployed only for master or manually for certain branches.
It makes sense to deploy the docs whenever they change which would be caused by either the documentation changing or one of the contracts it documents.

This should be setup in CI so that we have a preview for every PR as needed.

geoknee:
Coupled with a change to our netlify config (see below), this will allow website builds to be triggered only when files have changed in the `nitro-protocol` package (including for PRs), as well as cutting the build time roughly to a quarter of what it was (decoupling it from the building of other packages). 

Screenshot 2019-10-29 at 14 03 09

Closes #259 (although any change to the package triggers a build, rather than just the contracts or docs).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 # dependencies
 node_modules
 packages/**/yarn.lock
+# need this in order for netlify to use yarn
+!packages/nitro-protocol/yarn.lock 
 
 # production
 build/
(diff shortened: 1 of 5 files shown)


geoknee:
It's probably because we are using yarn workspaces, which uses only a single `yarn.lock` file in the monorepo root.

geoknee:
https://community.netlify.com/t/difficulty-with-new-monorepo-deployment-options/4381

Issue #299, opened 2019-10-28
[Wallet] Update protocols to use SignedState


lalexgap:
Update individual protocols to use nitro-protocol SignedState instead of Commitments. Since the CommitmentsReceived action contains both SignedCommitments and SignedStates we should be able to switch over protocols individually without breaking the others. The protocols can be worked on in parallel.

**TODO**: This should probably be split up by protocol.

Issue #298, opened 2019-10-28
[Wallet] Add SignedState test helpers


lalexgap:
Add test helpers to create nitro protocol `State` s similar to the helpers in `packages/wallet/src/domain/commitments/__tests__/index.ts`
    - These will be used when updating the protocols.

PR #321, opened 2019-10-31
[engine] Update application protocol to use SignedStates

lalexgap:
Updates the application protocol to work with SignedStates and add some SignedState test helpers.

Progress towards #299 and #298


lalexgap:
Oops, wrong branch

PR #326, opened 2019-10-31
[Hub] Import hub from magmo

kerzhner:
Fixes https://github.com/statechannels/monorepo/issues/249.
- Hub tests are disabled.
- Hub is very likely not usable.

We'll address the above in follow up PRs.  It's good to import the hub into the monorepo sooner than later so that the hub can be updated at the same time as the wallet engine in instances when the hub and the engine share code.


andrewgordstewart:
I might take a stab at branching off this to use the new contracts.

PR #327, opened 2019-11-01
[Engine] Update Consensus Update to work with Signed States

lalexgap:
Updates the consensus update protocol to work with Signed states. Adds some required test helpers and changes channel storage so that we always use the nitro protocol channel Id to store channels.

Some tests have been disabled for other protocols to support these changes. They should be re-enabled when they are updated to support SignedStates.

Progress towards #299 and #298 .


packages/engine/src/domain/commitments/__tests__/index.ts
```diff
@@ -22,10 +21,22 @@ export const participants: [string, string] = [asAddress, bsAddress];
...
+  return [
+    {
+      assetHolderAddress: ETH_ASSET_HOLDER_ADDRESS,
+      allocation: balances.map(b => {
+        const destination = b.address.length === 66 ? b.address : convertAddressToBytes32(b.address);
+        return {destination, amount: b.wei};
+      })
+    }
+  ];
+}
```
tomclose: :)

PR #334, opened 2019-11-01
[rps] Replace dummy Rock Paper Scissors contract with an actual implementation

geoknee:
Ported over from https://github.com/magmo/apps/blob/master/packages/rps/contracts/RockPaperScissorsGame.sol 

- [x] compiles
- [x] has complete logic
- [x] has a test
- [x] test passes

~https://deploy-preview-334--nitro-protocol.netlify.com/docs/natspec/rockpaperscissors~ the contract is now in the [rps] package and therefore does not generate a doc

The contract has much scope for improvement, but should be sufficient to develop the tests against. 

A future PR will flesh out the tests before DRYing out the contract and making other optimisations.


Alonski:
@geoknee Went ahead and merged master into this branch

PR #335, opened 2019-11-01
Add support for USE_NATIVE_SOLC

snario:
If you set `USE_NATIVE_SOLC` then all contract compilation will use `$(which solc)` to compile contracts. Using a native `solc`, such as the one you can get by running `brew install solidity` on a Mac, increases compilation times by up to 10x that of `solc-js`.


packages/nitro-protocol/package.json
```diff
@@ -70,6 +70,8 @@
     "@types/node": "12.7.5",
     "@types/web3": "1.0.19",
     "dotenv": "8.1.0",
+    "etherlime": "2.2.4",
```
geoknee: Do we need this in devDependencies as well as dependencies?

snario: Nope, good catch.

geoknee: Thanks for the fix, but I think we should have `etherlime` as a dependency and not a devDependency. If you check out the deploy preview https://deploy-preview-335--nitro-protocol.netlify.com/docs/api/contract-inheritance, you'll see there are no docs now since netlify uses a production environment and is now not able to compile the contracts (a necessary step in generating the documentation). 

This sort of thing should probably cause an error: https://app.netlify.com/sites/nitro-protocol/deploys/5dbc86fdbf2c650009b477b8

PR #356, opened 2019-11-04
[nitro-protocol] Make etherlime a regular dependency

geoknee:
As a devDependency, etherlime was not being installed in production environments, causing the API side of our docs site to break.


tomclose:
Would it be possible to solve this issue by getting the doc generation on netlify to run in the dev environment? It seems a bit strange to have to make etherlime a production dependency, when generating documentation isn't the purpose of the app.

geoknee:
I think so, will look into ti.

PR #364, opened 2019-11-04
Alon RPS Contract changes

Alonski:
### Description

Added my implementation (incomplete): RockPaperScissorsGame.sol

Updated RockPaperScissors.sol:
- Added requireValidREVEAL transition function
- Renamed many names for things to be more explicit.


packages/nitro-protocol/contracts/examples/RockPaperScissors.sol
```diff
@@ -16,69 +17,79 @@ import '../Outcome.sol';
...
+    enum Weapon {Rock, Paper, Scissors}
 
-    struct RPSData{
+    struct RPSData {
         PositionType positionType;
         uint256 stake;
         bytes32 preCommit;
-        Weapon bWeapon;
-        Weapon aWeapon;
+        Weapon playerFirstWeapon;
```
geoknee: Maybe `playerOneWeapon` and `playerTwoWeapon` would be better. In fact, we should probably use `playerAWeapon` and `playerBWeapon`, since the app uses that nomenclature.

packages/nitro-protocol/contracts/examples/RockPaperScissors.sol
```diff
@@ -16,69 +17,79 @@ import '../Outcome.sol';
...
         bytes32 salt;
     }
 
     /**
-    * @notice Deocdes the appData.
-    * @dev Deocdes the appData.
-    * @param appDataBytes The abi.encode of a RPSData struct describing the application-specific data.
+    * @notice Decodes the appData.
+    * @dev Decodes the appData.
+    * @param toGameDataBytes The abi.encode of a RPSData struct describing the application-specific data.
```
geoknee: `toGameDataBytes` sounds more like a function name (to me) than a variable. Also, `appDataBytes` is a nitro framework level concept so we should probably use that name here.

I guess you were thinking `to` and `from` for each commitment in the transition. But this function is applied to both. 

I would also like to keep with `app` rather than `game` if possible (throughout).

packages/nitro-protocol/contracts/examples/RockPaperScissors.sol
```diff
@@ -87,75 +98,146 @@ contract RockPaperScissors is ForceMoveApp {
...
+
+        // check hash matches
+        // need to convert Weapon -> uint256 to get hash to work
+        bytes32 hashed = keccak256(
+            abi.encodePacked(uint256(toGameData.playerFirstWeapon), toGameData.salt)
+        );
+        require(hashed == fromGameData.preCommit, 'The hash needs to match the precommit');
+
+        // calculate winnings
+        (playerFirstWinnings, playerSecondWinnings) = winnings(
```
geoknee: `playerAWinnings, playerBWinnings` etc

packages/nitro-protocol/contracts/examples/RockPaperScissors.sol
```diff
@@ -87,75 +98,146 @@ contract RockPaperScissors is ForceMoveApp {
...
+            toGameData.playerFirstWeapon,
+            toGameData.playerSecondWeapon,
+            toGameData.stake
+        );
+
+        Outcome.OutcomeItem[] memory outcomeFrom = abi.decode(fromPart.outcome, (Outcome.OutcomeItem[]));
```
geoknee: `fromOutcome` instead of `outcomeFrom` and so on would match the way we modify names elsewhere in this contract.

packages/nitro-protocol/contracts/examples/RockPaperScissorsGame.sol
```diff
@@ -0,0 +1,295 @@
+pragma solidity ^0.5.11;
```
geoknee: Should this file be deleted?

packages/nitro-protocol/contracts/examples/RockPaperScissorsGame.sol
```diff
@@ -0,0 +1,295 @@
...
+            (Outcome.AllocationItem[])
+        );
+        require(allocationB[0].destination == allocationA[0].destination,'RockPaperScissorsGame: Destimation playerFirst may not change');
+        require(allocationB[1].destination == allocationA[1].destination,'RockPaperScissorsGame: Destimation playerSecond may not change');
+        require(allocationB[0].amount == allocationA[0].amount,'RockPaperScissorsGame: Amount playerFirst may not change');
+        require(allocationB[1].amount == allocationA[1].amount,'RockPaperScissorsGame: Amount playerSecond may not change');
+        _;
+    }
+
+    modifier oneAssetType(
```
geoknee: We should consider putting these in a library for app developers to reuse (they might have to be functions rather than modifiers, then).

packages/nitro-protocol/contracts/examples/RockPaperScissors.sol
```diff
@@ -87,75 +98,146 @@ contract RockPaperScissors is ForceMoveApp {
...
+
+        require(
+            allocationTo[0].amount == allocationFrom[0].amount.add(playerFirstWinnings),
+            "Player First's allocation should be updated with the winnings."
+        );
+        require(
+            allocationTo[1].amount ==
+                allocationFrom[1].amount.sub(fromGameData.stake.mul(2)).add(playerSecondWinnings),
+            "Player Second's allocation should be updated with the winnings."
+        );
```
geoknee: We need to require that the the destinations are not changing, *globally* across all transitions.

i.e. `allocationTo[0].destination == allocationFrom[0].destination`

packages/nitro-protocol/contracts/examples/RockPaperScissors.sol
Alonski: I personally think `playerA` and `playerB` can be a bit confusing.
The reason for `playerFirst` is that in this turn based game `playerFirst` is the player who initiates the game.
If the DApp also uses this terminology do you think we should just stick with `A` and `B` or maybe refactor there as well?

packages/nitro-protocol/contracts/examples/RockPaperScissors.sol
Alonski: Woops this was a mistake :)
Lots of refactoring

packages/nitro-protocol/contracts/examples/RockPaperScissors.sol
geoknee: How about we stick with `A` and `B` for now, and revisit once the DApp is working again. I think at this stage it will cause us more pain to introduce a new nomenclature (even if it is superior).

PR #365, opened 2019-11-04
[monorepo] Add RPS from magmo/apps

lalexgap:
Add the RPS package to the monorepo from magmo/apps.


lalexgap:
Looks like I didn't do this quite right. Will try again.

PR #378, opened 2019-11-05
[rps] Get package compiling and tests running

geoknee:
This required some dependency upgrades.

- [x] Placeholder contract (compiles and deploys, has no business logic but correct interface)
When #364 and #334 are finished off, we can overwrite this placeholder with one containing the correct business logic.
- [x] Use etherlime instead of truffle
- [x] Upgrade dependencies and fix some TS errors
- [x] `npx tsc` works
- [x] Test file with `it.each` jest table format with single test case. Handles RPS `appData` and `outcome` and formats these appropriately for on chain contract call.
- [ ] `yarn start` works

Probably in separate PRs:
- [ ] Port legacy test cases (currently commented out)
- [ ] DRY out the test file, remove legacy TS
- [ ] Move some test functions from `nitro-protocol/test-helpers` to `devtools`.


packages/nitro-protocol/src/index.ts
```diff
@@ -48,4 +55,10 @@ export {
   getDepositedEvent,
   isAllocationOutcome,
   isGuaranteeOutcome,
+  VariablePart,
+  // test helperts -- TODO move these to devtools package
```
Alonski: Should this be `test helpers`?

geoknee: mos def. Hopefully I will get those functions moved and the comment wiped before this PR is ready :-)

geoknee:
Closing in favour of #401

PR #384, opened 2019-11-05
[api] API updates

tomclose:
Ongoing updates to the app-wallet api.


packages/app-wallet-interface/source/index.html.md
```diff
@@ -0,0 +1,321 @@
...
+    "turnNum": 7,
+    "allocations": [
+      {
+        "token": "0x...", // 0x0 for ETH
+        "allocationItems": [
+          {"destination": "0xa...", "amount": "18"},
+          {"destination": "0xb...", "amount": "6"}
+        ]
+      }
+    ],,
```
joelalejandro:
```suggestion
    ],
```

packages/app-wallet-interface/source/index.html.md
```diff
@@ -0,0 +1,321 @@
...
```
joelalejandro:
```suggestion
    ],
```

packages/app-wallet-interface/source/index.html.md
```diff
@@ -0,0 +1,321 @@
...
+## Create Channel
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "CreateChannel",
+  "id": 1,
+  "params": {
+    "participants": [
+      {"participantId": "user123", "signingAddress": "0x...", "destination": "0xa..."},
```
lalexgap: Is this `signingAddress` gotten from the wallet via another API call? If the app/client is providing it it means the `wallet` can't sign things states on their behalf.

packages/app-wallet-interface/source/index.html.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+| Parameter      | Type   | Description                                                     |
+| -------------- | ------ | --------------------------------------------------------------- |
+| participantId  | String | App allocated id, used for relaying messages to the participant |
+| signingAddress | String | Address used to sign channel updates                            |
+| destination    | String | Address of EOA to receive channel proceeds                      |
+
+Note: in the future we might replace the `appId` with a `contactAddress`, which would allow
+apps to get the wallet to relay messages itself. An example `contractAddress` would be something
```
lalexgap:
```suggestion
apps to get the wallet to relay messages itself. An example `contactAddress` would be something
```

packages/app-wallet-interface/source/index.html.md
tomclose: Yes. We'll need another API call to get the `signingAddress`.

Issue #398, opened 2019-11-06
[channel-client] Implement a public-facing API for dapps to use


joelalejandro:
This issue will track the creation of a `ChannelClient`.

The API is comprised of the following actions.

1. If someone calls `openChannel`, create the corresponding `opening` state (using fake data where required) and:
    1. Trigger a `channelUpdated(opening)` 
    2. Trigger a `sendMessage(opponent, channelProposed(opening))`
2. If someone calls `messageReceived(someEvent)`, then
    1. Trigger `someEvent`
3. If someone calls `joinChannel`
    1. Trigger a `channelUpdated`
    2. Trigger a `sendMessage(opponent, channelUpdated)`
4. If someone calls `updateState()`
    1. Trigger a `channelUpdated`
    2. Trigger a `sendMessage(opponent, channelUpdated)`

The API was expanded via #384 and #427.

### Tasks

- [x] Implement `openChannel`
- [x] Implement `sendMessage`
- [x] Implement event transmission for `channelUpdated`
- [ ] Implement error types
- [ ] Simulate data flow

Implementation and related discussions are happening in #402.

snario:
This is now being worked on by @andrewgordstewart (see #1933 for example)

PR #402, opened 2019-11-06
[rps] [channel-client] First implementation of a Channel Client

joelalejandro:
Contributes to #398.


joelalejandro:
@tomclose @geoknee @spersico - As of dc1f2f3, I've drafted a `ChannelClient` interface with the methods described in #398, based on our internal discussions. It wraps an `EventEmitter` instance that is used to trigger `channelUpdated` messages for each public method: `openChannel`, `joinChannel` and `updateState`.

For what I've read on the [MessageService saga](/statechannels/monorepo/tree/rps/channel-client/packages/rps/src/redux/message-service/saga.ts), the [`sendMessage` method](/statechannels/monorepo/tree/rps/channel-client/packages/rps/src/redux/message-service/state.ts) updates the state to trigger a reducer that eventually emits a message through Firebase.

The goal would be to extract out any messaging-related code from the saga to the `ChannelClient` itself. A first logical step would be to relocate the Firebase integration into the `ChannelClient`, maybe in the form of an abstract `MessagingService` layer, [similar to what we currently have on the `ChannelProvider`](/statechannels/monorepo/tree/rps/channel-client/packages/channel-provider/src/messaging-service.ts).

With these considerations in mind, the next few commits should be all about modifying the [`sendMessagesSaga()`](/statechannels/monorepo/tree/rps/channel-client/packages/rps/src/redux/message-service/saga.ts#L72-L118) to delegate to a new `ChannelClient.sendMessage()` method.

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,53 @@
...
+  async joinChannel(parameters: JoinChannelParameters) {
+    this.events.emit('channelUpdated', {...parameters, state: ChannelState.Open});
+  }
+
+  async updateState(parameters: MessageParameters) {
+    this.events.emit('channelUpdated', {...parameters, state: ChannelState.Open});
+  }
+
+  protected async sendMessage(method: string, parameters: MessageParameters) {
+    // TODO: What goes here?
```
tomclose: This should probably emit a `relayMessage` event. The app will bind its message relayer to these events, and send them over the wire to the opponent.

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,53 @@
...
+  channelId: string[];
+  state: 'open';
+}
+
+export class ChannelClient {
+  protected events: EventEmitter = new EventEmitter();
+
+  constructor() {
+    this.events.on('channelUpdated', (payload: MessageParameters) => {
+      this.sendMessage('chan_update', payload);
```
tomclose: I think you probably don't quite want this, as `channelUpdated` can be triggered in one of two ways: (1) when the app calls a method like `updateState`, (2) when a `channelUpdated` event arrives via `messageReceived`. You only want to `sendMessage` in the first of these situations, as otherwise you will end up in an infinite loop!

You probably want to explicitly trigger a `sendMessage` in the `openChannel`, `joinChannel` and `updateState` methods.

tomclose:
@joelalejandro the intention at the moment is to keep the message sending functionality in the app, rather than having any messaging-related code in the client. The client should emit a `sendMessage` event, which the app would bind to in a `message-relayer` (a saga), which would then send this off to firebase (in this case). 

This pattern will allow us to use the same `ChannelClient` for the web3torrent app, where the messages will be sent via a completely different mechanism (and potentially combined with other app-related messages).

tomclose:
The `ChannelClient` _will_ need something akin to the `messaging-service` that you linked to, for communicating with the wallet.

joelalejandro:
@tomclose As we discussed during our call today, I'll be refactoring the API and adding the types described in 433f477.

joelalejandro:
@tomclose @geoknee @spersico As of 1b83d74, I've implemented all of the necessary interfaces to reflect the data structures and API methods described in 433f477.

I've also left some open questions about the `channelProposed` and `channelUpdated` events.

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,219 @@
...
+    // TODO: This notification payload is incomplete, since it doesn't have the result
+    // of the channel creation. Where should this go?
+    this.notifyChannelUpdated(parameters as UpdateChannelResult);
+  }
+
+  updateChannel(parameters: UpdateChannelParameters) {
+    this.send(MethodName.UpdateChannel, parameters);
+  }
+
+  protected async send(methodName: MethodName, parameters: ActionParameters) {
```
tomclose: I was finding it a little confusing in the code to figure out what `send` meant. Maybe calling this `sendToWallet` or something would help with this?

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,219 @@
...
+    this.events.emit('message', {
+      jsonrpc: '2.0',
+      id: Date.now(),
+      method: methodName,
+      params: parameters,
+    });
+  }
+
+  protected async notify(notificationName: NotificationName, parameters: ActionParameters) {
```
tomclose: `notifyApp`?

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,219 @@
...
+  channelId: string;
+  status: ChannelStatus;
+  funding: Funding[];
+}
+
+export interface CreateChannelRequest
+  extends JsonRPCRequest<CreateChannelParameters, MethodName.CreateChannel> {}
+
+export interface CreateChannelResult extends ChannelResult {
+  turnNum: 0;
```
tomclose: I don't think it's important to enforce this at the typing level. `turnNum` can just be a bigNum.

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,219 @@
...
+        }
+      }
+    );
+  }
+
+  createChannel(parameters: CreateChannelParameters) {
+    this.send(MethodName.CreateChannel, parameters);
+
+    // TODO: This notification payload is incomplete, since it doesn't have the result
+    // of the channel creation. Where should this go?
```
tomclose: For faking purposes, this code should add a `channelId`, `turnNum: 0`, `status: 'opening'` etc. to the hash. The `ChannelClient` will also need to store the last state of the channel internally, to allow it to fake out the result of other methods.

Once we stop faking, this method will probably need to wait for the response to the `CreateChannel` call from the wallet. This method (and others) will need to be `async`. We should probably make that change now.

tomclose:
Thinking about it, there's actually only one response format:
```ts
interface Channel {
  channelId: Bytes32;
  status: "opening" | "funding" | "running" | "closing";   // tbc
  funding: Funds[];
  participants: Participant[];
  turnNum: BigNum;   // tbc
  allocations: Allocation[];
  appData: string;
}
```

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,219 @@
...
+
+export enum ChannelStatus {
+  Opening = 'opening',
+}
+
+export enum MethodName {
+  CreateChannel = 'CreateChannel',
+  JoinChannel = 'JoinChannel',
+  UpdateChannel = 'UpdateChannel',
+}
```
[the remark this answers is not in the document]
joelalejandro: I'm not just using it for types, but also as a value, hence the use of an Enum.

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,219 @@
...
+export interface Funding {
+  token: string;
+  amount: string;
+}
+
+export interface ParameterHash {
+  [key: string]: any;
+}
+
+export interface JsonRPCNotification<ParametersType = ParameterHash, MethodType = string> {
```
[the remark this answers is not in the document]
joelalejandro: It served a purpose to have a more flexible type before modeling the specific `CreateChannel*` types. We can safely remove this once the API's more stable, I think.

packages/rps/src/utils/channelClient.ts
joelalejandro: Sure!

packages/rps/src/utils/channelClient.ts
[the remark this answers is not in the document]
andrewgordstewart: If you use a union type, smart editors like vscode provides good autocompletion, meaning, when writing code, you can use string literals as values with less work than using enums as values.

Screen Shot 2019-11-08 at 9 53 17 AM

Do you get a further benefit from using an enum?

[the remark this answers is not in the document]
joelalejandro: Enums are transpiled as part of the codebase and are accessible to the consumer, but that's pretty much it. I don't particularly favor one style over the other, I'm just more used to enums, but I don't mind changing it.

packages/rps/src/utils/channelClient.ts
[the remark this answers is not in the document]
joelalejandro: Removed this while turning enums into literal types.

packages/rps/src/utils/channelClient.ts
```diff
@@ -0,0 +1,219 @@
...
+  | CreateChannelParameters
+  | JoinChannelParameters
+  | UpdateChannelParameters;
+
+export class ChannelClient {
+  protected events = new EventEmitter();
+
+  onMessageReceived(
+    notificationName: NotificationName,
+    callback: (parameters: ActionParameters) => void
```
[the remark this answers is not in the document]
joelalejandro: No longer needed, as there's a single `ChannelResult` type.

joelalejandro:
@tomclose @geoknee @spersico As of 25e4bd0, this PR fulfills our expectations for a first implementation. An upcoming PR will deal with implementing a fake state service to simulate responses.

PR #410, opened 2019-11-07
[wallet] Get Start Script working

### LINKED ISSUE #305: [Wallet] Update start script to function with latest devtools
The devtools have been updated to improve our testing workflow however our start script has not, so it no longer functions.

We should update the start script so it functions.

lalexgap:
Updates the start script to run. 

I've removed some of `react-dev-utils` function calls we were using to set up webpack as it seems like they were trying to suppress webpack output and replace it with their own. This resulted in the type checking seeming to wait indefinitely as the `react-dev-utils` function must not of been hooking correctly in to `forktscheckerwebpackplugin` for the typing checking results.

Fixes #305.


spersico:
I merged #411 into this, and I got to run `yarn` successfully in the root folder, and the `yarn start` script in the **wallet**  package. Haven't tested if the wallet actually works or not, but now at least it builds and runs!! :-D.

PR #424, opened 2019-11-07
Shared Contracts Infrastructure

cf19drofxots:
## Update

The packages that use the contract have all their tests passing and their READMEs updated with instructions on how to locally setup the needed contracts.

The updated README can be read [here](https://github.com/statechannels/monorepo/blob/0c50e8a6165155e6a6cc07ce0d8db5f216e6542d/readme.md#development-flow).

The three main notes of interest from this work are:
1. Now you can run `yarn start:ganache` from the root to spin up a ganache instance and have the relevant contracts deployed to it, and all other packages seamlessly have access to this information which is now being called the `NetworkContext` (name could change if a better one is suggested). This ganache server can be watched or manually restarted upon any contract code updates and other packages will pick this change up without any additional work.
- This is a property of the `NetworkContext` being written to the same path in every spun up instance of the ganache server and upon closing it this file _gets deleted_ (as described in the root README).

2. The way that packages access this `NetworkContext` is simply via an import statement from the `nitro-protocol` package. This is just an interim placement to get things working and could easily change. By importing the `NetworkContext` this way, we can both satisfy local development (as is the work of this PR) and reference static files that get checked in the repo which hold the contexts for production networks (i.e. `import `NetworkContext` from "@statechannels/nitro-protocols/deployments/main-net-network-context.json`) and have these contexts be bundled up into any consuming package when said package is being built for publication.

3. Although there is one `NetworkContext` that gets created initially, the current mechanism of deployments through the devtools package (specifically the `GanacheServer`) can be extended for any package to deploy additional contracts and update the global `NetworkContext`. This extension has not been implemented as it didn't seem necessary for it to be in the scope of this PR.

---------

## Motivation:

An implementation for an infrastructure that meets the acceptance criteria described in [this document](https://www.notion.so/elfour/Shared-Contracts-b617526c5e6b42229b4493089770b1b9).

Partially based on some work located [in this branch](https://github.com/statechannels/monorepo/tree/ags/dev-server)

Note that the added/removed LOC is mostly due to `yarn.lock` updates since dependencies changed in this branch.


packages/nitro-protocol/.env
```diff
@@ -2,3 +2,4 @@ DEFAULT_GAS=6721975
 DEFAULT_GAS_PRICE=20000000000
 GANACHE_HOST=0.0.0.0
 GANACHE_PORT=8545
+NODE_ENV=dev
```
andrewgordstewart: Should this be in `.env.development?`?

packages/nitro-protocol/ganache/index.ts
```diff
@@ -0,0 +1,36 @@
+import {configureEnvVariables, GanacheServer} from '@statechannels/devtools';
+import Koa = require("koa");
+import Router = require('koa-router');
+import {deploy} from '../deployment/deploy-test';
+import {logger} from './logger';
+// configureEnvVariables();
```
andrewgordstewart:
```suggestion
```

andrewgordstewart: Should this be run? or should it be deleted?

packages/nitro-protocol/package.json
```diff
@@ -46,7 +46,7 @@
     "docgen": "npx solidoc",
     "prettier:check": "npx prettier --check '{contracts,src,test}/**/*.{ts,tsx,sol}'",
     "prettier:write": "npx prettier --write '{contracts,src,test}/**/*.{ts,tsx,sol}'",
-    "startGanache": "node bin/start-ganache.js",
+    "startGanache": "npx tsc ganache/index.ts && NODE_ENV=dev npx node ganache/index.js",
```
andrewgordstewart:
```suggestion
    "startGanache": "NODE_ENV= npx ts-node ganache/index.js",
```

I think this should work (as long as `ts-node` is a devDep).

With this change, you can revert the change to `.gitignore`.

packages/nitro-protocol/ganache/index.ts
```diff
@@ -0,0 +1,36 @@
+import {configureEnvVariables, GanacheServer} from '@statechannels/devtools';
```
andrewgordstewart: Did you put any thought into whether `nitro-protocol/ganache` was a good location for this? I did not put much thought into the name `ganache`, and I'm not sure it's a good name.

Maybe `nitro-protocol/dev-environment` is better? Or, maybe it even belongs in its own package.

packages/devtools/src/utils/ganache-server.ts
```diff
@@ -48,18 +53,15 @@ export class GanacheServer {
 
     const oneMillion = ethers.utils.parseEther("1000000");
 
-    const opts = [`--networkId ${process.env.GANACHE_NETWORK_ID}`, `--port ${this.port}`].concat(
-      accounts.map(a => `--account ${a.privateKey},${a.amount || oneMillion}`)
-    );
-    const cmd = `ganache-cli ${opts.join(" ")}`;
-
-    this.server = spawn("npx", ["-c", cmd], {stdio: "pipe"});
```
andrewgordstewart: There was a reason for spawning ganache this way, which I don't fully remember -- unfortunately, the commit message is lost due to a squash commit. It was something like: "When I use `ganache` from `ganache-core`, each test case takes ~2s longer". I couldn't understand why I experienced it at the time. Did you experience anything like that here?

packages/nitro-protocol/.env.example
```diff
@@ -2,4 +2,5 @@
 DEFAULT_GAS=6721975
 DEFAULT_GAS_PRICE=20000000000
 GANACHE_HOST=0.0.0.0
-GANACHE_PORT=8545
\ No newline at end of file
+GANACHE_PORT=8545
+NODE_ENV=dev
```
andrewgordstewart: If we're checking in `.env`, then I don't think we need a `.env.example` file at all.

packages/nitro-protocol/.env
cf19drofxots: Yeah there shouldn't be a `.env` file checked in at all. Should only need `.env.example` and `.env.local` (or `.env.development`)

packages/nitro-protocol/ganache/index.ts
cf19drofxots: This is a temporary commenting out. I haven't thought much about how this configuration should be handled as of yet (getting to that soon so will update this line).

packages/nitro-protocol/package.json
cf19drofxots: Can we not direct `ts-node` to a `.ts` file?

Seems weird that we'd need to check in a `.js` file if we're using `ts-node` to begin with.

packages/nitro-protocol/ganache/index.ts
cf19drofxots: Wasn't too concerned with the name of it as the naming can always just change. Just wanted to get things up and running.

packages/devtools/src/utils/ganache-server.ts
cf19drofxots: Hmm I didn't but perhaps before I go much further with this I should.

packages/nitro-protocol/.env.example
cf19drofxots: https://github.com/statechannels/monorepo/pull/424#discussion_r343954686

packages/nitro-protocol/.env
andrewgordstewart: At magmo, we were checking in `.env` and `.env.$ENVIRONMENT` files, but not `.env.local`or `.env.$ENVIRONMENT.local` files. I know this goes against the 12 factor app philosophy, but it means
- everyone had good defaults, including circleci. they can overwrite the defaults with their `.local` env files
- nobody had to remember to create their own `.env.` file, or figure out what to put in it

What's wrong with this approach?

packages/nitro-protocol/package.json
andrewgordstewart: that was a typo: should have been `ganache/index.ts`

packages/nitro-protocol/.env
andrewgordstewart: I guess one idea is that
- `devtools` exports a `config` object
- `devtools` is responsible for creating sensible defaults, depending on the node environment
- Any environment is able to overwrite these defaults with an environment variable
- No `.env` files are checked in

Maybe this is more like the 12FA approach?

cf19drofxots: I'm not a fan of everyone needing to create their own `.env` either so +1 for that.

I don't think there's anything wrong with that. Given that the rules around picking up `.env.*` are pretty arbitrary, it's essentially our choice of whether we want

- `.env` to be the expected configuration for _only_ prod and have `.env.<X>` for other contexts or
- `.env` being the local configuration, custom configuration can be provided via `.env.<X>`, and prod context can be `.env.prod` (or something like that)

If anything I lean towards the latter as well.

cf19drofxots: > Any environment is able to overwrite these defaults with an environment variable

It seems to me that `.env.<environment>` would be doing something similar to this. It'd be providing its context-specific _default_ env vars, and if even _these_ defaults need to be overwritten then they can be specified via env vars.

If that's not how you're thinking about it then I guess you could elaborate on what you're envisioning.

andrewgordstewart: what I mean by an "environment variable" is something that ends up on `process.env`.

Suppose I want to know the configured value of `VAR`. Currently, the `configEnvVariables` sets `process.env.VAR` to be the value of the shell variable `$VAR` if it exists. Otherwise, it looks for a `VAR` in this set of files: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use

It's modified to suit our monorepo by looking in the monorepo root as well: https://github.com/statechannels/monorepo/blob/master/packages/devtools/src/config/env.ts#L23-L27

I think this logic currently doesn't make sense, since I believe that `packages/nitro-protocol/.env` takes precedence over `.env.production`.

packages/devtools/src/utils/ganache-server.ts
andrewgordstewart: Maybe just compare the test run times on circle once they pass?

cf19drofxots: I'd be surprised if there was a huge difference, since `ganache-cli` uses `ganache-core` under the hood.

packages/devtools/.prettierrc.js
```diff
@@ -2,7 +2,7 @@
 
 module.exports = {
     ...require('../../.prettierrc.js'),
-    "printWidth": 120,
+    "printWidth": 100,
```
cf19drofxots: I'll move this change out as well.

cf19drofxots:
> Do you think this infrastructure can satisfy both our testing needs and development needs?

I just saw this. Yeah, that's the goal.

packages/devtools/src/types.ts
```diff
@@ -14,8 +14,5 @@ export interface DeployedArtifact {
 }
 
 export interface DeployedArtifacts {
-  [artifactName: string]: {
-    address: string;
-    abi: string;
-  };
+  [artifactName: string]: DeployedArtifact;
```
andrewgordstewart: If you want to add some extra niceness, you can do
```
type ArtifactName = 'Adjudicator' | 'AssetHolder';
export interface DeployedArtifacts {
  [artifactName in keyof ArtifactName]?: DeployedArtifact
}
```

This way, `DeployedArtifacts` can only have whitelisted properties, making for good autocomplete (if that's useful).

andrewgordstewart: FYI this is the same as `export type DeployedArtifacts = Partial<Record<ArtifactName, DeployedArtifact>>`

packages/wallet/src/contract-tests/eth-asset-holder-events.test.ts
```diff
@@ -59,7 +59,7 @@ describe("ETHAssetHolder listener", () => {
     expect(action.destinationHoldings).toEqual(depositAmount);
   });
 
-  it("should handle an AssetTransferred event", async () => {
+  it.skip("should handle an AssetTransferred event", async () => {
```
cf19drofxots: The skip here was to unblock the remainder of the work. This needs to be un-skipped and the error it's raising is
```
VM Exception while processing transaction: revert Only the NitroAdjudicator is authorized
```

packages/wallet/src/contract-tests/transactions.test.ts
```diff
@@ -274,7 +271,7 @@ describe("transactions", () => {
     await testTransactionSender(transferAndWithdraw);
   });
 
-  it("should send a conclude, push outcome, and transfer all transaction", async () => {
+  it.skip("should send a conclude, push outcome, and transfer all transaction", async () => {
```
cf19drofxots: Skipping for the same reason as the other test. Will need to un-skip this and the error in this case is:
```
VM Exception while processing transaction: revert Deposit | holdings[destination] already meets or exceeds expectedHeld + amount
```

cf19drofxots:
The two comments made will have specific PRs dedicated for un-skipping those tests but since the nature of this branch shouldn't cause a functional change to the contracts, it's most likely a small issue that needs to be resolved.

packages/hub/readme.md
```diff
@@ -10,9 +10,13 @@ https://devcenter.heroku.com/articles/heroku-cli
 
 ### .env
 
-Copy `.env.example` to `.env`, and make it your own.
+Copy `.env.development` to `.env`, and make it your own.
```
andrewgordstewart: I think we would check in both `.env.development` and `.env` to git, right? (`.env.*.local` files should _not_ be checked in)

packages/nitro-protocol/ganache/deployer.ts
```diff
@@ -0,0 +1,44 @@
...
+    testAssetHolderArtifact2,
+    trivialAppArtifact,
+    countingAppArtifact,
+    singleAssetPaymentsArtifact,
+    nitroAdjudicatorArtifact,
+    testNitroAdjudicatorArtifact,
+    tokenArtifact,
+    {
+      artifact: erc20AssetHolderArtifact,
+      arguments: [testNitroAdjudicatorArtifact.contractName, tokenArtifact.contractName],
```
andrewgordstewart: 💡 if you take the suggestion above, this would be something like

```suggestion
      arguments: deployedArtifacts => [
        deployedArtifacts.testNitroAdjudicatorArtifact.address,
        deployedArtifacts.tokenArtifact.address
      ]```

File: packages/nitro-protocol/test/contracts/ForceMove/conclude.test.ts [hunk 13]
```diff
@@ -45,8 +44,7 @@ for (let i = 0; i < 3; i++) {
 beforeAll(async () => {
   networkMap = await getNetworkMap();
   ForceMove = await setupContracts(provider, ForceMoveArtifact);
-  networkId = (await provider.getNetwork()).chainId;
-  appDefinition = networkMap[networkId][countingAppArtifact.contractName]; // use a fixed appDefinition in all tests
+  appDefinition = networkMap[countingAppArtifact.contractName].address; // use a fixed appDefinition in all tests
```
andrewgordstewart: Maybe you can get rid of the duplicated comments by importing the counting app as `placeholderArtifact` or something?

packages/devtools/src/types.ts
```diff
@@ -0,0 +1,18 @@
+export interface Account {
+  privateKey: string;
+  amount?: string;
+}
+
+export interface Deployment {
+  artifact: any;
+  arguments?: any[];
```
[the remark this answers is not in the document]
andrewgordstewart: Actually, making it mandatory to supply it might be more pain than benefit.

packages/wallet/package.json
```diff
@@ -31,7 +31,7 @@
     "test:ci:app": "yarn test:app --all --ci --runInBand",
     "test:ci:contracts": "yarn contracts:compile && yarn test:contracts --all --ci --detectOpenHandles --bail",
     "test:ci": "yarn test:ci:contracts && yarn test:ci:app",
-    "test:contracts": "npx jest --runInBand -c ./config/jest/jest.contracts.config.js",
+    "test:contracts": "npx jest --runInBand -c ./config/jest/jest.contracts.config.js --detectOpenHandles --bail",
```
andrewgordstewart: 👍

packages/wallet/src/utils/contract-utils.ts
```diff
@@ -1,18 +1,21 @@
+import {State, validTransition} from "@statechannels/nitro-protocol";
+// @ts-ignore this is not supposed to exist at the prepare step
```
andrewgordstewart: I don't understand this comment.

packages/devtools/src/types.ts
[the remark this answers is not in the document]
cf19drofxots: I'm having trouble seeing this coupling.

> the code in `ganache-server.ts` isn't coupled to the constructor signatures of the contracts being deployed

AFAICT the code in `ganache-server.ts` is just detecting if the contract has any arguments, it doesn't care about what they are. So the calling code (in this case `deployer` in `nitro-protocol`) specifying these arguments as just the dependent contracts' names in a list, or their `DeployedArtifact` representations (i.e. said contracts' names _and_ their addresses) via a function doesn't seem to be a concern.

The function implementation would require access to a `deployedArtifacts` object holding the dependent contracts' addresses (as referenced in the second code snippet above), which is probably not something the _calling_ code should care about, right?

And to continue on that note:

> I'm not sure that's clear to someone without the context that the only constructor arguments are other contract addresses.

in the current circumstances of deploying the contracts, it just so happens to be other contracts' addresses, but the arguments can be anything, right? (hence, the usage of the `any` type)

Let me know if I got this backwards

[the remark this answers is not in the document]
andrewgordstewart: > I'm having trouble seeing this coupling.

Take [this line](https://github.com/statechannels/monorepo/pull/424/files/0c50e8a6165155e6a6cc07ce0d8db5f216e6542d#diff-9d77fecdba8b16354b02c5308eee735bR88).

It only works because the only arguments currently used in constructors are addresses of other contracts that are deployed. Here is [one](https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/contracts/ERC20AssetHolder.sol#L19) of them.

Let's say you change that constructor to 
```
    constructor(uint userId, address _AdjudicatorAddress, address _TokenAddress)
```
At that point, [this line](https://github.com/statechannels/monorepo/blob/0c50e8a6165155e6a6cc07ce0d8db5f216e6542d/packages/devtools/src/utils/ganache-server.ts#L88) no longer works: `arg` in that case some arbitrary number, and is not in `deployedArtifacts` object. So, [this check](https://github.com/statechannels/monorepo/blob/0c50e8a6165155e6a6cc07ce0d8db5f216e6542d/packages/devtools/src/utils/ganache-server.ts#L87) fails, and the `userId` that I supplied as an argument to the deployer doesn't get pushed onto the argument list that's passed to the constructor.

Essentially, the existing deployment code in `ganache-server.ts`
- assumes that constructor arguments are only addresses of contracts that have already been deployed
- doesn't actually accept the arguments themselves, but instead accepts the _names_ of the contract, which it converts to an address by looking at `deployedArtifacts`

> The function implementation would require access to a `deployedArtifacts` object holding the dependent contracts' addresses (as referenced in the second code snippet above), which is probably not something the _calling_ code should care about, right? Seems like the only thing the calling code should care about is specifying which other contracts it depends on.

The functional implementation doesn't require access to the `deployedArtifacts` object itself; it's an instruction on how to compute the arguments, given the `deployedArtifacts`. The reality is that the constructor for `EthAssetHolder` requires the address of the artifacts that have been already deployed, ie. it requires knowledge about `deployedArtifacts`. 


> in the current circumstances of deploying the contracts, it just so happens to be other contracts' addresses, but the arguments can be anything, right? (hence, the usage of the `any` type)

Yes, the arguments could be anything, but as pointed out above, if we change the constructor signature, arguments that are not contract addresses would not be passed on to the constructor.

packages/hub/readme.md
cf19drofxots: AFAICT we're not using the `.env.*.local` pattern anywhere, at least not yet.  But this should be updated anyway, since we already do check in both `.env` and `.env.development`

packages/hub/.env
```diff
@@ -4,13 +4,14 @@ SERVER_WALLET_DB_NAME=server_wallet_${NODE_ENV}
 ETH_ACCOUNT_MNENOMIC="a dozen random words"
 DEFAULT_GAS=6721975
 DEFAULT_GAS_PRICE=20000000000
-GANACHE_NETWORK_ID=123456789
+GANACHE_NETWORK_ID=9001
+CHAIN_NETWORK_ID=$GANACHE_NETWORK_ID
```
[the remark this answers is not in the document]
cf19drofxots: Right, we don't. Removed.

packages/devtools/src/types.ts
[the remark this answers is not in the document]
cf19drofxots: Yeah I had a sense of what you were referring to and I didn't like that coupling either. Updated it to be more decoupled as you suggested.

packages/nitro-protocol/test/contracts/ForceMove/conclude.test.ts
cf19drofxots: Done.

packages/wallet/src/utils/contract-utils.ts
cf19drofxots: Removed - the message behind it wasn't factual anyway.

cf19drofxots:
> The only suggestion that I feel strongly about is decoupling the deployment code from the contract constructors.

Thanks for pointing this out - updated to remove this decoupling.

PR #427, opened 2019-11-08
More JSON-RPC API changes

tomclose:
More changes to the JSON-RPC interface. In particular:

* Begins to define some errors
* Adds a `messageQueued` event, for when the wallet wants the app to relay a message
* Adds a `pushMessage` method, for when the app receives a relayed message, and needs to send it to the wallet
* Adds the `closeChannel` method
* Adds a `getAddresses` method, for the app to request one or more signing addresses from the wallet

### DIFF
--- a/packages/app-wallet-interface/source/index.html.md
+++ b/packages/app-wallet-interface/source/index.html.md
@@ -97,7 +97,7 @@ Used to push messages received from other participants into the wallet.
 {
   "jsonrpc": "2.0",
   "id": 1,
-  "result": "acknowledged"
+  "result": {"success": true}
 }
 ```
 
--- a/packages/app-wallet-interface/source/index.html.md
+++ b/packages/app-wallet-interface/source/index.html.md
@@ -36,7 +36,7 @@ API between the state channel wallet and application.
 | destination    | String | Address of EOA to receive channel proceeds                      |
 
 Note: in the future we might replace the `appId` with a `contactAddress`, which would allow
-apps to get the wallet to relay messages itself. An example `contractAddress` would be something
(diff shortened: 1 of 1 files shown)


packages/app-wallet-interface/source/index.html.md
```diff
@@ -245,14 +341,83 @@ Possible response to a `Channel Proposed` event.
...
+  }
+}
 ```
 
+> Example response
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
```
joelalejandro:
```suggestion
  "id": 0,
```

ID's should always match if present.

packages/app-wallet-interface/source/index.html.md
```diff
@@ -56,29 +56,75 @@ like `https://myserver.com/state_channel_callback`.
...
+}
+```
+
+> Example response
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": "acknowledged"
```
joelalejandro: While JSON-PRC allows for any type as a result, I recommend we stick to objects or arrays to keep the types more consistent in TypeScript.

```suggestion
  "result": { "success": true }
```

packages/app-wallet-interface/source/index.html.md
```diff
@@ -56,29 +56,75 @@ like `https://myserver.com/state_channel_callback`.
...
+}
+```
+
+> Example response
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": ["0x123..."]
```
lalexgap: Is there a reason this is an array?

PR #433, opened 2019-11-08
[wallet] Formatting changes

lalexgap:
- Enforces linting rules on `js` files
- Includes folders (like `config`) that contain `js` files outside of `src`
- Updated line width to 100


packages/wallet/package.json
```diff
@@ -19,12 +19,12 @@
 		"contracts:deploy:test": "npx etherlime deploy --compile=false --file=./deployment/deploy-test.js",
 		"contracts:deploy": "npx etherlime deploy --compile=false",
 		"deployContracts": "NODE_ENV=development npx deploy-contracts",
-		"lint:check": "yarn run tslint --config tslint.json --project .",
-		"lint:write": "yarn run tslint --config tslint.json --project . --fix",
+		"lint:check": "yarn run tslint --config tslint.json  '{src,config,deployments,gateways,migrations,scripts}/**/*.{js,ts,tsx}'",
```
andrewgordstewart: Should these things be whitelisted like this? Is there any reason not to lint every `jsx?|tsx?` file?

lalexgap: We want to avoid linting `node_modules`. However I think I can just exclude that rather than whitelisting everything.

PR #440, opened 2019-11-08
[wallet] Implement GetAddress json-rpc call in the wallet

### LINKED ISSUE #434: [wallet] Implement getAddress API Call
See https://app-wallet-interface.netlify.com/ for details of the call.

The wallet should accept a `JSON-RPC` getAddress call over `postMessage` and post a JSON-RPC result back with the address.

For the initial implementation we will only generate one address/privateKey that is used across all domains.

This will be similar to how we handle `INITIALIZE_REQUEST` currently.

lalexgap:
Fixes #434 

Adds some basic json-rpc handling over post message. I've refactored the messaging a little to hopefully make it easier to test.


packages/wallet/src/redux/sagas/message-handler.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function* messageHandler(jsonRpcMessage: string, fromDomain: string) {
+  const parsedMessage = jrs.deserialize(jsonRpcMessage);
+  switch (parsedMessage.type) {
+    case "notification":
+    case "success":
+      console.warn(`Received unexpected JSON-RPC message ${jsonRpcMessage}`);
+      break;
+    case "error":
+      throw new Error("TODO: Respond with error message");
```
lalexgap: Created #441

packages/wallet/src/redux/sagas/message-handler.ts
```diff
@@ -0,0 +1,28 @@
...
+      yield handleMessage(parsedMessage.payload as RequestPayloadObject, fromDomain);
+      break;
+  }
+}
+
+function* handleMessage(payload: jrs.RequestPayloadObject, domain: string) {
+  const {id} = payload;
+  switch (payload.method) {
+    case "GetAddress":
+      yield put(actions.addressRequest({id, domain}));
```
tomclose: It might be worth using this opportunity to start breaking out of the redux pattern. Especially in this case, where we just have a read operation, it seems like it would be possible to use the saga to read from the store and return a value, skipping the reducer etc. entirely.

packages/wallet/package.json
```diff
@@ -103,13 +107,9 @@
...
 		"@types/react-sidebar": "3.0.0",
 		"@types/reactstrap": "^8.0.0",
-		"@types/web3": "^1.0.2",
-		"@babel/runtime": "7.6.2",
-		"@statechannels/devtools": "0.1.21",
-		"@types/jest": "24.0.18",
-
 		"@types/redux-storage": "4.1.0",
 		"@types/storybook__react": "4.0.0",
+		"@types/web3": "^1.0.2",
```
tomclose: Do you know where we're still using this? Is it possible to completely remove in favour of ethers?

lalexgap: I think this can go

packages/wallet/src/redux/sagas/message-handler.ts
lalexgap: I've updated the saga to simply use a selector to get the `address` and then `fork` the messageSender so we skip the reducer entirely.

packages/wallet/src/redux/outbox/state.ts
```diff
@@ -11,7 +12,7 @@ export interface QueuedTransaction {
   processId: string;
 }
 export type DisplayOutbox = DisplayAction[];
-export type MessageOutbox = WalletEvent[];
+export type MessageOutbox = Array<WalletEvent | JsonRpcResponseAction>;
```
snario: What's the plan for dealing with these two: `WalletEvent` and `JsonRpcResponseAction`? Will there always be these two fundamental different types of outbound messages?

lalexgap: No, `WalletEvent`s will eventually be superseded by `JsonRpcResponseAction`s as we shift towards the json rpc API.  For now this is in place to not break our existing reducers.

PR #443, opened 2019-11-09
feat: Get RockPaperScissors tests setup and first test passing

Alonski:


yarn.lock
```diff
@@ -13788,18 +13788,6 @@ grpc@1.23.3:
     node-pre-gyp "^0.13.0"
```
geoknee: Should this file be changing as a part of this PR?

Issue #325, opened 2019-10-31
[web3torrent] Spike a balance-based payment model


joelalejandro:
Expands from #264.

The purpose of this issue is to spike a balance-based payment model.

In this model, the leecher pays a given amount of ETH, worth an amount X of data of the file. Once the balance is depleted, the leecher must pay again to continue.

- [ ] Define the minimum price a user can pay (in order to obtain a single, full piece)
- [ ] Connect to callbacks that monitor how much money remains in the balance to determine whether to keep sending or not

andrewgordstewart:
I've written some clarification over [here](https://github.com/statechannels/monorepo/issues/264).

Here's a possible payment/choking algorithm for a balance-based model.

### Paying
p1) Bob sends some ether along with a `request` message
p2) When he doesn't receive the requested chunk, ???.
p3) When he receives the requested chunk, he requests another chunk from the same peer.

### Choking
c1) Alice unchokes the four peers that currently have the highest pay-rate with her.
c4) When she sends a `choked` message, she returns any unused funds.

junjizhi:
@spersico Any more update on this issue? Can we move this issue to done in web3torrent project board?

spersico:
Yeah, I'll close this. This is what we are using now.

PR #448, opened 2019-11-11
[Web3Torrent] POC of buffer-based payment system

spersico:
### Description

This PR implements a buffer-based payment model for the file transfering of Web3Torrent (in which the Seeder maintains a buffer or balance of each Leecher, and answers each request accordingly, after deducing the data cost).
image

In this PoC, I make several assumptions:
- Each request is worth the same (which isn't true, as some requests are smaller than average).
- The Leecher always pays after being blocked, a predefined amount ($50, or 5 requests).
- The amount of funds is assumed valid (I just pass a stringified number, though I'm counting that at some point it's going to be a transaction hash to verify).

Furthermore, the UI has not been updated to reflect the changes, this change has been done at the **library** layer.

### Related issues

#325


joelalejandro:
Looking great!

PR #452, opened 2019-11-12
[rps] appData types and encoding/decoding

geoknee:
⚠️ base not master

### DIFF
--- a/packages/rps/src/core/app-data.ts
+++ b/packages/rps/src/core/app-data.ts
@@ -1,5 +1,5 @@
 import {Weapon} from './weapons';
-import {BigNumber, defaultAbiCoder, bigNumberify} from 'ethers/utils';
+import {BigNumber, defaultAbiCoder, bigNumberify, keccak256} from 'ethers/utils';
 import {HashZero} from 'ethers/constants';
 import {randomHex} from '../utils/randomHex';
 
@@ -66,4 +66,8 @@ export function decodeAppData(appDataBytes: string): RPSData {
     playerBWeapon: parameters[4],
     salt: parameters[5],
   };
+}
+
+export function hashPreCommit(weapon: Weapon, salt: string) {
+  return keccak256(defaultAbiCoder.encode(['uint256', 'bytes32'], [weapon, salt]));
 }
--- a/packages/rps/src/core/app-data.ts
+++ b/packages/rps/src/core/app-data.ts
@@ -20,18 +20,45 @@ export interface RPSData {
 export interface Start {
   type: 'start';
 }
+export interface RoundProposed {
+  type: 'roundProposed';
+  stake: BigNumber;
+  preCommit: string;
+}
 
-export type RoundProposed = Pick<RPSData, 'stake' | 'preCommit'> & {type: 'roundProposed'};
+export interface RoundAccepted {
+  type: 'roundAccepted';
+  stake: BigNumber;
+  preCommit: string;
+  playerBWeapon: Weapon;
+}
 
-export type RoundAccepted = Pick<RPSData, 'stake' | 'preCommit'> & {type: 'roundAccepted'};
-
-export type Reveal = Pick<RPSData, 'playerAWeapon' | 'playerBWeapon'> & {type: 'reveal'};
-
+export interface Reveal {
+  type: 'reveal';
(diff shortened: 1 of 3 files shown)


packages/rps/src/core/app-data.ts
```diff
@@ -1,26 +1,73 @@
...
 
-export interface Propose {
-  type: 'propose';
-  stake: number;
-  preCommit: string;
+export type Reveal = Pick<RPSData, 'playerAWeapon' | 'playerBWeapon'> & {type: 'reveal'};
+
+export type AppData = Start | RoundProposed | RoundAccepted | Reveal;
+
+export function toRPSData(appData: AppData): RPSData {
```
tomclose: Would it be possible to remove this function from the external app-data api? It seems that, as a user of this code, I'd rather do:
```ts
const start = { type: 'start' };

const encoded = encodeAppData(start);
```
instead of
```ts
const start = { type: 'start' };

const encoded = encodeAppData(toRPSData(start));
```
Is there a reason why a user would ever need access to the `toRPSData` method?

packages/rps/src/core/app-data.ts
```diff
@@ -1,26 +1,73 @@
...
+  preCommit: string; // bytes32
+  playerAWeapon: Weapon;
+  playerBWeapon: Weapon;
+  salt: string; // bytes32
+}
+export interface Start {
+  type: 'start';
+}
+
+export type RoundProposed = Pick<RPSData, 'stake' | 'preCommit'> & {type: 'roundProposed'};
```
tomclose: There's a lot more cognitive overhead to processing this line, than if you had written:
```ts
export type RoundProposed {
  type: 'roundProposed';
  stake: BigNumber;
  preCommit: string;
}
```
Would you be averse to replacing it?

packages/rps/src/core/app-data.ts
```diff
@@ -1,26 +1,73 @@
...
+
+export type RoundAccepted = Pick<RPSData, 'stake' | 'preCommit'> & {type: 'roundAccepted'};
```
tomclose: You need to add `bWeapon` to this. Might be worth re-writing as above too.

packages/rps/src/core/app-data.ts
geoknee: It's certainly possible and in fact I had it that way in an intermediate commit. 

I went this way because 
a) it enables the `decodeAppData` to be the inverse of `encodeAppData` :  I suppose it would be possible to do this even if the input type of `encodeAppData` was `appData` rather than `RPSData`, at the expense of quite a bit of branching logic to discriminate the union and return the correct thing.
b) I felt it was good to be quite explicit about when default values are being used for unspecified fields.

packages/rps/src/core/app-data.ts
geoknee: That's true  / I would not be.  I think my way is slightly safer, since it's harder for me to mistype the properties. But I can just make sure I don't mistype...

packages/rps/src/core/app-data.ts
tomclose: Yeah, I would expect `decode` to return an `AppData` in this case too. You'll need to switch on the state type to accomplish this. That's worth doing, as we will need to do this in the app, and otherwise that responsibility is also put on the caller.

You'd probably still end up calling `toRPSData` in the `encodeAppData` function, so you'd still be explicit about the defaults at that point. It would just be hidden in the app code.

geoknee: 👍

packages/rps/src/core/app-data.ts
geoknee: The `salt` as well, I believe.

packages/rps/src/core/app-data.ts
```diff
@@ -1,26 +1,130 @@
...
+  Start, // 0
+  RoundProposed, // 1
+  RoundAccepted, // 2
+  Reveal, // 3
 }
-
-export interface Propose {
-  type: 'propose';
-  stake: number;
+export interface RPSData {
```
tomclose: Probably don't need to export this now.

geoknee: It's used in the contract test file.

PR #458, opened 2019-11-12
[rps] [channel-client] Adds error codes

joelalejandro:
Contributes to #398.

### DIFF
--- a/packages/rps/src/utils/channelClient.ts
+++ b/packages/rps/src/utils/channelClient.ts
@@ -17,11 +17,11 @@ export type MessageNotificationName = 'MessageQueued';
 export type NotificationName = ChannelNotificationName | MessageNotificationName;
 
 export enum ErrorCodes {
-  SIGNING_ADDRESS_NOT_FOUND = -32100,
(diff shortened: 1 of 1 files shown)


packages/rps/src/utils/channelClient.ts
```diff
@@ -14,6 +16,14 @@ export type MessageNotificationName = 'MessageQueued';
 
 export type NotificationName = ChannelNotificationName | MessageNotificationName;
 
+export enum ErrorCodes {
+  SIGNING_ADDRESS_NOT_FOUND = -32100,
+  INVALID_APP_DEFINITION = -32101,
+  INVALID_APP_DATA = -32102,
+  UNSUPPORTED_TOKEN = -32103,
+  CHANNEL_NOT_FOUND = -32104,
```
tomclose: I don't think you're allowed to use these values for application errors. From [the spec](https://www.jsonrpc.org/specification):

> The error codes from and including -32768 to -32000 are reserved for pre-defined errors.

> The remainder of the space is available for application defined errors.

PR #471, opened 2019-11-13
Clearer naming of deposit related functions

snario:
This labels a few functions related to depositing to make them clearer in a multi-asset-holder world.


packages/nitro-protocol/src/index.ts
```diff
@@ -15,8 +15,9 @@ import {validTransition} from './contract/force-move-app';
 import {AllocationItem, encodeAllocation, encodeOutcome, Outcome} from './contract/outcome';
 import {Allocation, isAllocationOutcome, isGuaranteeOutcome} from './contract/outcome';
 import {State, VariablePart} from './contract/state';
+import {createERC20DepositTransaction} from './contract/transaction-creators/erc20-asset-holder';
 import {
-  createDepositTransaction,
+  createETHDepositTransaction,
   createTransferAllTransaction,
 } from './contract/transaction-creators/eth-asset-holder';
```
geoknee: As an alternative, could you have
```
import { createDepositTransaction as createETHDepositTransaction} from './contract/transaction-creators/eth-asset-holder'
```
and similarly for the ERC20 one?

My reason for suggesting is that the actual contract method is still called `deposit` in both `ETHAssetHolder` and `ERC20AssetHolder`. This way, you would be mirroring that in the `.ts` side of things (file name discriminates between ETH and ERC20, and function names are identical).

PR #474, opened 2019-11-13
Use concludePushOutcomeAndTransferAll in wallet

snario:
Addresses https://github.com/statechannels/monorepo/issues/28


packages/nitro-protocol/src/transactions.ts
```diff
@@ -52,6 +53,17 @@ export function createCheckpointTransaction(signedStates: SignedState[]): Transa
   });
 }
 
+export function createConcludePushOutcomeAndTransferAllTransaction(
```
lalexgap: We could add a test to `transactions.test` for this (although the test doesn't have a ton of value)

PR #475, opened 2019-11-13
[api] Update Errors

lalexgap:
Since I don't think we can actually verify `appData` and `appDefintion`(besides checking that there's a contract deployed at `appDefintion` I've changes those errors.

I've also added a `state transition` error on `joinChannel` as we can call `validTransition` once we have 2 states.

Also I've changed the numbering to reserve 100 error codes per a method. This allows us to introduce error codes for methods without messing with the rest.

### DIFF
--- a/packages/app-wallet-interface/source/index.html.md
+++ b/packages/app-wallet-interface/source/index.html.md
@@ -45,8 +45,8 @@ like `https://myserver.com/state_channel_callback`.
 {
   "token": "0x...", // 0x0 for ETH
   "allocationItems": [
-    {"destination": "0xa...", "amount": "0x1a"},
-    {"destination": "0xb...", "amount": "0x1a"}
+    { "destination": "0xa...", "amount": "0x1a" },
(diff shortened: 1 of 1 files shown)


packages/app-wallet-interface/source/index.html.md
```diff
@@ -275,9 +274,10 @@ Possible response to a `Channel Proposed` event.
 
 ### Errors
 
-| Code | Message           | Meaning                                                          |
-| ---- | ----------------- | ---------------------------------------------------------------- |
-| 1004 | Channel not found | The wallet can't find the channel corresponding to the channelId |
+| Code | Message                  | Meaning                                                          |
+| ---- | ------------------------ | ---------------------------------------------------------------- |
+| 1100 | Channel not found        | The wallet can't find the channel corresponding to the channelId |
+| 1101 | Invalid State Transition | The wallet contains invalid state data                           |
```
lalexgap: I think this needs a better description but I couldn't think of anything better.

PR #477, opened 2019-11-13
[wallet] Add UpdateChannel message handler

snario:


packages/wallet/src/redux/sagas/message-handler.ts
```diff
@@ -53,5 +56,22 @@ function* handleMessage(payload: jrs.RequestPayloadObject) {
...
+    case "UpdateChannel":
+      const {channelId} = payload.params;
+
+      // TODO: Error handling, what if the channel does not exist?
+      const mostRecentState = yield select(getLastStateForChannel, channelId);
+
+      const newState = createStateFromUpdateChannelParams(mostRecentState, payload.params);
+
+      // QUESTION: Should we do transition validation here?
+      // Comment (liam): I think it would be better done inside the protocol
```
lalexgap: Right now checking `validTransition` in a protocol will be painful since a calling `validTransition` on the contract (or using an `eVM` I think) will be `async`.

To do this in a reducer we'd have to add a `waitForValidation` state and a saga that would run validation and dispatch an action once `validTransition` is complete.

I think for now just doing the `validTransition` check in the saga is fine.

packages/wallet/src/redux/sagas/__tests__/message-handler.test.ts
```diff
@@ -101,4 +102,106 @@ describe("message listener", () => {
...
     });
   });
+
+  it.only("handles an update channel request", async () => {
+    const destinationA = Wallet.createRandom().address;
+    const signingAddressA = Wallet.createRandom().address;
+    const signingAddressB = Wallet.createRandom().address;
+    const destinationB = Wallet.createRandom().address;
+
+    // PUT  THIS INSIDE THE SHAREDDATA
```
lalexgap: What should be inside `SharedData`?

packages/wallet/src/redux/sagas/__tests__/message-handler.test.ts
```diff
@@ -101,4 +102,106 @@ describe("message listener", () => {
...
+      {
+        token: "0x0",
+        allocationItems: [
+          {destination: destinationA, amount: "12"},
+          {destination: destinationB, amount: "12"}
+        ]
+      }
+    ];
+
+    const testChannelStore = {
```
lalexgap: I think instead of manually constructing a channelStore, we should use `appState`,  `channelFromStates` and `setChannel` helper functions.

packages/wallet/src/redux/sagas/message-handler.ts
snario: We can write that to be sync, not async. I know how to do that.

packages/wallet/src/redux/sagas/__tests__/message-handler.test.ts
snario: Sorry that was a stale comment.

packages/wallet/src/redux/sagas/message-handler.ts
snario: https://github.com/armaniferrante/pure-evm

lalexgap: Created #478

packages/wallet/src/redux/sagas/message-handler.ts
```diff
@@ -35,9 +38,32 @@ function* handleMessage(payload: RequestObject) {
...
+      break;
+    case "UpdateChannel":
+      yield handleUpdateChannelMessage(payload);
   }
 }
 
+function* handleUpdateChannelMessage(payload: RequestObject) {
+  const {channelId} = payload.params as any;
+
+  // TODO: Error handling, what if the channel does not exist?
```
lalexgap: I think we can add a check for this and return an appropriate `Channel NotFound` error

packages/wallet/src/utils/json-rpc-utils.ts
```diff
@@ -65,3 +71,22 @@ export function createStateFromCreateChannelParams(params: JsonRpcCreateChannelP
...
 }
+
+// TODO: Error handling
+export function createStateFromUpdateChannelParams(
+  state: State,
+  params: JsonRpcUpdateChannelParams
+): State {
+  const {appData, allocations} = params;
+
+  // TODO: check for valid transition using EVM library
```
lalexgap: I don't think we need to check for a valid transition in `createStateFromUpdateChannelParams` so we can probably remove this `TODO`

packages/wallet/src/utils/json-rpc-utils.ts
```diff
@@ -65,3 +71,22 @@ export function createStateFromCreateChannelParams(params: JsonRpcCreateChannelP
...
+
+  // TODO: Check if this is a final state... I guess it couldn't be
```
lalexgap: In the case where the previous `state` `isFinal` I think `UpdateState` should fail, as the exchange of `isFinal` states will be handled by `ConcludeChannel`.

Assuming `s10` is a final state we already have, we can either:
- Have `UpdateChannel` construct `s11` with `isFinal:false` and fail `validTransition` returning an `Invalid Transition Error`
- Have `UpdateChannel` check if `s10` is final and return a bespoke error (Channel being Closed) 

I think we could either:
- cons

snario: Yeah good point, I'll add this in another PR

packages/wallet/src/redux/sagas/message-sender.ts
```diff
@@ -34,6 +34,8 @@ function* createResponseMessage(action: JsonRpcResponseAction) {
         turnNum,
         channelId
       });
+    case "WALLET.UPDATE_CHANNEL_RESPONSE":
```
lalexgap: Could we add test cases for these to `message-sender.test.ts`?

snario: I think I did add [one here](https://github.com/statechannels/monorepo/pull/477/files#diff-7ee60bb538c4c14602a93712ecb92032R296)

lalexgap: I think that test checks that the `message-handler` calls the `message-sender` with the correct `message`.

The `message-sender.test.ts` makes sure that the `message-sender` knows how to handle the message. (IE: If `message-sender` gets called with `WALLET.UPDATE_CHANNEL_RESPONSE` does it attempt to `postMessage` a proper json-rpc response. ) Mostly a sanity check.

snario: Oh sorry I misread the name of the file. Adding

lalexgap:
LGTM!

PR #482, opened 2019-11-14
[rps] Scrub out wallet

geoknee:
- [x] Login (spoof an address instead of getting one from the wallet)
- [x] join / create a game (spoof successful funding)
- [x] start playing game (choose weapons) (do not sign or validate commitments/states)
- [x] run out of funds and end game / resign (do not talk to wallet, send message directly to opponent)
- [x] challenge (button does nothing)

### DIFF
--- a/packages/rps/src/redux/message-service/saga.ts
+++ b/packages/rps/src/redux/message-service/saga.ts
@@ -1,19 +1,19 @@
-import {fork, take, call, put, select, actionChannel, takeEvery} from 'redux-saga/effects';
-import {buffers, eventChannel} from 'redux-saga';
-import {reduxSagaFirebase} from '../../gateways/firebase';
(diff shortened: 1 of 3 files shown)


packages/rps/src/redux/message-service/saga.ts
```diff
@@ -92,10 +92,13 @@ export function* sendMessagesSaga() {
     if (messageState.opponentOutbox) {
       const queue = Queue.GAME_ENGINE;
       const commitment = messageState.opponentOutbox.commitment;
-      const signature = yield signMessage(commitment);
+      // TODO:WALLET_SCRUBBED_OUT
```
geoknee: These comments should make it relatively easy to find the relevant places when we "unscrub" the wallet / replace with RPCs.

PR #497, opened 2019-11-15
[rps] Add messageQueuedListener saga

### LINKED ISSUE #455: [rps] Create MessageQueuedListener saga
The `MessageQueuedListener` saga is one of a few new sagas, which together will replace the `messaging-service`.

The saga should be initialized with a `ChannelClient`, and should bind to the `MessageQueued` event.

Every time the event is received, the saga should place a message in the `recipient`'s inbox in firebase.

Note that, we currently place messages into a _channel_ inbox, and not a per-participant inbox in firebase. Part of this issue will be to change that, which will require creating an `inbox/<user_id>` collection (or similar).

geoknee:
Closes #455 .

### DIFF
--- a/packages/rps/src/utils/channelClient.ts
+++ b/packages/rps/src/utils/channelClient.ts
@@ -275,6 +275,10 @@ export class ChannelClient {
     );
   }
 
+  unSubscribe(notificationName: NotificationName): void {
+    this.events.removeAllListeners('message');
+  }
+
   async createChannel(parameters: CreateChannelParameters) {
     await this.sendToWallet('CreateChannel', parameters);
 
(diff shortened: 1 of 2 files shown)


packages/rps/src/utils/channelClient.ts
```diff
@@ -275,6 +275,10 @@ export class ChannelClient {
     );
   }
 
+  unSubscribe(notificationName: NotificationName): void {
+    this.events.removeAllListeners('message');
+  }
+
```
geoknee: Note: This isn't currently narrowed to messages with a given `notificationName`.

PR #498, opened 2019-11-15
[rps] Add firebaseInboxListener()

### LINKED ISSUE #456: [rps] Create a FirebaseInboxListener saga
The `FirebaseInboxListener` saga is one of a couple of new sagas, which will together replace the `messaging-service`.

The `FirebaseInboxListener` saga will need to be initialized with a `ChannelClient`. It will also need to bind to updates on the user's firebase inbox (see #455). When an update is received in the firebase inbox, it will need to call `pushMessage` on the channel client.

geoknee:
Closes #456 .

TODO:
- [ ] test this by placing messages manually into firebase

### DIFF
--- a/packages/rps/src/redux/message-service/firebase-inbox-listener.ts
+++ b/packages/rps/src/redux/message-service/firebase-inbox-listener.ts
@@ -15,7 +15,7 @@ export function* firebaseInboxListener() {
   while (true) {
     const message = yield take(channel);
     const key = message.snapshot.key;
-    yield channelClient.pushMessage(message);
+    yield call(channelClient.pushMessage, message);
     yield call(reduxSagaFirebase.database.delete, `/messages/${address}/${key}`);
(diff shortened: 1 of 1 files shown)


packages/rps/src/redux/message-service/firebase-inbox-listener.ts
```diff
@@ -0,0 +1,21 @@
...
+  const channel = yield call(
+    reduxSagaFirebase.database.channel,
+    `/messages/${address.toLowerCase()}`,
+    'child_added',
+    buffers.fixed(10)
+  );
+  while (true) {
+    const message = yield take(channel);
+    const key = message.snapshot.key;
+    yield channelClient.pushMessage(message);
```
tomclose: I think this might need to be 
```ts
yield call(channelClient.pushMessage(message));
```

PR #502, opened 2019-11-17
[nitro-protocol] Replace TSLint with ESlint in Nitro Protocol

### LINKED ISSUE #35: [nitro-protocol] Switch from tslint to typescript-eslint
tslint is being deprecated in favour of typescript-eslint. If it doesn't take too much effort we should switch to typescript-eslint.

See https://github.com/typescript-eslint/tslint-to-eslint-config

Alonski:
### Description

This PR replaces TSLint with ESlint in Nitro Protocol.
TSLint is deprecated.

The rules can be changed.
What does everyone this?

### Related issues

Closes #35


packages/nitro-protocol/src/contract/outcome.ts
```diff
@@ -134,7 +132,7 @@ export function encodeAssetOutcome(allocationOrGuarantee: Allocation | Guarantee
 export type AssetOutcome = AllocationAssetOutcome | GuaranteeAssetOutcome;
 export type Outcome = AssetOutcome[];
 // ^ Note this is not the same structure
-// as the Outcome struct defined in Outcome.sol
+// As the Outcome struct defined in Outcome.sol
```
lalexgap: I'm not sure if we really need a rule to capitalize the start of comments, especially since many of comments are broken over multiple lines, so the capitalization doesn't make sense (like here).

lalexgap:
I think enforcing return types on functions might be a nice rule to enforce (https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md)

However I'm not sure of the effort to implement this rule right now. Might be worth a look if people agree?

packages/nitro-protocol/src/contract/outcome.ts
Alonski: @lalexgap I think that we should, in this case, have a multiline comment instead of multiple single line ones?

Alonski:
> I think enforcing return types on functions might be a nice rule to enforce (https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md)
> 
> However I'm not sure of the effort to implement this rule right now. Might be worth a look if people agree?

I agree. This was out of the scope of this PR. We can create an issue once this PR is merged

packages/nitro-protocol/src/index.ts
```diff
@@ -33,12 +41,12 @@ export interface SignedState {
 
 // TODO: Find a use case for this or remove.
 // @nsario I don't think we need this here -- it should be in the adjudicator state of the wallet
-// export interface ChannelStorage {
-//   turnNumRecord: BigNumberish;
-//   finalizesAt: BigNumberish;
-//   stateHash: string;
-//   challengerAddress: string;
-//   outcomeHash: string;
+// Export interface ChannelStorage {
```
tomclose: It doesn't make sense to capitalize this

PR #505, opened 2019-11-19
Start on Cypress Integration Testing

lalexgap:
Initial configuration of cypress and a start on integration testing. 

Currently integration tests are *not runnning* on circleCI due to issues I've had configuring it correctly. See #530.

Currently there's only one partial integration test: Player A opening a channel. There's some further work to be done in the API before this test can be taken further. (We need a way of pushing player B's state into the wallet).

I haven't bothered embedding the wallet in an `iframe` since `cypress` does that anyways and it will make the UI testing easier.

Unfortunately waiting for messages to be posted from the wallet is slightly painful (see `createParentPostMessagePromise` in `helpers.ts`.)


packages/nitro-protocol/src/contract/challenge.ts
```diff
@@ -1,12 +1,12 @@
-import {bigNumberify, defaultAbiCoder, keccak256, Signature} from 'ethers/utils';
+import {utils} from 'ethers';
```
lalexgap: This was changed to deal with https://github.com/ethers-io/ethers.js/issues/349 when `cypress` bundles everything using `webpack`.

packages/wallet/scripts/start.js
```diff
@@ -56,7 +56,7 @@ void (async () => {
   }
 
   // Tools like Cloud9 rely on this.
-  const DEFAULT_PORT = !!process.env.PORT ? parseInt(process.env.PORT, 10) : 3000;
+  const DEFAULT_PORT = !!process.env.PORT ? parseInt(process.env.PORT, 10) : 3055;
```
lalexgap: `ganache-deployer` starts up a web server on port 3000, so this was changed to avoid collisions.

Issue #503, opened 2019-11-18
[nitro-protocol | rps] Remove circular dependency in ganage/deployer


Alonski:
Currently, there is a circular dependency in `nitro-protocol` and `RPS`.
`RPS` depends on nitro-protocol.
`nitro-protocol/ganache/deployer.ts` depends on RPS contracts.

This causes problems during build time as `nitro-protocol` is build/linted/ran before RPS.

The fix is to move the `ganache/` functionality into a different package

lalexgap:
Currently `nitro-protocol/ganache/**` (which I will refer to as `contract-deployer`) is responsible for deploying contracts across all projects to a single ganache instance.

The `contract-deployer` currently depends on built contracts from `nitro-protocol` and `rps`.  However to start `rps` we need a started ganache instance with the `rps` and `nitro-protocol` contracts deployed.

This creates a kind of circular dependency where `contract-deployer` depends on `rps` building the contracts and `rps` depends on `contract-deployer` to deploy the contracts. This would not get solved by moving `contract-deployer` to it's own package.

One possible solution suggested by @andrewgordstewart is separating out the `rps` contracts into  a separate package.  (And possible `nitro-protocol` contracts, if the `contract-deployer` is needed by `nitro-protcol` to run contract tests?)

PR #507, opened 2019-11-19
Move Ganache Deployer to its own package

lalexgap:
This moves the contract deployment logic out of `nitro-protocol` into it's own `ganache-deployer`  package. 

I've also made some other changes:
- `ganache-deployer` exports a `startGanacheAndDeploy` function. This is useful if you want to start `ganache` in a start script (see `wallet/scripts/start.js`)
 - `ganache-deployer` includes an empty `ganache-network-context.json`. This allows for other packages to be built when ganache is not running. In the future we can use this to store test-net deployment info.
- When `ganache-deployer` starts up ganache it copies the existing `ganache-network-context.json` to `ganache-network-context.json.bak` and generates a new file containing the ganache deployment info.
- `ganache-deployer` has a `start-ganache-and-deploy` binary that can be used to trigger the original start-script. So you can have a script entry like this in another package: `   "start:ganache": "node_modules/.bin/start-ganache-and-deploy",`

NOTE:
- Since we update the git tracked file `ganache-network.json` when running `ganache` it's going to show up as a modified file in git. Once ganache is stopped the file will be reverted to the original file. 
- If `nitro-protocol` imports `ganache-deployer` we get a circular dependency (see #503).This will also occur with `rps`.  I think we'll need a way of packages being able to tell `ganache-deployer` what contracts to deploy.
- ~There also might be a CI config issue. Right now it attempts to run `start:ganache` in all packages. This works when `start:ganache` is only defined in `ganache-deployer` but will cause issues when multiple packages have a `start:ganache` script.~


andrewgordstewart:
>There also might be a CI config issue. Right now it attempts to run start:ganache in all packages. This works when start:ganache is only defined in ganache-deployer but will cause issues when multiple packages have a start:ganache script.

I think we should stop calling `lerna run start:ganache` in circle. Instead, we should run it directly from `ganache-deployer`.

packages/ganache-deployer/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "devDependencies": {
+    "typescript": "3.7.2",
+    "prettier": "1.18.2",
+    "tslint": "5.20.0",
+    "tslint-config-prettier": "1.18.0"
+  },
+  "dependencies": {
+    "@statechannels/devtools": "^0.1.21",
+    "@statechannels/nitro-protocol": "^0.0.1",
```
andrewgordstewart: I think it needs to depend on rps as well, for the time being.

packages/ganache-deployer/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import path from 'path';
+import writeJsonFile from 'write-json-file';
+import {deployContracts} from './deployer';
+
+const GANACHE_CONTRACTS_FILE = 'ganache-network-context.json';
+const GANACHE_CONTRACTS_PATH = path.join(__dirname, '../', GANACHE_CONTRACTS_FILE);
+const GANACHE_CONTRACTS_BACKUP_PATH = path.join(__dirname, '../', `${GANACHE_CONTRACTS_FILE}.bak`);
+
+export async function startGanacheAndDeploy(): Promise<GanacheServer> {
+  // TODO: This should probably merge the existing file instead of replacing it
```
andrewgordstewart: You mean it should do this?
1. read the existing file
2. Move it to `$BACKUP_PATH`
3. Merge the new network context with the existing one

lalexgap: Yup. I think ideally it should contain the existing test-net deployment info and the ganache info.

kerzhner:
Do you guys think that `ganache-deployer` should be in its own package? Or does it belong in the `devtools` package?

packages/ganache-deployer/src/index.ts
kerzhner: Should this package and the `ganache-network-context.json` file contain testnet information? From the package and files names, it seems like we should only store ganache addresses.

readme.md
```diff
@@ -100,38 +100,6 @@ To run all tests:
 yarn test
 ```
 
-### Development Flow
```
geoknee: Would be nice to still have some information here to explain how things work (for example, some distillation of the notes on this PR).

PR #510, opened 2019-11-20
Tidy up (given new approach to deployment)

geoknee:
- delete old deployment scripts
- update readmes
- Use undefined as the optional signing key in etherlime deployment scripts instead of the.signer. privatekey property of a new EtherlimeGanacheDeployer. Using `undefined` will cause etherlime to use the default signer.
Prevents confusing log output where we see the default port as well as process.env.GANACHE_PORT being logged.


packages/wallet/deployment/deploy.js
```diff
@@ -45,8 +45,7 @@ const deploy = async (network, secret, etherscanApiKey) => {
     }
   }
   const deployer = new etherlime.EtherlimeGanacheDeployer(
-    // The privateKey is optional, but we have to provide it in order to provide a port.
-    new etherlime.EtherlimeGanacheDeployer().signer.privateKey,
+    undefined,
```
geoknee: I think this file may now be unused and could be deleted.

packages/nitro-protocol/deployment/deploy.js
```diff
@@ -36,15 +36,14 @@ const migrate = async (deployer, startingMap, migrations) => {
 
 const deploy = async () => {
   const deployer = new etherlime.EtherlimeGanacheDeployer(
-    // The privateKey is optional, but we have to provide it in order to provide a port.
-    new etherlime.EtherlimeGanacheDeployer().signer.privateKey,
-    Number(process.env.GANACHE_PORT),
+    undefined,
+    Number(process.env.GANACHE_PORT)
```
geoknee: I think this file may now be unused and could be deleted.

readme.md
```diff
@@ -110,7 +110,7 @@ Run the following script _from the root of the repo_ to start a ganache server a
 yarn start:ganache
 ```
 
-The `NetworkContext` object is held at a specific path of the `nitro-protocol` package that can be referenced in other packages via `import NetworkContext from "@statechannels/nitro-protocol/ganache-ganache-network-context.json"`.
+The `NetworkContext` object is held at a specific path of the `nitro-protocol` package that can be referenced in other packages via `import NetworkContext from "@statechannels/nitro-protocol/ganache/ganache-network-context.json"`.
```
geoknee: Note that the path changes as a part of #507.

readme.md
```diff
@@ -100,6 +101,38 @@ To run all tests:
 yarn test
 ```
 
+### Development Flow
+
+The tests for the `rps`, `nitro-protocol`, `wallet` and `hub` packages need to interact with a local blockchain.
+
+Run the following script _from the root of the repo_ to start a ganache server, deploy the contracts listed [here](./packages/ganache-deployer/src/deployer.ts), and create a common `NetworkContext` object that holds the relevant contract information for the various packages that need to interface with the corresponding deployed contracts.
```
geoknee: I thought it was important to highlight that (currently) the `ganache-deployer` package actually *deploys* contracts (as its name suggests).

packages/nitro-protocol/deployment/deploy.js
geoknee: We agreed to delete this.

PR #519, opened 2019-11-20
[wallet] Add application state transition validation to wallet

snario:
Work in progress.

There seem to me several warnings in the wallets tests, but the tests pass. Need to understand that better. Also, we need figure out what the best way to store the bytecode of `ConsensusApp` is I think.


packages/wallet/src/redux/protocols/application/reducer.ts
```diff
@@ -32,11 +37,22 @@ export function initialize(
...
-    protocolState: states.waitForFirstState({channelId, privateKey, address, participants}),
-    sharedData: registerChannelToMonitor(sharedData, APPLICATION_PROCESS_ID, channelId, [])
+    protocolState: states.waitForFirstState({
+      channelId,
+      privateKey,
+      address,
+      participants,
+      bytecode
+    }),
+    sharedData: registerBytecodeForChannel(
```
lalexgap: I don't think we need to call `registerBytecodeForChannel` (or need that function?). Since `signAndInitialize`/`checkAndInitialize` accept a `byteCode` argument, they should initialize the `channelState` with the provided `byteCode`

lalexgap:
> Also, we need figure out what the best way to store the bytecode of `ConsensusApp` is I think.

We probably want to define this when the `wallet` starts up, and keep it on `sharedData` somewhere. That way if the `ConsensusApp` rules change we don't have to remember to update any hardcoded `byteCode` . (although I guess that doesn't really happen that often and will be defuncted by "null" channels) Then it could be accessed using a selector `selectors.getConsensusDataByteCode`. 

We could even do this for all `byteCode`. Then `checkAndStore(state: SharedData, signedState: SignedState)` could be updated to simply call a selector to get the `byteCode` for that `appDefinition`. This means that `byteCode` could be pulled off the `channelState`.

It does mean adding some more infrastructure though. We'd need to add a reducer to manage the bytecode portion of `sharedData` as well as a saga on start up that performs the Consensus App look up.

packages/wallet/src/redux/__tests__/state-helpers.ts
```diff
@@ -31,7 +31,14 @@ export const participants: ChannelParticipant[] = [
...
+
+export const consensusAppBytecode =
```
lalexgap: Could we move this to `constants.ts` for now (and add an issue to populate on load)?

packages/wallet/src/redux/channel-store/reducer.ts
```diff
@@ -68,10 +76,15 @@ export function signAndStore(store: ChannelStore, state: State): SignResult {
   let channel = getChannel(store, channelId);
 
   const signedState = Signatures.signState(state, channel.privateKey);
+
   if (!validTransition(channel, state)) {
     return {isSuccess: false, reason: "TransitionUnsafe"};
   }
 
+  // if (!validAppTransition(channel, signedState.state, bytecode)) {
```
lalexgap: Is there a reason we don't call the new `validAppTransition`?

packages/wallet/src/redux/protocols/reducer-helpers.ts
```diff
@@ -97,7 +97,11 @@ export function checkStates(
   states
     .filter(ss => ss.state.turnNum > turnNum)
     .map(ss => {
-      const result = checkAndStore(sharedData, ss);
+      const result = checkAndStore(
+        sharedData,
+        ss,
+        sharedData.bytecodeStorage[ss.state.appDefinition]
```
lalexgap: We should probably add a selector to `selector.ts` to get the `byteCode` from `sharedData`.

packages/wallet/src/redux/state.ts
```diff
@@ -277,8 +288,12 @@ export function checkAndInitialize(
   }
 }
 
-export function checkAndStore(state: SharedData, signedState: SignedState): CheckResult {
-  const result = checkAndStoreChannelStore(state.channelStore, signedState);
+export function checkAndStore(
```
lalexgap: Since this method already has access to the `sharedData` it doesn't have to get passed in the `byteCode`. That way protocols don't even have to know about `byteCode` at all.

We could have something like this:
```
export function checkAndStore(state: SharedData, signedState: SignedState): CheckResult {
 const byteCode = sharedData.bytecodeStorage[ss.state.appDefinition];
if (!byteCode){
    throw new Error('Uh oh');
}
const result = checkAndStoreChannelStore(state.channelStore, signedState, bytecode);
....
```

packages/wallet/src/redux/protocols/reducer-helpers.ts
snario: My mistake. Fixing in new commit...

packages/wallet/src/redux/channel-store/reducer.ts
snario: My mistake. Fixing in new commit...

PR #521, opened 2019-11-21
Changes to channel client

tomclose:
There are a lot of types defined in the channel client, which end up being a lot of information to process. This PR attempts to simplify the types and their relationships, e.g. by using less inheritance.

### DIFF
--- a/packages/rps/src/utils/channel-client.ts
+++ b/packages/rps/src/utils/channel-client.ts
@@ -5,17 +5,6 @@ export type JsonRPCVersion = '2.0';
 
 export type ChannelStatus = 'opening' | 'funding' | 'running' | 'closing';
 
-export type ChannelMethodName = 'CreateChannel' | 'JoinChannel' | 'UpdateChannel' | 'CloseChannel';
-export type AddressMethodName = 'GetAddress';
-export type MessageMethodName = 'PushMessage';
-
-export type MethodName = ChannelMethodName | AddressMethodName | MessageMethodName;
-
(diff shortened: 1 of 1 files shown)


packages/rps/src/utils/channel-client.ts
```diff
@@ -120,81 +94,64 @@ export interface JsonRPCErrorResponse<ErrorType extends JsonRPCError = JsonRPCEr
...
-
-export interface UpdateChannelRequest extends JsonRPCRequest<UpdateChannelParameters> {
-  method: 'UpdateChannel';
-}
+export type UpdateChannelRequest = JsonRPCRequest<'UpdateChannel', UpdateChannelParameters>;
 
-export interface PushMessageRequest extends JsonRPCRequest<Message> {
-  method: 'PushMessage';
-}
+export type PushMessageRequest = JsonRPCRequest<'PushMessage', Message>;
```
geoknee: Suggest replacing `Message` with `PushMessageParameters` (alias on L74)

packages/rps/src/utils/channel-client.ts
```diff
@@ -120,81 +94,64 @@ export interface JsonRPCErrorResponse<ErrorType extends JsonRPCError = JsonRPCEr
...
 }
 
-export interface PushMessageResponse extends JsonRPCResponse<PushMessageResult> {}
+export type PushMessageResponse = JsonRPCResponse<PushMessageResult>;
 
 export interface CloseChannelParameters {
   channelId: string;
 }
 
-export interface CloseChannelRequest extends JsonRPCRequest<CloseChannelRequest> {}
```
geoknee: Should this not have triggered an error?

PR #523, opened 2019-11-21
RPS storybook

geoknee:
Gets the app compiling, `start`ing and `storybook`ing


packages/rps/src/utils/rps-channel-client.ts
```diff
@@ -20,27 +24,27 @@ export class RPSChannelClient {
     bBal: string,
     appAttrs: AppData
   ): Promise<ChannelState> {
-    const participants = [
-      {participantId: aAddress, signingAddress: aAddress, destination: aAddress},
-      {participantId: bAddress, signingAddress: bAddress, destination: bAddress},
+    const participants=[
```
tomclose: It looks like there's some formatting issue here. If this is picking up a changed setting, I propose changing that setting back. It looks very cramped, and is making me feel claustrophobic ... :).

geoknee: Good catch, I had temporarily disabled prettier vscode extension (because it is causing lots of whitespace changes, because much of the rps package does not conform to the current prettier config). Would be good to run `prettier:write` on the whole package soon.

This file reformatted in 8b6044f.

PR #529, opened 2019-11-22
More RPS work

tomclose:


packages/rps/src/redux/game-v2/__tests__/player-a.test.ts
```diff
@@ -0,0 +1,214 @@
...
+describe('when joining an open game', () => {
+  it('calls createChannel and transitions to GameChosen', async () => {
+    const initialState = gameState(localStatesA.lobby);
+    const action = joinOpenGame(bName, bAddress, stake);
+    const state = channelStates.preFund0;
+
+    const { storeState } = await expectSaga(gameSaga as any, client)
+      .withReducer(reducer, initialState)
+      .dispatch(action)
+      .provide([callCreateChannel(state)])
```
geoknee: It looks like `callCreateChannel(state)` returns an array. Do you need to put that array in another array here?

packages/rps/src/redux/game-v2/__tests__/scenarios.ts
```diff
@@ -78,7 +80,7 @@ const propsA = {
   theirWeapon: playerBWeapon,
 };
 
-export const localStatesA: Record<string, LocalState> = {
```
geoknee: Why did you remove this type?

packages/rps/src/redux/game-v2/saga.ts
```diff
@@ -1,205 +1,267 @@
...
-          const state = yield call(rpsChannelClient.closeChannel, channelId);
-          yield put(updateChannelState(state));
+    const { localState, channelState }: ls.GameState = yield select(getGameState);
+
+    if (cs.isClosed(channelState) && localState.type !== 'GameOver') {
+      yield* transitionToGameOver(localState, channelState);
+      continue;
+    }
+
+    switch (localState.type) {
```
geoknee: 👍

packages/rps/src/redux/game-v2/__tests__/player-a.test.ts
tomclose: Yeah. You can `provide` many stubs at once, and each one is an array, for some reason.

packages/rps/src/redux/game-v2/__tests__/scenarios.ts
tomclose: Because I discovered it was actually removing type-safety: I could do `localStatesA.blah` and it wouldn't complain. Also, now I'm using constructors, typescript knows that all the objects are `LocalState`s anyway.

PR #531, opened 2019-11-22
[rps] Get screens for new states displaying in storybook

geoknee:
Next steps will be to add the rest of the `localState`s and make sure things display correctly.

- [x] most states wired up to old components
- [x] delete unused components
- [ ] rename used components
- [x] Add component for `ShuttingDown` state.
- [x] rebase off latest changes in `master`
- [x] Fix `GameContainer` test
- [ ] ensure all actions exist and are wired up correctly
- [ ] fix any todos

Will need to resolve discrepancy in number of screens / states vs old app.


packages/rps/src/redux/game-v2/__tests__/scenarios.ts
```diff
@@ -88,9 +88,9 @@ export const localStatesA = {
   weaponAndSaltChosen: s.weaponAndSaltChosen(propsA, salt),
   resultPlayAgain: s.resultPlayAgain(propsA, playerBWeapon, Result.YouWin),
   chooseWeapon2: s.chooseWeapon(propsA),
-  waitForRestart: s.waitForRestart(propsA),
-  shuttingDown: s.shuttingDown(propsA, 'InsufficientFundsOpponent'),
-  shuttingDownResign: s.shuttingDown(propsA, 'YouResigned'),
+  waitForRestart: s.waitForRestart(propsA, playerBWeapon, Result.YouWin),
+  shuttingDown: s.shuttingDown(propsA, playerBWeapon, Result.YouWin, 'InsufficientFundsOpponent'),
+  shuttingDownResign: s.shuttingDown(propsA, playerBWeapon, Result.Tie, 'YouResigned'),
```
tomclose: Should probably make this a completely different type of shutting down state. It's a bit nasty to have to put `Tie` in there as a default ..

tomclose: Can merge first though.

geoknee: Agreed. More recent changes show something less nasty, but still not the best solution.

PR #542, opened 2019-11-25
[hub] Replace TSLint with ESlint in Hub

Alonski:
### Description

This PR replaces TSLint with ESlint in Hub.
TSLint is deprecated.

The rules can be changed.
What does everyone think?


kerzhner:
The removal of `packages/channel-provider/tslint.json` probably doesn't belong in this PR.

packages/hub/.eslintignore
```diff
@@ -0,0 +1,11 @@
...
+/coverage/
+!.*
+/.vscode/
+/typings
+/migrations/
+**/*.json
+/website/
+/.etherlime-store/
+/lib/
+/dist/
```
kerzhner: Nit, should this contain a newline at the end of the file?

Alonski: I am on Windows 😬
Should we enforce line endings?

kerzhner: Ah gotcha. It might be nice to enforce a newline at the end of files.

Issue #546, opened 2019-11-25
Turn off capitalization of first word in comment linting rule


snario:
I doubt anyone cares very much about it, and it might be blocking on #545.

snario:
Fixed by #549

PR #549, opened 2019-11-25
Remove comment capitalization rule

snario:
I think this rule sucks. Who's with me!? #546

### DIFF
--- a/packages/channel-provider/.eslintrc.js
+++ b/packages/channel-provider/.eslintrc.js
@@ -2,7 +2,6 @@ const rules = {
   'arrow-body-style': 'error',
   'arrow-parens': ['off', 'as-needed'],
   camelcase: 'error',
-  'capitalized-comments': ['error', 'always', {ignoreConsecutiveComments: true}],
   complexity: 'off',
   'constructor-super': 'error',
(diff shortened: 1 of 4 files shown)


Alonski:
I like the rule 😭 
In English we type like this.
we don't type like this

andrewgordstewart:
> I like the rule 😭
> In English we type like this.
> we don't type like this
> 
> but if everyone else wants to type like this I am ok with that :)

I see we've got a prescriptionist over here 😛

snario:
I think it breaks some special rules that must be embedded in comments such as `/* tslint:disable */` and `/* webpackMode: 'eager' */`.

andrewgordstewart:
If `{t|e}slint` can autofix, then anyone who's indifferent shouldn't care, and people who do care are satisfied.

Then again, @snario makes a good point about special rules

Alonski:
@snario Those can be ignored.
https://eslint.org/docs/rules/capitalized-comments#ignorepattern

tomclose:
I think it's simplest just to disable this. You can write with comments with capital letters if you want, but you don't have to. :)

PR #552, opened 2019-11-25
Work on FakeChannelClient

tomclose:


packages/rps/src/core/channel-state.ts
```diff
@@ -8,8 +8,8 @@ export interface ChannelState<T = AppData> {
   status: ChannelStatus;
   aUserId: string;
   bUserId: string;
-  aDestination: string;
-  bDestination: string;
+  aAddress: string;
+  bAddress: string;
```
geoknee: We discussed that we aren't (yet) doing full justice to signing addresses vs beneficiary addresses.

packages/rps/src/redux/game-v2/__tests__/scenarios.ts
```diff
@@ -6,12 +6,12 @@ import * as s from '../state';
...
-const aUserId = 'userA';
-const bUserId = 'userB';
+export const aBal = bigNumberify(5).toString();
+export const bBal = bigNumberify(5).toString();
 export const aAddress = '0x5409ED021D9299bf6814279A6A1411A7e866A631';
 export const bAddress = '0x6Ecbe1DB9EF729CBe972C83Fb886247691Fb6beb';
-const aDestination = aAddress;
-const bDestination = bAddress;
+const aUserId = aAddress;
+const bUserId = bAddress;
```
geoknee: I would have expected "Alice" and "Bob" here?

packages/rps/src/redux/game-v2/saga.ts
```diff
@@ -34,74 +34,74 @@ const isPlayersTurnNext = (
 };
 
 export function* gameSaga(client: RPSChannelClient) {
-  while (true) {
-    yield take('*'); // run after every action
+  yield takeEvery('*', gameSagaRun, client);
```
geoknee: I believe the main difference is that now you can have several `gameSagaRun` tasks running concurrently? Is that desired?

(e.g. we can run another task without having to wait for the  previous wallet api call to resolve)

packages/rps/src/utils/channel-client.ts
```diff
@@ -29,40 +27,170 @@ export interface Allocation {
...
+  closeChannel: (channelId: string) => Promise<ChannelResult>;
+  pushMessage: (message: Message<Payload>) => Promise<PushMessageResult>;
+  getAddress: () => Promise<string>;
 }
 
+interface EventsWithArgs {
+  MessageQueued: [Message<ChannelResult>];
+  ChannelUpdated: [ChannelResult];
+}
+export class FakeChannelClient implements IChannelClient<ChannelResult> {
```
geoknee: 👍

packages/rps/src/utils/rps-channel-client.ts
```diff
@@ -20,87 +19,107 @@ export class RPSChannelClient {
...
-      appData: appAttrs,
-    };
+    const channelResult = await this.channelClient.createChannel(
+      participants,
+      allocations,
+      appDefinition,
+      appData
+    );
+
+    return convertToChannelState(channelResult);
```
geoknee: 👍

packages/rps/src/redux/game-v2/saga.ts
tomclose: Redux sagas don't run concurrently anyhow. I think it does mean that if the `gameSaga` emits events, it will run again on those events after it's done, which doesn't hurt. Mainly it seems nicer not to have a `while (true)` loop.

PR #556, opened 2019-11-26
[rps] Migrate to v2 game

geoknee:
Removes legacy game code
Renames v2 game folder
Fixes app so that is compiles, tests, storybooks and starts (involved adding actions and states)


packages/rps/src/redux/game/state.ts
```diff
@@ -1,362 +1,235 @@
...
-  name: StateName.CreatingOpenGame;
-  myName: string;
-  twitterHandle: string;
-  myAddress: string;
-  libraryAddress: string;
+  type: 'CreatingOpenGame';
+  name: string;
+  address: string;
+  // libraryAddress: string; // TODO
 }
```
geoknee: This state has been reintroduced (was in legacy code, but disappeared during refactor)

geoknee: See c63e033 , 1ecb37e and ca9befc for similar changes.

PR #561, opened 2019-11-26
Upgrade to Solidity 0.5.13

snario:


snario:
For some reason this has caused an issue with referencing the built contracts...

```
@statechannels/nitro-protocol:   5:41  error  Unable to resolve path to module '../../../../build/contracts/SingleAssetPayments.json'  import/no-unresolved
```

snario:
@kerzhner Turned out I had missed a change from `0.5.11` to `0.5.13` and I just didn't notice.

PR #565, opened 2019-11-27
[rps] Tweak start script 

geoknee:
Currently: `yarn start` triggers:
- the contracts to be deleted
- the contracts to be compiled
- ganache to be started
- serving on 3000, same as ganache-deployer (need to answer prompt every time)

On this branch
- compile contracts
- serve app on 3001

I think it makes more sense, given that deployment is currently handled by the `ganache-deployer` package, which only requires contract artifacts to exist.


packages/rps/package.json
```diff
@@ -20,7 +20,7 @@
     "postinstall": "rm -f node_modules/web3/index.d.ts",
     "prettier:check": "npx prettier --check 'src/**/*.{ts,tsx}'",
     "prettier:write": "npx prettier --write 'src/**/*.{ts,tsx}'",
-    "start": "npm run clearContracts && node scripts/start.js",
+    "start": "yarn contracts:compile && node scripts/start.js",
```
geoknee: Not sure that we ever really need `clearContracts` ? We could remove that script if it is unnecessary.

tomclose: I think it would be needed if we ever deleted a contract?

PR #568, opened 2019-11-27
Allow auto-player and/or opponent to be started from environment variables

tomclose:
Try
```
$ AUTO_OPPONENT=B yarn start
```

### DIFF
--- a/packages/rps/README.md
+++ b/packages/rps/README.md
@@ -6,11 +6,56 @@ This app is an example of a game built on our [ForceMove](https://magmo.com/forc
 
 ![splash](./screens.png 'screens')
 
-To run and/or develop the app on your machine, please:
+## Running locally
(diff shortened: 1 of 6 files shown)


geoknee:
Could you make the autoplayer open a game with a sensible amount of eth?
Screenshot 2019-11-27 at 18 03 18

For better or worse RPS looks horrible unless you are a big spender.

geoknee:
`ethers.constants.WeiPerEther` is handy for this ^

tomclose:
The auto-opponent writes directly to the `open-games` section of the store, so I don't run the `openGamesSaga` or `firebaseInboxListener` when the auto-opponent is running, to avoid conflicts. You either get to play other humans, or you get to play the auto-opponent, but not both!

PR #572, opened 2019-11-27
Update to TypeScript 3.7.2

snario:
This also updates some `redux-saga` packages because they were not compatible.

### DIFF
--- a/packages/rps/src/redux/message-service/firebase-inbox-listener.ts
+++ b/packages/rps/src/redux/message-service/firebase-inbox-listener.ts
@@ -1,5 +1,5 @@
 import { take, call } from 'redux-saga/effects';
-// import { buffers } from 'redux-saga';
+
 import { reduxSagaFirebase } from '../../gateways/firebase';
 import { RPSChannelClient } from '../../utils/rps-channel-client';
 import { Message } from '../../utils/channel-client';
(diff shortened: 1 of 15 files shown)


snario:
(work in progress)

packages/rps/src/redux/message-service/firebase-inbox-listener.ts
```diff
@@ -10,7 +10,8 @@ export function* firebaseInboxListener(client: RPSChannelClient) {
     reduxSagaFirebase.database.channel,
     `/messages/${address.toLowerCase()}`,
     'child_added',
-    buffers.fixed(10)
+    // TODO: @george why was buffers added here?
+    // buffers.fixed(10)
```
snario: `call` doesn't have an argument for this at the end, so I am a bit confused. It was throwing an error after upgrading TypeScript. @geoknee

geoknee: For context: You also updated `redux-saga`.

I'm not sure why this is here, though. Looks like it was probably my mistake: perhaps I saw `channel` on L10 and read `actionChannel` https://redux-saga.js.org/docs/api/#actionchannelpattern-buffer which does take a buffer argument.

geoknee: I think you're good to just remove it.

snario:
Blocked on #602

PR #573, opened 2019-11-28
[rps] Replace TSLint with ESlint in RPS

Alonski:
### Description

This PR replaces TSLint with ESlint in RPS.
TSLint is deprecated.

This also setups Prettier to lint SCSS and JSON files which can't be checked by ESLint, from what I could figure out.

The rules can be changed.
What does everyone think?


Alonski:
I managed to get VSCode + ESLint + Prettier playing nicely together :)
This is my `.vscode/settings.json` files:
```json
{
  "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
  "eslint.workingDirectories": [{"directory": "packages/rps", "changeProcessCWD": true}],
  "editor.formatOnSave": true
}
```
We will need to setup the `eslint.workingDirectories` option for each package.

I also downgraded Prettier in RPS to v1.18 so its consistent with other packages.
I also changed `ESLintrc` to not define Prettier rules but instead to inherit them from `.pretterrc`

tomclose:
Could you add your `vscode/settings.json` to the repo? I think we should definitely be sharing settings like these, if the development experience is broken without them.

packages/rps/package.json
```diff
@@ -15,11 +15,13 @@
...
-    "lint:check": "tslint -c tslint.json -p .",
-    "lint:write": "tslint -c tslint.json -p . --fix",
+    "lint:check": "run-s eslint:check prettier:check",
+    "lint:write": "run-s eslint:write prettier:write",
+    "eslint:check": "eslint \"**/*.{js,ts,tsx}\" --cache",
+    "eslint:write": "eslint \"**/*.{js,ts,tsx}\" --fix",
     "postinstall": "rm -f node_modules/web3/index.d.ts",
-    "prettier:check": "prettier --check '**/*.{scss,json,js,ts,tsx}' --ignore-path '.gitignore'",
-    "prettier:write": "prettier --write '**/*.{scss,json,js,ts,tsx}' --ignore-path '.gitignore'",
+    "prettier:check": "prettier --check '**/*.{scss,json}' --ignore-path '.gitignore'",
```
snario: I think `eslint` is running `prettier` so we shouldn't need special `prettier:check` scripts

packages/rps/package.json
```diff
@@ -34,7 +36,7 @@
     "precommit": "lint-staged --quiet"
   },
   "lint-staged": {
-    "{src,tests}/**/*.{ts,tsx}": "yarn lint:check"
+    "{src,tests}/**/*.{ts,tsx}": "eslint"
```
snario: Since you're using `.eslintignore` and `**/*.{js,ts,tsx}` for the linting, maybe you should use that regex here too.

packages/rps/package.json
Alonski: ESLint, from what I could manage to figure out, can't run on `scss` and `json` files
I want to run Prettier on those files though :)

packages/rps/package.json
snario: Also I recommend making `prettier` a `warn` and then doing `eslint --max-warnings=0`. I find it nice to separate linting issues versus build issues with yellow and red squiggly lines respectively in my editor.

Alonski: So I should change this back to `yarn lint:check`?

Alonski: So it checks Prettier as well?

Alonski: Oh nevermind you meant on the left side :)

packages/rps/src/utils/channel-client.ts
```diff
@@ -61,7 +61,7 @@ type UnsubscribeFunction = () => void;
 // The message Payload is designed to be opaque to the app. However, it's useful
 // to be able to specify the Payload type for the FakeChannelClient, as we'll be
 // manipulating it within the client.
-export interface IChannelClient<Payload = any> {
+export interface ChannelClient<Payload = any> {
```
geoknee: Why was this necessary?  I would have thought the prefixed `I` was pretty helpful in distinguishing this interface from some class that implements it.

geoknee:
Should we remove `tslint` and associated configuration as a part of this PR?

geoknee:
See #696

packages/rps/src/utils/channel-client.ts
Alonski: I see you resolved this :)

Alonski:
Thanks @geoknee!

PR #575, opened 2019-11-28
Remove index.ts in wallet

snario:
The `hub` is the only user of these functions; we should rip them out. There is no need for `@statechannels/wallet` to be creating dependencies on it with these.

### DIFF
--- a/packages/wallet/index.ts
+++ b/packages/wallet/index.ts
@@ -1,5 +1,0 @@
-import {unreachable} from "./src/utils/reducer-utils";
-import {addHex} from "./src/utils/hex-utils";
-import {convertAddressToBytes32} from "./src/utils/data-type-utils";
-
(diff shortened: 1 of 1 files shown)


lalexgap:
I believe the `hub` still uses the `wallet` so we need this for the time being.

snario:
@lalexgap this might be a good example where the `hub` could copy these helper methods, there are just three:

- `unreachable`
- `addHex`
- `convertAddressToBytes32`

`unreachable` is literally `x => x` and the other two just depend on `ethers` helper functions.

tomclose:
I agree that it's probably worth just copying these into the hub.

tomclose:
Sorry - closed this by mistake, and then deleted the branch on auto-pilot. Could you resubmit with the changes to the hub?

PR #578, opened 2019-11-28
Update comments about the adjudicatorStateUpdater

snario:
We do not have the luxury of on-chain data availability for our channel states (i.e., outcome, isFinalized, etc) anymore. The only piece of on-chain data we have is the current block number, which we will need to implement a different kind of saga / state updater for.

### DIFF
--- a/packages/wallet/src/redux/actions.ts
+++ b/packages/wallet/src/redux/actions.ts
@@ -143,7 +143,6 @@ export interface ChannelUpdate {
   type: "WALLET.ADJUDICATOR.CHANNEL_UPDATE";
   channelId: string;
   isFinalized: boolean;
-  balance: string;
 }
 // -------
 // Constructors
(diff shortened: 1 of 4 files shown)


andrewgordstewart:
> We do not have the luxury of on-chain data availability for our channel states (i.e., outcome, isFinalized, etc) anymore. The only piece of on-chain data we have is the current block number, which we will need to implement a different kind of saga / state updater for.

This isn't true. The `turnNumRecord` and `finalizesAt` are both viewable. There is a [`_getData`](https://github.com/statechannels/monorepo/blob/master/packages/nitro-protocol/contracts/ForceMove.sol#L693) function that fetches it for you, though it is marked as `internal`.

Using the `finalizesAt`, it's possible to narrow down the block number containing the `ForceMove` event down to a pretty small interval (say, 8 blocks) with high probability. You can binary search that interval from right to left to see when `finalizesAt` changes, and extract the event from that block.

If you missed it on both endpoints, you can fallback on a expand-then-narrow binary search in whichever direction the event was.

andrewgordstewart:
Also, circle failed on the merge. Is it possible to revisit "You must be ahead of `master` before merging", considering how frequently this is occuring?

snario:
Regarding data availability: good point, we can definitely use that technique to scan for the relevant data we would need. I guess we have partial data availability, but we still need to scan in the way you described to get the outcome.

PR #579, opened 2019-11-28
[rps] Allow playerB to cancel in WaitingRoom and disallow user actions in OpponentJoined

geoknee:
cancel0


geoknee:
in favour of #591

PR #583, opened 2019-11-28
Fix ganache-deployer's set adjudicator address for asset holder contracts

snario:
This was causing some tests to be skipped unnecessarily.


snario:
Also `transferAndWithdraw` is outdated.

PR #586, opened 2019-11-28
Add support for ERC20 deposits & withdrawals on the wallet

snario:
This is a work in progress. Some of the work will likely get split into separate PRs.

## Todo List
- [x] Figure out how to handle `AssetTransferred` events on `AssetHolder` contracts
- [x] Add support for ERC20 in ledger funding (see [here](https://github.com/statechannels/monorepo/blob/liam-erc20/packages/wallet/src/redux/protocols/existing-ledger-funding/reducer.ts#L271:L271))
- [x] Change `FundingReceivedEvent` to be `DepositedEvent`


packages/wallet/src/redux/selectors.ts
```diff
@@ -154,6 +154,18 @@ export const getChannelIds = (state: SharedData): string[] => {
...
+  Object.keys(
+    Object.values(state.channelStore).reduce(
+      (acc, val) => ({...acc, [val.signedStates[0].state.outcome[0].assetHolderAddress]: 1}),
+      {}
+    )
+  );
+
+// TODO: This is error prone I think
+export const getAssetHolderAddress = (state: SharedData, channelId: string): string =>
+  state.channelStore[channelId].signedStates[0].state.outcome[0].assetHolderAddress;
```
snario: Is there a better way to "get the asset holder" for a channel? I know there could be multiple in the future, but we only support one at the moment.

snario:
**Question** @lalexgap do we need `protocolId` and `protocolLocator` in `DepositedEvent`?

lalexgap:
> @lalexgap do we need `protocolId` and `protocolLocator` in `DepositedEvent`?

I think we still need these so the action gets routed to the correct `DirectFunding` protocol.

Although we could probably remove them just have `DepositedEvent` as a global action and a `DirectFunding` would do something if the `channelId` matches.

PR #588, opened 2019-11-28
Remove redundant action dispatching

snario:
Is this needed @lalexgap? It seems like the "process event actions" are getting dispatched many more times than need be. Also, two of them were being dispatched in the "event actions" helper function anyway.

### DIFF
--- a/packages/wallet/src/redux/sagas/adjudicator-watcher.ts
+++ b/packages/wallet/src/redux/sagas/adjudicator-watcher.ts
@@ -95,16 +95,8 @@ function* dispatchProcessEventAction(
       );
       break;
     case AdjudicatorEventType.ChallengeCleared:
-      const newTurnNumRecord = event.eventArgs[1].toNumber();
-      yield put(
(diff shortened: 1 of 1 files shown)


lalexgap:
> Is this needed @lalexgap? It seems like the "process event actions" are getting dispatched many more times than need be. Also, two of them were being dispatched in the "event actions" helper function anyway.

The original idea was that `ProccessEventAction`s would get dispatched to processes  and thus needed to have a `processId` and `protocolLocator` so they would get routed correctly.

However it looks like in refactor to `nitro-protocol` `challengeClearedEvent` and `concludedEvent` lost their `protocolLocator` /`processId`. I think those actions won't get routed correctly but I don't think any protocols really depend on them at the moment.

I would say remove them for now if no protocol depends on them. We can add them back and add process routing if/when a protocol requires it.

In the longer term I think we should look at overhauling the action routing so we don't have to do things like dispatch an action for every subscriber.

PR #591, opened 2019-11-29
[rps] Do not drop actions from gameSaga

geoknee:
gamesagafix

Additionally: 

- allow cancellation from `WaitingRoom` (player B)
- no user interaction required in OpponentJoined
- `client.updateChannel` a bit more robust


geoknee:
user-workflows-new-api

packages/rps/src/redux/game/saga.ts
```diff
@@ -135,7 +137,7 @@ function* generateSaltAndSendPropose(
   client: RPSChannelClient
 ) {
   // if we're player A, we first generate a salt
-  const salt = yield call(randomHex, 64);
+  const salt = randomHex(64);
```
tomclose: It might be useful to keep the `call` here, because then we can stub the call easily in the tests. Right now this isn't an issue, because `randomHex` isn't random, but it'll be hard otherwise to predict what the states should be.

PR #593, opened 2019-11-29
Remove @statechannels/wallet dependency by hub

snario:
Also moves some `bytes32` <--> `address` utilities to `nitro-protocol` since they are based on a convention that `AssetHolder` defines.


snario:
Actually I just recognized there are some direct file imports of some types and interfaces that the wallet uses. @kerzhner do we need to use these exact types?

e.g.,

```
import {
  RelayActionWithMessage,
  relayActionWithMessage,
  SignedStatesReceived,
  signedStatesReceived,
  strategyApproved,
  StrategyProposed
} from '@statechannels/wallet/lib/src/communication';
```

tomclose:
I've pulled this work into https://github.com/statechannels/monorepo/pull/594 btw.

snario:
@kerzhner I think it is fine if the types & interfaces are imported by direct file reference for the time being, but it requires adding `@statechannels/wallet` as a `devDependency`.

lalexgap:
These changes are in  #612

PR #594, opened 2019-11-29
Don't track ganache-network-context.json

tomclose:
**Update**: this PR has now ballooned into a large piece of work, as I kept having to add things to get the tests to pass. I think that, because this work touches most of the packages, I've ended up having to fix a number of completely unrelated problems that only manifest when tests from different packages run together :(.

Some things included:
* Changing the `ganache-deployer` to export a new deployer (the `GanacheNCacheDeployer`), so that contracts are now deployed from the packages themselves. This removes the circular dependency issue from below.
* Adding deploy scripts to each of the packages, to use the new deployer.
* Using different signers across packages, to avoid "wrong nonce" errors when multiple test suites run at once.
* Adding test versions of erc20AssetHolder and ethAssetHolder, which reference the TestNitroAdjudicator and should fix the problems seen in https://github.com/statechannels/monorepo/pull/602.
* Pulling in and extending Liam's work to remove the dependency of the hub on the wallet
* Exporting artifacts from the nitro-protocol package, so that other packages (a) are definitely using the same bytecode, and (b) don't have to do any compilation themselves
* Changing yarn commands so that contracts are recompiled as part of the test - compilation is part of the build step, which will happen beforehand.

---

This PR removes the `ganache-network-context.json` from git, which will remove the noise when committing while ganache is running.

In order to make this change, I first had to export a `getNetworkContext` function, to be used in consuming packages instead of linking directly to the `ganache-network-context.json` folder:
```ts
// instead of this:
import NetworkContext from '@statechannels/ganache-deployer/ganache-network-context.json';

// do this:
import {getNetworkContext} from '@statechannels/ganache-deployer';
const networkContext = getNetworkContext()
```
This means that projects can now compile when no `ganache-network-context.json` is present.


lalexgap:
Unfortunately I don't think this will work as it introduces a circular dependency. 

IE: `GanacheDeployer` depends on `rps` for the contract files. But `RPS` depends on `ganacheDeployer` for the deployment.

No matter what order `yarn` tries to build things it will always fail

tomclose:
Closing in favour of #612

PR #595, opened 2019-11-29
Add channelId argument to AssetTransferred event

snario:


snario:
It makes no sense to me why this test would fail.

```
@statechannels/wallet: FAIL src/contract-****s/eth-asset-holder-events.****.ts (67.358s)
@statechannels/wallet:   ● ETHAssetHolder listener › should handle an AssetTransferred event
@statechannels/wallet:     : Timeout - Async callback was not invoked within the 60000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 60000ms timeout specified by jest.setTimeout.Error:
@statechannels/wallet:       61 |   });
@statechannels/wallet:       62 | 
@statechannels/wallet:     > 63 |   it("should handle an AssetTransferred event", async () => {
@statechannels/wallet:          |   ^
@statechannels/wallet:       64 |     const channelNonce = getNextNonce();
@statechannels/wallet:       65 |     const chainId = (await provider.getNetwork()).chainId.toString();
@statechannels/wallet:       66 |     const channelId = nitroGetChannelId({
@statechannels/wallet:       at new Spec (../../node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
@statechannels/wallet:       at Suite.Object.<anonymous>.describe (src/contract-****s/eth-asset-holder-events.****.ts:63:3)
@statechannels/wallet:       at Object.<anonymous> (src/contract-****s/eth-asset-holder-events.****.ts:28:1)
```

snario:
Error in CI is related to #602

packages/wallet/src/redux/asset-holders-state/state.ts
```diff
@@ -82,3 +82,29 @@ export function recordDeposit(
     newAssetHolderChannelState
   );
 }
+
+export function recordAssetTransfer(
+  assetHoldersState: AssetHoldersState,
+  assetHolderAddress: string,
+  origin: string,
+  // @ts-ignore
+  destination: string, // TODO: Should we take this into account?
```
geoknee: It's possible in future that `destination` is removed so that it is not even part of the data emitted in the event. We might decide that it is redundant information. It was introduced mainly to make testing easier, but we could get the data from somewhere else most likely. In the case of an ERC20, for example, the Token contract itself will emit a `Transfer` event when the Tokens are transferred.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -96,7 +96,7 @@ contract AssetHolder is IAssetHolder {
             }
             if (_isExternalDestination(allocation[m].destination)) {
                 _transferAsset(_bytes32ToAddress(allocation[m].destination), payoutAmount);
-                emit AssetTransferred(allocation[m].destination, payoutAmount);
+                emit AssetTransferred(channelId, allocation[m].destination, payoutAmount);
             } else {
                 holdings[allocation[m].destination] += payoutAmount;
             }
```
geoknee: My comments above are because the event is only emitted inside a `_isExternalDestination` block. Do you think we ought to also emit an event when the holdings are updated? This could be the same event or a new one (with some appropriate name(s)). As it is now, no events will be emitted during e.g. a `transferAll` on a ledger channel that only allocates to other channels. The holdings mapping is public, however: but clients won't have a "trigger" to remind them to inspect it.

snario: We should make a decision on this at standup; it's not clear to me that it is important information for the wallet to keep track of.

packages/wallet/src/redux/actions.ts
```diff
@@ -56,7 +56,9 @@ export interface DisplayMessageSent {
 
 export interface AssetTransferredEvent {
   type: "WALLET.ASSET_HOLDER.ASSET_TRANSFERRED";
-
+  assetHolderAddress: string;
+  // This is either a `channelId` or an external destination (both bytes32).
+  origin: string;
   // This is either a `channelId` or an external destination (both bytes32).
   destination: string;
```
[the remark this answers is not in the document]
snario: Changed `origin` to `channelId`. I'll address the `destination` change in a separate PR.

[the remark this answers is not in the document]
snario: https://github.com/statechannels/monorepo/issues/605

packages/wallet/src/redux/actions.ts
```diff
@@ -56,9 +56,12 @@ export interface DisplayMessageSent {
 
 export interface AssetTransferredEvent {
   type: "WALLET.ASSET_HOLDER.ASSET_TRANSFERRED";
+  assetHolderAddress: string;
+  channelId: string;
 
   // This is either a `channelId` or an external destination (both bytes32).
```
geoknee:
```suggestion
  // This is either a `channelId` or an external destination (both bytes32):
```

PR #598, opened 2019-11-29
[Wallet] Fix Failing Test

lalexgap:
Master is currently failing due to a test failing associated with #586.  This fixes the test to get master passing again.


snario:
I think #595 also fixes this test

PR #600, opened 2019-11-30
Use TypeScript References where sensible

snario:


snario:
Test fails because of issue that is fixed in #601

PR #602, opened 2019-11-30
Attempt to fix pushOutcome test

snario:
Hey @geoknee I noticed that the `pushOutcome.test.ts` tests were failing with an `Only the NitroAdjudicator is authorized` error occurring. I realized this must have happened because I recently changed `ganache-deployer` to deploy `ETHAssetHolder` and `ERC20AssetHolder` with the real `NitroAdjudicator` contract as the `AdjudicatorAddress` argument (see https://github.com/statechannels/monorepo/pull/583). However this is problematic because the tests relied on those two asset holders allowing the `TESTNitroAdjudicator` as the `AdjudicatorAddress`.

Anyway three things:

1. I don't know how the CI did not catch this.

2. I tried adding the `constructor` from `ETHAssetHolder` to `TESTAssetHolder` and re-writing the tests (as you can see in this PR) but I am now seeing a `revert` on tests 1 and 4 for some reason that I cannot figure out.

3. In general, we need to de-couple contract deployment in the `nitro-protocol` package test environment from contract deployment for the `wallet`.. I think.. maybe.

### DIFF
--- a/packages/ganache-deployer/src/deployer.ts
+++ b/packages/ganache-deployer/src/deployer.ts
@@ -28,8 +28,6 @@ export async function deployContracts(chain: GanacheServer): Promise<object> {
     forceMoveAppArtifact,
     consensusAppArtifact,
     testForceMoveArtifact,
-    testAssetHolderArtifact1,
-    testAssetHolderArtifact2,
     trivialAppArtifact,
     countingAppArtifact,
     singleAssetPaymentsArtifact,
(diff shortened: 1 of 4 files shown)


tomclose:
On point 1: is it possible that the tests in a package [only run](https://github.com/statechannels/monorepo/blob/master/package.json#L20) if the code in that package has changed? If that's the case, I can see how this could happen, because changing the contracts that are linked in the deployer package could have broken the tests in an unrelated package.

PR #603, opened 2019-11-30
Allow WebAssembly in wallet only

snario:
Since `nitro-protocol` is ambiguous in terms of where it is meant to be used (i.e., browser or node) I'd rather not try injecting some WebAssembly in there.

The `wallet` is deliberately targeting the web. So, I'll add `pure-evm` to that package only.

However at the moment I'm not 100% sure where the best place to import `pure-evm` even is... some options are:

- In the same place that storage is loaded (i.e., `store.ts`) since it needs to do an async fetch
- In the `valid-transition.ts` file itself (which may cause a race condition-- this is where it is now)

Also, `jest` uses a different testing environment than the browser (i.e., `jsdom`) and so it can do a direct, synchronous import of the WebAssembly instance. However this is kind of ugly.

I want to try to find a way whereby I can decide at startup-time (either of the web app or of the jest suite of tests) how `pure-evm` is imported, but I'm not sure how to do that. Help appreciated!

**Note** I also edited the `pure-evm` patch file in this PR to reflect changes made in my recent PR to `pure-evm` which you can see here: https://github.com/armaniferrante/pure-evm/pull/2


lalexgap:
> * In the same place that storage is loaded (i.e., `store.ts`) since it needs to do an async fetch
> * In the `valid-transition.ts` file itself (which may cause a race condition-- this is where it is now)

Would it make sense to do this in a saga? It seems a natural fit for an async task. The saga could easily run on start up.

> Since `nitro-protocol` is ambiguous in terms of where it is meant to be used (i.e., browser or node) I'd rather not try injecting some WebAssembly in there.

Agreed!
> The `wallet` is deliberately targeting the web. So, I'll add `pure-evm` to that package only.
> 
> However at the moment I'm not 100% sure where the best place to import `pure-evm` even is... some options are:
> 
> * In the same place that storage is loaded (i.e., `store.ts`) since it needs to do an async fetch
> * In the `valid-transition.ts` file itself (which may cause a race condition-- this is where it is now)
> 
> Also, `jest` uses a different testing environment than the browser (i.e., `jsdom`) and so it can do a direct, synchronous import of the WebAssembly instance. However this is kind of ugly.
> 
> I want to try to find a way whereby I can decide at startup-time (either of the web app or of the jest suite of tests) how `pure-evm` is imported, but I'm not sure how to do that. Help appreciated!

We could define `PureEvm` as a global var in our `globals.d.ts` and have either a saga (for the web) or `testSetup.ts` (for jest) to define it.

The downside of this is that our code relies on a global variable and it might not be obvious where it's defined.

snario:
Made `PureEVM` a `wallet` global variable.

packages/wallet/package.json
```diff
@@ -68,7 +68,7 @@
     "object-hash": "^2.0.0",
     "promise": "8.0.1",
     "prop-types": "^15.6.2",
-    "pure-evm": "^0.1.2",
+    "pure-evm": "0.1.3",
```
snario: This was the change, by the way: https://github.com/armaniferrante/pure-evm/commit/c316a802e3305bb6f38557918bdf4da32e2433bc

Issue #569, opened 2019-11-27
vscode settings to show eslint errors/warning inline


kerzhner:
The following should be added to `settings.json`
```
"eslint.validate": [
    "typescript",
  ]
```

This setting should work if the editor is opened in the `monorepo` directory or in the individual package (such as the `wallet`).

Issue #585, opened 2019-11-28
Development pain-points


andrewgordstewart:
After stand-up this morning, we discussed some pain points that I found trying to develop in the mono-repo after a bit of a hiatus.

Here they are written down. Perhaps we can make issues for the pain points that we want to deal with.

- [x] I ran `yarn && yarn contract:compile` from `nitro-protocol` on master, and it failed. I needed to run `yarn` from the monorepo Misleadink .
- [x] .env.defaults is misleading
- [ ] .env.local isn't loaded by default when I run `yarn contract:compile`
- [ ] env variables weren't in .env
- [ ] I ran `yarn test` from `nitro-protocol`, and got a confusing error. The docs say you need to start ganache, so that's good
- [ ] I ran `yarn start:ganache` from `nitro-protocol`, doesn't work.
- [ ] I ran `yarn start:ganache` from monorepo root, and the tests killed the ganache server
- [ ] network-context.json is checked in, and gets changed when you start ganache
- [x] I saved a file from the devtools package, and prettier formatted my file on save. When I tried to commit that file, `yarn lint:check` failed

Alonski:
Taken from: https://github.com/statechannels/monorepo/pull/573#issuecomment-559705782

I managed to get VSCode + ESLint + Prettier playing nicely together :)
This is my .vscode/settings.json files:

{
  "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
  "eslint.workingDirectories": [{"directory": "packages/rps", "changeProcessCWD": true}],
  "editor.formatOnSave": true
}
We will need to setup the eslint.workingDirectories option for each package.

I also downgraded Prettier in RPS to v1.18 so its consistent with other packages.
I also changed ESLintrc to not define Prettier rules but instead to inherit them from .pretterrc

PR #607, opened 2019-12-02
[monorepo] Add VSCode settings

Alonski:
### Description

This unignores VSCode settings in the monorepo.
The settings are currently only related to ESLint.

### Related issues

#585 #573 

Fixes: #569


snario:
`eslint` is not installed on some of those packages, yet.

Alonski:
Hmm I'm not sure it matters but I'll update this tomorrow

snario:
It matters because our editors will start forcing us to format any files in all of those folders because of `editor.formatOnSave`. This happened to me when I checked out the branch. It caused conflicts with the linters in those packages that I couldn't resolve (because of the `editor.formatOnSave`) until I changed branched and restarted VS Code.

snario:
Could you also address / reference this issue here: https://github.com/statechannels/monorepo/issues/569

tomclose:
It might be worth adding the `.vscode/settings` on a project-by-project level, as you convert to eslint. For example, #573 is currently waiting on vscode settings to be merged. It would be fine to enable `editor.formatOnSave` there, as it's been converted. Then we could get that merged.

packages/devtools/.eslintrc.js
```diff
@@ -1,185 +1,185 @@
 const rules = {
```
snario: The changes here seem unrelated. I'm guessing they're because of my comment in the other PR? If so, I meant that we should avoid having literal object keys and quoted object keys. e.g., `{ "a": 1, b: 1 }`

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
```
snario: What's the point of having this root config and the package configs?

snario:
I get that @tomclose mentioned the idea of having per-package `settings.json` files, but two questions:

1. Is that even necessary? It seems like `"eslint.workingDirectories"` handles that.

2. If it is necessary, why do we need a repo level config?

.vscode/settings.json
andrewgordstewart: I believe that a root config is needed if I open the project from the root directory.

Alonski:
@snario Some people like to open packages standalone and work inside them and not in the whole repo. 
The root level settings is for people working with the whole monorepo open.
Each package level settings is for people working with a lone package open,

packages/devtools/.eslintrc.js
Alonski: This isn't related to the PR... But I ran format on this file and it changed :)
I can roll this back if its problematic to include

Alonski:
@geoknee Done :)

PR #611, opened 2019-12-02
[wallet] Replace TSLint with ESlint in Wallet

Alonski:
### Description

This PR replaces TSLint with ESlint in Wallet.
TSLint is deprecated.

The rules are currently **empty**, and will be updated in a future PR.


.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
+  "eslint.workingDirectories": [
+    {"directory": "packages/rps", "changeProcessCWD": true},
+    {"directory": "packages/jest-gas-reporter", "changeProcessCWD": true},
+    {"directory": "packages/wallet", "changeProcessCWD": true}
+  ],
+  "editor.formatOnSave": true
+}
```
snario: This file seems like it should not be in this PR.

.gitignore
```diff
@@ -20,7 +20,6 @@ coverage
 # misc
 package-lock.json
 .DS_Store
-.vscode
```
snario: Can remove this as it was added in #609

packages/wallet/.eslintignore
```diff
@@ -0,0 +1,11 @@
...
+/coverage/
+!.*
+/.vscode/
+/typings
+/migrations/
+**/*.json
+/website/
+/.etherlime-store/
+/lib/
+/dist/
```
snario: Many of these probably overlap with (a) `.gitignore` regexes and (b) other `.eslintignore` in other packages. Can we consolidate?

packages/wallet/.eslintrc.js
```diff
@@ -0,0 +1,241 @@
+const rules = {
+  "arrow-body-style": "error",
+  "arrow-parens": ["off", "as-needed"],
+  camelcase: "error",
```
snario: Can we be consistent with using quoted keys versus non-quoted keys?

packages/wallet/.eslintrc.js
```diff
@@ -0,0 +1,241 @@
+const rules = {
```
snario: You should edit this file to have the exact same settings as the existing `tslint.json` file so that when you run `yarn lint:write` no changes are made to any files. Then, we can begin toggling linting _settings_ in separate PRs instead of editing the settings in the same PR as introducing a new linting tool.

snario: Many of these rules seem to be duplicates of rules in other packages. Can we consolidate them in root `.eslintrc.js` file please?

packages/wallet/.eslintrc.js
```diff
@@ -0,0 +1,241 @@
...
+    react: {
+      version: "detect"
+    }
+  },
+  globals: {
+    // Maybe we should be importing these?
+    ethereum: "readonly",
+    web3: "readonly",
+    process: "readonly"
+  },
```
snario: Why do `globals` need to be defined twice? We also define them in `globals.d.ts` for example.

packages/wallet/.eslintrc.js
```diff
@@ -0,0 +1,241 @@
...
+      },
+      rules: {
+        "@typescript-eslint/no-var-requires": "off"
+      },
+      globals: {
+        jest: "readonly"
+      }
+    }
+  ]
+};
```
snario: In general this config seems very long and complicated with many opinionated rules added that I don't think we have discussed as a team. I recommended trying to find ways to shorten it. For example, in the root `tslint.json` we currently extend some industry standard linting rules:

```js
  "extends": [
    "tslint:recommended",
    "tslint-plugin-prettier",
    "tslint-config-prettier"
  ]
```

packages/wallet/package.json
```diff
@@ -16,12 +16,10 @@
     "build": "run-s clearContracts build:typescript build:webpack",
     "clearContracts": "rm -rf ./build/contracts",
     "contracts:compile": "node ./bin/compile.js",
-    "lint:check": "yarn run tslint --config tslint.json  './**/*.{ts,js,tsx,jsx}'",
-    "lint:write": "yarn run tslint --config tslint.json  './**/*.{ts,js,tsx,jsx}' --fix ",
+    "lint:check": "npx eslint \"*/**/*.ts\" --cache",
```
snario: I would be careful with this `--cache` option. When testing this branch I noticed it caused issues when deleting built contract artifacts, running the linter, and then rebuilding them, and running it again. The linter would fail despite the contract artifacts existing.

packages/wallet/package.json
```diff
@@ -16,12 +16,10 @@
...
+    "lint:write": "npx eslint \"*/**/*.ts\" --fix",
```
snario: I would avoid adding another use case of `npx`, see #617

.vscode/settings.json
tomclose: I think this is included because I asked for it [here](https://github.com/statechannels/monorepo/pull/573#issuecomment-559740236). I'm keen to make sure we commit any settings that are required for the setup to work, so that everything Just Works out of the box. We might be able to get away without this though, depending on whether it's truly required or not.

snario: There is a separate PR for this, though: https://github.com/statechannels/monorepo/pull/607

snario: And the one in this PR seems to be making changes related to `rps` and `jest-gas-reporter`, too.

.gitignore
tomclose: Just checking - are you saying that this change shouldn't be in the PR, as it was part of a previously merged PR (and we should just rebase)? Or are you saying that you want `.vscode` to be in `.gitignore`?

snario: 👇
>  this change shouldn't be in the PR, as it was part of a previously merged PR

packages/wallet/.eslintrc.js
Alonski: @snario This is quite hard as ESLint and TSLint work differently and have different rules.
I have been trying to get all packages to use the same or similar configs but there will be changes...

Alonski: This is the plan once all packages are using ESLint

packages/wallet/.eslintrc.js
Alonski: Yes thanks

packages/wallet/.eslintrc.js
Alonski: We define them as vars. This allows Typescript to understand them. ESLint though doesn't understand that they are globals.

packages/wallet/.eslintrc.js
Alonski: I used this to try and get the same rules:
https://github.com/typescript-eslint/tslint-to-eslint-config

I could just switch to using something similar to what you recommended but it might not have exactly the same settings.

packages/wallet/package.json
Alonski: I agree. If that happens I just run `lint:write` :)
Cache can save a lot of time when linting constantly. 
This is open for discussion.

packages/wallet/.eslintrc.js
andrewgordstewart: It doesn't seem necessary that all packages use ESLint. Those which use ESLint can extend the root `.eslintrc.js`; those which don't, would not.

andrewgordstewart:
> See my change requests. Importantly, I think you should try to structure this PR such that the only changed files are the following:
> 
> 1. `packages/wallet/.eslintignore`
> 2. `packages/wallet/.eslintrc.js`
> 3. `packages/wallet/package.json`
> 4. `packages/wallet/tslint.json`

Given that there is no 1-1 mapping between eslint and tslint rules, as @Alonski points out, this request could be
- feasible
- infeasible
- impossible (assuming we actually use eslint in this PR)

I think this PR should _limit_ changes outside of 1-4 as much as possible, but I don't think the above should be a hard requirement.

packages/wallet/.eslintrc.js
snario: What do you mean they work differently? Are there certain code styles that `tslint` can check for that `eslint` cannot?

snario: Why does the plan need to blocked on that? All you need are two packages using `eslint` for there to warrant a root level `.eslintrc.js` file.

snario:
> I think this PR should limit changes outside of 1-4 as much as possible, but I don't think the above should be a hard requirement.

I agree

Alonski:
I just ran `lint:check` on the wallet package with this config:
https://gist.github.com/Alonski/cae5ce15ea0c04351918d6a7ec509a37
These are the errors I got:
https://gist.github.com/Alonski/e64bf4e2e684631f1303b76b12363ba4

packages/wallet/.eslintrc.js
geoknee: The agreed approach now is to use empty rules in this PR, and tackle reintroduction of rules in a future PR.

PR #612, opened 2019-12-02
[wallet/nitro-protocol] Deploy to Independent Ganache Instance

lalexgap:
The `wallet` and `nitro-protcol` package have been updated to a new pattern of deploying contracts to their own instance of ganache.

This incorporates changes from #602 and #593, so they can be closed in favour of this.

This is the approach I've taken to allowing packages to start up their own ganache instance:
- Each package will have it's own `deploy.js` that deploys the contracts it needs (see `wallet/deployment/deploy.js`
- Any start scripts should startup a `GanacheServer` from `devtools` and call `deploy` from `deploy.js` and define the contract address in environment variables(see `wallet/scripts/start.js`)
- Tests that depend on contracts should startup a `GanacheServer` from `devtools` and call `deploy` from `deploy.js` (see `jest/contract-test-setup.ts`) and should kill the ganache server on teardown (see `jest/contract-test-teardown.ts`)
- Tests that do not rely on contracts should either mock out functions that are called to return contrat addresses (see `jest/test-setup.ts`) or just define the contract address environment variables to have some address in them.


packages/nitro-protocol/src/index.ts
```diff
@@ -1,4 +1,11 @@
 import {Signature} from 'ethers/utils';
+import NitroAdjudicatorArtifact from '../build/contracts/NitroAdjudicator.json';
```
andrewgordstewart: This approach, with the current .gitignore, means this package will be broken unless `yarn build` is run, since these files don't exist.

I'm not sure if this is the best approach:
- it's appealing _not_ to check in an artifact that's a pure function of the contracts (which are checked in)
- it's _not_ appealing that the project is broken (as in `npx tsc` fails) when I first clone the repo

If we remove `../build/contracts` from .gitignore, I recommend scrubbing the artifacts from anything other than: `abi`, `contractName` and `bytecode`

packages/wallet/jest/contract-test-setup.ts
```diff
@@ -0,0 +1,29 @@
...
+const {deploy} = require("../deployment/deploy");
+
+export default async function setup() {
+  const ganacheServer = new GanacheServer(
+    Number(process.env.GANACHE_PORT),
+    Number(process.env.CHAIN_NETWORK_ID)
+  );
+  await ganacheServer.ready();
+
+  const {
```
andrewgordstewart:
```suggestion
  const deployedArtifacts = await deploy();
```

packages/wallet/jest/contract-test-setup.ts
```diff
@@ -0,0 +1,29 @@
...
+    CONSENSUS_APP_ADDRESS,
+    TRIVIAL_APP_ADDRESS,
+    NITRO_ADJUDICATOR_ADDRESS,
+    ETH_ASSET_HOLDER_ADDRESS,
+    TEST_TOKEN_ADDRESS,
+    TEST_TOKEN_ASSET_HOLDER_ADDRESS
+  } = await deploy();
+
+  process.env.CONSENSUS_APP_ADDRESS = CONSENSUS_APP_ADDRESS;
```
andrewgordstewart:
```suggestion
  process.env = { ...process.env, deployedArtifacts };
```
(paired with the above suggestion)

packages/wallet/jest/contract-test-setup.ts
```diff
@@ -0,0 +1,29 @@
...
+  process.env.TRIVIAL_APP_ADDRESS = TRIVIAL_APP_ADDRESS;
+  process.env.NITRO_ADJUDICATOR_ADDRESS = NITRO_ADJUDICATOR_ADDRESS;
+  process.env.ETH_ASSET_HOLDER_ADDRESS = ETH_ASSET_HOLDER_ADDRESS;
+  process.env.TEST_TOKEN_ADDRESS = TEST_TOKEN_ADDRESS;
+  process.env.TEST_TOKEN_ASSET_HOLDER_ADDRESS = TEST_TOKEN_ASSET_HOLDER_ADDRESS;
+
+  (global as any).__GANACHE_SERVER__ = ganacheServer;
```
andrewgordstewart: I had a really hard time assigning `__GANACHE_SERVER__` to `global` in the past. I wish I had thought of this :P

packages/wallet/scripts/start.js
```diff
@@ -83,7 +84,27 @@ void (async () => {
...
 
   console.log(chalk.cyan("Starting ganache and deploying contract..."));
-  ganacheServer = await startGanacheAndDeploy();
+  ganacheServer = new GanacheServer(
+    Number(process.env.GANACHE_PORT),
+    Number(process.env.CHAIN_NETWORK_ID)
+  );
+  await ganacheServer.ready();
+
+  const {
```
andrewgordstewart: I wonder if it's worth it to consolidate this logic and the logic in contract-test-setup.

packages/nitro-protocol/src/index.ts
lalexgap: @tomclose and I had a quick chat about this and he raised a valid point about treating the contract files similar to our typescript files, as in they both must be compiled before the package can be used by another package.

> * it's _not_ appealing that the project is broken (as in `npx tsc` fails) when I first clone the repo

I think our approach here is that `npx tsc` is not considered "building" `nitro-protocol`. We expect people to run `yarn build` if they want to build the `nitro-protocol` package.

The advantage of this is that it allows consumers of `nitro-protocol` to not have to worry about compiling solidity files. 

Of course the downside is that building `nitro-protocol` now has the overhead of compiling the contracts.

andrewgordstewart: Yeah, I'm in favour of **not** checking them in. If you change your notion of what "broken" means, then it also seems fine to not check them in.

I think we're better off punting this issue, but before publishing to npm, we probably want to run some "production` build step, that
1. compiles the contracts
2. strip out unnecessary data from the artifacts
3. record the deployed addresses in a (published) network context file

packages/wallet/jest/test-setup.ts
```diff
@@ -0,0 +1,25 @@
...
+import Adapter from "enzyme-adapter-react-16";
+
+// Load environment variables from .env
+import "../config/env";
+import {AddressZero} from "ethers/constants";
+
+Enzyme.configure({adapter: new Adapter()});
+
+// Mock out util functions that require a networkContext
+jest.mock("../src/utils/contract-utils.ts", () => ({
```
tomclose: It might be nicer to set all the contract addresses to `AddressZero` in the `.env` file, along with an explanation that these are replaced by the deploy script. That way seems a little less magic.

andrewgordstewart: :+1:

packages/hub/src/communication/index.ts
```diff
@@ -0,0 +1,118 @@
+import {SignedState} from '@statechannels/nitro-protocol';
```
snario: Do we need to import _all_ of these? The entire file?

packages/nitro-protocol/contracts/test/TestErc20AssetHolder.sol
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+  * @dev This contract is a clone of the ERC20AssetHolder contract. It will be initialized to point to the TestNitroAdjudicator.
+*/
+contract TestErc20AssetHolder is ERC20AssetHolder {
+    constructor(address _AdjudicatorAddress, address _TokenAddress)
+        public
+        ERC20AssetHolder(_AdjudicatorAddress, _TokenAddress)
+    {}
+}
```
snario: See #602 which also fixes the issue that led to this file being written.

packages/nitro-protocol/test/contracts/TrivalApp/validTransition.test.ts
```diff
@@ -29,7 +29,7 @@ function getRandomVariablePart(): VariablePart {
 }
 
 beforeAll(async () => {
-  trivialApp = await setupContracts(provider, TrivialAppArtifact);
+  trivialApp = await setupContracts(provider, TrivialAppArtifact, process.env.TRIVIAL_APP_ADDRESS);
```
snario: If we're going to start using environment variables to share addresses for things in tests (which by the way seems unnecessary for unit tests as we discussed on standup), can we add a file which reads them from the environment and provides typed + validated accessors? For example, `environment.ts` might read `process.env. TRIVIAL_APP_ADDRESS` and if it sees `undefined` log an error in the console?

snario:
Regarding this comment... @lalexgap 

> Tests that depend on contracts should startup a GanacheServer from devtools and call deploy from deploy.js (see jest/contract-test-setup.ts) and should kill the ganache server on teardown (see jest/contract-test-teardown.ts)

I think it's important for us to recognize that the starting of ganache & deploying the contracts steps can be independent. For example in the past I would start ganache in a jest setup script and stop it in the tear down script, but then in each individual test suite in the `beforeAll` I would deploy contracts as needed for better unit test isolation.

packages/hub/src/communication/index.ts
tomclose: I think so. The way I constructed this file (on my branch, which Alex copied across) was to add the minimum and then keep adding until that compiled. We should be able to reduce this a lot as we firm up the wire protocol, but I think this is what's needed to decouple in the short-term.

packages/nitro-protocol/contracts/test/TestErc20AssetHolder.sol
tomclose: As far as I can see, the separate ganache instances also solve this, so these changes might not be necessary here at all.

PR #614, opened 2019-12-03
Switch wallet tsconfig to use esnext module

snario:
In order to use dynamic imports, we need to use the `import(...)` syntax. TypeScript doesn't know how to handle this with `commonjs`, so we need to use `esnext`.

In order to use WebAssembly in the browser we need to use dynamic imports.

In order to use `pure-evm`, we need to use WebAssembly, which means we need dynamic imports.

The integration tests use `ts-node`, which uses Node.js, which can't handle ES modules. See https://github.com/TypeStrong/ts-node/issues/436.

So, I need to use `commonjs` for the `puppeteer/funding.ts` test script. Note that this is _just_ for the test script, which launches a browser, which runs the webpack output, which was built from the TypeScript source compiled using `esnext`.

### DIFF
--- a/packages/wallet/config/webpack.config.js
+++ b/packages/wallet/config/webpack.config.js
@@ -273,8 +273,18 @@ module.exports = function(webpackEnv) {
         PnpWebpackPlugin.moduleLoader(module)
       ]
     },
+    
+    // Webpack has some issue with ts-loader, transpileOnly, and esnext
+    // https://github.com/TypeStrong/ts-loader/issues/751#issuecomment-376318718
+    stats: {
+      warningsFilter: /export .* was not found in/,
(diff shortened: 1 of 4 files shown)


snario:
I noticed `webpack`'s `stats.warningsFilter` doesn't seem to work. Also it doesn't work in the browser at all: https://github.com/webpack-contrib/webpack-hot-client/issues/94. That might be an issue.

Also `jsonrpc-lite` started raising this warning: https://github.com/teambition/jsonrpc-lite/issues/15

snario:
Ugh I think this might be blocked on #593 because the `hub` uses `ts-node` and imports the `wallet` for certain things.

PR #615, opened 2019-12-03
Hub: postgres db setup

kerzhner:
In this PR, we are:
1. Moving away from knex configuration `connection` string as it has unexpected behavior. When the `connection` string is invalid (such as `RequiredInLocalDotEnv/server_wallet_test`), knex defaults to a valid database connection with a default postgres server and database. This can lead to tests using an unexpected database.
1. Cleaning up duplicate and stale yarn commands.


packages/hub/config/knexfile.js
```diff
@@ -0,0 +1,38 @@
+const path = require('path');
```
snario: Why the switch to JavaScript over TypeScript?

kerzhner: Good question. I tried moving `knexfile.ts` to the `config` directory from the root package directory and saw this error `File 'statechannels/monorepo/packages/hub/config/knexfile.ts' is not under 'rootDir' 'statechannels/monorepo/packages/hub/src'. 'rootDir' is expected to contain all source files.`

`knexfile.ts` probably doesn't belong in the `src` directory. One option is to change the tsconfig settings. I adopted the path of least resistance and just made the file a `js` file since it doesn't use any typescript constructs anyways.

snario: Does it need to be in the root directory? Seems like there is a `typeof import("@types/knex").Config` type we could typecast the config as.

kerzhner: knexfile does not need to be in the root package directory, nor is it in the pull request. But, Typescript compiler is unhappy with any typescript files outside of the `tsconfig` rootDir, which is set to `src`.

Good call on knex types. I will look into those.

PR #621, opened 2019-12-03
Merge master into tc-network-context

snario:
Handles merge conflicts for `tc-network-context`.


lalexgap:
I think we can just obsolete this in favour of #612 if people are ok with that.

tomclose:
Yep - happy to discard this and `tc-network-context` in favour of #612

PR #622, opened 2019-12-03
[rps] Fixes to the endgame

geoknee:
- [x] Insufficient funds flow works
- [x] Resigning flow works
- [x] And other bug fixes. 
- [x] Diagram updated:
- [ ] update again to show resigning flow
user-workflows-new-api


packages/rps/src/containers/GameContainer.tsx
```diff
@@ -65,64 +71,73 @@ function RenderGame(props: GameProps) {
...
       );
+    case 'Resigned':
+      const ourTurn =
+        localState.player === 'A'
+          ? Number(channelState.turnNum) % 2 !== 0
+          : Number(channelState.turnNum) % 2 === 0;
+      const iResigned =
+        (ourTurn && channelState.status === 'running') ||
+        (!ourTurn && channelState.status === 'closing') ||
+        (ourTurn && channelState.status === 'closed');
```
tomclose: I feel like this should be stored in the state. It seems like too much logic to be in the view layer.

geoknee: Addressed in 4d98af6

packages/rps/src/redux/login/saga.ts
```diff
@@ -66,5 +67,6 @@ export default function* loginRootSaga() {
 }
 
 function getLibraryAddress() {
-  return process.env.RPS_CONTRACT_ADDRESS;
+  return '0x0'; // TODO fix this
+  // return process.env.RPS_CONTRACT_ADDRESS;
```
geoknee: I couldn't get this to work :-(

tomclose: I think you need this: https://github.com/statechannels/monorepo/blob/master/packages/wallet/config/env.js#L47-L53

PR #625, opened 2019-12-03
Update tsconfig.json of devtools

snario:
When I updated the monorepo `tsconfig` earlier this year I mentioned I would later go through each child package and try to merge any duplicate or unneeded keys to minimize complexity. Devtools was added later, so just doing this now.

### DIFF
--- a/packages/devtools/src/config/env.ts
+++ b/packages/devtools/src/config/env.ts
@@ -17,7 +17,7 @@ export function configureEnvVariables(monorepo = true) {
     NODE_ENV !== 'test' && `.env.local`,
     `.env.${NODE_ENV}`,
     '.env'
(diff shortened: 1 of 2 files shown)


packages/devtools/src/config/env.ts
```diff
@@ -17,7 +17,7 @@ export function configureEnvVariables(monorepo = true) {
     NODE_ENV !== 'test' && `.env.local`,
     `.env.${NODE_ENV}`,
     '.env'
-  ].filter(Boolean);
+  ].filter((x): x is string => !!x);
```
snario: Because https://github.com/microsoft/TypeScript/issues/16655#issuecomment-494907057

PR #626, opened 2019-12-03
Add typesync to monorepo

snario:
Adds [typesync](https://github.com/jeffijoe/typesync) to the repo. This adds missing types packages that we didn't explicitly include previously. It usually means the external packages we used, if not written in TypeScript, would give us `any` types. Now we get nice typed packages. It also runs on `postinstall` incase we added a new package but forgot the types.


snario:
`wallet` tests seem to just error with `@statechannels/wallet: Killed`... 😕

snario:
Ah that means ["out of memory"](https://support.circleci.com/hc/en-us/articles/115014359648-Exit-code-137-Out-of-memory).

snario:
This PR changes many packages, so when the tests run on all of them in parallel CircleCI cannot handle it, it seems. I changed the resource class to be `xlarge`, which I can change back if that passes. Otherwise I'll split this up into many PRs.

PR #630, opened 2019-12-03
[wallet] Use correct polyfill technique for wallet jest tests

snario:
Using the `globals` field in the jest config was the wrong approach for providing `window.PureEVM` in the jest tests for the `wallet`. Instead, this should have been a polyfill provided to `setupTests`. This fixed that.


snario:
As a note, this was causing `yarn test` to fail on the `wallet` because parallel tests are unable to use these global values.

PR #631, opened 2019-12-03
Cached deployer

tomclose:


packages/devtools/src/ganache/server.ts
```diff
@@ -10,6 +10,8 @@ import {Account, DeployedArtifacts, Deployment} from '../types';
 
 log.setDefaultLevel(log.levels.INFO);
 
+export const ganacheIsRunning = (port: number): boolean => false;
```
lalexgap: Is this needed or used?

packages/hub/scripts/start-ganache.js
```diff
@@ -0,0 +1,19 @@
+process.env.BABEL_ENV = "development";
```
lalexgap: Could these `start-ganache` scripts be consolidated into one script in `devtool`s? Or did you run into issues with that?

lalexgap:
Looks good to me!

PR #632, opened 2019-12-03
[hub] fix running hub locally

kerzhner:
For `development` environment (aka running the hub locally against ganache):
- Fix db migrations and seeding.
- Update readme.

Also:
- knexfile is now typescript to address @snario's [comment]( https://github.com/statechannels/monorepo/pull/615#discussion_r352978925).
- `.env` files cleaned up.


kerzhner:
@andrewgordstewart I did the following. Curious to hear if you think that's reasonable:
1. `knexfile.ts` is now under the `src` directory with the rest of the typescript source code. `knexfile.ts` is compiled (which happens during `yarn build`).
1. A painpont is with migration names stored in the `knex_migrations` table. When knex commands are run via command line like [here](https://github.com/statechannels/monorepo/blob/b4a7d02a1fdbefbfcba3bec80bc4489fec50b873/packages/hub/package.json#L12), migration files are `js` files. When knex commands are triggered by jest, like [here](https://github.com/statechannels/monorepo/blob/master/packages/hub/src/config/knexSetupTeardown.ts), migration files referenced are typescript files.
1. As a consequence, it's a bad idea to trigger knex commands via jest (or ts-node) AND the command line for the same database. You'll see migration errors.

andrewgordstewart:
> 3\. As a consequence, it's a bad idea to trigger knex commands via jest (or ts-node) AND the command line for the same database. You'll see migration errors.

We could address this pain point by making
`    "db:migrate": "npx ts-node knex/migrations.ts",`
or something like that

kerzhner:
Yep, we can likely come up with something to address that. Filed https://github.com/statechannels/monorepo/issues/634.

PR #636, opened 2019-12-04
Add USE_NATIVE_SOLC to CircleCI builds

snario:
This should shave off some time from CircleCI, maybe a few seconds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,5 @@
 version: 2.1
+
 orbs:
   slack: circleci/slack@3.4.1
   puppeteer: threetreeslight/puppeteer@0.1.2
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -124,7 +130,10 @@ jobs:
           package: nitro-protocol
 
   integration-test:
-    <<: *defaults
+    resource_class: medium
+    working_directory: /home/circleci/project
+    docker:
+      - image: circleci/node:10.16.3
```
kerzhner: What's the reason a different image is used for integration tests vs other jobs?

snario: The `counterfactual:statechannels` image is built ontop of `node:10.16-slim`, and to be honest I did not test it, but I suspect `puppeteer` won't run on there, and I did not want to deal with that at the moment. I know `circleci/node:10.16.3` seems to be fine with `puppeteer`, so just took the easy path.

kerzhner: Makes sense!

PR #638, opened 2019-12-04
Fix issues with lint-staged in various packages

snario:
Andrew and I noticed that the linting precommit hook wasn't working as expected, so I investigated. Turns out some packages did not have it set at all (i.e., `rps` and `hub`) and others had it set incorrectly (wrong regex to search for staged files). Also, it was implemented incorrectly in the first place in most cases, as if there _were_ a staged filed to be linted, it would run `yarn lint:check` which usually linted _every_ file and not just the sole changed one.

### DIFF
--- a/packages/embedded-wallet/package.json
+++ b/packages/embedded-wallet/package.json
@@ -18,8 +18,8 @@
     "test:ci": "react-scripts test --runInBand",
     "test:coverage": "react-scripts test --coverage --watchAll=false",
     "eject": "react-scripts eject",
-    "lint:check": "npx eslint \"*/**/*.{js,ts,tsx}\" --cache",
-    "lint:write": "npx eslint \"*/**/*.{js,ts,tsx}\" --fix"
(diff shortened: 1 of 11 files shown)


packages/embedded-wallet/package.json
```diff
@@ -33,7 +33,7 @@
     }
   },
   "lint-staged": {
-    "src/**/*.tsx?": "eslint"
+    "src/**/*.{ts,tsx}": "eslint"
```
andrewgordstewart: Why does `tsx?` not work?

snario: `lint-staged` only implements a subset of `regex` patterns, it seems. https://github.com/statechannels/monorepo/blame/199a552da29e6824b77d6b047adc73eccb8f48ef/packages/wallet/src/redux/adjudicator-state/reducer.ts#L47:L47

andrewgordstewart: The docs there indicate that it uses [micromatch](https://github.com/micromatch/micromatch), which claims to support the `*` and `?` wildcards.

PR #644, opened 2019-12-04
[channel-provider] Get channel provider working with wallet

lalexgap:
Updates the channel-provider to work with our current `wallet`.

It supports 
 - Hiding/Showing the wallet (once #643 is merged)
 - Sending messages, receiving a response (`const address = await window.channelProvider.send('GetAddress',{})`)
- Registering a listener for json-rpc notifications.

Currently subscriptions are not supported.


lalexgap:
> The idea with the `provider` is to keep the same interface as the Ethereum provider, as defined by [EIP 1193](https://github.com/ryanio/EIPs/blob/master/EIPS/eip-1193.md). That way, wallet builders don't have to change their interface with the browser to support state channels, they can just receive more messages down the same pipe. For this reason, we shouldn't add the `onNotification` action to the `provider`.
> 
> I'm guessing, but I think the interaction should look something like:
> 
> ```ts
> const messageQueuedSubscription = provider.subscribe('MessageQueued');
> provider.on(messageQueuedSubscripton, someCallback);
> // equivalent to
> provider.on('MessageQueued', someCallback);
> ```
> 
> For Ethereum events, I'm guessing you don't need to subscribe to e.g. `newBlock`, but you would need to subscribe to a particular stream of smart contract events. For state channels, it's not clear to me whether there are any events we'd need to subscribe to explicitly - and we can definitely side-step this by sending across all events at first. We probably implement a basic subscribe method though, so that the two approaches above both work.

I've updated the `channelProvider` to be able to register for json-rpc notifications using `on`/`off` (ie: provider.on('MessageQueued', someCallback);)

I've also added the ability to subscribe/unsubscribe to messages. Basically subscribing will just generate a guid that can be used in the `on` to register a callback for notifications.

packages/channel-provider/src/types.ts
```diff
@@ -13,33 +13,44 @@ export type JsonRpcResponse<ResultType = any> = {
...
   error: JsonRpcError;
 };
 
+export function isJsonRpcErrorResponse(message: any): message is JsonRpcErrorResponse {
+  return 'error' in message;
+}
+
 export interface IChannelProvider {
   enable(url?: string): Promise<void>;
   send<ResultType = any>(method: string, params?: any[]): Promise<ResultType>;
```
snario: Did you forget to change this `any[]` to `any`?

lalexgap: Yup, good catch!

Issue #618, opened 2019-12-03
Add solution to detect TypeScript build issues in test files before tests start runninga


snario:
I just wasted 10 minutes waiting for this CI to pass: https://circleci.com/gh/statechannels/monorepo/5201?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

And it turns out there was a TypeScript build error in one of the test files.

`hub/tsconfig.json` does not build `test/**/*` files so it makes sense it had to wait for `jest` to attempt to build them before running them.

However we _do have_ a `build` step with `yarn build`, but the annoying part is we don't want those test files in our `dist` files for the hub. 

So is there a solution?

snario:
Actually I might be crazy / very tired as that CI job failed on `yarn build`... but I thought it had not when I ran `yarn build` locally. A bit confused.

kerzhner:
To clarify, what is the desired outcome of resolving this issue? Is it to detect typescript issues during CI build step? Or is it to learn about typescript issues as quickly as possible? If it is the latter, perhaps a git pre-push hook might be handy.

snario:
It is to learn about typescript issues as quickly as possible.

PR #646, opened 2019-12-05
[monorepo] Consolidate tsconfig.json files

### LINKED ISSUE #64: [monorepo] Move common TypeScript configurations into root tsconfig.json
At the moment there are several redundant `tsconfig.json` settings in the various packages. We can roll up the common ones and put them in the root `/tsconfig.json`.

snario:
This was a leftover chore from 2 months ago. There were several redundant options set in the various `tsconfig.json` files. I combed through each and every one to check if they were actually necessary or not, and then set at the root level all of the shared / default values we use across every project for TypeScript code. In some cases very minor changes could be made to the code to adhere to the `strict` property. 

Also `build:typescript` is a script on every package's `package.json` now, which might be helpful for #618.

Fixes #64


packages/devtools/tsconfig.json
```diff
@@ -2,14 +2,14 @@
...
-    "composite": true,
     "lib": ["es6"],
     "module": "commonjs",
-    "noImplicitThis": true,
     "outDir": "lib",
-    "strictNullChecks": true,
     "target": "es2015",
     "types": ["node", "jest"],
+
+    /* The way setupGanache is written is hard to re-write in 5 min to fix this */
```
tomclose: Curious to hear a quick summary of what makes it hard?

snario: It just took longer than a couple minutes and I didn't want to start editing code. I think the quick summary is that it returns one of two objects which do not overlap with each other and behave quite differently.

PR #647, opened 2019-12-05
[Wallet] Return funding array from API

### LINKED ISSUE #645: [wallet] The wallet does not return funding information in the API.
Currently the `wallet` always returns `funding:[]` even after funding is complete.

lalexgap:
Fixes #645 

Sets the `funding` array on the API response for `UpdateState`. Currently only supports ETH.

### DIFF
--- a/packages/wallet/src/redux/sagas/message-sender.ts
+++ b/packages/wallet/src/redux/sagas/message-sender.ts
@@ -7,6 +7,8 @@ import {createJsonRpcAllocationsFromOutcome} from "../../utils/json-rpc-utils";
 import jrs from "jsonrpc-lite";
 import {unreachable} from "../../utils/reducer-utils";
 import {validateResponse, validateNotification} from "../../json-rpc-validation/validator";
+import {getChannelHoldings} from "../selectors";
+import {bigNumberify} from "ethers/utils";
 
 export function* messageSender(action: OutgoingApiAction) {
   const message = yield createResponseMessage(action);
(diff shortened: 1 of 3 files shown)


packages/wallet/src/redux/sagas/message-sender.ts
```diff
@@ -113,7 +115,13 @@ function* getChannelInfo(channelId: string) {
 
   const {participants} = channelStatus;
   const {appData, appDefinition, turnNum} = state;
-  const funding = [];
+
+  const channelHoldings = yield select(getChannelHoldings, channelId);
+  let funding: any[] = [];
+  // TODO: For now we assume ETH
```
snario: Why is it necessary to assume ETH?

lalexgap: I don't think the `wallet` has a mapping of `assetHolderAddress` to `token`, so we can't set the `token` value for anything but ETH.

snario: Oh I see. I had been operating as though they were interchangeable in my mind, but that was a mistake. It's actually easy to fetch the value you want though; you can check `await Contract(assetHolderAddress, ERC20AssetHolder.abi, provider).functions.Token()`

lalexgap: That's a good point!

 I think we need a few more changes to support multiple assetHolders though:
- We need to support multiple `assetHolder`s for different `token`s. (The `.env` file could probably contain an array of known assetHolder addresses)
- On `CreateChannel`/`JoinChannel` we need to check that `token` belongs to one of the `assetHolder`s the `wallet` knows.

It probably makes sense to store a mapping of `assetHolder` to `token` in the store, so we don't have to query the blockchain on each API call.

snario: > We need to support multiple assetHolders for different tokens.
https://github.com/statechannels/monorepo/issues/656

PR #649, opened 2019-12-05
[hub] use latest ganache deployer

kerzhner:
- Use the latest ganache deployer.
- Minimize the number of places where the environment variables are read in (from `.env` files).
- Remove dummy contract addresses from `.env`.


packages/hub/.env
```diff
@@ -15,10 +15,3 @@ TESTNET=ropsten
 INFURA_API_KEY=RequiredInLocalDotenv
 INFURA_JSON_RPC_ENDPOINT=https://${TESTNET}.infura.io/v3/${INFURA_API_KEY}
 GANACHE_JSON_RPC_ENDPOINT=http://${GANACHE_HOST}:${GANACHE_PORT}
-
-# These contract addresses get defined in the start/build scripts 
-# or the contract test setup with the actual adresses of the contracts
-CONSENSUS_APP_ADDRESS = 0x0000000000000000000000000000000000000000
-TRIVIAL_APP_ADDRESS = 0x0000000000000000000000000000000000000000
-NITRO_ADJUDICATOR_ADDRESS = 0x0000000000000000000000000000000000000000
-ETH_ASSET_HOLDER_ADDRESS = 0x0000000000000000000000000000000000000000
```
tomclose: In the wallet, these were kept, so that app-only tests could run without compiling the contracts. I imagine that might be useful in the hub at some point, but you could also just add them back in if it got to that point.

kerzhner: That makes sense. On the hub, the code that reads contract addresses is isolated so that:
1. Majority of tests can run without contract addresses defined.
1. The tests that read contract addresses need valid addresses and a running ganache instance.

I wonder if there is a way to adopt a similar approach for the wallet.

PR #650, opened 2019-12-05
[rps] Refactor game reducer to switch on state, not action

geoknee:
- saves quite a lot of lines
- is easier to compare with diagrams such as 
image

- is easier to catch bugs: in fact I found that one state was being entirely skipped over in a test (now fixed)

### DIFF
--- a/packages/rps/src/redux/game/reducer.ts
+++ b/packages/rps/src/redux/game/reducer.ts
@@ -1,44 +1,24 @@
 import {
-  WeaponChosen,
   weaponChosen,
   weaponAndSaltChosen,
   resultPlayAgain,
   LocalState,
   lobby,
-  waitForRestart,
   chooseWeapon,
   resigned,
   gameChosen,
-  waitingRoom,
-  opponentJoined,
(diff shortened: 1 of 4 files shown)


packages/rps/src/redux/game/reducer.ts
```diff
@@ -57,203 +37,92 @@ const localReducer: Reducer<LocalState> = (
...
+  let newState = state;
+  if (
+    action.type === 'Resign' &&
+    state.type !== 'Empty' &&
+    state.type !== 'NeedAddress' &&
+    state.type !== 'Lobby' &&
+    state.type !== 'WaitingRoom' &&
+    state.type !== 'CreatingOpenGame'
+  ) {
+    newState = resigned(state, action.iResigned);
```
tomclose: Might be nicer to just `return resigned(state, action.iResigned)` here - as someone reading the code, that's nice, because I know that's it and don't have to scroll down to the bottom to see if anything else happens to `newState`. Also removes a variable and a couple of lines 😛.

PR #651, opened 2019-12-05
[devtools] Format Devtools ESLint config

Alonski:
### Description

Format Devtools ESLint config


packages/devtools/.eslintrc.js
```diff
@@ -1,185 +1,185 @@
...
-    "number",
-    "String",
-    "string",
-    "Boolean",
-    "boolean",
-    "Undefined"
+  'arrow-body-style': 'error',
+  'arrow-parens': ['off', 'as-needed'],
+  camelcase: 'error',
+  complexity: 'off',
```
geoknee: Is now a good time to fix the `'key':` vs `key:` discrepancy?

Alonski: JavaScript allows us to define keys without quotes surrounding them if they are single words.
So `key: "something"` is valid while `this-key: "something"` is not.
The valid way is `"this-key": "something"`

geoknee: ...so do you think we should have all keys in double quotes? That way all keys have the same format.

Alonski: It is formatted automatically by Prettier. I don't think there is a config for it.
Take a look at this discussion:
https://github.com/prettier/prettier/issues/838

geoknee: Approved since this issue is not caused by this PR.

PR #653, opened 2019-12-05
[wallet] Create minimal ESLint config for wallet

snario:
After #611 was merged I was having issues with linting in the wallet being gone. It's pretty easy for this to lead to all kinds of annoying linting issues down the road, so I went through our old config and generated a new ESLint config that does 99% the same thing as that one, with some very minor changes.


Alonski:
@snario I will take a look at this tomorrow. Still not sure why you had to remove format on save. Which file was it changing for you?

packages/wallet/.eslintrc.js
```diff
@@ -1,49 +1,89 @@
...
+    "plugin:import/errors",
+    "plugin:import/warnings",
+    "plugin:import/typescript",
+    "plugin:react/recommended",
+    "plugin:prettier/recommended"
   ],
   rules: {
-    // "prettier/prettier": "error"
+    // Prettier violations should result in linting errors
+    "prettier/prettier": ["error", require("./.prettierrc.js")],
```
Alonski: The require here shouldn't be needed.
`"prettier/prettier": "error"` should have the same effect

packages/wallet/.eslintrc.js
```diff
@@ -1,49 +1,89 @@
...
   parserOptions: {
-    project: "tsconfig.json",
-    ecmaVersion: 2018,
-    sourceType: "module"
+    project: "tsconfig.json"
+  },
+  plugins: ["@typescript-eslint", "prettier", "jest", "react"],
+  settings: {
+    "settings": {
+      // TODO: Fix a warning being thrown despite the version being set here
```
Alonski: The settings are nested inside of settings :)
This needs to be:
```js
  plugins: ["@typescript-eslint", "prettier", "jest", "react"],
  settings: {
    react: {
      version: "detect"
    }
  },
  extends: [
```

packages/wallet/.eslintrc.js
```diff
@@ -1,49 +1,89 @@
+const generalRules = {
+  "no-var": "off",
```
Alonski: Why would we want this?
`let` and `const` should be used no?
Also, I am also a strong believer in using: `"prefer-const": "error",` as well.

packages/wallet/.eslintrc.js
```diff
@@ -1,49 +1,89 @@
...
+// From the tslint.json we used previously
+const leftoverTsLintRules = {
+  "@typescript-eslint/no-explicit-any": "off",
+  "@typescript-eslint/explicit-function-return-type": "off",
+  "@typescript-eslint/no-use-before-define": "off",
+  "@typescript-eslint/ban-ts-ignore": "off",
+  "@typescript-eslint/no-non-null-assertion": "off",
+  // TODO: Get rid of this, there are just a small number of cases
+  "@typescript-eslint/no-unused-vars": "off"
+};
```
Alonski: I think we should reexamine these in the future. Especially `no-explicit-any` and `explicit-function-return-type` but pretty much all of them...

We are losing out on the things that Typescript offers us when we do these things.

packages/wallet/package.json
```diff
@@ -14,8 +14,8 @@
     "build:typescript": "npx tsc -b",
     "build:webpack": "CI=false node scripts/build.js",
     "build": "run-s build:typescript build:webpack",
-    "lint:check": "eslint \"*/**/*.{js,ts,tsx}\" --cache",
-    "lint:write": "eslint \"*/**/*.{js,ts,tsx}\" --fix",
+    "lint:check": "eslint \"src/**/*.{js,ts,tsx}\" --cache",
+    "lint:write": "eslint \"src/**/*.{js,ts,tsx}\" --fix",
```
Alonski: Why do we not want to lint files that are not in `src`? We don't care how our config files look and act?

packages/wallet/.eslintrc.js
snario: This is not a rule I am enforcing on anyone, it is what was already used in the codebase.

packages/wallet/.eslintrc.js
snario: I agree. This PR is not meant to change any rules, it's just trying to recreate what we had before.

packages/wallet/package.json
snario: We can change it to something like `{src/config}` but I feel it is better to be explicit vs implicit with these kinds of tools lest they blow up due to some file / folder that's full of files. It also mimics how it worked before.

snario:
@Alonski the main point of this PR is not to add any new rules, it's just to mimic what we had before so as to not be disruptive. If we want to change code styles, those can be introduced as new rules, but it's really not a major priority for either @lalexgap or I who are mainly working in the `wallet` package at this moment.

Alonski:
@snario No problem. I just think that turning off various rules that, in my opinion are useful, is a bit of a waste of ESLints power :)

packages/wallet/.eslintrc.js
Alonski: @snario Once this is fixed this PR can be merged

packages/wallet/.eslintrc.js
```diff
@@ -55,11 +55,9 @@ module.exports = {
   },
   plugins: ["@typescript-eslint", "prettier", "jest", "react"],
   settings: {
-    "settings": {
-      // TODO: Fix a warning being thrown despite the version being set here
-      "react": {
-        "version": "detect"
-      }
+    // TODO: Fix a warning being thrown despite the version being set here
```
Alonski: This comment can be removed as well :)

snario:
> No problem. I just think that turning off various rules that, in my opinion are useful, is a bit of a waste of ESLints power :)
I agree, I'm just saying it is out of scope for this PR to change them back because it would require significant code changes / take up a bunch of time to rewrite all of the code to be compliant with a new set of rules it was not written for.

PR #655, opened 2019-12-05
Turn off formatOnSave

snario:
For me, this was affecting files in the `wallet` folder while I had open the `monorepo` as a root folder in my workspace. It was applying the `bracketSpacing: true` rule in particular every time I hit save.

Turning this flag off fixed the issue.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,5 +8,5 @@
     {"directory": "packages/jest-gas-reporter", "changeProcessCWD": true},
     {"directory": "packages/nitro-protocol", "changeProcessCWD": true}
   ],
-  "editor.formatOnSave": true
+  "editor.formatOnSave": false
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,5 @@
     {"directory": "packages/hub", "changeProcessCWD": true},
     {"directory": "packages/jest-gas-reporter", "changeProcessCWD": true},
     {"directory": "packages/nitro-protocol", "changeProcessCWD": true}
-  ],
-  "editor.formatOnSave": false
+  ]
 }


.vscode/settings.json
```diff
@@ -8,5 +8,5 @@
     {"directory": "packages/jest-gas-reporter", "changeProcessCWD": true},
     {"directory": "packages/nitro-protocol", "changeProcessCWD": true}
   ],
-  "editor.formatOnSave": true
+  "editor.formatOnSave": false
```
tomclose: Could you just remove this? That way any settings we have set locally still apply.

.vscode/settings.json
```diff
@@ -7,6 +7,5 @@
     {"directory": "packages/hub", "changeProcessCWD": true},
     {"directory": "packages/jest-gas-reporter", "changeProcessCWD": true},
     {"directory": "packages/nitro-protocol", "changeProcessCWD": true}
-  ],
-  "editor.formatOnSave": true
```
andrewgordstewart: You can override this option in the wallet `settings.json`, so that it remains for the other folders.

I think this problem is temporary due to
- eslint not running prettier
- vscode still using prettier to format on save in the wallet, but using the wrong prettier config (?)

Alonski:
This is because we have an empty ESLint file with no rules. I thought this might happen. @snario you can set this setting locally for now instead of for the workspace in your local settings

andrewgordstewart:
> This is because we have an empty ESLint file with no rules. I thought this might happen. @snario you can set this setting locally for now instead of for the workspace in your local settings

Workspace settings override user settings.

I looked, and didn't find a way to have something like `settings.local.json` that overrides `settings.json`.

How can I set `formatOnSave` locally that doesn't involve editing `settings.json`?

Alonski:
Woops I thought it was the other way

Alonski:
Then the fix is to have a good ESLint config in all packages

snario:
Actually it turns out the reason it was happening is because there is some [default formatter on VS Code](https://github.com/microsoft/vscode/tree/master/extensions/typescript-language-features) and that thing was being run.

Alonski:
@snario Yea sorry for that. I have always had Prettier set as my default formatter. 
Can we close this?

Alonski:
I thought we weren't going to merge this?

PR #657, opened 2019-12-05
[hub] update package.json

kerzhner:


packages/hub/package.json
```diff
@@ -12,8 +12,8 @@
     "db:migrate": "npx knex migrate:latest --knexfile lib/config/knexfile",
     "db:rollback": "npx knex migrate:rollback --knexfile lib/config/knexfile",
     "db:seed": "npx knex seed:run --knexfile lib/config/knexfile",
-    "hub:start": "node ./lib/hub/server.js",
-    "hub:watch": "nodemon --watch 'src/**/*' -e ts,tsx --exec ts-node ./src/hub/server.ts",
+    "hub:start:dev": "node ./lib/start-dev-server.js",
```
andrewgordstewart: Should this `ts-node` the `src/start-ev-server.ts` instead?

Also, a bit of a nit but `start-dev-server` should eventually _not_ be part of `lib` (the built hub wallet)

kerzhner: yep, we probably do want to use `ts-node` for starting the hub locally. Good point about `start-dev-server`.

Filed https://github.com/statechannels/monorepo/issues/658

PR #659, opened 2019-12-05
[channel-provider] Update ChannelProvider JSON-RPC Schema

snario:
Apparently `params` can either be an array or an object but not both.

Object is better I think. Also it's easier to type it.

See: https://www.jsonrpc.org/specification

Also this PR fixes a path mistake for the package's types.


snario:
@lalexgap is doing this in #644

PR #665, opened 2019-12-05
[channel-client] Add package for ChannelClient

snario:
Work in progress...

### DIFF
--- a/packages/channel-client/tests/index.test.ts
+++ b/packages/channel-client/tests/index.test.ts
@@ -1,0 +1,5 @@
+describe('ChannelClient', () => {
+  it('works', () => {
+    expect(true).toBeTruthy();
+  });
+});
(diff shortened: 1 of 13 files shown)


packages/channel-client/tests/index.test.ts
```diff
@@ -0,0 +1,15 @@
...
+// TODO: Figure out how to test
+describe('ChannelClient', () => {
+  it('instantiates', () => {
+    expect(
+      new ChannelClient({
+        enable: () => {
+          /* do nothing */
+        }
+      } as IChannelProvider)
+    ).toBeDefined();
```
cf19drofxots: Not directly relevant to the scope at hand but we should probably stick to the same interface naming convention (`IChannelProvider` here vs `ChannelClientInterface` elsewhere)

PR #666, opened 2019-12-05
Hub: env vars

kerzhner:
- migrate away from the depricated `magmo-devtools` dependency.
- do not populate environment variables in the hub application. Expect the environment variables to be populated before application start.

### DIFF
--- a/packages/hub/src/start-dev-server.ts
+++ b/packages/hub/src/start-dev-server.ts
@@ -1,5 +1,6 @@
 import {setupGanache} from '@statechannels/devtools';
 import {deploy} from '../deployment/deploy';
+import '../config/env'; // Note: importing this module has the side effect of modifying env vars
 import {startServer} from './hub/server';
 
 async function setupGanacheAndContracts() {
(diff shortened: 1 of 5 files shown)


packages/hub/src/start-dev-server.ts
```diff
@@ -1,3 +1,5 @@
+import '../config/env'; // Note: importing this module has the side effect of modifying env vars
```
tomclose: I wonder if it would be clearer to just import and run `configureEnvVariables()` here. That's all that `../config/env` does, and it seems less magic.

kerzhner: I thought about that as well. `env.js` is also used for setting env vars for [jest unit tests](https://github.com/statechannels/monorepo/blob/master/packages/hub/config/jest/jest.config.js#L7) and for setting up ganache for [unit tests that require ganache](https://github.com/statechannels/monorepo/blob/master/packages/hub/jest/test-setup.ts#L1).

My thinking was that it could be worthwhile to keep a module for setting env vars in case the module evolves to more than just the call to `configureEnvVariables()`. That being said, I don't feel strongly about it.

PR #667, opened 2019-12-05
[jest-gas-reporter] Lint jest-gas-reporter

snario:
Adds a super basic eslint config and fixes linting issues with it.


packages/jest-gas-reporter/.eslintrc.js
```diff
@@ -1,178 +1,16 @@
...
-    'plugin:import/warnings',
-    'plugin:import/typescript',
-    'prettier/@typescript-eslint',
-    'prettier',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:prettier/recommended'
   ],
-  rules: Object.assign(rules, TSRules),
+  rules: {
+    'prettier/prettier': ['warn']
```
Alonski: Hmm why not error?

packages/jest-gas-reporter/package.json
```diff
@@ -12,8 +12,8 @@
     "prepare": "yarn build",
     "build": "yarn lint:check && tsc -b",
     "build:typescript": "tsc -b",
-    "lint:check": "eslint \"*/**/*.ts\" --cache",
-    "lint:write": "eslint \"*/**/*.ts\" --fix"
+    "lint:check": "eslint \"src/**/*.ts\" --cache",
```
Alonski: Why is this only checking `src` and `ts` files?
Do we not want linting for all files?

packages/jest-gas-reporter/.eslintrc.js
snario: To distinguish between linting issues and build issues. I think it makes the code more readable.

Alonski: The question is do we want CI to fail on formatting issues?
Anyways your choice :)

snario: Good point. Turns out we can have the best of both worlds with `--max-warnings=0`.

Screen Shot 2019-12-06 at 10 50 08
Screen Shot 2019-12-06 at 10 50 25

PR #671, opened 2019-12-06
Strongly typed provider interface

tomclose:
Adds strong typing to the provider interface. Should be useful for developing the `ChannelClient`.


tomclose:
This doesn't work as I'd intended.

PR #674, opened 2019-12-06
[wallet] Use client-api-schema

snario:
Uses `client-api-schema` in the `wallet`.

### DIFF
--- a/packages/wallet/src/redux/sagas/messaging/message-handler.ts
+++ b/packages/wallet/src/redux/sagas/messaging/message-handler.ts
@@ -1,5 +1,11 @@
 import {select, fork, put, call} from "redux-saga/effects";
 import {getChannelId, State} from "@statechannels/nitro-protocol";
+import {
+  JoinChannelParams,
+  PushMessageParams,
+  UpdateChannelParams,
+  CreateChannelParams
+} from "@statechannels/client-api-schema";
 import jrs, {RequestObject} from "jsonrpc-lite";
 
 import * as outgoingMessageActions from "./outgoing-api-actions";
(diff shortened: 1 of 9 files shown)


packages/wallet/src/utils/json-rpc-utils.ts
```diff
@@ -3,68 +3,36 @@ import {
...
-}
-
-function createAllocationOutcomeFromParams(params: JsonRpcAllocations): Outcome {
+import {
+  UpdateChannelParams,
+  Allocations,
+  CreateChannelParams
+} from "@statechannels/client-api-schema";
+
+// TODO: Figure out how these should have been used
```
snario: @lalexgap

lalexgap: These should probably be moved somewhere else. These are interfaces for messages between `wallet`s.

The json-rpc api specifies that the `data` on a message can be anything. These interfaces define what the `wallet` currently expects/handles in the `data` field.

```
{
  "jsonrpc": "2.0",
  "method": "PushMessage",
  "id": 1,
  "params": {
    "recipient": "user123",
    "sender": "user456",
    "data": "0x123.."
  }
}
```

packages/wallet/src/redux/sagas/messaging/message-handler.ts
```diff
@@ -108,11 +113,12 @@ function* handleJoinChannelMessage(payload: RequestObject) {
 function* handlePushMessage(payload: RequestObject) {
   // TODO: We need to handle the case where we receive an invalid wallet message
   const {id} = payload;
-  const message = payload.params as JsonRpcMessage;
+  const message = payload.params as PushMessageParams;
   if (isRelayableAction(message.data)) {
     yield put(message.data);
     yield fork(messageSender, outgoingMessageActions.postMessageResponse({id}));
   } else {
+    // TODO: Add schema for PushMessageParams.data to client-api-schema/schema/push-message.json
```
lalexgap: I don't think we want to specify the `data` in the JSON schema. The specific messaging format shouldn't really be part of the API that the JSON schema specifies. The client should just blindly pass along `data`.

snario: Oh okay, I'll remove the todo.
```suggestion
```

PR #675, opened 2019-12-06
[hub] postgres db connection revamp

kerzhner:
knex encourages storage of all db configurations in [`knexfile`](http://knexjs.org/#knexfile). Configuration includes how to connect to a db as well as migrations and seeds. An application should not be aware of migrations and seed data (only db administration utilities need that data).

This PR:
- Separates db connection data into `db-config.ts`. The hub application needs only this configuration to build and run.
- Creates database admin utilities used by yarn commands and unit test setup scripts.
- Unites jest configuration under one directory.


packages/hub/.env.test
```diff
@@ -1,2 +1,2 @@
 HUB_ADDRESS=0x7892ea3A79ffB40388d6769Ef58642b3e4412Fbb
-HUB_PRIVATE_KEY=0xA39E13320C7752AEA43CE032CF40EF6B10BDC7753AA8B98A07F9477F30C8AC8F;
\ No newline at end of file
+HUB_PRIVATE_KEY=0xA39E13320C7752AEA43CE032CF40EF6B10BDC7753AA8B98A07F9477F30C8AC8F;
```
cf19drofxots: Why did the private key need to get updated?

packages/hub/jest/test-setup.ts
```diff
@@ -1,6 +1,6 @@
-import "../config/env";
-import {GanacheServer} from "@statechannels/devtools";
-import {deploy} from "../deployment/deploy";
+import '../env';
```
cf19drofxots: Do we enforce single or double quotes in linting? I couldn't find a linting file for the hub package.

snario: https://github.com/statechannels/monorepo/blob/f12dd05d24329f17f3be82b828bfdc878b1f6268/packages/hub/.eslintrc.js#L102

packages/hub/.env.test
snario: What does this even do?

packages/hub/jest/test-setup.ts
cf19drofxots: Ah there we go - `eslintrc.js`

packages/hub/.env.test
kerzhner: The private key was not updated. A newline was added to the end of the file. It would be nice to enforce newlines at the ends of files if that's easy to add.

`HUB_PRIVATE_KEY` env var is not wired up at the moment and can be deleted. I believe the intention of the env var was to store the key used for signing both states and transactions.

packages/hub/jest/test-setup.ts
kerzhner: Hmm, is the linting not enforced because this file is outside of `src`?

kerzhner: (since the linter was passing happily with double quotes in the imports before this PR)

packages/hub/.env.test
kerzhner: Removed `.env.test` for now and filed https://github.com/statechannels/monorepo/issues/677.

packages/hub/package.json
```diff
@@ -9,14 +9,14 @@
   "scripts": {
     "db:drop": "dropdb hub_wallet_${NODE_ENV}",
     "db:create": "createdb hub_wallet_${NODE_ENV}",
-    "db:migrate": "npx knex migrate:latest --knexfile lib/config/knexfile",
-    "db:rollback": "npx knex migrate:rollback --knexfile lib/config/knexfile",
-    "db:seed": "npx knex seed:run --knexfile lib/config/knexfile",
-    "hub:start:dev": "node ./lib/start-dev-server.js",
+    "db:migrate": "npx knex migrate:latest --knexfile lib/db-admin/knexfile",
```
cf19drofxots: I'd suggest adding a higher-level script like `db:setup` that'd run all of these as I expect in a development flow you're more often running these commands back to back instead of in isolation.

Though I could be wrong as @kerzhner has been iterating more heavily on the hub recently.

kerzhner: That's a great idea! TBH, I haven't really been running the hub much recently. I have been relying mostly on unit tests. And unit tests do not use these yarn commands.

PR #676, opened 2019-12-06
[rps] Replace ChannelResult types with Client API Schema types

snario:


snario:
Should `turnNum` be a `string` or `number` in the API? @tomclose @lalexgap

tomclose:
I think `turnNum` should probably be a string.

packages/client-api-schema/schema/close-channel.json
```diff
@@ -0,0 +1,52 @@
...
+        },
+        "status": {
+          "type": "string",
+          "const": "opening",
+          "enum": ["proposed", "opening", "funding", "running", "closing", "closed"]
+        },
+        "turnNum": {
+          "type": "string",
+          "const": "0x00"
+        }
```
tomclose: This set of attributes (elsewhere known as a `ChannelResult`) appears together in virtually all the responses. It therefore might be useful to introduce a new schema definition for this type, which help identify it a key concept.

That would allow us to flip some of this work on its head - reintroducing the `ChannelResult` type, but with that type defined automatically from the schema.

PR #682, opened 2019-12-07
[monorepo] Upgrade Prettier in all packages

Alonski:
### Description

This PR upgrades Prettier to latest and Prettiers all packages 

### Related issues

#585


package.json
```diff
@@ -29,12 +29,12 @@
     "state channels"
   ],
   "devDependencies": {
-    "@types/prettier": "1.18.2",
+    "@types/prettier": "1.19.0",
```
andrewgordstewart: Does it make sense to put prettier (and other common devDependencies) in the root `package.json`?

packages/devtools/.prettierrc.js
```diff
@@ -1,7 +1,7 @@
 // NOTE: This package has inconsistent use of trailing commas and bracket spacing
```
andrewgordstewart: Isn't it trivial to fix this with prettier? :P

package.json
andrewgordstewart: Here's a related eslint rule, which would let us enforce that:
- anything that is not listed as a `dependency` in a package, cannot be imported by that package

We would need to configure that rule to only apply to the source files of the project, since we would probably want use dev dependencies eg. in test files.

packages/devtools/.prettierrc.js
Alonski: Removed :) Not relevant anymore

package.json
Alonski: Hmm, I am not sure. We should discuss this.
I am going to leave this out of this PR.
Can you please open a ticket with these concerns?

packages/client-api-schema/types/response.d.ts
```diff
@@ -10,7 +10,7 @@
  */
 export type Response1 =
```
snario: This file should not be in here, it is gitignored.

PR #683, opened 2019-12-08
[hub] Jest setup script asserts for expected NODE_ENV env var

cf19drofxots:
Fixes https://github.com/statechannels/monorepo/issues/679

### DIFF
--- a/packages/hub/jest/test-setup.ts
+++ b/packages/hub/jest/test-setup.ts
@@ -3,6 +3,15 @@ import {GanacheServer} from '@statechannels/devtools';
 import {deploy} from '../deployment/deploy';
 
 export default async function setup() {
+  if (!process.env.NODE_ENV) {
+    throw Error(`The NODE_ENV env var must be set to "test"`);
+  }
+  if (process.env.NODE_ENV != 'test') {
+    throw Error(
+      `Run "yarn db:testSetup" and rerun the attempted script prefixed with "NODE_ENV=test"`
+    );
+  }
+
   const ganacheServer = new GanacheServer(
     Number(process.env.GANACHE_PORT),
     Number(process.env.CHAIN_NETWORK_ID)
--- a/packages/hub/readme.md
+++ b/packages/hub/readme.md
@@ -57,6 +57,20 @@ NODE_ENV=test yarn db:create
 yarn test:ci
 ```
 
+### Troubleshooting Test Errors
+
+If you run into the following error
+
+```shell
+    getaddrinfo ENOTFOUND RequiredInLocalDotenv RequiredInLocalDotenv:5432
+```
+
+Make sure you have a local .env file dedicated to testing, by running the following:
+
+```shell
+cp .env.development.local .env.test.local
+```
+
 ## Deploying
 
 Heroku is configured to automatically deploy from the watched `deploy` branch.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,6 @@ orbs:
   puppeteer: threetreeslight/puppeteer@0.1.2
 defaults: &defaults
   working_directory: /home/circleci/project
-  resource_class: large
   docker:
     - image: circleci/node:10.16.3
     - image: circleci/postgres:12.0-alpine


packages/hub/readme.md
```diff
@@ -57,6 +57,20 @@ NODE_ENV=test yarn db:create
...
+If you run into the following error
+
+```shell
+    getaddrinfo ENOTFOUND RequiredInLocalDotenv RequiredInLocalDotenv:5432
+```
+
+Make sure you have a local .env file dedicated to testing, by running the following:
+
+```shell
+cp .env.development.local .env.test.local
```
tomclose: Could we set the necessary details in `.env.test` to avoid this error completely?

kerzhner: Good thought, but it will not work unfortunately. Environment variables set outside the executable (and passed to the executable) take precedence over env vars read in from `.env` files.

packages/hub/jest/test-setup.ts
```diff
@@ -3,6 +3,15 @@ import {GanacheServer} from '@statechannels/devtools';
 import {deploy} from '../deployment/deploy';
 
 export default async function setup() {
+  if (!process.env.NODE_ENV) {
+    throw Error(`The NODE_ENV env var must be set to "test"`);
+  }
+  if (process.env.NODE_ENV != 'test') {
+    throw Error(
+      `Run "yarn db:testSetup" and rerun the attempted script prefixed with "NODE_ENV=test"`
```
kerzhner: I don't think we have defined this command yet.

packages/hub/readme.md
tomclose: I guess what I'm asking is could we set [these](https://github.com/statechannels/monorepo/blob/hub/assert_test_env_var/packages/hub/.env#L1-L2) to some standard values, that are likely to work on most laptops? Like `localhost` and `5432` or whatever the standard postgres port is?

kerzhner: There might be folks who run tests locally before setting up the development environment. In that case, `.env.development.local` would not exist. Would it be better to state something along the lines of:

The contents of your `.env.test.local` file define the connection to the postgres db via `HUB_DB_HOST` and `HUB_DB_PORT` variables. An example is:
```
HUB_DB_HOST=localhost
HUB_DB_PORT=5432
```

kerzhner: I like that idea. Do you know if port 5432 is standard for postgres on linux and windows?

Even if we set default host and port variables, but `NODE_ENV=development` is passed to jest tests, tests will fail. I spent a bit of time trying to understand why, but didn't get to the bottom of the issue. It seems that jest has a strong stance on NODE_ENV being set to `test`. I see hints about this NODE_ENV in some jest docs and issues. For example: https://jest-bot.github.io/jest/docs/getting-started.html.

kerzhner: Filed https://github.com/statechannels/monorepo/issues/693.

tomclose: I think if we just put `5432` in, then that should be fine. I'm hoping that knex will give a resonable error (e.g. "Can't connect to postgres at http://localhost:5432") if that isn't right, which should be enough to point them in the right direction. Might be worth verifying what happens (especially given my previous experience, where it just connected to the default database 🙀 ).

kerzhner: Yep, I am onboard with that. Good though on the default db. Tests pass with no `HUB_DB_NAME` defined (and the default db present on the database server).

kerzhner: This PR fixes the NODE_ENV issue and adds defaults to `.env.test`: https://github.com/statechannels/monorepo/pull/695

kerzhner: Actually, with https://github.com/statechannels/monorepo/pull/695, we can tell folks that `.env.test.local` is only needed if defaults in `.env.test` need to be overridden.

packages/hub/jest/test-setup.ts
kerzhner: Also, with https://github.com/statechannels/monorepo/pull/695, tests pass with any NODE_ENV.

packages/hub/readme.md
cf19drofxots: Hmm so as of #693, should we just close this (and #697) then?

packages/hub/jest/test-setup.ts
cf19drofxots: Hmm so as of #693, should we just close this (and #697) then?

packages/hub/readme.md
cf19drofxots: @kerzhner do you want to just include another commit in #695 then to update the README for the above comment? (only needing `.env.test.local` in case defaults need to be overridden)

packages/hub/jest/test-setup.ts
kerzhner: Yeah I think so.

PR #684, opened 2019-12-08
[monorepo] Cleanup Prettier and ESLint config

Alonski:
### Description

Some Prettier configs were extraneous in their configs overriding the root config unnecessarily.
In addition, ESLint config was declaring Prettier rules that should be taken from the Prettier configs.

This fixes both of these issues.


packages/wallet/.prettierrc.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   ...require("../../.prettierrc.js"),
-  printWidth: 100,
   singleQuote: false,
```
cf19drofxots: Here we declare `singleQuote` to be explicitly `false` but in other settings we remove it, and the lack of it implies `false` instead of an explicit setting. Perhaps the other places should also explicitly set it to `false`.

Alonski: I actually think the opposite. Why should we explicitly declare something that is the default? I will remove this.

Alonski: Oh, actually I misunderstood. Only the Wallet uses `singleQuote: false`. All others inherit from the root config that is `singleQuote: true`.
We should probably consolidate on one way of doing things across the whole monorepo no?

Alonski: Created this issue for discussion:
https://github.com/statechannels/monorepo/issues/730

PR #699, opened 2019-12-09
[client-api-schema] Update Client API Schema

snario:
Standardizes the result to be a `ChannelResult`


packages/wallet/src/json-rpc-validation/validator.ts
```diff
@@ -6,6 +6,8 @@ import getAddressSchema from "@statechannels/client-api-schema/schema/get-addres
 import joinChannelSchema from "@statechannels/client-api-schema/schema/join-channel.json";
 import updateChannelSchema from "@statechannels/client-api-schema/schema/update-channel.json";
 import definitionsSchema from "@statechannels/client-api-schema/schema/definitions.json";
+import channelResultSchema from "@statechannels/client-api-schema/schema/channel-result.json";
+import closeChannelSchema from "@statechannels/client-api-schema/schema/close-channel.json";
```
tomclose: It might be nicer to export the validator from `client-api-schema`, instead of importing all these definitions.

snario: I think I agree, I'll do that in a separate PR.

PR #702, opened 2019-12-09
[devtools] Significantly reduce .eslintrc.js size

snario:
Continuing to remove big linting config files


packages/devtools/src/ganache/deployer-with-cache.ts
```diff
@@ -1,7 +1,6 @@
 import * as fs from 'fs';
 import {CompiledContract} from 'etherlime-lib';
 import * as log from 'loglevel';
-// @ts-ignore etherlime-utils does not export any types
```
andrewgordstewart: I thought the typical solution to this was to declare that the type of the `etherlime-utils` module is `any` in some `globals.d.ts` file. (I've never figured out how that actually works)

PR #704, opened 2019-12-09
[channel-client] Move FakeChannelClient from rps package to channel-client package

cf19drofxots:
Addresses the first item in issue https://github.com/statechannels/monorepo/issues/688


packages/channel-client/tsconfig.json
```diff
@@ -2,7 +2,9 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     // This package is meant to emit webpack processed files
-    "noEmit": true
+    "noEmit": true,
+    // Needed for EventEmitter import
+    "allowSyntheticDefaultImports": true
```
snario: You can use import assignment instead.

```js
import EventEmitter = require('eventemitter3');
```

packages/channel-client/src/utils.ts
```diff
@@ -0,0 +1,3 @@
+export function sleep(time: number) {
```
snario: What's the point of a fake delay?

packages/channel-client/tsconfig.json
cf19drofxots: Isn't the `require` syntax an anti-pattern for TS imports?

packages/channel-client/src/utils.ts
cf19drofxots: To fake a delay to simulate two users interacting?

packages/channel-client/tsconfig.json
snario: I think
```
import EventEmitter = require('eventemitter3');
```
exposes that this _specific_ package has a default export for the type `EventEmitter` but not for the actual object `EventEmitter` which is non-default.

Adding `allowSyntheticDefaultImports` just masks the problem in my opinion.

packages/channel-client/src/utils.ts
geoknee: I think we were using this in RPS so that the user still sees all of the intermediate screens during interaction with the fake wallet.

snario: It's just faking the lag in messages being sent between participants? There is UI for this lag?

packages/channel-client/src/types.ts
```diff
@@ -69,3 +73,148 @@ export interface ChannelClientInterface<Payload = object> {
...
+
+export enum ErrorCodes {
+  SIGNING_ADDRESS_NOT_FOUND = 1000,
+  INVALID_APP_DEFINITION = 1001,
+  INVALID_APP_DATA = 1002,
+  UNSUPPORTED_TOKEN = 1003,
+  CHANNEL_NOT_FOUND = 1004
+}
+
+export class ChannelClientError implements JsonRpcErrorResponse {
```
tomclose: This and the other Error classes probably don't belong in types. They could probably be removed for the time being, as I don't think they're used anywhere.

PR #715, opened 2019-12-10
[rps] Experiment with @openzeppelin/test-environment

geoknee:
It claims to be blazing fast https://forum.openzeppelin.com/t/blazing-fast-contracts-testing/1873. 

The speedup they claim is impressive but might just be because they were using truffle tests beforehand. 

This PR is a work in progress to see if we can get any meaningful speedup or other advantage from migrating to this package.

### DIFF
--- a/packages/rps/jest/contract-test-setup.ts
+++ b/packages/rps/jest/contract-test-setup.ts
@@ -1,17 +1,15 @@
 import '../config/env';
-import {GanacheServer} from '@statechannels/devtools';
-import {deploy} from '../deployment/deploy';
+// import {GanacheServer} from '@statechannels/devtools';
+// import {deploy} from '../deployment/deploy';
 
 export default async function setup() {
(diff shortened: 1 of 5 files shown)


snario:
[This tool does similar things and I've found it to work well in the past.](**https://github.com/EthWorks/Waffle**)

geoknee:
https://github.com/OpenZeppelin/openzeppelin-test-environment/issues/29

pcowgill:
@snario I agree that Waffle is great, but I think the intended scope for this lib and Waffle overlaps somewhat but not entirely, right @frangio @geoknee?

frangio:
Waffle and OpenZeppelin Test Environment seem to have a similar API in terms of how the tests get hold of a Ganache provider. Aside from that, Waffle seems to bundle together a lot of functionality whereas we provide it in separate tools such as our Test Helpers and CLI.

PR #716, opened 2019-12-10
[hub] Channel.Open handler

kerzhner:
- respond to `Channel.Open` with `Channel.Joined`.
- Start adding `handleWalletMessage` unit tests. `handleWalletMessage` is the outermost function in the hub that is still simple to unit test.

### DIFF
--- a/packages/hub/src/communication/index.ts
+++ b/packages/hub/src/communication/index.ts
@@ -1,4 +1,5 @@
-import {SignedState} from '@statechannels/nitro-protocol';
+import {SignedState, getChannelId} from '@statechannels/nitro-protocol';
+import {ChannelParticipant} from '@statechannels/wallet/lib/src/redux/channel-store';
 
 export interface BaseProcessAction {
   processId: string;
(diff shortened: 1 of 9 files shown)


packages/hub/src/communication/index.ts
```diff
@@ -1,4 +1,5 @@
-import {SignedState} from '@statechannels/nitro-protocol';
+import {SignedState, getChannelId} from '@statechannels/nitro-protocol';
+import {ChannelParticipant} from '@statechannels/wallet/lib/src/redux/channel-store';
```
tomclose: It might be worth just copying this type across for the time being, instead of re-introducing the dependency of the hub on the wallet. In future, these types will probably be stored externally to both packages, in a new package that defines the wire protocol.

kerzhner: Good thinking. There are a couple other imports outside of this PR in the hub that rely on the client wallet. So, going to leave this cleanup for a future PR: https://github.com/statechannels/monorepo/issues/735.

One benefit of not copy/pasting (and importing modules from the `wallet` package) is that this makes it easy to track the shared pieces between the wallet and the hub.

PR #717, opened 2019-12-10
Fix name error that's breaking master

tomclose:


packages/rps/src/redux/game/reducer.ts
```diff
@@ -105,7 +105,7 @@ const localReducer: Reducer<LocalState> = (
       }
       if (action.type === 'CancelGame') {
         const {name, address} = state;
-        newState = lobby({name, address});
+        newState = Setup.lobby({name, address});
       }
```
geoknee: I think you probably want to just delete these lines, because the correct code is immediately above

PR #733, opened 2019-12-11
Start ganache

tomclose:
This PR makes some changes to make the `yarn start` behaviour simpler and more robust.

1. If you run `yarn start` without first running `yarn start:shared-ganache`, the script will now error instead of starting an individual ganache instance.
2. If package A starts an individual ganache instance on port P, and package B then tries to join it, thinking it is a shared ganache, then the start script will error.
3. Different packages use different deploy keys, solving https://github.com/statechannels/monorepo/issues/725.
4. The puppeteer integration tests still use an individual ganache instance on circle.


.gitignore
```diff
@@ -52,3 +52,5 @@ junit.xml
 
 
 ganache-deployments.json
+ganache-deployments*.json
+.ganache-deployment-cache/
```
snario: What folder does this get put into? Maybe if it's only in `packages/devtools` we can put this in `packages/devtools/.gitignore`.

packages/devtools/src/ganache/index.ts
```diff
@@ -8,6 +8,7 @@ import {Account} from '../types';
 import {GanacheDeployer} from './deployer';
 import {GanacheNCacheDeployer} from './deployer-with-cache';
 import {GanacheServer} from './server';
+import writeJsonFile = require('write-json-file');
```
snario: Is import assignment necessary here?

packages/devtools/src/ganache/index.ts
```diff
@@ -44,45 +46,88 @@ export const startOwnGanache = async (p: Partial<Params> = {}): Promise<GanacheS
...
+  const useShared = process.env.USE_GANACHE_DEPLOYMENT_CACHE === 'true';
+  const port = Number(process.env.GANACHE_PORT || 8545);
 
-  const deploymentsFile = process.env.GANACHE_DEPLOYMENTS_FILE;
-  const foundShared = await ganacheIsRunning(sharedPort);
-
-  if (!deploymentsFile) {
+  let server;
+  let deployer;
+  let type: 'shared' | 'individual';
```
snario: Isn't `type` a reserved keyword?

.gitignore
tomclose: It's in the root of the monorepo.

packages/devtools/src/ganache/index.ts
tomclose: No - I think vscode did that automatically and I didn't notice.

packages/devtools/src/ganache/index.ts
tomclose: Yes. Weird that that works ...

PR #737, opened 2019-12-11
[hub] ledger opening and funding test sequence

kerzhner:
An integration-like unit test that feeds ledger channel open + fund message sequence. The interesting changes are in `handle-wallet-message.test.ts`.


packages/hub/src/hub/handlers/__test__/handle-wallet-message.test.ts
```diff
@@ -7,15 +14,35 @@ function readSampleMessage(messageName) {
...
+}
 
-    const expectedAction: ChannelJoined = readSampleMessage('channel-joined');
-    const receivedAction = messageReleayRequested[0].actionToRelay;
-    // note: signatures are not validated
-    expect(receivedAction).toMatchObject(expectedAction);
+describe('handle-wallet-message', () => {
+  // note: signatures are not validated
+  it('open and fund ledger channel sequence', async () => {
+    await step<ChannelOpen, ChannelJoined>('channel-open', 'channel-joined');
```
tomclose: I'm wondering if providing the types adds anything here?

kerzhner: It would if there was a way to enforce a type on an object created from requiring a json module. That's something I am planning on looking into.

PR #738, opened 2019-12-11
[nitro-protocol] Bring nitro jest config in line with rps

geoknee:
- enables running of ts or contract tests separately
- no point running prettier before running tests
- no point recompiling contracts before running tests
- other tweaks to jest config


packages/rps/package.json
```diff
@@ -27,11 +27,11 @@
     "prepare": "yarn build:ci",
     "test:prepare": "yarn clearContracts && yarn contracts:compile",
     "test:app": "jest -c ./config/jest/jest.config.js",
-    "test:contracts": "yarn test:prepare && npx jest -c ./config/jest/jest.contracts.config.js",
-    "test": "run-s prettier:check test:contracts 'test:app --all'",
+    "test:contracts": "npx jest -c ./config/jest/jest.contracts.config.js",
+    "test": "run-s test:contracts 'test:app --all'",
     "test:ci": "yarn test:ci:contracts && yarn test:ci:app",
-    "test:ci:contracts": "yarn test:contracts --all --ci --runInBand --detectOpenHandles",
-    "test:ci:app": "yarn test:app --all --ci --runInBand --detectOpenHandles",
```
snario: Was `--detectOpenHandles` just here unnecessarily before?

geoknee: https://jestjs.io/docs/en/cli#--detectopenhandles 

> This option has a significant performance penalty and should only be used for debugging.

PR #739, opened 2019-12-11
[rps] Use <script> to import window.channelProvider

snario:
The `channel-provider` README states this is the recommended way of using it.

In the future perhaps somehow the embedded iframe might inject it in the page.

### DIFF
--- a/packages/rps/.eslintrc.js
+++ b/packages/rps/.eslintrc.js
@@ -52,11 +52,6 @@ module.exports = {
     es6: true,
   },
   parser: '@typescript-eslint/parser',
-  parserOptions: {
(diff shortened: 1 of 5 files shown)


lalexgap:
> The `channel-provider` README states this is the recommended way of using it.
> 
> In the future perhaps somehow the embedded iframe might inject it in the page.

I feel like there must be a better way of doing this using webpack, instead of just manually embedding the script on the page. Have you looked at https://webpack.js.org/plugins/provide-plugin/ ?

snario:
1. Blocked on #740 
2. I'll use the Provide Plugin

PR #740, opened 2019-12-11
[rps] Fix RPS linting configuration

snario:
Seems like the `lint:check` command was broken recently. This fixes it.

### DIFF
--- a/packages/rps/.gitignore
+++ b/packages/rps/.gitignore
@@ -16,6 +16,7 @@
 .env.test.local
 .env.production.local
 .vscode/settings.json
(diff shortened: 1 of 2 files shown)


Alonski:
I set this up to also run Prettier on files that ESLint doesn't check... Why is this breaking?

snario:
@Alonski sorry I missed that this was only for `scss` and `json`. Fixed.

geoknee:
I ran into the above problem again. It may be related to this https://github.com/okonet/lint-staged/issues/677.

PR #741, opened 2019-12-11
[hub] default environment variables for development

kerzhner:


kerzhner:
Closing and including defaults as part of virtual funding work.

PR #745, opened 2019-12-11
[rps] Inject window.channelProvider with webpack

snario:
Uses `webpack` to inject `window.channelProvider` into `rps`.

Note that I do _not_ use `ProvidePlugin` because that has some delay for when the actual `window.channelProvider` is injected which is hard to detect in the app. Instead, we load it inside `entry`. See [here](https://stackoverflow.com/questions/39888395/webpack-and-external-libs-provideplugin-vs-entry-vs-global-import/39891365) for the tradeoffs.

Also, this properly types `window` on `rps` and `wallet`.

Also, on `rps` this adds some state management for the existence of a `window.channelProvider` and does not let the UI render if there is none there.


snario:
The approach you're describing is already roughly how it works. However, since `rps` is built using webpack instead of relying on `index.html` (which the first version of this PR did, see https://github.com/statechannels/monorepo/pull/739), we use our webpack config. The point being, this webpack configuration is specifically the best way to do it for this `rps` app, not for all apps.

snario:
At the moment, `@statechannels/channel-provider` attaches to `window` inherently when it is imported onto the screen. See [the README](https://github.com/statechannels/monorepo/blob/master/packages/channel-provider/README.md#usage).

tomclose:
👍 I didn't quite understand what was going on here ... seems good to me now!

PR #753, opened 2019-12-12
Virtual funding for RPS

kerzhner:
https://github.com/statechannels/monorepo/pull/742 should precede this PR. 

Also, integration tests are failing on this PR.


lalexgap:
> Also, integration tests are failing on this PR. 

This is caused by the fact that we don't have a `hub` running when running the `integration-test`s and we're now trying to do `VirtualFunding`. I think I can make a quick change so that the strategy is an .env variable, so we can keep the `integration-test`s running with ledger funding for now.

kerzhner:
I think that's a solid idea. Would be good to get this work merged in quickly so that others can start playing around with virtual funding and discovering remaining issues.

lalexgap:
> I think that's a solid idea. Would be good to get this work merged in quickly so that others can start playing around with virtual funding and discovering remaining issues.

Updated here aa42884aa7f565ef142f0a23fac0a02c6a7ff386

geoknee:
I confirmed this worked on my machine 🌋 .

Issue #710, opened 2019-12-09
[wallet] The wallet should handle duplicate messages


lalexgap:
If a timeout occurs the `channel-provider` tries to resend the message to the `wallet`. Currently if this happens the `wallet` ends up getting in an invalid state since it doesn't realize the messages are duplicates. We should handle this scenario.

PR #756, opened 2019-12-12
[wallet] Add state to track in-progress JSON-RPC responses

snario:
Handles #710. Not sure how best to test this.

**Question**. Is storing this in `SharedData` the "right" strategy?

### DIFF
--- a/packages/wallet/src/redux/sagas/messaging/__tests__/message-handler.test.ts
+++ b/packages/wallet/src/redux/sagas/messaging/__tests__/message-handler.test.ts
@@ -44,6 +44,38 @@ describe("message listener", () => {
     );
   });
 
+  // TODO: Figure out how to test this behaviour
+  // it("does not execute code twice for the same request ID before responding", async () => {
+  //   const requestMessage = {
+  //     jsonrpc: "2.0",
+  //     method: "GetAddress",
+  //     id: 1,
+  //     params: {}
+  //   };
+
+  //   let responses = 0;
+
+  //   expectSaga(messageHandler, requestMessage, "localhost")
+  //     .withState(initialState)
+  //     .provide([
+  //       [matchers.fork.fn(messageSender), 0],
+  //       [matchers.select(getAddress), "poop"]
+  //     ])
+  //     .fork(messageSender, addressResponse({id: 1, address: "poop"}))
+  //     .run();
+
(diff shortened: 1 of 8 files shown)


packages/wallet/src/redux/sagas/messaging/__tests__/message-handler.test.ts
```diff
@@ -563,11 +595,11 @@ describe("message listener", () => {
         turnNum: 1
       };
 
-      expect(effects.put[0].payload.action).toMatchObject({
+      expect(effects.put[1].payload.action).toMatchObject({
         type: "WALLET.APPLICATION.OWN_STATE_RECEIVED",
         state: nextState
       });
-      expect(effects.put[1].payload.action).toMatchObject({
+      expect(effects.put[2].payload.action).toMatchObject({
```
snario: Also these kinds of diffs are annoying. @lalexgap any ideas on ways we could possibly handle these better? Maybe matching `put` calls based on the thing being put? e.g., "expect _some_ `put` to match object"

lalexgap: We could create a helper that does that similar to `itSendsThisMessage`

lalexgap:
> Handles #710. Not sure how best to test this.
> 
> **Question**. Is storing this in `SharedData` the "right" strategy?

Hmm, I thought I had commented on this PR but I must of forgot to hit submit :( 

Using `SharedData` doesn't seem right to me but I can't think of a better solution. I wonder if we should hold off on this for now. Now that we're starting to look at storage outside of redux there might be a better way of doing this.

Also I think this only protects against handling duplicate requests at the same time. However I think a more common scenario is receiving a message, handling it, and then receiving it again. To solve that I think we might need some constraints on the `id`. (IE: The wallet expects them to be unique)?

snario:
> Also I think this only protects against handling duplicate requests at the same time.
Right.

> However I think a more common scenario is receiving a message, handling it, and then receiving it again.
Maybe? I'm not sure. What are JSON-RPC servers supposed to do? Is there a standard?

geoknee:
Closing for now as this has been hanging around for a while and it's not clear what we want to do with it. Please reopen if keen on getting it merged.

PR #759, opened 2019-12-12
Hub: state channel address configuration

kerzhner:
With this PR:
- we can define `HUB_ADDRESS` and `HUB_PRIVATE_KEY` once in the root of the monorepo.
- have a clearer readme on starting the hub.

### DIFF
--- a/packages/hub/readme.md
+++ b/packages/hub/readme.md
@@ -37,7 +37,7 @@ $ yarn hub:watch (will rebuild app on file change)
 
 ```
 
-### Interacting with the hub from a browser
+### Establishing a virtual channel between clients through the hub
 
 **NOTE**: Running this package makes a connection to a shared external Firebase database. So, to avoid colliding with other developers also running this package, set the environment variable `HUB_ADDRESS` to one that is likely not being used by any other developer for local development purposes.
 
--- a/packages/hub/readme.md
+++ b/packages/hub/readme.md
@@ -39,15 +39,16 @@ $ yarn hub:watch (will rebuild app on file change)
 
 ### Interacting with the hub from a browser
 
-**NOTE**: it is important to define a unique `HUB_ADDRESS` if others are running the hub locally since all hub instances use the same Firebase realtime database.
+**NOTE**: Running this package makes a connection to a shared external Firebase database. So, to avoid colliding with other developers also running this package, set the environment variable `HUB_ADDRESS` to one that is likely not being used by any other developer for local development purposes.
 
 To connect to the `hub` from the browser `wallet`, the `hub` and the browser `wallet` need to:
 
 - Share the state-channel address of the hub. A good way to do so is to create a `.env.development.local` in the monorepo root with `HUB_ADDRESS` and `HUB_PRIVATE_KEY` defined.
 - Point to the same local Ganache server. Configure your `.env` files accordingly. This should work without any modifications.
+- Point to the same shared local Ganache server. This should work without any modifications. To see which ports are being used by the `hub` and `wallet`, and to verify they are the same, you can reference the `GANACHE_PORT` environment variable which by default is set in `.env` of each package.
 - Point to the same contract addresses on Ganache. This will be the case if the hub and the client wallet point to the same Ganache server.
 
-You will also need to make sure that the hub's blockchain address has funds. The default hub blockchain address `HUB_SIGNER_ADDRESS` is in [constants.ts](https://github.com/statechannels/monorepo/blob/master/packages/hub/src/constants.ts)
+You will also need to make sure that the hub's blockchain address has funds. The default hub blockchain address `HUB_SIGNER_ADDRESS` is in [constants.ts](https://github.com/statechannels/monorepo/blob/master/packages/hub/src/constants.ts). This address will have funds by default. Ganache is started with [these funded private keys](https://github.com/statechannels/monorepo/blob/hub-address/packages/devtools/src/constants.ts). Consequently, feel free to substitute any of these address/private key pairs for `HUB_SIGNER_ADDRESS` and `HUB_SIGNER_PRIVATE_KEY`.
 
 ## Testing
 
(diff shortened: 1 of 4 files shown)


snario:
What do you mean by defining them "in the root of the monorepo"? You mean in a `<root>/.env` file?

kerzhner:
@snario Apologies, poorly phrased PR comment. Mind looking at the readme diff of this PR and seeing if it makes more sense to you?

packages/hub/readme.md
```diff
@@ -39,13 +39,15 @@ $ yarn hub:watch (will rebuild app on file change)
 
 ### Interacting with the hub from a browser
```
snario: > a browser
Could be "from a `wallet`" to be precise

packages/hub/readme.md
```diff
@@ -39,13 +39,15 @@ $ yarn hub:watch (will rebuild app on file change)
...
 
+**NOTE**: it is important to define a unique `HUB_ADDRESS` if others are running the hub locally since all hub instances use the same Firebase realtime database.
```
snario: This could be clearer. Perhaps

> **NOTE**: Running this package makes a connection to a shared external Firebase database. So, to avoid colliding with other developers also testing this package, set the environment variable `HUB_ADDRESS` to one that is not being used by any other developer for testing or local development purposes.

packages/hub/readme.md
```diff
@@ -39,13 +39,15 @@ $ yarn hub:watch (will rebuild app on file change)
...
+
 To play against the hub from the browser client, the hub and the browser need to:
```
snario:
```suggestion
To connect to the `hub` from the browser `wallet`, the `hub` and the browser `wallet` need to:
```

packages/hub/readme.md
```diff
@@ -39,13 +39,15 @@ $ yarn hub:watch (will rebuild app on file change)
...
 
-- Share the state-channel address of the hub. A good way to do so is to create a `.env.development.local` in the monorepo root with HUB_ADDRESS and HUB_PRIVATE_KEY defined.
-- Point to the same local Ganache server. Configure your `.env.*.local` files accordingly.
-- Point to the same contract addresses on Ganache.
+- Share the state-channel address of the hub. A good way to do so is to create a `.env.development.local` in the monorepo root with `HUB_ADDRESS` and `HUB_PRIVATE_KEY` defined.
```
snario:
```suggestion
- Share the same address of the hub. A good way to do so is to create a `.env.development.local` in the monorepo root with `HUB_ADDRESS` and `HUB_PRIVATE_KEY` defined.
```

packages/hub/readme.md
```diff
@@ -39,13 +39,15 @@ $ yarn hub:watch (will rebuild app on file change)
...
+- Point to the same local Ganache server. Configure your `.env` files accordingly. This should work without any modifications.
```
snario:
```suggestion
- Point to the same shared local Ganache server. This should work without any modifications. To see which ports are being used by the `hub` and `wallet`, and to verify they are the same, you can reference the `GANACHE_PORT` environment variable which by default is set in `.env.development` of each package.
```

packages/hub/readme.md
```diff
@@ -39,13 +39,15 @@ $ yarn hub:watch (will rebuild app on file change)
...
+- Point to the same contract addresses on Ganache. This will be the case if the hub and the client wallet point to the same Ganache server.
 
-You will also need to make sure that the hub's address has funds. You can find the hub address in [constants.ts](https://github.com/statechannels/monorepo/blob/master/packages/hub/src/constants.ts)
+You will also need to make sure that the hub's blockchain address has funds. The default hub blockchain address `HUB_SIGNER_ADDRESS` is in [constants.ts](https://github.com/statechannels/monorepo/blob/master/packages/hub/src/constants.ts)
```
snario: Maybe add a note on how to send funds to this address for your local ganache instance? We could also possibly have some script which does this automatically.

andrewgordstewart: Perhaps the `start:ganache` script can do this on start.

snario: I think `ganache` allows you to set initial funding states of particular addresses with the `accounts` field. We could set `HUB_SIGNER_ADDRESS` to some big number.

packages/hub/readme.md
kerzhner: It's a bit tricky, as from a user's perspective, the interaction is happening from rps. Also, rps is the module sending messages to the hub.  What about "Establishing a virtual channel between clients through the hub"?

packages/hub/readme.md
kerzhner: There are 2 addresses the hub uses: one for transaction signing and one for state signing. If this line is not clear, open to suggestions though!

packages/hub/readme.md
kerzhner: At the moment, ganache env vars are set in `.env` (and not `.env.development`) in all packages.

packages/hub/readme.md
kerzhner: As long as `HUB_SIGNER_ADDRESS` is one of these addresses, it will have funds. Agreed that we can be clearer in the readme.

packages/hub/readme.md
kerzhner: Addressed in 04005dd79

packages/hub/readme.md
kerzhner: Hopefully clearer in 04005dd79e9bb15826f6d186e91b33ba99d4a0d6

packages/hub/readme.md
kerzhner: Hopefully clearer in 04005dd79e9bb15826f6d186e91b33ba99d4a0d6.

packages/hub/readme.md
kerzhner: 2255c2c49

PR #760, opened 2019-12-12
[wallet] Use symbolic link for web3.min.js

snario:
We use this for puppeteer testing but our `node_modules` already has the same file.


snario:
I'm struggling to understand why the `integration-test` job is failing. I `ssh`-ed into the job and verified the file is there and set up correctly. Also, it works when run locally so I assume it is not a `puppeteer` issue.

lalexgap:
> I'm struggling to understand why the `integration-test` job is failing. I `ssh`-ed into the job and verified the file is there and set up correctly. Also, it works when run locally so I assume it is not a `puppeteer` issue.

I just tried this locally and it seems to fail for me as well.

PR #763, opened 2019-12-12
[wallet] Integration Test Changes

lalexgap:
Two changes here to help (hopefully) with the stability of the `wallet` integration tests.

- Switched to the `networkidle0` condition when loading the `wallet` page to hopefully ensure it's fully loaded before being used.
- Clear the message listeners when the test is done before we close the browsers to prevent an error being thrown by a message listener trying to call `window.Postmessage` after the browsers have been closed.


lalexgap:
I'm still seeing issues on #764 so closing this for now.

PR #764, opened 2019-12-12
[wallet] Implement CloseChannel and get it working

lalexgap:
Create a `CloseChannel` method and gets it mostly working.

Currently some things are broken / still need to be investigated (I'll create issues for these):
- ~Concluding fails if player B attempts to conclude (on their opponent's turn)~
- The funds aren't sent to user's address (most likely because we're passing in the wallet address as the `destination`) when creating the channel.
- The `wallet` withdrawal screen is inaccurate. (It currently shows a random address instead of the `destination` on the `participants`
- Need to add tests for `CloseLedgerChannel`'s new `ClearedToSend` functionality.
- After concluding `rps` shows a `Withdraw` button even though the `wallet` has already triggered the withdraw.
-  There is an option to leave a channel "open" which probably doesn't work


packages/client-api-schema/schema/request.json
```diff
@@ -69,6 +76,14 @@
...
             "params": {"$ref": "update-channel.json#/definitions/UpdateChannelParams"}
           }
+        },
+        {
+          "if": {
+            "properties": {"method": {"const": "CloseChannel"}}
+          },
+          "then": {
+            "params": {"$ref": "close-channel.json#/definitions/CloseChannelParams"}
+          }
```
snario: Ah sorry I missed this.

packages/rps/src/redux/message-service/firebase-inbox-listener.ts
```diff
@@ -3,13 +3,17 @@ import {take, call} from 'redux-saga/effects';
 import {reduxSagaFirebase} from '../../gateways/firebase';
 import {RPSChannelClient} from '../../utils/rps-channel-client';
 import {Message} from '@statechannels/channel-client';
+import {buffers} from 'redux-saga';
 
 export function* firebaseInboxListener(client: RPSChannelClient) {
   const address: string = yield call([client, 'getAddress']);
+  // TODO: Redux saga firebase accepts a buffer but the type definition hasn't been updated in the latest version (0.15.0)
+  // so we just cast as any for now
```
snario: Is there an issue for this on their repo? I think I'm the one that removed this argument

lalexgap: I've opened this https://github.com/n6g7/redux-saga-firebase/issues/567

PR #768, opened 2019-12-13
[rps] Delete messages from firebase (actually)

### LINKED ISSUE #746: [rps] Clean up firebase messages after consumption

geoknee:
In the end it was just a `.toLowerCase()` issue.

Closes #746 .


packages/rps/src/redux/message-service/firebase-inbox-listener.ts
```diff
@@ -1,21 +1,30 @@
...
 import {Message} from '@statechannels/channel-client';
 
 export function* firebaseInboxListener(client: RPSChannelClient) {
-  const address: string = yield call([client, 'getAddress']);
+  const address: string = (yield call([client, 'getAddress'])).toLowerCase();
+  // ^ ensure this matches the to in message-queued-listener.ts
   const channel = yield call(
     reduxSagaFirebase.database.channel,
-    `/messages/${address.toLowerCase()}`,
+    `/messages/${address}`,
```
kerzhner: Question: what do the address keys in firebase look like? Are the keys lower case?

My preference is to always have the keys as lower case. Otherwise, it might be hard for the client and hub to agree on what keys should look like.

geoknee: With this PR they will always be lower case. I added comments to remind us to maintain that convention across the files that matter in this package.

geoknee: I removed the `.toLowerCase()` here because `address` is constructed to be lowercase above.

PR #770, opened 2019-12-13
[hub] pino for logging

kerzhner:
The goal of this PR is to create more manageable logs with pino (fixes https://github.com/statechannels/monorepo/issues/755). To accomplish this, the work expanded to:
- `firebase-feeder.ts`: a simple way to feed messages to the hub through firebase.
- no longer forking a firebase watcher process. Instead, the firebase listener is created in the main `server` process. The fork was legacy code from when the hub was an http server.
- clean ups: removed redundant Objection initialization, cleaned up mixing `async/away` and `then` syntax.

Also fixes https://github.com/statechannels/monorepo/issues/765.


snario:
Can we establish a convention for log levels? I think this would be very helpful

kerzhner:
That sounds good to me. Are you thinking of a log convention for all of our packages or just the hub? Is there an existing, standard convention that we should use?

snario:
It would be nice for all packages! But, just the hub is fine for now too. Right now logging just feels like something that is generally ignored / not thought about much across the team.

PR #771, opened 2019-12-13
[rps] Wrap rps test into an integration test

snario:
## TODO

- [x] Get `rps` running inside integration test environment\
- [x] Figure out how to use shared ganache in CircleCI I guess
- [x] Make sure environment variables are set correctly

Some problems encountered...
- `USE_VIRTUAL_FUNDING=TRUE` makes these tests hard to configure
- Unreliable, sometimes messaging doesn't go through

### DIFF
--- a/packages/wallet/README.md
+++ b/packages/wallet/README.md
@@ -47,6 +47,6 @@ There are a few integration tests inside the [`puppeteer`](./puppeteer) folder w
 NODE_ENV=test yarn ts-node -O '{"module":"commonjs","noUnusedLocals":false}' ./puppeteer/scripts/funding.ts
 ```
 
-You will want to call `setsetUpBrowser` with the `false` argument to use the Chromium UI on your Mac.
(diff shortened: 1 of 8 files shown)


packages/wallet/README.md
```diff
@@ -38,3 +38,15 @@ yarn test:ci:app
...
+
+## Integration testing with `puppeteer`
+
+There are a few integration tests inside the [`puppeteer`](./puppeteer) folder which can be run with the `yarn test:puppeteer` script. If you want to run these tests and debug the console of the instantiated Chromium windows at the same time, you might want to use `ts-node`. Here is an example, running the `funding` script.
+
+```bash
+NODE_ENV=test yarn ts-node -O '{"module":"commonjs","noUnusedLocals":false}' ./puppeteer/scripts/funding.ts
+```
+
+You will want to call `setsetUpBrowser` with the `false` argument to use the Chromium UI on your Mac.
```
lalexgap:
```suggestion
You will want to call `setupBrowser` with the `false` argument to use the Chromium UI on your Mac.
```

PR #772, opened 2019-12-15
[nitro-protocol] Gas savings study 

geoknee:
This PR includes:

- new function that outputs a gitignored `<testName>.gas.md` file showing gas consumed for each test case that calls it
- python script that generates a bar chart showing apples-to-apples comparison in gas consumption, with hard-coded values copy-pasted from these files on 
   - `master` (base branch)
   - `legacy-contracts` (branch that uses legacy contracts with our latest testing infrastructure)
- blog post around this chart, along with another image showing schematics of our gas optimization strategy

gas-savings


geoknee:
Approval / merge for this PR does not indicate that the blog post is in its final form.

PR #773, opened 2019-12-16
[rps] Remove dependence on web3

### LINKED ISSUE #694: [rps] Use ethereumProvider over web3
A warning in the `rps` demo you get is:

```
MetaMask: web3 will be deprecated in the near future in favor of the ethereumProvider
https://medium.com/metamask/4a899ad6e59e
```

Should be straightforward to resolve this.

### LINKED ISSUE #698: [rps] Check that you are connected to a working chain
If MetaMask is set to `localhost:ABCD` but there is no `ganache` running at `localhost:ABCD` then there will be this message in the console:

Screen Shot 2019-12-09 at 12 52 10

It's not obvious that MetaMask is just pointing to a bad URL.

geoknee:
Closes #694 .

Closes #698.

Lets `ethers` pick up the slack.

Also clears a warning about reloading on network change, and fixes the functionality whereby the app will not function if you have selected e.g. Ropsten in MM but are expecting a dev network via the TARGET_NETWORK environment variable.


packages/rps/src/redux/metamask/actions.ts
```diff
@@ -2,7 +2,6 @@ export const METAMASK_ERROR = 'METAMASK.ERROR';
 export const METAMASK_SUCCESS = 'METAMASK.SUCCESS';
 export const enum MetamaskErrorType {
   WrongNetwork = 'WrongNetwork',
-  NoWeb3 = 'NoWeb3',
```
snario: It seems to me this is catching the case of MetaMask not even being installed, and so we probably still need some additional error type for that.

packages/rps/src/redux/metamask/saga.ts
```diff
@@ -1,47 +1,52 @@
...
+  kovan: {
+    networkId: 42,
+  },
+};
 
-  try {
+export default function* checkMetamask() {
+  if (typeof window.ethereum !== 'undefined') {
+    try {
+      window.ethereum.autoRefreshOnNetworkChange = false;
```
snario: Why false?

geoknee: This is just to clear the error from the console, since the setting will soon vanish (see link in commit message). We are now essentially turning off a feature that is soon to disappear, before rolling our own solution just below (since we rely on the reloading behaviour).

packages/rps/src/redux/metamask/actions.ts
geoknee: Good point. Will fix.

geoknee: See 11c42d1

geoknee: Screenshot 2019-12-16 at 15 07 14

PR #774, opened 2019-12-16
[rps][wallet] Mirror vscode settings for packages migrated to eslint

geoknee:
- allows vscode eslint plugin pickup settings file correctly

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,8 @@
     {"directory": "packages/hub", "changeProcessCWD": true},
     {"directory": "packages/jest-gas-reporter", "changeProcessCWD": true},
     {"directory": "packages/nitro-protocol", "changeProcessCWD": true},
-    {"directory": "packages/channel-client", "changeProcessCWD": true}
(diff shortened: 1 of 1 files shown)


snario:
Is any package not on `eslint`? If not we could use a glob

andrewgordstewart:
> Is any package not on `eslint`? If not we could use a glob

Why is it important that all packages use `eslint`?

geoknee:
`app-wallet-interface` is not, for example. I added these packages only because they had a `.eslintrc.js` file.

.vscode/settings.json
```diff
@@ -7,6 +7,8 @@
     {"directory": "packages/hub", "changeProcessCWD": true},
     {"directory": "packages/jest-gas-reporter", "changeProcessCWD": true},
     {"directory": "packages/nitro-protocol", "changeProcessCWD": true},
-    {"directory": "packages/channel-client", "changeProcessCWD": true}
+    {"directory": "packages/channel-client", "changeProcessCWD": true},
```
kerzhner: Nit, might be nice to keep the package name list organized alphabetically.

PR #777, opened 2019-12-17
[rps] Resintate tests

### LINKED ISSUE #713: [rps] AutoOpponent/Rps Channel Client tests need to be fixed
Currently the tests in `rps-channel-client.test.ts` and `auto-opponent.test.ts` are disabled. They were failing after the switch in `rps` to use a real `channelClient`. They probably need to just mock the `channel-client` before running.

geoknee:
Closes #713 .

This PR undisables these tests, which were set to "skip" recently. 

1. The auto-opponent and test use the `FakeChannelClient`

2. The test for `RPSChannelClient` class mocks up a `ChannelClient` and does not assert anything that requires the underlying provider to exist (for example, an event being emitted). It also expands the test coverage to 

 - [x] `onMessageQueued()`
 - [x] `onChannelUpdated()`
 - [x] `createChannel()`
 - [x] `joinChannel()`
 - [x] `closeChannel()`
 - [x] `updateChannel()`
 - [x] `pushMessage()`
 - [x] `getAddress()`

👀 Additional modifications to non-test code:
- A wrapper `Client` will not spawn the underlying `Client`: it must be passed in, instead (makes testing via mocks far easier)
- (related) Channel Client's `enable` method no longer does anything.
- Encoding mismatch fixed (uncovered by new testing approach).
- Wrapper client's `pushMessage` method passes through (i.e. returns) the return value of the inner class instance's `pushMessage`

### DIFF
--- a/packages/rps/src/redux/store.ts
+++ b/packages/rps/src/redux/store.ts
@@ -1,5 +1,5 @@
 import {applyMiddleware, compose, createStore} from 'redux';
-import {fork, call} from 'redux-saga/effects';
+import {fork} from 'redux-saga/effects';
 import createSagaMiddleware from 'redux-saga';
 
 import reducer from './reducer';
--- a/packages/rps/src/redux/store.ts
+++ b/packages/rps/src/redux/store.ts
@@ -14,6 +14,7 @@ import {channelUpdatedListener} from './message-service/channel-updated-listener
 import {messageQueuedListener} from './message-service/message-queued-listener';
 import {gameSaga} from './game/saga';
 import {autoPlayer, autoOpponent} from './auto-opponent';
+import {ChannelClient} from '@statechannels/channel-client';
 
 const composeEnhancers = (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
 const enhancers = composeEnhancers(applyMiddleware(sagaMiddleware));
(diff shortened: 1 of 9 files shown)


packages/rps/src/redux/store.ts
```diff
@@ -26,9 +27,7 @@ function* rootSaga() {
   const channelProviderEnabled = yield stateChannelWalletSaga();
 
   if (channelProviderEnabled) {
-    const client = new RPSChannelClient();
-
-    yield call([client, client.enable]);
+    const client = new RPSChannelClient(new ChannelClient(window.channelProvider));
```
geoknee: These lines encapsulate fairly well the new approach (of this PR) to instantiating the double wrapper around `window.channelProvider`. There were comments in the existing code intimating that this was the desired approach in any case, and it also makes testing far easier.

packages/rps/src/utils/rps-channel-client.ts
```diff
@@ -1,21 +1,15 @@
...
 import {RPS_ADDRESS} from '../constants';
 import {bigNumberify} from 'ethers/utils';
 
 // This class wraps the channel client converting the request/response formats to those used in the app
 
 export class RPSChannelClient {
-  channelClient: ChannelClientInterface;
+  constructor(private readonly channelClient: ChannelClientInterface) {}
 
   async enable() {
```
cf19drofxots: Why not remove this method?

geoknee: It's not obvious from the syntax, but this constructor is going to set `this.channelClient` to what you pass in to it. So it is the same as writing {`this.channelClient = channelClient`} in the body. https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties

geoknee: (and we need `this.channelClient` to be available to most of the class methods).

cf19drofxots: Oh hah no I meant line 11 with the function implementation of `enable` being empty. I get what's happening on the line you explained :)

PR #784, opened 2019-12-18
[rps] Improve user login / metamask enable experience

geoknee:
rps-login

As per [recommendation](https://gist.github.com/rekmarks/d318677c8fc89e5f7a2f526e00a0768a#file-newprovider-js-L71-L73), metamask will not pop up until the user has confirmed they want to use the app. 

Note that metamask will remember the approval for a connection to `localhost` (or `rps.com`) and therefore will not re-prompt next time unless you delete the connection.

Screenshot 2019-12-18 at 11 07 58

I also disabled persistence for firebase auth, which was causing any user who had previously logged in to be bumped straight past the homepage to the "Name / Twitter Handle" modal without clicking anything.

References
---
See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1102.md

### DIFF
--- a/packages/rps/src/gateways/firebase.ts
+++ b/packages/rps/src/gateways/firebase.ts
@@ -14,6 +14,7 @@ const config = {
 };
 
 const fire = firebase.initializeApp(config);
+firebase.auth().setPersistence(firebase.auth.Auth.Persistence.NONE);
 export const reduxSagaFirebase = new ReduxSagaFirebase(fire);
 export const authProvider = new firebase.auth.GoogleAuthProvider();
 export default fire;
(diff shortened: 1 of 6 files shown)


tomclose:
What does this look like if metamask isn't installed? Or if you open the page in safari?

geoknee:
On safari, you can click "Start Playing", enter your name, and then "Submit" and then
Screenshot 2019-12-18 at 11 52 11

it should be the same for any broswer that does not have metamask.

If user rejects `ethereum.enable()`: 

Screenshot 2019-12-18 at 11 54 23
 and a refresh is required.

PR #785, opened 2019-12-18
[rps] Fix rootSaga to allow auto opponent to work

geoknee:
This mirrors the behaviour in `auto-opponent.test.ts` but switches on environment variables. Essentially, do not try and talk to the wallet, use a `FakeWalletClient`, and do not start certain sagas.

### DIFF
--- a/packages/rps/src/redux/store.ts
+++ b/packages/rps/src/redux/store.ts
@@ -14,7 +14,7 @@ import {channelUpdatedListener} from './message-service/channel-updated-listener
 import {messageQueuedListener} from './message-service/message-queued-listener';
 import {gameSaga} from './game/saga';
 import {autoPlayer, autoOpponent} from './auto-opponent';
-import {ChannelClient} from '@statechannels/channel-client';
+import {ChannelClient, FakeChannelClient} from '@statechannels/channel-client';
 
 const composeEnhancers = (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
 const enhancers = composeEnhancers(applyMiddleware(sagaMiddleware));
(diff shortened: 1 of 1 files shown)


packages/rps/src/redux/store.ts
```diff
@@ -24,28 +24,32 @@ const store = createStore(reducer, enhancers);
...
-    } else {
-      yield fork(firebaseInboxListener, client);
-      yield fork(openGameSaga);
-    }
+  }
+
+  yield fork(gameSaga, client);
+  yield fork(channelUpdatedListener, client);
+
+  if (process.env.AUTO_PLAYER === 'A') {
```
kerzhner: Nit, might be more elegant as:
``` typescript
if (process.env.AUTO_PLAYER) {
  yield fork(autoPlayer, process.env.AUTO_PLAYER);
}

File: packages/rps/src/redux/store.ts [hunk 2]
```diff
@@ -24,28 +24,32 @@ const store = createStore(reducer, enhancers);
...
+    yield fork(autoPlayer, 'A');
+  } else if (process.env.AUTO_PLAYER === 'B') {
+    yield fork(autoPlayer, 'B');
+  }
+
+  if (process.env.AUTO_OPPONENT === 'A') {
```
kerzhner: Nit, might be more elegant as:
``` typescript
if (process.env.AUTO_OPPONENT) {
  yield fork(autoOpponent, process.env.AUTO_OPPONENT, client);
}
```

packages/rps/src/redux/store.ts
```diff
@@ -24,28 +24,32 @@ const store = createStore(reducer, enhancers);
 function* rootSaga() {
   yield fork(loginSaga);
 
-  const channelProviderEnabled = yield stateChannelWalletSaga();
-
-  if (channelProviderEnabled) {
-    const client = new RPSChannelClient(new ChannelClient(window.channelProvider));
-    yield fork(channelUpdatedListener, client);
+  let client;
+  if (process.env.AUTO_OPPONENT === 'A' || process.env.AUTO_OPPONENT === 'B') {
```
kerzhner: Would this work?
```suggestion
  if (process.env.AUTO_OPPONENT) {
```

packages/rps/src/redux/store.ts
geoknee: It would be, but that causes a typescript error under the current implementation.

packages/rps/src/redux/store.ts
geoknee: It would be, but that causes a typescript error under the current implementation.

packages/rps/src/redux/store.ts
geoknee: I think it would, but I actually prefer the current way in case in future someone wants to set `AUTO_OPPONENT=none` or any other unexpected value.

PR #790, opened 2019-12-19
Move Wallet-Spec repo into monorepo

lalexgap:
Moves the `wallet-spec` work that @andrewgordstewart has done into the mono-repo.


packages/wallet-specs/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "test": "jest"
+  },
+  "dependencies": {
+    "xstate": "^4.7"
+  },
+  "devDependencies": {
+    "@types/jest": "^24.0.23",
+    "@types/node": "^12.12.7",
+    "fs": "^0.0.1-security",
+    "i": "^0.3.6",
```
andrewgordstewart: I don't recognize this.
```suggestion
```

packages/wallet-specs/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "jest": "^24.9.0",
+    "madge": "^3.6.0",
```
andrewgordstewart: This was installed to detect circular dependencies; it's not really needed as a dev dependency.
```suggestion
```

packages/wallet-specs/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "npm": "^6.13.1",
```
andrewgordstewart: This is probably accidental?
```suggestion
```

PR #793, opened 2019-12-19
[wallet-specs] Monorepo cleanups

snario:
- Adds eslint
- GItignores built files
- Adds some dependencies


packages/wallet-specs/package.json
```diff
@@ -24,6 +34,17 @@
     "tslint": "5.20.0",
     "tslint-config-prettier": "1.18.0",
     "tslint-plugin-prettier": "2.0.1",
-    "typescript": "^3.7.2"
+    "typescript": "^3.7.2",
+    "@typescript-eslint/eslint-plugin": "^2.7.0",
+    "@typescript-eslint/parser": "^2.8.0",
+    "eslint": "^6.6.0",
+    "eslint-config-prettier": "^6.7.0",
```
andrewgordstewart: Should these be dev dependencies of the root repo? The root `eslintrc.js` uses them ...

packages/wallet-specs/src/ChannelStoreEntry.ts
```diff
@@ -79,30 +73,28 @@ export class ChannelStoreEntry implements IChannelStoreEntry {
...
-      funding,
+      funding
     }: IChannelStoreEntry = this;
     return {
       supportedState,
       unsupportedStates,
       privateKey,
       participants,
       channel,
-      funding,
```
andrewgordstewart: I like my trailing commas. Not having them is really annoying in a few cases.

packages/wallet-specs/package.json
snario: Probably, yes

packages/wallet-specs/src/ChannelStoreEntry.ts
snario: I like trailing commas too. I'm in favour of turning them on in `packages/wallet-specs/.prettierrc.js` and eventually for all of the packages.

PR #795, opened 2019-12-19
[DO NOT MERGE] Remove build:ci yarn command from rps and wallet packages

kerzhner:
Fixes https://github.com/statechannels/monorepo/issues/794.

### DIFF
--- a/packages/rps/package.json
+++ b/packages/rps/package.json
@@ -9,9 +9,10 @@
     "firebase/**/grpc": "^1.16.1"
   },
   "scripts": {
+    "build": "run-s clearContracts build:typescript build:webpack",
+    "build:ci": "run-s build:typescript contracts:compile",
     "build:typescript": "npx tsc -b",
     "build:webpack": "CI=false node scripts/build.js",
-    "build": "run-s clearContracts build:typescript build:webpack",
     "clearContracts": "rm -rf ./build/contracts",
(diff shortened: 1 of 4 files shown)


packages/rps/package.json
```diff
@@ -23,7 +22,7 @@
     "eslint:write": "eslint \"**/*.{js,ts,tsx}\" --fix --max-warnings=0",
     "start": "yarn contracts:compile && node scripts/start.js",
     "start:shared-ganache": "NODE_ENV=development npx start-shared-ganache",
-    "prepare": "yarn build:ci",
+    "prepare": "run-s clearContracts build:typescript contracts:compile",
```
andrewgordstewart: I don't think `clearContracts` is needed.

kerzhner:
@andrewgordstewart do you think we should only have a `build` script defined for the packages that are a dependency in another monorepo package?

andrewgordstewart:
I wasn't quite correct in what I said: some packages should have their _`prepare`_ command removed. The `prepare` command should be run precisely in the packages where preparation is necessary.

packages/rps/package.json
kerzhner: Good call.

kerzhner:
Closing in favor of https://github.com/statechannels/monorepo/pull/798

PR #798, opened 2019-12-20
Remove unnecessary prepare work

andrewgordstewart:


packages/channel-provider/package.json
```diff
@@ -14,7 +14,6 @@
     "prepublish": "webpack",
     "build": "webpack",
     "build:typescript": "tsc -b .",
-    "prepare": "yarn build",
```
kerzhner: `channel-client` and `embedded-wallet` import `channel-provider`. `eslint` will fail during the circle `build` job.

packages/rps/package.json
```diff
@@ -9,11 +9,10 @@
     "firebase/**/grpc": "^1.16.1"
   },
   "scripts": {
-    "build:ci": "run-s clearContracts build:typescript contracts:compile",
+    "build:ci": "run-s build:typescript contracts:compile",
```
kerzhner: Do contracts need to be compiled? Contracts are already compiled (on every circle run) in `prepare` job.

packages/channel-provider/package.json
kerzhner: The prepare command should probably be switched to run `yarn build:typescript`

packages/rps/package.json
andrewgordstewart: Good question. On the one hand, it's nice for `build:ci` to do the complete build; the whole _point_ of `build:ci` is to check that the package can be built from start to finish. On the other hand, the current `build:ci` is a bit strange: contracts are deleted, then the typescript is compiled, then contracts are compiled, and webpack is _not_ run.

For now, perhaps I'll remove `contracts:compile` from `build:ci`

PR #806, opened 2020-01-06
Wallet store cleanup

andrewgordstewart:
- An array of signed states is stored, rather than two arrays of supported and unsupported signed states
- Some getter functions are removed from the store; the channel's store entry can instead expose these getters or alternatives.

### DIFF
--- a/packages/wallet-specs/src/ChannelStoreEntry.ts
+++ b/packages/wallet-specs/src/ChannelStoreEntry.ts
@@ -100,7 +100,7 @@ export class ChannelStoreEntry implements IChannelStoreEntry {
     if (!this.latestSupportedState) {
       return [];
     } else {
-      return this.states.slice(this.states.map(s => s.state).indexOf(this.latestSupportedState));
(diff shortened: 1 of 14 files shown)


packages/wallet-specs/src/ChannelStoreEntry.ts
```diff
@@ -106,9 +96,14 @@ export class ChannelStoreEntry implements IChannelStoreEntry {
     return getChannelId(this.channel);
   }
 
-  get states(): SignedState[] {
-    return this.unsupportedStates.concat(this.supportedState);
+  get unsupportedStates(): SignedState[] {
+    if (!this.latestSupportedState) {
+      return [];
+    } else {
+      return this.states.slice(this.states.map(s => s.state).indexOf(this.latestSupportedState));
```
andrewgordstewart:
```suggestion
      return this.states.slice(this.states.map(s => s.state).indexOf(this.latestSupportedState) + 1);
```

PR #810, opened 2020-01-08
[wallet] Allow UI to properly reflect strategy set in .env

geoknee:


packages/wallet/src/constants.ts
```diff
@@ -22,3 +22,8 @@ export const CHALLENGE_DURATION = 0x12c; // 5 minutes
 
 export const FUNDING_STRATEGY: FundingStrategy =
   process.env.USE_VIRTUAL_FUNDING === "TRUE" ? "VirtualFundingStrategy" : "IndirectFundingStrategy";
+
+export const FUNDING_SOURCE: {[key in FundingStrategy]: string} = {
```
geoknee: Not sure if this should live here...

lalexgap: Could we just export a `FUNDING_STRATEGY_DESCRIPTIVE_NAME` that's either `"a virtual channel"` or a `"reuse-able ledger channel"`?

snario: Probably this should be in the react component where the text is rendered

PR #811, opened 2020-01-08
Implement wallet using xState

lalexgap:
This is the start of creating a wallet that relies on xstate for the protocols.

Currently it allows funding of an RPS game (with depositing mocked out).


packages/xstate-wallet/src/storage/store.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  public getLatestConsensus(channelId: string) {
+    const {supportedState} = this.getEntry(channelId);
+    if (supportedState.length !== 1) {
+      throw new Error('Support contains multiple states');
+    }
+    return supportedState[0];
+  }
+
+  public getLatestSupport(channelId: string) {
```
andrewgordstewart: This stuff can be taken out with the latest wallet-specs `store.ts` code.

packages/xstate-wallet/package.json
```diff
@@ -10,11 +10,13 @@
...
   "dependencies": {
-    "ethers": "4.0.40",
-    "typescript": "3.7.3",
     "@statechannels/client-api-schema": "0.0.1",
     "@statechannels/nitro-protocol": "0.0.1",
-    "@statechannels/wallet-protocols": "1.0.0"
+    "@statechannels/wallet-protocols": "1.0.0",
+    "ethers": "4.0.40",
+    "jsonrpc-lite": "2.1.0",
+    "typescript": "3.7.3",
```
andrewgordstewart: This should be a `devDependency`

PR #812, opened 2020-01-08
[heroku] Configuration for the hub to run on Heroku

snario:
This adds a `Dockerfile` which can be used with the `heroku` CLI.

Here is the workflow...

```bash
# Build the Docker image
docker build . -t statechannels-hub

# Push the image to Heroku
heroku container:push -a statechannels-hub statechannels-hub

# Tell Heroku to update the dyno with this image
heroku container:release -a statechannels-hub statechannels-hub
```


snario:
Don't review until I confirm the deploy preview site is working ...

.circleci/config.yml
```diff
@@ -197,9 +197,9 @@ jobs:
           NODE_ENV: test
     environment:
       USE_VIRTUAL_FUNDING: 'TRUE'
+      HUB_DB_NAME: hub_wallet_development
```
snario: If this is not added, the `hub` will try to use `HUB_DB_NAME` as `root`.

.dockerignore
```diff
@@ -0,0 +1,2 @@
+packages/**/node_modules
+node_modules
```
snario: These are the only two ignored paths because we `COPY` the built JS version of the `hub` inside the docker image.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Run added dependnencies because of configureEnvVariables
+# Estimate 16 seconds
+RUN yarn add dotenv dotenv-expand
+
+# Copy nitro-protocol into dependencies of hub manually
+RUN mkdir ./node_modules/@statechannels && \
+    mv /statechannels/nitro-protocol ./node_modules/@statechannels
+
+# Set entrypoint to basic shell for ease-of-use on Heroku
```
snario: Theoretically it should let us to `heroku run bash` but in practice it hasn't worked for me. It might have something to do with dyno scale limits, as `heroku run` apparently launches a second dyno but we only pay for one. However, I also tried `heroku ps:exec` and it didn't work either.

packages/devtools/src/constants.ts
```diff
@@ -2,12 +2,14 @@ import {Account} from './types';
 
 export const ETHERLIME_ACCOUNTS: Account[] = [
   {
+    // This is used by Mike and Liam both on their local environments
```
snario: Is this still true @kerzhner?

packages/devtools/src/constants.ts
```diff
@@ -29,6 +31,7 @@ export const ETHERLIME_ACCOUNTS: Account[] = [
     privateKey: '0x6ca40ba4cca775643398385022264c0c414da1abd21d08d9e7136796a520a543'
   },
   {
+    // This is used by the hub on Heroku
     privateKey: '0xfac0bc9325ad342033afe956e83f0bf8f1e863c1c3e956bc75d66961fe4cd186'
   }
 ];
```
snario: As a side-note, there is no real reason why we use the key pairs from this file for various environments, the `HUB_ADDRESS` is for messaging and signing only, and doesn't need to have any ETH, I think.

packages/hub/src/constants.ts
```diff
@@ -1,3 +1,5 @@
+require('../env');
```
snario: I noticed this was missing, was that intentional @kerzhner ?

packages/devtools/src/constants.ts
kerzhner: I use the last private key for local development. It is easy for me to switch. Do you think that it's a good idea to leave comments like this in our source code? I wonder if there is a better way to keep track of what private keys folks are using. Probably a better solution is for each developer to generate a private key + address pair locally that is never checked into source control.

packages/hub/src/constants.ts
kerzhner: My personal opinion is that `require('../env')` is not a great way to set environment variables. But I am not sure I have a better solution in mind.

I tried to minimize the use of `require('../env')` to top-level entry points in the code base. `require('../env')` only needs to be called once during the execution of the program.

Dockerfile.hub.staging
```diff
@@ -0,0 +1,38 @@
...
+# TODO: Add package.json first, then do dependencies, then do rest of code
+
+COPY ./packages/hub /statechannels/hub
+COPY ./packages/nitro-protocol /statechannels/nitro-protocol
+
+# Remove the nitro-protocol dependency from @statechannels/hub
+RUN sed -ie "/@statechannels\/nitro-protocol/d" package.json
+
+# Remove the devtools devDependency from @statechannels/hub (avoid resolution)
+RUN sed -ie "/@statechannels\/devtools/d" package.json
```
kerzhner: Mind explaining why this is needed?

packages/devtools/src/constants.ts
snario: I don't think it is a good idea, no. I'll use new, unique, different keys for my local environment, CircleCI, and Heroku.

packages/hub/src/constants.ts
snario: I agree. However I noticed, for some reason, that `firebase-relay.ts` did not have the environment variables set when I ran it despite running from the entrypoint on `server.ts`

Dockerfile.hub.staging
snario: `yarn --production` still tries to resolve the `devDependencies` values, so it tries to find `@statechannels/devtools` on the `npm` registry, but it's not there.

packages/devtools/src/constants.ts
kerzhner: Yep, only the `HUB_SIGNER_ADDRESS` address needs to be an ethereum address with funds.

packages/devtools/src/constants.ts
snario: I updated Heroku's HUB_ADDRESS to 0x37e335623C9994F2473d791d0C3adA14E422022b and updated Netlify too.
Also updated HUB_ADDRESS on config.yml for my liam-heroku-hub branch which changes what CircleCI uses.
These should be unique and won't conflict with anyone's dev environment. (edited)

PR #814, opened 2020-01-08
Fix protocols crashing when there is no supported state

lalexgap:
Some protocols were still relying on `latestSupportState` to return `undefined` instead of throwing an error. I've added a `hasSupportedState` so protocols that aren't sure if a state is supported yet can check that before trying to access it.

### DIFF
--- a/packages/wallet-specs/src/protocols/advance-channel/protocol.ts
+++ b/packages/wallet-specs/src/protocols/advance-channel/protocol.ts
@@ -70,7 +70,7 @@ export const machine: MachineFactory<Init, any> = (store: Store, context?: Init)
       if (!latestEntry.hasSupportedState) {
         return false;
       }
-      return state.turnNum >= targetTurnNum;
+      return latestEntry.latestSupportedState.turnNum >= targetTurnNum;
     },
   };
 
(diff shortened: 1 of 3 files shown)


packages/wallet-specs/src/protocols/advance-channel/protocol.ts
```diff
@@ -66,8 +66,12 @@ export const mockOptions = {
 export const machine: MachineFactory<Init, any> = (store: Store, context?: Init) => {
   const guards: Guards = {
     advanced: ({ channelId, targetTurnNum }: Init, event, { state: s }) => {
+      const latestEntry = store.getEntry(channelId);
+      if (!latestEntry.hasSupportedState) {
+        return false;
+      }
       const { latestSupportedState: state } = store.getEntry(channelId);
```
andrewgordstewart:
```suggestion
```

packages/wallet-specs/src/protocols/advance-channel/protocol.ts
```diff
@@ -66,8 +66,12 @@ export const mockOptions = {
...
-      return !!state && state.turnNum >= targetTurnNum;
+      return state.turnNum >= targetTurnNum;
```
andrewgordstewart:
```suggestion
      return latestEntry.latestSupportedState.turnNum >= targetTurnNum;
```

PR #819, opened 2020-01-09
[rps] Remove unused dependencies

geoknee:
...as flagged by [depcheck](https://www.npmjs.com/package/depcheck).

- Node modules went from 1.26GB + 156MB to ~74MB~1.23GB + 73MB (global + local). ~That seems too good to be true... will investigate.~ (not sure what happened there, it only went down to 74MB in a state where `tsc` would fail...fixing that pushed it back up to 1.23GB)
- this has introduced errors, will need to fix: 
- [x] add openzeppelin contracts back in , and match version to other packages
- [x] fix linting issue in wallet-specs package

Note it flags up a lot of `@types/...` dependencies but I suspect a lot of these will be false positives. So I ignored all of those.


geoknee:
`build` job failing because of `eslint` in  `wallet-protocols`...

geoknee:
aka `wallet-specs` package

geoknee:
This changeset is snowballing so kicking this to a back burner for now.

PR #823, opened 2020-01-10
Re-wire RPS to have an address and outcomeAddress

snario:
Work in progress to fix https://github.com/statechannels/monorepo/issues/779


packages/rps/src/components/LobbyPage.tsx
```diff
@@ -10,7 +10,12 @@ import CreatingOpenGameContainer from '../containers/CreatingOpenGameContainer';
 
 interface Props {
   openGames: OpenGame[];
-  joinOpenGame: (opponentName: string, opponentAddress: string, roundBuyIn: string) => void;
+  joinOpenGame: (
+    opponentName: string,
+    opponentAddress: string,
+    opponentOutcomeAddress: string,
```
andrewgordstewart: is `outcomeAddress` another word for `destination`?

snario: Yes

packages/rps/src/core/channel-state.ts
```diff
@@ -10,6 +10,8 @@ export interface ChannelState<T = AppData> {
   bUserId: string;
   aAddress: string;
   bAddress: string;
+  aOutcomeAddress: string;
+  bOutcomeAddress: string;
```
geoknee: Could  we do better than `address`, `outcomeAddress` -- the former is ambiguous if you don't see it next to the latter.

Here we have https://app-wallet-interface.netlify.com/#introduction `signingAddress` and `destination`, and we should probably enforce consistency with a change here or there or both. The same consistency comment applies to the `nitro-protocol` package and docs website. 

Other possibilities: 
-  `channelAccount` and `chainAccount` 
-  `channelAddress` and `chainAddress`
-  `channelAddress` and `EOA` (externally owned account)
- `ephemeralAddress` and `outcomeAddress`

packages/rps/src/redux/game/actions.ts
```diff
@@ -26,6 +26,7 @@ export interface UpdateProfile {
 export interface GotAddressFromWallet {
   type: 'GotAddressFromWallet';
   address: string;
+  outcomeAddress: string;
```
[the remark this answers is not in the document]
snario: The counter-argument is that this the _chosen_ address, which happens to be from MetaMask, that your RPS game should use for payouts. So, in a way it is an RPS state. Also, there is your opponent's outcome address which is from _their_ MetaMask, which there is no stateful object that can be used to represent.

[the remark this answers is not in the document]
geoknee: Yeah fair point about the opponent's outcome address 👍

packages/rps/src/core/channel-state.ts
snario: I had initially used `participantAddress` and `outcomeAddress` but there was a difficult to find bug from changing `address` to `participantAddress` somewhere in the code. To save time on the weekend, I kept the former as `address`.

geoknee: Cool -- let's just punt this into an issue then, as this PR needn't be held up for this.

snario: https://github.com/statechannels/monorepo/issues/828

PR #825, opened 2020-01-13
[rps] Remove unused dependencies

geoknee:
This may not be comprehensive, but removes a fair bit of unused stuff. 

Seems to save about 30MB at the top level `node_modules` and about a further 30MB at the local `node_modules`.


packages/rps/package.json
```diff
@@ -97,77 +82,56 @@
...
-    "copyfiles": "^2.0.0",
-    "detect-port": "1.3.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
     "eslint": "^6.7.2",
     "eslint-plugin-jest": "^23.1.1",
     "eslint-plugin-prefer-arrow": "^1.1.7",
     "eslint-plugin-prettier": "^3.1.1",
     "eslint-plugin-react": "^7.17.0",
-    "etherlime": "2.2.4",
```
Alonski: Doesn't RPS use etherlime?

packages/rps/package.json
```diff
@@ -97,77 +82,56 @@
...
     "eventemitter3": "^4.0.0",
-    "extract-text-webpack-plugin": "4.0.0-beta.0",
     "file-loader": "^2.0.0",
     "fork-ts-checker-webpack-plugin": "0.4.9",
-    "ganache-cli": "^6.1.8",
     "html-webpack-plugin": "4.0.0-beta.5",
     "jest": "24.9.0",
     "lint-staged": "9.4.0",
-    "node-sass": "4.12.0",
```
Alonski: Does RPS use this to compile sass or something else?

packages/rps/package.json
geoknee: Yes: will reinstall

PR #827, opened 2020-01-13
Use nitro in wallet specs

andrewgordstewart:
This PR removes types defined in `wallet-protocols`, and imports those types from `nitro-protocol` instead.

## TODO
Future work
- [ ] calculate the updated ledger channel outcome -- doing so requires some planning around the management of participants, signing addresses, and destination addresses.
- [ ] write a `supported(state: State, states: SignedState[])` which determines if `state` is supported, given a set of signed states.


packages/wallet-specs/src/ChannelStoreEntry.ts
```diff
@@ -1,5 +1,6 @@
-import { Channel, getChannelId, SignedState, State } from '.';
+import { Channel, getChannelId, SignedState } from '.';
```
lalexgap: Minor but I think it would be nicer to import these from `@statechannels/nitro-protocol`

packages/wallet-specs/src/protocols/create-channel/protocol.ts
```diff
@@ -104,18 +104,18 @@ export const machine: MachineFactory<Init, any> = (store: Store, init: Init) =>
     const channel: Channel = {
       participants,
       channelNonce,
-      chainId: 'mainnet?',
+      chainId: '0x42', // TODO: determine chain Id
```
lalexgap: Should this be a property of the create channel event we send to the protocols?

packages/wallet-specs/src/protocols/virtual-defunding-as-hub/protocol.ts
```diff
@@ -17,23 +13,24 @@ type ChannelsSet = Init & {
   ledgerChannelIds: IDs;
 };
 
-export const assignChannels = assign(
-  ({ jointChannelId }: Init): ChannelsSet => {
-    const { guarantorChannelIds } = checkThat(store.getEntry(jointChannelId).funding, isGuarantee);
+// TODO: Use this in machine factory
```
lalexgap: Worth creating an issue for this?

packages/wallet-specs/src/store.ts
```diff
@@ -225,9 +228,11 @@ export class Store implements IStore {
   // PRIVATE
 
   private signState(state: State): SignedState {
+    const { privateKey } = this.getEntry(getChannelId(state.channel));
```
lalexgap: I think these changes will need to be applied to the `storage/store.ts` in `xstate-wallet` as well?

andrewgordstewart: Can we simplify the codebase for the moment and only define one `store.ts`?

packages/wallet-specs/src/protocols/virtual-defunding-as-hub/protocol.ts
andrewgordstewart: I think this is fine as a TODO: there are so many TODOs in this package that it would pollute the issues.

packages/wallet-specs/src/store.ts
lalexgap: The only real difference between the two `store.ts` is that the `xstate-wallet` has an implementation of `sendMessage` using `postMessage`.

If `wallet-specs` has a `store` that has a configurable messaging we could just use that for now.

andrewgordstewart: Can `xstate-wallet` extend the `Store` class from `wallet-specs`?

PR #832, opened 2020-01-13
Wallet styling tweaks

lalexgap:
Some tweaks to the wallet styling to get it looking like it used to.

The twitter address part of the profile has been removed since we don't need it.

image
funding_1
funding_2


packages/channel-provider/src/ui-service.ts
```diff
@@ -7,20 +7,22 @@ export enum UIElementNames {
   Container = 'channelProviderUiContainer',
   IFrame = 'channelProviderUi'
 }
-
+// TODO: Border radius shouldn't be set here it should be set in the wallet
```
geoknee: Is this in scope for this PR?

PR #839, opened 2020-01-14
[hub] Fix hub:start:dev

snario:
The environment variables were not being loaded for some reason when I ran `yarn hub:start:dev` unless I did this.


kerzhner:
That is curious. [knexfile imports `env` module](https://github.com/statechannels/monorepo/blob/master/packages/hub/src/db-admin/knexfile.ts#L2). Do you have a stale `knexfile.js` lying around by any chance?

snario:
> That is curious. [knexfile imports `env` module](https://github.com/statechannels/monorepo/blob/master/packages/hub/src/db-admin/knexfile.ts#L2). Do you have a stale `knexfile.js` lying around by any chance?

I do not

kerzhner:
Actually, never mind, I don't think that knexfile is relevant in this case.

PR #840, opened 2020-01-14
Monorepo linting

andrewgordstewart:
This PR tidies up some monorepo linting issues, including:
- `--max-warnings=0` is added when running `lint:check` in circle (see [comment](https://github.com/statechannels/monorepo/blob/master/.eslintrc.js#L15-L21))
- built files are ignored by eslint in nitro- and wallet-protocol
- `lint:*` is updated in e2e-tests
  * `web3.min.js` is needed to run puppeteer, and shouldn't be checked

----

For posterity, here is some context around why `web3.min.js` is a part of the repository

> why is a minified js file added to the git repo? is it needed to run the e2e tests on circle?
> Liam Horne:flag-ca: 1:59 PM
> Yes that's right
> Andrew 2:00 PM
> it's to avoid adding web3 as a dependency?
> Liam Horne:flag-ca: 2:02 PM
> I tried symlinking it to node_modules/web3/dist/web3.min.js or something like that, because web3 is in there, but I ran into some issue. I forget what the issue was, though. It was either that symlinking did not work within puppeteer or that the version of web3 in the node_modules wasn't the exact one we needed.
> 
> 
> The problem is that we need to mock window.ethereum in puppeteer and this is the only way we know how:
> await page.evaluateOnNewDocument(`window.web3 = new Web3("http://localhost:8547")`);
> await page.evaluateOnNewDocument(`window.ethereum = window.web3.currentProvider`);
> (edited)
> Maybe new ethers.providers.JsonRpcProvider would work instead, I don't know, haven't tried
> Andrew 2:03 PM
> but the point is that you need to supply an actual file containing web3. It's not enough to allow node to resolve the dependency
> right?
> Liam Horne:flag-ca: 2:23 PM
> We need to inject some code that contains Web3 into the frame that puppeteer creates
> I tried fixing it at one point: https://github.com/statechannels/monorepo/pull/760


package.json
```diff
@@ -43,7 +43,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lerna run --parallel --stream precommit --since HEAD"
+      "pre-commit": "lerna run --parallel --stream precommit --since HEAD --max-warnings=0"
```
andrewgordstewart:
```suggestion
      "pre-commit": "lerna run --parallel --stream precommit --since HEAD -- --max-warnings=0"
```

packages/channel-client/.eslintrc.js
```diff
@@ -1,10 +1,13 @@
...
 module.exports = {
+  ...baseConfig,
   env: {
     browser: true,
     es6: true
   },
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint', 'prettier', 'jest', 'import'],
+  plugins: [...baseConfig.plugins, '@typescript-eslint', 'prettier', 'jest', 'import'],
   extends: [
```
snario: You can actually add `../../.eslintrc` to this `extends` field instead of manually extending `module.exports` with `...baseConfig`, I learned.

PR #844, opened 2020-01-15
Remove unused dependencies

kerzhner:
- Remove `truffle`, `babel-register`, and `babel-runtime`. `truffle` is the second largest package in the monorepo.
- `firebase` package is also quite large, and most of the firebase functionality is unused. Firebase functionalities can be imported via firebase sub-packages. For the hub, needed sub-packages are imported. For rps, I tried the same approach but ran into typescript errors with `redux-saga-firebase`.


geoknee:
By my reckoning, this reduces the size of the `monorepo` (after build) from 1.82GB to 1.64GB, so roughly a 10% reduction 🌋

PR #846, opened 2020-01-15
[hub] Use TypeScript references correctly in hub

snario:
`hub` was not using [build mode](https://www.typescriptlang.org/docs/handbook/project-references.html#build-mode-for-typescript), now it is. Also added the `nitro-protocol` reference.

### DIFF
--- a/packages/hub/package.json
+++ b/packages/hub/package.json
@@ -21,7 +21,7 @@
     "hub:test": "npx jest --runInBand -c ./jest/jest.config.js",
     "test:ci": "run-s chain:test hub:test",
     "start:shared-ganache": "NODE_ENV=development npx start-shared-ganache",
-    "build": "npx tsc",
+    "build": "tsc -b .",
     "build:typescript": "tsc -b .",
     "prestart": "yarn build && npm run db:rollback && npm run db:migrate && npm run db:seed",
     "lint:check": "eslint \"*/**/*.{js,ts,tsx}\" --cache",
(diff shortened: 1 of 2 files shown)


packages/hub/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "esModuleInterop": true,
     "strict": false
   },
-  "references": [{"path": "../devtools"}],
+  "references": [{"path": "../devtools"}, {"path": "../nitro-protocol"}],
```
kerzhner: Do we need to enable `composite` setting in `nitro-protocol` and `devtools` for references to work properly? https://www.typescriptlang.org/docs/handbook/project-references.html#composite

snario: Yeah, and we do. It's in the root `tsconfig.json`.

packages/hub/package.json
```diff
@@ -21,7 +21,7 @@
     "hub:test": "npx jest --runInBand -c ./jest/jest.config.js",
     "test:ci": "run-s chain:test hub:test",
     "start:shared-ganache": "NODE_ENV=development npx start-shared-ganache",
-    "build": "npx tsc",
+    "build": "tsc -b .",
```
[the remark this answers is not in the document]
snario: The script ["can also be any locally installed executable that is inside `node_modules/.bin/`"](https://yarnpkg.com/lang/en/docs/cli/run/#toc-yarn-run-script)

packages/hub/tsconfig.json
kerzhner: Ok yep, that makes sense.

PR #847, opened 2020-01-15
[rps] Overwrite CHAIN_NETWORK_ID in production

geoknee:
This variable is needed for the UI to pickup the target network id. Previously, this was not being overwritten and so took on some default value for `ganache`.

There might be a more elegant fix that removes the redundancy with `TARGET_NETWORK` (which has the network namesting), but at least this change should un-break the deployed app.


snario:
I just set this on Netlify now as well.

PR #852, opened 2020-01-15
Updates to xstate wallet

lalexgap:
Various updates to get `xstate-wallet` working with the latest changes to `wallet-specs` to use nitro-protocol.

Currently allows for the fake funding of an rps game.


packages/wallet-specs/src/temp-store.ts
```diff
@@ -0,0 +1,5 @@
+// TODO: This is temporary until all protocols have been updated
```
andrewgordstewart: :+1:

packages/xstate-wallet/src/utils/json-rpc-utils.ts
```diff
@@ -1,35 +1,41 @@
...
+}
 
 export function createJsonRpcAllocationsFromOutcome(outcome: Outcome): Allocations {
   return outcome.map(o => {
     if (!isAllocationOutcome(o)) {
       throw new Error('Attempted to convert non allocation outcome to an allocation');
     }
     return {
-      token: o.assetHolderAddress,
+      token: '0x0',
```
andrewgordstewart: Use `AddressZero`?

lalexgap: Right now the API specifies that `token:0x0` is what the API will return for ETH so I'm going to leave as is for now.

PR #853, opened 2020-01-15
Add e2e test for withdrawal

snario:


package.json
```diff
@@ -8,6 +8,7 @@
     "node": "^10.16.3"
   },
   "scripts": {
+    "abc": "tsc",
```
lalexgap: Is this needed?

snario: Oh, woops, that was to show something to @kerzhner. Deleting...

snario:
```suggestion
```

PR #855, opened 2020-01-15
Basic deposit and getHolding methods for protocols

lalexgap:
Adds simple `Deposit` and `getHolding` functions that can be used by protocols.

I've verified that both work locally.

### DIFF
--- a/packages/xstate-wallet/package.json
+++ b/packages/xstate-wallet/package.json
@@ -13,7 +13,7 @@
     "@statechannels/client-api-schema": "0.0.1",
     "@statechannels/nitro-protocol": "0.0.1",
     "@statechannels/wallet-protocols": "1.0.0",
-    "ethers": "4.0.40",
+    "ethers": "4.0.39",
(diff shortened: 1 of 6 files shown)


packages/xstate-wallet/package.json
```diff
@@ -13,7 +13,7 @@
     "@statechannels/client-api-schema": "0.0.1",
     "@statechannels/nitro-protocol": "0.0.1",
     "@statechannels/wallet-protocols": "1.0.0",
-    "ethers": "4.0.40",
+    "ethers": "4.0.39",
```
snario: Why downgrade?

lalexgap: To keep it consistent with `nitro-protocol` and `wallet`.  I was getting issues with `BigNumber` from `ethers` not matching between `nitro-protocol` and `xstate-wallet`.

lalexgap:
Closing as going to work on this further with the state machine.

PR #856, opened 2020-01-15
Build and deploy netlify sites from circle

kerzhner:
Adding a release job to circle that builds + deploys static sites to netlify. For now, only `nitro-protocol` docs, `wallet`, and `rps` are deployed. Once this PR is merged in, continuous integration will be turned off in the netlify console. The reasons to migrate static site builds and deploys to Circle are:
- Netlify CI is expensive. Our monthly budget is around $200 (which is more than CircleCI).
- Static site build and deploy instructions are defined in code (config.yml) as opposed to in Netlify console.
- Circle is faster since `yarn install` is only run once for all site builds + deploys. On Netlify, `yarn install` is run once per each static site build + deploy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,9 @@
     "lint:write": "lerna run lint:write --parallel --no-bail",
     "publish": "lerna publish --yes from-package patch",
     "postinstall": "patch-package",
-    "build:netlify": "lerna run --stream --parallel --since $(git merge-base $(git rev-parse --abbrev-ref HEAD) origin/master) build:netlify",
-    "release:netlify": "lerna run --stream --parallel --since $(git merge-base $(git rev-parse --abbrev-ref HEAD) origin/master) release:netlify -- --auth $NETLIFY_ACCESS_TOKEN --message $(git rev-parse --short HEAD) $([[ $(git rev-parse --abbrev-ref HEAD) == master ]] && echo --prod)"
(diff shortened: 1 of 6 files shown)


kerzhner:
Once we are happy with this PR, we can raise a new PR for migrating Netlify static site CI for packages other than `nitro-protocol`, `wallet`, and `rps`.

package.json
```diff
@@ -22,7 +22,9 @@
     "lint:check": "lerna run lint:check --parallel --no-bail -- --max-warnings=0",
     "lint:write": "lerna run lint:write --parallel --no-bail",
     "publish": "lerna publish --yes from-package patch",
-    "postinstall": "patch-package"
+    "postinstall": "patch-package",
+    "build:netlify": "lerna run --stream --parallel build:netlify",
+    "release:netlify": "lerna run --stream --parallel release:netlify -- --auth $NETLIFY_ACCESS_TOKEN --message $(git rev-parse --short HEAD) $(case $(git rev-parse --abbrev-ref HEAD) in (master) echo --prod;; esac)"
```
geoknee: Am I right in thinking this command will always run? Could we use the `--since` flag for `lerna` (as in L20) to skip the netlify release when that package didn't change?

kerzhner: Yep you are right about that, good catch. We can probably skip `build:netlify` for unchanged packages as well.

PR #857, opened 2020-01-16
[wallet] Replace steps with a progress bar

geoknee:
progressbar

### DIFF
--- a/packages/wallet/package.json
+++ b/packages/wallet/package.json
@@ -79,6 +79,7 @@
     "redux-storage-decorator-filter": "1.1.8",
     "redux-storage-engine-indexed-db": "1.0.0",
     "resolve": "1.6.0",
+    "rimble-ui": "^0.11.1",
     "web3-utils": "1.2.2",
     "webpack": "4.29.5",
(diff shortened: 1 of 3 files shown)


geoknee:
Will wait for feedback on this idea before polishing up

- [ ] colors
- [ ] renaming

snario:
To be honest I kind of like the step-by-step UI as it lets you inspect what's happening a bit more easily and get a sense for the process being undergone.

geoknee:
The rationale is this: in the happy case, many screens / steps flick by very quickly making it pretty well impossible to track what is going on. When something is broken, or you are waiting on something, IMHO it seems best to focus on a particular message describing what is happening now, and not all the things that got successfully done or are yet to be attempted.

geoknee:
Current UI for reference: 
old
(It doesn't seem quite right anyways. I guess the whole UI needs an overhaul)

snario:
I can understand that rationale, makes sense

geoknee:
Closing since I put this idea into the x-state wallet instead #938 .

PR #861, opened 2020-01-17
Protocols connect to an IStore, not a Store

andrewgordstewart:
The `IStore` interface should specify all the methods are needed by the protocols. The `Store` implementation is still useful for trying things out, and contains some useful implementation details.


lalexgap:
Closing this in favour of #901. It was easier to redo the change then to resolve the merge conflicts.

PR #863, opened 2020-01-17
[ttt] [base !=master] Tweaks to test script

geoknee:
I think there is enough room to get the binary rep of the Xs / Os field into the table. This is nice because you don't have to scroll or do transcoding in your head to inspect these fields (e.g. and tell quickly if they are disjoint).

If I am wrong and the table is now too long, reverting to 1204dce would still be an improvement IMHO.


Alonski:
Hmm I'll take a look at this. Although I am pretty sure it won't all fit.

PR #864, opened 2020-01-17
[rps] Allow multiple serial games

### LINKED ISSUE #778: [wallet/rps] Cannot play a second game
After completing a game and concluding/withdrawing funds I'm unable to start a second game. It seems to get stuck on "Waiting for opponent to confirm".

Not sure if this issue is caused by `rps` or the `wallet` but I suspect it's something in the `wallet.

geoknee:
Three fixes: 

- Because most of the `rps` sagas `take` every action, there were hacks in place to prevent performing some tasks multiple times: in this case a mutex to track if Player A had joined a game. This mutex was never getting reset. The new approach is to do away with the mutex, and only join a game when a "join game" action was dispatched. 

- Some tasks (e.g. opening a channel) were only executed when the `channelState` was "empty". The new approach is to clear out the `channelState` when the game is over. 

Closes #778 .

(and, unrelated to this issue)

- In some scenarios the metamask address never gets injected, leading to runtime errors down the line. The new approach is to wait for the metamask address to exist before progressing with the game.

### DIFF
--- a/packages/e2e-tests/puppeteer/helpers.ts
+++ b/packages/e2e-tests/puppeteer/helpers.ts
@@ -33,9 +33,10 @@ export async function loadRPSApp(page: Page, ganacheAccountIndex: number): Promi
 }
 //
 
-export async function setUpBrowser(headless: boolean): Promise<Browser> {
+export async function setUpBrowser(headless: boolean, slowMo?: number): Promise<Browser> {
   const browser = await launch({
     headless,
+    slowMo,
     devtools: !headless,
     // Needed to allow both windows to execute JS at the same time
     ignoreDefaultArgs: [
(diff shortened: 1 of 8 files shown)


geoknee:
This PR also includes a considerable refactor of the e2e test.

packages/e2e-tests/puppeteer/helpers.ts
```diff
@@ -31,11 +31,18 @@ export async function loadRPSApp(page: Page, ganacheAccountIndex: number): Promi
     }
   });
 }
-//
 
-export async function setUpBrowser(headless: boolean): Promise<Browser> {
+export async function waitForAndClickButton(page: Page | Frame, button: string): Promise<void> {
```
tomclose: 👍

PR #867, opened 2020-01-17
CI: always build, test and release on master

kerzhner:
Fixes https://github.com/statechannels/monorepo/issues/860.

The goal is:
- for the `master` branch, run build, test, and release jobs on every CI run.
- for other branches, only run build, test, and release jobs for packages that have diverged since head of `master`.


package.json
```diff
@@ -10,22 +10,25 @@
   "scripts": {
     "prepare": "typesync && lerna run --concurrency 8 --stream prepare",
     "build": "lerna run build",
-    "build:ci": "lerna run --since $(git merge-base $CIRCLE_BRANCH origin/master) build:ci",
+    "build:ci:incremental": "yarn build:ci --since $(git merge-base $CIRCLE_BRANCH origin/master)",
+    "build:ci": "lerna run build:ci",
```
geoknee: What's the reason for using `yarn` on the incremental build but `lerna` on the master build?

package.json
```diff
@@ -10,22 +10,25 @@
...
     "run:w3t": "lerna run start --parallel --scope=**/web3torrent --scope=**/embedded-wallet",
     "run:ewt": "lerna run start --parallel --scope=**/embedded-wallet",
     "clean": "git clean -Xdf --exclude=\"!.env\"",
     "clean:dry": "git clean -Xdn --exclude=\"!.env\"",
     "start:shared-ganache": "cd packages/devtools && NODE_ENV=development yarn start:shared-ganache",
     "test": "lerna run --stream --parallel test",
-    "test:ci": "lerna run --stream --parallel test:ci --include-dependents --since $(git merge-base $(git rev-parse --abbrev-ref HEAD) origin/master) build:ci",
+    "test:ci:incremental": "yarn test:ci --since $(git merge-base $CIRCLE_BRANCH origin/master)",
+    "test:ci": "lerna run  test:ci --stream --parallel --include-dependents",
```
geoknee: Similar question about why we don't have the `--stream --parallel --include-dependents` flags on the incremental command?

package.json
kerzhner: `yarn` runs `build:ci` defined in the same `package.json`. In this case, `build:ci:incremental` will trigger `build:ci` aka `lerna run build:ci` with an additional `--since` flag.

package.json
kerzhner: The incremental command triggers the non-incremental command.

geoknee: Gotcha. Clearly wasn't looking carefully enough 😊

PR #870, opened 2020-01-20
[monorepo] Set concurrency for lerna to 4

geoknee:
We get an out of memory error on circle under current config. 

On this branch I will try a few different configs that avoid the error while keeping the duration as short as possble.

Tests in the `wallet` package take the most time (typically 6mins).

For all tests:

Concurrency = 2: 9mins
Concurrency = 4: 6 mins


geoknee:
Recommend we go for `--concurrency 4` as this seems close to optimal in the sense of not taking longer than the `wallet` tests do on their own.

PR #874, opened 2020-01-21
[rps, wallet] Correctly implement challenging e2e

snario:
Some notes...

  - [x] Added a `challengeChannel` method on `ChannelClient`.
  - [x] Documented method on Client API docs
  - [x] Added method to `client-api-schema`
  - [x] Added handler to `wallet`'s `message-handler`
  - [x] Added test case for `message-handler`'s tests
  - [x] Added e2e integration test for challenge workflow
  - [x] Added UI on RPS for when you're challenging
  - [x] Added UI on RPS for when you're challenging
  - [x] Added `challenging` and `responding` status to `ChannelResult` in `client-api-schema`
  - [x] Fixed off by one error in `NitroAdjudicator`'s check from PostFund to first state of a game
  - [x] Fixed "Invalid Signatures" error on `forceMove` (commented out sig verification atm)
  - [x] Added adjudicator watcher handling of `ChallengeCleared` event which checks transaction arguments of a `respond` call

### DIFF
--- a/packages/app-wallet-interface/source/index.html.md
+++ b/packages/app-wallet-interface/source/index.html.md
@@ -410,6 +410,62 @@ Possible response to a `Channel Proposed` event.
 | ---- | ----------------- | ---------------------------------------------------------------- |
 |      | Channel not found | The wallet can't find the channel corresponding to the channelId |
 
+## Challenge Channel
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "ChallengeChannel",
+  "id": 0,
+  "params": {
+    "channelId": "0xabc123"
+  }
+}
+```
+
+> Example response
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "channelId": "0xabc123...",
+    "status": "challenging",
+    "funding": [{"token": "0x0", "amount": "24"}],
+    "participants": [
+      {
+        "participantId": "user123",
+        "signingAddress": "0x...",
+        "destination": "0xa..."
+      },
+      {
+        "participantId": "user456",
+        "signingAddress": "0x...",
+        "destination": "0xb..."
+      }
+    ],
+    "turnNum": 7,
+    "allocations": [
+      {
+        "token": "0x...", // 0x0 for ETH
+        "allocationItems": [
+          {"destination": "0xa...", "amount": "18"},
+          {"destination": "0xb...", "amount": "6"}
+        ]
+      }
+    ],
+    "appData": "0x...."
+  }
+}
+```
+
+### Errors
+
+| Code | Message            | Meaning                                                                               |
+| ---- | ------------------ | ------------------------------------------------------------------------------------- |
+|      | Channel not found  | The wallet can't find the channel corresponding to the channelId                      |
+
 # Events
 
 Sent from the wallet to the app.
--- a/packages/channel-client/src/types.ts
+++ b/packages/channel-client/src/types.ts
@@ -53,6 +53,7 @@ export interface ChannelResult {
   status: ChannelStatus;
   // funding: Funds[]; // do we even need this?
   turnNum: string;
+  challengeExpirationTime?: number;
 }
 
 export type UnsubscribeFunction = () => void;
--- a/packages/channel-client/src/types.ts
+++ b/packages/channel-client/src/types.ts
@@ -51,7 +51,6 @@ export interface ChannelResult {
   appDefinition: string;
   channelId: string;
   status: ChannelStatus;
-  // funding: Funds[]; // do we even need this?
   turnNum: string;
   challengeExpirationTime?: number;
 }
--- a/packages/channel-client/src/types.ts
+++ b/packages/channel-client/src/types.ts
@@ -4,7 +4,17 @@ import {
   JsonRpcNotification
 } from '@statechannels/channel-provider';
 
-export type ChannelStatus = 'proposed' | 'opening' | 'funding' | 'running' | 'closing' | 'closed';
+// TODO: Several of these types are duplicates of those in @statechannels/client-api-schema
+
+export type ChannelStatus =
+  | 'proposed'
+  | 'opening'
+  | 'funding'
+  | 'running'
+  | 'challenging'
+  | 'responding'
+  | 'closing'
+  | 'closed';
 
 export interface Participant {
   participantId: string; // App allocated id, used for relaying messages to the participant
@@ -76,6 +86,7 @@ export interface ChannelClientInterface<Payload = object> {
     allocations: Allocation[],
     appData: string
   ) => Promise<ChannelResult>;
+  challengeChannel: (channelId: string) => Promise<ChannelResult>;
   closeChannel: (channelId: string) => Promise<ChannelResult>;
   getAddress: () => Promise<string>;
 }
--- a/packages/devtools/src/ganache/server.ts
+++ b/packages/devtools/src/ganache/server.ts
@@ -40,6 +40,7 @@ export class GanacheServer {
     const cmd = `ganache-cli ${opts}`;
 
     this.server = spawn('npx', ['-c', cmd], {stdio: 'pipe'});
+    this.server.stdout.on('data', (data: unknown) => console.log(`${data}`));
     this.server.stderr.on('data', (data: unknown) => {
       log.error(`Server threw error ${data}`);
       throw new Error('Ganache server failed to start');
--- a/packages/devtools/src/ganache/server.ts
+++ b/packages/devtools/src/ganache/server.ts
@@ -40,7 +40,6 @@ export class GanacheServer {
     const cmd = `ganache-cli ${opts}`;
 
     this.server = spawn('npx', ['-c', cmd], {stdio: 'pipe'});
-    this.server.stdout.on('data', (data: unknown) => console.log(`${data}`));
     this.server.stderr.on('data', (data: unknown) => {
       log.error(`Server threw error ${data}`);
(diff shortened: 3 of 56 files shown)


packages/app-wallet-interface/source/index.html.md
```diff
@@ -410,6 +410,62 @@ Possible response to a `Channel Proposed` event.
...
+    "channelId": "0xabc123"
+  }
+}
+```
+
+> Example response
+
+```json
+{
+  "jsonrpc": "2.0",
```
lalexgap: Are we returning the `challengeTimeout`? We should let the app know when the challenge will expire.

packages/devtools/src/ganache/server.ts
```diff
@@ -40,6 +40,7 @@ export class GanacheServer {
     const cmd = `ganache-cli ${opts}`;
 
     this.server = spawn('npx', ['-c', cmd], {stdio: 'pipe'});
+    this.server.stdout.on('data', (data: unknown) => console.log(`${data}`));
```
lalexgap: This probably shouldn't be included.

packages/app-wallet-interface/source/index.html.md
snario: Just added that.

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -333,7 +333,8 @@ contract ForceMove is IForceMove {
         for (uint256 i = 0; i < nParticipants; i++) {
             address signer = _recoverSigner(stateHashes[whoSignedWhat[i]], sigs[i]);
             if (signer != participants[i]) {
-                return false;
+                // TODO: Figure out why turnNum 3->4 RPS challenge hits this error case
+                continue; // return false;
```
snario: Flagging this to be fixed before merge.

packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
```diff
@@ -1,17 +1,18 @@
...
+import {Signature, Interface} from 'ethers/utils';
 
 // TODO: Currently we are setting some arbitrary gas limit
 // To avoid issues with Ganache sendTransaction and parsing BN.js
 // If we don't set a gas limit some transactions will fail
 const GAS_LIMIT = 3000000;
 
-const NitroAdjudicatorContractInterface = new utils.Interface(NitroAdjudicatorArtifact.abi);
+// @ts-ignore
+const NitroAdjudicatorContractInterface = new Interface(NitroAdjudicatorArtifact.abi);
```
snario: For some reason this is throwing a build error, not sure why.

packages/rps/contracts/RockPaperScissors.sol
```diff
@@ -116,7 +116,9 @@ contract RockPaperScissors is ForceMoveApp {
         private
         pure
         outcomeUnchanged(fromPart, toPart)
-        stakeUnchanged(fromGameData, toGameData)
+        // TODO: question (by @liam to @george) this is wrong right? the stake _will_ change
+        // because we are proposing a new round, so this line of code below should not be here?
+        // stakeUnchanged(fromGameData, toGameData)
```
snario: Question for @geoknee

packages/channel-client/src/types.ts
```diff
@@ -43,6 +53,7 @@ export interface ChannelResult {
   status: ChannelStatus;
   // funding: Funds[]; // do we even need this?
```
lalexgap: The api specifies that we return the amount funded to the channel which I think we're not doing right now. I think it's debatable if we need that even.

packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
lalexgap: Are the contract artifacts built?

packages/rps/contracts/RockPaperScissors.sol
geoknee: I think the stake could be allowed to change, so agree that arguably this modifier does not belong here. 

I would not go so far as to say the stake _will_ change. In our current app implementation, for example, it certainly will _not_ change. 

In fact it seems reasonable (if unnecessary) to bake this into the rules by leaving this modifier in place.

packages/nitro-protocol/src/contract/challenge.ts
```diff
@@ -1,8 +1,10 @@
+import NitroAdjudicatorArtifact from '../../build/contracts/NitroAdjudicator.json';
 import {utils} from 'ethers';
 import {decodeOutcome} from './outcome';
 import {FixedPart, hashState, State, VariablePart} from './state';
 import {Address, Bytes32, Uint256, Uint8} from './types';
 import {Channel, SignedState} from '..';
+import {Transaction, Interface, bigNumberify} from 'ethers/utils';
```
geoknee: Importing `ethers/utils` in two different ways here.

packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
geoknee: I see no build error. I do see linting errors

packages/wallet/src/redux/protocols/application/reducer.ts
```diff
@@ -166,6 +166,7 @@ function handleDisputeAction(
     newDisputeState.protocolState.type === "Challenging.Failure" ||
     newDisputeState.protocolState.type === "Responding.Success"
   ) {
+    // delete protocolState.disputeState; // TODO: @alex does this look right
```
geoknee: @lalexgap

geoknee:
This could probably be made a little more attractive: 

Screenshot 2020-01-22 at 11 13 03

packages/rps/src/redux/game/saga.ts
```diff
@@ -39,7 +39,14 @@ const isPlayersTurnNext = (
...
   const channel = yield actionChannel('*', buffers.fixed(10));
   while (true) {
-    yield take(channel);
+    const action = yield take(channel);
+
+    // TODO: @george @tom is this the right place for this?
+    if (action.type === 'Challenge') {
+      const {channelState} = yield select(getGameState);
+      yield challengeChannel(channelState, client);
+    }
```
geoknee: I think this is the right place, as all calls to the client live in this saga.

However, I am wondering if the call needs to provide the `channelState`, or even any data at all? If I am challenging, it is because it is not my turn: therefore my wallet should have all the information it needs (i.e. all the states) to launch the challenge because I have previously called `updateChannel` with them.

packages/nitro-protocol/contracts/ForceMove.sol
geoknee: Looking at the event emitted during the `forceMove`, somehow the `sigs` array is e.g.  `[bSig,bSig]` instead of `[aSig,bSig]`. So that is probably why the contract (rightly) reverts the transaction. Currently trying to track down where this error is introduced...

geoknee: Scratch that -- on further investigation the signatures seem to be taken out of the channel store and put into the transaction correctly. I did find a bug in the code that parses the `ChallengeRegistered` event, however, explaining what I saw there ^ #882

packages/rps/src/redux/game/saga.ts
geoknee: To clarify: the API call only supplies the `channelId`, which seems correct.

But the `challengeChannel` generator function also puts an `'UpdateChannelState'` action, and I am wondering if we really need that?

packages/e2e-tests/puppeteer/__tests__/rps.test.ts
```diff
@@ -42,7 +56,23 @@ describe('Playing a game of RPS', () => {
     expect(await rpsTabB.waitForXPath('//button[contains(., "Create a game")]')).toBeDefined();
     expect(await rpsTabA.waitForXPath('//button[contains(., "Create a game")]')).toBeDefined();
 
-    await clickThroughRPSUI(rpsTabA, rpsTabB);
+    await startAndFundRPSGame(rpsTabA, rpsTabB);
```
geoknee: This is arguably redundant with L72 : during the whole test suite we actually start 3 games now.  You could consider having test1: play a game, and close it out; test2: start a new game, raise and respond to a challenge?

packages/channel-client/src/types.ts
snario: @tomclose what do you think?

packages/rps/src/redux/game/saga.ts
snario: It actually only needs the `channelId` from that value

packages/channel-client/src/types.ts
tomclose: I think it might be useful to have this in the future, but I'm not averse to removing it until we get to that point.

packages/e2e-tests/puppeteer/__tests__/rps.test.ts
```diff
@@ -17,12 +25,15 @@ describe('Playing a game of RPS', () => {
   beforeAll(async () => {
     browserA = await setUpBrowser(HEADLESS, 10); // 10ms delay seems to prevent certain errors
     browserB = await setUpBrowser(HEADLESS, 10); // 10ms delay seems to prevent certain errors
+  });
 
+  beforeEach(async () => {
```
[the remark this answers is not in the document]
snario: Reloads the page so it re-runs the basic setup

PR #878, opened 2020-01-21
Add (passing) wallet test

andrewgordstewart:


packages/wallet-specs/package.json
```diff
@@ -8,9 +8,10 @@
   "license": "MIT",
   "scripts": {
     "prepare": "yarn build",
-    "build": "npx tsc && yarn generateConfigs",
+    "build": "npx tsc -b",
```
lalexgap: Is there a reason we don't want to generate configs on build? Speed?

andrewgordstewart: 👍 They aren't checked into the repository right now.

packages/wallet-specs/src/protocols/depositing/protocol.ts
```diff
@@ -110,32 +86,34 @@ type Services = {
...
-  funded: (context: Init) => boolean;
 };
 type Options = { services: Services; guards: Guards };
 export const machine: MachineFactory<Init, any> = (store: IStore, context: Init) => {
   const subscribeDepositEvent = (context: Init) => callback => {
-    return store.on('DEPOSITED', event => {
+    return store.on('DEPOSITED', (event: ChainEvent) => {
       if (event.type === 'DEPOSITED' && event.channelId === context.channelId) {
-        callback('DEPOSITED');
+        if (bigNumberify(event.total).gte(context.fundedAt)) {
```
lalexgap: I know we were talking before about making our protocols less reliant on events from the blockchain and instead relying on the state. We could this here by just dispatching one event (`CHAIN_UPDATED`)  but that makes the state machine more complex. Any thoughts?

andrewgordstewart: We could definitely do that. Even if we do, it could be nice for this callback to send `'FUNDED` or `'DEPOSITED` to the machine.

PR #879, opened 2020-01-21
Some wallet-protocols cleanup

andrewgordstewart:


packages/wallet-specs/.eslintrc.js
```diff
@@ -21,5 +21,7 @@ module.exports = {
     'no-debugger': 'off',
     'no-useless-catch': 'off',
     'no-empty-pattern': 'off',
+    'import/first': 2,
+    'import/order': [2, { 'newlines-between': 'always' }],
```
andrewgordstewart: I think this rule is beneficial as long as it violations are fixed **automatically on save**. It means smaller diffs and more organized imports.

PR #880, opened 2020-01-22
Application workflow

lalexgap:
This implements a basic "application" workflow in `xstate-wallet` that allows two players to fund and play RPS.

A visualization of the current workflow is here: https://xstate.js.org/viz/?gist=b602e01aa81d14cba8c7d62761a83119


packages/xstate-wallet/src/workflow-manager.ts
```diff
@@ -0,0 +1,53 @@
...
+      | OpenChannelEvent
+      | SendStates
+      | ChannelUpdated
+      | ApplicationWorkflowEvent
+  ) {
+    if (event.type && (event.type === 'CREATE_CHANNEL' || event.type === 'OPEN_CHANNEL')) {
+      const machine = interpret<any, any, any>(applicationWorkflow(this.store), {
+        devTools: true
+      })
+        .onTransition(state => {
```
andrewgordstewart: Should there be an `onDone` callback to remove oneself from `this.workflows`?

packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -0,0 +1,94 @@
...
+    id: 'openMachine',
+    src: 'invokeOpeningMachine',
+    data: (context, event) => event,
+    onDone: 'running',
+    autoForward: true
+  },
+  on: {SendStates: {actions: ['updateStore', 'sendChannelUpdated']}}
+};
+const running = {
+  on: {
```
andrewgordstewart: Are there times when one of these  events would arrive before entering the running state? How should they be handled then?

packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -0,0 +1,94 @@
...
+    PLAYER_STATE_UPDATE: {target: 'running', actions: ['sendToOpponent']},
+    SendStates: {target: 'running', actions: ['updateStore', 'sendChannelUpdated']}
+  }
```
andrewgordstewart: It might be worth writing out the logic of when you move to `'closing'` etc, with a mocked `CloseChannel` service. That would let us test starting a second game.

packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -0,0 +1,94 @@
...
+      return CreateChannel.machine(store);
+    } else {
+      return JoinChannel.machine(store, event);
+    }
+  };
+  const updateStore = (context, event: SendStates) => {
+    store.receiveStates(
+      // TODO: The outcome can get removed when going over the wire if it's empty
+      // For now we just add it back here
+      event.signedStates.map(ss => ({state: {outcome: [], ...ss.state}, signatures: ss.signatures}))
```
andrewgordstewart: Might be worth writing a `rehydrate(state)` helper that gets called within the walls of firebase-related code.

We could use a json-schema to validate an incoming `SignedState` object.

packages/xstate-wallet/src/workflow-manager.ts
lalexgap: Yup, I haven't done that yet since there is no way of finishing a workflow at the moment.

lalexgap: Going to handle this in a separate PR for https://github.com/statechannels/monorepo/issues/899

packages/xstate-wallet/src/workflows/application.ts
lalexgap: Going to handle this in a separate PR for https://github.com/statechannels/monorepo/issues/899

packages/xstate-wallet/src/workflows/application.ts
lalexgap: The slightly tricky bit is that the `data` payload in which the messages are embedded are not specified in the JSON schema. The idea being that clients shouldn't need to know about what the shape of the wire-protocol the protocols implement is.

Perhaps there should be a `validate`/`hydrate` in `wire-protocol.ts`?

packages/xstate-wallet/src/workflows/application.ts
lalexgap: This shouldn't ever happen but we could dump into an error state.

PR #883, opened 2020-01-22
Implement Conclude protocol

andrewgordstewart:


packages/wallet-specs/src/protocols/direct-funding/protocol.ts
```diff
@@ -104,18 +81,28 @@ type Services = {
...
+        .reduce(add, '0');
+      const holdings = await store.getHoldings(ctx.channelId);
 
-    if (gt(allocated, holdings)) {
-      throw new Error('Channel underfunded');
+      if (gt(allocated, holdings)) {
+        throw new Error('Channel underfunded');
+      }
+    } catch (e) {
+      if (/No supported state found/.test(e.message)) {
```
lalexgap: This seems a little strange

packages/wallet-specs/src/protocols/virtual-fund-as-leaf/protocol.ts
```diff
@@ -80,17 +79,17 @@ const createGuarantorChannel = {
...
-  ledgerId,
-  balances,
-}: ChannelsKnown): SupportStateArgs {
-  const amount = total(balances);
-  return {
-    channelId: ledgerId,
-    outcome: ethAllocationOutcome([{ destination: getChannelId(guarantorChannel), amount }]),
-  };
-}
+// function fundGuarantorArgs({
```
lalexgap: Can this go?

packages/wallet-specs/src/store.ts
```diff
@@ -53,6 +54,7 @@ export type Constructor = Partial<{
 }>;
 export class Store implements IStore {
   public static equals(left: any, right: any) {
+    // TODO: Delete this; we should use statesEqual and outcomesEqual
```
lalexgap: I've run into issues here where `outcome:[]` gets stripped out when being transported over firebase

packages/wallet-specs/src/protocols/virtual-fund-as-leaf/protocol.ts
andrewgordstewart: Why not.

It was commented out because
- it returns the wrong type, since `SupportState.Init` has changed
- it isn't used, but it probably will be soon.

PR #884, opened 2020-01-23
[base: ags/conclude] Clean up depositing machine

andrewgordstewart:
See commit messages for visualizations of the updates.


andrewgordstewart:
I was unable to actually get spawning to work properly with context. I had [similar problems](https://gitter.im/statecharts/statecharts?at=5ded84be5ac7f22fb585c96e) trying to get the documentation code to work.

PR #886, opened 2020-01-23
Add script to upload configuration as new gist

andrewgordstewart:


packages/wallet-specs/README.md
```diff
@@ -18,7 +18,8 @@ Run `yarn`
 After updating `protocol.ts`, run `yarn generateConfigs` to update `protocol.config.js`.
 Configs (`protocol.config.json`) can be visualized using https://xstate.js.org/viz/.
 
-(An upcoming update to the visualizer should allow CLI generation of the configurations!)
+Install [gist](https://github.com/defunkt/gist) and run `yarn uploadConfig $PROTOCOL_NAME` to save the current configuration file living in `src/protocols/$PROTOCOL_NAME/protocol.config.js`.
+Make sure to run `yarn generateConfigs` prior to uploading.
```
tomclose: It seems like this is a common error to make. Could you include config generation for `$PROTOCOL_NAME` as part of the upload task?

packages/wallet-specs/bin/uploadConfig.sh
```diff
@@ -0,0 +1,16 @@
...
+# > yarn uploadConfig $PROTOCOL
+#
+# Example
+# ❯ yarn uploadConfig direct-funding
+# yarn run v1.21.1
+# $ bin/uploadConfig.sh direct-funding
+# http://xstate.js.org/viz/?gist=7b60e4d910abec0a7b5368ac11f54ace
+
+PROTOCOL=$1
+FILENAME=src/protocols/${PROTOCOL}/protocol.config.js
```
lalexgap: Is there a way we could generalize this so we could use it in `xstate-wallet` as well for workflows?

Currently the `xstate-wallet` as a `workflow` folder that contains a `.ts` file per a workflow (IE: `Workflows/application.ts`) but I'm open to reorganizing that so we can reuse this.

Maybe the easiest way for now is just to create a `uploadConfig.sh` for `xstate-wallet` and we can pull out shared logic later to `devtools`

andrewgordstewart: we could just accept the filename as the 1st argument. Then:
- in `wallet-protocols`, run `bin/uploadConfig.sh src/protocols/${YARNS_ARG}/protocol.config.js` for the `uploadConfig` script

PR #887, opened 2020-01-23
Add env variable to control whether e2e tests run headless

tomclose:


packages/devtools/src/config/env.ts
```diff
@@ -1,6 +1,25 @@
...
 import * as path from 'path';
 
+export function getEnvBool(name: string, throwIfMissing = true): boolean {
+  const val = process.env[name];
+
+  if (throwIfMissing && val === undefined) {
+    throw Error(`Environment variable ${name} is not set.`);
+  }
+
+  switch (process.env[name]) {
```
geoknee: You could use `val` here?

PR #888, opened 2020-01-23
[nitro] Fix and test forceMove transaction generator

geoknee:
It was previously using a broken `createSignatureArguments` helper -- this is now altered to rip the signatures out of an array of `SignedStates`, and re-order them by participant (as is required by the contract code). 

This PR also increases test coverage such that tests would fail without the above fix (namely, when the array of states submitted does not begin with one owned by participant 0). 

Some scope for DRYing out the new tests.

### DIFF
--- a/packages/nitro-protocol/src/transactions.ts
+++ b/packages/nitro-protocol/src/transactions.ts
@@ -81,10 +81,13 @@ function createSignatureArguments(
   const {participants} = signedStates[0].state.channel;
 
   // Get a list of all unique states.
-  const states = signedStates.filter((s, i, a) => a.indexOf(s) === i).map(s => s.state);
-  const signatures = signedStates.map(s => s.signature);
+  const uniqueSignedStates = signedStates.filter((s, i, a) => a.indexOf(s) === i);
+  const states = uniqueSignedStates.map(s => s.state);
+
   // Generate whoSignedWhat based on the original list of states (which may contain the same state signed by multiple participants)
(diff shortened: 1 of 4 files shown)


packages/nitro-protocol/src/transactions.ts
```diff
@@ -81,10 +81,13 @@ function createSignatureArguments(
   const {participants} = signedStates[0].state.channel;
 
   // Get a list of all unique states.
-  const states = signedStates.filter((s, i, a) => a.indexOf(s) === i).map(s => s.state);
-  const signatures = signedStates.map(s => s.signature);
```
geoknee: This was the offending line.

packages/nitro-protocol/src/transactions.ts
```diff
@@ -81,10 +81,13 @@ function createSignatureArguments(
...
+  const uniqueSignedStates = signedStates.filter((s, i, a) => a.indexOf(s) === i);
+  const states = uniqueSignedStates.map(s => s.state);
+
   // Generate whoSignedWhat based on the original list of states (which may contain the same state signed by multiple participants)
   const whoSignedWhat = signedStates.map(s => participants.indexOf(getStateSignerAddress(s)));
+  const signatures = [];
+  participants.forEach((p, i) => signatures.push(uniqueSignedStates[whoSignedWhat[i]].signature));
```
geoknee: ...and this is the fix.

Issue #843, opened 2020-01-15
integration-test-with-hub flickers


andrewgordstewart:
[This job](https://app.circleci.com/jobs/github/statechannels/monorepo/8344) failed, and re-running it [passed](https://app.circleci.com/jobs/github/statechannels/monorepo/8371)

snario:
It _might_ be possible that if two `integration-test-with-hub` jobs are running simultaneously then one or both might fail as messages are shared in the same firebase database. This might not be the cause of the flickering, but it's one hypothesis to test.

andrewgordstewart:
<img width="1422" alt="Screen Shot 2020-01-15 at 8 57 25 AM" src="https://user-images.githubusercontent.com/8432675/72448959-233b1f80-3775-11ea-9023-dd8c6f3f9e30.png">

Good catch!

PR #890, opened 2020-01-23
Prefix all firebase messages with something unique to environment

snario:
This should theoretically prevent multiple environments from accidentally conflicting with one another (see #843)


packages/rps/src/redux/open-games/saga.ts
```diff
@@ -119,7 +119,7 @@ const openGameTransformer = dict => {
 function* openGameSyncer() {
   yield fork(
     reduxSagaFirebase.database.sync,
-    'challenges',
+    `${FIREBASE_PREFIX}/challenges`,
```
andrewgordstewart: Since it seemed easy to miss some of these, does it make sense to centralize the logic that constructs these paths?

packages/e2e-tests/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Container package for end to end tests, from contracts to browser apps",
   "main": "index.js",
   "scripts": {
-    "test": "wait-on http://localhost:3000 http://localhost:3055 && jest",
+    "test": "wait-on http://localhost:3000 http://localhost:3055 && jest --runInBand",
```
andrewgordstewart: Why's this?

snario: Because the individual tests themselves could conflict (there are 3) if run in parallel, I think

PR #907, opened 2020-01-24
Add a ChannelUpdated event and clean up Store

lalexgap:
Adds a channel updated event and accepts a messaging service as constructor argument. With these changes `xstate-wallet` should be able to run directly against the base `Store` by just configuring it with the correct messagingService,chain and registering for the channel updated listener.


packages/wallet-specs/src/store.ts
```diff
@@ -243,7 +267,15 @@ export class Store implements IStore {
   protected updateEntry(channelId: string, states: SignedState[]): ChannelStoreEntry {
     const entry = this.getEntry(channelId);
     this._store[channelId] = { ...entry, states: merge(states, entry.states) };
-
+    // TODO: Does this comparison work as expected?
```
lalexgap:
```suggestion
```

packages/wallet-specs/src/store.ts
```diff
@@ -61,24 +74,35 @@ export class Store implements IStore {
...
+      .initialize();
   }
 
   public async getHoldings(channelId: string) {
     return await this._chain.getHoldings(channelId);
   }
-  public on(chainEventType, listener) {
-    return this._chain.on(chainEventType, listener);
+  public on(eventType: StoreEventType, listener: StoreEventListener) {
+    if (eventType === 'CHANNEL_UPDATED') {
```
andrewgordstewart: Could we `switch` on `eventType`, and return `unreachable(eventType)` as the default?

PR #911, opened 2020-01-25
[RPS] Add Sentry.io Error Tracking and Monitoring

Alonski:
This PR sets up Sentry for RPS.

I created an Org called state-channels.

In Sentry settings I set it up so that it only logs errors from this url:
`https://rps.statechannels.org`

### DIFF
--- a/packages/rps/src/index.tsx
+++ b/packages/rps/src/index.tsx
@@ -1,12 +1,15 @@
 import * as React from 'react';
 import {render} from 'react-dom';
 import {Provider} from 'react-redux';
+import * as Sentry from '@sentry/browser';
 
 // Not adding in currently because it breaks most of our existing components
 // import 'bootstrap/dist/css/bootstrap.min.css';
 import './index.scss';
 import store from './redux/store';
 import SiteContainer from './containers/SiteContainer';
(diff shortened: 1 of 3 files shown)


packages/rps/src/index.tsx
```diff
@@ -1,13 +1,16 @@
...
 import {Provider} from 'react-redux';
+import * as Sentry from '@sentry/browser';
 
 // Not adding in currently because it breaks most of our existing components
 // import 'bootstrap/dist/css/bootstrap.min.css';
 import './index.scss';
 import store from './redux/store';
 import SiteContainer from './containers/SiteContainer';
 
+Sentry.init({dsn: 'https://db92051c360145b4b3537d33881c1bc3@sentry.io/1913622'});
```
snario: Can you add some information in the README on how this was attained? Also, perhaps add some info in our internal Notion on how we can access the admin dashboard etc (+ ensuring we have relevant invites).

Alonski: On how the DSN was obtained? I just followed the guides :)
Can you DM where exactly to add the info to the Notion? I haven't used it too much.

PR #912, opened 2020-01-25
[monorepo] Remove fmg-core and web3 dependencies

snario:
I tried upgrading to Node 12.7.0 and hit an issue related [to this Github issue](https://github.com/ethereum/web3.js/issues/2913). I don't think we even need `scrypt` since we don't need `web3`... so removing these.


snario:
This doesn't make Node 12 work by the way. We still need LimeChain to make [this small change](https://github.com/LimeChain/ethereum-transaction-debugger/issues).

PR #917, opened 2020-01-26
Get closing working in the xstate wallet

### LINKED ISSUE #893: Add Concluding to Application workflow
Currently the application workflow does not handle concluding at all. It should be expanded to handle concluding a channel by starting the appropriate concluding protocols and waiting for them to complete.

lalexgap:
Adds support for closing a channel to the application workflow. I've been able to close an RPS game using the application workflow but there are also a lot of times something goes wrong.

Fixes #893

 I've created #919 to follow up but I think this can be merged in as is.

Visualization of the workflow is available here: https://xstate.js.org/viz/?gist=4c2813fb86c609d9d189b33d7296272f


packages/xstate-wallet/src/workflows/tests/application.test.ts
```diff
@@ -0,0 +1,108 @@
...
+          channel: {chainId: '0x0', channelNonce: '0x0', participants: []}
+        },
+        signatures: []
+      }
+    ]
+  };
+  store.receiveStates = jest.fn();
+  const channelId = getChannelId(sendStates.signedStates[0].state.channel);
+  const service = interpret<any, any, any>(applicationWorkflow(store, {channelId}));
+  service.start('running');
```
lalexgap: I really like that we can just start at a certain state!

PR #920, opened 2020-01-27
Improve reliability of rps challenging scenario

snario:
I had been trying to make it the case that when you finish a challenge, the state can be updated despite the wallet process being in an `Application.WaitForDispute` state. This ended up causing race conditions not visible when testing against a fast local ganache network, but noticeable on Ropsten.

This PR gets rid of that feature, so if you want to continue a channel after a challenge, you need to click "Ok" on both prompts.

### DIFF
--- a/packages/wallet/src/redux/sagas/messaging/message-handler.ts
+++ b/packages/wallet/src/redux/sagas/messaging/message-handler.ts
@@ -282,7 +282,7 @@ function* handleUpdateChannelMessage(payload: RequestObject) {
     if (
       protocolState &&
       protocolState.type === "Application.WaitForDispute" &&
-      typeof protocolState.disputeState! !== "undefined"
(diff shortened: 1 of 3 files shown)


packages/wallet/src/redux/sagas/messaging/message-handler.ts
```diff
@@ -282,7 +282,7 @@ function* handleUpdateChannelMessage(payload: RequestObject) {
     if (
       protocolState &&
       protocolState.type === "Application.WaitForDispute" &&
-      typeof protocolState.disputeState! !== "undefined"
+      typeof protocolState.disputeState !== "undefined"
```
geoknee:
```suggestion
      protocolState.disputeState
```
I am aware this is not exactly equivalent code, but am wondering if it is actually as good or better for our needs here?

PR #931, opened 2020-01-28
Change the application workflow to support typesafe actions

tomclose:
Before this PR it was possible to call actions with the wrong type of event, which would cause runtime errors. By creating the machine using `Machine(config).withConfig(options, context)`, we weren't able to make use of xstate's action type-safety:
```ts
type Event1 = {type: 'FOO'; bar: string};
type Event2 = {type: 'FOO2'; bar: number};
type Event = Event1 | Event2;
const bazMachine = Machine<any, any, Event>({
  initial: 'start',
  states: {
    start: {
      on: {
        FOO: {
          actions: [
            (c, e) => e.bar, // here it knows that e of type Event1
            'badlyTypedActon'
          ]
        }
      }
    }
  }
}).withConfig({
  actions: {
    // but here the only check it does is that Event2 extends Event
    // it's lost the context of where badlyTypedAction is being called
    badlyTypedAction: (c, e: Event2) => e.bar ** 2
  }
});
```

This PR implements an approach to solving this problem:

1. There's a new type `Actions` which details the expected types:
    ```ts
    interface Actions {
      sendToOpponent: Action<Context, PlayerStateUpdate>;
      updateStore: Action<Context, SendStates>;
      hideUi: Action<Context, any>;
      displayUi: Action<Context, any>;
      assignChannelId: Action<Context, any>;
      sendChannelUpdatedNotification: Action<Context, ChannelUpdated>;
    }
    ```
2. There's a `generateConfig(actions: Actions)` function which generates a configuration.
3. Within this config, actions are referred to using the actions object `actions.hideUi`, instead of via strings

This gives type-safety, while still allowing the configuration to be defined separately from the action implementations.

### DIFF
--- a/packages/xstate-wallet/src/workflows/application.ts
+++ b/packages/xstate-wallet/src/workflows/application.ts
@@ -1,4 +1,4 @@
-import {MachineConfig, Machine, StateMachine, assign} from 'xstate';
+import {MachineConfig, Machine, StateMachine, assign, Action} from 'xstate';
 import {
   FINAL,
   MachineFactory,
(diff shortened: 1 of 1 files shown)


packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -222,13 +231,15 @@ const mockServices = {
...
+const mockActions: Actions = {
+  sendToOpponent: 'sendToOpponent',
+  updateStore: 'updateStore',
+  sendChannelUpdatedNotification: 'sendChannelUpdatedNotification',
+  hideUi: 'hideUi',
+  displayUi: 'displayUi',
+  assignChannelId: 'assignChannelId'
 };
+
+export const config = generateConfig(mockActions);
```
geoknee: Why do we need this? It has the same name as another `const` above (not exported), and does not appear to be imported anywhere.
```suggestion
```

tomclose: We need this so that `yarn generateConfigs` will work. It isn't ideal, but that's how it works for now.

PR #933, opened 2020-01-28
[monorepo] Run yarn install immediately after typesync

geoknee:
Running typesync _only_ alters `package.json`, adding values such as `@types/foo` if it finds `foo` as a dependency. As such, any nontrivial change (that does not have `yarn` run afterwards) is just going to cause everyone who bases work from that commit to regenerate `yarn.lock`.

This fix isn't particularly nice, so open to alternative suggestions (such as not having tyepsync be  part of `prepare` at all...).

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "node": "^10.16.3"
   },
   "scripts": {
-    "prepare": "typesync && yarn install --ignore-scripts && lerna run --concurrency 8 --stream prepare",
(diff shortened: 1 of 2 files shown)


geoknee:
I had that thought: but we will then be trying to run `typesync` before it has been installed. If we use `npx` , `typesync` will be downloaded and installed before being run. It takes about 6s to install on my machine. That might improve if it gets cached. See 2854010.

PR #934, opened 2020-01-28
[wallet-protocols] Clean up some configuration options

snario:


packages/wallet-protocols/package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "license": "MIT",
   "scripts": {
-    "prepare": "yarn build",
+    "prepare": "echo 'No prepare step for wallet-protocols'",
```
andrewgordstewart:
```suggestion
```

PR #935, opened 2020-01-29
Remove machine factory from index

andrewgordstewart:


packages/wallet-protocols/src/protocols/create-null-channel/protocol.ts
```diff
@@ -1,10 +1,12 @@
 import { assign, Machine } from 'xstate';
 import { HashZero, AddressZero } from 'ethers/constants';
+import { Outcome } from '@statechannels/nitro-protocol';
```
geoknee: This is an unused import: do we want  to add the `"noUnusedLocals": true,` option to `tsconfig.ts` for this package? Most of the other packages have it.

andrewgordstewart: We definitely want to add it!

PR #936, opened 2020-01-29
Virtual funding progress

andrewgordstewart:
This PR implements some of the required services for the virtual-funding protocols. The `fundLedger` step is not implemented, since it requires some small changes to the ledger-funding protocol.

I'm requesting to merge this now to avoid any more painful merge conflicts.


packages/wallet-protocols/src/protocols/virtual-fund-as-leaf/protocol.ts
```diff
@@ -1,131 +1,167 @@
...
+            { destination: targetChannelId, amount },
+            { destination: makeDestination(hubAddress), amount },
+          ],
+          store.ethAssetHolderAddress
+        ),
+      },
+    };
+  };
+}
+const fundTarget = getDataAndInvoke('fundTargetArgs', 'supportState', 'success');
```
tomclose: This isn't really a comment about this specific PR, but more about x-state, or at least the way we're using it. This line makes me feel really uneasy. As someone who's reading the code, essentially there are some magic strings being passed to a function. There doesn't seem to be an easy way for me to trace these back to where they're defined, or inspect their types, or understand what's going to be passed to them. I get that connecting services to machine config via strings allows for the implementation to be decoupled, but it seems like we have so much decoupling that it becomes harder to figure out what the actual code does.

It would be good to discuss this further, and figure out if there are patterns we can use to make our x-state code easier to read and understand.

PR #937, opened 2020-01-29
Hub: add Sentry alerts

kerzhner:


Alonski:
@kerzhner Did you set it up to only track errors thrown in production? I did this for RPS in Sentry settings by limiting the url that is tracked. Not sure if this can be done here as well.

kerzhner:
On the hub, we rely on `NODE_ENV` environment variable to distinguish between runtimes. At the moment, Slack notifications are triggered for the `staging` environment. We don't have a production environment at the moment.

One issue with the above: a developer can run the hub locally with `NODE_ENV=staging` and trigger Sentry alerts.

Alonski:
@kerzhner Yes that is exactly what I meant. We don't really want errors to be thrown locally. Could we introduce another env var that will only be used on servers to turn on Sentry?

kerzhner:
@Alonski, I introduced a new env var `RUNTIME_ENV`. @snario also enlightened me that [it's a bad idea](https://glebbahmutov.com/blog/do-not-use-node-env-for-staging/) to use anything besides `development` or `production` as `NODE_ENV`. So we need a new env var to distinguish between execution contexts anyways.

Alonski:
@kerzhner Ok great. So are we now filtering on Sentry based on Env? If its `development` it doesn't catch errors?

kerzhner:
@Alonski, Slack alerts are sent out when Sentry environment is `staging` (since we don't have a `production` environment at the moment). Sentry captures errors in all Sentry environments, but Sentry filters based on Sentry environment variable for alert triggering.

PR #938, opened 2020-01-29
[xstate wallet] Make UI prettier and extract data from invoked machines

### LINKED ISSUE #921: Add UI to create/ join channel process
We need a UI that provides basic information to the user when join/creating a channel.

We should base the UI on the current workflow state and possibly protocol state as well.

geoknee:
This PR contains 

- some groundwork to allow css in storybook and webpack
- proof-of-concept work that starts to extract information from machines and their children and display in a reasonably attractive way

Future work: add some buttons, and hook them up to `useService.send` where appropriate. One idea would be to get the same buttons as the current redux wallet so we can run the current e2e tests. 
Screenshot 2020-01-29 at 14 54 59
Screenshot 2020-01-29 at 14 55 43

Closes #921


geoknee:
Closes #921

PR #940, opened 2020-01-29
Update Fake Channel Provider

lalexgap:
Updates the Fake Channel Provider to handle new budget related calls, based on https://www.notion.so/Web3Torrent-Requirements-ce435f483e044d8291a8a01f570391d4#d815cee96f954fc6aaf04258992dd37e.

The Fake channel client has been removed since a regular `ChannelClient` can be started with a `FakeChannelProvider`

The JSON schema has been updated to include the `Budget` related calls.

I've also updated the `channel-provider` to use types from `client-api-schema` so return types should be more accurate and helpful.


packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -1,17 +1,18 @@
-import {ChannelProviderInterface} from '@statechannels/channel-provider/src';
+import {ChannelProviderInterface, MethodType} from '@statechannels/channel-provider/src';
```
tomclose: Might be worth moving this file into the `channel-provider` package.

packages/channel-client/src/channel-client.ts
```diff
@@ -5,10 +5,14 @@ import {
...
-  Allocation,
-  PushMessageResult
+  SiteBudget
 } from './types';
+import {
+  PushMessageResult,
+  Allocation,
+  Participant,
+  BudgetResult
+} from '@statechannels/client-api-schema/types';
```
tomclose: I think you could probably just import these from `@statechannels/client-api-schema`?

packages/channel-client/tests/fakes/fake-channel-provider.ts
lalexgap: I think this makes sense but it's annoying/non-trivial enough to do that I'm going to punt and create a separate issue for it (#961).

PR #946, opened 2020-01-30
[monorepo] Run yarn upgrade

snario:
Upgrades most dependencies.


snario:
Tests failing on `eslint` running out of memory.............

packages/e2e-tests/puppeteer/scripts/rps.ts
```diff
@@ -135,3 +135,25 @@ export async function bResigns(rpsTabA: Page, rpsTabB: Page): Promise<boolean> {
...
+
+    const rpsTabA = (await browserA.pages())[0];
+    const rpsTabB = (await browserB.pages())[0];
+
+    await loadRPSApp(rpsTabA, 0);
+    await loadRPSApp(rpsTabB, 1);
+
+    await login(rpsTabA, rpsTabB);
+  }
+})();
```
geoknee: Perhaps this belongs in a separate PR?

snario: #950

packages/nitro-protocol/src/contract/transaction-creators/eth-asset-holder.ts
```diff
@@ -10,7 +10,8 @@ import * as assetHolderTransactionCreator from './asset-holder';
 // If we don't set a gas limit some transactions will fail
 const GAS_LIMIT = 3000000;
 
-const EthAssetHolderContractInterface = new utils.Interface(EthAssetHolderArtifact.abi);
+// @ts-ignore
```
lalexgap: Can we add the github issue here as well?

PR #949, opened 2020-01-31
Use existing ledger channel

andrewgordstewart:
This PR
- Extracts the finding of a suitable ledger channel out from the `ledger-funding` protocol. This means that `ledger-funding` is fully-determined
- Uses the `findExistingLedgerId` method as a service in the `find-or-create-ledger-channel`


andrewgordstewart:
Closing as this will likely be unused.

PR #950, opened 2020-01-31
[rps] Add helpful script for rps.ts

snario:


packages/e2e-tests/puppeteer/scripts/rps.ts
```diff
@@ -135,3 +135,25 @@ export async function bResigns(rpsTabA: Page, rpsTabB: Page): Promise<boolean> {
   await Promise.all([playerA(rpsTabA), playerB(rpsTabB)]);
   return true;
 }
+
+/**
+ * Useful for local testing. Run with:
+ *
+ * yarn ts-node -O '{"module":"commonjs","noUnusedLocals":false}' ./puppeteer/scripts/rps.ts
```
geoknee: Let's have an alias for this in `package.json`?

packages/e2e-tests/puppeteer/scripts/rps.ts
```diff
@@ -135,3 +135,25 @@ export async function bResigns(rpsTabA: Page, rpsTabB: Page): Promise<boolean> {
...
+ *
+ * It opens two windows ready for you to manually use.
+ */
+(async (): Promise<void> => {
+  if (require.main === module) {
+    const browserA = await setUpBrowser(false, 100);
+    const browserB = await setUpBrowser(false, 100);
```
geoknee: You *might* get away without the `slowMo=100` here, because no messages are being sent yet and I think that's what causes the issues with the e2e test.

PR #964, opened 2020-02-04
Confirm-Create-Channel workflow

lalexgap:
Creates the workflow that handles confirming creating and joining a channel based on https://www.notion.so/0ee12a01ea684be095ca3b38909c9399?v=282f6c575af544e88924e7a1864a7a57&p=6ba5a9b3d85d467da9bb160986f8478d.

Visualizer here: https://xstate.js.org/viz/?gist=197d65fbc5f4dfa35865da66e8a156de

It seems that the actual `confirm-create-channel` workflow actually doesn't do much and can be re-used between either joining or creating a channel.

Marking as draft so we can discuss exactly what will be handled by the workflow versus handled outside of workflows.

I also experimented with some stricter typing.

Fixes #964

### DIFF
--- a/packages/xstate-wallet/src/workflows/confirm-create-channel.ts
+++ b/packages/xstate-wallet/src/workflows/confirm-create-channel.ts
@@ -1,0 +1,94 @@
+import {
+  MachineConfig,
+  Action,
+  DefaultGuardType,
+  GuardPredicate,
+  ConditionPredicate,
+  StateSchema
+} from 'xstate';
+import {Allocations, Participant} from '@statechannels/client-api-schema';
+
+interface WorkflowActions {
+  hideUi: Action<WorkflowContext, any>;
+  displayUi: Action<WorkflowContext, any>;
+}
+interface WorkflowGuards {
+  noBudget:
+    | GuardPredicate<WorkflowContext, WorkflowEvent>
+    | ConditionPredicate<WorkflowContext, WorkflowEvent>;
+}
+
+// While this context info may not be used by the workflow
+// it may be used when displaying a UI
+interface WorkflowContext {
+  participants: Participant[];
+  allocations: Allocations;
+  appDefinition: string;
+  appData: string;
+  chainId: string;
+  challengeDuration: number;
+}
+
+interface WorkflowStateSchema extends StateSchema<WorkflowContext> {
+  states: {
+    needUserConfirmation: {};
+    waitForUserConfirmation: {};
+    // TODO: Is it possible to type these as type:'final' ?
+    done: {};
+    failure: {};
+  };
+}
+interface UserApproves {
+  type: 'USER_APPROVES';
+}
+interface UserRejects {
+  type: 'USER_REJECTS';
+}
+type WorkflowEvent = UserApproves | UserRejects;
+const generateConfig = (
+  actions: WorkflowActions,
+  guards: WorkflowGuards
+): MachineConfig<WorkflowContext, WorkflowStateSchema, WorkflowEvent> => ({
+  initial: 'needUserConfirmation',
+  states: {
+    needUserConfirmation: {
+      on: {
+        '': [
+          {
+            target: 'waitForUserConfirmation',
+            cond: guards.noBudget,
+            actions: [actions.displayUi]
+          },
+          {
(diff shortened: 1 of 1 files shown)


packages/xstate-wallet/src/workflows/confirm-create-channel.ts
```diff
@@ -0,0 +1,118 @@
...
+  actions: WorkflowActions,
+  guards: WorkflowGuards
+): MachineConfig<WorkflowContext, WorkflowStateSchema, WorkflowEvent> => ({
+  initial: 'needUserConfirmation',
+  states: {
+    needUserConfirmation: {
+      on: {
+        '': [
+          {
+            target: 'waitForUserConfirmation',
```
lalexgap: Unfortunately `target` is just typed as string

packages/xstate-wallet/src/workflows/confirm-create-channel.ts
```diff
@@ -0,0 +1,118 @@
...
+interface UserRejects {
+  type: 'USER_REJECTS';
+}
+type WorkflowEvent = UserApproves | UserRejects;
+
+const generateConfig = (
+  actions: WorkflowActions,
+  guards: WorkflowGuards
+): MachineConfig<WorkflowContext, WorkflowStateSchema, WorkflowEvent> => ({
+  initial: 'needUserConfirmation',
```
lalexgap: `initial` gets strongly typed as `"waitForUserConfirmation"|"needUserConfirmation"|"done"|"failure"`

packages/xstate-wallet/src/workflows/confirm-create-channel.ts
```diff
@@ -0,0 +1,118 @@
...
+  MachineConfig,
+  Action,
+  DefaultGuardType,
+  GuardPredicate,
+  ConditionPredicate,
+  StateSchema,
+  Machine
+} from 'xstate';
+import {Allocations, Participant} from '@statechannels/client-api-schema';
+import {MachineFactory, Store} from '@statechannels/wallet-protocols/src';
```
snario: Avoid direct file imports

packages/xstate-wallet/src/workflows/confirm-create-channel.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+// While this context info may not be used by the workflow
+// it may be used when displaying a UI
+interface WorkflowContext {
+  participants: Participant[];
+  allocations: Allocations;
+  appDefinition: string;
+  appData: string;
+  chainId: string;
+  challengeDuration: number;
```
tomclose: Probably need to add the `requestId` here, to allow the JSON-RPC interface to include it in the result.

packages/xstate-wallet/src/workflows/confirm-create-channel.ts
```diff
@@ -0,0 +1,118 @@
...
+      sendDisplayMessage('Show');
+    },
+    hideUi: (context, event) => {
+      sendDisplayMessage('Hide');
+    }
+  };
+
+  const options = {guards, actions};
+  const config = generateConfig(actions, guards);
+  return Machine(config).withConfig(options, context);
```
tomclose: I don't think you need to pass in `options` here, given that you passed the `actions` and `guards` to `generateConfig`?

packages/xstate-wallet/src/workflows/confirm-create-channel.ts
lalexgap: Will the `requestId` live that long? When a user calls `create-channel` I'm assuming we will return success immediately instead of blocking until a user clicks something (the current `channel-client` will timeout and retry if we don't respond fairly quickly)

 Further updates will be provided to the client over json-rpc notifications.

tomclose: As it currently works, that call returns a `channel-result`, which I think would mean that we need to hang onto the `requestId` until we have the channel created. We could change that though.

lalexgap: Ah right. We probably don't want to to block the json-rpc call so we should probably change that.

PR #965, opened 2020-02-04
[wallet-protocols] Add sort-imports rules to wallet-protocols

snario:
I find this sorting style to look clean and be easy to understand what a file depends on relative to external modules and local files more easily than the previous unsorted way.

### DIFF
--- a/packages/wallet-protocols/.eslintrc.js
+++ b/packages/wallet-protocols/.eslintrc.js
@@ -23,7 +23,7 @@ module.exports = {
     'no-empty-pattern': 'off',
     'import/first': 2,
     "import/order": ["error", {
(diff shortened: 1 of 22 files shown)


packages/wallet-protocols/.eslintrc.js
```diff
@@ -22,7 +22,10 @@ module.exports = {
     'no-useless-catch': 'off',
     'no-empty-pattern': 'off',
     'import/first': 2,
-    'import/order': [2, { 'newlines-between': 'always' }],
+    "import/order": ["error", {
+      "groups": ["builtin", "external", "internal",  "parent", , "sibling", "index"],
```
geoknee:
```suggestion
      "groups": ["builtin", "external", "internal",  "parent", "sibling", "index"],
```

PR #967, opened 2020-02-04
[e2e] Add debugging code

geoknee:
- Refactors test to not use `Promise.race`
- on failures within `waitForAndClickButton`
    - saves a screenshot (gitignored, and uploaded on circle as test artifact)
    - throws an error with more context about the frame and selector


geoknee:
In commit 6f12074, circle was configured to run the integration test repeatedly. It passed 9 times and failed on the 10th: https://app.circleci.com/jobs/github/statechannels/monorepo/10974 . One of the new errors was thrown: 

```frame.click(#join) failed on frame Rock Paper Scissors```

This error is thrown when #join is found by `page.waitOnSelector` but then not found by `page.click`. 

By clicking on the artifacts tab, we can see the screenshot saved just afterwards:

integration-test-click error

Player A is in the lobby and there is no open game to click "join" on.

It seems that either

i) the game appeared and then disappeared before it could be clicked (currently `slowMo=100ms`)
ii) the game never appeared but puppeteer thought that it did.

PR #971, opened 2020-02-04
Remove json-rpc.ts file in wallet-protocols

snario:
Uses `client-api-schema`.

Also the term `Allocation` was overloaded in `client-api-schema/definition.json` and `Outcome.sol`. It meant both `AllocationItem[]` and `{ token: string, allocationItems: AllocationItem[] }`. Renamed the one that `nitro-protocol` relies on to `NitroAllocation`.


packages/nitro-protocol/src/contract/outcome.ts
```diff
@@ -29,26 +30,25 @@ export function decodeGuarantee(encodedGuarantee: Bytes): Guarantee {
...
-export type Allocation = AllocationItem[];
-export interface AllocationItem {
-  destination: Bytes32;
-  amount: Uint256;
-}
 
-export function encodeAllocation(allocation: Allocation): Bytes32 {
+// Note that a NitroAllocation is as defined in Outcome.sol, but an Allocation, used in
+// other parts of the codebase refers to what is used in the client-api-schema JSON-RPC
+export type NitroAllocation = AllocationItem[];
```
lalexgap: I'm not a huge fan of having a specially named `Allocation` since all our contracts and documentation refers to `Allocation`. However we can always rename it back so it's not a big deal.

snario: Yeah I think we need to decide on which is the true `Allocation`. Maybe the JSON-RPC parameter ought to be instead `AssetAllocations`?

packages/nitro-protocol/package.json
```diff
@@ -10,6 +10,7 @@
     "ethers": "4.0.44"
   },
   "devDependencies": {
+    "@statechannels/client-api-schema": "0.0.1",
```
geoknee: It doesn't seem right to me that `nitro-protocol` has our client-api-schema as a dependency (and why as a devDependency anyway?)

packages/nitro-protocol/src/index.ts
```diff
@@ -34,11 +34,10 @@ export {Channel, getChannelId} from './contract/channel';
 export {encodeConsensusData, decodeConsensusData, ConsensusData} from './contract/consensus-data';
 export {validTransition, ForceMoveAppContractInterface} from './contract/force-move-app';
 export {
-  AllocationItem,
   encodeAllocation,
   encodeOutcome,
   Outcome,
-  Allocation,
+  NitroAllocation as Allocation,
```
geoknee: I think contortions such as these are probably best avoided in this package which is at the very base of our stack.

packages/nitro-protocol/package.json
snario: It is in `outcome.ts` for the `AllocationItem` definition. Maybe it shouldn't need this? `devDependency` because it is just a type and so isn't needed in a production build.

packages/nitro-protocol/src/index.ts
snario: Ah I didn't notice this, will fix.

packages/nitro-protocol/package.json
geoknee: Good point on the `devDependency` thing 👍 .

andrewgordstewart: Doesn't tree shaking take care of that for us?

packages/hub/src/wallet/db/migrations/20191111110900_allocations.ts
```diff
@@ -1,7 +1,7 @@
 import * as Knex from 'knex';
 import {addBytesCheck} from '../utils';
 
-const TABLE_NAME = 'allocations';
```
kerzhner: @snario in theory, an existing migration should never change. If an existing migration changes, every database has to be wiped and migrated from empty state.

PR #973, opened 2020-02-05
[hub, rps] Only initialize Sentry in production environment

geoknee:
This prevents everyone from being alerted (emailed) when some developer gets an error locally.

### DIFF
--- a/packages/hub/src/hub/server.ts
+++ b/packages/hub/src/hub/server.ts
@@ -1,9 +1,11 @@
 import * as Sentry from '@sentry/node';
 
-Sentry.init({
-  dsn: 'https://5b818f025d1a4259a8cf086377b67025@sentry.io/2047255',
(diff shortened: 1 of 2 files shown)


Alonski:
It's also possible to filter error tracking by domains. This is what I do for RPS.

kerzhner:
@geoknee it looks like Sentry supports sending notifications only for specified environments. The danger of this PR: if the instantiation of Sentry has a runtime error, we would not notice the error until deployment to staging or production environments. To make the matters worse, Sentry would not report the error as Sentry would not be initialized properly.

geoknee:
I think that's a fair point @kerzhner. Happy to fix it so that errors in *all* environments go to sentry, but with enough context so that sentry can filter out everything but production / staging errors when notifying us via slack or email.

PR #975, opened 2020-02-05
Feed-based store

tomclose:


packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -0,0 +1,168 @@
...
+}
+interface Message {
+  signedStates?: SignedState[];
+  protocols?: Protocol[];
+}
+
+export type Protocol = CreateAndDirectFund;
+
+export interface StoreV2 {
+  onStateReceived: (channelId: string) => Observable<any>; // observable
```
andrewgordstewart: Curious about the names here. If I have a `store: StoreV2`, and I want to be notified about states received for some channel, I would call 
```
store.onStateReceived(channelId).subscribe({
  next(stateReceived) { ... }
})
```

`onStateReceived` sounds like calling `.on` on an event emitter. To me, a more natural thing to call is something like `store.stateReceivedFeed(channelId).subscribe({ ... })` -- a `Feed` is something you can subscribe to.

tomclose: Yeah, that's what I'm currently working with too. This interface currently isn't up-to-date with the latest approach, which is being implemented in the `MemoryStore` below.

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -0,0 +1,129 @@
...
+    return fromEvent(this._eventEmitter, 'newProtocol');
+  }
+
+  pushMessage(message: Message) {
+    const {signedStates, protocols} = message;
+
+    if (signedStates) {
+      // todo: check sig
+      // todo: check the channel involves me
+      signedStates.forEach(sstate => this._eventEmitter.emit('stateReceived', sstate.state));
```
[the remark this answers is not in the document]
tomclose: `sstate` here is deliberate, to stand for `signedState`. I should probably be more explicit.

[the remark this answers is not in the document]
andrewgordstewart: 🐍 right!

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -0,0 +1,302 @@
...
+  channelNonce: BigNumber;
+  appDefinition: string;
+  challengeDuration: BigNumber;
+}
+
+export interface State extends ChannelConstants, StateVariables {
+  channelId: string;
+}
+
+interface ChannelStorage {
```
snario: We already have a `ChannelStorage` in the Solidity code, I recommend not re-using that symbol.

https://github.com/statechannels/monorepo/blob/641d7d7987acc67e3ea035e77a1822da93ac92bf/packages/nitro-protocol/contracts/interfaces/IForceMove.sol#L39-L45

```solidity
    struct ChannelStorage {
        uint256 turnNumRecord;
        uint256 finalizesAt;
        bytes32 stateHash; // keccak256(abi.encode(State))
        address challengerAddress;
        bytes32 outcomeHash;
    }
```

PR #979, opened 2020-02-05
Virtual funding

andrewgordstewart:
Looks like: https://xstate.js.org/viz/?gist=96eeb6cdab8e13ba8c224d5c3ace2b09


andrewgordstewart:
Closing, as this will be replaced with workflow-based virtual funding work

PR #980, opened 2020-02-05
[wallet-protocols] Add CreateAndDirectFund

snario:


packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,120 @@
...
+export const advanceChannelArgs = (i: 1 | 3) => ({ channelId }: Init): AdvanceChannel.Init => ({
+  channelId,
+  targetTurnNum: i,
+});
+
+const constructFirstState = {
+  invoke: {
+    id: 'constructFirstState',
+    src: 'constructFirstState',
+    data: (ctx: Init) => ctx,
```
andrewgordstewart:
```suggestion
```

You get access to the context and the most recent event when invoke a promise. See the example [here](https://xstate.js.org/docs/guides/communication.html#invoking-promises)

packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,120 @@
...
+    id: 'preFundSetup',
+    src: 'advanceChannel',
+    data: advanceChannelArgs(1),
+    onDone: 'funding',
+  },
+  on: { CHANNEL_CLOSED: 'abort' },
+};
+
+// FIXME: Abort should not be success
+const abort = success;
```
andrewgordstewart:
```suggestion
```

I think it's safe to just get rid of `abort` for now, as we do not yet have a plan for managing failures.

packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,120 @@
...
+  states: {
+    constructFirstState,
+    preFundSetup,
+    abort,
+    funding: directFunding,
+    postFundSetup,
+    success: {
+      type: 'final' as 'final',
+
+      entry: sendParent({ type: 'CHANNEL_CREATED' }),
```
andrewgordstewart: I don't think we need to send `CHANNEL_CREATED` to the parent anymore. Likewise, the channelId is already known to the parent -- it supplied it as the `Init` args -- so it doesn't need to be returned in the next line.

packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,120 @@
...
+import { AdvanceChannel, DirectFunding } from '..';
+
+const PROTOCOL = 'create-and-direct-fund';
+
+export enum Indices {
+  Left = 0,
+  Right = 0,
+}
+
+export type Init = JsonRpcCreateChannelParams & {
```
andrewgordstewart: Since the channel is already initialized, the `Init` type can be
```
{
  channelId: string,
  ... // whatever data is needed to construct the first state
}
```

Or, if you want to bump the responsibility of constructing the first state to the caller of this protocol, it can be
```
{
  channelId: string;
  firstState: State
}
```

packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,122 @@
...
+  participants: Participant[];
+  allocations: AllocationAssetOutcome[];
+  appDefinition: string;
+  appData: string;
+  channelId: string;
+  challengeDuration: number;
+  index: Indices;
+};
+
+export type FirstStateConstructed = Init & { type: 'FIRST_STATE_CONSTRUCTED' };
```
andrewgordstewart:
```suggestion
```

packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,122 @@
...
+    directFunding,
+    postFundSetup,
+    success: {
+      type: 'final' as 'final',
+    },
+  },
+};
+
+export const machine: MachineFactory<Init, any> = (store: Store, init: Init) => {
+  async function constructFirstState(ctx: Init): Promise<FirstStateConstructed> {
```
andrewgordstewart: `constructFirstState` is an invoked service whose return value is unimportant. All the machine cares about is that the service resolved.

```suggestion
  async function constructFirstState(ctx: Init): Promise<void> {
```

packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,122 @@
...
+      channel: store.getEntry(channelId).channel,
+      appData,
+      appDefinition,
+      isFinal: false,
+      turnNum: 0,
+      outcome: [],
+      challengeDuration,
+    });
+
+    return { ...ctx, type: 'FIRST_STATE_CONSTRUCTED' };
```
andrewgordstewart:
```suggestion
```

packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,122 @@
...
+export type FirstStateConstructed = Init & { type: 'FIRST_STATE_CONSTRUCTED' };
+
+export const advanceChannelArgs = (i: 1 | 3) => ({ channelId }: Init): AdvanceChannel.Init => ({
+  channelId,
+  targetTurnNum: i,
+});
+
+const constructFirstState = {
+  invoke: {
+    id: 'constructFirstState',
```
andrewgordstewart:
```suggestion
```
The id isn't necessary if it has the same value as `src`

packages/wallet-protocols/src/protocols/create-and-direct-fund/protocol.ts
```diff
@@ -0,0 +1,122 @@
...
+const abort = success;
+
+const directFunding = {
+  invoke: {
+    src: 'directFunding',
+    data: ({ allocations, channelId }: FirstStateConstructed): DirectFunding.Init => {
+      return {
+        channelId,
+        // TODO: Error handling on non-0 case ... also multi-asset
+        minimalAllocation: allocations[0].allocationItems,
```
andrewgordstewart: There is a `getEthAllocation` function that could be used. Using it would help keep track of where updates are needed for multi-asset protocols.

PR #981, opened 2020-02-06
Use knex TypeScript support

snario:
Running `yarn hub:test` locally caused errors. See [here](https://statechannels.slack.com/archives/CMPHW40R0/p1575241726271600) for an example discussion thread on Slack. I [discovered](https://github.com/knex/knex/issues/2998#issuecomment-544822207) you can use `--cwd` to get TypeScript support, which enabled `knex` to validate migration files under `src` with `.ts` extensions as opposed to `.js` extensions which caused the errors in the earlier linked Slack thread.


kerzhner:
This is probably ok for now. In general, `migrate` command will be used outside of development environment to migrate staging and production databases. In those environments, we probably shouldn't use uncompiled typescript.

kibertoad:
@snario TypeScript support was broken prior to 0.20.9, you might want to update

snario:
Strange; seemed to work for me when I tested this. Will upgrade regardless.

PR #987, opened 2020-02-07
[wallet, rps] Call ethereum.enable() in the wallet

### LINKED ISSUE #955: rps is broken
I tried playing the game with two Chrome tabs but saw the below error. And the page stalled here forever: 

rps-screenshot

Did I do something wrong?

Environment: 
Mac OS 10.15.3
Chrome: Version 79.0.3945.130 (Official Build) (64-bit)
MetaMask: 7.7.4
Ropsten Testnet address: https://ropsten.etherscan.io/address/0x437864550474980e76f4559ee9b10e96b28ea6b5

geoknee:
Fixes #955 

Any ostensibly easy fix snowballed as I tried to get a reasonable user experience. Currently both the app and the wallet try to access `window.etherem.selectedAddress`, meaning that both should call `window.ethereum.enable()` at some early stage.

The app has some extra machinery to make sure that the user is 
 - not surprised by the ensuing metamask popup
 - not able to progress until authorizing

The wallet does not have that machinery, so the interim solution of this PR is:

-  to not release the state channel signing address to the app until the wallets own `window.ethereum.enable()` workflow resolves.
- to disable buttons in the lobby until the signing address is released

In practice the app and wallet make their `enable()` calls in quick succession, but the prompt from the wallet's call does not pop up (just get the little [1] on the fox's face). 

Starting from a fresh metamask (no connections):

mm-enable

TODO 

- [ ] fix unknown problem where cannot fund game

### DIFF
--- a/packages/rps/.env
+++ b/packages/rps/.env
@@ -12,7 +12,7 @@ CHAIN_NETWORK_ID=9001
 TARGET_NETWORK='development'
 FIREBASE_PROJECT='rock-paper-scissors-dev-4ff8d'
 FIREBASE_API_KEY='AIzaSyAOvhDzJir_El3O6SJ2xQlrpOisnObq6zw'
-WALLET_URL='http://localhost:3055'
+WALLET_URL='http://127.0.0.1:3055'
 BOT_URL=http://localhost:3002
 ##
 
--- a/packages/rps/.env
+++ b/packages/rps/.env
@@ -16,6 +16,10 @@ WALLET_URL='http://127.0.0.1:3055'
 BOT_URL=http://localhost:3002
 ##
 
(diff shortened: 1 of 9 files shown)


geoknee:
Latest thinking on this is to remove the connection between the app (`rps`) and metamask. 

- Having two connections makes for nasty UX.
- The wallet definitely needs to talk to metamask in order to submit transactions
- The app only needs the `outcomeAddress` from metamask. It could get this from the wallet instead, at the same time as getting the `signingAddress`. 

All of the existing UX around "Connect to MetaMask" and "wrong network selected" could remain in place but go via the wallet where necessary. (I think the network status will be provided even without calling `ethereum.enable()`).

geoknee:
Closing in favor of #991

PR #997, opened 2020-02-10
Create interface for Store/ Store Entries

lalexgap:
Stab at creating interfaces for the store and store entry.

### DIFF
--- a/packages/xstate-wallet/src/store/memory-channel-storage.ts
+++ b/packages/xstate-wallet/src/store/memory-channel-storage.ts
@@ -1,4 +1,4 @@
-import {ChannelConstants, StateVariables, SignedState, Participant} from './types';
+import {ChannelConstants, StateVariables, SignedState, Participant, State} from './types';
 import {
   getChannelId,
   signState,
@@ -6,13 +6,16 @@ import {
   hashState,
   getStateSignerAddress
 } from '@statechannels/nitro-protocol';
-import {joinSignature, splitSignature} from 'ethers/utils';
+import {splitSignature, joinSignature} from 'ethers/utils';
 
-interface ChannelStorage {
-  myIndex: number;
-  channelConstants: ChannelConstants;
-  stateVariables: Record<string, StateVariables>;
-  signatures: Record<string, (string | undefined)[]>;
+export interface ChannelStorage {
+  readonly channelId: string;
+  readonly myIndex: number;
+  readonly latestSupported: State | undefined;
+  readonly latestSupportedByMe: State | undefined;
+  readonly channelConstants: ChannelConstants;
+  readonly stateVariables: Record<string, StateVariables>;
+  readonly signatures: Record<string, (string | undefined)[]>;
 }
 
 export class MemoryChannelStorage implements ChannelStorage {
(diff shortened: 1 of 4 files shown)


packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -172,11 +190,11 @@ export class MemoryStore {
       });
     }
 
-    if (protocols) {
-      protocols.forEach(protocol => {
-        if (!this._protocols.find(p => _.isEqual(p, protocol))) {
-          this._protocols.push(protocol);
-          this._eventEmitter.emit('newProtocol', protocol);
+    if (objectives) {
+      objectives.forEach(objective => {
```
andrewgordstewart:
```suggestion
      objectives?.forEach(objective => {
```

Using [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining), we can get rid of the `if (objectives) { // type is narrowed }`

packages/xstate-wallet/src/store/memory-channel-storage.ts
```diff
@@ -6,13 +6,18 @@ import {
...
-import {joinSignature, splitSignature} from 'ethers/utils';
-
-interface ChannelStorage {
-  myIndex: number;
-  channelConstants: ChannelConstants;
-  stateVariables: Record<string, StateVariables>;
-  signatures: Record<string, (string | undefined)[]>;
+import {splitSignature, joinSignature} from 'ethers/utils';
+
+export interface ChannelStorage {
```
snario: I would not reuse the name `ChannelStorage`. Tom is using it in his PR and it's also in the Solidity code.

packages/xstate-wallet/src/store/types.ts
```diff
@@ -17,42 +16,45 @@ export interface StateVariables {
...
 
-interface MixedAllocation {
+export interface MixedAllocation {
   type: 'MixedAllocation';
   ethAllocation?: SimpleEthAllocation;
   tokenAllocations?: SimpleTokenAllocation[];
 }
-export type Outcome2 =
+// TODO: Better name?
+export type AuxillaryChannelOutcome =
```
andrewgordstewart: 💡 This could be `SimpleOutcome`?

lalexgap: That seems quite obvious in retrospect.

packages/xstate-wallet/src/store/memory-channel-storage.ts
```diff
@@ -11,11 +11,13 @@ import {splitSignature, joinSignature} from 'ethers/utils';
 export interface ChannelStorage {
   readonly channelId: string;
   readonly myIndex: number;
-  readonly latestSupported: State | undefined;
-  readonly latestSupportedByMe: State | undefined;
+  readonly latest: StateVariables | undefined; // TODO: Should this always be defined?
```
andrewgordstewart: I'm not sure if you'd ever call `createChannel` without passing the first state. It might be worth it to _require_ the first state, and in that case it should always be defined.

PR #998, opened 2020-02-10
Disable Integration with Hub in CircleCI

lalexgap:
Since the integration with hub test fails pretty consistently I think we should just remove it from circleCI now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -306,11 +306,11 @@ workflows:
       - integration-test:
           requires:
             - prepare
-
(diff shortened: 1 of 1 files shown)


tomclose:
I think it's worth doing this. For example, of the 6 PRs open at the moment, 2 are failing because of this test.

If we want to keep running this and working on getting it to pass reliably, I'd suggest setting it up on a schedule, to run once a day or something, so that it doesn't interrupt our dev workflow.

PR #1000, opened 2020-02-11
Define some Objectives

andrewgordstewart:


packages/xstate-wallet/src/store/wire-protocol.ts
```diff
@@ -0,0 +1,41 @@
+import {SignedState, Participant} from './types';
+
+type _Objective<Name, Data> = {
```
lalexgap: Is the `_` part of a naming convention?

andrewgordstewart: It seems to be used for unused variables. I couldn't think of a better name at the time.

PR #1001, opened 2020-02-11
[xstate-wallet]  Call ethereum enable and pass selectedAddress to app on request

geoknee:
Restores compatibility with RPS.

### DIFF
--- a/packages/app-wallet-interface/source/index.html.md
+++ b/packages/app-wallet-interface/source/index.html.md
@@ -109,9 +109,9 @@ Note: we don't return the state of the channel, as messages are not necessarily 
 | ---- | ----------------- | -------------------------------------------- |
 | 900  | Wrong Participant | The message is not addressed to this wallet. |
 
-## Get addresses
+## GetAddress
 
-Returns the signing address(es) for the current domain.
+Returns the signing address for the current domain.
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


packages/app-wallet-interface/source/index.html.md
```diff
@@ -132,6 +132,29 @@ Returns the signing address(es) for the current domain.
 }
 ```
 
+## GetEthereumSelectedAddress
+
+Returns the ethereum address selected in metamask. Typically used as the `destination`.
```
geoknee: We use `destination` in this doc, but the `rps` uses `outcomeAddress`. Should we change?

packages/app-wallet-interface/source/index.html.md
```diff
@@ -109,9 +109,9 @@ Note: we don't return the state of the channel, as messages are not necessarily
 | ---- | ----------------- | -------------------------------------------- |
 | 900  | Wrong Participant | The message is not addressed to this wallet. |
 
-## Get addresses
+## GetAddress
```
tomclose: I wonder if this should be `getParticipant`, and should return both addresses? It seems like we'll probably need the participant id too at some point (how this is set isn't currently defined, but it'll likely need some information about the tab so that the messages can be directed to the right app window).

geoknee: Maybe it could be in future. Possible reasons for keeping the calls separate: 
- the app may not need the `EthereumSelectedAddress` (it could prompt the user to input some other address to use as the `outcomeAddress` such as a contract that the user owns). Raises some trust concerns, though. 
- only one of the calls causes the wallet to call `window.ethereum.enable()`, which is quite a major threshold in the UX. Keeping 2 x json-RPCs allows for greater flexibility in managing that.

PR #1005, opened 2020-02-11
Port support state to xstate-wallet

andrewgordstewart:
Notes:
1. This includes the most recent changes from master as well as from https://github.com/statechannels/monorepo/pull/1003. Hence, `master` is the most natural base to merge into, though 1681fd53b0e06e99163a40147dd4a81399d28aa4 is not part of this PR proper.
2. I do not expect this to work; it simply compiles. In fact, I know it does not work, since `outcomesEqual` is not implemented.

### DIFF
--- a/packages/xstate-wallet/src/store/memory-store.ts
+++ b/packages/xstate-wallet/src/store/memory-store.ts
@@ -43,11 +43,10 @@ interface InternalEvents {
 }
 
 export interface Store {
-  channelUpdatedFeed(channelId: string): Observable<ChannelStoreEntry>;
-  newObjectiveFeed(): Observable<Objective>;
-
-  pushMessage: (message: Message) => void;
+  newObjectiveFeed: Observable<Objective>;
   outboxFeed: Observable<Message>;
+  pushMessage: (message: Message) => void;
+  channelUpdatedFeed(channelId: string): Observable<ChannelStoreEntry>;
 
   getAddress(): string;
   addState(channelId: string, stateVars: StateVariables);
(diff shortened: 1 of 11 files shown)


packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -187,4 +187,14 @@ export class MemoryStore {
       }
     });
   }
+
+  public async getEntry(channelId: string): Promise<ChannelStoreEntry> {
```
lalexgap: Ha, I added the same method on my branch

packages/xstate-wallet/src/utils.ts
```diff
@@ -0,0 +1,36 @@
+import {filter, map, flatMap, concatMap} from 'rxjs/operators';
+import {from, OperatorFunction} from 'rxjs';
+
+function flatFilter<T>(predicate: (event: T) => boolean | Promise<boolean>) {
```
lalexgap: It might be nice to isolate observable utils in `rxjs-utils` or something like that.

packages/xstate-wallet/src/workflows/supportState.ts
```diff
@@ -0,0 +1,76 @@
+import {AnyEventObject, AssignAction, MachineConfig, assign, spawn, Machine} from 'xstate';
+import {filter, map} from 'rxjs/operators';
+import {Store} from '../store';
+import {State} from '../store/types';
+import {statesEqual, outcomesEqual} from '../store/state-utils';
+
+const PROTOCOL = 'support-state';
```
lalexgap: minor, rename to `WORKFLOW`?

PR #1006, opened 2020-02-11
Create json rpc types by hand

tomclose:
Fixes https://github.com/statechannels/monorepo/issues/957


packages/channel-client/.vscode/launch.json
```diff
@@ -0,0 +1,72 @@
+{
```
geoknee: Does this belong in this PR?

packages/channel-client/.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
geoknee: Does this belong?

packages/channel-provider/src/types.ts
```diff
@@ -1,14 +1,14 @@
...
 import {
-  CreateChannelResult,
-  CloseChannelResult,
-  UpdateChannelResult,
-  PushMessageResult,
-  JoinChannelResult,
-  GetAddressResult,
-  GetEthereumSelectedAddressResult,
-  ChallengeChannelResult,
-  BudgetResult1 as BudgetResult
```
geoknee: :+1:

packages/client-api-schema/package.json
```diff
@@ -31,15 +17,32 @@
...
+  "devDependencies": {
+    "@types/eslint": "^6.1.3",
+    "@types/eslint-plugin-prettier": "^2.2.0",
+    "@typescript-eslint/eslint-plugin": "^2.10.0",
+    "@typescript-eslint/parser": "^2.10.0",
+    "eslint": "^6.7.2",
+    "eslint-plugin-prettier": "^3.1.1",
+    "json-schema-to-typescript": "^8.0.0",
+    "lint-staged": "9.4.0"
+  }
```
geoknee: Was it necessary to change these dependencies? Could this be the source of your linting woes?

tomclose: Hmm. This happened during a rebase. Will try and fix.

packages/channel-client/.vscode/settings.json
tomclose: Probably not strictly, but I think it's probably ok (?) - it's just importing the `.vscode` functionality that probably should have set up when the package was created.

Issue #2, opened 2019-09-20
[web3torrent] Merge external repo into the monorepo


joelalejandro:
Now that #1 is implemented, we can bring the `web3torrent-pse` repo from @prototypal to @statechannels.

PR #1011, opened 2020-02-12
Downgrade solc, do not use native solc in CI

kerzhner:
1. Downgrade `solc` npm package since our contracts still use solidity version `^0.5.13`. `openzeppelin` does not support solidity version 0.6 yet, and our contracts import `openzeppelin` contracts.
1. Maybe more controversial. Do not use native solc on CircleCI. The downside is that the prepare step is 30 seconds slower. The upside the following. With native solc (installed through apt-get in the docker image), we are not able to control what version of solc is used. As a result, every time we rebuild the docker image, we are forced to upgrade to the latest solc package.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,8 +9,6 @@ defaults: &defaults
   resource_class: medium
   docker:
     - image: counterfactual/statechannels:0.5.13 # Fast contract compilation with solc installed
-      environment:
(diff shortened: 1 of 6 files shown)


snario:
I think we can find a way to fix #2.

snario:
https://ethereum.stackexchange.com/a/64525/9000

PR #1009, opened 2020-02-11
Update xstate-wallet to use new Store

lalexgap:
Updates the application workflow and xstate wallet to work with the new `Store`
Adds the following protocols as workflows:
- DirectFunding
- Depositing
- CreateAndDirectFund
- AdvanceChannel

PR #1012, opened 2020-02-12
Store states when they arrive

andrewgordstewart:


packages/xstate-wallet/src/store/memory-channel-storage.ts
```diff
@@ -8,17 +9,14 @@ export interface ChannelStoreEntry {
...
-  readonly signatures: Record<string, (string | undefined)[]>; // TODO: Should we even expose this?
 }
 
 export class MemoryChannelStoreEntry implements ChannelStoreEntry {
   constructor(
     public readonly channelConstants: ChannelConstants,
     public readonly myIndex: number,
-    public stateVariables: Record<string, StateVariables> = {},
-    public signatures: Record<string, string[]> = {}
+    private stateVariables: Record<string, StateVariables> = {},
```
lalexgap: 👍

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -106,9 +106,8 @@ export class MemoryStore implements Store {
   public async createChannel(
     participants: Participant[],
     challengeDuration: BigNumber,
-    stateVars: StateVariables,
```
tomclose: What's the logic here? Don't you always know what the first state will be at the time you create the channel?

tomclose:
I reverted this PR, as it was not immediately compatible with Alex's work in https://github.com/statechannels/monorepo/pull/1009.

The PR also does a bit more than billed, making some non-trivial changes to the interface to the store. These changes aren't used anywhere in the PR, so I was hoping to understand more about why they were made. These include:

* Changing the behaviour of `createChannel`, so it no-longer creates the initial state.
    * Are there any examples where we don't know the initial state at the time we create the channel? 
* Changing `createChannel` to return a `ChannelStoreEntry`, instead of the `channelId`.
    * The original intention was to try and keep the surface area of the new store small. I understand that we decided to make a `ChannelStoreEntry` object back initially, to help with existing protocols, but is it required here?
* Splitting the existing `addState` method into two public methods: `signState` and `addState`.
    * It isn't clear why we need the other external method - it seems like we either add states from protocols (currently `addState`), or from external messsges (currently `pushMessage`).
    * I chose `addState` over `signState` originally, as I was concerned that `signState` didn't give any indication that state was being stored and synced, and it seemed more intuitive to understand that signing and syncing would happen automatically when adding a state to a channel. Some alternatives:  `signAndSendState`, `addToChannel`?

Happy for the PR to be resubmitted once it's been rebased onto the updated master.

packages/xstate-wallet/src/store/memory-store.ts
andrewgordstewart: It just became an unused variable.

andrewgordstewart:
> I reverted this PR, as it was not immediately compatible with Alex's work in #1009.
> 
> The PR also does a bit more than billed, making some non-trivial changes to the interface to the store. These changes aren't used anywhere in the PR, so I was hoping to understand more about why they were made.

The point of this PR is that `pushMessage` [does not store states](https://github.com/statechannels/monorepo/pull/1012/files#diff-550d8631da011300e059fa60a006db77L178-L181). Prior, there was no way to actually do this: there were only two ways to get states into the store: `addState` and `createChannel`. Neither of these methods accept a signature, and internally, the store signs states with the user's private key.

>These include:
> 
> * Changing the behaviour of `createChannel`, so it no-longer creates the initial state.
>   
>   * Are there any examples where we don't know the initial state at the time we create the channel?

I don't know. But if my wallet receives a signed state for you, it can't [store it](https://github.com/statechannels/monorepo/pull/1012/commits/52773bf61c3f54e261d79573f74464c8a5bed18f) if nothing exists at `this.channels[channelId]`. It's obviously unwise to blindly sign that state on arrival, so `createChannel` should not sign on the user's behalf.

> * Changing `createChannel` to return a `ChannelStoreEntry`, instead of the `channelId`.
>   * The original intention was to try and keep the surface area of the new store small. I understand that we decided to make a `ChannelStoreEntry` object back initially, to help with existing protocols, but is it required here?
Given that `ChannelStoreEntry` exposes a `channelId` property, this is strictly an increase in functionality:
`const channelId = createChannel(...)` can be written as `const { channelId } = createChannel(...)`

Returning a `ChannelStoreEntry` means you can replace [this](https://github.com/statechannels/monorepo/pull/1012/commits/52773bf61c3f54e261d79573f74464c8a5bed18f#diff-550d8631da011300e059fa60a006db77R167-R173) with something like
```
const channelStorage = await this.getEntry(channelId) || await this.createChannel(state)
```

> * Splitting the existing `addState` method into two public methods: `signState` and `addState`.
>   
>   * It isn't clear why we need the other external method - it seems like we either add states from protocols (currently `addState`), or from external messsges (currently `pushMessage`).
>   * I chose `addState` over `signState` originally, as I was concerned that `signState` didn't give any indication that state was being stored and synced, and it seemed more intuitive to understand that signing and syncing would happen automatically when adding a state to a channel. Some alternatives:  `signAndSendState`, `addToChannel`?

`pushMessage` currently does not add states to the store. This PR changes this, by having `pushMessage` call `addState`. It's more surprising to me that `addState` signs the state for me, than that `signState` stores the state for me. If I _ever_ witness a signed state, I want to store it unless I have a good reason to discard it. This includes states that I sign.

It doesn't matter if it's called `signState` or `signAndAdd` state, but there needs to be some method with signature `(signedState: SignedState) => any` which stores signed states; it's just as well for `signStateOrWhateverItsCalled` to call that method.

tomclose:
You're right that, as it's currently written, `pushMessage` doesn't store states, and that that bit needs to be added. It doesn't seem like that necessarily means that the other methods should change though? The intention was that there would be three ways of getting a state into the store:

* `createChannel`: called by the wallet owner's protocols, when they want to create a channel, and set the initial state
* `addState`: called by the wallet owner's protocols, when they want to add a state to an existing channel
* `pushMessage`: which unpacks messages from opponents, and adds the states, creating channels if necessary. Called directly from the messaging layer, and not by the wallet owner's protocols.

Regardless of what we call `add/sign/signAndAddState`, I don't believe we need to add an extra method to the external store interface?

andrewgordstewart:
> Regardless of what we call `add/sign/signAndAddState`, I don't believe we need to add an extra method to the external store interface?

Hm, yes, so the simple thing to do is remove `[addState](https://github.com/statechannels/monorepo/pull/1012/files#diff-550d8631da011300e059fa60a006db77R54)` from the store API (and then rename as we wish.)

tomclose:
I think (?) `addState` is one of the things that is now in use after #1009, so it might be better to just rename.

andrewgordstewart:
> I think (?) `addState` is one of the things that is now in use after #1009, so it might be better to just rename.

https://github.com/statechannels/monorepo/pull/1016/commits/f3e5a26f1f6967161ac9b5da7842dd6cf1de5602

PR #1015, opened 2020-02-12
sync deploy branch with master

geoknee:
This is my suggestion for how we do deploys. 

Screenshot 2020-02-12 at 16 08 29

Don't create any new branches. 
Don't need to use command line. 

The `deploy` branch is protected (needs PR reviews, requires linear history). 

We should merge with "rebase and merge". 

We should think about disabling "squash and merge", because that would introduce commits to `deploy` that are not part of `master`. 

The PR reviewer checks that the base branch is master.


geoknee:
One thing to note: between raising the PR and merging it, `master` might change. So it is more difficult to pin the deployment to a specific commit this way. That can still be done via command line and raising a PR from a new branch, though.

kerzhner:
> One thing to note: between raising the PR and merging it, `master` might change. So it is more difficult to pin the deployment to a specific commit this way. That can still be done via command line and raising a PR from a new branch, though.

That is one of the reasons I was thinking that a no PR strategy is perhaps simpler.

PR #1016, opened 2020-02-12
Revert "Revert "Store states when they arrive""

andrewgordstewart:
- Reverts statechannels/monorepo#1014
- Correctly uses `signState` in ported protocols
-  Replace `Record<string, SomeType>` with `Record<string, SomeType | undefined>`, forcing users to deal with non-existing entries.

### DIFF
--- a/packages/xstate-wallet/src/store/memory-store.ts
+++ b/packages/xstate-wallet/src/store/memory-store.ts
@@ -69,7 +69,7 @@ export class MemoryStore implements Store {
   protected _chain: Chain;
   private _channels: Record<string, MemoryChannelStoreEntry | undefined> = {};
   private _objectives: Objective[] = [];
-  private _nonces: Record<string, BigNumber> = {};
+  private _nonces: Record<string, BigNumber | undefined> = {};
   private _eventEmitter = new EventEmitter<InternalEvents>();
   private _privateKeys: Record<string, string> = {};
   // private _channels: Record<string, any> = {};
@@ -127,7 +127,7 @@ export class MemoryStore implements Store {
     const channelId = calculateChannelId(channelConstants);
 
     // TODO: There could be concurrency problems which lead to entries potentially being overwritten.
-    this.checkNonce(addresses, channelConstants.channelNonce);
+    this.setNonce(addresses, channelConstants.channelNonce);
     const entry =
       this._channels[channelId] || new MemoryChannelStoreEntry(channelConstants, myIndex);
 
@@ -148,9 +148,7 @@ export class MemoryStore implements Store {
       throw new Error("Couldn't find the signing key for any participant in wallet.");
     }
 
-    const currentNonce = this.getNonce(addresses);
-    const channelNonce = currentNonce ? currentNonce.add(1) : bigNumberify(0);
-    this.setNonce(addresses, channelNonce);
+    const channelNonce = this.getNonce(addresses).add(1);
     const chainId = '1';
 
     const entry = await this.initializeChannel({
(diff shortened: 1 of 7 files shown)


andrewgordstewart:
> Is this PR meant to be compatible with the approach outlined here, [#1012 (comment)](https://github.com/statechannels/monorepo/pull/1012#issuecomment-585275065)? Or are you intending to propose a new interface for the store?

> * `createChannel`: called by the wallet owner's protocols, when they want to create a channel, and set the initial state

What method is called in `pushMessage` when the channel doesn't exist?

tomclose:
> I guess there can be a private initializeChannel method which does not sign the first state.

👍

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -115,51 +113,70 @@ export class MemoryStore implements Store {
...
   private setNonce(addresses: string[], value: BigNumber) {
+    if (value.lte(this.getNonce(addresses))) throw 'Invalid nonce';
+
     this._nonces[this.nonceKeyFromAddresses(addresses)] = value;
   }
 
   private nonceKeyFromAddresses = (addresses: string[]): string => addresses.join('::');
 
-  addState(channelId: string, stateVars: StateVariables) {
+  signState(channelId: string, stateVars: StateVariables) {
```
lalexgap: Should this be `signAndAddState`?  From the name I would expect I need to call `signState` and then `addState` but it looks like `signState` adds it by calling `channelStorage.signAndAdd`?

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -177,24 +194,34 @@ export class MemoryStore implements Store {
...
       // todo: check the channel involves me
-      signedStates.forEach(signedState => {
-        const {signature, ...state} = signedState;
-        this._eventEmitter.emit('stateReceived', state);
-      });
+      await Promise.all(
+        signedStates.map(async signedState => {
+          const {signature, ...state} = signedState;
+          await this.addState(signedState);
+          this._eventEmitter.emit('stateReceived', state);
```
lalexgap: I think this should be switched to `channelUpdated` and we should just use the `ChannelUpdatedFeed` for now.

packages/xstate-wallet/src/store/memory-store.ts
andrewgordstewart: I don't think the store should ever sign a state and not add it: If I sign it, I should add it; otherwise, I risk giving someone else unilateral control over the channel. Likewise, if I sign a state, an honest wallet should always tell its peers; otherwise it is keeping control over the channel to itself, leading to broken protocols.

The API of `MemoryChannelStoreEntry` should be irrelevant, whether or not `signState` on the `Store` interface is called `signState` or `signAndAddState`. Do you prefer `signAndAddState` for clarity, or is `signState` good enough in light of this?

packages/xstate-wallet/src/store/memory-store.ts
andrewgordstewart: Agreed

andrewgordstewart: I would prefer deferring this to a different PR though -- even though `memory-store.test` is testing `stateReceivedFeed` and not `channelUpdatedFeed`, the `stateReceivedFeed` is not on the `Store` interface. So, protocols should not be using it.

packages/xstate-wallet/src/store/memory-store.ts
lalexgap: I think I prefer `signAndAddState` (or some variant of that) for clarity.

PR #1017, opened 2020-02-12
[xstate-wallet] Replace subscribeDepositEvent invoked callback with subscribeToFundingFeed invoked Observable

geoknee:
This PR includes 

- Replacing `subscribeDepositEvent` invoked callback with `subscribeToFundingFeed` invoked Observable. Why? I did this work in parallel. It's probably a sideways move at best but wondering what anyone else thinks
- A test that models sending events to an xstate machine, after a deposit is made to ganache and the new Observable emits the `ChainEvent`.
- test configuration ported over from the redux wallet. 
- some hacks around `ObsoleteStore`  and `Chain` interfaces that I did in order to hook up the `depositing` protocol to the Observable
- adds both adjudicator and an array of asset holders to the `ChainWatcher`

I hope some (if not all) of this will be useful.

TODO 

- [x] figure out why ganache is not being torn down correctly


packages/wallet-protocols/src/protocols/depositing/protocol.ts
```diff
@@ -1,10 +1,11 @@
 import { bigNumberify } from 'ethers/utils';
```
lalexgap: The `depositing` protocol has been migrated to `workflow\depositing`, I probably should have flagged this one as obsolete :(

geoknee:
Closing in favour of #1020

PR #1021, opened 2020-02-13
Generate json-rpc schema from typescript

tomclose:
And add a `parseRequest` method, which is guaranteed to return a `Request` object (or throw an error).


geoknee:
Should the generated schema be checked in to source control?

(In the previous pattern that we have not inverted, the generated types were not checked in). 

Also , we can probably remove `types` from the `.gitignore`.

tomclose:
I'm not sure. The reason I checked them in was to make it easy for someone browsing github to (a) see that we have a schema, and (b) to inspect it. Equally there are arguments for not checking it in (like we don't check in generated js etc. etc.).

geoknee:
Makes sense to me, just wanted to ask the question.

PR #1022, opened 2020-02-13
[base-not-master] [client-api-schema] Add annotations to types

geoknee:
So that the generated schema can include patterns to match against. Example: an Ethereum address is not just a string.


packages/client-api-schema/src/types.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * Ethereum Address
+ * @pattern  ^0x([a-fA-F0-9]{40})|0$
+ */
+type Address = string;
```
tomclose: I think if you do `export type Address = string` it will generate a definition in the schema for it.

PR #1023, opened 2020-02-13
Xstate MessagingService

tomclose:
Working towards a world where we can initialize the wallet like this (instantiating a class, and then connecting it to postMessage):
```ts
import {ChannelWallet} from './channel-wallet';

const channelWallet = new ChannelWallet();

window.addEventListener('message', async event => {
  channelWallet.pushMessage(event.data);
});
channelWallet.onSendMessage(m => window.parent.postMessage(m, '*'));
```
This should make it easier to run tests etc.


lalexgap:
Closing this in favour of #1024

PR #1024, opened 2020-02-13
XState Messaging Service

lalexgap:
Work to get a message service working based off of #1023 

The `workflowManager` has been removed. With some refactoring it seemed to make sense for the `ChannelWallet` to handle workflow duties and have one less service/object to worry about.

> Working towards a world where we can initialize the wallet like this (instantiating a class, and then connecting it to postMessage):
> 
> ```ts
> import {ChannelWallet} from './channel-wallet';
> 
> const channelWallet = new ChannelWallet();
> 
> window.addEventListener('message', async event => {
>   channelWallet.pushMessage(event.data);
> });
> channelWallet.onSendMessage(m => window.parent.postMessage(m, '*'));
> ```
> 
> This should make it easier to run tests etc.


tomclose:
I rebased this on top of master here: https://github.com/statechannels/monorepo/pull/1026

tomclose:
Suggest closing this?

PR #1025, opened 2020-02-14
Integration test for xstate wallet

lalexgap:
A simple "integration" test that tests that two wallets can fund an app. Blockchain interactions are faked using the `FakeChain`.


packages/xstate-wallet/src/integration-tests/basic.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      map(m => m as CreateChannelResponse)
+    )
+  );
+
+  await playerA.messagingService.receiveMessage(createEvent);
+
+  playerA.channelWallet.workflows[0].machine.send({type: 'USER_APPROVES'});
+
+  const response = await playerAResponsePromise;
+  expect(response.result).toBeDefined();
```
junjizhi: Can we be specific about the response result?

packages/xstate-wallet/src/integration-tests/basic.test.ts
```diff
@@ -0,0 +1,170 @@
...
+  get signingAddress() {
+    return new Wallet(this.privateKey).address;
+  }
+  store: Store;
+  messagingService: MessagingServiceInterface;
+  channelWallet: ChannelWallet;
+
+  get workflow(): Workflow | undefined {
+    return this.channelWallet.workflows.length > 0 ? this.channelWallet.workflows[0] : undefined;
+  }
```
junjizhi: This getter is only used in one place and everywhere else is `channelWallet.workflows[0]`. I don't think this is necessary, because you have use optional chaining to deal with the undefined value.

packages/xstate-wallet/src/integration-tests/basic.test.ts
lalexgap: Yup, updated.

packages/xstate-wallet/src/integration-tests/basic.test.ts
lalexgap: I've changed this to `WorkflowState` to make the tests cleaner.

PR #1026, opened 2020-02-14
Xstate Messaging Service (rebased)

tomclose:


packages/xstate-wallet/src/channel-wallet.ts
```diff
@@ -0,0 +1,65 @@
...
+import {convertToOpenEvent} from './utils/workflow-utils';
+export interface Workflow {
+  id: string;
+  machine: Interpreter<any, any, any>;
+  domain: string; // TODO: Is this useful?
+}
+export class ChannelWallet {
+  private workflows: Workflow[];
+
+  constructor(private store: Store, private messagingService: MessagingServiceInterface) {
```
tomclose: Might be worth initializing these inside, instead of passing them in (especially the MessagingService, which there will probably only be one version of).

packages/xstate-wallet/src/messaging.ts
```diff
@@ -1,67 +1,117 @@
...
-        request.method !== 'JoinChannel'
-      ) {
-        return false;
-      }
-      return request.params.channelId === channelId;
-    }),
-    map((e: MessageEvent) => {
-      return e.data.params;
-    })
+export interface MessagingServiceInterface {
```
tomclose: Wondering if it's worth having an interface here? It seems like we'll only have one `MessagingService`, so we could just use that everywhere.

lalexgap: Probably, but having a interface makes me feel safer :)

packages/xstate-wallet/src/channel-wallet.ts
lalexgap: I'm going to leave as is to allow for dependency injection for now.

packages/xstate-wallet/.vscode/launch.json
```diff
@@ -8,11 +8,20 @@
...
+
+        "${relativeFile}"
+      ],
+      "skipFiles": ["<node_internals>/**/*.js"],
       "cwd": "${workspaceRoot}",
       "protocol": "inspector",
       "console": "integratedTerminal",
-      "internalConsoleOptions": "neverOpen"
+      "internalConsoleOptions": "neverOpen",
+      "smartStep": true
```
andrewgordstewart: What was wrong with the debug config?

PR #1031, opened 2020-02-18
Replace embedded-wallet with channel-client in Web3Torrent

snario:


packages/web3torrent/package.json
```diff
@@ -23,6 +23,7 @@
     "webtorrent": "0.107.16"
   },
   "devDependencies": {
+    "@statechannels/channel-provider": "*",
```
geoknee: Should we have this as a hard dependency for now, even if that isn't the long term vision?

packages/web3torrent/src/index.tsx
```diff
@@ -1,9 +1,16 @@
+// TODO: Replace with injection via other means than direct app import
+// NOTE: This adds `channelProvider` to the `Window` object
 import '@statechannels/channel-provider';
+
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
 import {web3torrent, WebTorrentContext} from './clients/web3torrent-client';
 
+// TODO: Put inside better place than here where app can handle error case
```
geoknee: Perhaps we want to take our cues from RPS, bearing in mind that we need to get addresses from the wallet and that this can trigger the `window.ethereum.enable()` call.

packages/xstate-wallet/src/messaging/types.ts
```diff
@@ -0,0 +1,36 @@
+import {
```
geoknee: Does this belong in this PR?

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -29,6 +30,8 @@ const File: React.FC<RouteComponentProps> = () => {
   const [errorLabel, setErrorLabel] = useState('');
   const getLiveData = getTorrentAndPeersData(setTorrent, setPeers);
 
+  const channelClient = new ChannelClient(window.channelProvider);
```
geoknee: Could we add a comment here explaining that `window.channelProvider` might be a fake provider (referencing index.tsx)?

packages/web3torrent/package.json
snario: Yes I think so

packages/xstate-wallet/src/messaging/types.ts
snario: Oh, woops, nope.

packages/web3torrent/src/pages/file/File.test.tsx
```diff
@@ -9,18 +9,21 @@ import {WebTorrentAddInput} from '../../library/types';
...
+// import {JsonRpcResponse} from '@statechannels/channel-provider';
 
 const mockFileURL =
   '/file/#magnet:?xt=urn%3Abtih%3A148c62a7f7845c91e7d16ca9be85de6fbaed3a1f&dn=test.zip&xl=1398978&cost=0';
 
-const mockResponse: JsonRpcResponse = {jsonrpc: '2.0', id: 123, result: ''};
-
 Enzyme.configure({adapter: new Adapter()});
 
+// const mockResponse: JsonRpcResponse = {jsonrpc: '2.0', id: 123, result: ''};
```
junjizhi: Are these dead comments needed? Can we clean them up?

packages/web3torrent/src/pages/file/File.tsx
junjizhi: How would that comment help? As long as the channelProvider implements `ChannelProviderInterface`, `File` shouldn't care if it's fake or real. To me, a comment like that will be out of date once we replace it with real client.

packages/web3torrent/src/pages/file/File.test.tsx
```diff
@@ -84,13 +81,15 @@ describe('<File />', () => {
...
-  });
+  // TODO: Figure out how to test mocked channelClient
+  // eslint-disable-next-line
+  // it('should run approveBudgetAndFund functions when the File Button is clicked', async () => {
+  //   await act(async () => {
+  //     await component.find(testSelector('download-button')).simulate('click');
+  //   });
+  //   expect(ChannelClient.mock.instances[0].approveBudgetAndFund).toHaveBeenCalled();
+  //   expect(torrentFile).toHaveBeenCalled();
+  // });
```
junjizhi: Not sure asserting the private function `approveBudgetAndFund` is right. The test is too intrusive here. At least it should assert the functions on the `ChannelProviderInterface` function, like `send`.

junjizhi: hmm...I had this comment because I was reading `FakeChannelProvider` implementation, which declares `approveBudgetAndFund` as private. But in `ChannelClient` it is not.

packages/web3torrent/src/pages/file/File.test.tsx
snario: I only kept them because the commented out test relies on them.  Perhaps better to use `it.skip` instead and keep the code uncommented.

packages/web3torrent/src/pages/file/File.test.tsx
snario: > Update: After reading more of the code, now I see FakeChannelProvider actually implements the both provider and client logic. IMO, the test case should assert on the send function with the expected method and params.

I'm not sure that's accurate; a client with a fake provider is instantiated like:

```js
var client = new ChannelClient(new FakeChannelProvider())
```

I think your initial thought on the test being too intrusive might be right. Mostly we just want to ensure that the `ApproveBudgetAndFund` request is being sent to the wallet.

junjizhi: 👍  You are right.

PR #1032, opened 2020-02-18
Apply import rules to xstate-wallet

snario:


tomclose:
Closing this for the time being. Probably easiest to recreate given how stale it is.

PR #1034, opened 2020-02-20
[web3torrent] [base-not-master] Add comments for where we might add SC API calls

geoknee:


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -208,6 +218,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     });
 
     torrent.on(TorrentEvents.DONE, () => this.emit(ClientEvents.TORRENT_DONE, {torrent}));
+    // [ George ] Here we can call channelClient.closeChannel()
```
spersico: 👍

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -194,6 +203,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
           this.transferFunds(wire);
           break;
         case PaidStreamingExtensionNotices.START:
+          // [ George ] Here we can call channelClient.createChannel()
```
snario: Based on the diagram by @spersico (see [here](https://whimsical.com/Sq6whAwa8aTjbwMRJc7vPU)), this might happen many times for a single peer + file / channel. So, probably not the right place to create the channel.

PR #1035, opened 2020-02-20
Remove redux wallet

lalexgap:
Removes the old redux wallet.


tomclose:
Closing this for the time being. Given that it's now stale, it's probably easiest to recreate anyhow.

PR #1038, opened 2020-02-20
Some small tweaks to create-and-direct-fund.ts

snario:


packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -61,7 +61,7 @@ export interface Store {
   channelUpdatedFeed(channelId: string): Observable<ChannelStoreEntry>;
 
   getAddress(): string;
-  signAndAddState(channelId: string, stateVars: StateVariables);
+  signAndAddState(channelId: string, stateVars: StateVariables): void;
```
andrewgordstewart:
```suggestion
  signAndAddState(channelId: string, stateVars: StateVariables): Promise<void>;
```

(Adding a state in general is async.)

packages/xstate-wallet/src/workflows/create-and-direct-fund.ts
```diff
@@ -87,24 +82,22 @@ export const config: MachineConfig<Context, any, any> = {
 };
 
 export const machine: MachineFactory<Init, any> = (store: Store, init: Init) => {
-  async function constructFirstState(ctx: Init): Promise<void> {
-    const {appData, channelId, outcome} = ctx;
-
+  const constructFirstStateService = async ({appData, channelId, outcome}: Init): Promise<void> =>
     store.signAndAddState(channelId, {
```
andrewgordstewart:
```suggestion
    await store.signAndAddState(channelId, {
```

lalexgap:
I think this is pretty stale now, maybe we should just close it?

PR #1041, opened 2020-02-20
Remove Upload button on Web3Torrent UI

snario:


geoknee:
For context: I thought it would be good to remove this button as there is already a button on the homepage to take you to the upload page. And when you are on the upload page, the upload button remains very prominent but does nothing. 

Be good to get thoughts from @spersico . 

@snario looks like you might need to remove a test that is looking for this button

spersico:
**In the Upload page, before selecting any file, I agree that the Upload Button it's useless.**

But, on the download page:
image
**It allows the seeder to upload another file, and it serves to quickly let the leecher know that he also can upload files.**

On the home screen, it's true that there are 2 links to the Upload page, but I see it as having 2 call-to-action links. Considering that the design it's fairly clean and light (meaning... There's not a lot of stuff in the screen making things hard to understand). I would leave it.

What I think it would be better, is to make the menu more context sensitive, for example, if a seeder  has already uploaded a file, maybe the text could change to "_Upload Another File_" or something like that.

spersico:
About the code, it seems a fairly straightforward change.

spersico:
Another thing to consider, is where would we put the button for the Funds Status page, I think that the natural place for it would be just besides the upload button.... maybe. :-P

geoknee:
We discussed the possibility of having that information displayed on the current page, which would be another option.

geoknee:
Let's punt this for now.

PR #1047, opened 2020-02-21
[web3torrent] Add Web3TorrentChannelClient

geoknee:
Following `rps`, this client exposes a friendlier interface for the `web3torrent` app to interact with.

### DIFF
--- a/packages/web3torrent/src/clients/web3t-channel-client.ts
+++ b/packages/web3torrent/src/clients/web3t-channel-client.ts
@@ -63,10 +63,10 @@ export class Web3TorrentChannelClient {
     return this.channelClient.onMessageQueued(callback);
   }
 
-  // Accepts an rps-friendly callback, performs the necessary encoding, and subscribes to the channelClient with an appropriate, API-compliant callback
-  onChannelUpdated(rpsCallback: (channelState: ChannelState) => any) {
+  // Accepts an web3t-friendly callback, performs the necessary encoding, and subscribes to the channelClient with an appropriate, API-compliant callback
+  onChannelUpdated(web3tCallback: (channelState: ChannelState) => any) {
     function callback(channelResult: ChannelResult): any {
-      rpsCallback(convertToChannelState(channelResult));
+      web3tCallback(convertToChannelState(channelResult));
     }
     // These are two distinct events from the channel client
     // but for our purposes we can treat them the same
@@ -132,7 +132,7 @@ export class Web3TorrentChannelClient {
     this.updateChannel(channelId, aAddress, bAddress, aBal, bBal, aOutcomeAddress, bOutcomeAddress);
   }
 
-  async pushMessage(message: Message) {
+  async pushMessage(message: Message<ChannelResult>) {
     await this.channelClient.pushMessage(message);
   }
 
--- a/packages/web3torrent/src/clients/web3t-channel-client.ts
+++ b/packages/web3torrent/src/clients/web3t-channel-client.ts
@@ -5,14 +5,12 @@ interface ChannelState {
   channelId: string;
   turnNum: string;
   challengeExpirationTime;
-  aUserId: string;
-  bUserId: string;
-  aAddress: string;
-  bAddress: string;
-  aOutcomeAddress: string;
-  bOutcomeAddress: string;
-  aBal: string;
-  bBal: string;
+  seeder: string;
+  leecher: string;
+  seederOutcomeAddress: string;
+  leecherOutcomeAddress: string;
+  seederBalance: string;
+  leecherBalance: string;
 }
 
 // This class wraps the channel client converting the request/response formats to those used in the app
@@ -93,15 +91,25 @@ export class Web3TorrentChannelClient {
 
   async updateChannel(
     channelId: string,
-    aAddress: string,
-    bAddress: string,
-    aBal: string,
-    bBal: string,
-    aOutcomeAddress: string = aAddress,
-    bOutcomeAddress: string = bAddress
+    seeder: string,
+    leecher: string,
+    seederBalance: string,
+    leecherBalance: string,
+    seederOutcomeAddress: string,
+    leecherOutcomeAddress: string
   ) {
-    const allocations = formatAllocations(aOutcomeAddress, bOutcomeAddress, aBal, bBal);
-    const participants = formatParticipants(aAddress, bAddress, aOutcomeAddress, bOutcomeAddress);
+    const allocations = formatAllocations(
+      leecherOutcomeAddress,
+      seederOutcomeAddress,
+      leecherBalance,
+      seederBalance
+    );
+    const participants = formatParticipants(
+      leecher,
+      seeder,
+      leecherOutcomeAddress,
+      seederOutcomeAddress
+    );
 
     // ignore return val for now and stub out response
     const channelResult = await this.channelClient.updateChannel(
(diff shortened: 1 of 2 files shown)


packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,200 @@
...
+      channelId,
+      aAddress,
+      bAddress,
+      aBal,
+      bBal,
+      aOutcomeAddress,
+      bOutcomeAddress
+    } = channelState;
+    this.updateChannel(channelId, aAddress, bAddress, aBal, bBal, aOutcomeAddress, bOutcomeAddress);
+  }
```
geoknee: This is a new method I am proposing: it's not strictly necessary but I think is quite convenient for a unidirectional payment channel.

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+  bBal: string;
+}
+
+// This class wraps the channel client converting the request/response formats to those used in the app
+
+export class Web3TorrentChannelClient {
+  constructor(private readonly channelClient: ChannelClient) {}
+
+  mySigningAddress?: string;
+  myEthereumSelectedAddress?: string; // this state can be inspected to infer whether we need to get the user to "Connect With MetaMask" or not.
```
geoknee: If we don't have `myEthereumSelectedAddress`, we need to display some different UI and then call the getter on this class, which may cause metamask to pop up. Open for discussion about whether we want to store the information about whether this is necessary, here or elsewhere. In RPS (for example) we have an explicit state machine for the app which will not progress until we have these addresses.

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  async enable() {
```
snario: Probably doesn't need to be on the client

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+    aAddress: string,
+    bAddress: string,
+    aBal: string,
+    bBal: string,
+    aOutcomeAddress: string = aAddress,
+    bOutcomeAddress: string = bAddress
+  ): Promise<ChannelState> {
+    const participants = formatParticipants(aAddress, bAddress, aOutcomeAddress, bOutcomeAddress);
+    const allocations = formatAllocations(aOutcomeAddress, bOutcomeAddress, aBal, bBal);
+    const appDefinition = '0x0'; // TODO SingleAssetPayments address
```
snario: Maybe stick with `AddressZero` and reference the fact that this is the null app in a comment in the code

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+    this.myEthereumSelectedAddress = await this.channelClient.getEthereumSelectedAddress();
+    return this.myEthereumSelectedAddress;
+  }
+
+  onMessageQueued(callback: (message: Message) => void) {
+    return this.channelClient.onMessageQueued(callback);
+  }
+
+  // Accepts an web3t-friendly callback, performs the necessary encoding, and subscribes to the channelClient with an appropriate, API-compliant callback
+  onChannelUpdated(web3tCallback: (channelState: ChannelState) => any) {
```
snario: `void` instead of `any`?

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    return () => {
+      unsubChannelUpdated();
+      unsubChannelProposed();
+    };
+  }
+
+  async joinChannel(channelId: string) {
+    const channelResult = await this.channelClient.joinChannel(channelId);
+    return convertToChannelState(channelResult);
```
snario:
```js
    return convertToChannelState(this.channelClient.joinChannel(channelId));
```

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+      participants,
+      allocations,
+      'appData unused'
+    );
+
+    return convertToChannelState(channelResult);
+  }
+
+  // seeder may use this method to accept payments
+  async counterSign(channelState: ChannelState) {
```
snario: Perhaps rename to `acceptPayment` which defines the "action" taken by the seeder more closely.

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+      channelId,
+      aAddress,
+      bAddress,
+      aBal,
+      bBal,
+      aOutcomeAddress,
+      bOutcomeAddress
+    } = channelState;
+    this.updateChannel(channelId, aAddress, bAddress, aBal, bBal, aOutcomeAddress, bOutcomeAddress);
+  }
```
snario: Something we could do with Web3Torrent is the following:

- Use the null app
- Sender (leecher) sends state with the same `turnNum` to the Receiver (seeder) with `isFinal` set to `true`
- Receiver stores all of these inside their wallet

This gives the Receiver the option to close the channel unilaterally at any point in time by countersigning the state with `isFinal` as `true` on whichever is the most preferred to them. The client in Web3Torrent would select the most optimal one based on `aBal`/`bBal` and call a method on the wallet.

This would require the wallet to have some logic about "selecting" a state if there are multiple of them with the same turn number.

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+  challengeExpirationTime;
+  aUserId: string;
+  bUserId: string;
+  aAddress: string;
+  bAddress: string;
+  aOutcomeAddress: string;
+  bOutcomeAddress: string;
+  aBal: string;
+  bBal: string;
+}
```
snario: This could probably be a simpler data structure based on the nature of the Web3Torrent application.

I think every channel has a distinct `sender` and `receiver`, so we could use that instead of `a` and `b`. Then `aBal` and `bBal` could be stored as `amountSent` and `totalBudget` or something like that. I think `challengeExpirationTime` also can just be hard-coded for this application inside the Web3Torrent application code (i.e., in this file as a constant).

If we used the optimal strategy I mentioned, `turnNum` could always just either be `0` or `1`, too.

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -0,0 +1,193 @@
...
+  const {turnNum, channelId, participants, allocations, challengeExpirationTime} = channelResult;
+  return {
+    channelId,
+    turnNum: turnNum.toString(), // TODO: turnNum should be switched to a number (or be a string everywhere),
+    challengeExpirationTime,
+    aUserId: participants[0].participantId,
+    bUserId: participants[1].participantId,
+    aAddress: participants[0].destination,
+    bAddress: participants[1].destination,
+    aOutcomeAddress: participants[0].destination,
```
snario: `aOutcomeAddress` and `aAddress` seem to be the same thing

packages/web3torrent/src/clients/web3t-channel-client.ts
geoknee: I think for now we should target the way the wallet currently works and then switch to something more optimal later on.

packages/web3torrent/src/clients/web3t-channel-client.ts
geoknee: I think we have some freedom in whether we use the null app or the single assets payments app, but the choice will have implications on how the wallet should behave. I think this placeholder is probably ok until we decide which we want to do.

PR #1052, opened 2020-02-25
Upgrade solidity

junjizhi:
Got the contracts to compile for ~~nitro-protocol~~ all packages now.

Note that package @openzeppelin/contracts with v0.6.0 support is still in beta: 3.0.0-beta.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,6 +189,7 @@ jobs:
       USE_GANACHE_DEPLOYMENT_CACHE: true
       GANACHE_CACHE_FOLDER: ../../.ganache-deployments
       GANACHE_PORT: 8547
+      NODE_ENV: test
     steps:
       - checkout
       - puppeteer/install
--- a/packages/nitro-protocol/contracts/interfaces/IForceMove.sol
+++ b/packages/nitro-protocol/contracts/interfaces/IForceMove.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.13;
+pragma solidity ^0.6.0;
 pragma experimental ABIEncoderV2;
 
 import './ForceMoveApp.sol';
@@ -6,7 +6,7 @@ import './ForceMoveApp.sol';
 /**
   * @dev The IForceMove contract abstraction defines the interface that an implementation of ForceMove should implement. ForceMove protocol allows state channels to be adjudicated and finalized.
 */
-contract IForceMove {
+abstract contract IForceMove {
     struct Signature {
         uint8 v;
         bytes32 r;
@@ -65,7 +65,7 @@ contract IForceMove {
         Signature[] memory sigs,
         uint8[] memory whoSignedWhat,
         Signature memory challengerSig
-    ) public;
+    ) public virtual;
 
     /**
     * @notice Repsonds to an ongoing challenge registered against a state channel.
@@ -84,7 +84,7 @@ contract IForceMove {
         // variablePartAB[0] = challengeVariablePart
         // variablePartAB[1] = responseVariablePart
         Signature memory sig
-    ) public;
+    ) public virtual;
 
     /**
     * @notice Overwrites the `turnNumRecord` stored against a channel by providing a state with higher turn number, supported by a signature from each participant.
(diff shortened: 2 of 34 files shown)


.circleci/config.yml
```diff
@@ -189,6 +189,7 @@ jobs:
       USE_GANACHE_DEPLOYMENT_CACHE: true
       GANACHE_CACHE_FOLDER: ../../.ganache-deployments
       GANACHE_PORT: 8547
+      NODE_ENV: test
```
kerzhner: I don't think this is needed since `NODE_ENV` is set to `test` in CircleCI project environment variables: https://circleci.com/gh/statechannels/monorepo/edit#env-vars

packages/nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -1,12 +1,12 @@
...
+pragma solidity ^0.6.0;
 pragma experimental ABIEncoderV2;
 
 import './ForceMoveApp.sol';
 
 /**
   * @dev The IForceMove contract abstraction defines the interface that an implementation of ForceMove should implement. ForceMove protocol allows state channels to be adjudicated and finalized.
 */
-contract IForceMove {
+abstract contract IForceMove {
```
kerzhner: Nice!

patches/@counterfactual+cf-adjudicator-contracts+0.0.11.patch
```diff
@@ -4,7 +4,7 @@ index d823033..31595b9 100644
 +++ b/node_modules/@counterfactual/cf-adjudicator-contracts/contracts/interfaces/CounterfactualApp.sol
 @@ -1,4 +1,4 @@
 -pragma solidity 0.5.12;
-+pragma solidity ^0.5.13;
```
kerzhner: Not sure that this will work since the `.patch` file seems to include commit references. This patch might need to be regenerated. Or maybe not...

junjizhi:
Strange. The rps tests passed locally but failed in CI.

.circleci/config.yml
junjizhi: I added this because for anyone who in the future forks this repo will be tripped by this CircleCI settings like I did. 

Ideally,. all vars should be in this file (except maybe secrets?) so nobody needs to tinker with CircleCI. We have lots of env vars in `config.yml` anyway, why not add one extra.

kerzhner: @junjizhi you bring up a good point. I am ok defining env vars in `config.yml` with a couple caveats:
- An env variable should be defined either in `config.yml` or in the CircleCI admin console. But not in both.
- If we are moving away from defining env variables in CircleCI admin console (except for secrets), we should create reasonable defaults that most jobs inherit. A quick scan of `config.yml` shows that we might need to do a bit of clean up with env vars and override defaults only when needed.

junjizhi: @kerzhner I totally agree with the above two points. Having the setting in both places is bad for maintenance and newcomers. 

I don't have access to CircleCI admin, so I couldn't see env var settings there, so no way for me to compare with `config.yml` and tell 1) what are the defaults, 2) what gets overridden there. The cleanup may have to be done in a separate ticket.

packages/nitro-protocol/contracts/interfaces/IForceMove.sol
junjizhi: Solidity 0.6.0 has stricter checks on inheritance, so this is a must now. The compiler deserves the credit 😛

patches/@counterfactual+cf-adjudicator-contracts+0.0.11.patch
junjizhi: Seems to work fine for me. I could see that they archived `@counterfactual/cf-adjudicator-contracts` repo and stopped publishing new updates, so we had to patch it like this 🤷‍♂

.circleci/config.yml
kerzhner: Sounds excellent, filed https://github.com/statechannels/monorepo/issues/1054

junjizhi:
Finally got CI to pass. @snario @geoknee Can you guys approve again?

kerzhner:
@junjizhi fyi, we typically require just one approver unless you feel that a specific code owner should check off on a set of changes.

junjizhi:
@kerzhner Got it. I'm not authorized to merge. Can you hit the button for me? Thanks.

Here's what I see:
image

kerzhner:
@junjizhi Done. Appreciate the good work!

PR #1057, opened 2020-02-27
Set circleci default resource class to large

kerzhner:
Hopefully this will address the out of memory errors on CircleCI.


andrewgordstewart:
- Is there just one type of job that's failing? Can we set the resource class just on that job?
- Is it because there's too much concurrency? Should we limit that?
- Can we log to an artifact in the jobs with memory issues?

kerzhner:
Closing in favor of https://github.com/statechannels/monorepo/pull/1059.

PR #1058, opened 2020-02-27
[web3t] Add channel-client api calls

geoknee:
This is super patchy so far but just raising the PR so other interested parties can comment!


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -25,18 +30,37 @@ export * from './types';
...
+
+// TODO: Put inside better place than here where app can handle error case
+window.channelProvider.enable(process.env.REACT_APP_WALLET_URL);
+
 export default class WebTorrentPaidStreamingClient extends WebTorrent {
   allowedPeers: PeersByTorrent;
   pseAccount: string;
   torrents: PaidStreamingTorrent[] = [];
+  channelClient: Web3TorrentChannelClientInterface;
+  amSeeder: boolean = false;
```
geoknee: This is probably a bad idea in the long term since I can become a seeder at any point.

- The idea of a leecher being "not a seeder" is not really correct.
- We still want the "downloading peer" to open the channel during the opening handshake, if we can deduce who that is.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -25,18 +30,37 @@ export * from './types';
...
   constructor(opts: WebTorrentPaidStreamingClientOptions = {}) {
     super(opts);
+    this.channelClient = new Web3TorrentChannelClient(new ChannelClient(window.channelProvider));
     this.allowedPeers = {};
     this.pseAccount = opts.pseAccount || Math.floor(Math.random() * 99999999999999999).toString();
     log('ACCOUNT ID: ', this.pseAccount);
   }
 
+  async enable() {
+    this.pseAccount = await this.channelClient.getAddress();
```
geoknee: I'm pretty sure we want to use the wallet address for the torrenting account ids, and not to introduce another system.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -158,29 +183,67 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       wire.emit(PaidStreamingExtensionEvents.REQUEST, peerAccount);
     });
 
+    wire.paidStreamingExtension.once(PaidStreamingExtensionEvents.PSE_HANDSHAKE, async () => {
+      const peerAccount = wire.paidStreamingExtension && wire.paidStreamingExtension.peerAccount;
+      const pseAccount = wire.paidStreamingExtension && wire.paidStreamingExtension.pseAccount;
+
+      if (!this.amSeeder) {
```
geoknee: Is there a better way to determine if I should create the channel?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -158,29 +183,67 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-    }, 500);
+    const channel = await this.channelClient.createChannel(
+      '0x0000000000000000000000000000000000000000',
+      '0x0000000000000000000000000000000000000000',
+      '0x00',
+      '0x00',
+      '0x0000000000000000000000000000000000000000',
+      '0x0000000000000000000000000000000000000000'
+    );
+
```
geoknee: This doesn't really belong here, but we need the `channel` (in particular `channel.channelId` for the `updateChannel` call below (which I think *does* belong here).

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -158,29 +183,67 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
     // [ George ] Here the seeder can countersign the state update by an appropriate channelClient.updateChannel(), and pull the updated state channel balance off the ChannelResult before updating this.allowedPeers[infoHash][peerId].funds accordingly.
 
     // [ George ] If web3torrent is to run the Single Asset Payments ForceMoveApp, and the payments are going to be unidirectional we could wrap the channelClient in a web3tChannelClient which offers a countersign() convenience method. This allows the seeder to immediately accept the payment and for the leecher to be ready to send another one as quickly as possible.
 
-    // [ George ] NB all channel client methods are async so we would want to await them before continuing. That means a bunch of methods in this class will also need to be aysnc.
+    // [ George ] NB all channel channelClient methods are async so we would want to await them before continuing. That means a bunch of methods in this class will also need to be aysnc.
 
     const {funds} = this.allowedPeers[infoHash][peerId];
     this.allowedPeers[infoHash][peerId].funds = (Number(funds) + Number(paymentHash)).toString();
```
geoknee: We want to replace this with a proper, robust way of inferring a legitimate payment has been made. In `rps`, we listen to for `channelUpdated` events and mirror some of that information to the redux store, (which is then e.g. inspected by the UI).

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: It keeps going back to this, isn't it?

The problem that I keep finding is that a peer can behave as a seeder in a moment, and as a leecher in the next, with another peer. As the wire is a duplex connection, request for parts of files can come from both peers.

I thought about two options to handle this:
- We create the channel, and make the peers join **not** on the handshake, but when the first request is received. That way, we have certainty about who's gonna fill which role.
We could do that on: https://github.com/statechannels/monorepo/blob/89202f6ae7e8bb8d96e0a41c771f446930b62874/packages/web3torrent/src/library/web3torrent-lib.ts#L129


- Another thing that I think it could be possible, depending on the channel's support, would be to drop all pretences of differentiating between seeders and leechers, and making that the balance can be positive or negative, or even maintaining 2 balances per channel, where each peer A checks the balance of peer B when a request is made, and vice-versa.
That way, it doesn't really matter who's the seeder, and who's the leecher, both peers join the channel in equal conditions.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: Completely agree on this

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: Been testing what would be the best way of handling this.
I think that on
https://github.com/statechannels/monorepo/blob/89202f6ae7e8bb8d96e0a41c771f446930b62874/packages/web3torrent/src/library/web3torrent-lib.ts#L115
when we set the pseAccount, we could pass other variables, including something like "_seederMode_". That way, each peer will let the other know what role is having (**seeder** or **leecher**)
Though, as I commented before, I think this doesn't consider the duplex nature of wires, nor the role changing nature of peers.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: I would have to test and figure out if the peer balance check can be done asynchronously (f.e a check to the channel client)

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -25,18 +30,37 @@ export * from './types';
 
 export const REQUEST_RATE = 10;
 
+if (process.env.REACT_APP_FAKE_CHANNEL_PROVIDER === 'true') {
```
spersico: I think that the WebTorrentClient and the Web3TorrentChannelClient should be a little more independent of each other.
That way, Web3TorrentChannelClient could be another React Context Provider, and at the same time, be passed to the WebTorrentPaidStreamingClient as a constructor argument.

I know that this is a draft PR, so I don't want to judge too hard :-P

spersico: That way, all these checks could be done in the Web3TorrentChannelClient, at that client creation.

packages/web3torrent/src/library/web3torrent-lib.ts
geoknee: I like your first suggestion!

I think your suggestion would be quite elegant and match the true "peer-to-peer" nature of bittorrent, but  would be much more difficult given that ForceMove channels are turn-based. The way we currently handle the setup phase of a channel is really forcing us into having one proposer and one confirmer. 

As you say, whomever sends the first request can propose the channel. Otherwise, I can always accept channels that are being proposed (possible DoS vector here). 

If two peers are seeding and leeching from each other (which I think is possible), we can just have two unidirectional payment channels for that.

packages/web3torrent/src/library/web3torrent-lib.ts
geoknee: I'm working on getting this implemented 👍

geoknee:
Mar-03-2020 01-08-05
This screencast shows the following steps (using the `FakeChannelProvider`):
- seeder seeds / registers with tracker
- magnet link pasted + download button clicked -> leecher discovers seeder and makes a request
- on the first request, the seeder creates a channel with an opening balance that pays himself a small amount, and chokes the leecher
- when the Post FundSetup arrives from leecher, the seeder interprets this as a payment and unchokes the leecher
- the torrenting commences
- the leecher is re-choked because the initial payment only covered a small number of bytes
- the leecher notices this, and sends a payment
- is unchoked
- repeat
- the leecher runs out of money in the channel
- torrenting stops

TODO 

- [ ] communicate the `outcomeAddress` / `ethereumSelectedAddress` of leecher to seeder
- [ ] test with multiple leechers
- [ ] handle closing the channel
- [ ] general cleanup / insert some comments

tomclose:
This looks good. A minor point: it's important that the opening balances of the channel (in the pre-fund setup) match the amounts that will be deposited, as the wallet uses these to decide how much to deposit. The seeder needs to wait until _after_ the post-fund setup to request the first payment.

geoknee:
Good point, should be an easy fix though.

geoknee:
> The seeder needs to wait until _after_ the post-fund setup to request the first payment.

See 5615e81

packages/channel-client/src/types.ts
```diff
@@ -48,6 +48,13 @@ export interface ChannelClientInterface<Payload = object> {
   closeChannel: (channelId: string) => Promise<ChannelResult>;
   getAddress: () => Promise<string>;
   getEthereumSelectedAddress: () => Promise<string>;
+  approveBudgetAndFund(
+    playerAmount: string,
+    hubAmount: string,
+    playerDestinationAddress: string,
+    hubAddress: string,
+    hubDestinationAddress: string
```
snario: Where is the documentation on this method? I'm confused by `hubAddress` is necessary.

packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -41,6 +42,9 @@ export class FakeChannelProvider implements ChannelProviderInterface {
       case 'PushMessage':
         return this.pushMessage(params) as Promise<MethodType[K]>;
 
+      case 'GetEthereumSelectedAddress':
+        return new Promise(() => '0xEthereumSelectedAddress') as Promise<MethodType[K]>;
```
snario: Couldn't this just be `return '0xEthereumSelectedAddress'`?

packages/web3torrent/README.md
```diff
@@ -5,17 +5,19 @@
...
 
 ### The client
+
 As the client is a custom, extended version of [Webtorrent](https://github.com/webtorrent/webtorrent), most of the docs from [their docs page](https://webtorrent.io/docs) are applicable on this project.
 
 **There is a couple of extra properties and methods to take into account, the rest you can assume are pretty much the same**:
 
-* `pseAccount` it's an ID, wich identifies the client (peer). This can be set when the client is instantiated (by passing an object with the key `pseAccount`), or a random number is assigned.
-* `allowedPeers` holds a list of torrents (identified by infoHash), in wich every torrent has a list of allowed peers (identified by pseAccount)
+- `pseAccount` it's an ID, wich identifies the client (peer). This can be set when the client is instantiated (by passing an object with the key `pseAccount`), or a random number is assigned.
```
snario: This is now the address of the wallet, and is no longer a random number

packages/web3torrent/src/clients/web3t-channel-client.ts
```diff
@@ -146,6 +252,9 @@ export class Web3TorrentChannelClient {
 
   async pushMessage(message: Message<ChannelResult>) {
     await this.channelClient.pushMessage(message);
+    const channelResult: ChannelResult = message.data;
+    const channelId = channelResult.channelId;
+    this.openChannels = {...this.openChannels, [channelId]: convertToChannelState(channelResult)};
```
snario: Seems kind of odd to me. Wouldn't `pushMessage` need to cause the wallet to evaluate the message and then emit `ChannelUpdated`?

packages/channel-client/src/types.ts
geoknee: This change is just mirroring the method on the `ChannelClient` class to the interface for that class. Maybe a question for @lalexgap ?

packages/web3torrent/src/library/client-seeding.test.ts
```diff
@@ -88,7 +82,7 @@ describe('Seeding and Leeching', () => {
     });
   }, 10000);
 
-  it('should be able to unchoke and finish a download', done => {
+  it.skip('should be able to unchoke and finish a download', done => {
```
geoknee: I gave up getting this to work, for now. We might want to write an entirely new suite of tests that work well with the channel client.

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -2,11 +2,15 @@ import React from 'react';
 import {WebTorrentAddInput, WebTorrentSeedInput} from '../library/types';
 import WebTorrentPaidStreamingClient from '../library/web3torrent-lib';
 import {Status, Torrent} from '../types';
+import {web3TorrentChannelClient} from './web3t-channel-client';
+
+export const web3torrent = new WebTorrentPaidStreamingClient({
+  channelClient: web3TorrentChannelClient
```
spersico: :+1:

packages/web3torrent/src/library/types.ts
```diff
@@ -121,6 +121,7 @@ export type PeerByTorrent = {
   wire: PaidStreamingWire | PeerWire;
   allowed: boolean;
   funds: string;
+  seederBalance: string;
```
spersico: What's the difference between funds and seederBalance? Is this your way of having 2 sources of truth?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -124,92 +138,143 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
           break;
-        case PaidStreamingExtensionNotices.PAYMENT:
-          this.loadFunds(torrent.infoHash, wire.peerExtendedHandshake.pseAccount, data.hash);
-          this.unblockPeer(torrent.infoHash, wire, wire.peerExtendedHandshake.pseAccount);
+        case PaidStreamingExtensionNotices.MESSAGE:
+          await this.channelClient.pushMessage(JSON.parse(data.message));
+          turnNum = Number(JSON.parse(data.message).data.turnNum);
+          channelId = JSON.parse(data.message).data.channelId;
+          if (
+            JSON.parse(data.message).recipient === this.pseAccount &&
```
spersico: I think I read something about the turns that you guys use to handle payment channels. But is it relevant here?
In any case, couldn't this be something handled inside the channelClient? There's a lot of what it seems like channel-related logic here.

geoknee: Yes I think we can eventually hide a lot of this in the channel client.

packages/web3torrent/src/library/types.ts
geoknee: It might seem that way, and it's on my todo list to reduce potential confusion here.

Funds can go up and down. Seeder balance will typically only ever increase. Funds could be renamed to `buffer` or something like that.

spersico: okay! :-)

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: Yes

packages/web3torrent/src/clients/web3t-channel-client.ts
geoknee: Yes that is what happens: but the currently the class does not update the `openChannels`(aka `channelCache` when `ChannelUpdated` is emitted from the wallet.

The pattern that I will stick to (for now) is to update the cache with whatever the wallet _returns_, rather than with something it emits. It seemed weird to try and bind to those events inside this class.

PR #1061, opened 2020-02-28
Expose Chain directly on the Store, instead of proxying all methods

tomclose:
Currently methods from the `Chain` are made available on the `Store` through proxying - for each `Chain` method, there is a method with the same name and signature on the `Store`, which calls directly through to the corresponding `Chain` method.

In separate work, I've just added three new methods to the `Chain`, and it seems a bit sub-optimal to have to add each of these to the `Store` separately. There's also a question of whether the `Store` should be providing the `Chain` at all, or whether the `Chain` should be independently passed in to any protocols that require it; the `Store` itself doesn't currently call the `Chain` directly, other than to proxy the calls through.

This PR removes the proxied methods, instead exposing a `chain` method directly on the `Store`. This should make it easier to modify the `Chain` interface independently in the short term, while also making it (very slightly) easier to transition to passing the `Chain` into protocols separately at a later date, if required.

### DIFF
--- a/packages/xstate-wallet/src/store/memory-store.ts
+++ b/packages/xstate-wallet/src/store/memory-store.ts
@@ -77,14 +77,12 @@ export interface Store {
   // But I need one, in order to implement virtual funding.
   addObjective(objective: Objective): void;
 
-  // TODO: Should this be part of the store?
-  getChainInfo: Chain['getChainInfo'];
-  chainUpdatedFeed: Chain['chainUpdatedFeed'];
-  deposit: Chain['deposit'];
+  // TODO: should this be exposed via the Store?
+  chain: Chain;
 }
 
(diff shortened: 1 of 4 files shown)


packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -77,14 +77,12 @@ export interface Store {
...
-  getChainInfo: Chain['getChainInfo'];
-  chainUpdatedFeed: Chain['chainUpdatedFeed'];
-  deposit: Chain['deposit'];
+  // TODO: should this be exposed via the Store?
+  chain: Chain;
 }
 
 export class MemoryStore implements Store {
-  protected _chain: Chain;
+  public chain: Chain;
```
lalexgap: Could this be `readOnly`?

PR #1065, opened 2020-02-28
Throw on contract compilation errors

andrewgordstewart:
My `yarn prepare` script was passing, even when contracts weren't compiling due to an incorrect solidity version. This PR throws on compile error rather than logging.


kerzhner:
A good example where monads might be helpful.

PR #1066, opened 2020-02-29
Switch from BigNumber to Strings

lalexgap:
To prevent issues with storing and sending `BigNumber`s across the wire we've decided to use strings instead of `BigNumber`s.


packages/xstate-wallet/src/store/types.ts
```diff
@@ -1,4 +1,5 @@
-import {BigNumber} from 'ethers/utils';
+//This is used to identify strings that are the hex values of BN
+export type HexNumberString = string;
```
andrewgordstewart: We could create a new type that extends `String` and get stricter type checking:
```
interface HexString extends String {
  type: 'irrelevant';
}

function isHex(t: string | HexString): t is HexString {
  return t.startsWith('0x');
}

const foo = '0xf00';

function sayFoo(f: HexString) {
  console.log(f);
}

if (isHex(foo)) {
  sayFoo(foo);
} else {
  // type error
  sayFoo(foo);
}
```

lalexgap: I think this means we would need to use the type guard everywhere we return a string? Otherwise we get type errors.
```
function toHex(value: BigNumber): HexNumberString {
const hexValue = value.toHexString();
  return isHex(hexValue) ? hexValue : 0;
}
```

andrewgordstewart: Yes, but ideally you would only be passing around, and manipulating, `HexString` values. So, you would validate the string once, then the math helpers would maintain this type throughout the code

tomclose: This would be a way of ensuring that we don’t lose typesafety. Another approach would be to convert it to a `BigNumber` at the point of validation ...

lalexgap: Updated d01aeec0cb18538aea3e916d1722667a8268ae5f

lalexgap:
Closing in favour of #1071

PR #1069, opened 2020-02-29
[rps,web3t] Handle no metamask in Connection Banner

geoknee:
Screenshot 2020-02-29 at 11 46 46


spersico:
This is what web3torrent looks in firefox (with no extension) 
image
Could we move the banner on top of the page in web3torrent? In between the nav bar and the body it looks kind of weird.

geoknee:
I had avoided putting it there as it ended up being hidden from view. But I fixed that by scrolling the page up automatically: 

Screenshot 2020-03-02 at 11 26 15

geoknee:
(final?) fix in 392982a removes the need for auto-scrolling

PR #1071, opened 2020-03-01
Xstate wire serialization-deserialization

tomclose:
This PR adds serialization/deserialization utilities for the wire format to the xstate-wallet.

It represents a different approach to https://github.com/statechannels/monorepo/pull/1066: by serializing/deserializing at the boundary, we're free to use objects like `BigNumber` in our states internally. This will enable us to use richer objects (e.g. an `Outcome` class might be useful) internally going forward.


lalexgap:
Closing in favour of #1074

PR #1072, opened 2020-03-01
Change outcome format in xstate wallet

tomclose:
This PR changes the outcome format in the xstate wallet. In particular it:

1. Switches from storing the `token` to storing the `assetHolderAddress`
2. Consolidates `Simple{Eth,Token}{Allocation,Guarantee}` into `Simple{Allocation,Guarantee}`

The reasoning behind 1 is to remove ambiguity if multiple asset holders for the same token were ever deployed. To account for this scenario, we need to send outcomes containing `assetHolderAddress` over the wire between wallets, and to store the `assetHolderAddress` internally. We still want the client to only care about `token`, which is possible by storing a look-up table and converting between the two formats internally.

Note: in the long term, if we want to support multiple asset holders for a single token, we'll need to make the asset holder lookup token-dependent. In the short term, wallets can only store one asset holder per token, and an error will be thrown (currently at the point where the wallet first tries to sent a state to the app) if a non-recognised asset holder is sent. We should probably move this check to point where the state first arrives..


packages/xstate-wallet/src/utils/outcome.ts
```diff
@@ -1,51 +1,14 @@
...
-export const isMixedAllocation = outcomeGuard<MixedAllocation>('MixedAllocation');
-
-export const simpleEthAllocation = (...allocationItems: AllocationItem[]): SimpleEthAllocation => ({
-  type: 'SimpleEthAllocation',
+import {AllocationItem, SimpleAllocation, Outcome} from '../store/types';
+import {ETH_ASSET_HOLDER_ADDRESS} from '../constants';
+
+export function isSimpleEthAllocation(outcome: Outcome): outcome is SimpleAllocation {
+  return (
+    outcome.type === 'SimpleAllocation' && outcome.assetHolderAddress === ETH_ASSET_HOLDER_ADDRESS
```
andrewgordstewart: What was the benefit of removing the `SimpleEthAllocation` type and replacing it with this check? It seems like we still have a union type (eth allocations are fundamentally different from token allocations, since depositing is different) except that the compiler doesn't know about the union type (this type guard, `isSimpleEthAllocation`, does not assert that the input is a `SimpleEthAllocation`, but instead is a `SimpleAllocation`)

andrewgordstewart: This would make a bit more sense if this type guard checked the _token address_, but the fact that `constants.ts` exports a `ETH_ASSET_HOLDER_ADDRESS` muddles things a bit -- the intention of this PR is that there might not be a unique such address.

packages/xstate-wallet/src/constants.ts
```diff
@@ -10,3 +12,19 @@ export const CHALLENGE_DURATION = 0x12c; // 5 minutes
 export const NETWORK_ID = process.env.CHAIN_NETWORK_ID || '0';
 
 export const ETH_TOKEN = '0x0';
+
+export function assetHolderAddress(tokenAddress: string): string | undefined {
+  if (bigNumberify(tokenAddress).isZero()) {
+    return ETH_ASSET_HOLDER_ADDRESS;
+  }
+  // TODO: store mapping, implement lookup
```
andrewgordstewart: How is this supposed to work when we are supporting multiple asset holders for the token?

tomclose: The mapping needn't be one-to-one. We could store a mapping from asset holder address to token address, and a mapping from token address to the default asset holder address that our wallet would propose for that token.

This said, I'm not sure (?) if it makes sense to ever support multiple asset holders for a single token within a wallet. The overriding motivation in switching to `assetHolderAddress` is not to support multiple holders per token, but to be able to quickly detect if two wallets are pointing at different asset holders.

Thinking about it now, there isn't actually a danger of two wallets getting to the point where they deposit into different asset holders, as if they use different asset holder addresses, the state hashes won't match, and it will look like the state is not correctly signed. Regardless of this, it seems good to send the asset holder address in the inter-wallet states, as it (a) keeps the property that the state contains all the information required to verify the signature, and (b) it allows the different asset holder situation to be easily detected.

packages/xstate-wallet/src/utils/outcome.ts
tomclose: The intention really was to use `assetHolderAddress` instead of `token` in the wire format, and therefore also internally in the wallet. Once this change happens, there's no difference internally* between `SimpleEthAllocation` and `SimpleTokenAllocation`, so I decided to unify these types into `SimpleAllocation`. 

While making this change, I noticed that the protocols were using `isSimpleEthAllocation` to actually restrict the supported token type to eth. I didn't want to change that behaviour as part of the PR, so I ported the method across to work with the new formats. The intention is that `isSimpleEthAllocation` won't exist in the long-term, once we allow the wallet to support other tokens.

\* As you point out, there _are_ differences when it comes to interacting with the blockchain, My assumption was that these differences only manifest in the deposit transaction sent by the `Chain`, so it can do a last-minute switch on the asset holder address to accomplish this.

andrewgordstewart:
so I think I understand the point of this PR now: to ensure that the state hash is a pure function of the data sent across the wire.

PR #1073, opened 2020-03-01
Get RPS working

lalexgap:
Various fixes to get the xstate-wallet working with RPS. Supports funding and playing a game of RPS. Concluding and challenging to be handled in separate PRs.

Some caveats:
- ~The UI needs to be cleaned up, right now there's not a lot of relevant info to be found.~ I've done some minor clean up and created #1077
- ~There appears to be various race conditions during funding/updating that can cause RPS to get stuck.~ This was caused by the wallet sending out `ChannelUpdated` when the `xstate-wallet` receives a state from the player. I've left this behaviour as is and changed `rps` to handle this. 
- The change above broke the auto player test. I've just commented it out with a TODO for now.


packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -293,11 +296,17 @@ export const applicationWorkflow = (
...
+    spawnObservers: assign<ChannelIdExists>((context: WorkflowContext & ChannelIdExists) => {
+      if (!context.requestObserver || !context.updateObserver) {
+        return {
+          ...context,
+          updateObserver: spawn(notifyOnUpdate(context)),
+          requestObserver: spawn(notifyOnChannelRequest(context))
+        };
+      } else {
+        return context;
+      }
```
andrewgordstewart: IMO, this would be more clear as
```
    spawnObservers: assign<ChannelIdExists>(context => ({
      ...context,	
      updateObserver: context.updateObserver || spawn(notifyOnUpdate(context)),	
      requestObserver: context.requestObserver || spawn(notifyOnChannelRequest(context))	
    })),	
```

PR #1076, opened 2020-03-03
Unmock Outcome address & channelID

spersico:
> 1. Get `outcomeAddress` exchanged in the handshake

Adds the peerOutcomeAddress to the handshake, and unmocks the corresponding argument in the createChannel method call.

> 6. Add a `wire.channelId` property so that transferFunds is able to reference the correct channel for making a payment

Adds the peerChannelId to the STOP notice sent by the seeder, so that the leecher gathers and persists the channelId in it's end of the wire, unmocks the corresponding argument in the makePayment method call, and persists the channelId in the peersList object.


spersico:
@geoknee I just tested that:
It seems that there's some sort of issue with the payment when 2 leechers try to connect at the same time.
image
When the leechers go one after the other (One starts the download, after the other finished it, it worked, but when trying to download at the same time it throws that error).

Edit: I tested this again, with a smaller file, and it threw a different error (but now both peers were able to download, with the second kind of... waiting for the first one to finish :-P) :
image

packages/web3torrent/src/components/torrent-info/upload-info/UploadInfo.stories.tsx
```diff
@@ -27,7 +27,8 @@ storiesOf('Web3Torrent', module)
             },
             funds: '50',
             allowed: true,
-            seederBalance: '50'
+            seederBalance: '50',
+            channelId: '0x00012319283712'
```
snario: In general I like mock data to mock the structure of the real data. In this case it would be a length 42 string.. maybe use `0x0000000000000000000000000000001231927371`

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -14,13 +14,17 @@ export abstract class PaidStreamingExtension implements Extension {
...
   protected messageBus: EventEmitter;
   protected pseId: string = '';
+  protected pseOutcomeAddress: string = '';
 
   get name(): 'paidStreamingExtension' {
     return 'paidStreamingExtension';
   }
 
   peerAccount?: string;
+  peerOutcomeAddress?: string;
```
snario: Why do we need `?` if there is a default value  of `''`?

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -61,8 +74,13 @@ export abstract class PaidStreamingExtension implements Extension {
       this.peerAccount = handshake.pseAccount.toString();
     }
 
+    if (handshake.outcomeAddress) {
+      this.peerOutcomeAddress = handshake.outcomeAddress.toString();
```
snario: Why is `toString` necessary?

spersico: Because the fields in the handshake are not simple strings.
AFAIK, the fields/props are bencoded, or encoded in some way.
Without the .toString() that handshake.outcomeAddress looks like this:
```
Uint8Array(42) [48, 120, 100, 54, 101, 50, 54, 100, 56, 97, 99, 102, 100, 50, 57, 52, 56, 99, 48, 54, 48, 57, 56, 99, 54, 100, 101, 51, 56, 54, 99, 56, 57, 98, 49, 50, 101, 48, 102, 57, 49, 54]
```

packages/web3torrent/src/library/paid-streaming-extension.ts
spersico: If the default value that you are referring to is for the `pseOutcomeAddress`, which is set at the wire creation, not on the extended handshake.
I tried to respect the way the props were being handled previously.
The idea is that both `peerAccount` and `peerOutcomeAddress` are initially undefined because they are not set at the wire setup, while `pseId` and `pseOutcomeAddress` are defaulted to '', because they are set at the wire setup.

packages/web3torrent/src/components/torrent-info/upload-info/UploadInfo.stories.tsx
spersico: okay! fixed

PR #1079, opened 2020-03-04
Add a GetState ChannelClient API Method

snario:


geoknee:
Fixes #1096 .

geoknee:
Do you think we should use this method in `web3torrent`, possibly instead of caching states in the `PaymentChannelClient`?

geoknee:
I guess we _do_ want to do that, since we can then benefit from the superior persistence mechanisms the wallet has (i.e.the  IndexedDB store), and recover from refreshes / crashes.

PR #1084, opened 2020-03-04
Update web3torrent calls

junjizhi:
I renamed the method capitalized `CloseAndWithdraw` method to follow conventions with the rest.

Some methods still ignore the return values from channel-clients, e.g., 
- ~~acceptPayment~~
- approveBudgetAndFund

It's unclear whether we should take their return values and add them to cached states.


geoknee:
Perhaps the `acceptPayment` method should  just return the `Promise` that it gets from `.this.updateChannel`, rather than `await`ing it. We don't need to cache again because that is done inside of `updateChannel`.

junjizhi:
@geoknee I think the current implementation of `acceptPayment` is fine. It is an async function and returning a promise anyway.

PR #1086, opened 2020-03-04
Move event conversion to MessagingService

tomclose:
Previously we converted from the `client-api-schema` types in several places: the `MessagingService`, the `ChannelWallet`, and the `ApplicationWorkflow`. This PR:

1. Pulls out the internal event types (previously defined in the `ApplicationWorkflow`) out into their own file.
2. Moves all conversion from `client-api-schema` types to internal events into the `MessagingService`

### DIFF
--- a/packages/xstate-wallet/src/event-types.ts
+++ b/packages/xstate-wallet/src/event-types.ts
@@ -40,9 +40,11 @@ export interface PlayerRequestConclude {
   channelId: string;
 }
 
-export type WorkflowEvent =
+export type AppRequestEvent =
   | PlayerRequestConclude
   | PlayerStateUpdate
   | OpenEvent
   | ChannelUpdated
   | JoinChannelEvent;
+
+export type WorkflowEvent = AppRequestEvent;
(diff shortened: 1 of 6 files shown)


packages/xstate-wallet/src/event-types.ts
```diff
@@ -0,0 +1,50 @@
+import {SimpleAllocation, Participant} from './store/types';
```
lalexgap: I'm wondering if we should silo these to workflow. IE: Create an `application` folder containing `workflow.ts` and `events.ts`?

tomclose: Right now, I'm thinking of these events as "everything that is passed out of the messaging service". The plan is to add events like `ENABLE_ETHEREUM` to this collection. I thought by putting them at the top level, we'd be able to collect all of these events together. Given that there aren't very many of them, I think this should be ok for now, but prepared to revisit in the future. I think it would probably(?) make sense to scope UI events by protocol (but then again, there won't be very many of these either, so it might be just as easy to have all events in one place).

PR #1088, opened 2020-03-04
Move guarantor creation onto leaf workflow

andrewgordstewart:
Prior, the hub constructed the guarantor channels. This was convenient when writing the workflow, but is incompatible with a "simple hub", which only signs incoming states. Since we are focusing on getting web3torrent to work with the xstate-wallet, this PR switches the responsibility onto the leaves.

It became a bit case-driven to construct the two workflows in a single file (eg. `getDeductions`, and the `deductions` property stored on context, depends on if you're a hub or a leaf). Thus, I split it into two separate files, but still re-used some code.

Workflow as hub: https://xstate.js.org/viz/?gist=b730b04dc521ca40f83ea13b9e1fbcf3
Workflow as leaf: https://xstate.js.org/viz/?gist=02bdbdedc9a541127175c5767331b374


packages/xstate-wallet/src/workflows/virtual-funding-as-hub.ts
```diff
@@ -0,0 +1,168 @@
...
+          onDone: 'done'
+        }
+      },
+      done: {type: 'final'}
+    }
+  };
+};
+
+export const config: MachineConfig<Init, any, any> = {
+  key: 'virtual-funding',
```
kerzhner:
```suggestion
  key: 'virtual-funding-as-hub',
```

packages/xstate-wallet/src/workflows/virtual-funding-as-leaf.ts
```diff
@@ -70,85 +71,51 @@ const enum States {
...
-        }
-      };
-  }
-
-  return {...config, onDone: States.fundTargetChannel};
-};
-
-const generateConfig = (role: Role): MachineConfig<Init, any, any> => ({
+export const config: StateNodeConfig<Init, any, any> = {
   key: 'virtual-funding',
```
kerzhner:
```suggestion
  key: 'virtual-funding-as-leaf',
```

packages/xstate-wallet/src/workflows/virtual-funding-as-hub.ts
```diff
@@ -0,0 +1,168 @@
...
+
+type TEvent = AnyEventObject;
+
+const enum Actions {
+  assignDeductions = 'assignDeductions',
+  watchObjectives = 'watchObjectives'
+}
+
+const enum Services {
+  getDeductions = 'getDeductions',
```
kerzhner: Not sure I fully understand what deductions are. What is the benefit over current allocation items + desired allocation items?

tomclose:
@andrewgordstewart what's the status on this? Can it be merged?

packages/xstate-wallet/src/workflows/tests/dumb-hub.ts
```diff
@@ -0,0 +1,34 @@
...
+  public get outboxFeed(): Observable<Message> {
+    return fromEvent(this._eventEmitter, 'addToOutbox');
+  }
+
+  public async pushMessage({signedStates}: Message) {
+    signedStates?.map(signedState => {
+      const signature = signState(signedState, this.privateKey);
+      const {signatures, participants} = signedState;
+      const hubIdx = participants.findIndex(p => p.signingAddress === this.getAddress());
+      signatures[hubIdx] = signature;
```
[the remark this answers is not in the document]
andrewgordstewart: No, in the `MemoryChannelStoreEntry` stores signatures indexed by the participant index.

PR #1102, opened 2020-03-05
[WIP] XState-Wallet Store IndexedDB replacement

spersico:
So, the idea, is to replace the in-memory store with a indexedDB replacement.
Tom proposed that it could be a drop-in replacement:
> e.g. have something like memoryStore = new Store(new MemoryBackend()) vs indexedDBStore = new Store(new IndexedDbBackend()) or whatever

## Comments
We've got to remember that indexedDB is a DB that will be saved on disk, and as such, it handles objects, not classes with functions and state. So classes and complex objects (like BigNumbers) need to be parsed/transformed if we want to save them so we can get a more object-like element to save into the DB.

## The Pathway is:
- [x] Make a DBBackend interface, that both `MemoryBackend` and `IndexedDBBackend` have to implement.
- [x] Implement a `MemoryBackend` class, and move the internal props of the store, with async getters and setters.
- [x] Make all get and set to those internal props async methods to its internal DB object (by default the `MemoryBackend`). The idea is that when/if the `MemoryBackend` gets replaced by IndexedDBBackend it has the same behaviour.
- [x] Implement an `IndexedDBBackend` class, which has to be functionally equivalent to the `MemoryBackend`.
- [x] Deal with ~stuff like the app not working and~ tests not passing.
- [ ] Finish the integration of the new store, ensure tests pass and merge.

## Status
- **To remember:**
 I had to change the API of the store for a couple of events to make them async-able. No way around this.
- **All Store tests pass now!:**  
The last bugs that I encountered, were side effects not being saved, or stuff that wasn't correctly parsed to/from JSON, and nothing IndexedDB specific. So I feel kinda confident in the IndexedDB-Backend implementation now.

- **Got a Demo app to play with :smile::** 
 ~(it's what I used to develop and test stuff)~ **Outdated**, but still a good place to play with a basic version of this: [CodeSandbox](https://codesandbox.io/s/elastic-kare-m1jp8?fontsize=14&hidenavigation=1&theme=dark)


packages/xstate-wallet/src/store/store-indexedDB-backend.ts
```diff
@@ -0,0 +1,128 @@
...
+  public async getLedger(key: string) {
+    return this.get('ledgers', key);
+  }
+
+  // Individual Getters
+
+  public async setPrivateKey(key: string, value: string) {
+    return this.add('privateKeys', key, value);
+  }
+  public async setChannel(key: string, value: MemoryChannelStoreEntry) {
```
tomclose: You might need to unpack the `MemoryChannelStoreEntry` here. The `MemoryChannelStoreEntry` is a potentially large object, that stores the states received for a channel. It will probably be useful to have methods like `MemoryChannelStoreEntry.addState` on the `DBBackend` itself.

(To explain some of the context here: `MemoryChannelStoreEntry` was originally `MemoryChannelStorage`, and was just an implementation detail of the `MemoryStore`, that handled operations involving the storage related to a given channel. The name was then changed to `MemoryChannelStoreEntry`, and it became the object that was passed through to the protocols, in order to get the protocols working, providing the `ChannelStoreEntry` interface that they expect. This was more a short term convenience, than a long-term architectural decision. I don't think it really makes sense to be passing the object responsible for storage through to the protocols, and we could probably separate that out here.)

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -172,21 +183,24 @@ export class MemoryStore implements Store {
...
-    const ledgerId = this._ledgers[peerId];
+    const ledgerId = await this.backend.getLedger(peerId);
 
     if (!ledgerId) throw new Error(`No ledger exists with peer ${peerId}`);
 
     return await this.getEntry(ledgerId);
   }
 
   public setLedger(entry: MemoryChannelStoreEntry) {
+    // !! SP: this function is not async!.... so that saved me!
```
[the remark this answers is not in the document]
andrewgordstewart: This is not a store function. The virtual funding workflow requires a store to be set up a certain way, and that way has not been figured out yet because budgeting is still not done.

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -220,29 +234,30 @@ export class MemoryStore implements Store {
...
   }
 
-  private setNonce(addresses: string[], value: BigNumber) {
-    if (value.lte(this.getNonce(addresses))) throw 'Invalid nonce';
-
-    this._nonces[this.nonceKeyFromAddresses(addresses)] = value;
+  private async setNonce(addresses: string[], value: BigNumber) {
+    const nonce = await this.getNonce(addresses);
+    if (value.lte(nonce)) throw 'Invalid nonce';
+    await this.backend.setNonce(this.nonceKeyFromAddresses(addresses), value);
```
andrewgordstewart: This is unsafe async code: It could be that on L246, the nonce is valid. But then, on line 247, the nonce is no longer valid.

The backend needs to be responsible for ensuring nonce safety, and it would thus throw on line 247 in that case, making L246 redundant.

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -196,13 +210,13 @@ export class MemoryStore implements Store {
     appDefinition = AddressZero
   ): Promise<ChannelStoreEntry> {
     const addresses = participants.map(x => x.signingAddress);
-
-    const myIndex = addresses.findIndex(address => !!this._privateKeys[address]);
+    const privateKeys = await this.backend.privateKeys();
+    const myIndex = addresses.findIndex(address => !!privateKeys[address]);
```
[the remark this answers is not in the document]
spersico: Yeah, you are right.
In my head last night it was the way to "reduce the ammount of requests to the DB" but we can deal with optimizations later.

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -285,17 +302,18 @@ export class MemoryStore implements Store {
...
     objectives?.forEach(objective => {
-      if (!_.includes(this._objectives, objective)) {
-        this._objectives.push(objective);
+      if (!_.includes(objectives, objective)) {
+        objectives.push(objective);
         this._eventEmitter.emit('newObjective', objective);
       }
     });
+    this.backend.setObjectives(dbObjectives);
   }
```
andrewgordstewart: I think the backend should accept a list of objectives and return the list of new objectives. This could then be
```
(await this.backend.setObjectives(objectives)).forEach(o => this._eventEmitter.emit('newObjective', o);)
```

Otherwise, this code is not safe as asynchronous code.

packages/xstate-wallet/src/store/memory-store.ts
spersico: Umm... okay.
The solution would be to set it like this way?
```suggestion
    if (value.lte(await this.getNonce(addresses))) throw 'Invalid nonce';
    await this.backend.setNonce(this.nonceKeyFromAddresses(addresses), value);
```

Or I could move this entire method to the backend, but even then, the lookup to the indexedDB would render that move useless. 
The thing is, at some point, we will always have some asynchronicity if we use an indexedDB storage to store nonces.

andrewgordstewart: Yes, the point is the lookup to the backend is useless: we should just ask the backend to set the nonce, and expect that it might throw an error:
```
try {
  await this.backend.setNonce(this.nonceKeyFromAddresses(addresses), value);
} catch (e) {
  // ...
}
```

tomclose: So maybe we should have a method `getNextNonce(key)` on the `DBBackend`, which is responsible for returning a new, unique nonce?

Depending on the transaction capabilities of IndexedDB (which are slightly unclear to me), we might be forced to have some unsafe async code. It would be good to isolate this as much as possible (i.e. into the `DBBackend`) though.

andrewgordstewart: The `setNonce` method would not be used by the creator of the channel; if you create a channel and ask me to join it, I have to `setNonce` with the nonce you've proposed.

spersico: > 

Yes, the point is the lookup to the backend is useless: we should just ask the backend to set the nonce, and expect that it might throw an error:

try {
  await this.backend.setNonce(this.nonceKeyFromAddresses(addresses), value);
} catch (e) {
  // ...
}

I think that if the problem is the uniqueness of the nonce, we could make indexedDB ensure the uniqueness of the key. On indexedDB we can set a key to be used as the lookup key, and an optional rule, is to make that key unique.

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -82,30 +91,32 @@ export interface Store {
...
+    } else {
+      this.backend = new MemoryBackend();
+    }
+
     if (privateKeys && privateKeys.length > 0) {
       // load existing keys
       privateKeys.forEach(key => {
         const wallet = new Wallet(key);
-        this._privateKeys[wallet.address] = wallet.privateKey;
+        this.backend.setPrivateKey(wallet.address, wallet.privateKey);
```
lalexgap: Will this cause issues due to `setPrivateKey` being `async` ? If we construct the `store` and then immediately attempt to do something that requires the `privateKey` the `privateKey` might not be set in the `backend` yet.

packages/xstate-wallet/src/store/memory-store.ts
spersico: @andrewgordstewart @tomclose I've got a CRUD working now of IndexedDB, but I'm not clear on what should I do (if anything) about this.

packages/xstate-wallet/src/store/indexedDB-backend.ts
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  // Individual Getters
+
+  public async setPrivateKey(key: string, value: string) {
+    return this.put(ObjectStores.privateKeys, value, key);
+  }
+  public async setChannel(key: string, value: MemoryChannelStoreEntry) {
+    return this.put(ObjectStores.channels, value, key);
+  }
```
andrewgordstewart: The "obsolete" store worked like:
- The store just stored plain data
- the ChannelStoreEntry was a class which accepted that data in its constructor
- `store.getEntry(channelId)` would construct a new class using the stored data.

The current MemoryChannelStoreEntry almost works like that. If you add:
```
  public data() {
    return {
      stateVariables: this.stateVariables,
      channelConstants: this.channelConstants,
      signatures: this.signatures,
      funding: this.funding,
      myIndex: this.myIndex
    };
  }

  static fromJson(data: string) {
    const {stateVariables, channelConstants, signatures, funding, myIndex} = JSON.parse(data);
    return new MemoryChannelStoreEntry(
      channelConstants,
      myIndex,
      stateVariables,
      signatures,
      funding
    );
  }
```
to the `MemoryChannelStoreEntry`, then you can do:
```suggestion
  public async setChannel(key: string, value: MemoryChannelStoreEntry) {
    return this.put(ObjectStores.channels, value.data, key);
  }
  
  public async getChannel(key: string): MemoryChannelStoreEntry {
    return MemoryChannelStoreEntry.fromJson(await this.get(ObjectStores.channels, key))
  }
```

spersico: Thanks for that! Added

packages/xstate-wallet/src/store/memory-store.ts
spersico: fixed!

packages/xstate-wallet/src/store/memory-store.ts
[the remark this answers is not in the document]
spersico: All functions that needed so, have been made **async** for the moment.

lalexgap:
Superseded by #1260

PR #1103, opened 2020-03-05
[channel-client] Enhance the FakeChannelProvider to handle multiple channels

### LINKED ISSUE #1085: Support multiple channels in FakeChannelProvider
Although the `ChannelClient` class supports multiple channels (at leas partially), the fake channel provider that we are currently building web3torrent against does _not_. There's no hope for getting multiple simultaneous peers connected without fixing that.

https://github.com/statechannels/monorepo/blob/111fcce4342ca8b7efc90b66b02bf93be1658821/packages/channel-client/tests/fakes/fake-channel-provider.ts#L129-L131

geoknee:
Closes #1085

### DIFF
--- a/packages/channel-client/tests/fake-channel-client.test.ts
+++ b/packages/channel-client/tests/fake-channel-client.test.ts
@@ -122,7 +122,7 @@ describe('FakeChannelClient', () => {
 
       return new Promise(resolve => {
         clientBEventEmitter.once('ChannelProposed', () => {
-          expect(providerB.latestState).toEqual(states['proposed']);
+          expect(providerB.latestState[channelId]).toEqual(states['proposed']);
           resolve();
         });
 
@@ -137,7 +137,7 @@ describe('FakeChannelClient', () => {
       setProviderStates([providerB], states['proposed']);
       const channelResult = await clientB.joinChannel(channelId);
       expect(channelResult).toEqual(states['running']);
-      expect(providerA.latestState).toEqual(states['running']);
+      expect(providerA.latestState[channelId]).toEqual(states['running']);
     });
 
     it('the player whose turn it is not cannot accept a join proposal they sent', async () => {
@@ -156,7 +156,7 @@ describe('FakeChannelClient', () => {
         UPDATED_APP_DATA
       );
       expect(channelResult).toEqual(states['updated_app_data']);
-      expect(providerB.latestState).toEqual(states['updated_app_data']);
+      expect(providerB.latestState[channelId]).toEqual(states['updated_app_data']);
     });
 
     it('the player whose turn it is not cannot update the channel', async () => {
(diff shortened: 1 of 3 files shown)


geoknee:
This enabled me to get 2x leechers for the same torrent in `web3torrent` t. It worked pretty well! The seeder got paid twice :-)

packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -81,7 +81,7 @@ export class FakeChannelProvider implements ChannelProviderInterface {
   }
 
   setState(state: ChannelResult): void {
-    this.latestState = state;
+    this.latestState = {...this.latestState, [state.channelId]: state};
```
junjizhi: Why is key in brackets? Shouldn't it be
```typescript
{...this.latestState, state.channelId: state};
```
Is this a typescript thing?

tomclose: I think that's a [ES2015 thing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names). If the key isn't a string literal, you have to put it in brackets.

packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -263,9 +263,9 @@ export class FakeChannelProvider implements ChannelProviderInterface {
         break;
       // auto-close, if we received a close
       case 'closing':
-        this.latestState = {...this.latestState, turnNum, status: 'closed'};
-        this.notifyOpponent(this.latestState, 'ChannelUpdate');
-        this.notifyAppChannelUpdated(this.latestState);
+        this.setState({...this.latestState[params.data.channelId], turnNum, status: 'closed'});
+        this.notifyOpponent(this.latestState[params.data.channelId], 'ChannelUpdate');
+        this.notifyAppChannelUpdated(this.latestState[params.data.channelId]);
```
junjizhi: Nitpick: Instead of repeating `params.data`, why don't we assign it to variable `const channel = params.data;` and use `channel` from then on.

packages/channel-client/tests/fake-channel-client.test.ts
```diff
@@ -174,7 +174,7 @@ describe('FakeChannelClient', () => {
       // phase and the clients directly go to the channel 'closed' state
       const channelResult = await clientA.closeChannel(channelId);
       expect(channelResult).toEqual(states['closed']);
-      expect(providerB.latestState).toEqual(states['closed']);
+      expect(providerB.latestState[channelId]).toEqual(states['closed']);
     });
 
     it('player with invalid turn cannot make a valid close channel call', async () => {
```
junjizhi: It would be nice to add a test to cover the case of multiple channels.

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: Yeah good shout

packages/channel-client/tests/fake-channel-client.test.ts
geoknee: Agreed. I'll raise an issue for that: I don't want to hold up progress on `web3torrent`.

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: ddd8c18

PR #1108, opened 2020-03-05
[web3t] Overhaul buffer system

geoknee:
This PR tries to improve clarity around the role of the `buffer` (previously known as the `funds`, which is kept by the seeder per torrent and per leecher, and used to decide when to choke them / prompt them to pay us). 

Choices:

- The buffer should have units of "pieces" (in the near future, probably "bytes", but *not* ETH or wei)
- Conversion factor is a new constant, available to seeders and leechers and easily tuneable.
- The threshold for unchoking is when the buffer is nonzero (previously, at a positive constant called `REQUEST_RATE`)
- The number of pieces that a leecher is prepared to pay for upfront in each payment is another tuneable constant  `BUFFER_REFILL_RATE`

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -26,7 +26,14 @@ export type TorrentCallback = (torrent: Torrent) => any;
 
 export * from './types';
 
-export const REQUEST_RATE = bigNumberify(10);
+export const WEI_PER_PIECE = bigNumberify(1); // cost per credit / piece
+export const BUFFER_REFILL_RATE = bigNumberify(100); // number of credits / pieces the leecher wishes to increase the buffer by
+// These variables control the amount of (micro)trust the leecher must invest in the seeder
+// As well as the overall performance hit of integrating payments into webtorrent.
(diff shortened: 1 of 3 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -214,22 +219,24 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+    // infer payment using update balance and previously stored balance
     const payment = newSeederBalance.sub(
       bigNumberify(this.peersList[infoHash][peerId].seederBalance)
     );
+    // store new balance
+    this.peersList[infoHash][peerId].seederBalance = newSeederBalance.toString();
+    // convert payment into number of pieces
     this.peersList[infoHash][peerId].buffer = bigNumberify(this.peersList[infoHash][peerId].buffer)
-      .add(payment)
+      .add(payment.div(WEI_PER_PIECE)) // TODO allow this to vary by piece size
```
geoknee: We should ensure `buffer` remains an integer, since our check for choking is now `.isZero` and we decrement by `1`.

PR #1112, opened 2020-03-06
Basic Budgeting Functionality

lalexgap:
Adds support for budgets in the `xstate-wallet`

It adds support for the following API calls:
- `getBudget` : Returns the current budget for a `hubAddress`.
- `approveBudgetAndFund`: Starts a new workflow that asks for user approval and starts ledger funding(mocked out)

### DIFF
--- a/packages/channel-client/src/channel-client.ts
+++ b/packages/channel-client/src/channel-client.ts
@@ -96,16 +96,16 @@ export class ChannelClient implements ChannelClientInterface<ChannelResult> {
   async approveBudgetAndFund(
     playerAmount: string,
     hubAmount: string,
-    playerDestinationAddress: string,
+    playerOutcomeAddress: string,
     hubAddress: string,
-    hubDestinationAddress: string
+    hubOutcomeAddress: string
   ): Promise<SiteBudget> {
     return this.provider.send('ApproveBudgetAndFund', {
       playerAmount,
       hubAmount,
-      playerDestinationAddress,
+      playerOutcomeAddress,
       hubAddress,
-      hubDestinationAddress
+      hubOutcomeAddress
     });
   }
 
--- a/packages/channel-client/src/channel-client.ts
+++ b/packages/channel-client/src/channel-client.ts
@@ -109,11 +109,11 @@ export class ChannelClient implements ChannelClientInterface<ChannelResult> {
     });
   }
 
-  async getBudget(hubAddress: string): Promise<SiteBudget> {
+  async getBudget(hubAddress: string): Promise<SiteBudget | {}> {
     return this.provider.send('GetBudget', {hubAddress});
   }
 
-  async closeAndWithdraw(hubAddress: string): Promise<SiteBudget> {
+  async closeAndWithdraw(hubAddress: string): Promise<SiteBudget | {}> {
     return this.provider.send('CloseAndWithdraw', {hubAddress});
   }
 }
--- a/packages/client-api-schema/src/types.ts
+++ b/packages/client-api-schema/src/types.ts
@@ -167,9 +167,9 @@ export interface SiteBudget {
 
 export interface BudgetRequest extends Balance {
   site: string;
-  playerDestinationAddress: string;
+  playerOutcomeAddress: string;
   hubAddress: string;
-  hubDestinationAddress: string;
+  hubOutcomeAddress: string;
 }
 export type GetBudgetRequest = JsonRpcRequest<'GetBudget', {hubAddress: Address}>;
 export type GetBudgetResponse = JsonRpcResponse<SiteBudget | {}>;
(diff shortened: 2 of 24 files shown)


packages/channel-client/src/channel-client.ts
```diff
@@ -105,11 +105,11 @@ export class ChannelClient implements ChannelClientInterface<ChannelResult> {
     });
   }
 
-  async getBudget(hubAddress: string): Promise<SiteBudget> {
+  async getBudget(hubAddress: string): Promise<SiteBudget | {}> {
```
snario: Why `| {}`?

lalexgap: We need some return type when `getBudget` is called and no budget exists, I wasn't sure the best way to represent this over JSON-RPC but I used `{}` for now.

Another option would be to return an empty `SiteBudget` but I think that may be less clear? (Or maybe just a `null` result`)?

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
```diff
@@ -0,0 +1,120 @@
...
+    updateBudget: (context: WorkflowContext, event) => {
+      return store.updateOrCreateBudget(context.budget);
+    },
+    createAndFundLedger: (context: WorkflowContext, event) => {
+      // TODO: Hook up to workflow when it exists
+      return Promise.resolve();
+    }
+  };
+  const actions = {
+    // TODO: We should probably set up some standard actions for all workflows
```
snario: Mention this in a Github Issue?

packages/client-api-schema/src/types.ts
```diff
@@ -160,9 +160,18 @@ export interface SiteBudget {
   inUse: Balance;
   direct: Balance;
 }
+
+export interface BudgetRequest extends Balance {
+  site: string;
```
snario: Rename to `domain`?

packages/client-api-schema/src/types.ts
```diff
@@ -160,9 +160,18 @@ export interface SiteBudget {
...
+  playerDestinationAddress: string;
```
snario: Perhaps use `outcomeAddress`?

packages/client-api-schema/src/types.ts
```diff
@@ -160,9 +160,18 @@ export interface SiteBudget {
...
+  hubAddress: string;
```
snario: Why is the hub necessarily attached to a `BudgetRequest`?

packages/xstate-wallet/src/serde/app-messages/serialize.ts
```diff
@@ -1,11 +1,35 @@
...
+  Allocation,
+  AllocationItem,
+  SimpleAllocation,
+  SiteBudget,
+  BudgetItem
+} from '../../store/types';
+import {tokenAddress, ETH_ASSET_HOLDER_ADDRESS} from '../../constants';
 
+export function serializeSiteBudget(budget: SiteBudget): AppSiteBudget {
+  const assetBudget = budget.budgets[ETH_ASSET_HOLDER_ADDRESS];
```
snario: Can we avoid hard-coding ETH inside the wallet if it isn't too much extra to add an `assetType` or `tokenAddress` field to the budget object?

packages/client-api-schema/src/types.ts
lalexgap: What would be the case when there wouldn't be a hub? Are you suggesting the `wallet` gets to decide the `hub`?

packages/xstate-wallet/src/serde/app-messages/serialize.ts
lalexgap: Sure, I've added support for multiple assets.

packages/client-api-schema/src/types.ts
lalexgap: currently keeping as `site` to keep it consistent with all other `budget` related entities. `domain` probably is a better name tho

packages/client-api-schema/src/types.ts
snario: I don't think budgets are related to hubs

andrewgordstewart: An approved budget should result in a funded ledger channel with a hub.

lalexgap: I should have been clear that one of our simplifying assumptions for the initial implementation is that the `budget` just represents the funds allocated to the ledger channel with the `hub`.

tomclose: For the time being (to ship web3torrent):
1. There will be at most one hub per site
2. The site will be responsible for providing the hub address
3. The budget will represent the funds in a single ledger channel with this hub

This means that, for the time being, it makes sense to include the `hubAddress` in the budget request.

PR #1116, opened 2020-03-06
[web3t] Overhaul upload/download UI

geoknee:
- Integrates channel information and data transfer information. 
- adds a "Close" button to anticipate future functionality that we will need
- removes the ChannelList but preserves some of the styling 
- skips some UI tests (eek!) TODO make an issue to unskip them , or just delete

Screenshot 2020-03-06 at 15 11 19


packages/web3torrent/src/components/torrent-info/wires-list/WiresList.tsx
```diff
@@ -0,0 +1,56 @@
...
+        </tr>
+      )
+    );
+  }
+
+  return (
+    <section className="wires-list">
+      <table className="wires-list-table">
+        <tbody>{Object.values(torrent.wires).map(wireToTableRow)}</tbody>
+      </table>
```
geoknee: I've just realised an issue with this approach. When we finish downloading, it seems as if the wire is thrown away: that means it will disappear from the UI. The channel still exists, of course, and we want to be able to click "Close" so we can extract our money. 

A solution would be to have a "channels first" table, with the "downloaded/uploaded bytes" field grabbed with a reverse lookup through the existing wires. Then, if the wire is destroyed we will only lose this field (and we could then worry about saving the information somewhere so we never lose it).

junjizhi: I like the `channels first` approach. Otherwise we have to build another UI for users to close channels and extract money.

packages/web3torrent/src/components/torrent-info/wires-list/WiresList.tsx
```diff
@@ -0,0 +1,56 @@
...
+import './WiresList.scss';
+import {ChannelState} from '../../../clients/payment-channel-client';
+import {PaidStreamingWire} from '../../../library/types';
+
+export type UploadInfoProps = {
+  torrent: Torrent;
+  channelCache: Record<string, ChannelState>;
+  channelIds: string[];
+  peerType: 'seeder' | 'leecher';
+};
```
junjizhi: Nitpick: Based on your usage, the params can be distilled into just `{wires, channels, peerType}`.  It adds a little work for the call site, but the code in this component is cleaner because it does less work on channel id filtering and deconstructing `torrent`.

geoknee:
About the skipped tests: @junjizhi discussed this and agreed that deleting them is probably a good way forward. We have some kind of visual testing for these components via Storybook, which allows us to move a little faster than having jest tests.

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -1,11 +1,12 @@
 import {ChannelResult, Message, ChannelClientInterface} from '@statechannels/channel-client';
-import {bigNumberify} from 'ethers/utils';
+import {utils} from 'ethers';
```
snario: Why?

geoknee: It allowed storybook to not throw errors , see https://github.com/ethers-io/ethers.js/issues/349

PR #1127, opened 2020-03-06
[web3t] Strip out cost property

geoknee:
Compute cost from a CONSTANT * num_pieces [current behaviour]

( anticipated future behaviour might be CONSTANT * num_bytes instead )


packages/web3torrent/src/components/share-list/share-file/ShareFile.tsx
```diff
@@ -12,7 +12,7 @@ const ShareFile: React.FC<ShareFileProps> = ({file, goTo}: ShareFileProps) => {
       <td className="other-cell">{file.length}Mb</td>
       <td className="other-cell">{file.numPeers}S</td>
       <td className="other-cell">{file.numPeers}P</td>
-      <td className="other-cell">${file.cost}</td>
+      {/* <td className="other-cell">${cost}</td>  TODO reinstate this cell when we have a wer-per-byte model implpemented*/}
```
geoknee: When we get there, we can compute the cost directly from the `file`. As it is, we need the number of pieces, which is a property of the `torrent`.

PR #1142, opened 2020-03-07
[TTT] Inject Environment Variables into TTT

Alonski:
This PR sets up Env Var injection into TTT.
It uses `ember-cli-dotenv` for most of the work.

Something that I still haven't managed to setup is injecting the TTT_CONTRACT_ADDRESS.

This is generated at runtime so is a bit more complex. I believe this PR can be merged now and I will fix that in a different PR once I figure it out.


spersico:
Seems like you need to set some environment variables in the CI, or put some defaults or something:
image

Alonski:
Yes thanks @spersico I didn't setup everything for prod

PR #1150, opened 2020-03-09
[web3t] Leecher auto-closes channels when download finishes

### LINKED ISSUE #1104: [web3t] Close payment channels manually + automatically
We probably want to do this automatically if we manage to download the entire file, but also manually (e.g. with a new button in the `<ChannelList />` component) if we run out of funds before that happens.

geoknee:
Less clicking required :-)

Fixes #1104


geoknee:
We should also address the complementary issue, where if I close a channel manually (or my peer does) I should then stop sending requests (because now I am choked forever,

PR #1151, opened 2020-03-09
[web3t] Transform WiresList into ChannelsList

### LINKED ISSUE #1147: [web3torrent] Prevent dynamic re-ordering in WiresList
Given that there are buttons that I need to click in each table row, it's pretty awful UX that the ordering of the rows can change and they jump around. A simple fix might be to order by `channelId`. 

A more involved fix would maintain the list in the order that rows were added (preventing reordering when new rows are inserted).

geoknee:
And enforce static ordering on the table rows. 

Fixes #1147 

TODO

- [x] debug error where `PaymentChannelClient` complains that `ChannelClient` is undefined


packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    let uploaded;
+    let peerAccount;
+    wires.forEach(wire => {
+      if (peerType === 'leecher' && wire.paidStreamingExtension.peerChannelId === channelId) {
+        uploaded = wire.uploaded;
+        peerAccount = wire.paidStreamingExtension.peerAccount;
+      }
+    });
```
junjizhi: `wires.find(...)` is faster than going through the whole list.

packages/web3torrent/src/components/torrent-info/TorrentInfo.tsx
```diff
@@ -20,6 +21,8 @@ const TorrentInfo: React.FC<TorrentInfoProps> = ({
   channelCache = {},
   mySigningAddress
 }) => {
+  const closeChannel = useContext(WebTorrentContext).paymentChannelClient.closeChannel;
```
junjizhi: I think you can't assign function like this. `this` won't work correctly in `closeChannel`. Try just storing the `paymentChannelClient`.

geoknee: thanks. It has been a pain trying to get storybook to work again!

packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
```diff
@@ -0,0 +1,87 @@
...
+    );
+  }
+
+  render() {
+    return (
+      <section className="wires-list">
+        <table className="wires-list-table">
+          <tbody>
+            {_.keys(this.props.channels)
+              .sort((channelId1, channelId2) => Number(channelId1) - Number(channelId2))
```
geoknee: Because I renamed the file, it's hard to see what changed. This line is the main change, plus refactoring from a function component to a class component, to handle context in a better way.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -228,8 +228,10 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       this.paymentChannelClient.channelCache[channelId].beneficiaryBalance
     );
     // infer payment using update balance and previously stored balance
-    const payment = newSeederBalance.sub(
-      bigNumberify(this.peersList[infoHash][peerId].seederBalance)
+    const payment = bigNumberify(
+      this.peersList[infoHash][peerId]
+        ? newSeederBalance.sub(bigNumberify(this.peersList[infoHash][peerId].seederBalance))
+        : 0
```
junjizhi: This assignment is too long and hard to read. maybe store `this.peersList[infoHash][peerId]` in a var and avoid using `bigNumberify` twice.

PR #1161, opened 2020-03-09
Observables and side effects

andrewgordstewart:
I'm opening this PR as a way of communicating the result of a couple discussions with @kerzhner about observables and side effects.

# Past
The code was originally in [this form](https://github.com/statechannels/monorepo/commit/106b6ed951ab646a64306e554c25a98c5d5717b9):

```    this.messagingService.requestFeed.subscribe(r => {
      if (r.type === 'CREATE_CHANNEL' || r.type === 'JOIN_CHANNEL') {
        const workflow = this.startApplicationWorkflow();
        this.workflows.push(workflow);

        workflow.machine.send(r);
      }
    });
```

This code had the problem that `this.startApplicationWorkflow()` could error, and nobody would know about it: the execution of the subscription callback would halt, but the error would not propagate, making this issue difficult to debug.

# Present
Placing the side effect inside a `tap` (the current code) meant that the error triggers the `onError` callback of the subscription.

```    this.messagingService.requestFeed
      .pipe(
        filter(r => r.type === 'CREATE_CHANNEL' || r.type === 'JOIN_CHANNEL'),
        tap(r => {
          const workflow = this.startApplicationWorkflow();
          this.workflows.push(workflow);
          workflow.machine.send(r);
        })
      )
      .subscribe(undefined, console.error);
```

This is helpful to developers, but it runs the risk that:
- if an observable with multiple subscriptions has side effects, then the side effects may (unexpectedly) happen twice. For example, in this case, if multiple people subscribed to the new observable, then multiple workflows would be started.

In this particular example, since the side effect occurs in an inline observable, there's no risk of multiple subscriptions.

# Future
@kerzhner observed that an observable is best seen as a way to transform a stream of data in a predictable, ideally pure, fashion. This change represents one way to do this. `tap` is not used to perform side effects.

This is a bit subtle, since calling `start()` inside `createApplicationWorkflow` would result in side effects.

It's probably worth thinking more about this issue down the road, but I've decided to leave it at that for now.

### DIFF
--- a/packages/xstate-wallet/src/channel-wallet.ts
+++ b/packages/xstate-wallet/src/channel-wallet.ts
@@ -8,8 +8,9 @@ import {Wallet as WalletUi} from './ui/wallet';
 import {interpret, Interpreter, State} from 'xstate';
 import {Guid} from 'guid-typescript';
 import {Notification, Response} from '@statechannels/client-api-schema';
-import {filter, map, tap} from 'rxjs/operators';
+import {filter, map} from 'rxjs/operators';
 import {Message, OpenChannel} from './store/types';
+import {AppRequestEvent} from './event-types';
 
 export interface Workflow {
   id: string;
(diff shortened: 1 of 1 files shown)


andrewgordstewart:
Addendum: the general idea is to delay side effects, when possible, until either the subscription callback, or a `tap` operator near the end of the pipe.

When that is not possible, `publishLast` can be used to ensure that the side effects only happen once: https://rxjs.dev/api/operators/publishLast

snario:
Thanks for putting this together

andrewgordstewart:
Closing in anticipation of @tomclose's work which supersedes this.

In conclusion, here are the guidelines for observables and subscriptions:
1. When possible, avoid data transformation in a subscription ([Past](https://github.com/statechannels/monorepo/commit/106b6ed951ab646a64306e554c25a98c5d5717b9#diff-050186ea4f5d2757147267eed7eb95b2L52)).
2. When possible avoid side effects in an observable ([Present](https://github.com/statechannels/monorepo/commit/106b6ed951ab646a64306e554c25a98c5d5717b9#diff-050186ea4f5d2757147267eed7eb95b2R55-R56)).
3. If a side effect does happen in an observable, and the side effect should only happen once, then enforce this in the pipe with something like `publishLast`, `publish` + `refCount`, or `share`

PR #1163, opened 2020-03-10
Fix skipped test

junjizhi:
Fixes https://github.com/statechannels/monorepo/issues/1099

Fake client has a very subtle bug that makes the skipped test failed and the wrong turn error we saw:

image

~~The bug was due to fake client sets the wrong turn number when a participant joins the channel.~~

### DIFF
--- a/packages/web3torrent/src/library/client-seeding.test.ts
+++ b/packages/web3torrent/src/library/client-seeding.test.ts
@@ -4,8 +4,7 @@ import {
   defaultFileMagnetURI,
   defaultSeedingOptions,
   defaultTorrentHash,
-  mockMetamask,
-  mockChannelState
+  mockMetamask
 } from './testing/test-utils';
 import WebTorrentPaidStreamingClient, {
   ClientEvents,
@@ -31,7 +30,7 @@ describe('Seeding and Leeching', () => {
     });
     seeder.on('error', err => fail(err));
     seeder.on('warning', err => fail(err));
-    seeder.paymentChannelClient.channelCache = {0x0: mockChannelState};
+    seeder.paymentChannelClient.channelCache = {};
 
     leecher = new WebTorrentPaidStreamingClient({
       pseAccount: '2',
@@ -40,7 +39,7 @@ describe('Seeding and Leeching', () => {
     });
     leecher.on('error', err => fail(err));
     leecher.on('warning', err => fail(err));
-    leecher.paymentChannelClient.channelCache = {0x0: mockChannelState};
+    leecher.paymentChannelClient.channelCache = {};
   });
 
   it('should seed and remove a Torrent', done => {
@@ -82,25 +81,14 @@ describe('Seeding and Leeching', () => {
     });
   }, 10000);
 
-  it.skip('should be able to unchoke and finish a download', done => {
+  it('should unchoke when seeder unblocks leecher', done => {
     seeder.seed(defaultFile as File, defaultSeedingOptions(), seededTorrent => {
       seeder.once(ClientEvents.PEER_STATUS_CHANGED, ({peerAccount}) => {
-        seeder.togglePeer(seededTorrent.infoHash, peerAccount);
-
         seeder.once(ClientEvents.PEER_STATUS_CHANGED, ({torrentPeers}) => {
           expect(torrentPeers[`${leecher.pseAccount}`].allowed).toEqual(true);
+          done();
         });
-
-        seeder.on(ClientEvents.TORRENT_NOTICE, ({command}) => {
-          if (command == PaidStreamingExtensionNotices.ACK) {
-            leecher.once(ClientEvents.TORRENT_DONE, ({torrent: leechedTorrent}) => {
-              expect(seededTorrent.files[0].done).toEqual(leechedTorrent.files[0].done);
-              expect(seededTorrent.files[0].length).toEqual(leechedTorrent.files[0].length);
-              expect(seededTorrent.files[0].name).toEqual(leechedTorrent.files[0].name);
-              done();
-            });
-          }
-        });
+        seeder.togglePeer(seededTorrent.infoHash, peerAccount);
       });
 
       leecher.add(seededTorrent.magnetURI, {store: MemoryChunkStore});
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -176,11 +176,6 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
           response(false);
 
           this.blockPeer(torrent.infoHash, wire, peerAccount);
-          this.emit(ClientEvents.PEER_STATUS_CHANGED, {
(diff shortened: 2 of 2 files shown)


packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -174,7 +174,7 @@ export class FakeChannelProvider implements ChannelProviderInterface {
     // [assuming we're working with 2-participant channels for the time being]
     this.setState({
       ...latestState,
-      turnNum: bigNumberify(3).toString(),
+      turnNum: bigNumberify(2).toString(),
```
junjizhi: ☝️This is the subtle bug that causes issues.

snario: Hm, I've run into this in the past. What could we have done to make this more prevalent when this bug shows up?

packages/web3torrent/src/library/client-seeding.test.ts
```diff
@@ -78,32 +77,24 @@ describe('Seeding and Leeching', () => {
...
-              expect(seededTorrent.files[0].done).toEqual(leechedTorrent.files[0].done);
-              expect(seededTorrent.files[0].length).toEqual(leechedTorrent.files[0].length);
-              expect(seededTorrent.files[0].name).toEqual(leechedTorrent.files[0].name);
-              done();
-            });
-          }
-        });
+        // Add a delay to prevent race condition. Without the delay, leecher may
+        // transfer fund before it joins the channel.
+        await new Promise(r => setTimeout(r, 2000));
```
snario: Since we have both `leecher` and `seeder` in context, might there be some promise we could plug into and `await` as opposed to adding a delay?

junjizhi: Actually we don't need that. The race condition was caused by the fake channel client sending a wrong turn, somehow causing the leecher to get ahead of itself. Removed.

packages/channel-client/tests/fakes/fake-channel-provider.ts
junjizhi: I think the console log error log is one thing we shouldn't ignore. The other thing we can do is add a test, now that we know.

geoknee:
There shouldn't be a need to publish any packages: since we use a monorepo the packages get automatically linked.

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: There is already a test, and it asserts the *current* behaviour. Also, the error in the console does not always appear.  Clearly a channel runs successfully some or most of the time.

geoknee:
See #1164

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: To be clear: I don't think there was a bug here. When you join the channel, you have state 0 from your opponent: you should sign state 1, safely fund the channel, wait for your opponent to sign state 2, and then sign state 3. The FakeProvider shortcuts some of this, but the joining party (participant `1`) should always be signing odd-numbered-turnNum states.

junjizhi: Should calling `joinChannel` increase the turn number? Or the increase and signing happens after? Everything seems to work fine after I make this change.

junjizhi: I think I found out what really happened.

The below are logs from my PR (without the turnNum = 2 fix) rebased on George's PR: https://github.com/statechannels/monorepo/pull/1164

Seeder log:

image

Leecher log:

image

Seeder creates a channel, and sends a STOP event to leecher. Upon receiving STOP, leecher tries to transfer funds. At that moment, leecher hasn't even joined the channel yet. (For evidence, only after the 2nd STOP, we saw the `payment made...` message in the leecher log).

George's fix with a condition inside the `transferFund` isn't correct because it actually makes the leecher skip the first payment. That's why in the Seeder log, after the accepting the first payment,  `newSeederBalance` is still zero!

I think the safest fix is to hold the first STOP firing until leecher joins the channel. 

What do you think? @geoknee

junjizhi: It's such a coincidence that, when I put in turnNum = 2 fix, first eager STOP event actually makes leecher to transfer fund successfully and increase the turnNum correctly to 3 before it joins the channel. That could mean some checks may be missing in fake channel provider.

geoknee: > I think the safest fix is to hold the first STOP firing until leecher joins the channel.

This would lead to the leecher being unchoked and able to download until they join the channel, which doesn't seem right.

junjizhi: > This would lead to the leecher being unchoked and able to download until they join the channel, which doesn't seem right.

What I meant is seeder still chokes the wire but holds the first STOP until leecher joins the channel. STOP essentially equals `asking for payment` in our implementation. In other words, with this proposed fix, we decouple wire-choking and asking for payment. It does add little complexity to handle the initial state, but it is still reasonable? 🤷‍♂ 

The fix in https://github.com/statechannels/monorepo/pull/1164 results that, leecher joining the channel is equivalent to paying the first payment, therefore starts leeching right away. That doesn't seem right either.

junjizhi:
I've taken out the turnNum = 2 change in this PR. 

This PR can be rebased on #1164, or a future PR, that handles the transfer fund race condition, depending on what we decide.

packages/web3torrent/src/library/client-seeding.test.ts
```diff
@@ -78,32 +77,21 @@ describe('Seeding and Leeching', () => {
...
     });
   }, 10000);
 
-  it.skip('should be able to unchoke and finish a download', done => {
+  it('should unchoke when seeder unblocks leecher', done => {
     seeder.seed(defaultFile as File, defaultSeedingOptions(), seededTorrent => {
-      seeder.once(ClientEvents.PEER_STATUS_CHANGED, ({peerAccount}) => {
-        seeder.togglePeer(seededTorrent.infoHash, peerAccount);
-
+      seeder.once(ClientEvents.PEER_STATUS_CHANGED, async ({peerAccount}) => {
```
junjizhi: async is not needed.

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: Yes I think this is reasonable. Let's implement that after merging this and #1164 . 

(We discussed that having a little bit of free leeching might not be so bad, but that it's generally nicer to not request a payment if you don't need one, rather than requesting the payment and have the leecher decide not to pay).

junjizhi: Filed the new issue: https://github.com/statechannels/monorepo/issues/1178

junjizhi:
@geoknee Rebased. This PR is ready to be merged.

PR #1164, opened 2020-03-10
[channel-client][web3t] Finish multichannel enhancement to FakeChannelProvider

geoknee:
Turns out #1103 was only a job half done...

Also make changes in `web3torrent` to get things working (again), and to avoid "not your turn" errors.


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -297,24 +302,34 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
           break;
         case PaidStreamingExtensionNotices.MESSAGE:
           message = JSON.parse(data.message);
-          await this.paymentChannelClient.pushMessage(message);
-          channelId = message.data.channelId;
-          channelState = this.paymentChannelClient.channelCache[channelId];
```
geoknee: I think this was probably the source of the "not your turn" errors we were seeing. We were pulling information out of the `channelCache`, and this was only being updated asynchronously (I receive the message and push it into my wallet, and this promise resolves; some time later, a channel updated event fires and the `channelCache` is updated. It's not clear how to await that second event here, as we would need to if we wanted to avoid the race condition).

The solution here is to grab the state straight out of the received message. Downside is we don't do any kind of signature of balances (`validTransition`) checks on that state.

Previously, we updated the `channelCache` on the way "in", which is another way of avoiding the problem.

packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -193,17 +195,17 @@ export class FakeChannelProvider implements ChannelProviderInterface {
     const allocations = params.allocations;
     const appData = params.appData;
 
-    log.debug(`Player ${this.getPlayerIndex()} updating channel ${channelId}`);
+    log.debug(`Player ${this.getPlayerIndex(channelId)} updating channel ${channelId}`);
     const latestState = this.findChannel(channelId);
 
     const nextState = {...latestState, participants, allocations, appData};
-    if (nextState !== latestState) {
```
geoknee: This logic seems incompatible with how we are implementing a unidirectional payment channel: namely, an implementation where the beneficiary countersigns payments. In that scenario, we often pass precisely the same state to this function, which must increment the `turnNum` and store it.

packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -19,6 +19,8 @@ import {Wallet, utils} from 'ethers';
 
 const bigNumberify = utils.bigNumberify;
 
+type ChannelId = string;
```
snario: Is this the right place for this? Also, couldn't we use the wrapped type strategy such as what Andrew proposed / [newtype](https://github.com/gcanti/newtype-ts)?

packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -27,11 +29,11 @@ export class FakeChannelProvider implements ChannelProviderInterface {
   private events = new EventEmitter();
   protected url = '';
 
-  playerIndex?: number;
-  opponentIndex?: number;
+  playerIndex: Record<ChannelId, 0 | 1> = {};
```
snario: Perhaps rename to be plural since we're making it plural, e.g., `playerIndices`

packages/channel-client/tests/fakes/fake-channel-provider.ts
snario: Makes sense.

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -226,15 +226,15 @@ export class PaymentChannelClient {
     const turnNum = Number(channelState.turnNum);
     // doesn't guarantee that my balance increased
     if (channelState.beneficiary === this.mySigningAddress) {
-      // returns true for the second postFS if I am the beneficiary
-      // (I need to accept this 'payment' in order for another one to be sent)
-      return (channelState.status === 'running' && turnNum % 2 === 1) || turnNum === 3;
+      return channelState.status === 'running' && turnNum % 2 === 1;
     }
-    throw new Error(`${this.mySigningAddress} is not the beneficiary ${channelState.beneficiary}`);
+    // throw new Error(`${this.mySigningAddress} is not the beneficiary ${channelState.beneficiary}`);
```
snario: Maybe replace with `console.error`?

snario: or `log.debug`

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -256,14 +256,19 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-      channelId,
-      WEI_PER_BYTE.mul(BUFFER_REFILL_RATE).toString()
-    );
-    const newSeederBalance = bigNumberify(
-      this.paymentChannelClient.channelCache[channelId].beneficiaryBalance
-    );
-    log(`payment made for channel ${channelId}, newSeederBalance: ${newSeederBalance}`);
+    if (
+      this.paymentChannelClient.channelCache[channelId] &&
+      this.paymentChannelClient.channelCache[channelId].status === 'running'
```
snario: In what scenario would this method, `transferFunds`, be called where there is not a running channel?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -297,24 +302,34 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-              channelId
-            );
-            await this.paymentChannelClient.acceptPayment(
-              this.paymentChannelClient.channelCache[channelId]
-            );
-            this.unblockPeer(torrent.infoHash, wire, wire.paidStreamingExtension.peerAccount);
+          if (message.recipient === this.pseAccount) {
+            channelId = message.data.channelId;
+            channelState = await this.paymentChannelClient.pushMessage(message);
+            log(`State receivced with turnNum ${channelState.turnNum}`);
```
snario:
```suggestion
            log(`State received with turnNum ${channelState.turnNum}`);
```

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -297,24 +302,34 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+              await this.paymentChannelClient.acceptPayment(
+                this.paymentChannelClient.channelCache[channelId]
+              );
+              await this.refillBuffer(
+                torrent.infoHash,
+                wire.paidStreamingExtension.peerAccount,
+                channelId
+              );
+              this.unblockPeer(torrent.infoHash, wire, wire.paidStreamingExtension.peerAccount);
+              // TODO only unblock if the buffer is large enough
```
snario:
```suggestion
              // TODO: only unblock if the buffer is large enough
```

junjizhi:
@geoknee I tried this branch, and still saw `not your turn` error.

packages/web3torrent/src/library/web3torrent-lib.ts
geoknee: When the channel is `closed`, for example.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -256,14 +280,20 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-    );
-    const newSeederBalance = bigNumberify(
-      this.paymentChannelClient.channelCache[channelId].beneficiaryBalance
-    );
-    log(`payment made for channel ${channelId}, newSeederBalance: ${newSeederBalance}`);
+    if (
+      this.paymentChannelClient.channelCache[channelId] &&
+      this.paymentChannelClient.channelCache[channelId].status === 'running' &&
+      bigNumberify(this.paymentChannelClient.channelCache[channelId].turnNum).gt(3)
+    ) {
```
junjizhi: This may not be the correct fix. It may allow leecher may skip the first payment. See https://github.com/statechannels/monorepo/pull/1163#discussion_r390678890

geoknee: The `turnNum = 3` state is not really a payment, it is a setup state that confirms the initial balance of the channel. It is handled by the `joinChannel` call from the leecher.

This condition is there to stop the leecher sending a payment (e.g. `turnNum = 5` state) before they have received the "special" `turnNum = 4` state from the seeder. I'm going to call the `turnNum = 4` state the "spacer state" because it doesn't do anything useful save for increasing the `turnNum` and allowing the leecher to take the next step.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -227,6 +227,30 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     this.paymentChannelClient.onChannelProposed(async (channelState: ChannelState) => {
       await this.paymentChannelClient.joinChannel(channelState.channelId);
     });
+
+    this.paymentChannelClient.onChannelUpdated(async (channelState: ChannelState) => {
+      log(`State received with turnNum ${channelState.turnNum}`);
+      if (
+        this.paymentChannelClient.isPaymentToMe(channelState) ||
+        Number(channelState.turnNum) === 3
```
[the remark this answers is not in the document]
geoknee: I think this is the behaviour we want, this should trigger the seeder to respond with the `turnNum = 4` "spacer state" (see below comment).

Perhaps we should refactor the code so this is not associated with a "payment". The functionality should be the same, but we could call it "acceptSpacerState` or something like that. Currently, we just treat the `turnNum = 3` state as an empty payment, and the seeder "accepts" it, by sending the `turnNum = 4` spacer state.

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: We could but that is out of scope for this PR. I could just revert to using string throughout but I thought it made it easier to see that the lookup tables stored in this class are all being keyed by a channel id.

packages/web3torrent/src/library/web3torrent-lib.ts
geoknee: In the end I moved this code into an `onChannelUpdated` callback, which is definitely the best solution.

PR #1167, opened 2020-03-10
Create and fund workflow

andrewgordstewart:
The create-and-direct-fund workflow assumed direct funding. This PR renames it to create-and-fund, and its behaviour switches funding strategy based on an environment variable. (This could be improved to switching based on some kind of global config object.)

Master: https://xstate.js.org/viz/?gist=656e4b0ab0ec6c6e0b46ac446aeab172
This branch: https://xstate.js.org/viz/?gist=cf7c6ef762ee62489895ed7e313b718a

Within the branch, I removed `advance-channel` as a simplification, since it seems safe to use support-state to collectively.

I also made, then reverted, two additional changes:
1. Added temporary budget logic
2. Added `appData`/`appDefinition` validation to the `create-and-fund` workflow.

More details [here](https://github.com/statechannels/monorepo/commit/155de07d14a55edfde69cab9a00d9c8eed9bf487) and [here](https://github.com/statechannels/monorepo/pull/1144)


packages/xstate-wallet/src/actions.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line no-restricted-imports
+export {escalate} from 'xstate/lib/actions';
```
snario: Why does this warrant its own file?

packages/xstate-wallet/src/constants.ts
```diff
@@ -7,7 +8,14 @@ export const NITRO_ADJUDICATOR_ADDRESS =
   process.env.NITRO_ADJUDICATOR_ADDRESS || '0x0000000000000000000000000000000000000000';
 
 // TODO: Move top ENV variable
-export const HUB_ADDRESS = process.env.HUB_ADDRESS || '0x100063c326b27f78b2cBb7cd036B8ddE4d4FCa7C';
+// Corresponds to '0x8624ebe7364bb776f891ca339f0aaa820cc64cc9fca6a28eec71e6d8fc950f29' from test data
+export const HUB_ADDRESS = process.env.HUB_ADDRESS || '0xaaaa84838319627Fa056fC3FC29ab94d479B8502';
```
snario: It might be a good idea to add an `env.js` file, or `config.js` file (with a `Config` object)

packages/xstate-wallet/src/integration-tests/funding.test.ts
```diff
@@ -34,8 +34,7 @@ it('allows for two wallets to fund an app', async () => {
   const createEvent = generateCreateChannelRequest(playerA.participant, playerB.participant);
   const createPromise = playerA.messagingService.outboxFeed
     .pipe(
-      filter(m => 'id' in m && m.id === createEvent.id),
-      map(m => m as CreateChannelResponse),
+      filter((m): m is CreateChannelResponse => 'id' in m && m.id === createEvent.id),
```
snario: Why `is`? Just curious the use case

packages/xstate-wallet/src/constants.ts
andrewgordstewart: I'd like this as a separate PR

packages/xstate-wallet/src/actions.ts
andrewgordstewart: `escalate` is not exported from the `xstate` module. I thought it would be better to handle that issue once instead of on a per-use basis.

packages/xstate-wallet/src/integration-tests/funding.test.ts
andrewgordstewart: It serves as a type guard, which `pipe` can use to narrow the type later down the pipe. In this case, it serves to let [this](https://github.com/statechannels/monorepo/blob/4020bc8e988687d11ac5cee4b84ad3250fa32908/packages/xstate-wallet/src/integration-tests/funding.test.ts#L46) compile.

packages/xstate-wallet/src/integration-tests/funding.test.ts
```diff
@@ -34,8 +34,7 @@ it('allows for two wallets to fund an app', async () => {
...
       first()
     )
     .toPromise();
```
kerzhner: There is probably a way to write this code with observables instead of flipping back and forth between observables and promises. Might not be worth it right now.

PR #1168, opened 2020-03-10
Enable ethereum in xstate wallet

tomclose:


packages/xstate-wallet/src/chain.ts
```diff
@@ -28,6 +28,9 @@ export interface Chain {
   getChainInfo: (channelId: string) => Promise<ChannelChainInfo>;
   chainUpdatedFeed: (channelId: string) => Observable<ChannelChainInfo>;
   deposit: (channelId: string, expectedHeld: string, amount: string) => Promise<void>;
+  ethereumEnable: () => Promise<void>;
```
snario: Is there a different place than the `Chain` interface this could live? e.g., this is not relevant when run inside the `hub`, but everything else is relevant. Perhaps a `Web3Chain` extension of `Chain`?

packages/xstate-wallet/src/chain.ts
```diff
@@ -100,6 +115,35 @@ export class ChainWatcher implements Chain {
...
+
+  public get ethereumIsEnabled(): boolean {
+    if (window.ethereum) {
+      return !!window.ethereum.selectedAddress;
+    } else {
+      return false;
+    }
+  }
+
+  public get selectedDestination(): string {
```
geoknee: I think we ought to avoid conflating `destination` (which is a nitro term) and `selectedAddress` (which is an ethereum / MM term), throughout. We should use `selectedAddress` here (downstream code such as the app can decide what to do with it, such as using it as the `destination` in the outcome for a channel.

packages/xstate-wallet/src/channel-wallet.ts
```diff
@@ -49,32 +50,43 @@ export class ChannelWallet {
...
+    this.messagingService.requestFeed.subscribe(x => this.handleRequest(x));
+  }
+
+  private handleRequest(request: AppRequestEvent) {
+    switch (request.type) {
+      case 'CREATE_CHANNEL':
+      case 'JOIN_CHANNEL': {
         const workflow = this.startWorkflow(applicationWorkflow(this.store, this.messagingService));
         this.workflows.push(workflow);
+        // throw new Error('oops');
```
geoknee: Remove?

packages/xstate-wallet/src/workflows/ethereum-enable.ts
```diff
@@ -0,0 +1,128 @@
...
+  };
+  const actions = {
+    displayUi: () => {
+      sendDisplayMessage('Show');
+    },
+    hideUi: () => {
+      sendDisplayMessage('Hide');
+    },
+    sendResponse: (context: WorkflowContext, event) => {
+      messagingService.sendResponse(context.requestId, context.enabledAddress as string); // todo: typing
```
geoknee: `TODO: ` ?

packages/xstate-wallet/src/workflows/ethereum-enable.ts
```diff
@@ -0,0 +1,128 @@
...
+
+// And then there are services and actions and events and states
+
+// protocol interface:
+//   - methods you can call on the store / messaging service
+//   - actions
+//
+
+// but if you want to build something complex out of these, you still need to give it a display name
+// and then you need to mock these
```
geoknee: Did you mean to include this?

geoknee:
Seems pretty nice: 
Screenshot 2020-03-12 at 14 49 07


1. Can you add this screen to storybook relatively easily?
2. The use must now click twice to get ethereum enabled (once in the app, once in the wallet). I guess this is a better reflection of the current architecture, and more secure (the wallet should not expose sensitive data such as `myStateChannelsSigningKey` without user approval)   Perhaps we should change the button in RPS to be "Connect with StateChannels"?

geoknee:
Would now be the right time to fold `getAddress()`, `getEthereumSelectedAddress()` into a single `channelClient.enable()` API call which triggers your new workflow, including the `window.ethereum.enable()` step?

That way, the UI you have added is protecting the user against inadvertently exposing sensitive data such as `myStateChannelsSigningKey` to an app, at least without user approval. 

To me, that is the real value of having the new user approval screen -- as is, it is merely forcing the user to reassert the intention that they have *just* expressed. (Why would they click "Connect with Metamask" in the app, and then click "No" in the wallet? You could ask the same question about a non state channel Dapp: why click "Connect With Metamask" in the Dapp and not in Metamask... but there is a good answer for that. Metamask owns the sensitive data being requested and should have the final say about releasing it).

packages/xstate-wallet/src/workflows/ethereum-enable.ts
```diff
@@ -0,0 +1,128 @@
...
+    },
+    enabling: {
+      invoke: {
+        src: 'enableEthereum',
+        onDone: {target: 'done', actions: assign({enabledAddress: (context, event) => event.data})},
+        onError: 'failure'
+      }
+    }, // invoke ethereum enable
+    done: {type: 'final', entry: [actions.hideUi, actions.sendResponse]},
+    failure: {type: 'final'}
```
geoknee: Screenshot 2020-03-12 at 15 08 31

This seems unfinished? Maybe we should `actions.hideUI` and signal to the app to go back to a state where it can try again.

tomclose: I've decided to handle this in a separate PR, as it'll include the necessary changes to send an error response back over the json-rpc connection.

tomclose:
> Would now be the right time to fold getAddress(), getEthereumSelectedAddress() into a single channelClient.enable() API call which triggers your new workflow, including the window.ethereum.enable() step?

Sounds like a good idea to me, but I'd rather get this merged in first.

packages/xstate-wallet/src/ui/enable-ethereum-workflow.tsx
```diff
@@ -0,0 +1,31 @@
...
+      style={{
+        textAlign: 'center'
+      }}
+    >
+      <h1>In order to proceed you must connect to MetaMask.</h1>
+      <Button onClick={() => props.send('USER_APPROVES_ENABLE')}>Ok</Button>
+      <Button.Text onClick={() => props.send('USER_REJECTS_ENABLE')}>Cancel</Button.Text>
+    </div>
+  );
+  if (current.value.toString() === 'explainToUser') {
```
geoknee: Is there any way to make this type safe?

PR #1173, opened 2020-03-11
Create and Fund Ledger Workflow

### LINKED ISSUE #1113: Create workflow that opens and directly funds a ledger channel
We need a workflow that opens and directly funds a ledger channel with the hub.

lalexgap:
Adds the `create-and-fund-ledger` workflow and connects it to the `approve-budget-and-fund' workflow.

There are currently no unit tests but most of the functionality is tested by the integration test `approve-budget.test.ts`. 

Visualizer: https://xstate.js.org/viz/?gist=323f68c6c942c881719d9a11b0ced544

`create-and-fund-ledger` should be usable by the hub if it spawns a new `create-and-fund-ledger` workflow when it receives the `FundLedger` objective. 
Similar to https://github.com/statechannels/monorepo/blob/2f5a7cb98478c37164c110b47d355f5d9a204d3f/packages/xstate-wallet/src/integration-tests/approve-budget.test.ts#L32 

Fixes #1113


packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,184 @@
...
+        '': [{cond: 'doesChannelIdExist', target: 'waitForObjective'}, {target: 'initialize'}]
+      }
+    },
+    initialize: {
+      initial: 'initializeChannel',
+      states: {
+        initializeChannel: {
+          invoke: {
+            src: 'initializeChannel',
+            onDone: {target: 'createObjective', actions: []}
```
andrewgordstewart:
```suggestion
            onDone: 'createObjective'
```

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,184 @@
...
+    participants: context.participants,
+    data: {ledgerId: context.ledgerId}
+  };
+  return store.addObjective(objective);
+};
+const getPreFundState = (store: Store): WorkflowServices['getPreFundState'] => async context => {
+  const {latestState} = await store.getEntry(context.ledgerId);
+  return {state: latestState};
+};
+// TODO: Can this be shared with create-and-fund?
```
andrewgordstewart: I would just import this from `create-and-fund`

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,184 @@
...
+  assignChannelId: 'assignChannelId' as any
+};
+export const mockServices: WorkflowServices = {
+  initializeChannel: 'initializeChannel' as any,
+  supportState: 'supportState' as any,
+  createObjective: 'createObjective' as any,
+  depositing: 'depositing' as any,
+  getDepositingInfo: 'getDepositingInfo' as any,
+  getPreFundState: 'getPreFundState' as any
+};
```
andrewgordstewart: Are these needed? (and do they need to be exported?)
```suggestion
```

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export const mockOptions = {services: mockServices, actions: mockActions, guards: mockGuards};
```
andrewgordstewart:
```suggestion
export const mockOptions = {guards: mockGuards};
```

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,184 @@
...
+
+const createObjective = (store: Store): WorkflowServices['createObjective'] => async context => {
+  const objective: Objective = {
+    type: 'FundLedger',
+    participants: context.participants,
+    data: {ledgerId: context.ledgerId}
+  };
+  return store.addObjective(objective);
+};
+const getPreFundState = (store: Store): WorkflowServices['getPreFundState'] => async context => {
```
andrewgordstewart:
```suggestion
const getPreFundState = (store: Store): WorkflowServices['getPreFundState'] => async (context: LedgerIdExists) => {
```

lalexgap: There's no need to type `context` it is already typed by `WorkflowServices['getPreFundState'] `

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,184 @@
...
+};
+
+export const options = (
+  store: Store
+): {actions: WorkflowActions; services: WorkflowServices; guards: WorkflowGuards} => {
+  return {
+    actions: {
+      assignChannelId: assign((context: WorkflowContext, event: DoneInvokeEvent<string>) => {
+        return {...context, channelId: event.data};
+      })
```
andrewgordstewart:
```suggestion
      assignChannelId: assign(channelId: (_, {data}: DoneInvokeEvent<string>) => data)
```

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
lalexgap: I thought they were required by the `generateConfig` to create the config for the visualizer?

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,184 @@
...
+    },
+    guards: {
+      doesChannelIdExist: (context: WorkflowContext) => {
+        return !!context.ledgerId;
+      }
```
andrewgordstewart:
```suggestion
      doesChannelIdExist: (context: WorkflowContext) => !!context.ledgerId
```

lalexgap: If we have a preference this should we enforce it with a linter? (I'm assuming there's a rule that would apply to this)

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
andrewgordstewart: Only guards; if you try to execute with missing `actions` or `services`, then the interpreter just logs warnings

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
lalexgap: Personally I don't love destructured function parameters but I think that may just be that I'm not used to them.

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
andrewgordstewart: I would prefer it in this instance only because it makes the options very clean, and IMO more readable; each line is a single option.

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
andrewgordstewart: In this particular case, the bigger change is between the `assign` parameter:
```
assign({newProperty: (c, e) => 'foo'})
```

vs
```
assign((c, e) => ({...c, newProperty: 'foo'}))

It seems nicer not to have to spread the old context.

File: packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts (hunk 7 above)
Comment (andrewgordstewart) [reply packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts]: In case we do prefer this (I do), we can use [this rule](https://eslint.org/docs/rules/arrow-body-style#options) with `'as-needed'`

File: packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts (hunk 2 above)
Comment (lalexgap) [reply packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts]: I just realized that it's a slightly modified version due to different names of properties on the `context`

File: packages/xstate-wallet/src/store/types.ts [hunk 8]
```diff
@@ -100,8 +100,13 @@ export type FundGuarantor = _Objective<
     guarantorId: string;
   }
 >;
-
-export type Objective = OpenChannel | VirtuallyFund | FundGuarantor;
+export type FundLedger = _Objective<
```
andrewgordstewart: I suppose `OpenChannel` could serve as `FundLedger`.

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
```diff
@@ -95,9 +113,22 @@ export const approveBudgetAndFundWorkflow = (
     updateBudget: (context: WorkflowContext, event) => {
       return store.updateOrCreateBudget(context.budget);
     },
-    createAndFundLedger: (context: WorkflowContext, event) => {
-      // TODO: Hook up to workflow when it exists
-      return Promise.resolve();
+    createAndFundLedger: ({player, hub, budget}: WorkflowContext) => {
+      // TODO: Abstract into function that takes a budget and returns a outcome
+      const ethBudget = budget.budgets[ETH_ASSET_HOLDER_ADDRESS];
```
andrewgordstewart: `budget.budgets` makes it seem like one of `budget` or `budgets` is poorly named -- why does a budget have many budgets?. Maybe `budget.forToken[...]`?

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
```diff
@@ -95,9 +113,22 @@ export const approveBudgetAndFundWorkflow = (
...
+      const playerItem: AllocationItem = {
+        destination: player.destination,
+        amount: ethBudget.pending.playerAmount
+      };
+      const hubItem: AllocationItem = {
+        destination: player.destination,
```
andrewgordstewart: Shouldn't this be the hub's destination?

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
```diff
@@ -110,12 +141,37 @@ export const approveBudgetAndFundWorkflow = (
...
+      await messagingService.sendBudgetNotification(context.budget);
+    },
+    updateBudgetToFree: assign(
+      (context: WorkflowContext): WorkflowContext => {
+        const clonedBudget = _.cloneDeep(context.budget);
+        // TODO: There must be a nicer way of mapping from one record to a new one
+        Object.keys(clonedBudget.budgets).forEach(k => {
+          clonedBudget.budgets[k] = freeAssetBudget(clonedBudget.budgets[k]);
+        });
+        return {...context, budget: clonedBudget};
```
andrewgordstewart:
```suggestion
        const clonedBudget = _.mapValues(context.budget, freeAssetBudget);
        return {...context, budget: clonedBudget};
```

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,173 @@
...
+import {isSimpleEthAllocation} from '../utils/outcome';
+import {add} from '../utils/math-utils';
+
+type WorkflowActions = {
+  assignChannelId: AssignAction<WorkflowContext, DoneInvokeEvent<string>>;
+};
+type WorkflowGuards = {
+  doesChannelIdExist: ConditionPredicate<WorkflowContext, WorkflowEvent>;
+};
+export type WorkflowEvent = AnyEventObject;
```
andrewgordstewart: Do all these types need to be exported?

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,173 @@
...
+  for (let i = 0; i < allocationItems.length; i++) {
+    const {amount} = allocationItems[i];
+    if (i !== myIndex) depositAt = depositAt.add(amount);
+    else {
+      const totalAfterDeposit = depositAt.add(amount);
+      return {channelId: context.ledgerId, depositAt, totalAfterDeposit, fundedAt};
+    }
+  }
+
+  throw Error(`Could not find an allocation for participant id ${myIndex}`);
```
andrewgordstewart: This could be shoehorned to work, something like:

```
    return await CreateAndFund.getDepositingInfo({channelId: context.ledgerId, allocations: allocationItems});
```

though, you end up making two calls to the database.

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
```diff
@@ -0,0 +1,173 @@
...
+  };
+};
+
+export const mockGuards: WorkflowGuards = {
+  doesChannelIdExist: () => true
+};
+
+export const mockOptions = {guards: mockGuards};
+
+export const createAndFundLedgerWorkflow = (store: Store, context: WorkflowContext) =>
```
andrewgordstewart: Does this need to accept a `context`? Tests can do `createAndFundLedgerWorkflow(store).withContext(context)`

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
lalexgap: I've updated it to something similar to your proposal

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
lalexgap: Probably not, although I'm wondering if it might be good form to export `WorkflowEvents/States/Actions` etc. so it's easy to write a UI (or something else) against them.

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
lalexgap: In this case `createAndFundLedger` needs a defined `context` so I think it makes sense to enforce it by requiring it as a parameter.

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
lalexgap: Yup! That should be fixed

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
lalexgap: I've renamed to `forAsset` for now.

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
andrewgordstewart: Right, that makes sense for top-level workflows.

packages/xstate-wallet/src/workflows/create-and-fund-ledger.ts
lalexgap: I think I prefer to keep it separate for now so any changes in `CreateAndFund.getDepositingInfo` don't break an unrelated workflow.

PR #1174, opened 2020-03-11
[WIP] Fix inconsistent pricing 

### LINKED ISSUE #1148: [web3t] Fix inconsistent pricing
If I download a file that claims to "cost" 1300 wei, why do I end up paying far more than that? (Paying a little more is expected and probably ok, since we are using a basic buffered payment system).

spersico:
This PR should close #1148

There were quite a lot of issues that caused the issue:
- The **STOP** and **START** messages were being emitted and handled several times (on response to each blocked request). This provoked that sometimes funds were being transferred more than once. It also affected the process of jumpstarting the download and provoked several inconsistencies.
**Solutions:** 
I added a message interceptor at the wire level, that automates things, avoids message repetition at the leecher side, and adds some encapsulation to the wire.
I also added more checks in the stop() and start() methods to avoid spamming the leecher with messages, if possible.

- _The **REQUEST** interceptor implementation was horrible_, and I hate myself for leaving it as it was 'till now. It was hard to read and understand, and it seems that it added inconsistencies in its execution, provoked, I believe, by the stack not being always the same.
**Solution:**
I now pass a callback with the REQUEST event, that gives the client the power to allow or block a request.

- _The **STOP** and  **START** logic was/is problematic:_ for some reason, as it was (and until an hour ago) if the seeder choked the leecher, but the leecher had still some responses of requests left to process, all those responses got dropped. 
**Solution:**
I make the seeder wait 'till it finishes to process all requests (_by using setTimeout, yes, I hate it, I need to see if I can avoid it_), and only then emit the START event. 

Besides that, I added some more improvements to the codebase:
- Now we can see when a request is responded (by intercepting the _onPiece method).
- I merged the listeners of the **REQUEST** event to a single one so that it gets easier to read, and all requests share the same interface and behaviour.
- I removed unnecessary method calls in the jumpstart function (I believe they were unnecessary, but we don't have the tests that could help check that, like the one that tested if the leecher could start downloading files after 2+ minutes idle).

**Everything seems to be working, but I need to do some more manual testing. With the last basic runs that I made to test this, it worked great, but if someone else wants to give it a run, that would be great.**

### DIFF
--- a/packages/web3torrent/src/library/paid-streaming-extension.ts
+++ b/packages/web3torrent/src/library/paid-streaming-extension.ts
@@ -25,7 +25,9 @@ export abstract class PaidStreamingExtension implements Extension {
 
   pseChannelId: string;
   peerChannelId: string;
+
   isForceChoking = false;
+  isBeingChoked = false;
 
   constructor(wireToUse: PaidStreamingWire) {
     this.wire = wireToUse;
@@ -87,15 +89,19 @@ export abstract class PaidStreamingExtension implements Extension {
   }
 
   stop() {
-    this.isForceChoking = true;
-    this.wire.choke();
-    this.executeExtensionCommand(PaidStreamingExtensionNotices.STOP, this.pseChannelId);
+    if (!this.isForceChoking) {
+      this.isForceChoking = true;
+      this.executeExtensionCommand(PaidStreamingExtensionNotices.STOP, this.pseChannelId);
+    }
   }
 
   start() {
-    this.isForceChoking = false;
-    this.wire.unchoke();
-    this.executeExtensionCommand(PaidStreamingExtensionNotices.START);
+    if (this.isForceChoking) {
+      setTimeout(() => {
+        this.isForceChoking = false;
+        this.executeExtensionCommand(PaidStreamingExtensionNotices.START);
+      }, 0);
+    }
   }
 
   ack() {
@@ -111,11 +117,34 @@ export abstract class PaidStreamingExtension implements Extension {
   onMessage(buffer: Buffer) {
     try {
       const jsonData = bencode.decode(buffer, undefined, undefined, 'utf8');
-      this.messageBus.emit(PaidStreamingExtensionEvents.NOTICE, jsonData);
+      this.messageHandler(jsonData);
     } catch (err) {
-      log('ERROR: decoding', err);
+      log('ERROR: onMessage decoding', err);
       return;
     }
+  }
+
+  protected messageHandler({command, data}) {
+    switch (command) {
+      case PaidStreamingExtensionNotices.ACK:
+        return;
+      case PaidStreamingExtensionNotices.START:
+        log(`START received from ${this.peerAccount}`);
+        this.isBeingChoked = false;
+        this.wire.requests = [];
+        this.wire.unchoke();
+        break;
+      case PaidStreamingExtensionNotices.STOP:
+        log(`STOP received from ${this.peerAccount}`);
+        this.peerChannelId = data;
+        if (this.isBeingChoked) return;
+        this.isBeingChoked = true;
+        break;
+      default:
+        log(`MESSAGE received from ${this.peerAccount}`, data);
+    }
+    this.ack();
+    this.messageBus.emit(PaidStreamingExtensionEvents.NOTICE, {command, data});
   }
 
   protected executeExtensionCommand(command: PaidStreamingExtensionNotices, data = {}) {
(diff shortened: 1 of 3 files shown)


packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -87,15 +89,19 @@ export abstract class PaidStreamingExtension implements Extension {
...
 
   start() {
-    this.isForceChoking = false;
-    this.wire.unchoke();
-    this.executeExtensionCommand(PaidStreamingExtensionNotices.START);
+    if (this.isForceChoking) {
+      setTimeout(() => {
+        this.isForceChoking = false;
+        this.executeExtensionCommand(PaidStreamingExtensionNotices.START);
+      }, 0);
```
geoknee: Is this a timeout for 0 ms?

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -111,13 +117,36 @@ export abstract class PaidStreamingExtension implements Extension {
...
-      log('ERROR: decoding', err);
+      log('ERROR: onMessage decoding', err);
       return;
     }
   }
 
+  protected messageHandler({command, data}) {
+    switch (command) {
+      case PaidStreamingExtensionNotices.ACK:
+        return;
```
geoknee: Do you think we ought to use either `return` or `break` exclusively?

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -111,13 +117,36 @@ export abstract class PaidStreamingExtension implements Extension {
...
+      case PaidStreamingExtensionNotices.START:
+        log(`START received from ${this.peerAccount}`);
+        this.isBeingChoked = false;
+        this.wire.requests = [];
+        this.wire.unchoke();
```
geoknee: Could you explain why the leecher is choking/unchoking? Is it their way of reducing unecessary requests to the seeder?

packages/web3torrent/src/library/paid-streaming-extension.ts
snario: It is 😕

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -111,13 +117,36 @@ export abstract class PaidStreamingExtension implements Extension {
...
-      this.messageBus.emit(PaidStreamingExtensionEvents.NOTICE, jsonData);
+      this.messageHandler(jsonData);
     } catch (err) {
-      log('ERROR: decoding', err);
+      log('ERROR: onMessage decoding', err);
       return;
     }
   }
 
+  protected messageHandler({command, data}) {
```
snario: Perhaps rename to `handleMessage`?

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -111,13 +117,36 @@ export abstract class PaidStreamingExtension implements Extension {
...
+      case PaidStreamingExtensionNotices.START:
+        log(`START received from ${this.peerAccount}`);
+        this.isBeingChoked = false;
+        this.wire.requests = [];
+        this.wire.unchoke();
+        break;
+      case PaidStreamingExtensionNotices.STOP:
+        log(`STOP received from ${this.peerAccount}`);
+        this.peerChannelId = data;
+        if (this.isBeingChoked) return;
```
snario: Why does this case `return` and skip lines 146 and 147?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -27,7 +26,7 @@ export type TorrentCallback = (torrent: Torrent) => any;
 export * from './types';
 
 export const WEI_PER_BYTE = bigNumberify(1); // cost per credit / piece
-export const BUFFER_REFILL_RATE = bigNumberify(100000); // number of credits / pieces the leecher wishes to increase the buffer by
+export const BUFFER_REFILL_RATE = bigNumberify(80000); // number of credits / pieces the leecher wishes to increase the buffer by
```
snario: Perhaps we should put this inside a `constants.ts` file so they're more easily understood to be app-level constants.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -140,77 +138,71 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-        //   'Seeder Balance: ' + knownPeerAccount.seederBalance
-        // );
-      }
-    });
-
-    wire.paidStreamingExtension.once(
+    wire.paidStreamingExtension.on(
       PaidStreamingExtensionEvents.REQUEST,
-      async (index: number, offset: number, size: number) => {
+      async (index: number, _: number, size: number, response: (allow: boolean) => void) => {
```
snario: Since we use `lodash`, maybe we ought not name a local variable `_`.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -343,27 +325,18 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
     }
+
+    log(`JUMPSTART: Wire requests.`, JSON.stringify(wire.requests, null, 1));
+    torrent.pieces.forEach(piece => {
+      if (piece && !!piece._reservations) {
+        piece._reservations--;
+      }
+    });
+    torrent._update();
+    // torrent._startDiscovery(); // useful for long blocked torrents, forces the client to re-start searching for peers
```
snario: Does the client _stop_ searching for peers at some point?

packages/web3torrent/src/library/paid-streaming-extension.ts
geoknee: I read that this is a reasonably common hack, equivalent to some small browser-specific timeout. It's basically the minimum timeout possible.

snario: I saw this once, might explain it https://github.com/feross/queue-microtask

snario: > **Why not use setTimeout(fn, 0)?**
This approach is the most compatible, but it has problems. Modern browsers throttle timers severely, so setTimeout(…, 0) usually takes at least 4ms to run. Furthermore, the throttling gets even worse if the page is backgrounded. If you have many setTimeout calls, then this can severely limit the performance of your program.

andrewgordstewart: What's the purpose of the 0ms timeout here?

geoknee:
I am going to merge this so we can unblock people; the discussion can continue and actions arising should be addressed in followup PRs.

packages/web3torrent/src/library/paid-streaming-extension.ts
spersico: The idea is to let the seeder finish responding all the requests that the leecher made. It dont really like this _"solution"_. I could check the _queue-microtask_ package as a replacement,  or no timeout at all.

packages/web3torrent/src/library/paid-streaming-extension.ts
spersico: Umm.. maybe, the idea was to avoid an unending cycle of ACKs between peers, and repeating code...  but yeah, it gets harder to read. I could change it in another PR later

geoknee: If there's a good reason for the difference, then that's fine and you could probably just add a comment (or not). It just popped out at me when I scanned over the code.

PR #1179, opened 2020-03-11
simple-hub: remove xstate-wallet dependency

kerzhner:
Fixes https://github.com/statechannels/monorepo/issues/1177


kerzhner:
> I assume the code is copy-pasted?

Yes. I started by copy pasting the functions from `xstate-wallet` used by the hub. Then I copy-pasted functions called by that code (+ removing `export` statements).

PR #1188, opened 2020-03-12
Fix master

tomclose:


packages/simple-hub/src/wallet/xstate-wallet-internals/index.ts
```diff
@@ -40,7 +40,7 @@ interface MixedAllocation {
 }
 
 type Allocation = SimpleAllocation | MixedAllocation;
-type Outcome = Allocation | SimpleGuarantee;
+export type Outcome = Allocation | SimpleGuarantee;
```
snario: This definition is at odds with the definition in `nitro-protocol` btw

https://github.com/statechannels/monorepo/blob/5955bb864c0e9e8d089e9d1099f3023e8f6c7a54/packages/nitro-protocol/src/contract/outcome.ts#L38-L39

packages/simple-hub/src/wallet/xstate-wallet-internals/index.ts
```diff
@@ -285,3 +290,19 @@ export function signState(state: State, privateKey: string): string {
...
+
+export const firstState = (
+  outcome: Outcome,
+  {channelNonce, chainId, challengeDuration, appDefinition, participants}: ChannelConstants,
+  appData?: string
+): State => ({
+  appData: appData || '0x',
+  isFinal: false,
+  turnNum: bigNumberify(0),
+  chainId: chainId || '0x01',
```
snario: Maybe the default `chainId` shouldn't be mainnet... just in case

packages/simple-hub/src/wallet/xstate-wallet-internals/index.ts
tomclose: I think the shapes are genuinely different - the xstate wallet `Outcome` is a format that's more convenient to work with internally.

PR #1189, opened 2020-03-12
[web3t] Display UploadInfo unconditionally 

geoknee:
Reflects fact that I can start uploading at any point after I start leeching, and in particular *before* I become a seeder. A seeder is someone who has the whole file.

With the existing condition in place, there's no evidence of the channels that I open to allow peers to download from me (unless I happen to be a seeder).

Screenshot 2020-03-12 at 11 42 55

### DIFF
--- a/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
+++ b/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
@@ -9,7 +9,7 @@ import {WebTorrentContext} from '../../../clients/web3torrent-client';
 export type UploadInfoProps = {
   wires: PaidStreamingWire[];
   channels: Dictionary<ChannelState>;
-  peerType: 'seeder' | 'leecher';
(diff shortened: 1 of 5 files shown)


packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
```diff
@@ -42,19 +42,19 @@ class ChannelsList extends React.Component<UploadInfoProps> {
     );
 
     const uploaded = wire ? wire.uploaded : 0;
-    const peerAccount = wire ? wire.paidStreamingExtension.peerAccount : 'unknown';
+    const peerAccount =
+      pseType === 'leecher' ? channels[channelId].beneficiary : channels[channelId].payer;
```
geoknee: This is more robust against the `wire` getting dropped (something that seems pretty likely to happen when no more requests are being made).

geoknee:
We should think about overhauling the use of 'seeder' throughout the codebase, to avoid confusion.

PR #1190, opened 2020-03-12
[web3t] Clarify logic around "spacer" state

### LINKED ISSUE #1178: Decouple wire choking and sending STOP
**Describe the bug :bug:**
web3torrent currently has a behaviour that allows a little free leeching when leecher is joining the channel.  

See discussions in: https://github.com/statechannels/monorepo/pull/1163#discussion_r391046090

**To Reproduce**
1. Start seeder and leecher in two windows
2. Turn on dev console and filter web3torrent log only
3. See the behaviour described in this [comment](https://github.com/statechannels/monorepo/pull/1163#discussion_r390678890)

**Expected behavior**
Leecher shouldn't be able to start leeching before it makes the payment.

**Additional context**

geoknee:
The "spacer state" is the state channel update with `turnNum = 4`. 

It's awkward to handle, but an important inevitable consequence of the following choices: 

- unidirectional ForceMove payment channel with beneficiary in slot `0`
- vanilla ForceMove turn-taking

At the state channel layer, we should have the following sequence, with uploader as `0` and downloader as `1`:
```
0: createChannel, turnNum = 0
1: joinChannel() turnNum = 1
0: turnNum = 2
1: turnNum = 3 <<-- fake provider updates straight to here on joinChannel()
0: acceptChannelUpdate(), turnNum = 4 <<--- "spacer state"
1: makePayment(), turnNum = 5
0: acceptChannelUpdate(), turnNum = 6
...
1: makePayment(), turnNum = 2n+3
0: acceptChannelUpdate(), turnNum = 2n+4
```

According to this PR, in the web3torrent code, we do not prompt the downloader to make a payment (i.e. send him a STOP message) until we have sent the spacer state. That way, no special logic is required on the downloader's client: they can always respond to this prompt with a payment. 

(Previously we had a bit of a hacky solution where we would treat `turnNum=3` as a payment -- which it isn't -- and accepting this payment triggered the spacer state to be sent. We were also skipping payments on the downloader's client until we received the spacer state).

Closes #1178 .

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -198,7 +198,7 @@ export class PaymentChannelClient {
     }
   }
   // beneficiary may use this method to accept payments
-  async acceptPayment(channelState: ChannelState) {
+  async acceptChannelUpdate(channelState: ChannelState) {
     const {
       channelId,
       beneficiary,
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -6,6 +6,8 @@ import {ChannelStatus} from '@statechannels/client-api-schema';
 import {SiteBudget} from '@statechannels/client-api-schema';
 
 const bigNumberify = utils.bigNumberify;
+const FINAL_SETUP_STATE = 3; // for a 2 party ForceMove channel
+
 export interface ChannelState {
   channelId: string;
   turnNum: string;
@@ -233,7 +235,7 @@ export class PaymentChannelClient {
 
   shouldSendSpacerState(channelState: ChannelState): boolean {
     const turnNum = Number(channelState.turnNum);
-    return turnNum === 3 ? true : false;
+    return turnNum === FINAL_SETUP_STATE ? true : false;
   }
 
   async pushMessage(message: Message<ChannelResult>) {
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -231,6 +231,11 @@ export class PaymentChannelClient {
     return false; // only beneficiary may receive payments
   }
 
+  shouldSendSpacerState(channelState: ChannelState): boolean {
+    const turnNum = Number(channelState.turnNum);
+    return turnNum === 3 ? true : false;
+  }
+
   async pushMessage(message: Message<ChannelResult>) {
     await this.channelClient.pushMessage(message);
     return convertToChannelState(message.data);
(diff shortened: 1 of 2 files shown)


packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -231,6 +231,11 @@ export class PaymentChannelClient {
     return false; // only beneficiary may receive payments
   }
 
+  shouldSendSpacerState(channelState: ChannelState): boolean {
+    const turnNum = Number(channelState.turnNum);
+    return turnNum === 3 ? true : false;
```
junjizhi: Suggestion: instead of hard coding 3 that looks like a magic number, assign it to a constant.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -213,21 +213,28 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-        // returns true for the second postFS if I am the beneficiary
-        // (I need to countersign this state in order for the first payment to be sent)
-      ) {
+      if (this.paymentChannelClient.shouldSendSpacerState(channelState)) {
+        // send "spacer" state
+        await this.paymentChannelClient.acceptChannelUpdate(
+          this.paymentChannelClient.channelCache[channelState.channelId]
+        );
+        wire.paidStreamingExtension.stop(); // prompt peer for a payment
+        return;
```
junjizhi: Generally we shouldn't exit function like this. We only have another if below. use an `else if` is cleaner.

I point it out this because when people reading the code, it's easy to miss the a block that suddenly exits and jumps somewhere, like a `goto` statement. when that happens, people would try adding a statement down below and expect it to execute. but look and behold, it never reaches there. It's anther debugging session. So we should usually `return` at the end.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -237,24 +244,29 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
         );
         this.unblockPeer(torrent.infoHash, wire, wire.paidStreamingExtension.peerAccount);
         // TODO: only unblock if the buffer is large enough
+        return;
```
junjizhi: unnecessary `return`.

junjizhi:
Can the spacer turn num be 1 instead of 3?

geoknee:
> Can the spacer turn num be 1 instead of 3?

It has to be 4. In a 2 party ForceMove channel, state 4 is the first state that is allowed to change the outcome https://protocol.statechannels.org/docs/forcemove-and-nitro/force-move#core-transition-rules

packages/web3torrent/src/library/web3torrent-lib.ts
geoknee: 079f1f2

packages/web3torrent/src/library/web3torrent-lib.ts
geoknee: 079f1f2

geoknee:
> Can the spacer turn num be 1 instead of 3?

400a5ea

packages/web3torrent/src/clients/payment-channel-client.ts
geoknee: 400a5ea

PR #1193, opened 2020-03-12
[w3t] Remove commented out discovery code

snario:
This is commented out, but it is unclear to me why it is here. Maybe we can discuss here.


spersico:
This is code that was used before, that helped the leecher to restart the download if the seeder resumed access after being unresponsive for a long time (more than 1 minute).

The tests that checked if this is needed or not were removed. 
It can be removed with no issues for now, I left it there in case it was needed, while I was dealing with the inconsistent payment issue.

PR #1195, opened 2020-03-12
[PoC] Use local firebase-server on simple-hub

snario:
Simplified and speeds up local development.


packages/simple-hub/.env.development
```diff
@@ -1,5 +1,5 @@
-FIREBASE_PROJECT='rock-paper-scissors-dev-4ff8d'
```
kerzhner: Where is this env var intended to be used?

packages/simple-hub/src/message/firebase-relay.ts
```diff
@@ -13,11 +13,7 @@ const log = logger();
 
 const config = {
   apiKey: process.env.FIREBASE_API_KEY,
-  authDomain: `${process.env.FIREBASE_PROJECT}.firebaseapp.com`,
-  databaseURL: `https://${process.env.FIREBASE_PROJECT}.firebaseio.com`,
-  projectId: process.env.FIREBASE_PROJECT,
```
kerzhner: Good catch!

packages/simple-hub/.env.development
snario: Maybe on staging?

PR #1196, opened 2020-03-12
[web3t] Fix downloaded = 0 in the UI

geoknee:
Currently the `DownloadInfo` component does not display the number of bytes downloaded. 

This was because 

1. TypesScript was telling me that `downloaded` was not a property on my `wire` (when actually it is). 
2. I had thought that perhaps `uploaded` would be the same as `downloaded` on a wire. But it's a duplex connection so I guess my `uploaded` is your `downloaded`. Fine.

### DIFF
--- a/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
+++ b/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
@@ -41,9 +41,16 @@ class ChannelsList extends React.Component<UploadInfoProps> {
         wire.paidStreamingExtension.pseChannelId === channelId
     );
 
-    const transferred = pseType === 'seeder' ? wire.uploaded : wire.downloaded;
-    const peerAccount =
-      pseType === 'leecher' ? channels[channelId].beneficiary : channels[channelId].payer;
+    let transferred: string;
+    let peerAccount: string;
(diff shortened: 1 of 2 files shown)


packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
```diff
@@ -41,17 +41,24 @@ class ChannelsList extends React.Component<UploadInfoProps> {
...
-      pseType === 'leecher' ? channels[channelId].beneficiary : channels[channelId].payer;
+    let transferred: string;
+    let peerAccount: string;
+    if (wire) {
+      transferred = pseType === 'seeder' ? prettier(wire.uploaded) : prettier(wire.downloaded);
+      peerAccount =
+        pseType === 'leecher' ? channels[channelId].beneficiary : channels[channelId].payer;
+    } else {
+      transferred = 'NOWIRE';
+      peerAccount = 'NOWIRE';
```
geoknee: We tend to lose the `wire` when there haven't been any requests recently. This UI should probably be seen as a stop-gap until we have a robust way of dealing with that.

Suggestions: 
i. Maintain some "memory" in this component: do not overwrite the current fields if the props change to `undefined`

PR #1197, opened 2020-03-12
Concurrent app channels

andrewgordstewart:


packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -167,12 +169,42 @@ export class MemoryStore implements Store {
     }
     channelEntry.setFunding(funding);
   }
-  public async getLedger(peerId: string) {
-    const ledgerId = this._ledgers[peerId];
 
-    if (!ledgerId) throw new Error(`No ledger exists with peer ${peerId}`);
+  public async lockLedger(participantId: string): Promise<LedgerLocked> {
```
snario: Usually the term used is `acquire` so maybe `acquireLockForLedgerChannel` is a better name?

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -67,16 +68,24 @@ interface InternalEvents {
...
+
 export class MemoryStore implements Store {
   readonly chain: Chain;
   protected _channels: Record<string, MemoryChannelStoreEntry | undefined> = {};
   private _objectives: Objective[] = [];
   private _nonces: Record<string, BigNumber | undefined> = {};
   private _eventEmitter = new EventEmitter<InternalEvents>();
   private _privateKeys: Record<string, string | undefined> = {};
-  protected _ledgers: Record<string, string | undefined> = {};
+  protected _ledgers: Record<string, LedgerStatus | undefined> = {};
```
snario: Perhaps name this more explicitly as `_ledgerChannelLockStatuses`?

packages/xstate-wallet/src/store/store.ts
```diff
@@ -18,14 +18,16 @@ export interface Store {
     appDefinition?: string
   ): Promise<ChannelStoreEntry>;
   getEntry(channelId): Promise<ChannelStoreEntry>;
+
+  ledgerFeed: Observable<LedgerUpdated>;
+  lockLedger(participantId: string): Promise<LedgerUpdated>;
```
snario: If we ever wanted to allow for replace-by-incentive apps in the wallet, we'll also need some sort of lock mechanism on app channels I think. Should we implement a more general locking API on a channel ID instead of specifically for ledger channels?

packages/xstate-wallet/src/store/store.ts
```diff
@@ -18,14 +18,16 @@ export interface Store {
...
+  releaseLedger(status: LedgerStatus): Promise<void>;
   getLedger(peerId: string): Promise<ChannelStoreEntry>;
-  // TODO: This is awkward. Might be better to set the funding on create/initialize channel?
+
   setFunding(channelId: string, funding: Funding): Promise<void>;
-  // TODO: I don't know how the store is mean to send outgoing messages.
-  // But I need one, in order to implement virtual funding.
   addObjective(objective: Objective): void;
-  // TODO: should this be exposed via the Store?
```
snario: Did we decide it should be?

packages/xstate-wallet/src/store/memory-store.ts
andrewgordstewart: It's also the lookup for the `ledgerId` for a given `participantId`

packages/xstate-wallet/src/workflows/create-and-fund.ts
```diff
@@ -133,7 +147,20 @@ export const config: MachineConfig<Init, any, any> = {
   }
 };
 
-const services = (store: Store) => ({
+const acquireLock = (store: Store) => async (): Promise<LedgerUpdated> =>
+  await store.ledgerFeed
+    .pipe(
+      filter(s => s.participantId === HUB.participantId),
+      first(s => !s.lock),
+      map(async s => await store.lockLedger(s.participantId))
```
snario: I think you can omit `async` and `await` here

packages/xstate-wallet/src/store/memory-store.ts
snario: Is that necessary to cache assuming it is inside the `MemoryChannelStoreEntry`?

packages/xstate-wallet/src/store/store.ts
andrewgordstewart: That might be a good idea; it's probably worth separating the `ledgerId` lookup from the ledger channel's lock status

andrewgordstewart:
This lock strategy does not prevent deadlocks:
## Scenario:
- A, B concurrently to fund C1, C2 virtually, through H
- A has L1 open with H
- B has L2 open with H

## Notation: below, step1 & step2 together form an atomic action.
```
n. {
  a. step1
  b. step2
}
```

## Sequence
1. {
  a. A acquires lock on L1
  b. A triggers `virtuallyFund(C1)`. This creates joint channel J1.
  c. A is blocked on jointChannelUpdate(J1)
}
2. {
  a. B acquires lock on L2
  b. B triggers `virtuallyFund(C2)`. This creates joint channel J2.
  c. B is blocked on jointChannelUpdate(J2)
}

[A, B] are deadlocked -- neither will support the other's `jointChannelUpdate`

packages/xstate-wallet/src/store/memory-store.ts
andrewgordstewart: I don't quite follow.

On master, `_ledgers` serves this purpose:
1. I want to virtually fund `C` through `H`
2. I get to the step where I need to indirectly fund my guarantor
3. I need to find the `ledgerId` with the hub

The `ledgerId` is not stored on any `channelStoreEntry` until it is determined, and `_ledgers` is the discovery mechanism.

I think `_ledgers` should probably remain unchanged, and I can just add a `_locks: Record<channelId, Guid>` lookup.

andrewgordstewart:
Closing in favour of https://github.com/statechannels/monorepo/pull/1200

PR #1198, opened 2020-03-12
[web3t] Remove incorrect usages of 'seeder'

geoknee:
These mostly arise out of the (tempting but incorrect) assumption that anyone who uploads is a seeder.

### DIFF
--- a/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
+++ b/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
@@ -9,7 +9,7 @@ import {WebTorrentContext} from '../../../clients/web3torrent-client';
 export type UploadInfoProps = {
   wires: PaidStreamingWire[];
   channels: Dictionary<ChannelState>;
-  pseType: 'seeder' | 'leecher';
+  participantType: 'payer' | 'beneficiary';
 };
 
 class ChannelsList extends React.Component<UploadInfoProps> {
@@ -19,7 +19,7 @@ class ChannelsList extends React.Component<UploadInfoProps> {
     channelId: string,
     channels: Dictionary<ChannelState>,
     wires: PaidStreamingWire[],
-    pseType: 'seeder' | 'leecher'
+    participantType: 'payer' | 'beneficiary'
   ) {
     let channelButton;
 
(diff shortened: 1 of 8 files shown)


geoknee:
@spersico we currently have 

https://github.com/statechannels/monorepo/blob/5ece4ea6918d15ab60e6fa309c6aef1980b9fe5e/packages/web3torrent/src/types.ts#L3-L13

Which doesn't do as much justice as I would like to this issue. 

Should we have another status such as `Downloading&Uploading` ? Or rename `Downloading`?  And add this to to `UploadingStatuses`?

spersico:
@geoknee Ummm I think Downloading it's a valid state (for example when there are only one seeder and only one leecher, the leecher will only download). 

Also, that status was UI related, mostly (that's why it's not in the library, but on the client, and it's maintained by the torrent-status-checker utils). 

I like the idea of adding a status of Downloading&Uploading and add that to UploadingStatuses.

And we should stop showing one OR the other, and allow for **both sections** to be shown at the same time: https://github.com/statechannels/monorepo/blob/7343f348409fdb7717de2f2e0bcb9d4b69013493/packages/web3torrent/src/components/torrent-info/TorrentInfo.tsx#L40-L51

geoknee:
@spersico Sounds good, I can address this in a future PR.

The current state of the UI is to *always* show `UploadInfo`, and show `DownloadInfo` conditional on the download status. The plan would be to independently show both, but each has its own condition on the status. And also to introduce a `Downloading&Uploading` status that will activate both components.

packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
```diff
@@ -9,7 +9,7 @@ import {WebTorrentContext} from '../../../clients/web3torrent-client';
 export type UploadInfoProps = {
   wires: PaidStreamingWire[];
   channels: Dictionary<ChannelState>;
-  pseType: 'seeder' | 'leecher';
+  participantType: 'payer' | 'beneficiary';
```
tomclose: Wondering whether `data-sender` and `data-receiver` would be even clearer?

geoknee: This component is about Channels primarily, though. And these are the terms we are using in the `PaymentChannelClient`.

PR #1201, opened 2020-03-12
Add Close Ledger and Withdraw Workflow

### LINKED ISSUE #1109: Simple Withdrawal
The withdraw button only becomes active if all channels are closed. Clicking this button will trigger a closeLedgerChannel command, which will run the closeLedgerAndWithdraw protocol in the wallet. All funds will be returned to user and hub.

lalexgap:
Fixes #1109 

Adds a workflow that closes a ledger channel and withdraws the funds. The workflow is complete but an UI needs to be implemented (#1245)


packages/xstate-wallet/src/store/memory-channel-storage.ts
```diff
@@ -68,6 +68,18 @@ export class MemoryChannelStoreEntry implements ChannelStoreEntry {
...
+    return this.isSupported && this.supported.isFinal;
+  }
+
+  get finalizationProof(): SignedState[] {
+    if (!this.isFinalized) {
+      throw new Error('Channel is not finalized');
+    }
+    return this.sortedByDescendingTurnNum
+      .filter(s => s.isFinal)
+      .map(s => ({...this.channelConstants, ...s}));
```
andrewgordstewart: Shouldn't this just return `this.supported` when `this.isFinalized` is true?

lalexgap: I think that only works when everyone signs a single final state. If we're dealing with the case where everyone is signing their own state I think we need to provide both states in the finalization proof.

andrewgordstewart: Right.

There's no guarantee that these states form a support proof though:
- `this.supported` might be supported by signatures on _it_
- It could also depend on signatures on another state (probably just the previous one)

There should be a `this.support` method which returns the support.


Also, I would recommend that the `conclude-channel` protocol has everyone sign the same final state.
- It's simple enough to reason about
- It means we can use `supportState`
- it's more gas-efficient: less call data, no `validTransition` calls, and no `whoSignedWhat` logic

lalexgap: Yeah, I should have tagged this with a `TODO`. This isn't a proper check but relies on participants playing nicely.

snario:
What's the benefit of `closeAndWithdraw` as an API method over `close` and `withdraw`?

lalexgap:
> What's the benefit of `closeAndWithdraw` as an API method over `close` and `withdraw`?

Is there a case where'd you want to close your ledger channel but not withdraw the funds?

packages/xstate-wallet/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "allowSyntheticDefaultImports": true,
     "baseUrl": ".",
     "esModuleInterop": true,
-    "lib": ["es6", "dom"],
+    "lib": ["es6", "ES2019", "dom"],
```
lalexgap: This lets us use nice things like `Array.flatmap`

packages/client-api-schema/src/types.ts
```diff
@@ -181,6 +175,10 @@ export type GetBudgetResponse = JsonRpcResponse<SiteBudget | {}>;
 
 export type ApproveBudgetAndFundRequest = JsonRpcRequest<'ApproveBudgetAndFund', BudgetRequest>;
 export type ApproveBudgetAndFundResponse = JsonRpcResponse<SiteBudget>;
+
+export type CloseAndWithdrawParams = {site: string; player: Participant; hub: Participant};
```
lalexgap: Should  `site` just be pulled from the domain the json-rpc request originates from?

tomclose: Probably. It seems like we should be able to figure out which `player` we are too?

packages/xstate-wallet/src/workflows/close-ledger-and-withdraw.ts
```diff
@@ -0,0 +1,157 @@
...
+      }
+    },
+
+    createObjective: {
+      invoke: {
+        src: 'createObjective',
+        onDone: {target: 'closeLedger', actions: ['assignLedgerId']}
+      }
+    },
+    closeLedger: getDataAndInvoke({src: 'getFinalState'}, {src: 'supportState'}, 'withdraw'),
```
tomclose: I was thinking about this pattern the other day. I was wondering whether an alternative would be to `getFinalState` earlier in the process and store this data on the workflow (e.g. it seems like we'd probably want to display this on the UI). It would probably also make the error handling easier in the case where you can't `getFinalState`, for whatever reason.

packages/client-api-schema/src/types.ts
lalexgap: I've created #1258 to make sure our API applies this across the board.

PR #1202, opened 2020-03-12
[web3t] Detect public wifi / firewall / lack of internet

### LINKED ISSUE #1129: 2. Investigate public wifi situation
**Problem description :confused:**
On a cafe or an airport, where the firewall is stricter, it would be nice to let the user know that web3torrent might not work correctly.

**Describe the solution you'd like :rocket:**
We should research for a way to test the connection initially, and show a banner or some kind of UI element to the user.

**Describe alternatives you've considered**
We could use some of the legal/open/free torrents that we use in the home page of web3torrent, start the download to see if the client can reach the seeder and if it can't, show a banner or something.

geoknee:
The simple idea is to try and download the example torrent used on the webtorrent.io homepage when we enable the `WebTorrentPaidStreamingClient`. As soon as we get a WIRE event, cancel and raise a success flag. If this doesn't resolve in 3 seconds, raise a failure flag. 

We can hook this into another `ConnectionBanner` type component that informs the user: "Looks like you are behind a firewall or do not have an internet connection".

I tested this with my home WiFI on/off. 

TODO 
- test on public wifi / with typical firewall or blocked ports
- is a WIRE event enough of a signal that things should work?

Closes #1129

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -59,8 +59,6 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     log('got ethereum address');
     log('ACCOUNT ID: ', this.pseAccount);
     log('THIS address: ', this.outcomeAddress);
-    const canTorrent = await this.testTorrentingCapability(3000);
-    console.log(canTorrent ? 'Can Torrent!' : 'Cannot torrent');
   }
 
   async testTorrentingCapability(timeOut: number) {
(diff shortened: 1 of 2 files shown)


geoknee:
No Wifi: 
Screenshot 2020-03-13 at 08 24 59

Wifi + refresh page:

Screenshot 2020-03-13 at 08 34 40

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -60,6 +61,22 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+        torrent.once('wire', wire => {
+          this.remove(torrent.infoHash);
+          resolve(true);
+        });
+      });
+    });
+    const timer = new Promise(function(resolve, reject) {
+      setTimeout(resolve, timeOut);
+    });
+    return Promise.race([gotAWire, timer]);
```
geoknee: If you are on a slow computer / connection, you might see the warning unnecessarily. 

Increasing the timeout would fix that, but you don't see the warning until the timeout lapses.

PR #1203, opened 2020-03-12
Add back test to cover the completed torrent download case

junjizhi:
This is one part of https://github.com/statechannels/monorepo/issues/1182

There was another subtle bug that prevents the torrent to finish downloaded: Both the seeder and leecher was never `enabled` or initialized, so seeder's `mySigningAddress` var is always null, and never be able to accept payments, and the leecher is permanently blocked after turnNum 4.

I haven't got the time to refactor the code and add more tests yet, but at least we have something to cover the full case for now.

### DIFF
--- a/packages/web3torrent/src/library/client-seeding.test.ts
+++ b/packages/web3torrent/src/library/client-seeding.test.ts
@@ -81,14 +81,23 @@ describe('Seeding and Leeching', () => {
     });
   }, 10000);
 
-  it('should unchoke when seeder unblocks leecher', done => {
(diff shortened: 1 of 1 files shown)


junjizhi:
The function name `enable` is a bit deceptive because other tests didn't fail and only impacts the leecher from the second transfer request in a subtle way.

geoknee:
We could throw an error inside the client if you try to do something before successfully calling `enable`?

junjizhi:
Yeah that's doable. I will add a check in my next PR.

PR #1206, opened 2020-03-13
Update ChannelProviderInterface implementations

snario:


packages/channel-client/src/channel-client.ts
```diff
@@ -103,9 +103,17 @@ export class ChannelClient implements ChannelClientInterface<ChannelResult> {
     return this.provider.send('ApproveBudgetAndFund', {
       playerAmount,
       hubAmount,
-      playerOutcomeAddress,
-      hubAddress,
-      hubOutcomeAddress
+      site: 'abc.com', // TODO: use real value
```
snario: I suppose it is fine to just do `document.domain` here.

PR #1207, opened 2020-03-13
Add more web3torrent tests

junjizhi:
Fixes https://github.com/statechannels/monorepo/issues/1182

Added a test to cover the case of one seeder and multiple leechers.


packages/web3torrent/src/library/client-seeding.test.ts
```diff
@@ -6,44 +6,47 @@ import {
...
+    leecherA = await defaultClient();
+    leecherB = await defaultClient();
+  });
+
+  it('should throw when the client is not enabled', async done => {
+    await seeder.disable();
+    expect(() => {
+      seeder.seed(defaultFile as File, defaultSeedingOptions(false));
+    }).toThrow();
+    done();
```
geoknee: 👍

packages/web3torrent/src/library/client-seeding.test.ts
```diff
@@ -74,38 +77,63 @@ describe('Seeding and Leeching', () => {
...
+    let finishCount = 0;
+    async function downloadFinished() {
+      finishCount += 1;
+      if (finishCount == 2) {
+        done();
+      }
+    }
+
+    leecherA.once(ClientEvents.TORRENT_DONE, downloadFinished);
+    leecherB.once(ClientEvents.TORRENT_DONE, downloadFinished);
```
junjizhi: I don't really like putting more logic into the test cases. But typescript doesn't have event listener that can run n times only, e.g., `seeder.twice(EVENT, ...)`, and we decided to stick to the callback style tests, so 🤷‍♂ .

geoknee: Presumably you will eventually add an assertion on `finishCount`? Might it be nicer to keep a separate flag for each leecher? It might make debugging easier.

junjizhi: Currently it calls `done()` when finishCount reaches 2, so it is equivalent to asserting on finishCount. 

Keeping a separate flag inside a leecher won't  get rid of this problem, because we still have to wait for both flag to turn true to finish the test.

spersico:
I don't get the whole `ensureEnabled` thing:
Is a realistic case a seeder disabling his own client?.
All the checks are done in the seeder (basically, the one who could get payed for seeding).
Wouldn't in any case, be better to perform those checks in the `add()` and `setup()` methods?

junjizhi:
@spersico `ensureEnabled()` is called inside `seed()` and `add()`. And it is protected, so it is implementation details and not visible to the client user. 

> Is a realistic case a seeder disabling his own client?
We could imagine seeder / leecher can turn on/off their web3torrent in the future UI? I think that's reasonable to anticipate a use case like this. 

Besides, it's just odd that we have `enable` but not `disable`.

junjizhi:
I think in general we should write tests that are resistant to changes, because the implementation can/will change all the time, but our tests are not supposed to (unless we change the class interface) and should help us guard against regressions.

To achieve that, we shouldn't couple the tests with implementation details. The fact that we are accessing `peersList` of the client directly is against the OO principles. We can live with that for now, but it's good to keep that in mind.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -116,6 +126,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
   }
 
   blockPeer(torrentInfoHash: string, wire: PaidStreamingWire, peerAccount: string) {
+    this.ensureEnabled();
```
spersico: Are you sure this check should be here?
`blockPeer` and `unblockPeer` are run exclusively on a sharing peer.

So, by doing this, you are inserting checks that would disrupt the flow of messages, on a place that (I think) we really shouldn't. This would force the seeder **to not** let the leecher know that it won't respond to more requests, no matter how many funds it sends.

I think that a much better way of doing this would be that, on disabling the client, the client **blocks all it's peers, for all its torrents**, sending a BLOCK event of some sort, maybe with a special message (or with the STOP event, without the channelId as an argument that it sends right now) letting the leechers know that they shouldn't send more funds.

**Another way** is that that block gets done on a try-catch block inside the `blockPeer`, by handling the thrown error.
That way, if the user disables it's client, on `blockPeer `you could set the channelId as null or something, so that eventually when the funds of the leecher dry out, the seeder blocks new funds. And on `unblockPeer`, you just block the unblocking, because the user disabled its own funds handling capabilities.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -126,6 +137,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
   }
 
   unblockPeer(torrentInfoHash: string, wire: PaidStreamingWire, peerAccount: string) {
+    this.ensureEnabled();
```
spersico: Same that the other comment.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: > blocks all it's peers, for all its torrents
this would also involve adding the handling of STOP events that don't have the `channelID `set as parameter.

junjizhi: You raise a good point.

> This would force the seeder to not let the leecher know that it won't respond to more requests, no matter how many funds it sends. 

Assuming you are on master branch, if the seeder suddenly closes the browser, would the leecher end up in the same place?

> Another way is that that block gets done on a try-catch block inside the blockPeer, by handling the thrown error.

I like the second way.

junjizhi: I think the situation is a bit like scenario described in https://github.com/statechannels/monorepo/issues/1215, where we need to handle peer failure (e.g., unresponsive wire, wifi going offline, browser crashing, etc.). 

We may have a "nice" peer that notifies us with a  `BLOCK` message or some sort, but we always need the logic to handle peers that purposely go offline or drop wires with malicious intent. This type of failure should probably be handled ultimately in the `ForceMove` contract / blockchain level. 

Anyway, with the discussions in https://github.com/statechannels/monorepo/issues/1215 going on, we would devise a failure handling strategy. I will take out the checks for `(un)blockPeer()` for now. The checks in `seed()` and `add()` will be still kept because these are the first calls before peers setting the wire, so it is reasonable to add there.

spersico: I agree, this is a not so simple matter, and probably requires further discussion, to implement a wire dropped handler and stuff like that.

The seed and add checks look fine to me.

junjizhi:
@geoknee @snario Can you guys merge?

PR #1208, opened 2020-03-13
[web3t] Introduce prettyPrintWei() function

geoknee:
This is an experiment and I'm not 100% sure if we want to do it:
Screenshot 2020-03-13 at 15 21 56

Downsides: 

- increased "change blindness" (not so easy to see you are actually making payments when the numbers are big)
- rounding is not done correctly (not a big deal and definitely fixable with a bit more work).


spersico:
> increased "change blindness" (not so easy to see you are actually making payments when the numbers are big)

this could be avoided 'till a certain point by having an "on Hover" state that showed the user the exact amount of Wei, unconverted. Or by clicking on mobile screens.

I really like this, for big torrents the Wei quantities can get hard to read.

Issue #1135, opened 2020-03-06
8. Use real / xstate-wallet


spersico:
**Problem description :confused:**
The idea is to start migrating the web3torrent app to see if it can be used in a more realistic environment.

geoknee:
Trying this out as of f9e4d8843cb1d81433130ca61a077a7bd9c44bc2, I get a validation error: 

![Screenshot 2020-03-16 at 09 53 06](https://user-images.githubusercontent.com/1833419/76744472-3023c180-676c-11ea-989e-63c107636e4e.png)


in `client-api-schema/src/parse-request.ts`. It's not the most useful error message...

geoknee:
#1264

geoknee:
As of 9c6b6ed75a8f59c76e9d32b038d1d8fe10af297f I get the following error message: 
![Screenshot 2020-03-21 at 11 33 32](https://user-images.githubusercontent.com/1833419/77225535-ea9a3680-6b67-11ea-83a7-1fa9487f3f54.png)

geoknee:
```
errors.js:76 Uncaught (in promise) Error: cannot convert null value to array (arg="value", value=null, version=4.0.45)
    at Object.throwError (errors.js:76)
    at arrayify (bytes.js:45)
    at Object.splitSignature (bytes.js:248)
    at Object.getSignerAddress (state-utils.ts:66)
    at MemoryChannelStoreEntry.addState (memory-channel-storage.ts:144)
    at memory-store.ts:365
    at Array.forEach (<anonymous>)
    at MemoryStore.<anonymous> (memory-store.ts:365)
    at Generator.next (<anonymous>)
    at memory-channel-storage.ts:178
```

PR #1219, opened 2020-03-16
[web3t[[xstate-wallet] web3t with xstate wallet

geoknee:
As an intermediate goal, I'm trying to get web3t working with the xstate wallet using *direct funding*. Hopefully this will iron out some of the issues that might arise with the differences in API implementation versus the `FakeProvider`, which is what we have been building against so far. 

To be clear, ultimately we will use *virtual funding* (and associated concepts such as the hub and budgets etc) and the wallet will not show as much UI.

Progress toward #1135


geoknee:
The first issue I hit was that `onChannelProposed ` is not supported in the xstate wallet. If that is not an easy fix, we can revert to using `onChannelUpdated` for this step, too. 

(In that case we should remove it from the API altogether).

geoknee:
Closing in favour of #1231

PR #1221, opened 2020-03-16
[monorepo] Run syncpack and yarn

snario:
Ran these commands in order:

```bash
npx syncpack set-semver-ranges                                                                          
npx syncpack format                                                                                     
npx syncpack fix-mismatches                                                                             
yarn 
```

**fix-mismatches**
Ensure that multiple packages requiring the same dependency define the same version, so that every package requires eg. react@16.4.2, instead of a combination of react@16.4.2, react@0.15.9, and react@16.0.0.

**format**
Organise package.json files according to a conventional format, where fields appear in a predictable order and nested fields are ordered alphabetically. Shorthand properties are used where available, such as the "repository" and "bugs" fields.

**set-semver-ranges**
Ensure dependency versions used within "dependencies", "devDependencies", and "peerDependencies" follow a consistent format.


Alonski:
@snario Maybe we can run these in CI and if there are changes fail the build?

kerzhner:
> Maybe we can run these in CI and if there are changes fail the build?

@Alonski Is the PR CI run not sufficient?

Alonski:
@kerzhner Not sure. Are we already doing it? :)

kerzhner:
https://circleci.com/workflow-run/8b9a178e-6b76-4101-abed-fe0d94b4e38d

PR #1222, opened 2020-03-16
channel-client: Add multiple channels tests

junjizhi:
Fixes https://github.com/statechannels/monorepo/issues/1107


packages/channel-client/tests/constants.ts
```diff
@@ -2,6 +2,7 @@ import {AddressZero, Zero, One} from 'ethers/constants';
 
 export const PARTICIPANT_A = '0xeF3c144c6F1B974AcDD5f7AF18057DEc22B2DCe5';
 export const PARTICIPANT_B = '0x56C6ded44Acd7c8643e1501b47285Ad5d722225B';
+export const PARTICIPANT_C = '0xe1501b47285Ad5d722225B56C6ded44Acd7c8643';
```
geoknee: Maybe use `PEER` instead of `PARTICIPANT`?

PR #1223, opened 2020-03-16
Get xstate wallet to send ChannelProposed, instead of ChannelUpdated

tomclose:
https://github.com/statechannels/monorepo/pull/1219#issuecomment-599460805


geoknee:
I wanted to check that this doesn't break `rps` before merging; that is currently broken for (I think) unrelated reasons. Will merge this when things become clearer.

PR #1236, opened 2020-03-17
Web3torrent: Improve channel closing UI and workflow

junjizhi:
Fixes https://github.com/statechannels/monorepo/issues/1216

~~WIP: Auto close channel when leecher runs out of funds~~

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -205,6 +205,9 @@ export class PaymentChannelClient {
           beneficiaryOutcomeAddress,
           payerOutcomeAddress
         );
+      } else {
+        console.error('Insufficient fund to make a payment. Closing channel.');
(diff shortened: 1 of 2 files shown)


geoknee:
It might be nice if the auto-closing was a part of this PR. That way, we won't go through an intermediate state on `master` where we can't close a channel if we run out of funds.

junjizhi:
Yeah I had it working locally. Just trying to polish it a bit more.

junjizhi:
I was trying to work on the UI to show leechers insufficient fund, but based on [this comment](https://github.com/statechannels/monorepo/issues/1216#issuecomment-600163373), we may not have this situation in the first place, so I stopped there. It may deserve its own ticket.

My latest commit simply closes the channel. The seeder will see the channel is closed. The leecher would see the progress bar being stuck and the channel is closed. Again, the more user friendly UI work may be in another ticket.

Issue #1237, opened 2020-03-17
Deposits start before both players accept joining/creating a channel


lalexgap:
It seems that after player A selects create channel they're prompted to deposit, even if player B has not chosen to join the channel yet.

andrewgordstewart:
It's because of [this commit](https://github.com/statechannels/monorepo/commit/870455dda98029400b2f7407512cb9fefb4dae76), which explains its reasoning.

We probably want to revisit that reasoning. Technically, even if the app is malicious, the joiner's funds are not at risk, since the pop-up prevents the joiner from depositing. But, it does aggravate the creator, who might end up depositing before the joiner refuses to join the channel. Therefore, the wallet should, as a courtesy, refrain from signing the first state until the user clicks ok.

We should defer this issue since it is not a part of the workflow hit by web3torrent.

lalexgap:
This has been resolved by other changes.

PR #1239, opened 2020-03-17
Handle duplicate Join Channel

### LINKED ISSUE #1235: Join channel UI does not work with latest Create and Fund changes

lalexgap:
Fixes #1235 

Currently `rps` will call `JoinChannel` any time it receives a `ChannelUpdated` event. Right now it seems that we're sending out multiple `ChannelUpdated` events (possibly related to #1237)

The `xstate-wallet` ends up creating multiple workflows for each `JoinChannel` request, ending up with multiple `Application` workflows running, causing issues with the UI.

To safeguard this I've added deterministic workflowIds for some of the events so we can detect the duplicate call and avoid starting a workflow.

Longer term:
-`RPS` should probably be more careful about making duplicate calls 
- The `xstate-wallet` should not be sending out duplicate `ChannelUpdated` events (this might be resolved by #1237?)

### DIFF
--- a/packages/xstate-wallet/src/channel-wallet.ts
+++ b/packages/xstate-wallet/src/channel-wallet.ts
@@ -53,14 +53,41 @@ export class ChannelWallet {
     this.messagingService.requestFeed.subscribe(x => this.handleRequest(x));
   }
 
+  private isWorkflowIdInUse(workflowId: string): boolean {
+    return this.workflows.map(w => w.id).indexOf(workflowId) > -1;
+  }
+
+  // Deterministic workflow ids for certain workflows allows us to avoid spawning a duplicate workflow if the app sends duplicate requests
+  private calculateWorkflowId(request: AppRequestEvent): string {
+    switch (request.type) {
+      case 'JOIN_CHANNEL':
+        return `${request.type}-${request.channelId}`;
+      case 'APPROVE_BUDGET_AND_FUND':
+        return `${request.type}-${request.player.participantId}-${request.hub.participantId}`;
+      default:
+        return `${request.type}-${Guid.create().toString()}`;
+    }
+  }
   private handleRequest(request: AppRequestEvent) {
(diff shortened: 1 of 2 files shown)


geoknee:
RE: `rps`.  We make that API call in a saga every time a redux action is dispatched, as long as the channel status is 'proposed':

https://github.com/statechannels/monorepo/blob/855d07e75287daf15405a8dff78fdf2e55dd689c/packages/rps/src/redux/game/saga.ts#L86-L87

I suppose previous versions of the wallet, as well as the fake provider, may have changed the status of the channel more rapidly, preventing any more API calls being made? (And that might explain why things worked better then?)

From the saga's point of view, the API call is blocking. If the channel was guaranteed to be progressed out of 'proposed' when the `joinChannel` call resolves, then (I think) we wouldn't make multiple calls unnecessarily (except in case of error)?

lalexgap:
> RE: `rps`. We make that API call in a saga every time a redux action is dispatched, as long as the channel status is 'proposed':
> 
> https://github.com/statechannels/monorepo/blob/855d07e75287daf15405a8dff78fdf2e55dd689c/packages/rps/src/redux/game/saga.ts#L86-L87
> 
> I suppose previous versions of the wallet, as well as the fake provider, may have changed the status of the channel more rapidly, preventing any more API calls being made? (And that might explain why things worked better then?)
> 
> From the saga's point of view, the API call is blocking. If the channel was guaranteed to be progressed out of 'proposed' when the `joinChannel` call resolves, then (I think) we wouldn't make multiple calls unnecessarily (except in case of error)?

Hmm, that's a good point. Perhaps we should be sending a `CreateChannel/JoinChannel` response only once the channel is funded?  Also related to this is #1241 where `RPS` was missing a `ChannelUpdated` event since it was blocked on `JoinChannel`. 

I think the two PRs can go in as is for now but I've created #1243 to consider when we should send out the responses/channel updated.

PR #1242, opened 2020-03-17
Add EnableEthereum API method

snario:


packages/app-wallet-interface/source/index.html.md
```diff
@@ -109,6 +109,29 @@ Note: we don't return the state of the channel, as messages are not necessarily
...
+}
+```
+
+> Example response
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {"success": true}
```
tomclose: Might be worth sending back the selectedAddress here. That's what happens with `window.ethereum`.

packages/xstate-wallet/src/messaging.ts
```diff
@@ -134,17 +134,22 @@ export class MessagingService implements MessagingServiceInterface {
     const {id: requestId} = request;
 
     switch (request.method) {
+      case 'EnableEthereum':
+        if (this.store.chain.ethereumIsEnabled) {
+          window.parent.postMessage(jrs.success(requestId, {success: true}), '*');
+        } else {
+          this.eventEmitter.emit('AppRequest', {type: 'ENABLE_ETHEREUM', requestId});
+          // TODO: @tom should it send {success: true} at this point?
```
tomclose: No - the workflow would need to do that, otherwise it would return success even if the user doesn't approve. If we decide to return the selected address, then we won't need to change the existing workflow.

PR #1244, opened 2020-03-17
Add budgets to create-and-fund workflow.

andrewgordstewart:
This adds a state `reservingFunds` before invoking the virtual-funding machine. This state invokes `reserveFunds`, which calls a new store method `reserveFunds`.

### DIFF
--- a/packages/xstate-wallet/package.json
+++ b/packages/xstate-wallet/package.json
@@ -9,6 +9,7 @@
     "@statechannels/wire-format": "0.0.1",
     "@xstate/react": "0.8.1",
     "ethers": "4.0.45",
+    "async-lock": "1.2.2",
     "eventemitter3": "4.0.0",
     "filter-async-rxjs-pipe": "0.1.5",
     "guid-typescript": "1.0.9",
--- a/packages/xstate-wallet/src/serde/app-messages/serialize.ts
+++ b/packages/xstate-wallet/src/serde/app-messages/serialize.ts
@@ -9,19 +9,25 @@ import {
   AllocationItem,
   SimpleAllocation,
   SiteBudget,
-  BudgetItem
+  BudgetItem,
+  AssetBudget
 } from '../../store/types';
 import {tokenAddress} from '../../constants';
 import {AddressZero} from 'ethers/constants';
+import {checkThat, exists} from '../../utils';
 
 export function serializeSiteBudget(budget: SiteBudget): AppSiteBudget {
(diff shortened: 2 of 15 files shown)


packages/xstate-wallet/src/serde/app-messages/serialize.ts
```diff
@@ -9,19 +9,25 @@ import {
...
 export function serializeSiteBudget(budget: SiteBudget): AppSiteBudget {
-  const budgets = Object.keys(budget.forAsset).map(assetHolderAddress => ({
-    token: tokenAddress(assetHolderAddress) || AddressZero,
-    pending: serializeBudgetItem(budget.forAsset[assetHolderAddress].pending),
-    free: serializeBudgetItem(budget.forAsset[assetHolderAddress].free),
-    inUse: serializeBudgetItem(budget.forAsset[assetHolderAddress].inUse),
-    direct: serializeBudgetItem(budget.forAsset[assetHolderAddress].direct)
-  }));
+  const budgets = Object.keys(budget.forAsset).map(assetHolderAddress => {
+    const assetBudget = checkThat<AssetBudget>(budget.forAsset[assetHolderAddress], exists);
```
kerzhner: Can probably use 2 `map` operations to avoid a local `assetBudget` variable. Not convinced the code will be any clearer with though.

packages/xstate-wallet/src/store/memory-store.ts
```diff
@@ -109,14 +112,60 @@ export class MemoryStore implements Store {
     }
   }
 
-  public getBudget(site: string): Promise<SiteBudget | undefined> {
-    return Promise.resolve(this._budgets[site]);
+  public getBudget(site: string): Promise<SiteBudget> {
```
kerzhner: curious, what's the reason for returning a `Promise`?

packages/xstate-wallet/package.json
```diff
@@ -9,6 +9,7 @@
     "@statechannels/wire-format": "0.0.1",
     "@xstate/react": "0.8.1",
     "ethers": "4.0.45",
+    "async-lock": "1.2.2",
```
kerzhner: Should we also add `"@types/async-lock": "1.1.1"`?

packages/xstate-wallet/src/store/memory-store.ts
andrewgordstewart: It should probably be `async`, but every store function like this should return promises.

PR #1252, opened 2020-03-18
[web3t] Detect message queued in unexpected format

geoknee:
I am just posting this here to show how queued messages come back from the provider in differing formats on the real vs fake providers.

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -281,7 +281,12 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     );
 
     // If the wallet queues a message, send it across the wire
-    this.paymentChannelClient.onMessageQueued((message: Message) => {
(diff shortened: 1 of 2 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -281,16 +281,26 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     );
 
     // If the wallet queues a message, send it across the wire
-    this.paymentChannelClient.onMessageQueued((message: Message) => {
+    this.paymentChannelClient.onMessageQueued((message: Message | any) => {
+      // TODO remove any
```
snario: Why could it be `any`?

geoknee: Good question, this is what I am trying to figure out. I put the `any` here otherwise my runtime check is a type error: it's a reflection that something other than a `Message` is sometimes getting passed to this function.

geoknee:
Will close this if #1264 gets merged.

PR #1253, opened 2020-03-18
Better ethereum enable error handling

tomclose:
Work towards fixing https://github.com/statechannels/monorepo/issues/1240. Probably worth merging now, as it includes work on the json-rpc error format, which will be useful elsewhere.

### DIFF
--- a/packages/xstate-wallet/src/messaging.ts
+++ b/packages/xstate-wallet/src/messaging.ts
@@ -15,6 +15,7 @@ import {
   ApproveBudgetAndFundRequest,
   ChannelProposedNotification,
   CloseAndWithdrawRequest
+  ErrorResponse
 } from '@statechannels/client-api-schema';
 
 import * as jrs from 'jsonrpc-lite';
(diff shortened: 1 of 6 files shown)


tomclose:
New screens: 

Screen Shot 2020-03-19 at 12 27 44 PM

Screen Shot 2020-03-19 at 12 28 25 PM

geoknee:
New screens look good: is the alignment a bit off on the first one?

tomclose:
Yes :) - and I tried to fix it for a while, but then decided it wasn't the highest priority ...

packages/xstate-wallet/src/messaging.ts
```diff
@@ -85,6 +87,11 @@ export class MessagingService implements MessagingServiceInterface {
     this.eventEmitter.emit('SendMessage', response);
   }
 
+  public async sendError(id: number, error: ErrorResponse['error']) {
+    const response = {id, jsonrpc: '2.0', error} as ErrorResponse; // typescript can't handle this otherwise
```
geoknee: Any idea why?

andrewgordstewart: It's probably the `'2.0'` being typed as `string` in `response`.
```suggestion
    const response: Response = {id, jsonrpc: '2.0', error};
```

If you assign the type `Response` to `response`, it has a similar effect while providing type safety -- `'1.0'` is _not_ assignable to `'2.0'`, and the types of the other properties must match as well.

PR #1255, opened 2020-03-18
[wallet] Ensure EthereumEnable returns address

snario:
Missed in #1242


tomclose:
We might be at the point where we can delete the old wallet now ..

snario:
@tomclose I really think we should wait until `rps` works exactly as it did with `wallet`, with no regressions, before deleting it

PR #1257, opened 2020-03-18
[api] Add WalletVersion API Method

snario:
I'd like to add this as a sort of "test" API method that the app / client is actually connected to a wallet and to get some info about what wallet it is connected to.


geoknee:
Is there a better name than `ClientVersion`? There are so many clients...

Why not `WalletVersion`?

snario:
Renamed to `WalletVersion`.

PR #1260, opened 2020-03-18
Implement an IndexedDb Store

lalexgap:
Implements indexedDB in the store. Based on @spersico's work in #1102 .

I've done a basic funding test with `rps` and the `indexedDb` store and things seem to work.

Note: when using indexed db each player must be in a separate browser (or incognito window) since the indexed DB is shared across all tabs in a browser.


packages/xstate-wallet/src/store/memory-channel-storage.ts
```diff
@@ -175,4 +187,87 @@ export class MemoryChannelStoreEntry implements ChannelStoreEntry {
...
+      }
+      if (state.challengeDuration) {
+        state.challengeDuration = parserFunction(state.challengeDuration);
+      }
+      state.outcome = MemoryChannelStoreEntry.toogleBigNumberOutcome(state.outcome, parserFunction);
+    }
+    return stateVariables;
+  }
+
+  private static toogleBigNumberOutcome(
```
tomclose: toggle?

spersico: lol, that was me, sorry :-P

PR #1263, opened 2020-03-19
web3torrent: handle dropped wires

junjizhi:
Fix: https://github.com/statechannels/monorepo/issues/1215

Added a fix to fall back to channel cache to calculate the download size.

Also put in a fix to handle a random crash when the channel cache is not available.

WIP: When leechers cancels the download, wire is dropped but the channel is still running.

### DIFF
--- a/packages/web3torrent/src/utils/calculateWei.ts
+++ b/packages/web3torrent/src/utils/calculateWei.ts
@@ -1,10 +1,16 @@
 import {WEI_PER_BYTE} from '../library/web3torrent-lib';
 import {utils} from 'ethers';
+import prettier from 'prettier-bytes';
 
 export const calculateWei = (fileSize: number | string) => {
   if (!isNaN(Number(fileSize))) {
     return WEI_PER_BYTE.mul(fileSize);
   } else return utils.bigNumberify(NaN);
+};
+
+export const prettyPrintBytes = (wei: utils.BigNumber): string => {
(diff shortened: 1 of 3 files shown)


junjizhi:
@geoknee CI fails with a weird error. and I couldn't rerun the jobs.

packages/web3torrent/src/utils/calculateWei.ts
```diff
@@ -1,12 +1,18 @@
...
   if (!isNaN(Number(fileSize))) {
     return WEI_PER_BYTE.mul(fileSize);
   } else return utils.bigNumberify(NaN);
 };
 
+export const prettyPrintBytes = (wei: utils.BigNumber): string => {
+  const bytes = wei.div(WEI_PER_BYTE);
+  return prettier(bytes.toNumber());
+};
+
```
geoknee: Seems nice to me to have a function `calculateBytes` or `calculateBytesFromWei` that just does the unit conversion, and no formatting. 

The `prettier()` call could then be (optionally) called in the context where `calculateBytes` is called.

This would match `prettyPrintWei(calculateWei(...))`

junjizhi: Yeah. I thought about that, but we had another prettier function down below, so I follow that. The case where we want `calculate` not `prettier` is not clear yet, so I didn't bother to refactor it.

PR #1264, opened 2020-03-19
[channel-provider/client] Add types to ChannelProvider events

geoknee:
This work was spurred by seeing runtime errors due to unexpected types emitted from the provider, as well as discrepancy in types coming from fake vs real providers. 

I believe the second issue to be solved by this PR.

The first issue is not (and possibly cannot be) solved by tinkering with types in the provider. But we can probably diagnose the issue by doing validation of messages *from* the wallet against our schema (see #1266 ) . That work should be much easier given the greater clarity that (I hope) this PR brings to the various JSON RPC types.

### DIFF
--- a/packages/channel-client/src/types.ts
+++ b/packages/channel-client/src/types.ts
@@ -3,15 +3,9 @@ import {
   ChannelResult,
   Participant,
   Allocation,
-  SiteBudget
+  SiteBudget,
+  Message
 } from '@statechannels/client-api-schema';
-
-export interface Message<T = object> {
-  recipient: string; // Identifier of user that the message should be relayed to
-  sender: string; // Identifier of user that the message is from
-  data: T; // Message payload. Format defined by wallet and opaque to app.
-  // But useful to be able to specify, for the purposes of the fake-client
-}
 
 export type UnsubscribeFunction = () => void;
 
--- a/packages/channel-client/src/types.ts
+++ b/packages/channel-client/src/types.ts
@@ -9,22 +9,19 @@ import {
 
 export type UnsubscribeFunction = () => void;
 
-// The message Payload is designed to be opaque to the app. However, it's useful
-// to be able to specify the Payload type for the FakeChannelClient, as we'll be
-// manipulating it within the client.
-export interface ChannelClientInterface<Payload = object> {
+export interface ChannelClientInterface {
   /*
     Queuing a message is meant for when the app receives messages from
     the wallet meant for the opponent's app (and hence the opponent's wallet).
   */
-  onMessageQueued: (callback: (message: Message<Payload>) => void) => UnsubscribeFunction;
+  onMessageQueued: (callback: (message: Message<object>) => void) => UnsubscribeFunction;
   onChannelUpdated: (callback: (result: ChannelResult) => void) => UnsubscribeFunction;
   onChannelProposed: (callback: (result: ChannelResult) => void) => UnsubscribeFunction;
   /*
     Pushing a message is meant for when the app receives a message from
     the opponent's app meant for the wallet.
   */
-  pushMessage: (message: Message<Payload>) => Promise<PushMessageResult>;
+  pushMessage: (message: Message<object>) => Promise<PushMessageResult>;
   createChannel: (
     participants: Participant[],
     allocations: Allocation[],
--- a/packages/channel-client/src/types.ts
+++ b/packages/channel-client/src/types.ts
@@ -14,14 +14,14 @@ export interface ChannelClientInterface {
     Queuing a message is meant for when the app receives messages from
     the wallet meant for the opponent's app (and hence the opponent's wallet).
   */
-  onMessageQueued: (callback: (message: Message<object>) => void) => UnsubscribeFunction;
+  onMessageQueued: (callback: (message: Message) => void) => UnsubscribeFunction;
   onChannelUpdated: (callback: (result: ChannelResult) => void) => UnsubscribeFunction;
   onChannelProposed: (callback: (result: ChannelResult) => void) => UnsubscribeFunction;
   /*
     Pushing a message is meant for when the app receives a message from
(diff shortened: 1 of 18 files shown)


packages/channel-provider/src/types.ts
```diff
@@ -120,11 +121,15 @@ export type MethodRequestType =
...
   send(request: MethodRequestType): Promise<MethodResponseType[MethodRequestType['method']]>;
-  on(event: string, callback: ListenerFn): void;
-  off(event: string, callback?: ListenerFn): void;
-  subscribe(subscriptionType: string, params?: any): Promise<string>;
-  unsubscribe(subscriptionId: string): Promise<boolean>;
+  on: OnType;
+  off: OffType;
+  // subscribe(subscriptionType: string, params?: any): Promise<string>;
+  // unsubscribe(subscriptionId: string): Promise<boolean>;
 }
```
geoknee: This captures the main idea: `callback` now has a type that depends on the type of `event`, meaning passing in the wrong type / mismatched of parameter to the callback will be a type error. E.g. `.on('BudgetUpdated', (channelResult: ChannelResult) => void)`.

Instead of explicitly constructing the conditional types, here we just let the EventEmitter class do that for us, by passing in `<NotificationType>` to its constructor and grabbing a type from the resulting instance.

`NotificationType` is a lookup from event name to params for that event.

packages/client-api-schema/src/types.ts
```diff
@@ -181,20 +181,37 @@ export type GetBudgetResponse = JsonRpcResponse<SiteBudget | {}>;
...
   | ChannelUpdatedNotification
   | ChannelClosingNotification
   | BudgetUpdatedNotification
   | MessageQueuedNotification;
 
+type FilterByMethod<T, Method> = T extends {method: Method} ? T : never;
+
+export type NotificationType = {
+  [T in Notification['method']]: [FilterByMethod<Notification, T>['params']];
+};
```
geoknee: Constructs a type that is a lookup from the allowed `method`s to the corresponding `params`.

packages/channel-client/src/types.ts
```diff
@@ -3,16 +3,10 @@ import {
...
+  Message
 } from '@statechannels/client-api-schema';
 
-export interface Message<T = object> {
-  recipient: string; // Identifier of user that the message should be relayed to
-  sender: string; // Identifier of user that the message is from
-  data: T; // Message payload. Format defined by wallet and opaque to app.
-  // But useful to be able to specify, for the purposes of the fake-client
-}
-
```
geoknee: This was duplicated, have pointed consumers to the original in `client-api-schema`.

packages/channel-provider/src/channel-provider.ts
```diff
@@ -49,48 +56,42 @@ class ChannelProvider implements ChannelProviderInterface {
...
     return subscriptionId;
   }
 
   async unsubscribe(subscriptionId: string): Promise<boolean> {
-    Object.keys(this.subscriptions).map(e => {
-      this.subscriptions[e] = this.subscriptions[e]
-        ? this.subscriptions[e].filter(s => s !== subscriptionId)
-        : [];
+    Object.keys(this.subscriptions).forEach(method => {
+      this.subscriptions[method as Notification['method']] = this.subscriptions[
```
geoknee: https://github.com/microsoft/TypeScript/pull/12253#issuecomment-263132208

packages/client-api-schema/src/types.ts
```diff
@@ -184,18 +184,28 @@ export type CloseAndWithdrawParams = {site: string; player: Participant; hub: Pa
 export type CloseAndWithdrawRequest = JsonRpcRequest<'CloseAndWithdraw', CloseAndWithdrawParams>;
 export type CloseAndWithdrawResponse = JsonRpcResponse<{success: boolean}>;
 // Notifications
+// these notifications come *from* the wallet, which is not strictly how JSON-RPC should work
+// (since we treat the wallet as the 'server')
```
geoknee: We discussed this in standup.

packages/channel-provider/src/channel-provider.ts
```diff
@@ -49,48 +56,42 @@ class ChannelProvider implements ChannelProviderInterface {
...
     }
 
-    if (isJsonRpcNotification(message)) {
-      const eventName = message.method;
-      if (eventName === 'UIUpdate') {
+    if (isJsonRpcNotification<keyof NotificationType>(message)) {
+      // TODO: use schema validations as better type guards
+      const notificationMethod = message.method;
+      const notificationParams = message.params;
+      this.events.emit(notificationMethod, notificationParams);
```
geoknee: Here we strip out the JSON-RPC headers; the message just came across that interface so the rest of the app does not need this information. The implication is that the channel client never sees the `message`, only an event with event string `message.method` and data `message.params`.

packages/channel-client/src/types.ts
```diff
@@ -3,34 +3,25 @@ import {
...
-// to be able to specify the Payload type for the FakeChannelClient, as we'll be
-// manipulating it within the client.
-export interface ChannelClientInterface<Payload = object> {
+export interface ChannelClientInterface {
   /*
     Queuing a message is meant for when the app receives messages from
     the wallet meant for the opponent's app (and hence the opponent's wallet).
   */
-  onMessageQueued: (callback: (message: Message<Payload>) => void) => UnsubscribeFunction;
+  onMessageQueued: (callback: (message: Message<object>) => void) => UnsubscribeFunction;
```
tomclose: I think we should change `Message` so it doesn't accept a type, and the `payload` is of type `Unknown`.

geoknee: d1c659f

Alonski:
This PR missed this file :)
https://github.com/statechannels/monorepo/blob/master/packages/tic-tac-toe/app/services/ttt-channel-client.ts#L4

PR #1267, opened 2020-03-19
RPS with virtual funding

andrewgordstewart:
# Instructions for testing virtual funding
1. Set `USE_VIRTUAL_FUNDING` to `'true'` (or any truthy value) to use virtual funding.
2. Set `FIREBASE_URL` to the same value between `simple-hub` and `rps`.
3. Accept the deposits that automatically happen when you initialize the xstate-wallet.


packages/simple-hub/src/wallet/deposit.ts
```diff
@@ -27,17 +27,24 @@ export function depositsToMake(
 ): {channelId: string; amountToDeposit: BigNumber}[] {
   const simpleAllocationStates = _.filter(isSimpleAllocationState, message.signedStates);
   return simpleAllocationStates
-    .filter(state => state.participants.length === 2)
```
kerzhner: No need for a closure and return syntax as far as I can tell

packages/simple-hub/src/wallet/deposit.ts
```diff
@@ -27,17 +27,24 @@ export function depositsToMake(
...
+    })
+    .filter(state => {
+      return (
         _.findIndex(
-          allocationItem => allocationItem.destination === cHubChainAddress,
+          allocationItem => allocationItem.destination === makeDestination(cHubChainAddress),
           state.outcome.allocationItems
         ) === 0
-    )
-    .filter(state => state.turnNum.eq(ethers.constants.Zero))
```
kerzhner: No need for a closure and return as far as I can tell

packages/simple-hub/src/start-dev-server.ts
```diff
@@ -13,18 +13,18 @@ async function setupGanacheAndContracts() {
 }
 
 async function startLocalFirebaseServer() {
-  const server = new FirebaseServer(5555, 'localhost');
```
kerzhner: Should this be deleted for now?

packages/simple-hub/src/wallet/deposit.ts
andrewgordstewart: Ah, these changes were made as it seems finicky to add breakpoints without a closure + return statement.

packages/simple-hub/src/start-dev-server.ts
andrewgordstewart: It would be nice to get it working, so that we can use a local firebase server. But it was not my priority at the time.

andrewgordstewart:
Closing as this is superceded by either https://github.com/statechannels/monorepo/pull/1272 (which is unfortunately already out of date) or a different PR.

PR #1271, opened 2020-03-21
[TTT] Fix type import errors

Alonski:
Fixes a type import error.
An import was moved in this PR and wasn't update for TTT.
https://github.com/statechannels/monorepo/pull/1264

This also adds `@statechannels/client-api-schema` as a devDependency of RPS and TTT as they both import from this package.

### DIFF
--- a/packages/tic-tac-toe/app/services/ttt-channel-client.ts
+++ b/packages/tic-tac-toe/app/services/ttt-channel-client.ts
@@ -1,10 +1,10 @@
 import Service from '@ember/service';
 import {
   ChannelResult,
-  Message,
   ChannelClientInterface,
   UnsubscribeFunction
 } from '@statechannels/channel-client';
+import {Message} from '@statechannels/client-api-schema';
 import {AppData, encodeAppData, decodeAppData} from '../core/app-data';
 import {ChannelState} from '../core/channel-state';
 import ENV from '@statechannels/tic-tac-toe/config/environment';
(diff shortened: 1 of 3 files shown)


packages/tic-tac-toe/app/services/ttt-channel-client.ts
```diff
@@ -1,10 +1,10 @@
 import Service from '@ember/service';
 import {
   ChannelResult,
-  Message,
   ChannelClientInterface,
   UnsubscribeFunction
 } from '@statechannels/channel-client';
+import {Message} from '@statechannels/client-api-schema';
```
andrewgordstewart: This appears to be a real dependency, not a dev dependency. Is there a reason you only include it as a dev dependency?

Alonski: Hmm, doesn't really matter but this is a TS type import. So if it's important to anyone I can move it to dependencies.

snario: @andrewgordstewart I think type information imports make sense as dev dependencies

snario: It depends on the distributed file in the `package.json` but in the case of `tic-tac-toe` it's a JS bundle, so seems reasonable to make type information a dev dependency

andrewgordstewart: Ah, I hadn't clued in that it it was for type information only.

Alonski: So can we merge this in? 😁

PR #1272, opened 2020-03-22
General fixes around virtual funding

andrewgordstewart:
This PR includes a few fixes required to make virtual funding work "in practice".

The most widespread change is around destinations. Some places used 20 bytes (addresses) as destinations, some 32 bytes. Some places used lower-case addresses to form destinations, some used checksum-cased addresses. I introduced a `Destination` "nominal" type to ensure a consistent usage of destinations.

Virtual funding was tested manually by populating the store by hand with a ledger channel and a budget on wallet initialization. With this strategy, it worked. I then tried to create the ledger channel by triggering a `ApproveBudgetAndFund` workflow on initialization, but ran into firebase/RPS issues that made this difficult.


packages/simple-hub/jest/jest.chain.config.js
```diff
@@ -1,5 +0,0 @@
-const config = require('./jest.config');
```
kerzhner: I wonder if this jest configuration should remain. We could have chain tests in the near future, and I don't see much harm in leaving the configuration if there are no tests that match the `testMatch` pattern. Of course, to play devil's advocate, we shouldn't have any dead code in the repo.

andrewgordstewart: One reason I deleted it was, the `chain:test` command was failing but not the jest process was not exiting.

I've restored `jjest.chain.config.js` for now.

andrewgordstewart: >I don't see much harm in leaving the configuration if there are no tests that match the testMatch pattern.

Unfortunately, this results in a jest failure :P

packages/hub/package.json
```diff
@@ -73,7 +73,6 @@
     "prestart": "yarn build && npm run db:rollback && npm run db:migrate && npm run db:seed",
     "prettier:check": "npx prettier --check 'src/**/*.{ts,tsx}'",
     "prettier:write": "npx prettier --write 'src/**/*.{ts,tsx}'",
-    "start:shared-ganache": "NODE_ENV=development npx start-shared-ganache",
-    "test:ci": "run-s chain:test hub:test"
```
snario: ?

packages/simple-hub/src/blockchain/eth-asset-holder.ts
```diff
@@ -11,20 +11,28 @@ const walletWithProvider = new ethers.Wallet(cHubChainPK, provider);
...
     // To avoid this we only allow deposit transactions to happen serially.
     await Blockchain.attachEthAssetHolder();
 
-    return lock.acquire('depositing', async () => {
-      const tx = await Blockchain.ethAssetHolder.deposit(channelID, expectedHeld, value, {
-        value
-      });
+    return lock.acquire('depositing', async release => {
+      const expectedHeld: BigNumber = await Blockchain.ethAssetHolder.holdings(channelID);
+      if (expectedHeld.gte(value)) return;
```
snario: Why not `release()` here?

packages/simple-hub/src/constants.ts
```diff
@@ -4,7 +4,8 @@ export const cFirebasePrefix = process.env.FIREBASE_PREFIX || 'default-prefix';
 
 export const cHubChannelPK =
   process.env.HUB_CHANNEL_PK ||
-  '0x1b427b7ab88e2e10674b5aa92bb63c0ca26aa0b5a858e1d17295db6ad91c049b';
+  '0x8624ebe7364bb776f891ca339f0aaa820cc64cc9fca6a28eec71e6d8fc950f29';
+// '0xaaaa84838319627Fa056fC3FC29ab94d479B8502'
```
snario: Is there any significance to this pair (e.g., pre-funded)

packages/hub/package.json
andrewgordstewart: The hub is no longer in use. I don't want to fix tests that we are going to delete.

packages/simple-hub/src/constants.ts
andrewgordstewart: It starts with `aaaa`, so it's easy to identify who it is.

packages/simple-hub/package.json
```diff
@@ -66,6 +65,6 @@
     "prettier:check": "npx prettier --check 'src/**/*.{ts,tsx}'",
     "prettier:write": "npx prettier --write 'src/**/*.{ts,tsx}'",
     "start:shared-ganache": "NODE_ENV=development npx start-shared-ganache",
-    "test:ci": "run-s chain:test hub:test"
+    "test:ci": "yarn hub:test"
```
lalexgap: Should we still be skipping `SimpleHub` tests?

packages/wallet/src/contract-tests/adjudicator-events.test.ts
```diff
@@ -41,7 +41,7 @@ describe("adjudicator listener", () => {
     await depositIntoETHAssetHolder(signer, channelId);
   });
 
-  it("should not handle an event when no process has registered", async () => {
+  it.skip("should not handle an event when no process has registered", async () => {
```
lalexgap: Can these be re-enabled?

packages/wallet/src/redux/protocols/consensus-update/__tests__/reducer.test.ts
```diff
@@ -9,7 +9,7 @@ import {
 
 import * as scenarios from "./scenarios";
 
-describe("Two Players", () => {
+describe.skip("Two Players", () => {
```
lalexgap: Outside the scope of this PR but I think it might make sense to remove the `wallet` package now.

packages/xstate-wallet/src/messaging.ts
```diff
@@ -241,18 +241,33 @@ export function sendDisplayMessage(displayMessage: 'Show' | 'Hide') {
   window.parent.postMessage(message, '*');
 }
 
+export function convertToInternalParticipant(participant: {
```
lalexgap: Should this be moved into the `serde` folder?

packages/simple-hub/package.json
andrewgordstewart: `chain:test` is testing facilities that we are no longer using -- watching for deposits, and matching them.

packages/wallet/src/contract-tests/adjudicator-events.test.ts
andrewgordstewart: I already spent some time trying to get these wallet tests to pass. Since we're going to delete the package, it doesn't make sense to continue with that.

lalexgap: Sorry, I thought these were test in the `simple-hub` package not `wallet`

packages/xstate-wallet/src/messaging.ts
```diff
@@ -239,18 +239,33 @@ export function sendDisplayMessage(displayMessage: 'Show' | 'Hide') {
...
+  destination: string;
+  signingAddress: string;
+  participantId: string;
+}): Participant {
+  return {
+    ...participant,
+    destination: makeDestination(participant.destination)
```
kerzhner: Nit, maybe keep the naming scheme consistent between `convertToInternalParticipant` and `makeDestination`? In other words, maybe `convertToInternalParticipant` should be `makeParicipant`?

andrewgordstewart: I think I might stick with `convertToInternalParticipant`, since it matches `convertToInternalEvent`?

packages/rps/src/constants.ts
```diff
@@ -4,3 +4,8 @@ export const FIREBASE_PREFIX = process.env.FIREBASE_PREFIX || 'default-prefix';
 
 // todo: load the correct address
 export const RPS_ADDRESS = process.env.RPS_CONTRACT_ADDRESS;
+
+export const HUB = {
+  signingAddress: '0xaaaa84838319627Fa056fC3FC29ab94d479B8502',
```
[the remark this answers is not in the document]
andrewgordstewart: Agreed, but these need to be changed across several packages. 

I'd like to define cross-package things like this at the root level, but our current environment configuration makes that a bit hard.

PR #1274, opened 2020-03-22
Web3torrent: Handle peer cancelling downloads

junjizhi:
Fixes https://github.com/statechannels/monorepo/issues/1230

The UI looks like below after Cancel Download is pressed:
image

### DIFF
--- a/packages/web3torrent/src/clients/web3torrent-client.test.ts
+++ b/packages/web3torrent/src/clients/web3torrent-client.test.ts
@@ -100,7 +100,7 @@ describe('Web3TorrentClient', () => {
         );
     });
 
-    it('should return the ID of the removed client', async () => {
+    it('should return the infohash of the paused client', async () => {
       const mockInfoHash = '124203';
       const result = await cancel(mockInfoHash);
 
--- a/packages/web3torrent/src/clients/web3torrent-client.test.ts
+++ b/packages/web3torrent/src/clients/web3torrent-client.test.ts
@@ -3,7 +3,7 @@ import {PaidStreamingTorrent, WebTorrentAddInput, WebTorrentSeedInput} from '../
 import {TorrentCallback} from '../library/web3torrent-lib';
 import {Status} from '../types';
 import {createMockTorrent, createMockTorrentPeers} from '../utils/test-utils';
-import {download, getTorrentPeers, remove, upload, web3torrent} from './web3torrent-client';
+import {download, getTorrentPeers, cancel, upload, web3torrent} from './web3torrent-client';
 
 describe('Web3TorrentClient', () => {
   describe('download()', () => {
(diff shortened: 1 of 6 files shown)


packages/web3torrent/src/clients/web3torrent-client.test.ts
```diff
@@ -102,7 +102,7 @@ describe('Web3TorrentClient', () => {
 
     it('should return the ID of the removed client', async () => {
```
geoknee:
```suggestion
    it('should return the infoHash of the paused torrent', async () => {
```

packages/web3torrent/src/components/torrent-info/download-info/DownloadInfo.tsx
```diff
@@ -29,18 +29,26 @@ const DownloadInfo: React.FC<DownloadInfoProps> = ({
   const totalSpent = myPayingChannelIds
     .map(id => channelCache[id].beneficiaryBalance)
     .reduce((a, b) => bigNumberify(a).add(bigNumberify(b)), bigNumberify(0));
+
+  const displayProgress = !(torrent.done || torrent.paused);
```
geoknee: I think we need to consider what the user sees in `.paused`. Right now, it's not clear that they are in that state. It would be nice to let them know that
- They didn't finish the download
- They may have still paid some money out

Although (as we discussed) the low level `pause` method is better than `remove` (when the user clicks [ Cancel Download ] ), from a UI perspective the torrent isn't really paused. Pausing suggests the possibility of restarting, we which aren't currently supporting.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -420,6 +427,14 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
   }
 
+  protected async closeChannels(torrent: PaidStreamingTorrent) {
+    torrent.wires.forEach(async wire => {
+      if (wire.paidStreamingExtension && wire.paidStreamingExtension.peerChannelId) {
+        await this.paymentChannelClient.closeChannel(wire.paidStreamingExtension.peerChannelId);
+      }
+    });
+  }
+
```
geoknee: Can we add a comment here explaining that we only close channels that our peer opened (i.e. that we are paying *out* in) ?

We might even want to rename `closeDownloadingChannels` or something like that. 

Obviously this behaviour means that I can still get paid for uploading the bits of the file I still have.

packages/web3torrent/src/components/torrent-info/download-info/DownloadInfo.tsx
junjizhi: Actually, the status of the torrent is shown as `Cancelled`.

The UI looks like below after Cancel Download is pressed:
image

packages/web3torrent/src/library/web3torrent-lib.ts
junjizhi: fixed.

junjizhi:
@geoknee It seems CI was failing for no good reasons. But I think the PR is good to merge.

snario:
Screen Shot 2020-03-23 at 14 02 37

For future reference, this should be click-able when there is an irregular CI error

junjizhi:
@snario It's not clickable for me.
image

PR #1275, opened 2020-03-23
W3T add dev tracker

spersico:
This PR adds a tracker server to the start script of web3torrent.
The reason for this is explained in [here](https://statechannels.slack.com/archives/CMJFENMEX/p1584820990099200).

For now, this won't work in the public URL (I think), because the tracker is not deployed to any public domain. 

It's a development only feature, for now, to be able to work on the web3torrent project (as the tracker servers are still down at the moment of creation of this PR). 

The tracker should be fairly easy to deploy though.


packages/web3torrent/src/constants.ts
```diff
@@ -4,13 +4,17 @@ import {ChannelState} from './clients/payment-channel-client';
...
-  'udp://tracker.opentrackr.org:1337',
-  'wss://tracker.btorrent.xyz',
-  'wss://tracker.openwebtorrent.com'
+  // 'udp://explodie.org:6969',
+  // 'udp://tracker.coppersurfer.tk:6969',
+  // 'udp://tracker.empire-js.us:1337',
+  // 'udp://tracker.leechers-paradise.org:6969',
+  // 'udp://tracker.opentrackr.org:1337',
+  // 'wss://tracker.btorrent.xyz',
+  // 'wss://tracker.openwebtorrent.com'
```
geoknee: Is it necessary to disable the public trackers?

If they are online (I think since you raised this PR they came back on), and I didn't run `yarn  start:tracker`, it would be nice if I could still connect to them.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -112,18 +112,19 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+        torrentId = torrent.infoHash;
+        torrent.once('wire', wire => resolve(true));
       });
     });
     const timer = new Promise(function(resolve, reject) {
       setTimeout(resolve, timeOut);
     });
-    return Promise.race([gotAWire, timer]);
+    const raceResult = await Promise.race([gotAWire, timer]);
+    this.remove(torrentId);
```
geoknee: 👍

packages/web3torrent/tracker/index.js
```diff
@@ -0,0 +1,75 @@
+const Server = require('bittorrent-tracker').Server;
```
geoknee: If we were to deploy this tracker so that it was publicly accessible, would it track non-web3torrent torrents? If so, is there any easy way for it to only track web3torrents?

packages/web3torrent/src/constants.ts
geoknee: I notice that this works anyway, despite these lines being commented out.  I had previously been similarly confused, when I tried to remove a single tracker but saw that the client still trying to connect to it (via the console). 

Could you un-confuse me about where trackers can be added/removed or at least enabled/siabled?

geoknee: I also just spotted that you run the tracker as a part of the `start` script, so it is not necessary to do that separately. 

Would be nice to have 2/3 sentences in readme to explain that.

spersico: > (I think since you raised this PR they came back on)

You gotta be kidding me, I can't believe it.
...
...
yeah, the public tracker works now.

packages/web3torrent/package.json
```diff
@@ -131,6 +131,7 @@
     "prettier:check": "prettier --check '**/*.{scss,json}' --ignore-path '.eslintignore'",
     "prettier:write": "prettier --write '**/*.{scss,json}' --ignore-path '.eslintignore'",
     "start": "node scripts/start.js",
+    "start:tracker": "node tracker/index.js",
```
geoknee: It doesn't currently seem possible to run this script if I already ran `yarn start` (a tracker is already running on that port). 

That means this command can only allow me to run a local tracker on its own, without the web3torrent client, which seems of limited use.

packages/web3torrent/scripts/start.js
```diff
@@ -61,4 +61,18 @@ void (async () => {
...
+    console.log(data.toString());
+  });
+
+  trackerServer.stderr.on('data', data => {
+    throw data.toString();
+  });
+
+  trackerServer.on('close', code => {
+    process.exit(code);
+  });
```
geoknee: I am not sure this server is being torn down correctly. If I run `yarn start`, kill the process and then run `yarn start` again, I get this:

```
➜  web3torrent git:(sp/w3t-tracker) ✗ yarn start 
yarn run v1.19.1
$ node scripts/start.js
The USE_GANACHE_DEPLOYMENT_CACHE option is set. Using ganache in shared mode with cached deployments. Port = 8547.
Found shared ganache instance running on http://localhost:8547.
Using the deployments cache at /Users/georgeknee/statechannels/monorepo/.ganache-deployments/ganache-deployments-8547.json.
Contract SingleAssetPayments already exists address: 0x430869383d611bBB1ce7Ca207024E7901bC26b40
$ node tracker/index.js

$ /Users/georgeknee/statechannels/monorepo/node_modules/.bin/react-scripts start


/Users/georgeknee/statechannels/monorepo/packages/web3torrent/scripts/start.js:72
    throw data.toString();
    ^
listen EADDRINUSE: address already in use :::8000

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
➜  web3torrent git:(sp/w3t-tracker) ✗ 
```

packages/web3torrent/src/constants.ts
spersico: > I notice that this works anyway, despite these lines being commented out. 

Per WebTorrent code, if no trackers are defined, WebTorrent uses the public ones. If the public ones work, all is fine and dandy, if not, nothing will work.

There are two (or three) places where trackers matter:
[Here](https://github.com/statechannels/monorepo/pull/1275/files#diff-679cd7fa88a4e6378ca287a6bd6de080R10): That sets the default trackers to be used when seeding.
[Here](https://github.com/statechannels/monorepo/pull/1275/files#diff-a84153ed9f3ab2bd13c41bca43b584c7R141): That sets the tracker for each torrent when seeded (I just noticed that this is probably unnecessary).
[and Here](https://github.com/statechannels/monorepo/blob/8068eabc7702526cbed3f44d19cdcf6e2cc3b4b5/packages/web3torrent/src/utils/magnet.ts#L16): That sets the tracker for each torrent when parsing the url from the browser.

packages/web3torrent/tracker/index.js
spersico: Yeah, I read about [this](https://github.com/webtorrent/bittorrent-tracker#server) on the project, we could add some sort of secret or something to the clients, and pass somehow that secret to the tracker so that it only serves web3torrent clients.

packages/web3torrent/package.json
spersico: It accepts the TRACKER_PORT environment variable. That allows for multiple trackers to be run at the same time in the same computer.

spersico: By the way, why would you want to run it independently? From this PR onwards we would run it as part of the `yarn start` script.

spersico: image
That way, we don't have to run yet another terminal

packages/web3torrent/scripts/start.js
spersico: I.... can't seem to get that error. I ran it like 20 times, hitting `CTRL` + `C` at different moments, and it always restarts okay.
Are you sure that you are not running the tracker in another tab or something?

packages/web3torrent/package.json
geoknee: Yeah we probably wouldn't need to run it separately; I just spotted the `yarn start:tracker` script and wondered what that was for :-) . I tried it out because I thought I might be able to see the logs that way.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: uh, I forgot I left that there. It avoids some inconsistent behaviour and some console errors that I noticed.

packages/web3torrent/scripts/start.js
geoknee: Ok I killed all my processes that listen on `8000` and now the problem seems to go away 🤷‍♂

packages/web3torrent/tracker/index.js
geoknee: Yeah I think anything like that would be good -- it probably doesn't need to be super secure. 

We could filter on the `pseAccount` being a valid Ethereum address https://ethereum.stackexchange.com/questions/1374/how-can-i-check-if-an-ethereum-address-is-valid

spersico: uhh, that's cool. Maybe I can add it in another PR? I think it involves changing the internal peerId (which is supposed to be read-only) after the client was instantiated. And we get the pseAccount after the web3torrent is created.

packages/web3torrent/scripts/start.js
spersico: I gave it another try, trying to show the debug messages, and everything breaks.
I pushed another commit. Now if you run `DEBUG="web3torrent:tracker" yarn start` it should show you all the logs of the tracker in the console

packages/web3torrent/tracker/index.js
geoknee: Yeah separate PR is cool. 

 I'm not sure what ends up in the `params` of the server `filter` function: anything that picks out web3Torrent uniquely would work, I guess.

packages/web3torrent/scripts/start.js
geoknee: Thanks! I think I had tried this once maybe that's why I had issues.

packages/web3torrent/src/constants.ts
geoknee: So is there a way of blacklisting a tracker? Say, if one public tracker in particular was to go down permanently?

spersico: AFAIK, with this PR, I've effectively stopped using the entire _free-open-tracker_ list. Unless a torrent explicitly states that a specific tracker must be used, W3T it will assume that the trackers to use are specified in the default tracker list here.

By that same logic, if we wanted to, say, add or remove trackers, we could do that just adding or removing them from the `defaultTrackers` now

PR #1279, opened 2020-03-23
Handle the case of no metamask / wrong network

tomclose:
Screen Shot 2020-03-23 at 10 49 12 AM

Screen Shot 2020-03-23 at 10 49 08 AM

### DIFF
--- a/packages/xstate-wallet/src/ui/enable-ethereum-workflow.tsx
+++ b/packages/xstate-wallet/src/ui/enable-ethereum-workflow.tsx
@@ -1,8 +1,12 @@
-import React from 'react';
+import React, {useContext} from 'react';
 import {EventData} from 'xstate';
 import './wallet.scss';
-import {Button, MetaMaskButton} from 'rimble-ui';
+import {Button, Box, Flex, Icon, Text, MetaMaskButton, Flash} from 'rimble-ui';
+
+import ConnectionBanner from '@rimble/connection-banner';
+import RimbleUtils from '@rimble/utils';
 import {WorkflowState} from '../workflows/approve-budget-and-fund';
+import {WindowContext} from './window-context';
 
 interface Props {
   current: WorkflowState;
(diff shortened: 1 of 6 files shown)


packages/xstate-wallet/src/ui/stories/enable-ethereum.stories.tsx
```diff
@@ -17,19 +18,34 @@ const testContext = {
...
+      devTools: true
+    }
+  ); // start a new interpreted machine for each story
+  machine.onEvent(event => console.log(event.type)).start(state);
+  storiesOf('Workflows / Enable Ethereum', module).add(
+    name,
+    renderComponentInFrontOfApp(
+      <WindowContext.Provider value={window}>
+        <EnableEthereum current={machine.state} send={machine.send} />
+      </WindowContext.Provider>
```
geoknee: Wondering if you could have used `.withContext` as on L28?

tomclose: I think that's a different sort of context: it's the xstate machine context, rather than the react `useContext` context.

packages/xstate-wallet/src/ui/enable-ethereum-workflow.tsx
```diff
@@ -34,6 +121,16 @@ export const EnableEthereum = (props: Props) => {
...
   };
 
+  const connectionBanner = (
+    <ConnectionBanner currentNetwork={currentNetwork} requiredNetwork={targetNetwork}>
+      {{
+        notWeb3CapableBrowserMessage: <NotWeb3Browser />,
+        noNetworkAvailableMessage: <NoNetwork />,
+        onWrongNetworkMessage: <WrongNetwork />
+      }}
+    </ConnectionBanner>
```
geoknee: Am I right in thinking this is pretty close to just using the default values? Is the main advantage that you get to resize the banner by doing this?

packages/xstate-wallet/src/ui/stories/enable-ethereum.stories.tsx
geoknee: Yes you're right.

 I was getting confused with the `.addDecorator` that we use here https://github.com/statechannels/monorepo/blob/master/packages/web3torrent/.storybook/config.js . 

I guess what you have is fine for now but if we wanted to reuse L37-39  ^ this might be an option.

snario:
I think the wallet should send back an error to `EnableEthereum` if it detects that there is no MetaMask installed. Otherwise that API call will hang and timeout. It is unlikely the user will install MetaMask, this UI will update, and the user will continue in the timeout of the API call.

PR #1281, opened 2020-03-23
[channel-provider] Update window.channelProvider object interface

snario:
1. `.mountWalletComponent(url)` is what `.enable(url)` used to do. It hooks up the app window to the wallet window by connecting a PostMessage listener and mounting the wallet UI in the DOM.

2. `.enable()` is now what `.send({method: "EthereumEnable"})` would do; this means it tells the wallet to connect to a Web3 provider (i.e., MetaMask). If it succeeds, it also populates `.address`, `.selectedAddress` and `.walletVersion` on the `window.channelProvider` object.

### App Workflow

1. Ensure `.mountWalletComponent` is called to add the wallet to the app. In the future this will be done by a Chrome Extension, so the app won't need to use this at all.

2. Check `window.channelProvider.address` (or `.walletVersion` or `.selectedAddress`) to see if the wallet has been enabled or not. They'll be `undefined` if not enabled.

3. If not enabled, call `window.channelProvider.enable()` which will trigger wallet UI. Otherwise, move forward with the app.


packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -34,19 +34,30 @@ type ChannelId = string;
...
+  public selectedAddress?: string;
+  public walletVersion?: string;
+
   private events = new EventEmitter<EventType>();
   protected url = '';
 
   playerIndex: Record<ChannelId, 0 | 1> = {};
   opponentIndex: Record<ChannelId, 0 | 1> = {};
-  address: string = Wallet.createRandom().address;
+  internalAddress: string = Wallet.createRandom().address;
```
geoknee: We are using `signingAddress` elsewhere, and I'm mindful that we don't want to proliferate too many names for the same thing. 

To further muddy the waters / potentially clarify the terminology, @kerzhner made the point that we could call this `channelPublicKey` to reflect its limited role in the context in which we use it, and help disambiguate from other things like `outcomeAddress`  / `ethereumAddress`.
https://statechannels.slack.com/archives/CQJCSTKA7/p1584026202014600

Perhaps best to stick with `signingAddress` for this PR, and then consider a global change of terminology in a separate one?

packages/channel-provider/src/channel-provider.ts
```diff
@@ -76,6 +87,12 @@ class ChannelProvider implements ChannelProviderInterface {
 
   off: OffType = (method, params) => this.events.off(method, params);
 
+  private async populateProviderProperties() {
+    this.internalAddress = await this.send({method: 'GetAddress', params: {}});
+    this.selectedAddress = await this.send({method: 'GetEthereumSelectedAddress', params: {}});
+    this.walletVersion = await this.send({method: 'WalletVersion', params: {}});
+  }
```
geoknee: I like this. It seems to hint to us to combine these RPC calls into one at some point in the future. 

Should we make the `WalletVersion` check *first*? If that fails, we should get a meaningful error and not bother with the other calls.

packages/channel-client/tests/fakes/fake-channel-provider.ts
snario: Agree, see https://github.com/statechannels/monorepo/pull/1281/commits/db5b4067a32b83bb6a4d18e2abe95926a677e275

packages/channel-provider/src/channel-provider.ts
snario: I think there is a separate thing we can do which is to determine whether or not the iFrame is in fact loaded and the PostMessage bridge is behaving correctly. Ideally there would be some `testConnection` method or something like that. We could have a property on `ChannelProvider` which is `connectionEstablished: boolean` which would be set to `true` on an event monitoring the frame [like this one](https://stackoverflow.com/a/36155560/2680092) being triggered.

snario:
The `tic-tac-toe` tests are unrelated.

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: I think some instances of `internalAddress` still snuck in?

PR #1288, opened 2020-03-23
Remove build:netlify from packages with manual netlify deployment

kerzhner:
The upside of including `build:netlify` is that CI tests that all packages hosted on netlify can be built for netlify when the package changes and on all master CI runs. The downside is that we are seeing [out-of-memory errors](https://app.circleci.com/pipelines/github/statechannels/monorepo/4464/workflows/7f137d70-7283-49b9-a440-68e22a5ba936/jobs/16026) even with concurrency set to 2.

### DIFF
--- a/packages/rps/package.json
+++ b/packages/rps/package.json
@@ -186,7 +186,6 @@
   "scripts": {
     "build": "run-s build:typescript build:webpack",
     "build:ci": "run-s build:typescript contracts:compile",
-    "build:netlify": "NODE_ENV=production yarn build:webpack",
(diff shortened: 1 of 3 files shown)


kerzhner:
The other solution to the out-memory errors is to set concurrency to 1 for the top level `build:netlify` command. With that concurrency setting, `release:netlify` (which triggers `build:netlify`)  becomes the longest running command by 4 minutes.

Of course, we can always increase the circleci resource class for the `release:netlify` job if we deem that the solution proposed in this PR is unsatisfactory.

PR #1290, opened 2020-03-24
[channel - provider] Do not populateProviderProps() until enabled

geoknee:
If I clear my metamask connections, then I get a metamask not enabled error:

Screenshot 2020-03-24 at 10 34 23

We should only ever try to populate these properties once we have enabled ethereum in the wallet (in web3torrent we do this immediately afterwards anyway).

### DIFF
--- a/packages/channel-provider/src/channel-provider.ts
+++ b/packages/channel-provider/src/channel-provider.ts
@@ -48,7 +48,6 @@ class ChannelProvider implements ChannelProviderInterface {
     this.messaging.setUrl(this.url);
     await this.ui.mount();
     console.info('Application successfully mounted Wallet iFrame inside DOM.');
-    await this.populateProviderProperties();
   }
 
   async enable() {


geoknee:
@tomclose This fix prevents the error happening before the pop-up happens... but I still see the error after enabling in metamask.

snario:
I think it _should_ populate if `window.ethereum` is already enabled. This is how `window.ethereum` works. If it was enabled once in the past, then `window.ethereum.selectedAddress` is filled when you reload the page. I think `window.channelProvider.selectedAddress` should act similarly.

geoknee:
We can get that behaviour just by calling `channelProvider.enable()` before trying to populate the addresses (it is idempotent). Or otherwise wrapping some logic around the `populate()` call: i.e. make it conditional on ethereum being enabled.

In any case, the problem I am trying to solve is when we have *never* enabled, which currently gives rise to an error.

snario:
@geoknee I think the wallet should throw out an `EthereumNotEnabled` error in that case, no?

geoknee:
That's exactly what it does, if you look at the screenshot. 

Unless I am missing something, this seems bad. How am I ever going to get ethereum enabled in the wallet?

snario:
I was referring to a particular JSON RPC Error response type that describes "the wallet not being enabled" as opposed to the wallet throwing a runtime error.

> Unless I am missing something, this seems bad. How am I ever going to get ethereum enabled in the wallet?

You should call `.enable()` and then the wallet will run the enable workflow. You shouldn't need to see this JSON RPC error to know to call this method though, you should just check `window.channelProvider.selectedAddress` for being `undefined`.

My point from above is that, if the enable workflow has _already_ been run, then you should not need to call `.enable()` a second time. If it's idempotent then sure that's fine, but in my opinion `.enable()` should be triggered upon a User UI click (meaning you would not run it on every page load).

tomclose:
Once the fix in https://github.com/statechannels/monorepo/pull/1296 lands, we shouldn't need this anymore. 

> You shouldn't need to see this JSON RPC error to know to call this method though, you should just check window.channelProvider.selectedAddress for being undefined.

Small point: I think this should be `null` not `undefined` - that's the behaviour with `window.ethereum` anyhow.

PR #1293, opened 2020-03-24
[xstate-wallet] Validate messages in + out

geoknee:
Currently tests will fail due to the increased checking that is being done.  Assuming the `wire-format` is correct, looks like there is some work to do ensuring messages end up being serialized in a way that is consistent with that.


packages/xstate-wallet/src/serde/wire-format/serialize.ts
```diff
@@ -20,12 +21,22 @@ export function serializeMessage(message: Message, recipient: string, sender: st
   };
 }
 
+function bigNumberToUint256(bigNumber: BigNumber): string {
+  // our wire protocol calls for hex strings of this exact format
+  return (
```
lalexgap: We could probably use `ethers` `hexZeroPad` here

PR #1295, opened 2020-03-24
Close and withdraw ui

tomclose:
Not done yet, but the first screen is there. Have made a couple of changes to the wallet screen framework, which are probably worth merging, in case anyone else wants to touch the UI.

Screen Shot 2020-03-24 at 4 59 42 PM

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,6 +5,20 @@
   "version": "0.2.0",
   "configurations": [
     {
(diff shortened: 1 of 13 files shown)


tomclose:
Screen Shot 2020-03-25 at 3 18 57 PM

geoknee:
Possibly not of high priority, but could we align text to the left (justified)?

PR #1296, opened 2020-03-24
[api] Combine GetAddress, WalletVersion, GetEthereumSelectedAddress

snario:


packages/xstate-wallet/src/messaging.ts
```diff
@@ -143,8 +143,15 @@ export class MessagingService implements MessagingServiceInterface {
     const {id: requestId} = request;
 
     switch (request.method) {
-      case 'WalletVersion':
-        window.parent.postMessage(jrs.success(requestId, 'xstate-wallet@VersionTBD'), '*'); // TODO: inject git or build info
+      case 'GetWalletInformation':
+        window.parent.postMessage(
```
tomclose: This should use `this.sendResponse` instead of calling `window.parent.postMessage` directly.

snario:
As a note, we saw an error in the `App.test.tsx` case that seems to be caused by `jsdom` not implementing `setInterval` correctly. [Here](https://github.com/webtorrent/bittorrent-protocol/issues/44) is an issue on `webtorrent` that relates to that issue.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -63,22 +63,26 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+
     log('Enabling Channel Provider...');
     await window.channelProvider.enable();
-    log('Enabling WebTorrentPaidStreamingClient...');
-    this.pseAccount = this.paymentChannelClient.mySigningAddress;
+
     log('set pseAccount to sc-wallet signing address');
+    this.pseAccount = this.paymentChannelClient.mySigningAddress;
+
+    log('set outcomeAddress to sc-wallet web3 wallet address');
```
geoknee:
```suggestion
    log('set outcomeAddress to web3 wallet address');
```

PR #1298, opened 2020-03-24
Implement budget UI

junjizhi:
Fixes https://github.com/statechannels/monorepo/issues/1171

The budget comes from mock. Next I will try to wire up the budget update call.

The budget UI looks like this:

image

### DIFF
--- a/packages/web3torrent/src/components/torrent-info/download-info/DownloadInfo.tsx
+++ b/packages/web3torrent/src/components/torrent-info/download-info/DownloadInfo.tsx
@@ -9,6 +9,8 @@ import {ChannelState} from '../../../clients/payment-channel-client';
 import {utils} from 'ethers';
 import {ChannelsList} from '../channels-list/ChannelsList';
 import {prettyPrintWei} from '../../../utils/calculateWei';
+import {ChannelBudget} from '../channel-budget/ChannelBudget';
+import {createMockBudget} from '../../../utils/test-utils';
 
 const bigNumberify = utils.bigNumberify;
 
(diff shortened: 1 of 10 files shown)


packages/web3torrent/src/components/torrent-info/download-info/DownloadInfo.tsx
```diff
@@ -66,11 +68,14 @@ const DownloadInfo: React.FC<DownloadInfoProps> = ({
...
-            Total Spent: <strong className="total-spent">{prettyPrintWei(totalSpent)}</strong>
-          </p>
-        </section>
+        <>
+          <section className="totalPayed">
+            <p>
+              Total Spent: <strong className="total-spent">{prettyPrintWei(totalSpent)}</strong>
+            </p>
+          </section>
+          <ChannelBudget budget={createMockBudget()} />
```
geoknee: If I am uploading & downloading, it looks like I will see the `<ChannelBudget/>` component twice. Is that what we want?

geoknee: Perhaps it would be best to call the component `<SiteBudget/>` ?

I think it is best thought of as an integral part of the setup of web3torrent, and therefore shown as soon as a budget exists (rather than conditional on what is happening in any particular payment channel, or waiting until torrenting has begun).

junjizhi: > If I am uploading & downloading, it looks like I will see the <ChannelBudget/> component twice. Is that what we want?

Good catch. I thought peer can only be download or upload for the demo, so it would only show once. Anyway, I moved the component up level.

> Perhaps it would be best to call the component <SiteBudget/> ?
I renamed it to `SiteBudgetTable` so that it doesn't have a name conflict with `SiteBudget` type.

junjizhi:
> Do you think we are displaying redundant information with the Total Spent and Total Received fields (which should be the same as the Spent and Earned fields of your new component)?

I removed the Total Spent / Received in the text and replaced them with the budget table. See the screenshot above.

packages/web3torrent/src/components/torrent-info/download-info/DownloadInfo.tsx
geoknee: 👍 

(The plan is to allow non seeding peers to upload and download simultaneously)

geoknee:
On second thoughts, maybe we are labelling the  information incorrectly here. `inUse.hubAmount` is not the amount that I have received: it is the *maximum* amount I could receive if all of my uploading payment channels were exhausted. Similarly `inUse.playerAmount` is similarly the maximum amount I could pay out if all my downloading channels got exhausted.

geoknee:
I'll close this for now but we should reopen when #1302 lands and this component is rebuilt against the new interface.

PR #1302, opened 2020-03-25
Budget Refactor

lalexgap:
Updates the `budget` based on the discussion in stand up and https://www.notion.so/State-Channels-4cc770686d05438bb23b3c37f4069d7e?p=ce435f483e044d8291a8a01f570391d4

### DIFF
--- a/packages/channel-client/tests/fakes/fake-channel-provider.ts
+++ b/packages/channel-client/tests/fakes/fake-channel-provider.ts
@@ -323,7 +323,7 @@ export class FakeChannelProvider implements ChannelProviderInterface {
           token: '0x0',
           availableReceiveCapacity: params.requestedReceiveCapacity,
           availableSendCapacity: params.requestedSendCapacity,
-          channels: {}
+          channels: []
         }
       ]
     };
--- a/packages/channel-client/tests/fakes/fake-channel-provider.ts
+++ b/packages/channel-client/tests/fakes/fake-channel-provider.ts
@@ -10,7 +10,7 @@ import log = require('loglevel');
 
 import {EventEmitter} from 'eventemitter3';
 import {
-  BudgetRequest,
+  TokenBudgetRequest,
   CloseAndWithdrawParams,
   ChannelResult,
   CloseChannelParams,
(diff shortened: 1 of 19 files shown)


packages/client-api-schema/src/types.ts
```diff
@@ -167,28 +162,43 @@ export type ChallengeChannelRequest = JsonRpcRequest<'ChallengeChannel', {channe
...
+
 export interface TokenBudget {
   token: string;
-  pending: Balance;
-  free: Balance;
-  inUse: Balance;
-  direct: Balance;
+  availableReceiveCapacity: string;
+  availableSendCapacity: string;
+  channels: ChannelBudget[];
```
lalexgap: I switched this from a `Record` as specified in the notion doc to keep it consistent with `budgets`. I think we should either using a `Record` or `Array` consistently in the API types but I'm ok using either depending on what's easiest for `web3-torrent`.

packages/client-api-schema/src/types.ts
```diff
@@ -167,28 +162,43 @@ export type ChallengeChannelRequest = JsonRpcRequest<'ChallengeChannel', {channe
 export type ChallengeChannelResponse = JsonRpcResponse<ChannelResult>;
 
 // Budget
+
+export interface ChannelBudget {
+  channelId: string;
+  status: 'pending' | 'available';
```
tomclose: I think we decided that we didn't need the `status` - you can check the channel status to find this out.

packages/channel-client/tests/fakes/fake-channel-provider.ts
```diff
@@ -312,20 +312,18 @@ export class FakeChannelProvider implements ChannelProviderInterface {
...
-    const {hub, site, playerAmount, hubAmount} = params;
-
+  private async approveBudgetAndFund(params: TokenBudgetRequest): Promise<SiteBudget> {
     // TODO: Does this need to be delayed?
-    const result = {
-      hub: hub.signingAddress,
-      site,
+    const result: SiteBudget = {
+      hubAddress: params.hub.signingAddress,
+      domain: 'web3torrent.statechannels.org',
```
geoknee: Should we have a reference to web3torrent baked into the fake channel provider? Other apps such as RPS might (do) use it.

packages/client-api-schema/src/types.ts
geoknee: 2f84c95

packages/channel-client/tests/fakes/fake-channel-provider.ts
geoknee: ecd818d

PR #1307, opened 2020-03-25
Fix scary bug - Fix overpaying for last piece - Event Handling improvements

### LINKED ISSUE #1270: Leftover Budget Handling
**Problem description :confused:**
Once a leecher downloads his desired torrent from one or more seeders, there's always a leftover buffer in the seeders.

For example, with a *BUFFER_REFILL_RATE* of 20000 wei, a *WEI_PER_BYTE* of 1 wei, and a torrent size of 62kb, with a single seeder and a single leecher, after the leecher downloads the entire file, the seeder will have 18000 wei.
That buffer should be returned somehow to the leecher.

This, I believe, is the remaining issue in order to solve inconsistent costs of torrents, of a torrent cost not being what the UI tells the leecher it is.

**Describe the solution you'd like :rocket:**
We should implement a way for the seeders to return the leftover buffer to the leecher. Maybe this could be done before closing the channel?.

spersico:
It got bigger than I wanted.
closes #1270 
This PR:
- (1) Fixes a bug, where big torrents (seen on torrents from +30mb) wouldn't ever get downloaded.
- Adds a fix to avoid overpaying for the last piece (the first to download, ~_this was super easy_~ _It wasn't that easy at the end_).
- While trying to debug (1), the Notice event handler got somewhat refactored. The biggest change is that the messages that get to the library, are already parsed, and filtered.
- **Changes the way that requests are rejected and resumed:**
**Before this**, requests on the seeder side got dropped when the wire got blocked, and on a re-allowed wire, the leecher cleared up the request queue, and a couple of other data structures, and re-started making requests.
**Now** requests on the seeder side, when the leecher gets blocked, get saved in an array, that when the leecher pays and it's allowed, the seeder starts answering them. This avoids a/several race conditions, that made the leecher download more than once some parts of torrents, sometimes.


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -229,31 +229,31 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
             allowed: false,
-            channelId: channel.channelId
+            channelId
           };
+          log('> BLOCKED: ' + index + ' NEW PEER - Channel Created, id:' + channelId);
           response(false);
-
           this.blockPeer(torrent.infoHash, wire, peerAccount);
         } else if (!knownPeerAccount.allowed || reqPrice.gt(knownPeerAccount.buffer)) {
-          response(false);
```
spersico: this logs got moved around so that the log messages respect the actual order of execution.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -229,31 +229,31 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
             this.pseAccount, // seeder
             peerAccount, // leecher
             INITIAL_SEEDER_BALANCE.toString(), // seederBalance,
             INITIAL_LEECHER_BALANCE.toString(), // leecherBalance,
             this.paymentChannelClient.myEthereumSelectedAddress, // seederOutcomeAddress,
             peerOutcomeAddress // leecherOutcomeAddress
           );
-          log(`> created channel with id ${channel.channelId}`);
-          wire.paidStreamingExtension.pseChannelId = channel.channelId;
+          // eslint-disable-next-line require-atomic-updates
```
spersico: I don't know where this came out, I never saw it before, but the error kept showing up, so I added that line

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -175,15 +178,20 @@ export abstract class PaidStreamingExtension implements Extension {
...
-          _onRequest.apply(wire, [index, offset, length]);
-          log(`_onRequest PASS - ${index}`);
-        } else {
-          wire._onCancel(index, offset, length);
-          log(`_onRequest CHOKED - ${index}`);
-        }
-      });
+      if (this.paidStreamingExtension.isForceChoking) {
+        wire._onCancel(index, offset, length);
+        log(`_onRequest AUTO CHOKE - ${index}, ${offset}, ${length}`);
```
spersico: This makes it so that the library doesn't even get to see choked requests if it's supposed to be choking them.

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -131,17 +131,20 @@ export abstract class PaidStreamingExtension implements Extension {
...
         this.peerChannelId = data;
         if (this.isBeingChoked) return;
         this.isBeingChoked = true;
         break;
-      default:
+      case PaidStreamingExtensionNotices.MESSAGE:
+        data = JSON.parse(data.message);
+        if (data.recipient !== this.pseAccount) {
+          return;
+        }
```
spersico: this parses the message in the wire and filters it (by not re-emitting it)

packages/web3torrent/src/constants.ts
```diff
@@ -1,9 +1,9 @@
 import {Status, Torrent} from './types';
 import {ChannelState} from './clients/payment-channel-client';
 import {bigNumberify} from 'ethers/utils';
-
+export const BLOCK_LENGTH = 1 << 14; // Standard request length. Setting less than this doesn't make sense.
```
spersico: Looking into Webtorrent's, I saw that the chunk is always of a maximum of this length, so this looks like a good idea. Make the transfers always be a multiple of the maximum chunk length.
Most chunks are that size, except the last chunk of a torrent, and that one gets handled as an exception.

geoknee: How does a chunk or a block relate to a piece?

From below it looks like block = piece?

I don't see either term here https://en.wikipedia.org/wiki/Glossary_of_BitTorrent_terms

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -367,85 +370,86 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
     return torrent;
   }
 
+  /**
+   * Define the amount to pay, and makes the payment
+   */
+  protected async makePayment(torrent: PaidStreamingTorrent, wire: PaidStreamingWire) {
+    const {peerChannelId} = wire.paidStreamingExtension;
+    let amountToPay = WEI_PER_BYTE.mul(BUFFER_REFILL_RATE).toString();
+    if (torrent.downloaded === 0) {
```
geoknee: Perhaps add a comment here that the last piece is always downloaded first.

Is this *always* the case?

packages/web3torrent/src/constants.ts
spersico: A piece is composed of chunks. On small torrents (10mb or less, AFAIK), each piece has only one chunk. So in those torrents, the terms piece and chunks are pretty much the same thing.

On bigger torrents, pieces are split on smaller chunks, of a máximum of 16kib. That's what the offset in each request is for, to _tell_ the peer, which **part** of the piece is asking for (which chunk of the piece).

The issue that I found, is that as I lately tested this with small torrents, is that the jumpstart process didn't account for chunks. The details of the logic behind which piece is selected or ignored, is not... completely clear to me. I made a fix, that fixed that, and I'm trying to resolve all edge cases (for example, I'm now having issues, where multiple leechers end up paying more than they should, even with this fix).

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: Sure, I'll add it.

As far as I tested it, yes. The last piece is always requested first. I should test in 1gb+ torrents to make sure, but I observed that behaviour in all my test files (60kb, 120kb, 1.1mb, 3mb 9mb and 33mb)

packages/web3torrent/src/constants.ts
spersico: Looking into this, the model change is made at the 25mb threshold, it seems. Before that, 1 piece is equivalent to 1 chunk.

packages/web3torrent/src/constants.ts
```diff
@@ -3,13 +3,19 @@ import {ChannelState} from './clients/payment-channel-client';
 import {bigNumberify} from 'ethers/utils';
 
 export const WEI_PER_BYTE = bigNumberify(1); // cost per byte
-export const BUFFER_REFILL_RATE = bigNumberify(2e4); // number of bytes the leecher wishes to increase the buffer by
+export const BLOCK_LENGTH = 1 << 14; // Standard request length.
+export const PEER_TRUST = 1; //amount of trust between peers. It's equivalent to the amount of request to pre-pay.
+// The recomended value is 5 ( the size of the queue of requests made by the leecher to the seeder)
```
geoknee: Perhaps we should set to the recommended value?

packages/web3torrent/src/library/paid-streaming-extension.ts
```diff
@@ -91,19 +94,31 @@ export abstract class PaidStreamingExtension implements Extension {
   stop() {
     if (!this.isForceChoking) {
       this.isForceChoking = true;
-      this.executeExtensionCommand(PaidStreamingExtensionNotices.STOP, this.pseChannelId);
+      setTimeout(() => {
+        this.executeExtensionCommand(PaidStreamingExtensionNotices.STOP, this.pseChannelId);
+      }, 0);
```
geoknee: Any clue why we need this?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -186,7 +186,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       torrentInfoHash,
       peerAccount
     });
-    log('SEEDER: > unblockedPeer', peerAccount, 'from', Object.keys(this.peersList));
+    log('<< unblockedPeer', peerAccount, 'from', Object.keys(this.peersList));
```
geoknee: What's the thinking behind this notation?

geoknee:
I think I am still seeing overpayments:

Screenshot 2020-03-30 at 11 34 33

Note discrepancy between the price in the magnet link, and the balance of the beneficiary when the download was finished.

geoknee:
I was able to finish a 55MB file, by setting `PEER_TRUST` to `50` (I tried with `1` and `5` but it was very slow, and I found the download speed seemed to significantly degrade over time).

packages/web3torrent/src/constants.ts
geoknee: It seems to me that `PEER_TRUST` should be proportional to an amount of money, rather than an amount of requests. My concern is that with different sized files, we have different sized pieces, and for the same `PEER_TRUST` we might be pre-paying more or less money?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -367,85 +365,90 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+    // On each wire, the algorithm tries to download the uneven piece (which is always the last piece)
+
+    const lastPieceReservations: PaidStreamingWire[] =
+      torrent._reservations[torrent.pieces.length - 1];
+    const lastPieceReservationsIncludesThisWire =
+      lastPieceReservations &&
+      lastPieceReservations.find(
+        wire => wire && wire.paidStreamingExtension.peerAccount === peerAccount
+      );
+    if (torrent.downloaded === 0 && lastPieceReservationsIncludesThisWire) {
```
junjizhi: It would be cleaner if we can wrap the Line 410-417 logic in a separate function, e.g., `isLastPiece(torrent, wire)`

packages/web3torrent/src/constants.ts
spersico: It isn't, although the size of the pieces can vary, the maximum size of a request is constant. It's the BLOCK_LENGTH size. That's what I was explaining on my comment about chunks and pieces.

This allows us to not having any variance in payment once the "uneven" parts change, using a simple variable. For example, usually, the queue is made up of 5 requests to each wire. Setting the PEER_TRUST to 5 is, to me, the optimum size, as it would optimize the communication flow between the peers.

packages/web3torrent/src/library/paid-streaming-extension.ts
spersico: This was actually added on the previous PR about Payments, and I still quite don't get why it's necessary. It's about clearing the call stack (to respond to the requests in time) before letting the leecher know that it should stop. But I don't know if it's still necessary.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: Everything which is preceded by `<<` it's done by the peer in a Leecher role.
Everything which is preceded by `>>` it's done by the peer in a Seeder role.
This allows me to filter what was important to me in multi-peer situations.

packages/web3torrent/src/library/types.ts
```diff
@@ -91,13 +91,18 @@ export type OverridenTorrentProperties = 'pieces';
 
 export type ExtendedTorrentPiece = WebTorrent.TorrentPiece & {
   _reservations: number;
+  _chunks: number;
+  length: number;
+  _buffered: number;
+  _buffer: any[];
+  _cancellations: any[];
```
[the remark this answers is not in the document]
spersico: Umm, no, it seems like it's not. I'll remove them, as soon as I confirm what's going on with @geoknee's case, just to make sure that I don't need them.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -294,7 +292,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
         // do not pass a channelId, since this is the first we heard about this channel and it won't be cached
         // only join if counterparty proposed
         await this.paymentChannelClient.joinChannel(channelState.channelId);
-        log(`Joined channel ${channelState.channelId}`);
+        log(`<<Joined channel ${channelState.channelId}`);
```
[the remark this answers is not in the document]
spersico: :-P yes! I didn't see it!.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: :+1:

packages/web3torrent/src/library/paid-streaming-extension.ts
junjizhi: What happen if we undo this `setTimeout()`?

spersico: if we run the code one tab beside the other, the download never finishes, it gets stuck at the beginning. I've got yet to test what happens if run in multiple browsers, but that first one seemed like a good enough reason to leave it as is

PR #1309, opened 2020-03-25
xstate-wallet: add arrow-body-style rule

kerzhner:
Probably not the best time to add this rule as the wallet is under rapid development, but this would be nice in the near future. Makes our code more readable in my opinion.

### DIFF
--- a/packages/xstate-wallet/.eslintrc.js
+++ b/packages/xstate-wallet/.eslintrc.js
@@ -13,7 +13,6 @@ const leftoverTsLintRules = {
   '@typescript-eslint/ban-ts-ignore': 'off'
 };
 
-
 module.exports = {
   env: {
     browser: true,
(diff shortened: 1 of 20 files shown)


lalexgap:
It might be a `vscode` setting but I find it really hard to place breakpoints when we have inlined functions like this.

kerzhner:
@lalexgap, @andrewgordstewart and I chatted about this today. In VSCode, you can add an inline breakpoint. Until the most recent version of VSCode, this was available in the right click menu. In VSCode 1.43.2, an inline breakpoint can be added either through menu selection or via a shortcut.

For chrome debugging:
Screen Shot 2020-03-25 at 8 53 14 PM

andrewgordstewart:
> @lalexgap, @andrewgordstewart and I chatted about this today. In VSCode, you can add an inline breakpoint. Until the most recent version of VSCode, this was available in the right click menu. In VSCode 1.43.2, an inline breakpoint can be added either through menu selection or via a shortcut.

FYI the vs-code shortcut is shift+F9 to place an inline breakpoint at your cursor.

PR #1314, opened 2020-03-26
[xstate-wallet][channel-provider] Wait for wallet to be ready

geoknee:
...before returning from `mountWalletComponent()`.


packages/channel-provider/src/channel-provider.ts
```diff
@@ -39,6 +39,14 @@ class ChannelProvider implements ChannelProviderInterface {
     this.messaging = new MessagingService();
   }
 
+  walletReady = new Promise(resolve => {
+    window.addEventListener('message', event => {
+      if (event.data === 'WalletReady') {
+        resolve();
+      }
+    });
```
andrewgordstewart:
```suggestion
    window.addEventListener('message', event => event.data === 'WalletReady' && resolve());
```

PR #1315, opened 2020-03-26
Fix firebase communication in web3torrent.

andrewgordstewart:
This also adds an optional auto-`approveBudgetAndFund` workflow, triggered by setting `REACT_APP_AUTO_FUND_LEDGER` to a truthy value.


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -14,13 +14,13 @@ import {
   WebTorrentSeedInput,
   WireEvents
 } from './types';
-import {utils} from 'ethers';
+import {utils, ethers} from 'ethers';
```
geoknee: Not sure you want to import both.

PR #1318, opened 2020-03-27
[web3torrent] Remove {} from type for budgetCache

geoknee:
When there is no budget, use `undefined`.

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -37,7 +37,7 @@ if (process.env.REACT_APP_FAKE_CHANNEL_PROVIDER === 'true') {
 // The payer joins the channel, and makes payments
 export class PaymentChannelClient {
   channelCache: Record<string, ChannelState> = {};
-  budgetCache: SiteBudget | {} = {};
+  budgetCache?: SiteBudget;
 
   get mySigningAddress(): string | undefined {
     return this.channelClient.signingAddress;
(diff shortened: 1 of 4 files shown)


packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -61,15 +61,14 @@ const File: React.FC<RouteComponentProps & Props> = props => {
...
-            Object.keys(budgetCache).length > 0 && Object.keys(channelCache).length > 0;
+          const showBudget = budgetCache && Object.keys(channelCache).length > 0;
           return (
             <>
               <TorrentInfo torrent={torrent} channelCache={channelCache} mySigningAddress={me} />
               <br />
               {showBudget ? (
                 <SiteBudgetTable
-                  budgetCache={budgetCache as SiteBudget}
+                  budgetCache={budgetCache}
```
geoknee: No need for an assertion any more.

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -37,7 +37,7 @@ if (process.env.REACT_APP_FAKE_CHANNEL_PROVIDER === 'true') {
 // The payer joins the channel, and makes payments
 export class PaymentChannelClient {
   channelCache: Record<string, ChannelState> = {};
-  budgetCache: SiteBudget | {} = {};
+  budgetCache?: SiteBudget;
```
geoknee: Property is optional.

PR #1319, opened 2020-03-27
[web3torrent] Fix possible race condition

geoknee:
When transitioning from the fake provider to a real wallet, channel updates are going to get significantly slower due to the burden of managing signatures. Meanwhile, web3torrent messages that do not involve the wallet will be relatively fast. 

This means that although the seeder will send the spacer state, *then* prompt the leecher to make a payment, the leecher may well respond to the prompt before the channel has been updated in their client.

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -183,7 +183,7 @@ export class PaymentChannelClient {
       ) {
         resolve();
       } else {
-        this.channelClient.onChannelProposed(channelResult => {
+        this.channelClient.onChannelUpdated(channelResult => {
           const channelState = convertToChannelState(channelResult);
           channelState.channelId === channelId &&
             this.isAcceptanceOfMyPayment(channelState) &&
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -23,8 +23,8 @@ export interface ChannelState {
 }
 
 enum Index {
-  Leecher = 0,
-  Seeder = 1
+  Payer = 0,
+  Beneficiary = 1
 }
 
 // This class wraps the channel client converting the
@@ -187,7 +187,7 @@ export class PaymentChannelClient {
         state &&
         state.status === 'running' &&
         state.payer === this.mySigningAddress &&
-        state.turnNum.mod(2).eq(Index.Leecher);
+        state.turnNum.mod(2).eq(Index.Payer);
 
       const currentState = this.channelCache[channelId];
       if (readyToPay(currentState)) resolve(currentState);
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -21,6 +21,11 @@ export interface ChannelState {
   payerBalance: string;
 }
 
+const enum Index {
+  Leecher = 0,
+  Seeder = 1
+}
+
 // This class wraps the channel client converting the
 // request/response formats to those used in the app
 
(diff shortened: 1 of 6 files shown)


geoknee:
It works 🎉 !

packages/xstate-wallet/.env
```diff
@@ -29,8 +29,7 @@ TEST_ASSET_HOLDER2_ADDRESS = 0x0000000000000000000000000000000000000000
 
 # TODO: This should be something the wallet/user decides not a .env variable. 
 # For now this allows integration tests to test ledger funding
-USE_VIRTUAL_FUNDING=TRUE
```
kerzhner: Is the removal of environment variables intentional?

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -176,6 +176,24 @@ export class PaymentChannelClient {
...
   // payer may use this method to make payments (if they have sufficient funds)
   async makePayment(channelId: string, amount: string) {
+    const channelReady = new Promise(resolve => {
+      if (
+        this.channelCache[channelId] &&
+        this.isAcceptanceOfMyPayment(this.channelCache[channelId])
+      ) {
+        resolve();
+      } else {
+        this.channelClient.onChannelUpdated(channelResult => {
```
kerzhner: Should unsubscribe get called before promise resolution?

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -21,6 +21,11 @@ export interface ChannelState {
   payerBalance: string;
 }
 
+const enum Index {
+  Leecher = 0,
+  Seeder = 1
+}
+
```
geoknee: I like this approach but would prefer to use `payer` and `beneficiary` terminology; this class need not use any torrent specific terms as it could in principle be used in other apps. 

(Also as previously discussed, we can end up with payment channels between peers, none of whom are seeders).

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -41,7 +42,7 @@ if (process.env.REACT_APP_FAKE_CHANNEL_PROVIDER === 'true') {
 // The beneficiary proposes the channel, but accepts payments
 // The payer joins the channel, and makes payments
 export class PaymentChannelClient {
-  channelCache: Record<string, ChannelState> = {};
+  channelCache: Record<string, ChannelState | undefined> = {};
```
geoknee: Why do we need this? Is there a situation where you want to store the `channelId` but do not yet have a state for it?

packages/xstate-wallet/.env
geoknee: No, but it does reflect the conditions under which these changes led to a click through test succeeding. Looks like we leave them in place but preserve those conditions in bcabeed.

packages/web3torrent/src/clients/payment-channel-client.ts
geoknee: Good catch, see a0072ba

packages/web3torrent/src/clients/payment-channel-client.ts
geoknee: See 176c58b

geoknee:
Confirmed this works at 6a53184, with web3t started with `REACT_APP_AUTO_FUND_LEDGER= yarn start` and xstate-wallet started with `ADD_LOGS=true USE_VIRTUAL_FUNDING= yarn start`.

packages/web3torrent/src/clients/payment-channel-client.ts
andrewgordstewart: This is for type safety -- if you access `this.channelCache['someRandomString']`, you get a value of type `ChannelState` even though the value is probably `undefined`.

geoknee: Good point. https://fnune.com/typescript/2019/01/30/typescript-series-1-record-is-usually-not-the-best-choice/

geoknee: There are some other places we could think about fixing this, including most places where we do `Record<string,T>` and https://github.com/statechannels/monorepo/blob/20112d7a213340b42429247fc195b885873fc599/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx#L14

PR #1326, opened 2020-03-27
xstate-wallet: add launch config for attaching to chrome

kerzhner:
Great reference doc: [https://github.com/microsoft/vscode-chrome-debug](https://github.com/microsoft/vscode-chrome-debug)

- Launch chrome via `/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --remote-debugging-port=9222 --user-data-dir=[PATH TO DIR STORING CHROME USER DATA]/chrome-debug`. Reading [this](https://github.com/microsoft/vscode-chrome-debug#cannot-connect-to-the-target-connect-econnrefused-1270019222) will clarify the command.
- Run *Attach to running wallet* in VSCode.
- Set breakpoints to your heart's content.

### DIFF
--- a/packages/xstate-wallet/.vscode/launch.json
+++ b/packages/xstate-wallet/.vscode/launch.json
@@ -18,7 +18,6 @@
         "webpack:///./~/*": "${webRoot}/node_modules/*"
       }
     },
-
     {
       "name": "Jest Current",
       "type": "node",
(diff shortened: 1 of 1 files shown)


kerzhner:
Also of note (@geoknee might be useful to you): it is easy to start many chrome instances with different `user-data-dir` paths. Each chrome instance will have its own Metamask extension and its own IndexedDB instances. The benefit is being able to use Chrome debug tools (for all participants) instead of Firefox or Opera if those are either not working or have an insufficient feature set.

PR #1329, opened 2020-03-28
Web3torrent: Add end-to-end tests with Puppeteer

junjizhi:
Addresses: https://github.com/statechannels/monorepo/issues/1328

Assumption: The browser setup of web3torrent is the same as rps, so I reuse the helper functions. 

- [x] Not clear whether / how much xstate-wallet UI with web3torrent will change. That might impact the pupeteer DOM elements listening.
- [x] Implement the helpers to listen to DOM elements and invoke events.
- [x] Figure out how to set up web3torrent apps in CI (right now it listens to localhost:3000 which I think is point to rps.

### DIFF
--- a/packages/e2e-tests/puppeteer/__tests__/rps.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/rps.test.ts
@@ -2,7 +2,7 @@
 import {Page, Browser} from 'puppeteer';
 import {configureEnvVariables, getEnvBool} from '@statechannels/devtools';
 
-import {setUpBrowser, loadWeb3App} from '../helpers';
+import {setUpBrowser, loadDapp} from '../helpers';
 import {
   login,
   aChallenges,
@@ -29,8 +29,12 @@ describe('completes game 1 (challenge by A, challenge by B, resign by B) and beg
     rpsTabA = (await browserA.pages())[0];
     rpsTabB = (await browserB.pages())[0];
 
-    await loadWeb3App(rpsTabA, 0);
-    await loadWeb3App(rpsTabB, 1);
+    await loadDapp(rpsTabA, 0);
+    await loadDapp(rpsTabB, 1);
+
+    const url = 'http://localhost:3000';
+    await rpsTabA.goto(url, {waitUntil: 'load'});
+    await rpsTabB.goto(url, {waitUntil: 'load'});
 
     await login(rpsTabA, rpsTabB);
   });
--- a/packages/e2e-tests/puppeteer/__tests__/rps.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/rps.test.ts
@@ -2,7 +2,7 @@
 import {Page, Browser} from 'puppeteer';
 import {configureEnvVariables, getEnvBool} from '@statechannels/devtools';
 
-import {setUpBrowser, loadRPSApp} from '../helpers';
+import {setUpBrowser, loadWeb3App} from '../helpers';
 import {
   login,
   aChallenges,
@@ -29,8 +29,8 @@ describe('completes game 1 (challenge by A, challenge by B, resign by B) and beg
     rpsTabA = (await browserA.pages())[0];
     rpsTabB = (await browserB.pages())[0];
 
-    await loadRPSApp(rpsTabA, 0);
-    await loadRPSApp(rpsTabB, 1);
+    await loadWeb3App(rpsTabA, 0);
+    await loadWeb3App(rpsTabB, 1);
 
     await login(rpsTabA, rpsTabB);
   });
--- a/packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
@@ -57,6 +57,8 @@ describe('Supports torrenting among peers with channels', () => {
     await waitForClosingChannel(web3tTabB);
     await waitForClosingChannel(web3tTabA);
 
+    await web3tTabA.waitFor(1500);
+    await web3tTabB.waitFor(1500);
     console.log('Checking exchanged amount between downloader and uploader...');
     const earnedColumn = await web3tTabA.$('td.earned');
     const earned = await web3tTabA.evaluate(e => e.textContent, earnedColumn);
--- a/packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
@@ -50,6 +50,7 @@ describe('Supports torrenting among peers with channels', () => {
     await waitAndOpenChannel(web3tTabA);
     await waitAndOpenChannel(web3tTabB);
 
+    // Let the download cointinue for some time
     await web3tTabB.waitFor(3000);
     console.log('B cancels download');
     await cancelDownload(web3tTabB);
@@ -57,8 +58,10 @@ describe('Supports torrenting among peers with channels', () => {
     await waitForClosingChannel(web3tTabB);
     await waitForClosingChannel(web3tTabA);
 
+    // Inject some delays. Otherwise puppeteer may read the stale amounts and fails.
     await web3tTabA.waitFor(1500);
     await web3tTabB.waitFor(1500);
+
     console.log('Checking exchanged amount between downloader and uploader...');
     const earnedColumn = await web3tTabA.$('td.earned');
     const earned = await web3tTabA.evaluate(e => e.textContent, earnedColumn);
--- a/packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
@@ -54,5 +54,12 @@ describe('Supports torrenting among peers with channels', () => {
 
     await waitForClosingChannel(web3tTabB);
     await waitForClosingChannel(web3tTabA);
+
+    // Assert exchanged amount are the same on both sides
+    const earnedColumn = await web3tTabA.$('td.earned');
+    const earned = await web3tTabA.evaluate(e => e.textContent, earnedColumn);
+    const paidColumn = await web3tTabB.$('td.paid');
+    const paid = await web3tTabB.evaluate(e => e.textContent, paidColumn);
+    expect(paid).toEqual(`-${earned}`);
   });
 });
--- a/packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
@@ -39,7 +39,7 @@ describe('Supports torrenting among peers with channels', () => {
   });
 
   it('allows peers to start torrenting', async () => {
-    console.log('A uploads a file...');
+    console.log('A uploads a file');
     const url = await uploadFile(web3tTabA);
 
     console.log('B starts downloading...');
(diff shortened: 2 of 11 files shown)


geoknee:
> * Figure out how to set up web3torrent apps in CI (right now it listens to localhost:3000 which I think is point to rps.

I think *both* rps and web3torrent are configured to serve on this port, currently. Here's where the integration test on circle decides to run the old wallet and rps:

https://github.com/statechannels/monorepo/blob/master/.circleci/config.yml#L263-L268

Here's where the test is disabled:

https://github.com/statechannels/monorepo/blob/master/.circleci/config.yml#L315-L318

For a start, you could just clone the workflow, undisable it, and swap rps/wallet for web3t/xstate-wallet

packages/e2e-tests/puppeteer/__tests__/rps.test.ts
```diff
@@ -2,7 +2,7 @@
 import {Page, Browser} from 'puppeteer';
 import {configureEnvVariables, getEnvBool} from '@statechannels/devtools';
 
-import {setUpBrowser, loadRPSApp} from '../helpers';
+import {setUpBrowser, loadWeb3App} from '../helpers';
```
geoknee, line 5: Not sure you want to touch this file?

packages/e2e-tests/puppeteer/scripts/rps.ts
```diff
@@ -1,6 +1,6 @@
 import {Page} from 'puppeteer';
 
-import {waitForAndClickButton, setUpBrowser, loadRPSApp} from '../helpers';
+import {waitForAndClickButton, setUpBrowser, loadWeb3App} from '../helpers';
```
geoknee, line 3: Not sure you want to touch this file either.

packages/e2e-tests/puppeteer/helpers.ts
```diff
@@ -92,3 +105,17 @@ export async function setUpBrowser(headless: boolean, slowMo?: number): Promise<
 
   return browser;
 }
+
+export async function waitAndOpenChannel(page: Page): Promise<void> {
+  const createChannelButton =
+    'body > div:nth-child(3) > div > div > div:nth-child(2) > div > button';
```
geoknee, line 111: Should we make this more robust to future changes in the UI? I think elsewhere we have used id selectors, which might be better.

packages/e2e-tests/puppeteer/scripts/web3torrent.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+import {Page} from 'puppeteer';
+
+export async function uploadFile(page: Page): Promise<string> {
+  await page.waitForSelector('input[type=file]');
+  const fileToUpload = 'file.pdf';
```
geoknee, line 6: Is this just the name of a file, or actually the file contents itself? If the latter, it seems as if the download might finish extremely quickly. Depending on what the test expects to happen, that might lead to problems?

packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    console.log('B starts downloading...');
+    await startDownload(web3tTabB, url);
+
+    await waitAndOpenChannel(web3tTabA);
+    await waitAndOpenChannel(web3tTabB);
+
+    await web3tTabB.waitFor(3000);
+    console.log('B cancels download...');
+    await cancelDownload(web3tTabB);
```
geoknee, line 53: Would be great if we could expect at least some data to have been transferred. This test may pass even when a lot of the functionality is broken.

packages/e2e-tests/puppeteer/scripts/web3torrent.ts
geoknee: I understand that it must be the name of a file. I had to make `file.pdf` in the `e2e-test` directory to get this to work. Recommend that you check this file in.

I think the point still stands that the test behaviour will generally depend on the file size, as well as the speed of the internet connection of the machine the test is run on.

packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  beforeAll(async () => {
+    // 100ms sloMo avoids some undiagnosed race conditions
+    browserA = await setUpBrowser(HEADLESS, 100);
+    browserB = await setUpBrowser(HEADLESS, 100);
+
+    web3tTabA = (await browserA.pages())[0];
+    web3tTabB = (await browserB.pages())[0];
+
+    await loadWeb3App(web3tTabA, 0, 'file/new', true);
+    await loadWeb3App(web3tTabB, 0, '', true);
```
geoknee, line 29: To speed things up a bit, you don't need to load web3torrent in tab B, since you will navigate away from the homepage later anyway.

junjizhi:
> I tried to get this working locally and saw

Did you allow chromium for the incoming traffic? There was a prompt every time I ran puppeteer and I have to press `allow`.

packages/e2e-tests/puppeteer/__tests__/rps.test.ts
junjizhi: It's a simple rename so I can reuse the function.

packages/e2e-tests/puppeteer/scripts/rps.ts
junjizhi: See the comment above.

packages/e2e-tests/puppeteer/__tests__/rps.test.ts
geoknee: Ah I see. You're using `Web3App` to stand for rps or web3torrent or something of that type. Other possible names:

- `loadApp` 
- `loadDapp`
- `loadStateChannelApp`.

geoknee:
I do not see a prompt.

packages/e2e-tests/puppeteer/helpers.ts
junjizhi: done.

packages/e2e-tests/puppeteer/scripts/web3torrent.ts
junjizhi: I added a step in CI to create a 4MB file with random data for seeding.

packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
junjizhi: done.

packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
junjizhi: Actually we need that step to setup the browser for web3. So this step is needed.

packages/e2e-tests/puppeteer/scripts/web3torrent.ts
geoknee: OK. Nitpick follows!

I'm curious as to the advantages of that approach versus checking in a file. The downsides I can see are that this test won't run out of the box in local development, and that we are introducing some indeterminism into our tests which may not be desirable. (Highly likely that the random data doesn't matter, but then why make it random in the first place?).

packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
geoknee: That step could be broken into separate functions so we can avoid the unnecessary page load.  I suppose it takes a little more refactoring to achieve, so we could tackle it in a future PR. Bottom line, we call `web3tTabB.goto` twice and we need only call it once. (It takes about 5s on my machine, so maybe it is not a big deal).

packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    await waitForClosingChannel(web3tTabB);
+    await waitForClosingChannel(web3tTabA);
+
+    // Assert exchanged amount are the same on both sides
+    const earnedColumn = await web3tTabA.$('td.earned');
+    const earned = await web3tTabA.evaluate(e => e.textContent, earnedColumn);
+    const paidColumn = await web3tTabB.$('td.paid');
+    const paid = await web3tTabB.evaluate(e => e.textContent, paidColumn);
+    expect(paid).toEqual(`-${earned}`);
```
geoknee, line 63: 👍

packages/e2e-tests/puppeteer/scripts/web3torrent.ts
junjizhi: My thinking was to reduce the git commit size. But I agree. It's not worth the hassle to create the file locally by hand. I added the file in the latest git commit.

packages/e2e-tests/puppeteer/__tests__/web3torrent.test.ts
junjizhi: Done. I took out `page.goto(...)` from the helper function.

junjizhi:
@geoknee 

The updated test may have caught a bug:

```
Supports torrenting among peers with channels › allows peers to start torrenting

    expect(received).toEqual(expected) // deep equality

    Expected: "-180.2 kwei"
    Received: "-475.1 kwei"

      63 |     const paidColumn = await web3tTabB.$('td.paid');
      64 |     const paid = await web3tTabB.evaluate(e => e.textContent, paidColumn);
    > 65 |     expect(paid).toEqual(`-${earned}`);
         |                  ^
      66 |   });
      67 | });
      68 | 

      at Object.it (puppeteer/__tests__/web3torrent.test.ts:65:18)Supports torrenting among peers with channels › allows peers to start torrenting

    expect(received).toEqual(expected) // deep equality

    Expected: "-180.2 kwei"
    Received: "-475.1 kwei"

      63 |     const paidColumn = await web3tTabB.$('td.paid');
      64 |     const paid = await web3tTabB.evaluate(e => e.textContent, paidColumn);
    > 65 |     expect(paid).toEqual(`-${earned}`);
         |                  ^
      66 |   });
      67 | });
      68 | 

      at Object.it (puppeteer/__tests__/web3torrent.test.ts:65:18)
```

packages/e2e-tests/puppeteer/scripts/web3torrent.ts
andrewgordstewart: Can you programmatically generate a random 4MB file (that doesn't get checked in?)

junjizhi: That's what I did. @geoknee suggests we commit it in the git so that when you run the tests locally you don't have to remember generating the file. 4M is not that big anyway.

See https://github.com/statechannels/monorepo/pull/1329/files#diff-9ed3cedce43d8747b5dc9942f0b66cedR6

geoknee: I'll just note that we can have all of

- deterministic tests
- run out of the box
- no large files checked in

We just need to avoid 1) generating something *random*, 2) relying on something outside test script such as a command on circle. So perhaps a deterministic generation of a file (or Buffer?) within the jest test https://nodejs.org/api/buffer.html#buffer_class_method_buffer_alloc_size_fill_encoding

geoknee:
@junjizhi do you see this problem on a standard, manual click-through?

junjizhi:
@geoknee Yeah. It turns out, after rebasing on master, the amount update became slower. And puppeteer was reading the stale value. I injected a small delay to make it pass. A bit ugly, but it seems to work fine.

packages/e2e-tests/puppeteer/scripts/web3torrent.ts
junjizhi: In the latest commit, I got rid of the checked-in large file and programmatically create a `/tmp/web3torrent-test` instead. `/tmp/` is usually RAM backed partition so it should be fast to read and write data there without hitting the disk.

PR #1334, opened 2020-03-30
xstate-wallet: force USE_VIRTUAL_FUNDING into a boolean

kerzhner:


packages/xstate-wallet/src/constants.ts
```diff
@@ -37,3 +37,12 @@ export function tokenAddress(assetHolderAddress: string): string | undefined {
...
   return assetHolderAddress;
 }
+
+function envVarToBoolean(envVar: string | undefined): boolean {
+  if (!envVar) return false;
+  const upperEnvVar = envVar.toLocaleUpperCase();
+  if (upperEnvVar === 'FALSE') return false;
+  if (upperEnvVar === 'TRUE') return true;
+  throw new Error(`Environment variable ${envVar} must be a boolean`);
+}
```
geoknee: Just wanted to point out that we have this utility in devtools:

https://github.com/statechannels/monorepo/blob/20112d7a213340b42429247fc195b885873fc599/packages/devtools/src/config/env.ts#L4-L21

kerzhner: Good catch

kerzhner: f170ebb7c

PR #1337, opened 2020-03-31
Strengthen Request validation

geoknee:
Has the side effect that concluding now works :-).

### DIFF
--- a/packages/client-api-schema/src/__tests__/good_sample_requests.ts
+++ b/packages/client-api-schema/src/__tests__/good_sample_requests.ts
@@ -95,8 +95,14 @@ export const createChannel = {
       {
         token: '0x0',
         allocationItems: [
-          {destination: '0x63e3fb11830c01ac7c9c64091c14bb6cbaac9ac7', amount: '0x06f05b59d3b20000'},
-          {destination: '0x63e3fb11830c01ac7c9c64091c14bb6cbaac9ac7', amount: '0x06f05b59d3b20000'}
+          {
+            destination: '0x63e3fb11830c01ac7c9c64091c14bb6cbaac9ac7',
+            amount: '0x00000000000000000000000000000000000000000000000006f05b59d3b20000'
+          },
+          {
+            destination: '0x63e3fb11830c01ac7c9c64091c14bb6cbaac9ac7',
+            amount: '0x00000000000000000000000000000000000000000000000006f05b59d3b20000'
+          }
         ]
       }
     ],
@@ -148,11 +154,11 @@ export const pushMessage2 = {
             {
               allocationItems: [
                 {
-                  amount: '0x06f05b59d3b20000',
+                  amount: '0x00000000000000000000000000000000000000000000000006f05b59d3b20000',
                   destination: '0x00000000000000000000000063e3fb11830c01ac7c9c64091c14bb6cbaac9ac7'
                 },
                 {
-                  amount: '0x06f05b59d3b20000',
+                  amount: '0x00000000000000000000000000000000000000000000000006f05b59d3b20000',
                   destination: '0x00000000000000000000000063e3fb11830c01ac7c9c64091c14bb6cbaac9ac7'
                 }
               ],
--- a/packages/client-api-schema/src/generated-schema.json
+++ b/packages/client-api-schema/src/generated-schema.json
@@ -29,7 +29,7 @@
       "additionalProperties": false,
       "properties": {
         "amount": {
-          "type": "string"
+          "$ref": "#/definitions/Uint256"
         },
         "destination": {
           "$ref": "#/definitions/Address"
@@ -126,6 +126,11 @@
       ],
       "type": "object"
     },
+    "Bytes32": {
+      "description": "Bytes32",
+      "pattern": "^0x([a-fA-F0-9]{64})$",
+      "type": "string"
+    },
     "ChallengeChannelParams": {
       "additionalProperties": false,
       "properties": {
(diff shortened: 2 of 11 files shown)


packages/client-api-schema/src/__tests__/good_sample_requests.ts
```diff
@@ -95,8 +95,14 @@ export const createChannel = {
       {
         token: '0x0',
         allocationItems: [
-          {destination: '0x63e3fb11830c01ac7c9c64091c14bb6cbaac9ac7', amount: '0x06f05b59d3b20000'},
-          {destination: '0x63e3fb11830c01ac7c9c64091c14bb6cbaac9ac7', amount: '0x06f05b59d3b20000'}
+          {
+            destination: '0x63e3fb11830c01ac7c9c64091c14bb6cbaac9ac7',
+            amount: '0x00000000000000000000000000000000000000000000000006f05b59d3b20000'
```
andrewgordstewart: Why does this need to be padded?

packages/client-api-schema/src/generated-schema.json
```diff
@@ -1401,6 +1406,11 @@
       ],
       "type": "object"
     },
+    "Uint256": {
+      "description": "Uint256",
+      "pattern": "^0x([a-fA-F0-9]{64})$",
```
andrewgordstewart: I don't think this needs to be 32 bytes. 
```suggestion
      "pattern": "^0x([a-fA-F0-9]{0,64})$",
```

geoknee:
Painful: `ajv` errors are hardly human readable. I had tried using https://github.com/atlassian/better-ajv-errors which is much better, but unfortunately not support in the browser at present. 

Horrible: so many `hexZeroPad` everywhere :-(

packages/client-api-schema/src/types.ts
```diff
@@ -16,6 +16,18 @@ export type ChannelId = string;
...
 
+/**
+ * Bytes32
+ * @pattern  ^0x([a-fA-F0-9]{64})$
+ */
+export type Bytes32 = string;
+
+/**
+ * Uint256
+ * @pattern  ^0x([a-fA-F0-9]{64})$
```
andrewgordstewart:
```suggestion
 * @pattern  ^0x([a-fA-F0-9]{0,64})$
```

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -318,8 +320,14 @@ const convertToChannelState = (channelResult: ChannelResult): ChannelState => {
...
-    beneficiaryBalance: bigNumberify(allocations[0].allocationItems[0].amount).toString(),
-    payerBalance: bigNumberify(allocations[0].allocationItems[1].amount).toString()
+    beneficiaryBalance: hexZeroPad(
+      bigNumberify(allocations[0].allocationItems[0].amount).toHexString(),
+      32
+    ),
+    payerBalance: hexZeroPad(
+      bigNumberify(allocations[0].allocationItems[1].amount).toHexString(),
+      32
+    )
```
andrewgordstewart: Unnecessary if the validation schema is changed as per above.

packages/xstate-wallet/src/workflows/conclude-channel.ts
```diff
@@ -19,6 +19,7 @@ const finalState = (store: Store) => async (context: Init): Promise<SupportState
 
   // If we've received a new final state that matches our outcome we support that
   if (outcomesEqual(latestSupportedByMe.outcome, latestFinalState?.outcome)) {
+    // outcomesEqual function will return false if there is inconsistent padding of hex strings
```
andrewgordstewart: What is this comment for? It's my understanding that `outcomesEqual` should _not_ return `false` in this case.

geoknee: The comment is telling you that the function does not do what you think it should.

andrewgordstewart: Shouldn't it be on `outcomesEqual` then?

packages/client-api-schema/src/types.ts
geoknee: I copied this from the wire-format validator in the xstate wallet.

packages/client-api-schema/src/generated-schema.json
geoknee: See below comment (this file autogenerated anyway)

packages/client-api-schema/src/__tests__/good_sample_requests.ts
geoknee: In order to pass the schema after the latter changed (merit of which we are discussing in other threads on this page).

packages/xstate-wallet/src/workflows/conclude-channel.ts
geoknee: I propose that we do as you suggest and make this function behave independently of the formatting of the outcomes, say by using `bigNumberify`.

geoknee: (And then remove the comment)

packages/client-api-schema/src/types.ts
geoknee: If you have fewer than 64 nibbles, then you have fewer than 256 bits.

And if you have something other than a multiple of 8 nibbles, then you don't have a valid solidity data type at all.

 Of course most things would likely still work regardless, but if we are going to call something a Uint256, it seems to me that it should have 256 bits.

packages/xstate-wallet/src/workflows/conclude-channel.ts
geoknee: #1342

PR #1338, opened 2020-03-31
Add env variable to clear storage on load

lalexgap:
Introduces a`CLEAR_STORAGE_ON_LOAD` env variable that clears out the indexedDb backend on load.

### DIFF
--- a/packages/xstate-wallet/src/index.ts
+++ b/packages/xstate-wallet/src/index.ts
@@ -5,14 +5,16 @@ import {MessagingService} from './messaging';
 import {ChainWatcher} from './chain';
 import {IndexedDBBackend} from './store/indexedDB-backend';
 import {MemoryBackend} from './store/memory-backend';
-import {TestStore} from './workflows/tests/store';
+import {XstateStore} from './store';
 
 (async function() {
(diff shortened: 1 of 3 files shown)


packages/xstate-wallet/src/index.ts
```diff
@@ -5,14 +5,16 @@ import {MessagingService} from './messaging';
...
 
 (async function() {
   const {privateKey} = ethers.Wallet.createRandom();
   const chain = new ChainWatcher();
+
   const backend = process.env.USE_INDEXED_DB ? new IndexedDBBackend() : new MemoryBackend();
-  const store = new TestStore(chain, backend);
-  await store.initialize([privateKey]);
+  const store = new XstateStore(chain, backend);
+  const cleanSlate = !!process.env.CLEAR_STORAGE_ON_START || false;
```
kerzhner: In recent discussions, we have concluded that it might be a good pattern to never access `process.env` keys outside of `constants.ts`. So in this case, `cleanState` would be defined in and exported by `constants.ts`. 

It also might be good to use [`getEnvBool`](https://github.com/statechannels/monorepo/blob/master/packages/devtools/src/config/env.ts#L4) for boolean environment variables.

PR #1342, opened 2020-04-01
[xstate-wallet] Improve outcomesEqual utility function.

geoknee:
Currently this function does not understand that (e.g.)   '0x02' and '0x2' are equal.

### DIFF
--- a/packages/xstate-wallet/src/store/state-utils.ts
+++ b/packages/xstate-wallet/src/store/state-utils.ts
@@ -107,12 +107,12 @@ export function outcomesEqual(left: Outcome, right?: Outcome) {
     return simpleGuaranteesEqual(left, right);
   }
   if (left.type === 'MixedAllocation' && right?.type === 'MixedAllocation') {
-    if (left.simpleAllocations.length === right.simpleAllocations.length) {
-      return left.simpleAllocations
-        .map((value, index, array) =>
-          simpleAllocationsEqual(left.simpleAllocations[index], right.simpleAllocations[index])
-        )
-        .reduce((a, b) => a && b);
+    return (
+      left.simpleAllocations.length === right.simpleAllocations.length &&
+      _.every(left.simpleAllocations, (_, index) =>
+        simpleAllocationsEqual(left.simpleAllocations[index], right.simpleAllocations[index])
+      )
+    );
     }
   }
   return false;
--- a/packages/xstate-wallet/src/store/state-utils.ts
+++ b/packages/xstate-wallet/src/store/state-utils.ts
@@ -1,4 +1,12 @@
-import {State, ChannelConstants, Outcome, AllocationItem, SignedState} from './types';
+import {
+  State,
+  ChannelConstants,
+  Outcome,
+  AllocationItem,
+  SignedState,
+  SimpleAllocation,
+  SimpleGuarantee
+} from './types';
 import {
   State as NitroState,
   SignedState as NitroSignedState,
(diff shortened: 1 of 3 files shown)


packages/xstate-wallet/src/store/state-utils.ts
```diff
@@ -70,9 +78,46 @@ export function statesEqual(left: State, right: State) {
...
+      return left.allocationItems
+        .map(
+          (value, index, array) =>
+            value.destination === right.allocationItems[index].destination &&
+            value.amount.eq(right.allocationItems[index].amount)
+        )
+        .reduce((a, b) => a && b);
+    }
+  }
+  return false;
```
andrewgordstewart, line 93: I think this is more understandable as a sequence of conditionals joined with `&&`
```suggestion
  return (
    left.assetHolderAddress === right.assetHolderAddress &&
    left.allocationItems.length === right.allocationItems.length &&
    _.every(
      left.allocationItems,
      (value, index) =>
        value.destination === right.allocationItems[index].destination &&
        value.amount.eq(right.allocationItems[index].amount)
    )
  );
```

:caution: ^ Needs a Lodash import, can be replaced with `.map(...).reduce(...)` as above.

packages/xstate-wallet/src/store/state-utils.ts
```diff
@@ -70,9 +78,46 @@ export function statesEqual(left: State, right: State) {
   return hashState(left) === hashState(right);
 }
 
+function simpleAllocationsEqual(left: SimpleAllocation, right: SimpleAllocation) {
+  if (left.assetHolderAddress === right.assetHolderAddress) {
+    if ((left.allocationItems.length = right.allocationItems.length)) {
```
andrewgordstewart, line 83:
```suggestion
    if ((left.allocationItems.length === right.allocationItems.length)) {
```

packages/xstate-wallet/src/store/state-utils.ts
```diff
@@ -70,9 +78,46 @@ export function statesEqual(left: State, right: State) {
...
+  if (left.type === 'SimpleGuarantee' && right?.type === 'SimpleGuarantee') {
+    return simpleGuaranteesEqual(left, right);
+  }
+  if (left.type === 'MixedAllocation' && right?.type === 'MixedAllocation') {
+    if (left.simpleAllocations.length === right.simpleAllocations.length) {
+      return left.simpleAllocations
+        .map((value, index, array) =>
+          simpleAllocationsEqual(left.simpleAllocations[index], right.simpleAllocations[index])
+        )
+        .reduce((a, b) => a && b);
```
andrewgordstewart, line 117:
```suggestion
    return (
      left.simpleAllocations.length === right.simpleAllocations.length &&
      _.every(left.simpleAllocations, (_, index) =>
        simpleAllocationsEqual(left.simpleAllocations[index], right.simpleAllocations[index])
      )
    );
```

packages/xstate-wallet/src/store/tests/outcomes-equal.test.ts
```diff
@@ -0,0 +1,26 @@
+import {outcomesEqual} from '../state-utils';
+import {SimpleAllocation} from '../types';
+import {AddressZero, HashZero} from 'ethers/constants';
+import {bigNumberify} from 'ethers/utils';
+
+const simpleAllocation1 = {
+  type: 'SimpleAllocation',
+  assetHolderAddress: AddressZero,
+  allocationItems: [{destination: HashZero, amount: bigNumberify('0x2')}]
+} as SimpleAllocation;
```
andrewgordstewart, line 10: Can you use a type declaration rather than assertion?

geoknee: Not as far as I know, since `Destination` is tagged: 

https://github.com/statechannels/monorepo/blob/6d3df1a56aadce744c980e502d73acaa30544dc0/packages/xstate-wallet/src/store/types.ts#L33

I suppose the assertion could be made on a finer grain level, at least.

andrewgordstewart: You could also export a `DestinationZero` constant from a test data file which has type `Destination`.

PR #1343, opened 2020-04-01
[web3torrent] set FIREBASE_PREFIX based on an env var

kerzhner:
Unit tests [initialize a firebase client](https://github.com/statechannels/monorepo/blob/master/packages/web3torrent/src/library/web3torrent-lib.ts#L43) that is hooked up to a hosted firebase instance. This ends up potentially costing Google Cloud funds.

There is not a need to hook up unit tests to a local firebase server since the unit tests do not use firebase.


geoknee:
In favour of #1346.

PR #1346, opened 2020-04-01
[web3torrent] unit tests

kerzhner:
Note, this superseeds https://github.com/statechannels/monorepo/pull/1343.

This PR moves away from `react-script` for testing. Current unit tests have the following issues:
- I was not able to set vscode breakpoints in unit tests.
- some tests have typescript errors but somehow still pass (reference `Upload.test.tsx`)

There is an outstanding issue with a couple unit tests where jest shows "Jest did not exit one second after the test run has completed."  This is currently dealt with on master using the [`--forceExit` flag on CI](https://github.com/statechannels/monorepo/blob/master/packages/web3torrent/package.json#L140).


packages/web3torrent/.env.test
```diff
@@ -0,0 +1,3 @@
+TRACKER_URL=http://localhost:4242/announce
+REACT_APP_SINGLE_ASSET_PAYMENT_CONTRACT_ADDRESS = 0x430869383d611bBB1ce7Ca207024E7901bC26b40
+REACT_APP_TARGET_NETWORK = development
```
spersico, line 3:
```suggestion
REACT_APP_SINGLE_ASSET_PAYMENT_CONTRACT_ADDRESS=0x430869383d611bBB1ce7Ca207024E7901bC26b40
REACT_APP_TARGET_NETWORK=development
```

packages/web3torrent/src/App.test.tsx
```diff
@@ -21,6 +21,10 @@ import {web3torrent, WebTorrentContext} from './clients/web3torrent-client';
 // Currently the App.tsx component is arguably responsible for too much: e.g. enabling the web3torrent client and state channel provider. We should figure out a way to test the rendering of App.tsx in isolation from that; possibly by moving the WebTorrentPaidStreamingClient.enable() call elsewhere.
 // https://github.com/statechannels/monorepo/issues/1291
 
+afterAll(() => {
+  web3torrent.destroy();
+});
```
spersico, line 26: Can this be moved to the global teardown?, so that it doesn't need to be repeated?

kerzhner: I think this is an idea worth exploring.

PR #1350, opened 2020-04-02
[web3t] Torrents on welcome screen are *our* pre-seeded torrents

### LINKED ISSUE #1331: [web3t] Remove or modify example torrents
If you load the app and click "Download" next to (e.g.) the Sintel torrent, you start downloading without any state channels stuff getting in your way. 

We could either

a) delete these for now
b) add a check for downloaders to not connect to non web3t clients (uploaders already have one, I think)

geoknee:
Current behaviour: There are 3 torrents on the Welcome page. Clicking on Download will connect you, via a public tracker, to any webtorrent client (bypassing state channels flow). 

With this PR: The Welcome page is now a class that wraps a bittorrent tracker client, connects to *only* our whitelisted tracker(s) (not the public ones!) and will display a torrent if it can find at least one seeder on that tracker. When clicking, you have to do the proper state-channels-enabled flow as you would when sharing a magnet link. 

Closes #1331 .

TODO: 

- [ ] Swap `Sintel.mp4` for some more interesting content: `ForceMove.pdf`? A video presentation someone has made / will make? Suggestions welcome.
- [ ] Figure out if a tracker will drop a seeder if they go offline / close their browser

### DIFF
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -68,7 +68,7 @@ export const EmptyTorrent = ({
 // Pre Seeded Constants (by StateChannels team)
 export const preSeededTorrents: Array<Partial<Torrent>> = [
   {
-    name: 'Sintel',
+    name: 'Sintel.mp4',
     length: 129241752,
     infoHash: 'c53da4fa28aa2edc1faa91861cce38527414d874',
     magnetURI:
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -70,9 +70,7 @@ export const mockTorrents: Array<Partial<Torrent>> = [
   {
     name: 'Sintel',
     length: 129302391,
-    numSeeds: 47,
-    numPeers: 12,
-    files: [],
+    infoHash: 'c53da4fa28aa2edc1faa91861cce38527414d874',
     magnetURI: 'magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&xl=129302391'
   },
   {
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -69,9 +69,10 @@ export const EmptyTorrent = ({
 export const mockTorrents: Array<Partial<Torrent>> = [
   {
     name: 'Sintel',
-    length: 129302391,
+    length: 129241752,
     infoHash: 'c53da4fa28aa2edc1faa91861cce38527414d874',
-    magnetURI: 'magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&xl=129302391'
+    magnetURI:
+      'magnet:?xt=urn%3Abtih%3Ac53da4fa28aa2edc1faa91861cce38527414d874&dn=Sintel.mp4&xl=129241752'
   },
   {
     name: 'Big Buck Bunny',
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -73,8 +73,7 @@ export const mockTorrents: Array<Partial<Torrent>> = [
     numSeeds: 47,
     numPeers: 12,
     files: [],
-    magnetURI:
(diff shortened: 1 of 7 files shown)


packages/web3torrent/src/constants.ts
```diff
@@ -65,37 +65,24 @@ export const EmptyTorrent = ({
...
-    name: 'Sintel',
-    length: 129302391,
-    numSeeds: 47,
-    numPeers: 12,
-    files: [],
+    name: 'Sintel.mp4',
+    length: 129241752,
+    infoHash: 'c53da4fa28aa2edc1faa91861cce38527414d874',
     magnetURI:
-      'magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2F&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent'
```
geoknee, line 77: Here lies the only remains of any public trackers. This is the reason we were still trying to connect to the defunct `tracker.fastcast.nz` despite removing it from our list of trackers.

packages/web3torrent/src/constants.ts
```diff
@@ -65,37 +65,24 @@ export const EmptyTorrent = ({
...
+export const preSeededTorrents: Array<Partial<Torrent>> = [
   {
-    name: 'Sintel',
-    length: 129302391,
-    numSeeds: 47,
-    numPeers: 12,
-    files: [],
+    name: 'Sintel.mp4',
+    length: 129241752,
+    infoHash: 'c53da4fa28aa2edc1faa91861cce38527414d874',
```
geoknee, line 73: Things that confuse me: 
1. The string that follows `btih` in the previous magnet link (which I assume means bit torrent info hash) is rejected by our tracker as an invalid infoHash, but the magnet link still works when you paste into our web3torrent client.
2. Changing the magnet link seems to change whether the paid streaming extension is used, or not.

Can you shed any light @spersico ? 


@spersico

spersico: 1 - What are you testing? And how is it getting rejected?.
According to [this wiki](https://en.wikipedia.org/wiki/Magnet_URI_scheme), the minimum viable magnet format is:
`magnet:?xt=urn:btih:INFOHASH`
The `btih` part is what clients and servers use to figure out what's the format of the hash. . Maybe you are not including that in the magnet that you send to the tracker?

2- Changing the magnet link how? We parse the magnet from the URL in [here](https://github.com/statechannels/monorepo/blob/25aa4b93cd877b897e20f112c8005278d7d6c2cd/packages/web3torrent/src/utils/magnet.ts#L5), and that's all that we do with that.

geoknee: 1. I passed in that *part* of the magnetlink as an info hash in the bittorrent tracker client. In the logs of our tracker, I see an "Invalid infoHash" error (I never send it a magnet link). It doesn't matter so much, since I got it working by regenerating the magnet link and infoHash from our client. Perhaps I am just wrong to try and naively grab the infohash from the magnet link.

2. Using the magnet links that are in the mockConstants on `master`*, we seem to be able to connect to i) public trackers and ii) clients that do not run our paid streaming extension, and therefore do not demand any payment or channel setup. Probably ii is a consequence of i, and the fact the downloaders won't pay uploaders if they aren't prompted to do so! So I think I was wrong to say the paid streaming extension is not used; most likely it is only used by the downloader. Does this sound right to you?

3. The magnet link is different when we upload through our client (compared to the ones that are called `mockTorrents` on master). No big deal, but this tripped me up for a while. 

*(You can try this out on the deployed version of the app by clicking on one of the example torrents).

spersico: 1. I would, have to test what the tracker considers valid torrents. I don't quite know what it expects, currently.
2 is definitely correct. If the seeders don't seem to use the PSE extension, nor STOP and ask for funds, the leecher should run the download as is, without paying nor using anything extra from web3torrent. This was done on purpose, to maintain compatibility with non-web3torrent-torrents.
3 I think, was my first approach to make the url shorter and more readable, in a way. I don't quite remember right now what has been done exactly. 


I think that at the future, it would be a good idea to just drop the magnet format that we currently use, and just pass the infohash, name and size of the torrent as params or queryParams in the URL, instead of fragments in the URL.

packages/web3torrent/src/pages/welcome/Welcome.tsx
```diff
@@ -1,50 +1,94 @@
...
+  constructor(props) {
+    super(props);
+    this.trackerClient = new Client({...requiredOpts, ...optionalOpts});
+    this.updateIfSeederFound = this.updateIfSeederFound.bind(this);
+    this.trackerClient.on('update', this.updateIfSeederFound);
+    this.trackerClient.start();
+  }
+
+  componentWillUnmount() {
+    this.trackerClient.off('update', this.updateIfSeederFound);
```
spersico, line 50: According to [the tracker docs](https://github.com/webtorrent/bittorrent-tracker), it seems that running stop or destroy is a good idea on this case:

> // stop getting peers from the tracker, gracefully leave the swarm
> client.stop()
> 
> // ungracefully leave the swarm (without sending final 'stop' message)
> client.destroy()

geoknee: Good catch bfb14f1

PR #1351, opened 2020-04-02
Set initial leecher balance as torrent size

### LINKED ISSUE #1327: [web3t] Set INITIAL_LEECHER_BALANCE according to file size
Currently, all payment channels are initialised with a constant `payer` balance, meaning that above some file size, the leecher will run out of funds and be unable to finish the download.

Until such a time as we have channel top-ups (and perhaps even then), we should make sure the leecher will have enough funds to pay for the entire file.

spersico:
closes #1327


geoknee:
I guess this will turn anything that was an overpayment error into an out-of-funds error. Hopefully we will never run into those again... if we do, we could set the initial balance to 101% of the torrent cost or something.

spersico:
@geoknee For now, while web3torrent is getting finished, I think it's healthy that in case of an overpay to throw an overpay error. 

Maybe on these days, with the testing this project it's sure to get, with this we find if there's some case that we have an overpayment.

PR #1352, opened 2020-04-02
PaymentChannelClient initialization refactor, Loading Upload UI

### LINKED ISSUE #1291: [web3t] Do not trigger wallet enabling immediately
Currently, as soon as you land on `web3torrent`, you will get our wallet pop up and ask you to connect to metamask. This is considered bad UX:

https://rimble.consensys.design/guides/ux/connect-a-wallet-conditions

If we simply move this command

https://github.com/statechannels/monorepo/blob/efb9da5271b8ee217bf2ab92133b6f6cc9b26457/packages/web3torrent/src/App.tsx#L36

(which enables the web3torrent client, which enables everything else including the provider)

to a later point in the UX, such as after having clicked [ Choose files ] / [ Start Download ], this should be an improvement.

spersico:
closes #1291
**PaymentChannelClient initialization**

This PR moves the WebTorrentPaidStreamingClient and PaymentChannelClient initialization to the execution of the download and upload functions.

**PaymentChannelClient refactor**
Also, I moved part of the  `WebTorrentPaidStreamingClient.enable()` code, to the `PaymentChannelClient.enable()`, as that looks like something specific the the PaymentChannelClient.

**Loading Upload UI**
As this inserts a delay in the upload page, and considering that this delay can also happen when seeding big files, I added a loading spinner to that screen, while the torrent is loading up:
image


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -126,10 +76,10 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     const gotAWire = new Promise(resolve => {
       super.add(testTorrent.magnetURI, (torrent: Torrent) => {
         torrentId = torrent.infoHash;
-        torrent.once('wire', wire => resolve(true));
+        torrent.once('wire', () => resolve(true));
       });
     });
-    const timer = new Promise(function(resolve, reject) {
+    const timer = new Promise(function(resolve, _) {
```
geoknee, line 82:
```suggestion
    const timer = new Promise(function(resolve, () => {}) {
```

geoknee: This was my mistake, apologies.

PR #1353, opened 2020-04-02
Add proper support check, refactor state storage

### LINKED ISSUE #1344: Garbage Collection of Old states in ChannelStoreEntry uses a lot of CPU
Currently most of the cpu usage in chrome is taken by attempting to clean up old states. Specifically this call https://github.com/statechannels/monorepo/blob/d2ab09459ca502ce460122dd3e9d313709d80596/packages/xstate-wallet/src/store/channel-store-entry.ts#L180

We should probably optimize this or remove it (perhaps GC on start up?)

lalexgap:
Changes the storage of states to use a simple array and update the `support` and `supported` functions to perform an actual supported check instead of relying on the amount of signatures.

Running after this change I'm seeing less cpu usage and download speeds of `2-4MB/s`.

Fixes #1344

### DIFF
--- a/packages/xstate-wallet/src/serde/wire-format/example.ts
+++ b/packages/xstate-wallet/src/serde/wire-format/example.ts
@@ -7,20 +7,20 @@ export const wireStateFormat = {
   participants: [
     {
       destination: '0x00000000000000000000000063E3FB11830c01ac7C9C64091c14Bb6CbAaC9Ac7',
-      participantId: '0xAE363d29fc0f6A9bbBbEcC87751e518Cd9CA83C0',
-      signingAddress: '0xAE363d29fc0f6A9bbBbEcC87751e518Cd9CA83C0'
+      participantId: '0x11115FAf6f1BF263e81956F0Cc68aEc8426607cf',
+      signingAddress: '0x11115FAf6f1BF263e81956F0Cc68aEc8426607cf' // key: 0x95942b296854c97024ca3145abef8930bf329501b718c0f66d57dba596ff1318
     },
     {
       destination: '0x00000000000000000000000063E3FB11830c01ac7C9C64091c14Bb6CbAaC9Ac7',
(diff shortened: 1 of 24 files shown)


packages/xstate-wallet/src/serde/wire-format/example.ts
```diff
@@ -7,20 +7,20 @@ export const wireStateFormat = {
   participants: [
     {
       destination: '0x00000000000000000000000063E3FB11830c01ac7C9C64091c14Bb6CbAaC9Ac7',
-      participantId: '0xAE363d29fc0f6A9bbBbEcC87751e518Cd9CA83C0',
-      signingAddress: '0xAE363d29fc0f6A9bbBbEcC87751e518Cd9CA83C0'
+      participantId: '0x11115FAf6f1BF263e81956F0Cc68aEc8426607cf',
+      signingAddress: '0x11115FAf6f1BF263e81956F0Cc68aEc8426607cf' // key: 0x95942b296854c97024ca3145abef8930bf329501b718c0f66d57dba596ff1318
```
andrewgordstewart, line 11: :+1:

packages/xstate-wallet/src/store/types.ts
```diff
@@ -130,12 +134,12 @@ export interface Message {
...
 
 export type ChannelStoredData = {
-  stateVariables: Record<string, StateVariables>;
+  stateVariables: Array<StateVariablesWithHash>;
   channelConstants: Omit<ChannelConstants, 'challengeDuration' | 'channelNonce'> & {
     challengeDuration: BigNumber | string;
     channelNonce: BigNumber | string;
   };
-  signatures: Record<string, Array<string | undefined>>;
+  signatures: Record<string, Array<SignatureEntry>>;
```
andrewgordstewart, line 142: Does it make sense to refactor this to just store the `signatures` on the entry of the `stateVariables` array? (so that they are guaranteed to be aligned)

PR #1355, opened 2020-04-03
web3torrent: initialize firebase only when configuration is defined

kerzhner:


geoknee:
I profiled our firebase database while running the `client-seeding.test.ts`. On `master`, I saw 40 operations recorded (0 bytes written or read to the DB, but there will have been overhead data transfer which we pay for). 

I believe these operations are 

- when we register for events using `myFirebaseRef.on` (disabling this brought the number down to 21).
- `myFirebaseRef.onDisconnect` (disabling brings the number down to 2) 
- creating `myFirebaseRef` and `hubFirebaseRef` in the first place. (disabling those commands brings the number down to 0).

On this branch, I saw 0 operations recorded.

geoknee:
At roughly 4000 circle jobs a month, and at least 40 firebase operations per circle job., this means we run 160K firebase operations per month. And it seems like we download at least 3.5KB per operation as an overhead cost. (https://firebase.google.com/docs/database/usage/billing). 

That’s 560MB downloaded per month, which I think begins to explain our high costs.

geoknee:
The overhead cost far outweighs the amount of _actual_ data downloaded, implying that we shouldn't bother trying to optimise that.

One rule we could try to live by is to not hit firebase at all from our circle jobs. And of course at the very least to follow @kerzhner's example here and reduce operations that don't do anything for us anyway).

kerzhner:
Thanks for the details @geoknee. This is great. There remains a piece of our firebase cost that I do not fully understand. It seems that the overhead of connecting to a firebase instance is proportional to the amount of data stored in firebase.

kerzhner:
On more thought: web3torrent would connect to a valid firebase prefix that contained entries. `on_child` added event fires with initial data when the event listener is added. So the data for the prefix is repeatedly downloaded.

PR #1357, opened 2020-04-03
Application workflow updates

andrewgordstewart:
This journey started off with two goals:
1. Use virtual funding based on the `createChannel` call
    - This required a change to the client-api-schema, which required changes across many files.
2. Get rid of the UI when using `'Virtual'` funding strategy.

This lead to
3. (1) lead me to start the application workflow when the 2nd participant receives the `OpenChannel` message  (from the opponent), rather than the `JoinChannel` event (from the app). This makes sense from a broader point of view: there should be something to handle application-related things (such as remembering what the `fundingStrategy` is!)
  - This resulted in a fairly large refactor of the application workflow, resulting in [this diagram](https://xstate.js.org/viz/?gist=22de47394d05a9e6204be288d43bd119)
  - One nice thing about the refactor is the workflow now accepts a discriminated union as its `Init`, rather than being sent an event right away /shrug/

4. The `confirmCreateChannel` workflow doesn't really have anything to do with creating a channel -- it's just a "confirm" machine, so it is stripped down and renamed `confirm`

5. The application workflow tests (`application.test.ts` as well as those in `integration-tests`, somewhat) were tightly coupled to the state schema of the application workflow. They are still coupled, but a little less so, and I think they are closer to testing the intended behaviour rather than the implementation.

6. After trying to test it in browser, I found various other bugs to fix (near the end of the commits). It just about works, though someone isn't signing a state in the joint channel. I would rather get this merged and tackle the broken virtual-funding workflow in a separate PR.


geoknee:
I didn't look at the code (yet), but the description sounds good to me! 

I like the linked diagam. Things that confuse me a bit: 

1.  On load I seem to get dumped into the failure sub-state of the `joiningChannel` state
2.  I seem to be allowed to transition from that failure state into (eventually) a `fundingChannel` state
3.  Joining happens before creating (was this somehow intentional?)

andrewgordstewart:
> I didn't look at the code (yet), but the description sounds good to me!
> 
> I like the linked diagam. Things that confuse me a bit:
> 
> 1. On load I seem to get dumped into the failure sub-state of the `joiningChannel` state

The guards in the config are mocked to always return true. So, the first guarded transition is always accepted.
> 2. I seem to be allowed to transition from that failure state into (eventually) a `fundingChannel` state

You can indeed transition by manually sending the `'done.joiningChannel'` event. This is an event which the interpreter sends to itself on reaching a `final` state (those surrounded with double lines). In this diagram, `'done'` is the only final child state of `'joiningChannel'`

> 3. Joining happens before creating (was this somehow intentional?)

- Only one of "joining" and "creating" actually happen.
- If you are the creator (`amCreator` guard), you first enter `joiningChannel.joining` and then transition to `joiningChannel.done`, which triggers the `'done.joiningChannel'` event, moving right away to either `'confirmingWithUser'` or `'creatingChannel'`, depending on the funding strategy

geoknee:
Thanks for the explanation. On 3, the approach seems fine to me 👍

packages/xstate-wallet/src/constants.ts
```diff
@@ -39,9 +38,22 @@ export function tokenAddress(assetHolderAddress: string): string | undefined {
   return assetHolderAddress;
 }
 
-export function useVirtualFunding(): boolean {
-  return getEnvBool('USE_VIRTUAL_FUNDING', false);
-}
+export function getBool(val: string | undefined, throwIfMissing = true): boolean {
```
lalexgap, line 41: Could you throw a TODO to eventually use `getEnvBool`

packages/xstate-wallet/src/constants.ts
```diff
@@ -39,9 +38,22 @@ export function tokenAddress(assetHolderAddress: string): string | undefined {
   return assetHolderAddress;
 }
```
lalexgap, line 40:
```suggestion
// TODO: Use getEnvBool from devtools once working
```

packages/xstate-wallet/src/integration-tests/funding.test.ts
```diff
@@ -1,21 +1,19 @@
...
 import {
   Player,
   generateCreateChannelRequest,
   hookUpMessaging,
   generateJoinChannelRequest
 } from './helpers';
+import {isChannelUpdated, isChannelProposed} from '../messaging';
 
-jest.setTimeout(30000);
+jest.setTimeout(20000);
```
lalexgap, line 12: Nit: We probably should switch to a constant timeout

packages/xstate-wallet/src/integration-tests/funding.test.ts
```diff
@@ -1,21 +1,19 @@
...
 
 it('allows for two wallets to fund an app', async () => {
+  /*
```
lalexgap, line 15:
```suggestion
```

packages/xstate-wallet/src/integration-tests/funding.test.ts
```diff
@@ -1,21 +1,19 @@
...
+   */
```
lalexgap, line 16:
```suggestion
```

packages/xstate-wallet/src/messaging.ts
```diff
@@ -103,7 +110,14 @@ export class MessagingService implements MessagingServiceInterface {
   public async sendChannelNotification(
     method: ChannelClosingNotification['method'] | ChannelUpdatedNotification['method'],
     notificationData: ChannelResult
-  ) {
+  );
+  // eslint-disable-next-line no-dupe-class-members
+  public async sendChannelNotification(
```
lalexgap, line 115: Can this be removed?

lalexgap: The overload?

andrewgordstewart: The overload is intentional:
- if you want to respond to `ChannelProposed`, you _must_ provide the `fundingStrategy`
- otherwise, you _must not_ provide the `fundingStrategy`

PR #1363, opened 2020-04-03
Private Tracker Deployment Preparation, Client Capability Testing improvements

spersico:
This PR
- adds support for secure WebSocket connections to the tracker (_wss_).
- moves the Web3TorrentClient network capability testing to the Web3TorrentClient constructor, but only for non-testing environments as this causes problems with tests.
- refactors the app component, and the welcome and upload pages.
- un-skips all the unit tests that were being skipped, and removes non-representative tests.

### DIFF
--- a/packages/web3torrent/src/App.test.tsx
+++ b/packages/web3torrent/src/App.test.tsx
@@ -2,30 +2,11 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
 import {web3torrent, Web3TorrentContext} from './clients/web3torrent-client';
-
-// This tests fails with the following error message
-// /Users/georgeknee/statechannels/monorepo/node_modules/webtorrent/lib/torrent.js:232
-//
-// if (this._rechokeIntervalId.unref) this._rechokeIntervalId.unref()
-// ^
-//
-// TypeError: Cannot read property 'unref' of undefined
-// at Torrent._onParsedTorrent (/Users/georgeknee/statechannels/monorepo/node_modules/webtorrent/lib/torrent.js:232:33)
-// at process.nextTick (/Users/georgeknee/statechannels/monorepo/node_modules/webtorrent/lib/torrent.js:205:14)
-// at process._tickCallback (internal/process/next_tick.js:61:11)
-
-// The issue seems to be related to https://github.com/webtorrent/bittorrent-protocol/issues/44
-
-// The test will pass if we disable the call to WebTorrentPaidStreamingClient.testTorrentingCapability() inside App.tsx
-
-// Currently the App.tsx component is arguably responsible for too much: e.g. enabling the web3torrent client and state channel provider. We should figure out a way to test the rendering of App.tsx in isolation from that; possibly by moving the WebTorrentPaidStreamingClient.enable() call elsewhere.
-// https://github.com/statechannels/monorepo/issues/1291
-
 afterAll(() => {
   web3torrent.destroy();
 });
 
-it.skip('renders app without crashing', () => {
+it('renders app without crashing', () => {
   const div = document.createElement('div');
 
   ReactDOM.render(
--- a/packages/web3torrent/src/App.tsx
+++ b/packages/web3torrent/src/App.tsx
@@ -11,13 +11,15 @@ import File from './pages/file/File';
 import Upload from './pages/upload/Upload';
 import {RoutePath} from './routes';
 import {Web3TorrentContext} from './clients/web3torrent-client';
+import {TorrentClientCapabilities, ClientEvents} from './library/types';
 
 const history = createBrowserHistory();
 class App extends React.Component {
   state = {
     currentNetwork: 'ethereum' in window && Number(window.ethereum.networkVersion),
     requiredNetwork: Number(process.env.REACT_APP_CHAIN_NETWORK_ID),
-    canTorrent: true
+    canTorrent: true,
+    showNetworkWarning: false
   };
 
   static contextType = Web3TorrentContext;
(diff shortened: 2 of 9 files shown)


packages/web3torrent/src/App.test.tsx
```diff
@@ -2,30 +2,11 @@ import React from 'react';
...
-// at Torrent._onParsedTorrent (/Users/georgeknee/statechannels/monorepo/node_modules/webtorrent/lib/torrent.js:232:33)
-// at process.nextTick (/Users/georgeknee/statechannels/monorepo/node_modules/webtorrent/lib/torrent.js:205:14)
-// at process._tickCallback (internal/process/next_tick.js:61:11)
-
-// The issue seems to be related to https://github.com/webtorrent/bittorrent-protocol/issues/44
-
-// The test will pass if we disable the call to WebTorrentPaidStreamingClient.testTorrentingCapability() inside App.tsx
-
-// Currently the App.tsx component is arguably responsible for too much: e.g. enabling the web3torrent client and state channel provider. We should figure out a way to test the rendering of App.tsx in isolation from that; possibly by moving the WebTorrentPaidStreamingClient.enable() call elsewhere.
-// https://github.com/statechannels/monorepo/issues/1291
```
spersico, line 22: Moved to https://github.com/statechannels/monorepo/issues/1369

packages/web3torrent/src/constants.ts
```diff
@@ -78,6 +73,15 @@ export const preSeededTorrents: Array<Partial<Torrent>> = [
   }
 ];
 
+// Welcome Page Tracker creation options
+export const welcomePageTrackerOpts = {
+  infoHash: [preSeededTorrents[0].infoHash],
+  announce: defaultTrackers,
+  peerId: '2d5757303030372d37454e613073307937495630', // random
```
kerzhner, line 80: Should this be a randomly generated value instead of a hardcoded string? If there are many users using the Welcome page, will shared `peerId`s cause issues?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -44,12 +45,15 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
 
+  clientCapability: TorrentClientCapabilities = TorrentClientCapabilities.NOT_TESTED;
+
   constructor(opts: WebTorrent.Options & Partial<PaidStreamingExtensionOptions> = {}) {
     super(opts);
     this.peersList = {};
     this.pseAccount = opts.pseAccount;
     this.paymentChannelClient = opts.paymentChannelClient;
     this.outcomeAddress = opts.outcomeAddress;
+    if (process.env.NODE_ENV !== 'test') this.testTorrentingCapability();
```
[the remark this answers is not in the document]
kerzhner: What is the benefit of adding this to the constructor? `testTorrentCapability` has side effects, which is not ideal for constructor code.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -83,9 +88,13 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     });
     const raceResult = await Promise.race([gotAWire, timer]);
     if (torrentId) {
-      this.remove(torrentId);
+      super.remove(torrentId);
```
kerzhner, line 91: If `resolve(torrentId)` replaces `resolve(true)`, we can avoid having to define the `torrentId` variable.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -69,7 +73,8 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     this.outcomeAddress = null;
   }
 
-  async testTorrentingCapability(timeOut: number) {
+  async testTorrentingCapability(timeOut: number = 3000): Promise<void> {
+    this.emit(ClientEvents.CLIENT_CAPABILITY_TEST, this.clientCapability);
     log('Testing torrenting capability...');
     let torrentId;
```
kerzhner, line 79:
```suggestion
```

packages/web3torrent/src/App.tsx
```diff
@@ -11,37 +11,55 @@ import File from './pages/file/File';
...
 import {Web3TorrentContext} from './clients/web3torrent-client';
+import {TorrentClientCapabilities, ClientEvents} from './library/types';
 
 const history = createBrowserHistory();
 class App extends React.Component {
   state = {
     currentNetwork: 'ethereum' in window && Number(window.ethereum.networkVersion),
     requiredNetwork: Number(process.env.REACT_APP_CHAIN_NETWORK_ID),
-    canTorrent: true
+    canTorrent: true,
```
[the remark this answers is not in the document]
spersico: It's used for the _Looks like you do not have an internet connection or are behind a firewall.... warning_

packages/web3torrent/src/constants.ts
spersico: Haven't tested that out, I _think_ it shouldn't make any problems, considering that this code doesn't actually download anything or makes any updates on the tracker

spersico: I could make a random number every time though, so it doesn't become an issue

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: Thing is, and correct me if I'm wrong, Promise.race only returns the first promise that gets solved, and drops the other. If the torrent is created but no peers connect on time (meaning, if the test fails), we won't have the torrentId.

spersico:
> What is the benefit of adding this to the constructor? testTorrentCapability has side effects, which is not ideal for constructor code.

The idea was to start that function as fast as possible, in order to show as soon as possible the warning. I could run that function on the initialize though.

packages/web3torrent/src/App.tsx
```diff
@@ -11,37 +11,55 @@ import File from './pages/file/File';
...
+    showNetworkWarning: false
   };
 
   static contextType = Web3TorrentContext;
 
   // Adds typing information to this.context
   context!: React.ContextType<typeof Web3TorrentContext>;
 
-  async componentDidMount() {
+  constructor(props) {
```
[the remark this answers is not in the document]
spersico: it seems that there's quite a bit of overlap between this PR and yours, so I'll let you finish up yours first, and work from there

packages/web3torrent/src/library/web3torrent-lib.ts
kerzhner: How about something like that?
```
    const gotAWire = new Promise<string>(resolve => {
      super.add(testTorrent.magnetURI, (torrent: Torrent) => {
        torrent.once('wire', () => resolve(torrent.infoHash));
      });
    });
    const timer = new Promise<void>(function(resolve, _) {
      setTimeout(resolve, timeOut);
    });
    const torrentId = await Promise.race([gotAWire, timer]);
    if (torrentId) {
      super.remove(torrentId);
    }
    this.clientCapability = torrentId
      ? TorrentClientCapabilities.CAPABLE
      : TorrentClientCapabilities.NOT_CAPABLE;
```

packages/web3torrent/src/App.tsx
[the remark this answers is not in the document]
kerzhner: Filed https://github.com/statechannels/monorepo/issues/1412. My 2c is that we should finish up this PR and refactor `App.tsx` later.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: but... doesn't that has the same problem? If timer resolves first, we lose the Promise result of `gotAWire`, and then we can't remove the torrent from the client

kerzhner: Resolved on a call, thanks @spersico!

PR #1364, opened 2020-04-04
(Mostly) budget-related fixes

andrewgordstewart:
With these changes, I was able to virtually fund a web3torrent channel:

```
$ state-channels-monorepo
❯ yarn start:shared-ganache

$ xstate-wallet
❯ CLEAR_STORAGE_ON_START=true USE_INDEXED_DB=true  ADD_LOGS=true yarn start

$ simple-hub
❯ FIREBASE_PREFIX=web3t-andrew  NODE_ENV=development yarn hub:watch

$ web3-torrent
❯ REACT_APP_FIREBASE_PREFIX=web3t-andrew REACT_APP_AUTO_FUND_LEDGER=true yarn start
```

TODO:
`CLEAR_STORAGE_ON_START=true` and `REACT_APP_AUTO_FUND_LEDGER=true` seem to be necessary because on reload, the chain does not fetch the data from the adjudicator's current state.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,7 +233,7 @@ jobs:
       - image: circleci/node:10.16.3
     environment:
       FIREBASE_PREFIX: << pipeline.git.branch >>-1
-      USE_VIRTUAL_FUNDING: 'FALSE'
+      REACT_APP_FUNDING_STRATEGY: Direct
       USE_GANACHE_DEPLOYMENT_CACHE: true
       GANACHE_CACHE_FOLDER: ../../.ganache-deployments
       GANACHE_PORT: 8547
--- a/packages/xstate-wallet/src/workflows/application.ts
+++ b/packages/xstate-wallet/src/workflows/application.ts
@@ -62,6 +62,7 @@ type WorkflowGuards = Guards<
 >;
 
 export interface WorkflowActions {
+  setApplicationSite: Action<ChannelIdExists, JoinChannelEvent>;
   sendUpdateChannelResponse: Action<any, PlayerStateUpdate>;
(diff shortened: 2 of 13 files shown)


andrewgordstewart:
Note that this does not work when rebased on master: see https://github.com/statechannels/monorepo/issues/1366#issue-594004137

packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -305,28 +311,30 @@ export const workflow = (
...
+        applicationSite
       );
 
       // Create a open channel objective so we can coordinate with all participants
       await store.addObjective({
         type: 'OpenChannel',
-        data: {targetChannelId: channelId, fundingStrategy},
-        participants
+        data: {targetChannelId, fundingStrategy},
+        participants: [participants[1]]
```
lalexgap, line 335: Why is this just one participant?

andrewgordstewart: Because if I am creating a channel, I don't want to also run the `OpenChannel` objective (which triggers a second application workflow)

.circleci/config.yml
```diff
@@ -233,7 +233,7 @@ jobs:
       - image: circleci/node:10.16.3
     environment:
       FIREBASE_PREFIX: << pipeline.git.branch >>-1
-      USE_VIRTUAL_FUNDING: 'FALSE'
+      REACT_APP_FUNDING_STRATEGY: Direct
```
junjizhi, line 236: Can you add this var to `integration-test-web3torrent` config? That would make the CI pass. I tried it on master.

PR #1397, opened 2020-04-08
Store signs states at most once

andrewgordstewart:


andrewgordstewart:
(This has has newer changes from `master`, so I've marked it as a draft until @junjizhi 's branch is merged)

PR #1400, opened 2020-04-09
Add prefundsetup check to ledger funding

andrewgordstewart:
While working on virtual funding puppeteer tests with @snario, we found the reason why the simple-hub was not able to read messages: during serialization, guarantee outcomes are stripped: https://github.com/statechannels/monorepo/commit/7c6aeb79662cb5bc759b1b9ef78b12efbccf2b5b.

This lead me to realize that the virtual-funding workflows create the guarantor channels, then move to the next step before that channel has a supported "prefund setup" state.

This PR adds a defensive "checking target" state for validating the state of the target channel when ledger funding. (The virtual funding bug will be fixed in a later PR.)


andrewgordstewart:
> (The virtual funding bug will be fixed in a later PR.)

Actually, I realized that it needs to be fixed in _this_ PR to make the tests pass 😄

PR #1414, opened 2020-04-10
Add concurrency safety to store + various fixes

snario:
- Add a private `withLock` function to the store. Functions which call `this.backend.setChannel` should be wrapped with a `withLock` call to guarantee safety
- The `simple-hub` logs the exact string that it sends, as to not mislead developers when debugging
- The `simple-hub` serializes guarantee outcomes
- The `web3torrent-integration-test-with-hub` sets the same firebase prefix between `web3torrent` and `simple-hub`

TODO: 
- [ ] Devise a scheme to prevent calling `this.backend.setChannel` without first acquiring a log.
- [ ] Clean up the environment variable mess that we've created.
- [ ] Add explicit tests for multiple concurrent writes to channel entries?


andrewgordstewart:
It would be nice to merge this as is, the benefits being:
1. we can hopefully sustain ✅ 's on `master`
2. we can resolve some possible unpredictable behaviour

andrewgordstewart:
> Devise a scheme to prevent calling this.backend.setChannel without first acquiring a log.

Locally, I was playing around with changing the type of `this.backend.setChannel` to be
`(key: string, data: () => Promise<ChannelStoredData>)`.

Then, the backend is responsible for concurrency-safety.

packages/simple-hub/src/wallet/xstate-wallet-internals/index.ts
```diff
@@ -171,18 +173,24 @@ function deserializeAllocationItem(allocationItem: AllocationItemWire): Allocati
   return {destination, amount: bigNumberify(amount)};
 }
 
-function serializeOutcome(outcome: Outcome): OutcomeWire {
+export function serializeOutcome(outcome: Outcome): OutcomeWire {
```
kerzhner, line 176: Does this need to be exported?

packages/simple-hub/src/wallet/xstate-wallet-internals/index.ts
```diff
@@ -309,3 +317,13 @@ export const firstState = (
   participants,
   outcome
 });
+
+export function makeDestination(addressOrDestination: string): string {
```
kerzhner, line 321: Should [`deposit.ts`](https://github.com/statechannels/monorepo/blob/ags/wallet-fixes/packages/simple-hub/src/wallet/deposit.ts#L35) be updated to use this function?

snario:
@andrewgordstewart It looks like the test failed, but the job succeeded

https://app.circleci.com/pipelines/github/statechannels/monorepo/4987/workflows/d19ec482-f2b4-443d-8e11-262db7e86dd6/jobs/18094

```
Killed
error Command failed with exit code 137.
```

andrewgordstewart:
Xstate wallet did start. Did you read the logs in the puppeteer test?

PR #1415, opened 2020-04-10
web3torrent: refactor App component as a function component

kerzhner:
The main benefit of the refactor is that it flagged most state as unneeded. `requiredNetwork `, `canTorrent` and `showNetworkWarning` states keys are removed.

My preference is to use React function component going forward when possible.

### DIFF
--- a/packages/web3torrent/src/App.tsx
+++ b/packages/web3torrent/src/App.tsx
@@ -1,7 +1,7 @@
 import ConnectionBanner from '@rimble/connection-banner';
 import {Flash} from 'rimble-ui';
 import {createBrowserHistory} from 'history';
-import React from 'react';
+import React, {useState, useContext, useEffect} from 'react';
 import {Route, Router, Switch} from 'react-router-dom';
 import './App.scss';
 import {LayoutFooter, LayoutHeader} from './components/layout';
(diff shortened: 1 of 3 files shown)


packages/web3torrent/src/App.tsx
```diff
@@ -11,88 +11,59 @@ import File from './pages/file/File';
...
-      this.setState({...this.state, currentNetwork: Number(window.ethereum.networkVersion)});
-      window.ethereum.on('networkChanged', this.updateState);
+  useEffect(() => {
+    if (window.ethereum) {
+      window.ethereum.on('networkChanged', chainId => {
+        setCurrentNetwork(Number(chainId));
+      });
     }
-  }
+  }, []);
```
spersico, line 31: Is the approach of adding the window.ethereum.networkVersion to the array of dependencies of the useEffect something you tried?

I think that using that would remove the necessity of maintaining a listener

kerzhner: Not sure I fully understand unfortunately. Could you come up with a (pseudo)code of what you are thinking about?

spersico: Something like
```
useEffect(() => {
    setCurrentNetwork(window.ethereum && Number(window.ethereum.networkVersion));
  }, [window.ethereum, window.ethereum.networkVersion]);
```
This way, React is the one checking for changes inside the dependencies, and executing the function specified inside the useEffect, **and** we don't have to mantain the listener.

Though I don't know if it would actually work, I think I remember a case where a prop was changing in the web3torrent library and the change wasn't being picked up by React

kerzhner: Thanks for the clarification, and that's a clever thought. But I think the above doesn't quite work. `useEffect` hook runs after every component render. So `window.ethereum` would only be checked after the `App` component renders. And what if `window.ethereum` changes in the background without triggering `App` component re-rendering?

PR #1418, opened 2020-04-11
[WIP] Fixes RPS firebase auth error

Alonski:
fix?: Don't load RPS LoginSaga
Firebase Auth was removed from RPS

Caused by: https://github.com/statechannels/monorepo/pull/1396


geoknee:
@Alonski does this fix the problem? If so would be good to finish up this PR and merge it. I want to reenable e2e tests with rps + dappeteer and this is blocking me somewhat.

Alonski:
@geoknee This doesn't completely fix RPS. If you check out this branch it'll fail somewhere else.

geoknee:
I'm trying to push this forward in #1467

PR #1427, opened 2020-04-13
Improve Torrent URLs

spersico:
Prior to this PR, the URL to the file page (where peers interact) was:
`HOST/file/#magnet:?xt=urn%3Abtih%3A3839624a79988b6c2246aff5a6cef287dfc98a16&dn=file_name.pdf&xl=14200&cost=0`

Now, the format is:
`HOST/file/3839624a79988b6c2246aff5a6cef287dfc98a16?name=file_name.pdf&length=14200`

Seems small, but when reloading the page, using segments in the URL could be problematic (for example, changing the URL by hand and pressing `ENTER` had to be pressed twice to reload the page). This way, the URL is much more readable (by using queryParams with names that make sense, and at the cost of dropping support for instant.io URL magnet link parsing), and page reloads are much more consistent.


packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -31,21 +31,27 @@ interface Props {
...
   const [, setPeers] = useState({});
   const [loading, setLoading] = useState(false);
   const [buttonLabel, setButtonLabel] = useState('Start Download');
   const [errorLabel, setErrorLabel] = useState('');
 
   const getLiveData = getTorrentAndPeersData(web3Torrent, setTorrent, setPeers);
 
+  useEffect(() => {
+    setTorrent(parseURL(infoHash, queryParams));
+    // eslint-disable-next-line
```
kerzhner, line 46: Are you sure this is needed? I am not seeing any eslint errors or warnings when I run ` yarn eslint:check` in `web3torrent`

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -31,21 +31,27 @@ interface Props {
...
+  }, []);
+
   useEffect(() => {
     if (torrent.infoHash) {
       getLiveData(torrent);
     }
     // eslint-disable-next-line
```
kerzhner, line 53: Same comment about linting as above.

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -31,21 +31,27 @@ interface Props {
...
+  const queryParams = useQuery();
+  const [torrent, setTorrent] = useState<Torrent>(EmptyTorrent);
   const [, setPeers] = useState({});
   const [loading, setLoading] = useState(false);
   const [buttonLabel, setButtonLabel] = useState('Start Download');
   const [errorLabel, setErrorLabel] = useState('');
 
   const getLiveData = getTorrentAndPeersData(web3Torrent, setTorrent, setPeers);
 
+  useEffect(() => {
```
kerzhner, line 44: Do we need to `useEffect` here? As written, this will run only on first render and will initialize `torrent` to `parseURL(infoHash, queryParams)`.

packages/web3torrent/src/utils/magnet.ts
```diff
@@ -1,39 +1,34 @@
...
-export const parseMagnetURL: (rawMagnetURL: string) => Torrent = (rawMagnetURL = '') => {
-  const emptyTorrentData = {...EmptyTorrent, magnetURI: ''} as Torrent;
-  if (!rawMagnetURL.trim()) {
-    return emptyTorrentData;
-  }
-  const magnetParams = new URLSearchParams(rawMagnetURL.trim().replace(/#magnet:/g, ''));
-  if (!magnetParams.has('xt')) {
-    return emptyTorrentData;
-  }
+export const useQuery = () => {
```
kerzhner, line 6: I found this a bit confusing at first. Do you think it would be useful to document that this is the recommended approach in the official docs: https://reacttraining.com/react-router/web/example/query-parameters

packages/web3torrent/src/utils/magnet.ts
```diff
@@ -1,39 +1,34 @@
...
+  return new URLSearchParams(useLocation().search);
+};
 
-  if (!magnetParams.has('tr')) {
-    defaultTrackers.map(tr => magnetParams.append('tr', tr));
+export const parseURL: (infoHash: string, queryParams: any) => Torrent = (
+  infoHash,
+  queryParams
+) => {
```
kerzhner, line 13:
```suggestion
export function parseURL(infoHash: string, queryParams: URLSearchParams): Torrent {
```

kerzhner: This syntax is cleaner. And we can add type information for `queryParams`.

packages/web3torrent/src/utils/magnet.ts
```diff
@@ -1,39 +1,34 @@
...
+  if (!infoHash || !infoHash.trim()) {
+    return EmptyTorrent;
   }
+  const name = queryParams.get('name') || queryParams.get('dn') || 'unknown';
+  const length = Number(queryParams.get('length')) || 0;
```
kerzhner, line 18: `||` works here, but, in general, [nullish coalescing operator (??)](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing) is safer.

packages/web3torrent/src/utils/magnet.ts
```diff
@@ -1,39 +1,34 @@
...
 
-  return {
-    ...emptyTorrentData,
-    name: magnetParams.get('name') || magnetParams.get('dn') || 'unknown',
-    magnetURI: decodeURIComponent(`magnet:?${magnetParams.toString()}`),
-    infoHash: (magnetParams.get('xt') as string).substring(9),
-    length: Number(magnetParams.get('xl')) || 0,
-    cost: magnetParams.get('cost') || '0'
-  };
+const buildMagnetURI = (name = 'unknown', length = 0, infoHash, trackers) => {
```
kerzhner, line 23:
```suggestion
export function buildMagnetURI(
  name: string = 'unknown',
  length: number = 0,
  infoHash: string,
  trackers: string[]
) {```

File: packages/web3torrent/src/utils/magnet.ts [hunk 8]
```diff
@@ -1,39 +1,34 @@
...
+  const parameters = new URLSearchParams(`?xt=urn:btih:${infoHash}&dn=${name}&xl=${length}`);
+  trackers.map(tr => parameters.append('tr', tr));
+  return decodeURIComponent(`magnet:?${parameters.toString()}`);
 };
 
-export const generateMagnetURL = (torrent: Torrent) => {
-  if (!torrent.magnetURI) {
+export const generateURL = ({magnetURI, infoHash, files, length, name}: Torrent) => {
```
kerzhner, line 29: I recommend adding type information and refactoring to use the `function`syntax as suggested for the function above

packages/web3torrent/src/utils/magnet.ts
```diff
@@ -1,39 +1,34 @@
...
-    name: magnetParams.get('name') || magnetParams.get('dn') || 'unknown',
-    magnetURI: decodeURIComponent(`magnet:?${magnetParams.toString()}`),
-    infoHash: (magnetParams.get('xt') as string).substring(9),
-    length: Number(magnetParams.get('xl')) || 0,
-    cost: magnetParams.get('cost') || '0'
-  };
+const buildMagnetURI = (name = 'unknown', length = 0, infoHash, trackers) => {
+  const parameters = new URLSearchParams(`?xt=urn:btih:${infoHash}&dn=${name}&xl=${length}`);
+  trackers.map(tr => parameters.append('tr', tr));
+  return decodeURIComponent(`magnet:?${parameters.toString()}`);
```
kerzhner, line 26: Not sure I fully understand this. Is there a reason to use `URLSearchParams` object and then immediately convert the object to a string? Would using a dictionary work for the `parameters` variable?

packages/web3torrent/src/utils/magnet.ts
kerzhner: Also, `files` parameter is unused.

packages/web3torrent/src/utils/magnet.ts
```diff
@@ -1,39 +1,34 @@
...
 };
 
-export const generateMagnetURL = (torrent: Torrent) => {
-  if (!torrent.magnetURI) {
+export const generateURL = ({magnetURI, infoHash, files, length, name}: Torrent) => {
+  if (!magnetURI) {
```
kerzhner, line 30: Is this a legitimate use of the function? Should we throw an error?

packages/web3torrent/src/pages/file/File.tsx
spersico: In my other PR I [had to add it](https://github.com/statechannels/monorepo/pull/1428/commits/ea8bae6091fd8e8c67c29ef726b4568063a418c2) or it [wouldn't pass the CI tests](https://circleci.com/gh/statechannels/monorepo/18289)

packages/web3torrent/src/utils/magnet.ts
spersico: I like the nullish coalescing (though I need to get used to it, I sometimes forget that I can use it).
In this case,  `||` fits better because the queryParam can be an empty string (e.g. `?name=&length=14200`) and this catches that case too.

packages/web3torrent/src/pages/file/File.tsx
spersico: Answered above

packages/web3torrent/src/utils/magnet.ts
spersico: Changed to a reduce function, that seems to work just as fine, and it's cleaner and shorter.

packages/web3torrent/src/utils/magnet.ts
spersico: I don't quite remember why that was there. I checked this out, and it happens when the upload function tries to create an empty Torrent (seen only on a test, so I don't know how realistic is this).

I will remove it completely, as I wasn't able to create an empty torrent.

packages/web3torrent/src/utils/magnet.ts
spersico: I agree with everything, except adding type information, as the type information is already provided by the Torrent type.

packages/web3torrent/src/pages/file/File.tsx
spersico: It was the result of me discovering [this](https://statechannels.slack.com/archives/CMPHW40R0/p1586807045141000).
Said that, I will change that with
```const [torrent, setTorrent] = useState<Torrent>(() => parseURL(infoHash, queryParams));```

Which avoids the re-running of the function on each render, and it's cleaner IMO.

Comment (kerzhner) [reply packages/web3torrent/src/pages/file/File.tsx]: is `parseURL` an expensive call? Also, not sure the above will work. It will set the `torrent` variable to a parameterless function (and not invoke the function). Unless my understanding of `useState` is flawed...

Comment (kerzhner) [reply packages/web3torrent/src/pages/file/File.tsx]: [Might have answered my own question](https://reactjs.org/docs/hooks-reference.html#functional-updates)

Comment (spersico) [reply packages/web3torrent/src/pages/file/File.tsx]: It works fine on my PC, and the tests pass. And according to the [god-awful docs](https://en.reactjs.org/docs/hooks-reference.html#lazy-initial-state) (that I don't know why they put the `useState` documentation in 2 different places, I didn't like this entire quirk of the useState API), it _should_ work, I think.

File: packages/web3torrent/src/pages/file/File.tsx (hunk 1 above)
Comment (spersico) [reply packages/web3torrent/src/pages/file/File.tsx]: resolved with https://github.com/statechannels/monorepo/pull/1427/commits/edc85738a5512fedcc5f8a36419665afe95884d7 and https://github.com/statechannels/monorepo/pull/1427/commits/7e198de4d82917d8a8e2403ec0b1532e6876abcd

Comment (kerzhner) [reply packages/web3torrent/src/pages/file/File.tsx]: Thanks! I _think_ you also need to add the following rules for eslint to flag issues correctly:
```
'react-hooks/rules-of-hooks': 'error'
'react-hooks/exhaustive-deps': 'warn'
```

You can add the rules [here](https://github.com/statechannels/monorepo/blob/sp/w3t/magnet/packages/web3torrent/.eslintrc.js#L54).

packages/web3torrent/src/utils/magnet.ts
kerzhner: If `queryParams.get('length')` is an empty string, I _think_ `??` would still work correctly since `Number('')` returns 0.

packages/web3torrent/src/pages/file/File.tsx
kerzhner: Also, I think you might need to add `eslint-plugin-react-hooks` to dev dependencies and rerun `yarn`.

spersico: done! :smile: 
Take a look at https://github.com/statechannels/monorepo/pull/1427/commits/9648f9ab425e3bb310ed0ec4b78d4422ae50b678  and https://github.com/statechannels/monorepo/pull/1427/commits/f02939144a5a1d73572bb9a066aab54fcd2a7f39

kerzhner: My comments are:
-  the new rules do not belong in the `otherViolations` list as in this case we are turning on eslint rules as opposed to turning off rules that our codebase violates.
- we have adopted the heuristic that package versions are exact. So remove `^` in the package version.

PR #1428, opened 2020-04-13
Fix multi-file Torrents naming

spersico:
With Torrents that consisted of several files, the naming logic in WebTorrent was not working (at least on Linux), the name of the created torrent was simply one of the files selected, which is kind of misleading for the leecher.

This PR generates a name for the torrent if the torrent has multiple files. The name consists of `"various_"+ the last 5 digits of the Date.now() + ".zip"`.

Added to that, this also sets that name to the zip file when downloaded.
image

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -104,17 +104,13 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     callback?: TorrentCallback
   ): PaidStreamingTorrent {
     this.ensureEnabled();
-    let torrent: PaidStreamingTorrent;
 
-    if (typeof optionsOrCallback === 'function') {
-      torrent = super.seed(
-        input,
-        {createdBy: this.pseAccount, announce: defaultTrackers} as TorrentOptions,
-        optionsOrCallback
-      ) as PaidStreamingTorrent;
-    } else {
-      torrent = super.seed(input, optionsOrCallback, callback) as PaidStreamingTorrent;
-    }
+    const options =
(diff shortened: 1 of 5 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -104,17 +104,13 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-      ) as PaidStreamingTorrent;
-    } else {
-      torrent = super.seed(input, optionsOrCallback, callback) as PaidStreamingTorrent;
-    }
+    const options =
+      typeof optionsOrCallback === 'function'
+        ? {createdBy: this.pseAccount, announce: defaultTrackers}
+        : {createdBy: this.pseAccount, announce: defaultTrackers, ...optionsOrCallback};
+
+    const torrent = super.seed(input, options, callback) as PaidStreamingTorrent;
```
geoknee, line 113:
```suggestion
    const torrent: PaidStreamingTorrent = this.seed(input, options, callback);
```

geoknee: Ok this was a bad suggestion ^ so **don't commit it**.  

I wanted to avoid using the type assertions. I'll make an issue for this.

spersico:
> If I understand correctly, the timing of the user's click will now influence the torrent infoHash, meaning that seeding the same collection of files will likely result in different magnet links (for example). That makes it almost impossible to have multiple seeders for the same collection of files.

This change would only affect the torrent's name, and the filename of the zip file that gets downloaded at the end. It doesn't affect the infohash at all.

Though, you present a good idea, we could use the infohash as the filename, and be done with it!

PR #1429, opened 2020-04-14
Fix web3torrent integration test artifacts upload

junjizhi:
Fix the CI configs that are related to artifacts upload

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -252,19 +252,27 @@ jobs:
       - checkout
       - puppeteer/install
       - log_stats:
-          file: integration-test-stats
+          file: integration-test-web3torrent-stats
       - attach_workspace:
           at: /home/circleci/project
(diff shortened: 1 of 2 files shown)


geoknee:
@junjizhi might be nice to see more output in `e2e.log`:

```
yarn run v1.17.3
$ wait-on http://localhost:3000 http://localhost:3055 && jest puppeteer/__tests__/web3torrent/
Done in 80.18s.
```

On the other hand, the circle console shows much more information: does that make `e2e.log`  (and/or the other log files) somewhat redundant ?

junjizhi:
@geoknee 

> On the other hand, the circle console shows much more information: does that make e2e.log

Hmm. Is it because those extra output are stderr? Wonder if we should pipe them to `tee` there as well: https://stackoverflow.com/a/692407/8175889

PR #1430, opened 2020-04-14
More robust integration test

andrewgordstewart:
- Add a `waitForNthState` e2e helper, and use in the test. This ensures we are testing that some data is actually being transferred
- Fixes an issue in web3torrent where the `channelCache` may be undefined when it's used
- Adds a hack to `updateChannelCache` to always set the channel cache, even when the channel is unknown (see https://github.com/statechannels/monorepo/issues/1438#issuecomment-614159943)

### DIFF
--- a/packages/xstate-wallet/src/workflows/conclude-channel.ts
+++ b/packages/xstate-wallet/src/workflows/conclude-channel.ts
@@ -1,4 +1,4 @@
-import {Machine, StateNodeConfig, ActionFunction} from 'xstate';
+import {Machine, StateNodeConfig} from 'xstate';
 import {Store} from '../store';
 import {SupportState, VirtualDefundingAsLeaf} from '.';
 import {getDataAndInvoke} from '../utils';
@@ -7,7 +7,6 @@ import {outcomesEqual} from '../store/state-utils';
 import {State} from '../store/types';
 import {map} from 'rxjs/operators';
 import {ParticipantIdx} from './virtual-funding-as-leaf';
-import {ETH_ASSET_HOLDER_ADDRESS} from '../constants';
 
 const WORKFLOW = 'conclude-channel';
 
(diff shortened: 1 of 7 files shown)


andrewgordstewart:
I expect the integration test with hub to fail due to what is _probably_ a virtual-defunding bug.

andrewgordstewart:
(Ignore this close -- testing a github projects feature)

lalexgap:
> I expect the integration test with hub to fail due to what is _probably_ a virtual-defunding bug.

Do we have a issue for this?

> If we take the opinion that the integration test should _pass_, then perhaps we should relax the test requirements on a short-term basis.

I'd prefer we have a test that passes for now. Even if the test is not super robust it can still be helpful at catching possible issues when creating a PR (and I really like seeing all  ✅s )

andrewgordstewart:
> > I expect the integration test with hub to fail due to what is _probably_ a virtual-defunding bug.
> 
> Do we have a issue for this?

https://github.com/statechannels/monorepo/issues/1442
 
> > If we take the opinion that the integration test should _pass_, then perhaps we should relax the test requirements on a short-term basis.
> 
> I'd prefer we have a test that passes for now. Even if the test is not super robust it can still be helpful at catching possible issues when creating a PR (and I really like seeing all ✅s )

:+1:

packages/xstate-wallet/src/workflows/conclude-channel.ts
```diff
@@ -69,7 +68,6 @@ const getRole = (store: Store) => (ctx: Init) => async cb => {
 
 const virtualDefunding = {
   initial: 'gettingRole',
-  entry: ['releaseFunds'],
```
andrewgordstewart, line 72: We'll need to release funds appropriately in the virtual-defunding workflow.

PR #1431, opened 2020-04-14
Handle User declining budget

### LINKED ISSUE #1380: Handle cases where user declines
Most of the UI dialog ask the user to confirm an action (for example during approve and fund we ask the user to the approve the budget)

We need to handle the case where the user selects no/decline. 

Areas where this might apply:
- Approve Budget and Fund UI
- Confirm Channel UI
- Withdraw UI

lalexgap:
Fixes #1380.

Handles the case where the user declines a budget in web3-torrent. Just shows a simple warning

image


packages/client-api-schema/src/types.ts
```diff
@@ -46,7 +46,11 @@ interface JsonRpcNotification<NotificationName, NotificationParams> {
   params: NotificationParams;
 }
 
-interface JsonRpcError<Code, Message, Data = undefined> {
+export const UserDeclinedErrorCode = 200;
```
lalexgap, line 49: I ran into `TS1209: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.` in `web3-torrent` when trying to use an enum here, so I just used simple `const` variables.

lalexgap:
This is small enough that it's included in #1462

PR #1436, opened 2020-04-15
[web3torrent] add react eslint plugin

kerzhner:


packages/web3torrent/.eslintrc.js
```diff
@@ -40,13 +58,14 @@ module.exports = {
     browser: true,
     es6: true
   },
-  plugins: [...baseConfig.plugins, "react-hooks", 'jest'],
+  plugins: [...baseConfig.plugins, 'jest', 'react', 'react-hooks'],
```
geoknee, line 61: Nit: I don't think you need this because of the use of `'plugin:react/recommended'` below.

```suggestion
  plugins: [...baseConfig.plugins, 'jest', 'react-hooks'],
```

kerzhner: Good catch!

kerzhner:
> LGTM. Out of interest, was there a particular rule (or rules) that you think will be particularly helpful to us?

To be honest, here was the sequence of events:
1. We found `react-hooks` eslint plugin. It flagged real issues with our code that would have been difficult to debug.
1. I added `react` eslint plugin for posterity.

PR #1434, opened 2020-04-15
Fix integration tests and update direct funding UI

junjizhi:
This PR updates direct funding PR and also has a fix for the failing integration tests. The torrent upload URL has changed.

Fixes https://github.com/statechannels/monorepo/issues/1383

PR #1437, opened 2020-04-15
Change web3torrent e2e test URL

snario:


snario:
Nevermind this is in #1434

PR #1441, opened 2020-04-15
Add "stress" e2e test

andrewgordstewart:
This adds a test file setting up
- one file
- one seeders
- three leechers

There is a more complex, slightly stale test that is skipped. It should be updated to ensure that each actor signs with its own private key

Some other changes were made along the way trying to troubleshoot these tests:
- mostly adding logs
- the simple-hub does not crash on failed deposits


andrewgordstewart:
(Ignore this close: testing a github projects feature)

snario:
What's the status of this PR?

packages/e2e-tests/puppeteer/__tests__/web3torrent/stress.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    A = 'A',
+    B = 'B',
+    C = 'C',
+    D = 'D'
+  }
+
+  const leechers: Label[] = [Label.B, Label.C, Label.D];
+  const labels: Label[] = leechers.concat([Label.A]);
+
+  type T = {browser: Browser; metamask: Dappeteer; tab: Page};
```
lalexgap, line 42: Nit: could this be a more descriptive type name?

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -170,12 +170,15 @@ export class PaymentChannelClient {
   }
 
   insertIntoChannelCache(channelState: ChannelState) {
-    this.channelCache = {...this.channelCache, [channelState.channelId]: channelState};
+    // TODO: The joinChannel promise never resolves, meaning this function is not called
```
lalexgap, line 173: Is this still the case? I thought we fixed the `joinChannel` not resolving but I could be wrong.

andrewgordstewart: It's an out of date comment, thanks to a three-way merge

PR #1443, opened 2020-04-15
[web3torrent] Clean up File component

kerzhner:
- Use `function` syntax when possible. I am finding a constant that defines a function (with Typescript types sprinkled in there) pretty hard to parse. If others agree, please use `function` syntax going forward (when possible).
- `File` component has a `useContext` hook now, so we can remove `Web3TorrentContext.Consumer`. In general, I might be a good idea to minimize the logic inside jsx and instead use jsx primarily to describe UI.

### DIFF
--- a/packages/web3torrent/src/pages/file/File.tsx
+++ b/packages/web3torrent/src/pages/file/File.tsx
@@ -12,14 +12,14 @@ import './File.scss';
 import WebTorrentPaidStreamingClient from '../../library/web3torrent-lib';
 import _ from 'lodash';
 
-const getLiveData: (
+function getLiveData(
   web3Torrent: WebTorrentPaidStreamingClient,
   setTorrent: React.Dispatch<React.SetStateAction<Torrent>>,
   torrent: Torrent
-) => void = (web3Torrent, setTorrent, torrent) => {
+): void {
   const liveTorrent = torrentStatusChecker(web3Torrent, torrent, torrent.infoHash);
   setTorrent(liveTorrent);
-};
+}
 
 interface Props {
   ready: boolean;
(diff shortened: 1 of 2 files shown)


packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -46,41 +46,33 @@ const File: React.FC<Props> = props => {
...
-                  mySigningAddress={me}
-                />
-              ) : (
-                false
-              )}
-            </>
-          );
-        }}
-      </Web3TorrentContext.Consumer>
+      <>
```
spersico, line 62: this react fragment is not necessary

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -46,41 +46,33 @@ const File: React.FC<Props> = props => {
...
+        <br />
+        {showBudget ? (
+          <SiteBudgetTable
+            budgetCache={budgetCache}
+            channelCache={channelCache}
+            mySigningAddress={me}
+          />
+        ) : (
+          false
+        )}
```
spersico, line 73: this could be 
```showBudget && (<SiteBudgetTable ....``` to avoid the `: (false)`. 
`false` values don't get shown in the render.

File: packages/web3torrent/src/pages/file/File.tsx [hunk 3]
```diff
@@ -46,41 +46,33 @@ const File: React.FC<Props> = props => {
...
+      </>
+      }
```
spersico, line 75: Ups! You missed one :stuck_out_tongue: 
image

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -46,41 +46,33 @@ const File: React.FC<Props> = props => {
...
       {torrent.status === Status.Idle ? (
```
spersico, line 76: this could also be `torrent.status && Status.Idle ? (....`

packages/web3torrent/src/pages/file/File.tsx
kerzhner: 37e02c957

packages/web3torrent/src/pages/file/File.tsx
kerzhner: 37e02c957

packages/web3torrent/src/pages/file/File.tsx
kerzhner: Good catch! 37e02c957

packages/web3torrent/src/pages/file/File.tsx
kerzhner: 37e02c957

PR #1444, opened 2020-04-15
Assign Request Id for joiner in application workflow

### LINKED ISSUE #1438: Join channel call doesn't always resolve

lalexgap:
Fixes #1438 .

We weren't assigning the `requestId` as the joiner, which meant that the `JoinChannel` response would not contain an id and thus the `channel-provider` would never resolve the `joinChannel` call.

### DIFF
--- a/packages/xstate-wallet/src/messaging.ts
+++ b/packages/xstate-wallet/src/messaging.ts
@@ -91,6 +91,9 @@ export class MessagingService implements MessagingServiceInterface {
   }
 
   public async sendResponse(id: number, result: Response['result']) {
+    if (!id) {
+      throw new Error(`No id provided for the response`);
+    }
     const response: Response = {id, jsonrpc: '2.0', result};
     this.eventEmitter.emit('SendMessage', response);
   }
--- a/packages/xstate-wallet/src/workflows/application.ts
+++ b/packages/xstate-wallet/src/workflows/application.ts
@@ -196,7 +196,7 @@ const generateConfig = (
 
     //This could handled by another workflow instead of the application workflow
     closing: {
-      entry: actions.displayUi,
+      entry: [actions.displayUi, actions.assignRequestId],
       exit: actions.hideUi,
       invoke: {
         id: 'closing-protocol',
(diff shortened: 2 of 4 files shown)


packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -135,7 +136,7 @@ const generateConfig = (
             ],
             JOIN_CHANNEL: {
               target: 'settingSite',
-              actions: [actions.sendJoinChannelResponse]
+              actions: [actions.assignRequestId, actions.sendJoinChannelResponse]
```
andrewgordstewart, line 139: The context here seems a bit off to me -- it suggests that the application workflow is responsible for a single request.

packages/xstate-wallet/src/messaging.ts
```diff
@@ -91,6 +91,9 @@ export class MessagingService implements MessagingServiceInterface {
   }
 
   public async sendResponse(id: number, result: Response['result']) {
+    if (!id) {
+      throw new Error(`No id provided for the response`);
+    }
```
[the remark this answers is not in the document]
lalexgap: It looks like enabling response validation resulted in a bunch of schema errors so I've created #1447 to tackle that.

packages/xstate-wallet/src/workflows/application.ts
lalexgap: I think the `application` workflow will only handle a `Create/JoinRequest` and a `ConcludeRequest` right now so perhaps we should store them separately on the `context` ?

andrewgordstewart: In any case, seems unrelated to this change.

packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -196,7 +196,7 @@ const generateConfig = (
 
     //This could handled by another workflow instead of the application workflow
     closing: {
-      entry: actions.displayUi,
+      entry: [actions.displayUi, actions.assignRequestId],
```
andrewgordstewart, line 199: Whether or not the UI is displayed should probably be the responsibility of the closing protocol (UI shouldn't be displayed when virtually-funded)

PR #1445, opened 2020-04-15
[web3torrent] remove synchronous, pure effect

kerzhner:
I _think_ that `useEffect` in the file component is unnecessary (and might introduce insidious race conditions).

Also, it looks to me like the `buttonLabel` state is directly derived from the `loading` state. I created a pure function to derive the `buttonLabel` from `loading` instead of having two state variables. The advantages are:
- The component is rendered fewer times as state is set once instead of twice.
- `buttonLabel` and `loading` won't have incompatible values.


packages/web3torrent/src/utils/torrent-status-checker.ts
```diff
@@ -76,9 +76,6 @@ export function torrentStatusChecker(
       length: live.length || previousData.length,
       downloaded: live.downloaded || 0,
       status: getStatus(live, web3Torrent.pseAccount),
-      uploadSpeed: live.uploadSpeed,
```
spersico, line 79: I think these 3 lines are making problems:
image

spersico: The number of peers and the download/upload speed it's not showing on the leecher nor the seeder.

spersico:
Hey! I found this, I think it's got something to do with the files that you are modifying, want to give it a look in this PR? Otherwise, I can do it tomorrow.

packages/web3torrent/src/utils/torrent-status-checker.ts
kerzhner: Good catch! Do you know if this an issue on the current `master`, or just on this branch?

kerzhner: Regarding the lines this comment is referencing: these lines are deleted because the lines will never have any effect.

spersico: > Good catch! Do you know if this an issue on the current master, or just on this branch?

Just this branch, I think

kerzhner: I believe this should be fixed in a4a137545.

kerzhner: Screen Shot 2020-04-16 at 1 55 26 PM

spersico: awesome!

kerzhner:
> so sad to see it go... 😢

It was an epic piece of code, I must admit.

kerzhner:
This PR addresses https://github.com/statechannels/monorepo/issues/1454 by simplifying torrent state.

PR #1449, opened 2020-04-15
remove global network check, add initial  torrent check

### LINKED ISSUE #1423: Fix firewall check

spersico:
Adds status checks for individual torrents (closes #1423 ) . The status check test consists of a separate tracker client than the one on the web3torrent client. This tracker looks up the torrent on web3torrent's tracker server, with 3 possible results:
- **Seeders found:** the tracker client had responses from the server, and was able to find an active peer. The warning is not shown at all.
- **No seeders found:** the tracker client got responses from the tracker but wasn't able to find an active peer. Might mean that the torrent doesn't have seeders, but it could also mean that the tracker is slow, or the seeder has a slow connection.
- **No connection:** there was no answer at all from the tracker server, which might mean network connections.

_Warnings don't disable any buttons or anything like that._

## Screenshots
image

It works for seeders too! (if they can't connect to the tracker)
Screenshot 2020-04-16 05:36:00

### DIFF
--- a/packages/web3torrent/src/App.tsx
+++ b/packages/web3torrent/src/App.tsx
@@ -1,6 +1,5 @@
 import ConnectionBanner from '@rimble/connection-banner';
-import {Flash} from 'rimble-ui';
-import React, {useState, useContext, useEffect} from 'react';
+import React, {useState, useEffect} from 'react';
 import {Route, Switch, BrowserRouter} from 'react-router-dom';
 import './App.scss';
 import {LayoutFooter, LayoutHeader} from './components/layout';
@@ -9,12 +8,9 @@ import Welcome from './pages/welcome/Welcome';
 import File from './pages/file/File';
 import Upload from './pages/upload/Upload';
 import {RoutePath} from './routes';
-import {Web3TorrentContext} from './clients/web3torrent-client';
-import {TorrentClientCapabilities} from './library/types';
 import {requiredNetwork} from './constants';
 
 const App: React.FC = () => {
-  const Web3Torrent = useContext(Web3TorrentContext);
   const [currentNetwork, setCurrentNetwork] = useState(
     window.ethereum ? Number(window.ethereum.networkVersion) : undefined
   );
@@ -31,16 +27,9 @@ const App: React.FC = () => {
   }, []);
 
   const ready = currentNetwork === requiredNetwork;
-  const showNetworkWarning = Web3Torrent.clientCapability === TorrentClientCapabilities.NOT_CAPABLE;
   return (
     <BrowserRouter>
       <main>
-        {showNetworkWarning && (
-          <Flash variant="danger">
-            Looks like you do not have an internet connection or are behind a firewall. Web3Torrent
-            may not work on some public wifi networks.
-          </Flash>
-        )}
         <ConnectionBanner
           currentNetwork={currentNetwork}
           requiredNetwork={requiredNetwork}
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -17,11 +17,17 @@ export const BUFFER_REFILL_RATE = utils.bigNumberify(WEI_PER_BYTE.mul(BLOCK_LENG
 
 export const INITIAL_SEEDER_BALANCE = utils.bigNumberify(0); // needs to be zero so that depositing works correctly (unidirectional payment channel)
 
-const VERSION_STR = '0.7.2'.replace(/\d*./g, v => `0${Number(v) % 100}`.slice(-2)).slice(0, 4);
+const randomNumberGenerator = (length: number) => {
+  // Programatic way of getting fixed length number, based of https://stackoverflow.com/a/21816636/6569950
+  const base = Math.pow(10, length - 1);
+  return Math.floor(base + Math.random() * 9 * base);
+};
+
 export function generateRandomPeerId() {
-  return Buffer.from(
-    `-WW${VERSION_STR}-` + btoa((Math.random() * 99999999999).toFixed(0).slice(0, 9))
-  ).toString('hex');
+  // based of webtorrent's way of generating random peerId's
+  // https://github.com/webtorrent/webtorrent/blob/ed2809159585d2611dff24a48f25748baf78e8fb/index.js#L52
+  const randomNumber = String(randomNumberGenerator(9));
+  return Buffer.from(`-WW0007-${btoa(randomNumber)}`).toString('hex');
 }
 
 // firebase setup
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -85,13 +85,15 @@ export const preSeededTorrents: Array<Partial<Torrent>> = [
 ];
 
 // Welcome Page Tracker creation options
-export const welcomePageTrackerOpts = {
+export const defaultTrackerOpts = {
   infoHash: [preSeededTorrents[0].infoHash],
   announce: defaultTrackers,
-  peerId: '2d5757303030372d37454e613073307937495630', // random
+  peerId: generateRandomPeerId(), // random
   port: 6881,
   getAnnounceOpts: () => ({
     pseAccount: '0x7F0126D6c4270498b6514Cb934a3274898f68777',
+    // the pseAccount is only used to allow the client on the tracker server, see
+    // https://github.com/statechannels/monorepo/blob/d0c6b1be3a637c88880c0c937d9a6ebc8078799c/packages/web3torrent/tracker/index.js#L16
     uploaded: 0,
     downloaded: 0
   }) // dummy pseAccount, but it works
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -77,7 +77,11 @@ export const welcomePageTrackerOpts = {
   announce: defaultTrackers,
   peerId: '2d5757303030372d37454e613073307937495630', // random
   port: 6881,
-  getAnnounceOpts: () => ({pseAccount: '0x7F0126D6c4270498b6514Cb934a3274898f68777'}) // dummy pseAccount, but it works
+  getAnnounceOpts: () => ({
+    pseAccount: '0x7F0126D6c4270498b6514Cb934a3274898f68777',
+    uploaded: 0,
+    downloaded: 0
+  }) // dummy pseAccount, but it works
 };
 
 export const testTorrent = {
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -17,6 +17,13 @@ export const BUFFER_REFILL_RATE = utils.bigNumberify(WEI_PER_BYTE.mul(BLOCK_LENG
 
 export const INITIAL_SEEDER_BALANCE = utils.bigNumberify(0); // needs to be zero so that depositing works correctly (unidirectional payment channel)
 
+const VERSION_STR = '0.7.2'.replace(/\d*./g, v => `0${Number(v) % 100}`.slice(-2)).slice(0, 4);
+export function generateRandomPeerId() {
+  return Buffer.from(
+    `-WW${VERSION_STR}-` + btoa((Math.random() * 99999999999).toFixed(0).slice(0, 9))
+  ).toString('hex');
+}
+
 // firebase setup
 export const HUB = {
   signingAddress: '0xaaaa84838319627Fa056fC3FC29ab94d479B8502',
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -1,6 +1,5 @@
 import debug from 'debug';
 import WebTorrent, {Torrent, TorrentOptions} from 'webtorrent';
-import {Client} from 'bittorrent-tracker';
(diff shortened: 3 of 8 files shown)


packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -33,6 +46,19 @@ const File: React.FC<Props> = props => {
...
+    const testResult = async () => {
+      const torrentCheckResult = await checkTorrent(web3Torrent, infoHash);
+      setWarning(torrentCheckResult);
+    };
+
+    if (infoHash) {
+      testResult();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
kerzhner, line 61: Should this effect get executed every time `web3Torrent` or `infoHash` changes? Why wouldn't it?

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -9,8 +9,21 @@ import {parseURL, useQuery} from '../../utils/magnet';
 import {torrentStatusChecker} from '../../utils/torrent-status-checker';
 import {useInterval} from '../../utils/useInterval';
 import './File.scss';
-import WebTorrentPaidStreamingClient from '../../library/web3torrent-lib';
+import WebTorrentPaidStreamingClient, {TorrentTestResult} from '../../library/web3torrent-lib';
 import _ from 'lodash';
+import {Flash} from 'rimble-ui';
+
+const checkTorrent = async (web3Torrent, infoHash) => {
```
kerzhner, line 16: Use function syntax and add types to parameters.

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -9,8 +9,21 @@ import {parseURL, useQuery} from '../../utils/magnet';
...
+  const testResult = await web3Torrent.checkTorrentInTracker(infoHash);
+  switch (testResult) {
+    case TorrentTestResult.NO_CONNECTION:
+      return `Your connection to the tracker may be limited, you might have unexpected functionality`;
```
kerzhner, line 20: Better to use `` ` `` for strings instead of `'`

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -9,8 +9,21 @@ import {parseURL, useQuery} from '../../utils/magnet';
...
+    case TorrentTestResult.NO_SEEDERS_FOUND:
+      return `Seems like the torrent doesn't have any seeders. You can give it a try nonetheless.`;
```
kerzhner, line 22:
```suggestion
      return 'Seems like the torrent doesn't have any seeders. You can give it a try nonetheless.';
```

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -9,8 +9,21 @@ import {parseURL, useQuery} from '../../utils/magnet';
...
+    default:
+      return ``;
```
kerzhner, line 24:
```suggestion
      return '';
```

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -172,6 +147,37 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     }
   }
 
+  async checkTorrentInTracker(infoHash: string) {
+    const trackerClient: Client = new Client({
```
kerzhner, line 151: Why is a new client needed? Why can't we use the existing `WebTorrent` client?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -172,6 +147,37 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+      });
+      trackerClient.once('update', updateIfSeederFound);
+      trackerClient.start();
+    });
+    const timer: Promise<undefined> = new Promise((resolve, _) => setTimeout(resolve, 5000));
+    const raceResult = await Promise.race([gotAWire, timer]);
+    trackerClient.stop();
+    trackerClient.destroy();
+    if (raceResult) {
+      return TorrentTestResult.SEEDERS_FOUND; // Could connect connect to a peer
```
geoknee, line 172:
```suggestion
      return TorrentTestResult.SEEDERS_FOUND; // Could connect to a peer
```

packages/web3torrent/src/constants.ts
```diff
@@ -17,6 +17,13 @@ export const BUFFER_REFILL_RATE = utils.bigNumberify(WEI_PER_BYTE.mul(BLOCK_LENG
 
 export const INITIAL_SEEDER_BALANCE = utils.bigNumberify(0); // needs to be zero so that depositing works correctly (unidirectional payment channel)
 
+const VERSION_STR = '0.7.2'.replace(/\d*./g, v => `0${Number(v) % 100}`.slice(-2)).slice(0, 4);
```
geoknee, line 20: This is.... eye-catching. Could you add a comment explaining what is going on here?

packages/web3torrent/src/library/web3torrent-lib.ts
kerzhner: Have we tried using `noPeers` event: `torrent.on('noPeers', function (announceType) {}))` described in the [docs](https://webtorrent.io/docs)?

packages/web3torrent/src/constants.ts
spersico: sure! BTW, this is a rewrite from [here](https://github.com/webtorrent/webtorrent/blob/ed2809159585d2611dff24a48f25748baf78e8fb/index.js#L52)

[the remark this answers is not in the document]
spersico: this is a rewrite of [this](https://github.com/webtorrent/webtorrent/blob/ed2809159585d2611dff24a48f25748baf78e8fb/index.js#L52). But I hardcoded the version of WebTorrent that last night I thought that we were using (super wrong about that, In my head it was 0.7.2, and it is 0.107.16)

packages/web3torrent/src/pages/file/File.tsx
spersico: this should be executed once, and just once, at the moment that the client has the infoHash.

On the seeder side, it allows the user to know if the torrent was announced to the tracker.
On the leecher side, it allows the user to know if it could communicate with the tracker, and if the tracker came up with any peers that could seed the file.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: > Why is a new client needed? Why can't we use the existing WebTorrent client?

I tried to, but it didn't work. I can try later again, but the issue that I had is that the test results never found peers if I did that. That's why I even make a randomized peerId, so it behaves like a different tracker client than the one on the webtorrent client.

> Have we tried using noPeers event: torrent.on('noPeers', function (announceType) {})) described in the docs?

Noup, I'll take a look

packages/web3torrent/src/pages/file/File.tsx
kerzhner: Got it. But what about when the `web3Torrent` client changes? When the `web3Torrent` client is updated (which doesn't happen in the code currently, but could in the future), shouldn't the `testResult` get retriggered?

What about when the `infoHash` is update? Isn't the `testResult` invalid in that case as well?

Maybe a bigger point is the following. The linter error is saying: 
"Hey, `web3Torrent` and `infoHash` are variables that can change when the component is re-rendered. Are you sure you don't want to rerun the effect when these things change?"

It strikes me that we are ignoring this linter error because we state: "We know better. `web3Torrent` and `infoHash never change". But that's an assumption made in the codebase at this instant. What is the downside of including `web3Torrent` and `hashInfo` in the effect second argument array?

packages/web3torrent/src/library/web3torrent-lib.ts
kerzhner: No worries, I was mostly just wondering.

packages/web3torrent/src/pages/file/File.tsx
spersico: The test doesn't check for anything regarding the `webtorrent` client at the moment. Nor the web3torrent client.
The torrent check could be a utility function, as it doesn't depend in any way on the webtorrent client (and now that I think of it, I'm wondering why I thought it was a good idea to put it there... I blame sleepiness, but I can move it to another file so that the webtorrent client doesn't get more cluttered).

About infoHash updates. I was being careful, in order to avoid running 2 times this function at the same time, thus creating 2 trackerClients. I agree that adding infoHash doesn't have a negative consequence. But to be fair, I don't expect the infoHash to ever change. Once the value is set, I expect to be constant through the whole life of this component. Remember that the context of this page is a single torrent, and the infoHash is, in a way, a torrent's id

packages/web3torrent/src/pages/file/File.tsx
spersico: here this suggestion would make problems. because _doesn't_ has a `'`. I can change it to a `"` though

packages/web3torrent/src/pages/file/File.tsx
kerzhner: Ok that makes a lot more sense, thank you for explanation 🙏! I think removing the `web3Torrent` dependency clear up my mental model.

spersico: cool. Thank you too!. It helped me too, to realize that I should move the function to another file so that the web3torrent client doesn't get cluttered with stuff that it doesn't make sense to be there.

I think my reasoning at the time was that _" because the global checker was there, this should be there too"_

packages/web3torrent/src/pages/file/File.tsx
kerzhner: Good catch! Updated the suggestion

packages/web3torrent/src/pages/file/File.tsx
spersico:
```suggestion
      return 'Your connection to the tracker may be limited, you might have unexpected functionality';
```

packages/web3torrent/src/App.tsx
```diff
@@ -31,16 +27,9 @@ const App: React.FC = () => {
   }, []);
 
   const ready = currentNetwork === requiredNetwork;
-  const showNetworkWarning = Web3Torrent.clientCapability === TorrentClientCapabilities.NOT_CAPABLE;
```
geoknee, line 34: @spersico following our discussion earlier about `useContext`: was it the case here that if `Web3Torrent.clientCapability` changed value, the component would not necessarily re-render?

spersico: yeah... this code wasn't working on master (at least on my PC). The previous state of this, before @kerzhner's refactor, I think it used to emit events in the Web3Torrent library that the UI listened so that the changes triggered a re-render. I just ran a quick test:
I put a console log at the end of the global test, to see if it was supposed to show the warning, the test failed (as always) and.. no warning shown:
image


Doesn't really matter now, as all that was removed though.

kerzhner: I think having a mutable context object is an antipattern. So yes.

kerzhner: If it's not hard, would you be able to point to where the we used to listen to the events (before the refactor)?

geoknee: @spersico yeah it doesn't matter, but good to understand what was suboptimal previously

spersico: @kerzhner sure, here it is : https://github.com/statechannels/monorepo/commit/68bb9d606838aa014992667a178767d7607fa0f4#diff-aaa8f7e584ffc4ebc36d61e9313a7075R44

kerzhner: 🙏 Good catch. Mind filing an issue? Shouldn't be too hard to fix I think...

spersico: It's not necessary, I think, that code get's removed by this PR.

packages/web3torrent/src/constants.ts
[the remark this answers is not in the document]
spersico: https://github.com/statechannels/monorepo/pull/1449/commits/cc79225268d6337a8d5eae8e7764788d3d583483

spersico: https://github.com/statechannels/monorepo/pull/1449/commits/cc79225268d6337a8d5eae8e7764788d3d583483

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: decided to test and touch more stuff tomorrow, so we can merge this one, and I don't block your refactor.

packages/web3torrent/src/pages/file/File.tsx
spersico: https://github.com/statechannels/monorepo/pull/1449/commits/227e2312f728634016f0b2000704e71dc73c4cc1

packages/web3torrent/src/pages/file/File.tsx
spersico: https://github.com/statechannels/monorepo/pull/1449/commits/94c9a9125ce5465046ebb16f4a2eb594017087d1

PR #1450, opened 2020-04-15
Enable validation of outgoing responses

### LINKED ISSUE #1447: Enable response validation (by the schema) and fix any validation errors.
We should enable validation on outgoing responses from the wallet and fix any existing issues around validation.

lalexgap:
Fixes #1447 

Enables validating of outgoing responses. I've also fixed a few different formatting errors that was causing the response validation to fail.

### DIFF
--- a/packages/xstate-wallet/src/serde/app-messages/serialize.ts
+++ b/packages/xstate-wallet/src/serde/app-messages/serialize.ts
@@ -3,31 +3,35 @@ import {
   Allocations as AppAllocations,
   AllocationItem as AppAllocationItem,
   SiteBudget as AppSiteBudget,
-  TokenBudget
+  TokenBudget,
+  ChannelResult,
+  ChannelStatus
 } from '@statechannels/client-api-schema';
 import {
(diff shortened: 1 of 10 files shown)


packages/xstate-wallet/src/serde/app-messages/serialize.ts
```diff
@@ -63,6 +67,41 @@ function serializeSimpleAllocation(allocation: SimpleAllocation): AppAllocation
...
+}
+
+export async function serializeChannelEntry(
+  channelEntry: ChannelStoreEntry
+): Promise<ChannelResult> {
+  const {latest, channelId} = channelEntry;
+  const {appData, turnNum} = latest;
+  const {participants, appDefinition} = channelEntry.channelConstants;
+
+  const outcome = latest.outcome;
```
andrewgordstewart, line 81:
```suggestion
  const {appData, turnNum, outcome} = latest;
  const {participants, appDefinition} = channelEntry.channelConstants;
```

PR #1452, opened 2020-04-16
[xstate-wallet] Return selectedAddress from Chain property

geoknee:
...not from the `window` property (which is updated async, and therefore might not exist yet).

This should fix the following race condition: 

- call `window.ethereum.enable()` and wait for the promise to resolve
- inspect / use `window.ethereum.selectedAddress` and expect it to exist (with some p > 0, it does not yet exist)

The solution is that `window.ethereum.enable` resolves to `[selectedAddress]`, i.e. that which will be set on `window.ethereum.selectedAddress` some time in the future. So we wait for the enable promise to resolve, store the value, and use that as our source of truth when querying the selectedAddress.

### DIFF
--- a/packages/xstate-wallet/src/chain.ts
+++ b/packages/xstate-wallet/src/chain.ts
@@ -39,7 +39,7 @@ export interface ChannelChainInfo {
 export interface Chain {
   // Properties
   ethereumIsEnabled: boolean;
-  selectedAddress: string | null;
+  selectedAddress: string | undefined;
 
   // Feeds
   chainUpdatedFeed: (channelId: string) => Observable<ChannelChainInfo>;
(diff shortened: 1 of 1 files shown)


packages/xstate-wallet/src/chain.ts
```diff
@@ -248,8 +249,19 @@ export class ChainWatcher implements Chain {
...
     if (window.ethereum) {
-      const [selectedAddress] = await window.ethereum.enable();
-      return selectedAddress;
+      try {
+        this.selectedAddress = (await window.ethereum.enable())[0];
+        if (typeof this.selectedAddress === 'string') {
+          return this.selectedAddress;
+        } else {
+          console.error('Ethereum enabled but no selected address is defined');
+          return Promise.reject('Ethereum enabled but no selected address is defined');
```
tomclose, line 258: Out of interest, how / when would this happen?

geoknee: It shouldn't ever happen, as long as `window.ethereum.enable()` returns a promise that resolves to an array of strings. But we have no typings in place to be sure of that. 

I suppose we could just assert the type here, but dealing with the case where it resolves to something unexpected means we are building to the interface that we have, rather than the current implementation of metamask.

PR #1462, opened 2020-04-16
Flesh out Withdrawal UI

### LINKED ISSUE #1379: Clean up withdrawing screens
Update the withdrawing UI in xstate wallet. We should use the existing approve and fund UI as a guide.

### LINKED ISSUE #1407: Ensure Budget is only updated when withdraw suceeds
Right now the budget gets cleared out even if the withdraw transaction fails. We should only update the budget after the transaction has suceeded.

lalexgap:
Updates the withdrawal workflow and UI. The UI basically copies the styling from ApproveBudgetAndFund.

image

Fixes #1379 
Fixes #1407


packages/xstate-wallet/src/ui/close-ledger-and-withdraw.tsx
```diff
@@ -1,37 +1,120 @@
...
-  const current = props.current;
-  const prompt = (
+  const [current, send] = useService(props.service);
+
+  const waitForUserApproval = ({waiting, budget}: {waiting: boolean; budget: SiteBudget}) => {
+    const {playerAmount, hubAmount} = getAmountsFromBudget(budget);
+    return (
+      <Flex alignItems="left" flexDirection="column">
+        <Heading textAlign="center" mb={0}>
+          Close your App Budget
```
tomclose, line 24: Maybe "Withdraw Funds"

packages/xstate-wallet/src/ui/close-ledger-and-withdraw.tsx
```diff
@@ -1,37 +1,120 @@
...
+        </Heading>
+        <Heading textAlign="center" as="h4" mt={0} mb={2}>
+          {budget.domain}
+        </Heading>
+
+        <Text fontSize={1} pb={2}>
+          Close your budget with <strong>{budget.domain}</strong> and withdraw your funds?
```
tomclose, line 31: Maybe "Close your hub connection for ... and withdraw your funds?"

packages/xstate-wallet/src/ui/close-ledger-and-withdraw.tsx
```diff
@@ -1,37 +1,120 @@
...
+        </Text>
+
+        <Flex justifyContent="center" pb={2}>
+          <Box>
+            <Text>Send: {formatEther(playerAmount)} ETH</Text>
+            <Text>Receive: {formatEther(hubAmount)} ETH</Text>
+          </Box>
+        </Flex>
```
tomclose, line 39: "Current balances: You ..., Hub ..." ? Might not even need this table. All that really matters at this point is the amount that they stand to receive.

packages/xstate-wallet/src/ui/close-ledger-and-withdraw.tsx
```diff
@@ -1,37 +1,120 @@
...
+
+        <Text pb={3} fontSize={1}>
+          You will receive {formatEther(playerAmount)} ETH and the budget will be closed with the
+          channel hub.
```
tomclose, line 43: "and your connection to the hub will be closed"?

packages/xstate-wallet/src/utils/workflow-utils.ts
```diff
@@ -72,3 +83,53 @@ export const debugAction = (c, e, {state}) => {
   // eslint-disable-next-line no-debugger
   debugger;
 };
+
+export const displayUI = (
+  messagingService: MessagingServiceInterface
+): ActionFunction<any, any> => () => messagingService.sendDisplayMessage('Show');
```
tomclose, line 89: There's a genuine change here from the code that it replaced:
```ts
const displayUI = (messagingService: MessagingServiceInterface): ActionObject<Context, Event> => ({
  type: 'displayUI',
  exec: () => messagingService.sendDisplayMessage('Show')
});
```
The original code includes a `type`, which is what is displayed on the xstate viewer, if you display the machine directly. Would be good to keep that property. (Unless there's something going on here that I don't understand. Did you manage to get your machine to display in xstate devtools?)

packages/xstate-wallet/src/workflows/close-ledger-and-withdraw.ts
```diff
@@ -1,87 +1,173 @@
...
+        src: Services.observeFundsWithdrawal
+      },
+      states: {
+        submitTransaction: {
+          invoke: {
+            id: 'submitTransaction',
+            src: Services.submitWithdrawTransaction,
+            onDone: {
+              target: 'waitMining',
+              actions: [Actions.setTransactionId]
```
tomclose, line 148: Not expecting a change in this PR, but I'm strongly in favour of removing away from defining config in terms of strings, and move towards putting the functions in directly. This opens the path to better typesafety (you see it immediately with actions on named events, which are checked against the actual action type they will be passed), and also to better code discovery, by being able to directly find the code that will execute at that point, instead of having to trace through to where it is plugged together.

packages/xstate-wallet/src/workflows/close-ledger-and-withdraw.ts
```diff
@@ -1,87 +1,173 @@
...
       invoke: {
-        src: 'createObjective',
-        onDone: {target: 'closeLedger', actions: ['assignLedgerId']}
+        src: Services.createObjective,
+        onDone: {target: 'closeLedger', actions: [Actions.assignLedgerId]}
       }
     },
-    closeLedger: getDataAndInvoke({src: 'getFinalState'}, {src: 'supportState'}, 'withdraw'),
+    closeLedger: getDataAndInvoke<LedgerExists>(
+      {src: Services.constructFinalState},
```
tomclose, line 128: Would be an option here to separate these out into two steps in this machine, and store the final state on this machine itself. It seems like that would give you slightly better control if you wanted to handle the case where one of these steps failed, in terms of what you show to the user / the next actions you take.

packages/xstate-wallet/src/workflows/close-ledger-and-withdraw.ts
lalexgap: Does defining the action functions inline work with the xstate visualizer? I noticed that currently approve budget and fund is not generating a config for the visualizer.

lalexgap: Is `xstate-devtools` the visualizer or the chrome extension?

packages/xstate-wallet/src/utils/workflow-utils.ts
lalexgap: Is `xstate-devtools` the visualizer? The `hideUI/displayUI` seem to show ok up https://xstate.js.org/viz/?gist=8d968ac87a9dd068c51f1c6c8ca8f9bf

packages/client-api-schema/src/types.ts
```diff
@@ -46,7 +46,11 @@ interface JsonRpcNotification<NotificationName, NotificationParams> {
   params: NotificationParams;
 }
 
-interface JsonRpcError<Code, Message, Data = undefined> {
+export const UserDeclinedErrorCode = 200;
+export const EthereumNotEnabledErrorCode = 100;
+
+export type ErrorCode = typeof UserDeclinedErrorCode | typeof EthereumNotEnabledErrorCode;
```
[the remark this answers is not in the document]
lalexgap: I think `ErrorCode` type is mostly used for the `JsonRpcError<Code extends ErrorCode, Message, Data = undefined>` interface, so I've stopped exporting `ErrorCode`

PR #1465, opened 2020-04-17
Improved xstate-wallet logging

andrewgordstewart:
This adds pino to xstate-wallet, and sets the default behaviour (when `LOG_DESTINATION=='console'` is true) to pretty print to the console:
Screen Shot 2020-04-16 at 8 03 38 PM

When the `LOG_DESTINATION` env variable is present, but not `'console'` it is configured to log to that file _if `NODE_ENV` is `'test'`_. Otherwise, it is assumed you're running in the browser, where logging directly to a file is not possible. In this case, the message is serialized as a string before passed to `window.console`, since `window.console.log` would otherwise print `JSHandle@object`.

The circleCI integration tests configure `LOG_DESTINATION` to be `/temp/xstate-wallet.log`, resulting in JSON blobs being logged [as an artifact](https://19048-209829093-gh.circle-artifacts.com/0/tmp/xstate-wallet.log). This log can be parsed with utilities like `jq`, `pino-pretty`, or `grep`.

Resolves https://github.com/statechannels/monorepo/issues/1458


packages/xstate-wallet/src/logger.ts
```diff
@@ -0,0 +1,20 @@
+import pino from 'pino';
+import {LOG_FILE} from './constants';
+
+// TODO: Is there a better way to determine if we're in a browser context?
+const isBrowserContext = process.env.NODE_ENV === 'test';
```
junjizhi, line 5: Do you mean `isTestContext`?

.circleci/config.yml
```diff
@@ -248,6 +248,7 @@ jobs:
       GANACHE_CACHE_FOLDER: ../../.ganache-deployments
       GANACHE_PORT: 8547
       NODE_ENV: test
+      LOG_FILE: /tmp/xstate-wallet.log
```
junjizhi, line 251: How about `XSTATE_WALLET_LOG_FILE`? If we don't plan to reuse this var for all packages and pipe their logs into the same file, having a more specific name makes easier to understand.

packages/xstate-wallet/src/constants.ts
```diff
@@ -59,4 +59,4 @@ export const CLEAR_STORAGE_ON_START = getBool(process.env.CLEAR_STORAGE_ON_START
 export const USE_INDEXED_DB = getBool(process.env.USE_INDEXED_DB, false);
 
 export const LOG_FILE = process.env.LOG_FILE;
-export const ADD_LOGS = LOG_FILE ? true : getBool(process.env.ADD_LOGS);
+export const ADD_LOGS = LOG_FILE ? true : getBool(process.env.ADD_LOGS || 'false');
```
junjizhi: Ideally, we just use one env var to denote whether we do logging. Now we have two... that would make it confusing to anyone who looks at this.

packages/xstate-wallet/src/logger.ts
andrewgordstewart: I actually mean `isBrowserContext`. But the formula is wrong!

```suggestion
const isBrowserContext = process.env.NODE_ENV !== 'test';
```

.circleci/config.yml
andrewgordstewart: I think we should be piping logs into the same file. Having logs from multiple sources in a single file, with "close-enough" proximity of log statements, seems more useful than having multiple files that require cross-referencing.

As long as there is a time stamp and an origin marker (`name` in xstate-wallet as of this PR) it doesn't matter that much, since you can either filter-by-name to split into multiple files, or you can concat and sort-by-timestamp to combine multiple files.

packages/xstate-wallet/src/constants.ts
andrewgordstewart: There are two almost orthogonal issues:
1. do you want logs to happen?
2. where should logs go?

An alternative could be a single `LOG_DESTINATION` variable which:
- when `undefined`, turn it off
- when is `'console'`, log to console
- otherwise, assume `LOG_DESTINATION` is a filename, and log to the file

What do you suggest?

junjizhi: I like the `LOG_DESTINATION` approach. People don't have to wrap the head around to juggle two vars. Less is better, to me.

PR #1468, opened 2020-04-17
[rps] Fix the app

geoknee:
In light of xstate wallet integration. 

- strips out almost everything to do with "logging in".
- Does not / will not handle virtual funding in RPS (stick to direct funding)
- Does not yet tackle e2e tests, saving that for a quick (?) follow up PR.


packages/rps/src/redux/store.ts
```diff
@@ -54,16 +50,18 @@ function* rootSaga() {
...
-    yield call(
-      [client, 'approveBudgetAndFund'],
-      '0x8ac7230489e80000', // 10 eth
-      '0x8ac7230489e80000',
-      window.channelProvider.selectedAddress,
-      HUB.signingAddress,
-      HUB.outcomeAddress
-    );
+
+    setupFirebase(client, window.channelProvider.signingAddress);
```
geoknee, line 54: I don't know why this was necessary, but I like the change because it does the firebase messaging setup without using a saga (fewer lines, matches web3t pattern). 

Syncing open games is unaffected, still uses a saga.

PR #1470, opened 2020-04-17
[web3torrent] torrent used by React is now immutable

kerzhner:
The most important changes are in the `torrent-status-checker.ts`.

The File component is the UI for a torrent. The component now uses an immutable `TorrentUI` object that is created from a computation based on the url of the page and the object pulled out from the `web3torrent client` (using the hash id).

Other benefits of this PR:
- Removing the dangerous force casting of object `as Torrent` in `torrent-status-checker`.
- Passing only the data React components need instead of the full fledged object pulled out from the web3torrent client.
- starting to tease apart data management and UI management.

Fixes https://github.com/statechannels/monorepo/issues/1464


packages/web3torrent/src/utils/magnet.ts
```diff
@@ -13,28 +12,6 @@ export const useQuery = () => {
...
-
-export function buildMagnetURI(name: string, length: number, infoHash: string): string {
-  return defaultTrackers.reduce(
-    (magnetURI, tracker) => magnetURI + '&tr=' + tracker,
-    `magnet:?xt=urn:btih:${infoHash}&dn=${encodeURIComponent(name)}&xl=${length}`
-  );
-}
-
-export function generateURL({infoHash, length, name}: Torrent): string {
+export function generateURL({infoHash, length, name}: TorrentUI): string {
```
spersico, line 15: this file probably needs a new name, as it generates a URL for a torrent share, it's got nothing to do with a magnet at this point

packages/web3torrent/src/utils/torrent-status-checker.test.ts
```diff
@@ -28,10 +27,10 @@ describe('Torrent Status Checker', () => {
     it('should return a torrent with a status of Idle when the torrent is no longer live', () => {
```
spersico, line 27:
```suggestion
    it('should return a torrent with a status of Idle when the torrent is not live', () => {
```

geoknee:
Unfortunately I don't have time to review this quickly @kerzhner so hopefully @spersico can do so.

Without looking at the code, I am on board with the direction of the PR, judging by the description 👍

kerzhner:
> Unfortunately I don't have time to review this quickly @kerzhner so hopefully @spersico can do so.

Appreciate the quick look! The review can also wait until later if others aren't planning on significant `web3torrent` work today or over the weekend.

packages/web3torrent/src/constants.ts
```diff
@@ -84,6 +100,32 @@ export const preSeededTorrents: Array<Partial<Torrent>> = [
   }
 ];
 
+export const preseededTorrentsUI: TorrentUI[] = preSeededTorrents.map(partialTorrent => ({
+  ...partialTorrent,
+  ...getStaticTorrentUI(partialTorrent.infoHash, partialTorrent.name, partialTorrent.length)
+}));
+
+export function getStaticTorrentUI(
```
spersico, line 108: I don't know if I would put this function here..... it's not exactly a constant, isn't it?

spersico:
Gave it a quick test, everything looks like it's working okay!

packages/web3torrent/src/utils/magnet.ts
kerzhner: Good call. 1ae612f2c

packages/web3torrent/src/constants.ts
kerzhner: When I move `getStaticTorrentUI` into another module, I am running into into a circular dependency issue as `constant.ts` uses `getStaticTorrentUI` (and many other modules use the `constants` module). So I decided to move the function into constants. If you have other suggestions for getting around a circular dependency, I am all ears.

PR #1471, opened 2020-04-17
Simple hub logs to file/console based on LOG_DESTINATION

andrewgordstewart:
As a result, we have a single [`pino.log` artifact](https://19211-209829093-gh.circle-artifacts.com/0/tmp/pino.log) containing the logs from both simple-hub and xstate-wallet in ndjson format.


packages/channel-provider/src/messaging-service.ts
```diff
@@ -88,14 +88,14 @@ export class MessagingService {
     callback?: (result: ResultType) => void
   ) {
     const guid = Guid.create().toString();
-    console.log(`${guid} - CREATE`);
+    log('CREATE', guid);
```
lalexgap, line 91:
```suggestion
```

packages/channel-provider/src/messaging-service.ts
```diff
@@ -88,14 +88,14 @@ export class MessagingService {
...
     const listener = (event: MessageEvent) => {
       if (event.data && event.data.jsonrpc && event.data.result && event.data.id === message.id) {
         if (callback) {
           callback(event.data.result);
         }
         this.acknowledge();
-        console.log(`${guid} - REMOVE`);
+        log('REMOVE', guid);
```
lalexgap, line 98:
```suggestion
        log('REMOVE', guid);
```

packages/channel-provider/src/messaging-service.ts
lalexgap: I added this for debugging and forgot to remove it 🤦

packages/e2e-tests/puppeteer/helpers.ts
```diff
@@ -5,7 +5,7 @@ import * as path from 'path';
 
 const pinoLog =
   process.env.LOG_DESTINATION && process.env.LOG_DESTINATION !== 'console'
-    ? fs.createWriteStream(process.env.LOG_DESTINATION, {flags: 'w'})
+    ? fs.createWriteStream(process.env.LOG_DESTINATION, {flags: 'a'})
```
lalexgap, line 8: Minor but it might be nice to drop a comment about what the `a` flag does

andrewgordstewart: https://github.com/statechannels/monorepo/pull/1475/commits/559e5a0cc1cf4d5ef84e5d517a418a76d6c66d9f

packages/simple-hub/package.json
```diff
@@ -39,7 +39,7 @@
     "lint-staged": "10.0.4",
     "nodemon": "1.19.4",
     "npm-run-all": "4.1.5",
-    "pino-pretty": "3.5.0",
+    "pino-pretty": "^4.0.0",
```
kerzhner, line 42: I believe we aim to have exact dependencies
```suggestion
    "pino-pretty": "4.0.0",
```

packages/simple-hub/src/logger.ts
```diff
@@ -1,10 +1,12 @@
...
-  debug: _logger.debug.bind(_logger),
-  info: _logger.info.bind(_logger),
-  error: _logger.error.bind(_logger),
-  fatal: _logger.fatal.bind(_logger)
-};
+const destination = LOG_TO_FILE ? pino.destination(LOG_DESTINATION) : undefined;
+const prettyPrint = LOG_TO_CONSOLE ? {translateTime: true, color: true} : false;
+
+const opts = {name, prettyPrint};
+export const log = destination ? pino(opts, destination) : pino(opts);
```
kerzhner, line 12: Surprised this works. I was seeing some issues with pino + observables and had to add `this` binding to get around those.

andrewgordstewart: You mean `pino().info` was not working from the context of an observable?

kerzhner: I think that `log.info` was not working where log was defined outside of observable.

PR #1473, opened 2020-04-17
Use pino in web3torrent

andrewgordstewart:
This completes the trifecta: web3torrent, xstate-wallet, and simple-hub all logging to a [single artifact](https://19243-209829093-gh.circle-artifacts.com/0/tmp/pino.log).

Note that some of the web3torrent logs are _very_ large: the longest line in this file is 264010 characters long. I think some torrent/wire classes are being logged containing actual data. We could define pino serializers which eg. serialize a `wire` object to strip data when storing in a file, if the log level is at most `50`.

Closes https://github.com/statechannels/monorepo/issues/1460


packages/web3torrent/src/logger.ts
```diff
@@ -0,0 +1,42 @@
...
+  LOG_TO_FILE && IS_BROWSER_CONTEXT
+    ? {write: o => console.log(JSON.stringify({...o, name}, getCircularReplacer()))}
+    : undefined;
+
+const prettyPrint = LOG_TO_CONSOLE ? {translateTime: true} : false;
+
+const opts = {name, prettyPrint, browser};
+export const logger = destination ? pino(opts, destination) : pino(opts);
+
+// export default prefix => (...x) => logger.info(prefix, ...x);
```
kerzhner, line 42: Is this intentional?

andrewgordstewart, line 42:
```suggestion
```

PR #1476, opened 2020-04-17
fix tracker check, add logs

### LINKED ISSUE #1472: RTCErrorEvent
**Describe the bug :bug:**
A error message is appearing in console. There isn't a consequence that I can see of this error. 
image

**To Reproduce**
Seed and Leech a Torrent (in different tabs) you should see this error in both tabs.

**Additional context**

spersico:
should fix #1472 
To understand the change of this PR, taking a look at the [bittorrent-tracker docs](https://github.com/webtorrent/bittorrent-tracker#client) its advisable.

The previous implementation of the check was a deeper emulation of the web torrent's tracker client behaviour, but it seems that emulating that and suddenly stopping and destroying the tracker client is problematic.

So basically, what this does, is that the [tracker client waits for a couple of seconds](https://github.com/statechannels/monorepo/compare/sp/tracker-check-fix?expand=1#diff-54c06577239cf26700fa05ec63e79884R15), and then scraps the tracker server for information. This way, the tracker server gives us a more accurate response. The implementation gets simpler too.

I also added some logs for the test so it gets easier to see in the CI.

### DIFF
--- a/packages/web3torrent/src/utils/checkTorrentInTracker.ts
+++ b/packages/web3torrent/src/utils/checkTorrentInTracker.ts
@@ -5,7 +5,7 @@ import debug from 'debug';
 const log = debug('web3torrent:tracker-check');
 
 export async function checkTorrentInTracker(infoHash: string) {
-  log(`Scrapping tracker for torrent ${infoHash}`);
+  log(`Scraping tracker for torrent ${infoHash}`);
   const trackerClient: Client = new Client({
     ...defaultTrackerOpts,
     infoHash: [infoHash]
--- a/packages/web3torrent/src/utils/checkTorrentInTracker.ts
+++ b/packages/web3torrent/src/utils/checkTorrentInTracker.ts
@@ -1,33 +1,33 @@
 import {Client} from 'bittorrent-tracker';
 import {defaultTrackerOpts} from '../constants';
 import {TorrentTestResult} from '../library/types';
+import debug from 'debug';
+const log = debug('web3torrent:tracker-check');
 
 export async function checkTorrentInTracker(infoHash: string) {
+  log(`Scrapping tracker for torrent ${infoHash}`);
   const trackerClient: Client = new Client({
     ...defaultTrackerOpts,
     infoHash: [infoHash]
   });
-  let completePeers;
-  const gotAWire: Promise<boolean> = new Promise(resolve => {
-    const updateIfSeederFound = data => {
-      completePeers = data.complete;
-    };
-    trackerClient.once('peer', function() {
-      resolve(true);
-    });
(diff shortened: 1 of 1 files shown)


packages/web3torrent/src/utils/checkTorrentInTracker.ts
```diff
@@ -1,33 +1,33 @@
 import {Client} from 'bittorrent-tracker';
 import {defaultTrackerOpts} from '../constants';
 import {TorrentTestResult} from '../library/types';
+import debug from 'debug';
+const log = debug('web3torrent:tracker-check');
 
 export async function checkTorrentInTracker(infoHash: string) {
+  log(`Scrapping tracker for torrent ${infoHash}`);
```
kerzhner, line 8:
```suggestion
  log(`Scraping tracker for torrent ${infoHash}`);
```

packages/web3torrent/src/utils/checkTorrentInTracker.ts
```diff
@@ -1,33 +1,33 @@
...
-  trackerClient.stop();
-  trackerClient.destroy();
-  if (raceResult) {
-    return TorrentTestResult.SEEDERS_FOUND; // Could connect to a peer
-  }
-  if (Number.isInteger(completePeers)) {
-    // was able to connect to the tracker, but couldn't connect to a peer
-    return TorrentTestResult.NO_SEEDERS_FOUND;
+
+  const raceResult = await Promise.race([trackerScrape, timer]);
```
kerzhner, line 19: Is the race still needed now that scrape is called after 2.5 seconds?

packages/web3torrent/src/utils/checkTorrentInTracker.ts
```diff
@@ -1,33 +1,33 @@
 import {Client} from 'bittorrent-tracker';
 import {defaultTrackerOpts} from '../constants';
 import {TorrentTestResult} from '../library/types';
+import debug from 'debug';
```
kerzhner, line 4: The logging will need to change when https://github.com/statechannels/monorepo/pull/1473/files is merged.

packages/web3torrent/src/utils/checkTorrentInTracker.ts
spersico: umm... haven't thought about it. I'll test it out tomorrow, and if it's not, I'll make a new PR.

packages/web3torrent/src/utils/checkTorrentInTracker.ts
spersico: well... it's not merged.
As this log is not essential at all, I would merge this PR, so that the CI error stops appearing, and later change the logging

PR #1482, opened 2020-04-18
Use Guid to generate message IDs

andrewgordstewart:
Fixes https://github.com/statechannels/monorepo/issues/1456


andrewgordstewart:
Changing the type to string breaks too many things for my taste :P

PR #1483, opened 2020-04-18
Unique request ids

andrewgordstewart:
(Hopefully) fixes https://github.com/statechannels/monorepo/issues/1456

### DIFF
--- a/packages/channel-provider/src/messaging-service.ts
+++ b/packages/channel-provider/src/messaging-service.ts
@@ -49,14 +49,16 @@ export class MessagingService {
     }
   }
 
+  private requestNumber = 0;
   async request<ResultType = any>(
     target: Window,
     message: JsonRpcRequest,
     callback?: (result: ResultType) => void
   ): Promise<ResultType> {
-    if (!message.id) {
-      message.id = Date.now();
-    }
+    // TODO: I don't know the requirements on message IDs, but it's important
+    // that they be unique
+    message.id = this.requestNumber++;
+    message.time = Date.now();
 
     return new Promise<ResultType>((resolve, reject) => {
       window.addEventListener(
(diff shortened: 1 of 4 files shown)


packages/channel-provider/src/types.ts
```diff
@@ -28,13 +28,15 @@ import {
 
 export interface JsonRpcRequest<MethodName = string, RequestParams = any> {
   id?: number;
+  time?: number;
```
lalexgap, line 31: I don't think this follows the official JSON-RPC spec?

Do we need a `time` property?

andrewgordstewart: I don't know. The `request.id` was set as `Date.time()`, so I thought I should keep the time in the request.

I have no reason other than that for keeping it, so I'll remove it.

packages/channel-provider/src/messaging-service.ts
```diff
@@ -49,14 +49,20 @@ export class MessagingService {
...
-    if (!message.id) {
-      message.id = Date.now();
-    }
+    // Some tests rely on being able to supply the id on the message
+    // We should not allow this in production, as we cannot guarantee unique
+    // message ids.
+    if (message.id) console.error('message id should not be defined');
+
+    // TODO: I don't know the requirements on message IDs, but it's important
+    // that they be unique
```
geoknee, line 64: From what I can tell, from the JSON-RPC spec side, this would meet requirements 

> id
An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [1] and Numbers SHOULD NOT contain fractional parts [2]

```suggestion
    // message IDs should be unique
```

PR #1487, opened 2020-04-20
[e2e] Use dappeteer

geoknee:
Latest attempt at integrating metamask into our e2e tests. 

Everything passes locally.

~(save for rps, which needs a tiny bit more work).~

Problem: to test using chrome extensions (such as metamask) we need to run puppeteer head*ful* not head*less*. 

~Our current config on circle does not allow for that~
~TODO: investigate if this is possible~

A new circle config allows for this, but I am now running into other issues on circle -- possibly to do with running out of memory.

Refs
--
https://github.com/marketplace/actions/puppeteer-headful
https://stackoverflow.com/questions/52999911/bitbucket-pipelines-non-headless-puppeteer


geoknee:
This has been neatened / focussed up in #1515

PR #1490, opened 2020-04-20
W3T torrent check refactor

spersico:
This PR adds support for the Pino logs that @andrewgordstewart implemented last week, and reactors the checkTorrentInTracker function so that it becomes more readable.

I tried removing the entire logic behind the previous _timer_ promise, but as the tracker client is event-based, the clean up after running the function because there's a lot of possible results and you end up repeating code. 
I think that it's better to clean the function and just remove the promise race while maintaining the timer logic.

### DIFF
--- a/packages/web3torrent/src/utils/checkTorrentInTracker.ts
+++ b/packages/web3torrent/src/utils/checkTorrentInTracker.ts
@@ -7,18 +7,15 @@ const log = logger.child({module: 'tracker-check'});
 export async function checkTorrentInTracker(infoHash: string) {
   log.info(`Scraping tracker for torrent ${infoHash}`);
 
-  const trackerClient: Client = new Client({
-    ...defaultTrackerOpts,
-    infoHash: [infoHash]
-  });
+  const client = new Client({...defaultTrackerOpts, infoHash: [infoHash]});
   const trackerScrape: Promise<number> = new Promise(resolve => {
-    trackerClient.once('scrape', data => resolve(data.complete));
-    setTimeout(() => trackerClient.scrape(), 2500); // waits ~2 seconds as that's the tracker update tick
+    client.once('scrape', data => resolve(data.complete));
+    setTimeout(() => client.scrape(), 2500); // waits ~2 seconds as that's the tracker update tick
     setTimeout(() => resolve(-1), 5000); // returns -1 if there was no answer from the tracker(s) for ~2.5 seconds
   });
   const scrapeResult = await trackerScrape;
-  trackerClient.stop();
-  trackerClient.destroy();
+  client.stop();
+  client.destroy();
 
   if (scrapeResult > 0) {
     log.info(`Test Result: SEEDERS_FOUND (${scrapeResult})`);
--- a/packages/web3torrent/src/utils/checkTorrentInTracker.ts
+++ b/packages/web3torrent/src/utils/checkTorrentInTracker.ts
@@ -1,33 +1,32 @@
 import {Client} from 'bittorrent-tracker';
 import {defaultTrackerOpts} from '../constants';
 import {TorrentTestResult} from '../library/types';
-import debug from 'debug';
-const log = debug('web3torrent:tracker-check');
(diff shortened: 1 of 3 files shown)


packages/web3torrent/src/utils/checkTorrentInTracker.ts
```diff
@@ -1,33 +1,29 @@
 import {Client} from 'bittorrent-tracker';
```
kerzhner, line 1: Should we rename this file so that it follow the kebab-case convention? For example, `check-torrent-in-tracker.ts`?

packages/web3torrent/src/utils/checkTorrentInTracker.ts
```diff
@@ -1,33 +1,29 @@
...
-  });
+  log.info(`Scraping tracker for torrent ${infoHash}`);
+
+  const client = new Client({...defaultTrackerOpts, infoHash: [infoHash]});
   const trackerScrape: Promise<number> = new Promise(resolve => {
-    trackerClient.once('scrape', data => resolve(data.complete));
-    setTimeout(() => trackerClient.scrape(), 2500); // waits ~2 seconds as that's the tracker update tick
+    client.once('scrape', data => resolve(data.complete));
+    setTimeout(() => client.scrape(), 2500); // waits ~2 seconds as that's the tracker update tick
+    setTimeout(() => resolve(-1), 5000); // returns -1 if there was no answer from the tracker(s) for ~2.5 seconds
```
kerzhner, line 14: should we use clearTimeout in case this callback it loses the race?

packages/web3torrent/src/utils/checkTorrentInTracker.ts
```diff
@@ -1,33 +1,29 @@
 import {Client} from 'bittorrent-tracker';
 import {defaultTrackerOpts} from '../constants';
 import {TorrentTestResult} from '../library/types';
-import debug from 'debug';
-const log = debug('web3torrent:tracker-check');
+import {logger} from '../logger';
+const log = logger.child({module: 'tracker-check'});
 
 export async function checkTorrentInTracker(infoHash: string) {
```
kerzhner, line 7: Should we add a return type to this function?
```suggestion
export async function checkTorrentInTracker(infoHash: string): Promise<TorrentTestResult> {
```

packages/web3torrent/src/utils/checkTorrentInTracker.ts
spersico: oh, good catch! will do

packages/web3torrent/src/utils/checkTorrentInTracker.ts
spersico: Are you talking about the first setTimeout or the second?
For the second, I don't think it's necessary, there's no side-effect there, and once a promise resolves, all subsequent tries to resolve simply don't have any effect.

If you are talking about the first one (`setTimeout(() => client.scrape(), 2500);`) I think it's impossible for the other setTimeout to run before this one. And the possible side-effects are handled after with the `client.stop()` and `client.destroy()`

packages/web3torrent/src/utils/checkTorrentInTracker.ts
spersico: https://github.com/statechannels/monorepo/pull/1490/commits/3952c25531b950f53d5f9031024f668171ad4b2c

packages/web3torrent/src/utils/checkTorrentInTracker.ts
spersico: https://github.com/statechannels/monorepo/pull/1490/commits/3952c25531b950f53d5f9031024f668171ad4b2c

packages/web3torrent/src/utils/checkTorrentInTracker.ts
kerzhner: I think your reasoning is solid. Good thinking!

PR #1491, opened 2020-04-20
Dump store to logs in circle

andrewgordstewart:
Aiming to reproduce https://github.com/statechannels/monorepo/issues/1478, I wrote code to dump the store on command.

While I attempt to [reproduce the issue on circle](https://github.com/statechannels/monorepo/commit/7d109642599dd77f8aeab3a946ff53ea7f4dc5a2) (I have been unable to locally) I think this code will be useful in `master`.

```
❯ cat seed-download.e2e.log | grep '"store":' | jq .store.ledgers
{}
{
  "firebase:simple-hub": "0x3d59eb841d26f1e7c5be21c1043bc3902b6b991581c22982772f2511c419b274"
}
{}
{
  "firebase:simple-hub": "0x71ea10595d1bbee25a7cb4c3b2cdf89752cb9ad0e47a86df22050e617256d32c"
}
```

[This artifact](https://19680-209829093-gh.circle-artifacts.com/0/tmp/seed-download.pino.log) is still under 1MB, so it does not seem to generate very large logs

### DIFF
--- a/packages/client-api-schema/src/generated-schema.json
+++ b/packages/client-api-schema/src/generated-schema.json
@@ -705,13 +705,6 @@
       ],
       "type": "object"
     },
-    "ErrorCode": {
-      "enum": [
-        200,
-        100
-      ],
-      "type": "number"
-    },
     "ErrorResponse": {
       "anyOf": [
         {
--- a/packages/xstate-wallet/src/channel-wallet.ts
+++ b/packages/xstate-wallet/src/channel-wallet.ts
@@ -111,7 +111,7 @@ export class ChannelWallet {
         this.getWorkflow(this.calculateWorkflowId(request)).service.send(request);
         break;
       case 'APPROVE_BUDGET_AND_FUND': {
-        const workflow = this.startWorkflow(
+        this.startWorkflow(
           ApproveBudgetAndFund.machine(this.store, this.messagingService, {
             player: request.player,
             hub: request.hub,
@@ -122,7 +122,6 @@ export class ChannelWallet {
           true // devtools
         );
 
-        workflow.service.send(request);
         break;
       }
       case 'CLOSE_AND_WITHDRAW': {
--- a/packages/xstate-wallet/src/channel-wallet.ts
+++ b/packages/xstate-wallet/src/channel-wallet.ts
@@ -14,7 +14,7 @@ import {ApproveBudgetAndFund, CloseLedgerAndWithdraw, Application} from './workf
 import {ethereumEnableWorkflow} from './workflows/ethereum-enable';
 import {AppRequestEvent} from './event-types';
 import {serializeChannelEntry} from './serde/app-messages/serialize';
-import {ADD_LOGS} from './constants';
+import {ADD_LOGS, IS_PRODUCTION} from './constants';
 import {logger} from './logger';
 
 export interface Workflow {
@@ -152,6 +152,9 @@ export class ChannelWallet {
     const service = interpret(machine, {devTools})
       .onTransition((state, event) => ADD_LOGS && logTransition(state, event, workflowId))
       .onDone(() => (this.workflows = this.workflows.filter(w => w.id !== workflowId)))
+      .onStop(async () => {
+        if (!IS_PRODUCTION) logger.info({store: await this.store.dumpBackend()});
+      })
       .start();
     // TODO: Figure out how to resolve rendering priorities
     this.renderUI(service);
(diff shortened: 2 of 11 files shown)


packages/client-api-schema/src/generated-schema.json
```diff
@@ -705,13 +705,6 @@
       ],
       "type": "object"
     },
-    "ErrorCode": {
-      "enum": [
-        200,
-        100
-      ],
-      "type": "number"
-    },
```
snario, line 714: Why is this here?

packages/xstate-wallet/src/channel-wallet.ts
```diff
@@ -146,13 +145,23 @@ export class ChannelWallet {
...
+    const service = interpret(machine, {devTools})
       .onTransition((state, event) => ADD_LOGS && logTransition(state, event, workflowId))
+      .onTransition(
+        async () =>
+          IS_PRODUCTION &&
+          logger.info({workflowId, store: await this.store.dumpBackend()}, 'Done workflow')
+      )
       .onDone(() => (this.workflows = this.workflows.filter(w => w.id !== workflowId)))
+      .onStop(
+        async () =>
```
snario, line 161: Does this need to be `async`?

packages/xstate-wallet/src/store/indexedDB-backend.ts
```diff
@@ -17,7 +17,7 @@ enum ObjectStores {
 
 // A running, functioning example can be seen and played with here: https://codesandbox.io/s/elastic-kare-m1jp8
 export class IndexedDBBackend implements DBBackend {
-  private _db: any;
+  private _db: IDBDatabase;
```
snario, line 20: Where is this type defined?

packages/xstate-wallet/src/store/indexedDB-backend.ts
```diff
@@ -61,11 +62,40 @@ export class IndexedDBBackend implements DBBackend {
       request.onerror = err => reject(err);
       request.onsuccess = () => {
         this._db = request.result;
+
         resolve(request.result);
       };
     });
   }
 
+  public async dump(): Promise<object> {
```
snario, line 71: Is `dump` the right name for this? Maybe `serialize` or `toJSON` or something?

packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -121,6 +122,7 @@ const generateConfig = (
   actions: WorkflowActions,
   guards: WorkflowGuards
 ): MachineConfig<WorkflowContext, WorkflowStateSchema, WorkflowEvent> => ({
+  strict: !IS_PRODUCTION,
```
snario, line 125: Why?

packages/client-api-schema/src/generated-schema.json
andrewgordstewart: I needed to run `yarn prepare` to update something (I think?), and it resulted in this change: https://github.com/statechannels/monorepo/pull/1491/commits/e47d753dae92d898c1a4585c215cdb4c3bcaec3a

packages/xstate-wallet/src/channel-wallet.ts
andrewgordstewart: Not anymore!
```suggestion
        () =>
```

packages/xstate-wallet/src/store/indexedDB-backend.ts
andrewgordstewart: I would guess in the indexedDB type package?

packages/xstate-wallet/src/workflows/application.ts
andrewgordstewart: I was attempting to be able to log the store dump only on error: https://github.com/statechannels/monorepo/pull/1491/commits/8b7d3df4b01072235eb7b918e1ddcbf57de20ed1

Eventually, I decided to log the store on every transition, because it doesn't not really pollute the logs.

Perhaps it should just be set to `strict: true`, but I am not sure if we want that.

packages/xstate-wallet/src/store/indexedDB-backend.ts
andrewgordstewart: Perhaps, but it is not something I put a lot of thought into.

The purpose of this is to log the store contents for debugging purposes. So, `dump` may seem more appropriate than `serialize`, considering its usage in programming.

I'm not fussed either way.

andrewgordstewart:
Closed, as the indexedDB backend is now removed.

PR #1493, opened 2020-04-20
setApplicationSite acquires lock before updating store

andrewgordstewart:
(Hopefully) fixes https://github.com/statechannels/monorepo/issues/1478


snario:
Were you able to confirm if it fixes that issue?

andrewgordstewart:
> Were you able to confirm if it fixes that issue?

As I'm unable to reproduce the issue, I can't confirm it. I am pretty confident that `setApplicationSite` _can_ cause this issue, and that this change removes that possibility. But there's no guarantee that there's no other call which would overwrite the application site without acquiring a lock.

PR #1497, opened 2020-04-21
[xtstate-wallet] Add component for challenging UI

snario:


packages/xstate-wallet/src/ui/selectors.ts
```diff
@@ -31,6 +31,12 @@ export function getConfirmCreateChannelService(
   return applicationWorkflowState.children.invokeCreateChannelConfirmation as any;
 }
 
+export function getChallengeChannelService(
+  applicationWorkflowState: AppWorkflowState
+): Interpreter<any> {
+  return applicationWorkflowState.children.invokeChallengingProtocol as any;
```
snario, line 37: This value is `undefined` when testing on Storybook, and I'm unsure why. I'm trying to fetch to internal state / context of the `ChallengeMachine`.

packages/xstate-wallet/src/ui/application-workflow.tsx
```diff
@@ -44,7 +46,7 @@ export const ApplicationWorkflow = (props: Props) => {
       <Heading textAlign="center" mb={0}>
         {messages[getApplicationStateValue(current)]}
       </Heading>
-      {!isConfirmCreateChannel(current) && (
+      {!isConfirmCreateChannel(current) && !isApplicationChallenging(current) && (
```
andrewgordstewart, line 49: Should `isConfirmCreateChannel` ever return `true` when `isApplicationChallenging` returns `true`?

What seems to be going on here is:
- the UI is always in exactly one "display mode"
- we have `isApplicationX` boolean functions defined in `./selectors'`
- so, we should probably have a `applicationDisplayMode` which returns an enum/union type
- we display the UI based on the return value of `applicationDisplayMode`

PR #1501, opened 2020-04-21
Introduce SC_ENV

kerzhner:
This PR addresses the often asked question of: How do I set up my local environment variables to get multiple packages to work together?

The workflow is:
1. `.env.custom-env` is created by the developer who is creating the reproducible scenario. The file is added to monorepo root with every environment variable needed for every package developers are going to run. The file can be checked into version control.
1. The developer who is trying to reproduce a scenario runs every package with `SC_ENV=custom-env` prepended to the command. `SC` stands for state channels.

When `SC_ENV` is defined, only one environment file is read in. No hierarchy and no local overrides are accepted. Included in this PR is the `.env.virtual-funding` that has all of the environment variables needed for the e2e virtual funding test.

Closes https://github.com/statechannels/monorepo/issues/1488.


packages/xstate-wallet/.env
```diff
@@ -1,12 +1,3 @@
-
```
kerzhner, line 1: It appears that below are no longer used.

tomclose:
Sounds good to me. Is it possible we'll need a single override `.env.custom-env.local`, in case there are sensitive variables that shouldn't be checked into the repo?

kerzhner:
> Sounds good to me. Is it possible we'll need a single override `.env.custom-env.local`, in case there are sensitive variables that shouldn't be checked into the repo?

Thanks for the suggestion! Per discussion during standup, we'll introduce this capability when a use case comes up.

kerzhner:
Closing since https://github.com/statechannels/monorepo/pull/1507 was merged.

PR #1502, opened 2020-04-21
[xstate-wallet] fix: Return window.ethereum.selectedAddress for selectedAddress

Alonski:
There were cases where ethereumIsEnabled returned true
However this.mySelectedAddress was null.
This makes sure that if ethereum is enabled it will return the correct address


packages/xstate-wallet/src/messaging.ts
```diff
@@ -164,9 +164,7 @@ export class MessagingService implements MessagingServiceInterface {
       case 'GetWalletInformation':
         await this.sendResponse(requestId, {
           signingAddress: await this.store.getAddress(),
-          selectedAddress: this.store.chain.ethereumIsEnabled
-            ? this.store.chain.selectedAddress
-            : null,
+          selectedAddress: this.store.chain.selectedAddress,
```
andrewgordstewart, line 167:
```suggestion
          selectedAddress: this.store.chain.selectedAddress ?? null,
```

packages/xstate-wallet/src/chain.ts
```diff
@@ -273,7 +273,8 @@ export class ChainWatcher implements Chain {
   }
 
   public get selectedAddress(): string | null {
-    return this.mySelectedAddress || null;
+    this.mySelectedAddress = this.mySelectedAddress || window.ethereum.selectedAddress;
```
geoknee, line 276: What if `window.ethereum` is not defined?

geoknee: if `this.mySelectedAddress` is `undefined`, this has the same effect as L247 above (only without a check about `window.ethereum existing, see preceding comment).

This makes me wonder if this changeset only fixes the issue you describe because this getter is called some time *after* L247. So the fix may be brittle.

The underlying issue seems to be that, if ethereum is enabled from metamask's point of view, the `window.ethereum` object does not immediately update to reflect this (even if waiting for a `window.ethereum.enable()` promise to resolve).

Alonski: @geoknee From what we saw together when we paired. Part of the issue was that we were setting `mySelectedAddress` before it was defined.
With this way if its undefined it will try to get it from `window.ethereum`

I could change to this:
```ts
this.mySelectedAddress = this.mySelectedAddress || window.ethereum?.selectedAddress ?? null;
```
What do you think?

Alonski: Also maybe if `window.ethereum` isn't defined we should just throw and show a modal anyways?

geoknee: If `window.ethereum` is not defined: we already use the Rimble Connection Banner to alert the user that the app may not work for them in this case.

Alonski: Ok. This happens only in RPS or is the Wallet responsible for it?
I will update the code to not throw though and return null instead

PR #1505, opened 2020-04-21
Update production env vars for web3torrent

snario:


snario:
After this is merged, we should be able to delete hard-coded env vars on the Netlify dashboard: https://app.netlify.com/sites/sc-web3torrent/settings/deploys#environment

Issue #1433, opened 2020-04-15
Investigate why the wait for the hub takes so long


lalexgap:
The wallet seems to wait a long time before noticing the hub's deposit.

We should look into why that is and if we can do anything about it.

kerzhner:
In some cases, a wallet never notices a client deposit no matter how long I wait. Seems to happen fairly frequently with the leecher in `web3torrent`.

lalexgap:
Initially it seems like this might be related to `metamask`. We don't see any of these issues from the integration tests when using a direct `web3` provider.

 Manually forcing a query to the blockchain before the `Deposit` returns a `ChainView` that does not reflect the deposit. It seems like `metaMask` itself might be slow to update itself against the lastest chain?

lalexgap:
So far I've identified two pieces of this so far:

- `ethers.js` polling: This plays a relatively small part in the delay (The default polling interval is 4 seconds), but we can probably enforce a faster polling period when `chainUpdatedFeed` is called to easily alleviate this.
`metamask` caching/polling. It looks like `metamask` only polls the ganache instance every 20 seconds(you can see this by inspecting the background.js view of the metamask extension). This would explain the drastic difference between integration tests and manual testing.  Not sure if there is a way to mitigate this, will investigate further.

Based on these two factors the worst case is a wait of ~24 seconds before the `wallet` gets the latest chain info.

lalexgap:
> In some cases, a wallet never notices a client deposit no matter how long I wait. Seems to happen fairly frequently with the leecher in `web3torrent`.

Are you still running into this @kerzhner ?  The delay can be up to ~24 seconds but I'm seeing the `deposit` event eventually (when I wait :P ) Is this something that happens quite frequently?

kerzhner:
Thanks for investigating this! I am beginning to think that I may be misattributing fault. It is possible that, when I don't see the deposit event, I am running packages with incorrect environment variables.

The reason I believe this is that I have not seen the missing deposit event scenario since I started using `.env.virtual-funding` in the monorepo root.

lalexgap:
Closing as it with the change to port `8545` `metamask` no longer caches results.

PR #1506, opened 2020-04-21
Enable faster polling in ethers.js

lalexgap:
Based off my investigation in #1433.

`ethers` default polling time is 4 seconds which can be a bit long. This sets it to 50 ms when the `chainUpdatedFeed ` is subscribed to.

Most of #1433 is caused by metamask 20 second polling time so this only helps slightly.


lalexgap:
> This seems good to me. My only questions are around the 50ms interval you chose for the "fast" setting. It wouldn't make much sense to poll at an interval significantly lower than the typical transaction confirmation interval: so I am wondering whether we should only use this fast setting when targeting ganache?

I think we're only really concerned about `ganache` but I don't think there's too much harm doing it against an actual blockchain for now.

Longer term we could set `ETHERS_REGULAR_POLLING_INTERVAL`, `ETHERS_QUICK_POLLING_INTERVAL` to appropriate values for different networks but I'm not sure it's worth the effort now.

PR #1507, opened 2020-04-22
Environments for integration tests

kerzhner:
https://github.com/statechannels/monorepo/pull/1501 precedes this PR.

Concentrate all environment variables for web3torrent integration tests in corresponding files. The benefits are:
- Easier to understand all environment variable values that are set for the tests.
- Makes it simpler to run the integration tests locally.
- Makes it simpler to play around with variables to try out different values.


.circleci/config.yml
```diff
@@ -242,21 +242,9 @@ jobs:
...
       - image: circleci/node:10.16.3
     environment:
-      # For multiple
-      NODE_ENV: test
-      # For shared-ganache
-      USE_GANACHE_DEPLOYMENT_CACHE: true
-      GANACHE_CACHE_FOLDER: ../../.ganache-deployments
-      GANACHE_PORT: 8547
-      # For simple-hub
-      FIREBASE_PREFIX: << pipeline.git.branch >>-1
```
snario, line 252: I think this would need to be also set for `simple-hub`

snario:
Also we could change the format of the files to `VAR_NAME = 'value'` like was set in #1508

.circleci/config.yml
kerzhner: I dont think this integration test uses the hub.

kerzhner:
> Also we could change the format of the files to `VAR_NAME = 'value'` like was set in #1508

Done in de0afa7a0

PR #1508, opened 2020-04-22
Standardize env var file format + add VS code extension

snario:


packages/devtools/.env.development
```diff
@@ -1,7 +1,6 @@
+GANACHE_CACHE_FOLDER = '../../.ganache-deployments'
```
kerzhner, line 1: Can no longer use `source` to load `.env` files when there are spaces around `= `

snario: Do we ever use `source`?

kerzhner: Good question. We have in the past. Not sure we do at this point? If others do not use source, I am ok losing that functionality if this PR gives us a benefit like syntax highlighting.

kerzhner: There is also an argument for readability (regardless of syntax highlighting) with this PR.

packages/nitro-protocol/.env
```diff
@@ -1,20 +1,20 @@
...
+SINGLE_ASSET_PAYMENT_ADDRESS = ' 0x0000000000000000000000000000000000000000'
+CONSENSUS_APP_ADDRESS = ' 0x0000000000000000000000000000000000000000'
+TRIVIAL_APP_ADDRESS = ' 0x0000000000000000000000000000000000000000'
+TEST_FORCE_MOVE_ADDRESS = ' 0x0000000000000000000000000000000000000000'
+TEST_NITRO_ADJUDICATOR_ADDRESS = ' 0x0000000000000000000000000000000000000000'
+TEST_TOKEN_ADDRESS = ' 0x0000000000000000000000000000000000000000'
+TEST_ETH_ASSET_HOLDER_ADDRESS = ' 0x0000000000000000000000000000000000000000'
+TEST_TOKEN_ASSET_HOLDER_ADDRESS = ' 0x0000000000000000000000000000000000000000'
+TEST_ASSET_HOLDER_ADDRESS  = ' 0x0000000000000000000000000000000000000000'
+TEST_ASSET_HOLDER2_ADDRESS = ' 0x0000000000000000000000000000000000000000'
```
geoknee, line 20:
```suggestion
TEST_ASSET_HOLDER2_ADDRESS = '0x0000000000000000000000000000000000000000'
```

etc? Nitpicking, for sure: but this PR is about standardising the format.

PR #1510, opened 2020-04-22
[nitro-protocol] Take zeppelin dependencies out of beta

snario:


snario:
Seems like they depend on solidity at version 0.6.2. We can table this for now then, and wait until we upgrade Solidity.

PR #1513, opened 2020-04-22
Remove embedded wallet

snario:


davidangel:
Hi there, I stumbled on this while searching for good examples of how to approach `netlify-ignore.sh`...

I noticed it was removed in this PR. Curious if it was working okay, though. Thanks!

snario:
Might be good to ask @geoknee who wrote that

geoknee:
@davidangel this PR removed an entire package, hence the change to `netlify-ignore.sh`. We have also been transferring responsibilities from netlify to circle-ci, so we aren't in need of `netlify-ignore.sh` any more. But it was working pretty well when we needed it to.

PR #1514, opened 2020-04-22
Remove netlify-ignore script

snario:


geoknee:
Why not remove `netlify-ignore.sh` and `netlify.toml` while you are at it?

snario:
Strange, I thought I had.

PR #1516, opened 2020-04-22
Try to get puppeteer test to run in headful in CI

geoknee:
trying GH actions first, might revert to circle later.


geoknee:
Progress toward #1549 .

geoknee:
We should pursue a solution via #1548 instead.

PR #1521, opened 2020-04-22
Add ReplaySubject to ChannelClient, use it on PaymentChannelClient

spersico:
The title is fairly straightforward.

This PR adds a subscribable ReplaySubject on the channelClient. This subject emits the last state to new subscribers, ensuring that no matter when the subscriber subscribes, it never loses the latest state of the channel.

See [here](https://www.learnrxjs.io/learn-rxjs/subjects) for more data about the ReplaySubject.

### DIFF
--- a/packages/channel-client/src/channel-client.ts
+++ b/packages/channel-client/src/channel-client.ts
@@ -36,9 +36,7 @@ export class ChannelClient implements ChannelClientInterface {
 
   constructor(readonly provider: ChannelProviderInterface) {
     this.channelState = new ReplaySubject(1);
-    this.provider.on('ChannelUpdated', result => {
-      this.channelState.next(result);
-    });
+    this.provider.on('ChannelUpdated', result => this.channelState.next(result));
   }
 
   onMessageQueued(
--- a/packages/channel-client/src/channel-client.ts
+++ b/packages/channel-client/src/channel-client.ts
@@ -16,10 +16,12 @@ import {
 } from '@statechannels/client-api-schema';
 import {HUB} from './constants';
 import {ETH_TOKEN_ADDRESS} from '../tests/constants';
+import {ReplaySubject} from 'rxjs';
 
 type TokenAllocations = Allocation[];
 
 export class ChannelClient implements ChannelClientInterface {
+  channelState: ReplaySubject<ChannelResult>;
   get signingAddress(): string | undefined {
     return this.provider.signingAddress;
   }
@@ -32,7 +34,12 @@ export class ChannelClient implements ChannelClientInterface {
     return this.provider.walletVersion;
   }
 
-  constructor(readonly provider: ChannelProviderInterface) {}
+  constructor(readonly provider: ChannelProviderInterface) {
+    this.channelState = new ReplaySubject(1);
+    this.provider.on('ChannelUpdated', result => {
+      this.channelState.next(result);
+    });
+  }
 
   onMessageQueued(
     callback: (result: MessageQueuedNotification['params']) => void
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -261,11 +261,11 @@ export class PaymentChannelClient {
 
       this.channelClient.channelState
         .pipe(
-          map((result: ChannelResult) => convertToChannelState(result)),
+          map(convertToChannelState),
           takeWhile((latestChannelState: ChannelState) => !readyToPay(latestChannelState), true),
           filter(readyToPay)
         )
-        .subscribe(channelState => resolve(channelState));
+        .subscribe(resolve);
     });
 
     const {payerBalance} = channelState;
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -16,6 +16,7 @@ import {AddressZero} from 'ethers/constants';
 import * as firebase from 'firebase/app';
 import 'firebase/database';
 import _ from 'lodash';
+import {takeWhile, map, filter, take} from 'rxjs/operators';
 import {logger} from '../logger';
 const log = logger.child({module: 'payment-channel-client'});
 const hexZeroPad = utils.hexZeroPad;
@@ -72,7 +73,7 @@ export class PaymentChannelClient {
   }
 
   constructor(private readonly channelClient: ChannelClientInterface) {
-    this.channelClient.onChannelUpdated(channelResult => {
+    this.channelClient.channelState.subscribe(channelResult => {
       this.updateChannelCache(convertToChannelState(channelResult));
     });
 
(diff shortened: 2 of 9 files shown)


packages/web3torrent/src/pages/file/File.test.tsx
```diff
@@ -13,7 +13,8 @@ import WebTorrentPaidStreamingClient from '../../library/web3torrent-lib';
 
 Enzyme.configure({adapter: new Adapter()});
 
-jest.mock('@statechannels/channel-client');
+// this mock was making the entire test fail, and doesn't looks to be used anywhere
+// jest.mock('@statechannels/channel-client');
```
kerzhner, line 17: Should we delete these lines?

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -258,15 +259,13 @@ export class PaymentChannelClient {
...
-
-      const unsubscribeListener = this.channelClient.onChannelUpdated(cu => {
-        if (readyToPay(convertToChannelState(cu))) {
-          unsubscribeListener();
-          resolve(convertToChannelState(cu));
-        }
-      });
+      this.channelClient.channelState
+        .pipe(
+          map((result: ChannelResult) => convertToChannelState(result)),
```
kerzhner, line 264:
```suggestion
          map(convertToChannelState)
          ,
```

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -258,15 +259,13 @@ export class PaymentChannelClient {
...
+          takeWhile((latestChannelState: ChannelState) => !readyToPay(latestChannelState), true),
+          filter(readyToPay)
+        )
+        .subscribe(channelState => resolve(channelState));
```
kerzhner, line 268:
```suggestion
        .subscribe(resolve);
```

packages/channel-client/src/channel-client.ts
```diff
@@ -32,7 +34,12 @@ export class ChannelClient implements ChannelClientInterface {
     return this.provider.walletVersion;
   }
 
-  constructor(readonly provider: ChannelProviderInterface) {}
+  constructor(readonly provider: ChannelProviderInterface) {
+    this.channelState = new ReplaySubject(1);
```
kerzhner, line 38: Are you sure that a ReplaySubject is needed? Is the latest event delivered to an observer on subscription without a ReplaySubject?

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -258,15 +259,13 @@ export class PaymentChannelClient {
...
-      const unsubscribeListener = this.channelClient.onChannelUpdated(cu => {
-        if (readyToPay(convertToChannelState(cu))) {
-          unsubscribeListener();
-          resolve(convertToChannelState(cu));
-        }
-      });
+      this.channelClient.channelState
+        .pipe(
+          map((result: ChannelResult) => convertToChannelState(result)),
+          takeWhile((latestChannelState: ChannelState) => !readyToPay(latestChannelState), true),
```
kerzhner, line 265: `takeUntil` might be more clear

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -258,15 +259,13 @@ export class PaymentChannelClient {
...
+          filter(readyToPay)
```
kerzhner, line 266: [takeLast](https://rxjs-dev.firebaseapp.com/api/operators/takeLast) might be more clear

packages/channel-client/src/channel-client.ts
```diff
@@ -32,7 +34,12 @@ export class ChannelClient implements ChannelClientInterface {
...
+    this.provider.on('ChannelUpdated', result => {
+      this.channelState.next(result);
+    });
```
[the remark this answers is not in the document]
spersico: It's not the same, that makes the tests stop working, so I changed it to:
```    
this.provider.on('ChannelUpdated', result => this.channelState.next(result));
```

packages/channel-client/src/channel-client.ts
spersico: > Is the latest event delivered to an observer on subscription without a ReplaySubject?

It's not, that's why I chose to use ReplaySubject instead of a normal Subject. It ensures that it will always emit the last state of the channel.

packages/channel-client/src/channel-client.ts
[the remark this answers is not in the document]
kerzhner: Hmm I wonder why my suggestion is not working. It seems a bit weird to write a one parameter function that, when invoked, calls another one parameter function with the exact same parameter. Seems like unnecessary verboseness given first-class functions.

Do you think it's an issue with `this`?

packages/channel-client/src/channel-client.ts
kerzhner: Yep, and I wonder if a `Subject` is equivalent to `ReplaySubject(1)`. I might be wrong though

packages/web3torrent/src/clients/payment-channel-client.ts
spersico: well... its not the same. takeUntil expects a number of events to complete the subscription. I didn't want to assume that either the first, second or Nth event would be the one that would close the subscription.

Talking with George, we refactored this code to something better

packages/web3torrent/src/clients/payment-channel-client.ts
spersico: take a look at https://github.com/statechannels/monorepo/pull/1521/commits/b31af613ca8f8d85e206bea2c401cc792a19a5c1

packages/channel-client/src/channel-client.ts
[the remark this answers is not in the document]
andrewgordstewart: I think it is an issue with `this`. I have tried similar things in the past.

packages/channel-client/src/channel-client.ts
andrewgordstewart: I think `ReplaySubject(1)` is equivalent to `BehaviourSubject`, but not `Subject`

packages/web3torrent/src/clients/payment-channel-client.ts
kerzhner: Nice, looks good!

packages/web3torrent/src/clients/payment-channel-client.ts
kerzhner: Thanks!

packages/channel-client/src/channel-client.ts
spersico: > Yep, and I wonder if a `Subject` is equivalent to `ReplaySubject(1)`. I might be wrong though

It's not, it doesn't save the last emitted value.: https://jsbin.com/yekaqov/edit?js,console


> I think `ReplaySubject(1)` is equivalent to `BehaviourSubject`, but not `Subject`

When the replaySubject constructor param is 1, is equivalent to BehaviourSubject, but it doesn't need an initial value.

packages/web3torrent/src/pages/file/File.test.tsx
spersico: sure

PR #1523, opened 2020-04-22
Add Dexie backend

andrewgordstewart:
Adds a Dexie backend which has feature parity with the indexedDB backend.

This backend should be updated to use proper Dexie features, like indexes, transactions, etc. This PR takes the approach that we can do this incrementally with a working dexie backend.

### DIFF
--- a/packages/xstate-wallet/src/index.ts
+++ b/packages/xstate-wallet/src/index.ts
@@ -1,7 +1,6 @@
 import {ChannelWallet} from './channel-wallet';
 import {MessagingService} from './messaging';
 import {ChainWatcher} from './chain';
-import {IndexedDBBackend} from './store/indexedDB-backend';
 import {MemoryBackend} from './store/memory-backend';
 import {XstateStore} from './store';
 import * as constants from './constants';
@@ -9,12 +8,13 @@ import Url from 'url-parse';
 import './render';
 
 import {logger} from './logger';
+import {Backend} from './store/dexie-backend';
 const log = logger.info.bind(logger);
 
 (async function() {
   const chain = new ChainWatcher();
 
-  const backend = constants.USE_INDEXED_DB ? new IndexedDBBackend() : new MemoryBackend();
+  const backend = constants.USE_INDEXED_DB ? new Backend() : new MemoryBackend();
   const store = new XstateStore(chain, backend);
 
   await store.initialize([], constants.CLEAR_STORAGE_ON_START);
--- a/packages/xstate-wallet/src/store/dexie-backend.ts
+++ b/packages/xstate-wallet/src/store/dexie-backend.ts
@@ -161,24 +161,6 @@ export class Backend implements DBBackend {
     return this.put(ObjectStores.objectives, value, Number(key)) as Promise<Objective>;
   }
 
-  /**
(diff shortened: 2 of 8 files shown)


packages/xstate-wallet/src/store/dexie-backend.ts
```diff
@@ -0,0 +1,203 @@
...
+  }
+  public async getLedger(key: string) {
+    return this.get(ObjectStores.ledgers, key);
+  }
+
+  // Individual Setters
+
+  public async setPrivateKey(signingAddress: string, privateKey: string) {
+    const pksPutted = await this.put(ObjectStores.privateKeys, privateKey, signingAddress);
+    return pksPutted;
```
spersico, line 144:
```suggestion
    return this.put(ObjectStores.privateKeys, privateKey, signingAddress);
```

packages/xstate-wallet/src/store/dexie-backend.ts
```diff
@@ -0,0 +1,203 @@
...
+  public async setChannel(key: string, value: ChannelStoredData) {
+    return this.put(ObjectStores.channels, value, key);
+  }
+  public async setLedger(key: string, value: string) {
+    return this.put(ObjectStores.ledgers, value, key);
+  }
+  public async setNonce(key: string, value: BigNumber) {
+    await this.put(ObjectStores.nonces, value.toString(), key);
+
+    return await this._db[ObjectStores.nonces].get(key);
```
spersico, line 155: I think this is no longer necessary because in [here](https://github.com/statechannels/monorepo/pull/1523/files#diff-5b0ab37ac1fdf8ed411d298edc8c8e0fR190) we already return the saved value

packages/xstate-wallet/src/index.ts
```diff
@@ -1,20 +1,20 @@
...
 
 import {logger} from './logger';
+import {Backend} from './store/dexie-backend';
 const log = logger.info.bind(logger);
 
 (async function() {
   const chain = new ChainWatcher();
 
-  const backend = constants.USE_INDEXED_DB ? new IndexedDBBackend() : new MemoryBackend();
+  const backend = constants.USE_INDEXED_DB ? new Backend() : new MemoryBackend();
```
lalexgap, line 17: This is unrelated to the PR but I think we should retire `MemoryBackend` soon and just rely on `CLEAR_STORAGE_ON_START` if we want to start fresh on refresh.

packages/xstate-wallet/src/store/types.ts
```diff
@@ -149,6 +149,8 @@ export type ChannelStoredData = {
 };
 export interface DBBackend {
   initialize(cleanSlate?: boolean): Promise<any>;
+  // TODO: Perhaps the backend API should look more like this?
+  // privateKeys(): Promise<Record<string, {signingAddress: string; privateKey: string} | undefined>>;
```
lalexgap, line 153: 👍

packages/xstate-wallet/src/index.ts
andrewgordstewart: It's not clear how easy a task that is: https://github.com/statechannels/monorepo/issues/1526

PR #1528, opened 2020-04-23
WIP: Budget UI

lalexgap:


packages/web3torrent/src/hooks/use-budget.ts
```diff
@@ -0,0 +1,48 @@
+import {web3torrent} from '../clients/web3torrent-client';
+import {INITIAL_BUDGET_AMOUNT} from '../constants';
+import {useState, useEffect} from 'react';
+import {SiteBudget} from '@statechannels/client-api-schema';
+import {useChannelClient} from './use-web3-torrent-client';
+
+export function useBudget() {
```
lalexgap, line 7: This is a new pattern I'm looking at for accessing functionality of the clients. Instead of referencing properties off of the context we instead create hooks when needed to access different parts of the context.



We could also use something like [constate](https://github.com/diegohaz/constate) if we wanted with this pattern (but I'm not sure if I see the benefit of using the context over custom hooks)

packages/web3torrent/src/App.tsx
```diff
@@ -9,12 +9,21 @@ import File from './pages/file/File';
...
 import {requiredNetwork} from './constants';
+import {Budgets} from './pages/budgets/Budgets';
+import {InitializationContext} from './contexts/initialization-context';
 
 const App: React.FC = () => {
   const [currentNetwork, setCurrentNetwork] = useState(
     window.ethereum ? Number(window.ethereum.networkVersion) : undefined
   );
 
+  const {initialize, isInitialized} = useContext(InitializationContext);
```
lalexgap, line 20: This isn't strictly necessary since `initialize` will be called whenever another context is used, but it's nice to start loading the `iFrame` as soon as we hit the page.

packages/web3torrent/src/contexts/omni-provider.tsx
```diff
@@ -0,0 +1,21 @@
...
+import {InitializationContext, useInitializationContext} from './initialization-context';
+import React from 'react';
+import {ChannelContext, useChannelContext} from './channel-context';
+import {TorrentContext, useTorrentContext} from './torrent-context';
+
+export const OmniProvider: React.FC = ({children}) => {
+  const initializationContext = useInitializationContext();
+  // TODO: Probably a more elegant way of doing this
+  return (
+    <InitializationContext.Provider value={initializationContext}>
```
lalexgap, line 11: The idea behind constructing a `InitializationContext` and passing it to other contexts is that other contexts can initialize the "wallet" if needed.

This means that if the UI calls `useContext(BudgetContext)` the budget context can use the initialize context to intialize the "wallet" so the UI doesn't have to care about whether `initialize` has been called or not.

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -81,17 +79,17 @@ export class PaymentChannelClient {
     });
   }
 
+  async initialize() {
```
lalexgap, line 82: I'm not sure if there's a better naming convention. Right now it's slightly confusing that there's an `initialize` and an `enable` on the payment channel client.

packages/web3torrent/src/contexts/budget-context.ts
```diff
@@ -0,0 +1,47 @@
...
+    setFetching(false);
+  };
+  const closeBudget = async () => {
+    setFetching(true);
+    await paymentChannelClient.closeAndWithdraw();
+
+    setFetching(false);
+  };
+
+  return {budget: paymentChannelClient.budgetCache, loading, createBudget, closeBudget};
```
lalexgap, line 46: Right now we're just connecting `budget` to the `paymentChannelClient.budgetCache` but if we wanted more advanced state management we can do that in this context. We could use `useState` or `useReducer` to store the `budget`

packages/web3torrent/src/contexts/initialization-context.ts
```diff
@@ -0,0 +1,26 @@
...
+import {useState} from 'react';
+import React from 'react';
+
+export const InitializationContext = React.createContext<
+  ReturnType<typeof useInitializationContext>
+>(undefined);
+export function useInitializationContext() {
+  const {paymentChannelClient} = web3torrent;
+
+  const [isInitialized, setInitialized] = useState(false);
```
kerzhner, line 11: Should this be one state with 3 possible values (like an enum)? The values can be 'notInitialized', 'initializing', and ' initialized'.

packages/web3torrent/src/contexts/torrent-context.ts
```diff
@@ -0,0 +1,23 @@
+import {download, upload, web3torrent} from '../clients/web3torrent-client';
+import {useEffect} from 'react';
+
+import React from 'react';
+import {getTorrentUI} from '../utils/torrent-status-checker';
+
+export const TorrentContext = React.createContext<ReturnType<typeof useTorrentContext>>(undefined);
+
+export function useTorrentContext({initializationContext}) {
+  const {initialize, isInitialized} = initializationContext;
```
kerzhner, line 10: Question. Is the intention here to use a React context? Or just to use the object value passed in? I think that without the `useContext` hook, this function just uses the value of the object passid in.

lalexgap: Originally I was hoping to use the context normally like `const {initialize, isInitialized} = useContext(InitializationContext)`. The problem is that since the `useTorrentContext` hook is being called in the context of setting a value for `TorrentContext.Provider` react hasn't wired up `InitializationContext` yet, so `useContext(InitializationContext)` returns `undefined`.

To work around this I just pass the actual context object in the props. It's slightly hacky but it got things working 😛

kerzhner:
I likely don't understand something, but I have a question about the InitializationContext:
1. `OmniProvider` hold on to state of whether the `paymentChannelClient` has been initialized. Let's call this state `init`. `init` starts as `false`.
1. `App` component rendering triggers an effect that triggers the start of initialization.
1. When initialization is completed, `init` state is set to true. This triggers the entire component tree to re-render.
1. It strikes me that components do not need to trigger `initialize()`. They just need to check `init` state and maybe short-circuit some logic.

Thoughts?

lalexgap:
> I likely don't understand something, but I have a question about the InitializationContext:
> 
> 1. `OmniProvider` hold on to state of whether the `paymentChannelClient` has been initialized. Let's call this state `init`. `init` starts as `false`.
> 2. `App` component rendering triggers an effect that triggers the start of initialization.
> 3. When initialization is completed, `init` state is set to true. This triggers the entire component tree to re-render.
> 4. It strikes me that components do not need to trigger `initialize()`. They just need to check `init` state and maybe short-circuit some logic.

I think that you're right, components should never call `initialize`. One thing I didn't really think of is that this all occurring at startup when the initial context is set up (not when the context is actually accessed by the Components).

I guess this means that the Components loses the ability to determine when to perform `initialize` but I think that's fine.

packages/web3torrent/src/contexts/omni-provider.tsx
lalexgap: My comment is slightly misleading as it makes it sound that `initialize` will occur when the context is accessed. However the `initialize` will be called at startup (see https://github.com/statechannels/monorepo/pull/1528#issuecomment-618591176)

packages/web3torrent/src/contexts/omni-provider.tsx
```diff
@@ -0,0 +1,21 @@
...
+      <TorrentContext.Provider value={useTorrentContext({initializationContext})}>
+        <ChannelContext.Provider value={useChannelContext({initializationContext})}>
+          <BudgetContext.Provider value={useBudgetContext({initializationContext})}>
```
kerzhner, line 14: Is it true that the values of these providers will be updated only when the `OmniProvider` is re-rendered? If so, is the main purpose of finegrain providers to wire finegrain data to the component that needs the data?

My understanding is that when `OmniProvider` component is update, the entire React tree needs to be re-rendered. So I am not sure that wiring finegrain data to components is a performance gain. Though I am not 100% sure about that.

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -1,46 +1,50 @@
...
   paymentChannelClient: paymentChannelClient,
   tracker: {announce: defaultTrackers}
 });
 
-export const Web3TorrentContext = React.createContext(web3torrent);
-
 export const getTorrentPeers = infoHash => web3torrent.peersList[infoHash];
 
-export const download: (torrent: WebTorrentAddInput) => Promise<ExtendedTorrent> = torrentData => {
+const doesBudgetExist = async (): Promise<boolean> => {
```
kerzhner, line 15: Nit, might be cleaner with `function` syntax.

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -1,46 +1,50 @@
...
+  const budget = await web3torrent.paymentChannelClient.getBudget();
+  return !!budget && !_.isEmpty(budget);
+};
+
+export const download: (
```
kerzhner, line 20: Nit, might be cleaner with `function` syntax.

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -1,46 +1,50 @@
...
 });
 
-export const Web3TorrentContext = React.createContext(web3torrent);
-
 export const getTorrentPeers = infoHash => web3torrent.peersList[infoHash];
 
-export const download: (torrent: WebTorrentAddInput) => Promise<ExtendedTorrent> = torrentData => {
+const doesBudgetExist = async (): Promise<boolean> => {
+  const budget = await web3torrent.paymentChannelClient.getBudget();
+  return !!budget && !_.isEmpty(budget);
```
kerzhner, line 17: Never really understood the reasons to use `!!`... 🤷

packages/web3torrent/src/components/site-budget-table/SiteBudgetTable.tsx
```diff
@@ -1,79 +1,64 @@
...
+        </thead>
+        <tbody>
+          <tr className="budget-info">
+            <td className="budget-button">
+              <button id="budget-withdraw" onClick={closeBudget}>
+                Withdraw
+              </button>
+            </td>
+            <td className="budget-number">
+              {' '}
```
kerzhner, line 52: Are these needed? For me, sometimes, prettier (or maybe eslint?) insert these for reasons I don't understand.

packages/web3torrent/src/contexts/budget-context.ts
```diff
@@ -0,0 +1,47 @@
...
+import {useState, useEffect} from 'react';
+
+import React from 'react';
+
+export const BudgetContext = React.createContext<ReturnType<typeof useBudgetContext>>(undefined);
+
+export function useBudgetContext({initializationContext}) {
+  const {paymentChannelClient} = web3torrent;
+
+  const [loading, setFetching] = useState(true);
```
kerzhner, line 12: I think it might be more conventional
```suggestion
  const [loading, setLoading] = useState(true);
```

kerzhner:
Thanks for putting this work together! I am done with my review of the work as is here. More than happy to take another look if it evolves.

packages/web3torrent/src/contexts/budget-context.ts
lalexgap: Yup, just forgot to rename the setter.

packages/web3torrent/src/contexts/omni-provider.tsx
lalexgap: > Is it true that the `value` properties of these providers will be updated only when the `OmniProvider` is re-rendered? If so, is the main purpose of finegrain providers to wire finegrain data to the component that needs the data?

I think it's kind of the opposite way around. Whenever the `value` changes it forces a rerender. My understanding is that `react` is smart enough to only render components that are affected by a context change.

From [react docs](https://reactjs.org/docs/context.html#contextprovider):
 >All consumers that are descendants of a Provider will re-render whenever the Provider’s value prop changes"

A consumer is a component calling the `useContext` hook or the old style
```
<MyContext.Consumer>
  {value => /* render something based on the context value */}
</MyContext.Consumer>
```

In our case a consumer is something like  `useContext(BudgetContext)`. `react` is smart enough( if my understanding is correct) to rerender whatever components call `useContext(BudgetContext)`

So by having narrower contexts you would only cause rerenders in specific components.

All that being said, our react app is pretty simple (from a UI perspective) so I don't think the performance aspect really matters much. (And I haven't really bothered to look at exactly when things are being rerendered)

My preference for narrower contexts is that it seems more natural in the components to access a narrow piece of data, instead of having to navigate some complex context object.

kerzhner: I agree with all of your points, and your understand of React matches mine. I think it is true that `BudgetContext.Provider` `value` property is only updated when `OmniProvider` is updated.

Also, I think [this caveat](https://reactjs.org/docs/context.html#caveats) is something to be aware of. For us, I think this means that the `value` property will be updated every time `OmniProvider` is re-rendered.

kerzhner: I gave this some more thought, and I think the following will happen:
1. Something triggers a state update for `OmniProvider` component. This can be one of the children components calling a `setState` function provided by a contexts.
1. The OmniProvider is re-rendered. During re-rendering, every `value` property for every context is recomputed.
1. All context consumers are re-rendered due to `value` updates.

Not a big deal given that the application is fairly small. But I think the implication is that every component that is a context consumer is re-rendered every time `OmniProvider` component is re-rendered (which happens any time finegrain state-setter is called).

packages/web3torrent/src/contexts/omni-provider.tsx
```diff
@@ -0,0 +1,21 @@
...
+import React from 'react';
+import {ChannelContext, useChannelContext} from './channel-context';
+import {TorrentClientContext, useTorrentClientContext} from './torrent-context';
+
+export const OmniProvider: React.FC = ({children}) => {
+  const web3TorrentClientContext = useWeb3TorrentClientContext();
+  // TODO: Probably a more elegant way of doing this
+  return (
+    <Web3TorrentClientContext.Provider value={web3TorrentClientContext}>
+      <TorrentClientContext.Provider value={useTorrentClientContext({web3TorrentClientContext})}>
```
kerzhner, line 12: To be honest, I am little surprised that `useState` and `useEffect` work inside a component property computation...

packages/web3torrent/src/contexts/omni-provider.tsx
lalexgap: Good point! I hadn't really considered that `OmniProvider` owns all the various states of the custom hooks it's using, which forces a rerender when changed.

I think we could avoid this by not custom hooks to construct a context . I'm going to take a second look at if we even need to write the context as custom hooks.

lalexgap: > I think we could avoid this by not custom hooks to construct a context . I'm going to take a second look at if we even need to write the context as custom hooks.

So I've spent some time looking at just doing this with classes for the context and it seems more awkward than using these custom hooks.

kerzhner: I have tossing ideas around my head and trying to distill the essence of this code. So I think it is:
1. We would like to define state at the top level of the component tree instead of leaves.
1. We would like the ability to pass this state and corresponding state setters down the leaves without painful prop chaining.
1. We would like this state to be more finegrain as opposed to a large global state object. 
  1. This is done to limit the scope of knowledge for a component. 
  1. And the hope with this was to optimize re-rending.

Generic contexts give us #1 and #2. Seems like 3b will be tricky to get. Then I guess it all comes down to how strongly we feel about 3a.

packages/web3torrent/src/context/channel-context.ts
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
junjizhi, line 1: Too many context classes. Don't see any async calls here. Is this class necessary? 😛

packages/web3torrent/src/context/client-initialization-context.ts
```diff
@@ -0,0 +1,37 @@
...
+    'Not Initialized'
+  );
+
+  const initialize = () => {
+    if (initializationStatus === 'Not Initialized') {
+      setInitializationStatus('Initializing');
+
+      w3tClient.paymentChannelClient.initialize().then(() => {
+        setInitializationStatus('Initialized');
+      });
```
junjizhi, line 32: This is too intrusive to initialize a member variable of an object for it. 

I think we can do the following:

- Rename `w3tClient.enable()` to `w3tClient.initialize()`
- Call `w3tClient.initialize()` here
- Make sure when we pass in `paymentChannelClient` to build `w3tClient`, `paymentChannelClient` has already been initialized. 
- Add a check to `w3tClient.initialize()` to make sure `paymentChannelClient` is intialized.

What do you think?

packages/web3torrent/src/context/torrent-client-context.ts
```diff
@@ -0,0 +1,83 @@
...
+      w3tClient.add(torrentData, (torrent: any) => resolve({...torrent, status: Status.Connecting}))
+    );
+  };
+
+  const upload: (input: WebTorrentSeedInput) => Promise<ExtendedTorrent> = async input => {
+    await w3tClient.enable();
+    // TODO: This only checks if a budget exists, not if we have enough funds in it
+    if (!(await doesBudgetExist())) {
+      await budgetContext.createBudget();
+    }
```
junjizhi, line 53: Can we put the logic in the budgetContext?

junjizhi:
image

@lalexgap I was able to run this branch locally and see the budget UI. Nice work!

What's left for this PR? Is it mainly refactoring work?

lalexgap:
> image
> 
> @lalexgap I was able to run this branch locally and see the budget UI. Nice work!

> What's left for this PR? Is it mainly refactoring work?

@junjizhi Right now I'm in the process of reverting some of my context changes. 

Initially I had hoped to simplify state management, but the end result still seems pretty complex and there's still some things not working.

 Considering that is essentially a demo application I don't think it's worth spending too much time on state management (as long as things work)

So my plan is to revert most of these changes and solve the initial problem I ran into by simply passing something in via `props`. Hoping to have this done soon🤞

lalexgap:
Closing as most of the context work has been reverted and a new PR has been created.

PR #1533, opened 2020-04-23
W3T Restart Fixing

### LINKED ISSUE #1448: The download button dissapears after cancelling the download.
**Describe the bug :bug:**
I found a bug, where if I press the cancel button, the cancel button disappears, and it doesn't appear again.
The download seems to be correctly cancelled, but the button doesn't reappear. 
**To Reproduce**
Steps to reproduce the behaviour:
1. Upload a torrent
2. Download a torrent (from another tab)
3. Cancel the download

**Expected behaviour**
The button to download should reappear.

**Screenshots (if applicable)**
After cancelling:
image

**Environment:**
- Tested on master

spersico:
closes #1448 
The button now it's shown.

Needs #1754 to be merged first (so that tests pass)

### DIFF
--- a/packages/web3torrent/src/clients/web3torrent-client.ts
+++ b/packages/web3torrent/src/clients/web3torrent-client.ts
@@ -51,13 +51,14 @@ export const upload: (input: WebTorrentSeedInput) => Promise<ExtendedTorrent> = 
   );
 };
 
-export const cancel = (id: string = '') => {
+export const cancel = (torrentId: string = '') => {
   return new Promise((resolve, reject) =>
-    web3torrent.cancel(id, err => {
+    web3torrent.remove(torrentId, err => {
       if (err) {
         reject(err);
       } else {
-        resolve(id);
+        // TODO: clean up channel info
+        resolve(torrentId);
       }
     })
   );
--- a/packages/web3torrent/src/clients/web3torrent-client.ts
+++ b/packages/web3torrent/src/clients/web3torrent-client.ts
@@ -53,7 +53,7 @@ export const upload: (input: WebTorrentSeedInput) => Promise<ExtendedTorrent> = 
 
 export const cancel = (torrentId: string = '') => {
   return new Promise((resolve, reject) =>
-    web3torrent.remove(torrentId, err => {
+    web3torrent.cancel(torrentId, err => {
       if (err) {
         reject(err);
       } else {
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -112,7 +112,8 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     log.info('> Peer cancels download. Pausing torrents');
     const torrent = this.torrents.find(t => t.infoHash === torrentInfoHash);
     if (torrent) {
-      torrent.pause();
+      await this.closeDownloadingChannels(torrent);
+      torrent.destroy();
     }
   }
 
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -108,12 +108,14 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     return torrent;
   }
 
-  async cancel(torrentInfoHash: string, callback?: (err: Error | string) => void) {
+  async cancel(infoHash: string, callback?: (err: Error | string) => void) {
     log.info('> Peer cancels download. Pausing torrents');
-    const torrent = this.torrents.find(t => t.infoHash === torrentInfoHash);
+    const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
       await this.closeDownloadingChannels(torrent);
-      torrent.destroy();
+      this.remove(infoHash, callback);
+    } else {
+      callback(new Error('No torrent found'));
     }
   }
 
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -395,6 +395,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
   protected async makePayment(torrent: PaidStreamingTorrent, wire: PaidStreamingWire) {
     if (torrent.paused) {
       log.info('Torrent Paused - makePayment early exit', torrent);
+      return;
     }
     const {peerChannelId, peerAccount} = wire.paidStreamingExtension;
     let amountToPay = BUFFER_REFILL_RATE.sub(
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -108,15 +108,25 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     return torrent;
   }
 
-  async cancel(infoHash: string, callback?: (err: Error | string) => void) {
-    log.info('> Peer cancels download. Pausing torrents');
+  // TODO: refactor "pause" and "cancel" functions. It's an ugly mess right
+  pause(infoHash: string, callback?: (err?: Error | string) => void) {
+    log.info('> Peer pauses download: Pause torrent, eventual close PaymentChannels');
+    const torrent = this.torrents.find(t => t.infoHash === infoHash);
+    if (torrent) {
+      torrent.pause(); // the paymentChannelClosing is done at the moment of payment
+    } else {
+      return callback(new Error('No torrent found'));
+    }
+  }
+
+  async cancel(infoHash: string, callback?: (err?: Error | string) => void) {
+    log.info('> Peer cancels download. Remove Torrent, close PaymentChannels');
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
       await this.closeDownloadingChannels(torrent);
-      torrent.pause();
-      // pausing does not close the wires, it only prevents the torrent to search new wires.
+      torrent.destroy(callback);
     } else {
-      callback(new Error('No torrent found'));
+      return callback(new Error('No torrent found'));
     }
   }
 
@@ -177,12 +187,12 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       PaidStreamingExtensionEvents.REQUEST,
       async (index: number, size: number, response: (allow: boolean) => void) => {
         const reqPrice = bigNumberify(size).mul(WEI_PER_BYTE);
-        const {peerAccount: peer, peerChannelId} = wire.paidStreamingExtension;
+        const {pseChannelId, peerAccount: peer} = wire.paidStreamingExtension;
         const knownPeer = this.peersList[torrent.infoHash][peer];
 
-        if (!knownPeer || !peerChannelId) {
+        if (!knownPeer || !pseChannelId) {
           await this.createPaymentChannel(torrent, wire);
-          log.info(`${peer} >> REQUEST BLOCKED (NEW PEER): ${index}`);
+          log.info(`${peer} >> REQUEST BLOCKED (NEW WIRE): ${index}`);
           response(false);
           this.blockPeer(torrent.infoHash, wire, peer);
         } else if (!knownPeer.allowed || reqPrice.gt(knownPeer.buffer)) {
@@ -202,7 +212,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
           };
 
           const {buffer, uploaded} = this.peersList[torrent.infoHash][peer];
-          log.info(`${peer} >> REQUEST ALLOWED: ${index} BUFFER: ${buffer} UPLOAD: ${uploaded}`);
+          log.info(`${peer} >> REQUEST ALLOWED: ${index} BUFFER: ${buffer} UPLOADED: ${uploaded}`);
           response(true);
         }
       }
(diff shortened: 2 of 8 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -105,11 +105,14 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+  async cancel(infoHash: string, callback?: (err: Error | string) => void) {
     log.info('> Peer cancels download. Pausing torrents');
-    const torrent = this.torrents.find(t => t.infoHash === torrentInfoHash);
+    const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
-      torrent.pause();
+      await this.closeDownloadingChannels(torrent);
+      this.remove(infoHash, callback);
+    } else {
+      callback(new Error('No torrent found'));
```
kerzhner, line 115: Is the expectation that the callback should handle this error? I note that the error is not thrown.

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -51,13 +51,14 @@ export const upload: (input: WebTorrentSeedInput) => Promise<ExtendedTorrent> =
   );
 };
 
-export const cancel = (id: string = '') => {
+export const cancel = (torrentId: string = '') => {
```
kerzhner, line 54: Note that when the `DownloadInfo` react component calls this function, there is wait or callback for the action to complete. Is that the intention?

spersico: this function call is useless, I've got a bunch of things that I want to refactor about this function.
The `web3torrent.cancel()` function is `async` (both in this PR and in master), so this function having a callback is only used to show errors (and using a callback for that is useless in this case).
I think that initially was the function was set up this way because the `webtorrent.cancel` has a callback, but right now it doesn't make any sense as it is.

My plan is to refactor this `cancel()` function, the new `pause()` and the `resumeTorrent()` ones, on another PR.
I want to merge this PR, it's getting kind of old and large, and I want to make sure that cancelling works consistently, at least.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: https://github.com/statechannels/monorepo/pull/1533#discussion_r421271946

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -252,10 +237,16 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-      ) {
+      const {pseChannelId, peerChannelId, peerAccount} = wire.paidStreamingExtension;
+
+      if (channelState.channelId === pseChannelId || channelState.channelId === peerChannelId) {
+        if (channelState.status === 'closed' || channelState.status === 'closing') {
+          wire.paidStreamingExtension.pseChannelId = null;
+          wire.paidStreamingExtension.peerChannelId = null;
+          log.info(
+            `PeerAccount ${peerAccount} - ChannelId ${channelState.channelId} Channel Closed||Closing`
+          );
```
spersico, line 248: this does most of the work.
This updates the thing that gets checked [here](https://github.com/statechannels/monorepo/pull/1533/files#diff-a84153ed9f3ab2bd13c41bca43b584c7R190).
That triggers a new channel creation if the leecher makes new requests.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -266,18 +257,41 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+      beneficiaryBalance: '0', // (wei)
+      allowed: false,
+      channelId, // TODO: remove this prop, it isn't as trustworthy as the wire one, and it's just repeating data
+      uploaded: 0
+    };
+
+    log.info(`${peerAccount} >> Channel Created - ID: ${channelId}`);
+    return channelId;
+  }
+
```
spersico, line 294: I moved all this to another function because it was getting hard to read

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -325,7 +339,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
           if (torrent.paused) {
             // We currently treat pausing torrent as canceling downloads
             await this.closeDownloadingChannels(torrent);
-          } else if (!torrent.done) {
+          } else if (!torrent.done || !torrent.destroyed) {
```
spersico, line 342: this fixes a race condition that came up a couple of times, where just after cancelling the download, the leecher tried to make a new payment (it threw an error because the channel was already closed by that point, but I wanted to avoid that)

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -346,19 +360,37 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       await this.closeDownloadingChannels(torrent);
     });
 
-    torrent.on(TorrentEvents.ERROR, err => {
-      log.error({err, torrent}, 'Torrent ERROR: ');
-      this.emit(ClientEvents.TORRENT_ERROR, {torrent, err});
+    torrent.on(TorrentEvents.ERROR, error => {
+      log.error('Torrent ERROR: ', torrent, error);
+      if (error && (error as Error).message.includes('Cannot add duplicate torrent')) {
+        this.resumeTorrent(torrent.infoHash); // horrible way of doing things, will do until we make pause/resume
```
spersico, line 366: this is really, really messy.
I promise that I will fix this one in another PR. The idea with this PR was to make an initial approach to pausing too. To see what could be happening, and make it work semi-okay

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -346,19 +360,37 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
 
+  // TODO: make this publicly accesible and not in an error handler. (when we get a pause/resume button)
+  // TODO: make this work okay (takes a long time to restart)
+  protected resumeTorrent(infoHash: string) {
+    const torrent = super.get(infoHash);
+    if (torrent && torrent.paused) {
+      torrent.resume();
+    } else {
+      throw new Error('Invalid infoHash of torrent to resume: ' + infoHash);
+    }
```
spersico, line 384: same than https://github.com/statechannels/monorepo/pull/1533/files#r421276349

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -397,8 +429,15 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
    */
   protected async closeDownloadingChannels(torrent: PaidStreamingTorrent) {
     torrent.wires.forEach(async wire => {
-      if (wire.paidStreamingExtension && wire.paidStreamingExtension.peerChannelId) {
-        await this.paymentChannelClient.closeChannel(wire.paidStreamingExtension.peerChannelId);
+      const {peerChannelId} = wire.paidStreamingExtension;
+      if (peerChannelId) {
+        log.info(`About to close channel ${peerChannelId}`);
+        wire.paidStreamingExtension.peerChannelId = null;
```
spersico, line 435: with this one, we clean up the downloading channels when pausing, to ensure that the peer won't be able to pay if it tries.
It also removes an unnecessary check and adds a couple of extra logs that I think are usefull

packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -111,7 +111,7 @@ const File: React.FC<Props> = props => {
           withdraw={closeBudget}
         />
       )}
-      {torrent.status === Status.Idle && (
+      {(torrent.status === Status.Idle || torrent.status === Status.Paused) && (
```
spersico, line 114: this is what fixes the download button not appearing

spersico:
## the apparent reason for the E2E tests failing
image
(although I don't think that the seed download test it's supposed to cancel the seeding)

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -170,61 +184,32 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-            peerAccount, // leecher
-            hexZeroPad(INITIAL_SEEDER_BALANCE.toHexString(), 32), // seederBalance,
-            hexZeroPad(WEI_PER_BYTE.mul(torrent.length).toHexString(), 32), // leecherBalance,
-            this.paymentChannelClient.myEthereumSelectedAddress, // seederOutcomeAddress,
-            peerOutcomeAddress // leecherOutcomeAddress
-          );
-          // eslint-disable-next-line require-atomic-updates
-          wire.paidStreamingExtension.pseChannelId = channelId;
+        const {pseChannelId, peerAccount: peer} = wire.paidStreamingExtension;
+        const knownPeer = this.peersList[torrent.infoHash][peer];
```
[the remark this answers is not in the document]
spersico: It's not possible, because by the time this event can happen, this one https://github.com/statechannels/monorepo/blob/d30fb32bfee3463198aa88d2252c5f47ebf5e261/packages/web3torrent/src/library/web3torrent-lib.ts#L329 must happen first

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -346,19 +360,38 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       await this.closeDownloadingChannels(torrent);
     });
 
-    torrent.on(TorrentEvents.ERROR, err => {
-      log.error({err, torrent}, 'Torrent ERROR: ');
-      this.emit(ClientEvents.TORRENT_ERROR, {torrent, err});
+    torrent.on(TorrentEvents.ERROR, error => {
+      log.error('Torrent ERROR: ', torrent, error);
+      if (error && (error as Error).message.includes('Cannot add duplicate torrent')) {
```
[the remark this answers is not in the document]
spersico: yeah, this is a ~horrible~ temporal thing to do, to lay the groundwork for pausing. This was the easiest way to avoid having more buttons to pause and resume.
The idea is to merge this and immediately after start working on the pause button. But ensuring that you (and the e2e tests, and whatever else) can start using and looking at this.
I'll see if adding this the codebase tomorrow in a better approach is easy to do so that I don't even leave this.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -252,10 +237,16 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     });
 
     this.paymentChannelClient.onChannelUpdated(async (channelState: ChannelState) => {
-      if (
-        channelState.channelId === wire.paidStreamingExtension.pseChannelId ||
-        channelState.channelId === wire.paidStreamingExtension.peerChannelId
-      ) {
+      const {pseChannelId, peerChannelId, peerAccount} = wire.paidStreamingExtension;
```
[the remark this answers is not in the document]
spersico: **In a 2 peer situation, with peer A and peer B:**
Peer A `peerChannelId`, is peer B's `pseChannelId` and vice-versa.
And each peer's `peerChannelId` is the id of the channel to pay **the other peer** for data.

So, each time that peer A needs to pay peer B, peer A, uses the peerChannelId of the wire.

The pseChannelId is just for to have a mirror of the channel id that the other peer has.

----------------------------
For example, in https://github.com/statechannels/monorepo/blob/d30fb32bfee3463198aa88d2252c5f47ebf5e261/packages/web3torrent/src/library/web3torrent-lib.ts#L190
the seeder checks if the leecher has a `pseChannelId` (meaning, if the leecher has an active channel to pay), and if not, he creates a payment channel (remember that the peer that receives the data requests is the one responsible for creating payment channels in web3torrent).

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -346,19 +360,38 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       await this.closeDownloadingChannels(torrent);
     });
 
-    torrent.on(TorrentEvents.ERROR, err => {
-      log.error({err, torrent}, 'Torrent ERROR: ');
-      this.emit(ClientEvents.TORRENT_ERROR, {torrent, err});
+    torrent.on(TorrentEvents.ERROR, error => {
+      log.error('Torrent ERROR: ', torrent, error);
```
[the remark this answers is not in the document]
spersico: oh that's why he put the errors like that

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -170,61 +184,32 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+        if (!knownPeer || !pseChannelId) {
+          await this.createPaymentChannel(torrent, wire);
+          log.info(`${peer} >> REQUEST BLOCKED (NEW WIRE): ${index}`);
           response(false);
-          this.blockPeer(torrent.infoHash, wire, peerAccount);
-        } else if (!knownPeerAccount.allowed || reqPrice.gt(knownPeerAccount.buffer)) {
-          const {uploaded} = this.peersList[torrent.infoHash][peerAccount];
-          log.info(`${peerAccount} >> REQUEST BLOCKED: ${index} UPLOADED: ${uploaded}`);
+          this.blockPeer(torrent.infoHash, wire, peer);
+        } else if (!knownPeer.allowed || reqPrice.gt(knownPeer.buffer)) {
```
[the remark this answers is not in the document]
spersico: The buffer, when refilled is converted from wei to bytes: https://github.com/statechannels/monorepo/blob/7f76ba8a11d83ab2b4e28179a5e24e0b2d57f39e/packages/web3torrent/src/library/web3torrent-lib.ts#L319-L322) 

So, we compare bytes to bytes.

PR #1547, opened 2020-04-24
Downgrade solc to 0.6.1

kerzhner:


kerzhner:
A short term fix for the failing master CI introduced via https://github.com/statechannels/monorepo/pull/1545

PR #1558, opened 2020-04-24
Only sign one state per turn number

andrewgordstewart:


packages/xstate-wallet/src/store/channel-store-entry.ts
```diff
@@ -255,7 +255,7 @@ export class ChannelStoreEntry {
     }
 
     const {signedStates} = this;
-    const turnNums = _.map(signedStates, s => s.turnNum.toHexString());
+    const turnNums = _.map(signedStates, s => parseInt(s.turnNum.toHexString(), 16));
```
andrewgordstewart, line 258: :nit: Does this work with turn numbers bigger than 53 bits?

Perhaps, while we're still using `BigNumber` for turn numbers (I think we should _stop_ doing that), we should change `isReverseSorted` to accept `BigNumber[]`, and work correctly.

packages/xstate-wallet/src/workflows/support-state.ts
```diff
@@ -16,57 +17,50 @@ Do we abort? Or do we try to reach consensus on a later state?
...
+  const {isSupportedByMe} = entry;
+  // We only sign the state if we haven't signed it already
+  if (!isSupportedByMe || !statesEqual(entry.latestSignedByMe, state)) {
     await store.signAndAddState(channelId, state);
-  } else {
-    throw 'Not safe to send';
+  } else if (statesEqual(entry.latestSignedByMe, state)) {
+    // The support state machine was started with a state that we already support
+    // That's fine but we output a warning in case that's unexpected
+    warn({state}, 'The state is already supported');
```
andrewgordstewart, line 50: 1. I'm not 100% sure what to do here. Maybe you should send it again (what if it's supported in my store but not yours?)

2. What if it's not safe to send? We should probably be throwing, so that the parent machine catches (an explicit error) and (potentially) retries with a different state

packages/xstate-wallet/src/store/channel-store-entry.ts
andrewgordstewart, line 258: Will this work with proper `BigNumber`s? While we're using `BigNumber` for `turnNum` (I think we should _stop_ doing that) perhaps we should just set `isReverseSorted` to accept `BigNumber[]`

packages/xstate-wallet/src/workflows/support-state.ts
andrewgordstewart, line 50: 1. I'm not 100% sure what to do here. Maybe you should send it again (what if it's supported in my store but not yours?)

2. What if it's not safe to send? We should probably be throwing, so that the parent machine catches (an explicit error) and (potentially) retries with a different state

lalexgap: > 1. I'm not 100% sure what to do here. Maybe you should send it again (what if it's supported in my store but not yours?)

I think `support-state` shouldn't have to worry about sending the state. We can just rely on the `store`/`messaging` to worry about syncing states and `support-state` can just deal with whether a state is signed or not.

> 2. What if it's not safe to send? We should probably be throwing, so that the parent machine catches (an explicit error) and (potentially) retries with a different state

I think the `stale state` check will catch cases where we're attempting to support  a stale state. Are there other checks we should be doing?

packages/xstate-wallet/src/store/channel-store-entry.ts
snario: It seems to work with `ethers.constants.MaxUint256`:
```node
> parseInt(MaxUint256.toHexString(), 16) < parseInt(MaxUint256.sub(1).toHexString(), 16)
false
```

packages/xstate-wallet/src/store/channel-store-entry.ts
```diff
@@ -229,14 +239,38 @@ export class ChannelStoreEntry {
 
     this.clearOldStates();
 
+    this.checkInvariants();
+
     return this.state(entry);
   }
 
+  private checkInvariants() {
```
snario, line 247: Should we call this in the constructor? I didn't notice that we had tests explicitly defining the order and that caught me up when I was trying to figure out what was wrong.

PR #1564, opened 2020-04-27
Use dynamic firebase prefix

snario:
#1553 changed the code to use a fixed prefix, but the dynamic prefix was there to prevent two jobs running in parallel from conflicting with each other. 

Actually as I type this I realize that, if we're using `firebase-server` (which we are?) then this shouldn't matter... so maybe this PR is irrelevant.


andrewgordstewart:
Is this needed, considering https://github.com/statechannels/monorepo/pull/1565/files?

snario:
I don't think it is needed at all since we use a local firebase. But I realized it at the very last moment. I'll just close it.

PR #1572, opened 2020-04-27
Add drift chatbox on web3torrent

snario:
Adds a chatbox using Drift. Let's users send a message. At the moment it just pings me on Slack, but we can figure out a way to share with the whole team.


packages/web3torrent/src/index.tsx
```diff
@@ -1,11 +1,14 @@
...
 import ReactDOM from 'react-dom';
+import Drift from 'react-driftjs';
+
 import App from './App';
 import {web3torrent, Web3TorrentContext} from './clients/web3torrent-client';
 
 ReactDOM.render(
   <Web3TorrentContext.Provider value={web3torrent}>
     <App />
+    {process.env.REACT_APP_INCLUDE_DRIFT_CHATBOX && <Drift appId="kp3gf29uk9c3" />}
```
junjizhi, line 11: How about defining the appId as the env var: `DRIFT_APPID` and removing `REACT_APP_INCLUDE_DRIFT_CHATBOX ` instead of hard coding? That would make it easier to update.

geoknee: I addressed this in f8788a2. Do we think the appId should be (changed and) stored as a secret (on netlify or circle or wherever we do that) ?

PR #1573, opened 2020-04-27
[e2e] Run integration tests in a shell script

geoknee:
So they can be launched with a single command.


packages/e2e-tests/run-e2e.sh
```diff
@@ -0,0 +1,29 @@
...
+cd ../e2e-tests
+yarn run wait-on -d 5000 http://localhost:3000 http://localhost:3055
+yarn test $1
+code=$?
+
+kill $ganache
+kill $wallet
+kill $app
+kill $hub
+killall node
```
geoknee, line 28: After spending some time trying to track all the descendant processes carefully and clean them up, I decided to use this command to make sure. It's not ideal, but we can remove it once we ensure all of our various start scripts clean up after themselves properly when they finish or are killed.

packages/web3torrent/scripts/start.js
```diff
@@ -1,23 +1,21 @@
...
 // Makes the script crash on unhandled rejections instead of silently
 // ignoring them. In the future, promise rejections that are not handled will
 // terminate the Node.js process with a non-zero exit code.
 process.on('unhandledRejection', err => {
-  if (ganacheServer) {
-    ganacheServer.close();
-  }
   if (devServer) {
-    devServer.close();
+    devServer.kill();
```
geoknee, line 15: `devServer` does not have a `.close()` method, since it is a `ChildProcess`. We can `kill()` it, but its children might not be killed.

packages/e2e-tests/run-e2e.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+cd ../devtools
+yarn start:shared-ganache & 
+ganache=$!
+
+cd ../$1 && yarn run wait-on ../../.ganache-deployments/ganache-deployments-8545.json
```
geoknee, line 7: Previously, we waited on the *folder*, not on the file. The file is cleaned up when ganache winds down, but the folder is not. If starting from a clean hard drive (i.e. on circle) this is not a problem. On your local machine, it meant that this command would not wait properly after the first time you run it.

packages/e2e-tests/run-e2e.sh
```diff
@@ -0,0 +1,29 @@
...
+yarn start &
+app=$!
+
+cd ../xstate-wallet
+yarn start &
+wallet=$!
+
+cd ../simple-hub
+yarn hub:watch &
+hub=$!
```
geoknee, line 17: I decided to run the `hub` every time, even for e2e tests that do not require it.

Related issue: I don't think we can `wait-on` the hub to be ready, although we probably want to do that. There might be a workaround?

packages/e2e-tests/run-e2e.sh
snario: Wouldn't the filename change if `GANACHE_PORT` was changed?

packages/e2e-tests/run-e2e.sh
snario: Probably we should add a `lock.pid` file which has the process ID and get deleted when the hub shuts down.

packages/e2e-tests/run-e2e.sh
geoknee: Yes it would. But because of our `SC_ENV` solution, not easy to get the actual `GANACHE_PORT` into this script. 

It's the same problem to `wait-on localhost:3000`.

packages/e2e-tests/run-e2e.sh
geoknee: Would that help us wait for the hub to be 'ready'?

snario: Hm, no I suppose not. I guess there are two viable options:
1. Have a message type that the hub is expected to reply to. Like `{ recipient: "firebase:simple-hub", value: "ping" }` that gets a reply `"pong"`. Then have a little script which pings until it gets a pong.
2. Add a very lightweight HTTP server and add a `/heartbeat` endpoint that returns `"OK"` when everything is "ready"
I'd say we should do #2 (in a separate PR) as it relates to an item on my [wishlist](https://www.notion.so/elfour/0ee12a01ea684be095ca3b38909c9399?v=282f6c575af544e88924e7a1864a7a57&p=87f9f19a1da04c89849829d79920ac10)

PR #1575, opened 2020-04-27
Decouple create budget call from enable()

junjizhi:
Addresses https://github.com/statechannels/monorepo/issues/1453

### DIFF
--- a/packages/web3torrent/src/pages/file/File.tsx
+++ b/packages/web3torrent/src/pages/file/File.tsx
@@ -53,6 +53,13 @@ const File: React.FC<Props> = props => {
   );
 
   useEffect(() => {
+    const initResult = async () => {
+      await web3Torrent.init();
+    };
+    initResult();
+  }, [web3Torrent]);
+
+  useEffect(() => {
     const testResult = async () => {
       const torrentCheckResult = await checkTorrent(infoHash);
       setWarning(torrentCheckResult);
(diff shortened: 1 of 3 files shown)


packages/web3torrent/src/pages/file/File.tsx
```diff
@@ -52,6 +52,13 @@ const File: React.FC<Props> = props => {
     })
   );
 
+  useEffect(() => {
+    const initResult = async () => {
+      await web3Torrent.init();
+    };
+    initResult();
+  }, [web3Torrent]);
```
junjizhi, line 60: Not sure if this is the correct usage of useEffect, but I will leave it for now.

lalexgap:
@junjizhi  just a heads up that I've done something similar in my PR https://github.com/statechannels/monorepo/pull/1528

packages/web3torrent/src/pages/file/File.tsx
snario: cc @kerzhner / @geoknee / @spersico

spersico: I think that Alex's PR supersedes this one. Might be useful to coordinate efforts to merge both pieces of code

junjizhi: Good catch. This is mostly an investigation so the code here is not polished. I will focus on Alex's PR instead.

PR #1576, opened 2020-04-27
Rename XstateStore to Store

andrewgordstewart:


andrewgordstewart:
I am conflicted about exporting the `StoreInterface` type. Defining it as an interface seems like useful documentation, so I think defining the interface has some value.

Exporting it, and referring to the interface in the machines, means that we only use store methods defined on the interface. On the other hand, it is a level of indirection: when I cmd+click on `store.getEntry`, it brings me to the interface definition, and I have to remember to use the "Go to implementation" feature (not available on github)

junjizhi:
> I am conflicted about exporting the StoreInterface type. 

Looks like we aren't going to have another store implementation soon. The only thing I've seen is `TestStore` which extends Store. So this `StoreInterface` indirection hurts us especially when the interface is in flux ( I see the file change history and it has lots of changes recently). We can get rid of the interface and use duck typing.

PR #1578, opened 2020-04-28
CI: build and release simple hub

kerzhner:
On pushes to https://github.com/statechannels/monorepo/tree/deploy-web3torrent-xstate-wallet, CircleCI builds the `simple-hub` docker image and releases the docker image to Heroku.

The main issue with this PR is that the prepare + release is very slow, about 13 minutes. Some ideas on how to speed that up:
- Coalesce `prepare` and `hub_deploy` into one job so that persisting the workspace and attaching to the workspace are not needed.
- Do not run `yarn` inside the docker container. This would require running yarn without hoisting in the prepare job so that `node_modules` for the entire monorepo do not have to copied into the `simple-hub` docker container.
- Do not use the docker executor for `hub_deploy`. Machine executors do not have to use `remote_docker`, which is quite slow with large workspaces.


.circleci/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN apt-get install -y \
     tar \
     zlib1g-dev
 
-RUN curl -o /usr/bin/solc -fL https://github.com/ethereum/solidity/releases/download/v0.6.6/solc-static-linux \
+RUN curl -o /usr/bin/solc -fL https://github.com/ethereum/solidity/releases/download/v0.6.1/solc-static-linux \
```
snario, line 19: Can we set the version as a build variable?

.circleci/config.yml
```diff
@@ -55,6 +56,30 @@ restore_lib: &restore_lib
...
+    - run: heroku container:release -a simple-hub-staging simple-hub
+  before_build:
+    - attach_workspace:
+        at: /home/circleci/project
+    - run: curl https://cli-assets.heroku.com/install.sh | sh
+    - run: cd packages/simple-hub && yarn build
+  deploy: false
+  dockerfile: packages/simple-hub/docker/simple-hub.dockerfile
+  docker-password: HEROKU_API_KEY
+  docker-username: HEROKU_USERNAME
```
snario, line 72: What's this? Also why aren't these prefixed with a `$`?

.circleci/config.yml
```diff
@@ -55,6 +56,30 @@ restore_lib: &restore_lib
...
+    - run: heroku container:login
+    - run: docker push registry.heroku.com/simple-hub-staging/simple-hub
+    - run: heroku container:release -a simple-hub-staging simple-hub
+  before_build:
+    - attach_workspace:
+        at: /home/circleci/project
+    - run: curl https://cli-assets.heroku.com/install.sh | sh
+    - run: cd packages/simple-hub && yarn build
+  deploy: false
+  dockerfile: packages/simple-hub/docker/simple-hub.dockerfile
```
snario, line 70: Where is the build cache?

.circleci/Dockerfile
kerzhner: yep, but that should be done in a separate PR

.circleci/config.yml
kerzhner: https://circleci.com/orbs/registry/orb/circleci/docker#executors-docker: Name of environment variable storing your Docker password

.circleci/config.yml
kerzhner: Nowhere :). Would probably speed things up.

snario:
Does this trigger a deploy hook into Slack? (I think it ought to right since we use ChatOps)

kerzhner:
> Does this trigger a message to be sent into #notify-deployments in Slack? (I think it ought to right since we use ChatOps)

Is that hook enabled and sending slack messages? I haven't seen any slack messages for hub deploys.

snario:
> > Does this trigger a message to be sent into #notify-deployments in Slack? (I think it ought to right since we use ChatOps)
> 
> Is that hook enabled and sending slack messages? I haven't seen any slack messages for hub deploys.

In theory https://statechannels.slack.com/archives/C0125NB072B/p1587999635002100

snario:
It might be helpful to add some notes in the README, or perhaps in some kind of dev guide, that describes some of the nuances of this config (motivation, todo items like the ones you listed in the description of this github issue, etc).

kerzhner:
> It might be helpful to add some notes in the README, or perhaps in some kind of dev guide, that describes some of the nuances of this config (motivation, todo items like the ones you listed in the description of this github issue, etc).

Good call: https://github.com/statechannels/monorepo/issues/1583. I wonder where this type of explanation belongs. I was thinking that README might have just the explanation for how to deploy the hub: something like, push `deploy-web3torrent-xstate-wallet` branch and expect the hub to be deployed. I wonder if we should add more comments to the circleci config

Issue #1579, opened 2020-04-28
[simple-hub] More verbose logging around deposit transaction


kerzhner:
Logs currently look like this:

```
2020-04-28T13:26:36.298231+00:00 app[simple-hub.1]: [2020-04-28 13:26:36.297 +0000] INFO  (simple-hub/3 on 7a63f567-5fa6-4732-bc4b-2c4ee2b0f38f): Depositing
2020-04-28T13:26:36.298231+00:00 app[simple-hub.1]: [2020-04-28 13:26:36.297 +0000] INFO  (simple-hub/3 on 7a63f567-5fa6-4732-bc4b-2c4ee2b0f38f): Responding with message
2020-04-28T13:26:36.298231+00:00 app[simple-hub.1]: messages: []
2020-04-28T13:26:36.298232+00:00 app[simple-hub.1]: [2020-04-28 13:26:36.297 +0000] INFO  (simple-hub/3 on 7a63f567-5fa6-4732-bc4b-2c4ee2b0f38f): Deposited
```

kerzhner:
Fixed in https://github.com/statechannels/monorepo/commit/92a0ee3c3f1a45b2733a1be4222d7a1056afdf19.

PR #1581, opened 2020-04-28
Add slightly more logs to deposit function on hub

snario:
Maybe helpful towards #1579


packages/simple-hub/src/blockchain/eth-asset-holder.ts
```diff
@@ -42,12 +42,18 @@ export class Blockchain {
         return;
       }
 
+      log.info(
```
kerzhner, line 45: Maybe we should include the channel id as well.

PR #1582, opened 2020-04-28
Handle miscellaneous TODOs in codebase

snario:


packages/xstate-wallet/src/store/state-utils.ts
```diff
@@ -194,7 +194,7 @@ export function convertToNitroOutcome(outcome: Outcome): NitroOutcome {
       ];
     case 'MixedAllocation':
       // FIXME: is this a typo?
```
geoknee, line 196: ?

packages/xstate-wallet/src/workflows/ethereum-enable.ts
```diff
@@ -97,15 +97,15 @@ export const ethereumEnableWorkflow = (
     hideUi: () => {
       messagingService.sendDisplayMessage('Hide');
     },
-    sendResponse: async (context: WorkflowContext, event) => {
+    sendResponse: async (context: WorkflowContext) => {
       messagingService.sendResponse(context.requestId, {
         signingAddress: await store.getAddress(),
         walletVersion: WALLET_VERSION,
         selectedAddress: context.enabledAddress as string
-      }); // TODO: typing
```
geoknee, line 105: Is this still a TODO, but untracked? Or is this now done?

packages/xstate-wallet/src/workflows/virtual-funding-as-hub.ts
```diff
@@ -120,7 +119,7 @@ export const config: MachineConfig<Init, any, any> = {
       },
       onDone: States.fundTargetChannel
     }
-  } as any // TODO: This is to deal with some flickering compilation issues.
```
geoknee, line 123: 👍

packages/xstate-wallet/src/store/state-utils.ts
snario: The code is literally broken, but there was an unused var that I removed

packages/xstate-wallet/src/workflows/ethereum-enable.ts
snario: No, it was fixed (the arguments are typed), but the TODO was never removed.

packages/xstate-wallet/src/chain.ts
```diff
@@ -413,22 +413,22 @@ export class ChainWatcher implements Chain {
...
+    //   filter(
+    //     ({
+    //       blockNumber,
+    //       chainInfo: {
+    //         channelStorage: {finalizesAt}
+    //       }
+    //     }) => finalizesAt.gt(0) && finalizesAt.lte(blockNumber)
+    //   ),
+    //   map(({chainInfo}) => ({channelId, ...chainInfo}))
+    // );
```
geoknee, line 431: Taken at face value, this seems a little heavy handed -- can you justify?

Is it better than just leaving the  code intact with its accompanying FIXME?

snario: The variable was unused and did not work anyway because you can't pass in an ethers style event emitter as an argument. Still need to work on this further.

packages/xstate-wallet/src/store/state-utils.ts
geoknee: Ok. If you have any more context / understanding of this comment, could you improve it? The comment is a question, but it sounds like you are sure that this is broken. 

Appreciate that this is outside scope of your PR, but it seems like a good opportunity.

packages/xstate-wallet/src/chain.ts
junjizhi: Opinion: If we are sure we don't use this, better delete the whole thing.

packages/xstate-wallet/src/workflows/conclude-channel.ts
```diff
@@ -83,6 +83,7 @@ const virtualDefunding = {
   onDone: 'success'
 };
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const withdraw = (store: Store) => async (ctx: Init) => {
```
junjizhi, line 87: We can't remove the param?

snario: In this particular case I'm not sure why `withdraw` is a "no-op". @andrewgordstewart @lalexgap ?

PR #1586, opened 2020-04-28
Split apart config from constants in xstate-wallet

snario:
When developing I found it a bit confusing that some things were constants, some were config items potentially drawn from `process.env`. Hoping this is a step in the direction of cleaning up the `xstate-wallet ` bit so configuration options are separate from codebase constants.


packages/xstate-wallet/src/config.ts
```diff
@@ -0,0 +1,34 @@
...
+import {hexZeroPad, bigNumberify} from 'ethers/utils';
+import {Destination} from './store';
+
+export const WALLET_VERSION: string = process.env.WALLET_VERSION || 'xstate-wallet@VersionTBD';
+
+export const APP_ENV: string | undefined = process.env.APP_ENV;
+
+export const CHAIN_NETWORK_ID: string = process.env.CHAIN_NETWORK_ID || '0';
+
+export const CLEAR_STORAGE_ON_START = Boolean(process.env.CLEAR_STORAGE_ON_START);
```
andrewgordstewart, line 11:
```
> Boolean('false')
true
```

packages/xstate-wallet/src/constants.ts
```diff
@@ -1,45 +1,8 @@
...
-export const HUB_DESTINATION = hexZeroPad(HUB_ADDRESS, 32) as any;
-export const HUB = {
-  destination: HUB_DESTINATION,
-  signingAddress: HUB_ADDRESS,
-  participantId: 'firebase:simple-hub'
-};
-export const CHALLENGE_DURATION = bigNumberify(0x12c); // 5 minutes
-export const NETWORK_ID = process.env.CHAIN_NETWORK_ID || '0';
-
 // TODO: Use getEnvBool from devtools once working
```
andrewgordstewart, line 6: This is now unused.

packages/xstate-wallet/.eslintrc.js
```diff
@@ -44,5 +45,22 @@ module.exports = {
...
+      rules: {
+        'no-process-env': 'off'
+      }
+    },
+    // process.env allowed in src/config.js
+    {
+      files: ['src/config.ts'],
+      rules: {
+        'no-process-env': 'off'
+      }
```
andrewgordstewart, line 63: :+1:

snario:
@andrewgordstewart I think this is just one small step in the direction of defining a wallet configuration object which could perhaps eventually be passed in via CLI, or as a `.toml` config, or some other abstract configuration option set as opposed to relying heavily on `process.env` and injected variables from `jest` or `webpack`. It doesn't do much right now besides clearly label things which are config options.

There are also places in the codebase where we legitimately use constant values that could be put in `constants.ts`. Also, some constants (e.g., `CHALLENGE_DURATION`) could be config options.

andrewgordstewart:
The integration tests look broken. Are you sure firebase is configured correctly?

snario:
> The integration tests look broken. Are you sure firebase is configured correctly?

I think it should be fine, now. Earlier it was broken because I changed `IS_BROWSER_CONTEXT` to use `webpack.DefinePlugin` which I think is a better indicator for "is browser" but the tests failed. I'm not sure why, but they pass when the check is based on `NODE_ENV !== "test"`.

andrewgordstewart:
Ok, are you sure that the failing tests are a flicker then?

andrewgordstewart:
> Ok, are you sure that the failing tests are a flicker then?

I [re-ran](https://app.circleci.com/pipelines/github/statechannels/monorepo/5643/workflows/64c6e3de-8a53-40df-b90a-a8642f4708af/jobs/21892/steps) the failing test, and it failed again with similar logs. I don't believe this to be a flickering test, considering that almost every other test run on master is passing. Given that
- this PR makes a change to the configuration that is used in the integration test
- the test failed twice in a row, with similar behaviour

it seems much more likely that there is a bonafide error.

snario:
@andrewgordstewart I think it was a flickering test 😆 Made another two commits that seem irrelevant and it passed.

andrewgordstewart:
> @andrewgordstewart I think it was a flickering test 😆 Made another two commits that seem irrelevant and it passed.

It seems difficult to check now, since a `force-push` changes history, which means:
- we can't re-run the failed tests (the commit where the test failed no longer exists)
- we can't compare the code when the test failed to the code with the two seemingly unrelated commits

The last twelve merged PRs have had green check marks*, and the same commit failed twice for the same reason (and I have not personally seen a different flickering test failure for that reason.) It's good that this branch now has a ✅ , but this seems like a case where force-pushing has negative consequences.

* maybe they have green checkmarks because the developers re-ran flickering tests to go away. The latest track record on master is [not so good](https://app.circleci.com/pipelines/github/statechannels/monorepo?branch=master).

packages/xstate-wallet/src/logger.ts
```diff
@@ -1,8 +1,9 @@
 import pino from 'pino';
-import {LOG_DESTINATION, ADD_LOGS} from './constants';
 
-// TODO: Is there a better way to determine if we're in a browser context?
-const IS_BROWSER_CONTEXT = process.env.NODE_ENV !== 'test';
+import {LOG_DESTINATION, ADD_LOGS, NODE_ENV} from './config';
+
```
andrewgordstewart, line 4: By removing this comment, do you mean to say that you think this is the best way to determine if we're in a browser context?

packages/xstate-wallet/src/messaging.ts
```diff
@@ -283,7 +284,7 @@ async function convertToInternalEvent(
         participants: request.params.participants.map(convertToInternalParticipant),
         outcome,
         challengeDuration: bigNumberify(CHALLENGE_DURATION),
-        chainId: NETWORK_ID,
+        chainId: CHAIN_NETWORK_ID,
```
andrewgordstewart, line 287: 👍

packages/xstate-wallet/src/logger.ts
snario: No, I'll add it back

PR #1590, opened 2020-04-28
Fix dexie backend

andrewgordstewart:
- Fixes dexie import
- Fixes nested transaction issue by emitting `'ChannelUpdated'` events _after_ the transaction is committed
- sets `USE_INDEXED_DB = 'true'` in root .env files

### DIFF
--- a/.env.virtual-funding
+++ b/.env.virtual-funding
@@ -17,6 +17,7 @@ RPC_ENDPOINT = http://${GANACHE_HOST}:${GANACHE_PORT}
 ## xstate-wallet
 XSTATE_WALLET_DEPLOYER_ACCOUNT_INDEX = '3'
 LOG_DESTINATION = 'console'
+USE_INDEXED_DB = 'true'
 
 
 ## web3torrent
--- a/packages/xstate-wallet/src/store/dexie-backend.ts
+++ b/packages/xstate-wallet/src/store/dexie-backend.ts
@@ -11,7 +11,11 @@ import {
 } from './types';
 import * as _ from 'lodash';
 
-import {Transaction, Dexie, TransactionMode} from 'dexie';
+// eslint-disable-next-line import/no-duplicates
+import Dexie from 'dexie';
+// eslint-disable-next-line import/no-duplicates
+import {Transaction, TransactionMode} from 'dexie';
+
 import {unreachable} from '../utils';
 import {logger} from '../logger';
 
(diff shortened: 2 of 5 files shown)


packages/xstate-wallet/src/store/store.ts
```diff
@@ -358,20 +367,21 @@ export class Store {
...
+          const channelId = calculateChannelId(state);
+          const memoryChannelStorage =
+            (await this.backend.getChannel(channelId)) || (await this.initializeChannel(state));
+          // TODO: This is kind of awkward
+          state.signatures.forEach(sig => memoryChannelStorage.addState(state, sig));
+          await this.backend.setChannel(channelId, memoryChannelStorage.data());
+          return memoryChannelStorage;
+        }
+      )
+      .then(entry => this._eventEmitter.emit('channelUpdated', entry));
```
lalexgap, line 384: Might be worthwhile leaving a comment explaining that we want this outside the transaction scope.

packages/xstate-wallet/src/store/dexie-backend.ts
```diff
@@ -11,7 +11,11 @@ import {
 } from './types';
 import * as _ from 'lodash';
 
-import {Transaction, Dexie, TransactionMode} from 'dexie';
+// eslint-disable-next-line import/no-duplicates
+import Dexie from 'dexie';
```
lalexgap, line 15: I think this can be written as `import Dexie, {Transaction, TransactionMode} from 'dexie';` so we don't have to ignore the eslint rules

packages/xstate-wallet/src/store/store.ts
```diff
@@ -247,7 +252,8 @@ export class Store {
   public setLedger = (ledgerId: string) =>
     this.backend.transaction(
       'readwrite',
-      [ObjectStores.ledgers, ObjectStores.channels],
+      // FIXME: Needs a unit test -- was missing privateKeys store
```
lalexgap, line 255: `FIXME` still need to be fixed?

lalexgap:
Thanks for figuring this out!

.env.virtual-funding
```diff
@@ -17,6 +17,7 @@ RPC_ENDPOINT = http://${GANACHE_HOST}:${GANACHE_PORT}
 ## xstate-wallet
 XSTATE_WALLET_DEPLOYER_ACCOUNT_INDEX = '3'
 LOG_DESTINATION = 'console'
+USE_INDEXED_DB = 'true'
```
lalexgap, line 20: This might be worth announcing to the dev group as a "breaking change", like we were talking about in today's standup.

PR #1594, opened 2020-04-29
Sp/headfull ci

spersico:


packages/e2e-tests/Dockerfile
```diff
@@ -1,29 +1,37 @@
...
+# ADD https://api.github.com/repos/statechannels/monorepo/git/refs/heads/$BRANCH version.json
 RUN  git pull 
 RUN  yarn
+# the following lines copy stuff into the image manually, so that we can try things without commiting and pushing
+# COPY .env.circle-integration-w3t /home/appuser/monorepo/.env.circle-integration-w3t
+# COPY ./puppeteer/helpers.ts /home/appuser/monorepo/packages/e2e-tests/puppeteer/helpers.ts
 
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true
+ENV DISPLAY=:99.0
+ENV PUPPETEER_EXEC_PATH=google-chrome-stable
```
spersico: moved the env variables from the entrypoint to here, seems like the correct approach

packages/e2e-tests/Dockerfile
```diff
@@ -1,29 +1,37 @@
...
      && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
      && apt-get update \
      && apt-get install -y google-chrome-stable --no-install-recommends \
      && rm -rf /var/lib/apt/lists/*
-
+     
+# RUN useradd -r -u 1001 appuser -m
+# USER appuser
+RUN mkdir /home/appuser
+WORKDIR /home/appuser
```
spersico, line 19: this entire _appuser_ thing is an attempt to run chrome without using the root user (the default in docker). Didn't work, failed with the following 
image

packages/e2e-tests/entrypoint.sh
```diff
@@ -1,12 +1,10 @@
...
-set -e
-exec "$@"
\ No newline at end of file
+if [[ $@ == *"bash"* ]]
+then
+  echo "Interactive console: "
+  set -e
+  exec "$@"
+else
+  Xvfb -ac :99 -screen 0 1280x800x24 -ac -nolisten tcp -dpi 96 +extension RANDR > /dev/null 2>&1 & yarn test:e2e:w3t
```
spersico, line 9: if the run command includes bash (it works with bash and /bin/bash) use that, otherwise run the test

geoknee:
In favour of #1613

Issue #1605, opened 2020-04-29
Wallet gets stuck on Closing Channel screen when downloader cancels download


lalexgap:
When the downloader cancels the download they now seem to get stuck on the "Closing Channel" screen.

I think our current integration tests are not strict enough that this gets detected as failure.

It seems to be happening pretty consistently for me (it's all I've run into)

It looks the wallet gets stuck in the `supportState` machine of conclude channel
```
browser.js:186 JOIN_CHANNEL-0x0f8c60f0c2bc5fd0ef523084db5cbcfd493412027e11e4c596fe84cac1dbebdc/conclude-channel.concludeChannel.supportState - TRANSITION FROM Object13:28:43.161browser.js:186 JOIN_CHANNEL-0x0f8c60f0c2bc5fd0ef523084db5cbcfd493412027e11e4c596fe84cac1dbebdc/conclude-channel.concludeChannel.supportState/support-state - STARTED "support-state" TRANSITIONED TO "sendState"13:28:43.161browser.js:186 JOIN_CHANNEL-0x0f8c60f0c2bc5fd0ef523084db5cbcfd493412027e11e4c596fe84cac1dbebdc/conclude-channel.concludeChannel.supportState - TRANSITION FROM Object EVENT PLAYER_REQUEST_CONCLUDE TO Object13:28:43.161browser.js:186 JOIN_CHANNEL-0x0f8c60f0c2bc5fd0ef523084db5cbcfd493412027e11e4c596fe84cac1dbebdc/conclude-channel.concludeChannel.supportState/support-state - STARTED "support-state" TRANSITIONED TO "sendState"13:28:43.161browser.js:186 JOIN_CHANNEL-0x0f8c60f0c2bc5fd0ef523084db5cbcfd493412027e11e4c596fe84cac1dbebdc/conclude-channel.concludeChannel.supportState - TRANSITION FROM Object EVENT PLAYER_REQUEST_CONCLUDE TO Object13:28:43.161browser.js:186 JOIN_CHANNEL-0x0f8c60f0c2bc5fd0ef523084db5cbcfd493412027e11e4c596fe84cac1dbebdc/conclude-channel.concludeChannel.supportState/support-state - STARTED "support-state" TRANSITIONED TO "sendState"
```

lalexgap:
I think this could be the root cause: https://github.com/statechannels/monorepo/issues/1456#issuecomment-615521106

lalexgap:
[...]

PR #1607, opened 2020-04-29
More checks on the integration tests

### LINKED ISSUE #1606: Add stricter visibility checks in the integration tests
Currently we're not checking whether elements are `visible` before attempting to click them. We should add more checks so we can be sure the `wallet` is being displayed when we expect it.

### LINKED ISSUE #1608: Closing a channel displays the Wallet UI when Virtual Funding

lalexgap:
Fixes #1606.

Adds some stricter checks around visibility in the integration tests.

Fixes #1608

### DIFF
--- a/packages/e2e-tests/puppeteer/helpers.ts
+++ b/packages/e2e-tests/puppeteer/helpers.ts
@@ -8,6 +8,16 @@ import {ETHERLIME_ACCOUNTS} from '@statechannels/devtools';
 
 const logDistinguisherCache: Record<string, true | undefined> = {};
 
+export const waitForWalletToBeDisplayed = async (page: Page): Promise<void> => {
+  const walletIframe = page.frames()[1];
+  await walletIframe.waitForSelector(':root', {visible: true});
+};
+
+export const waitForWalletToBeHidden = async (page: Page): Promise<void> => {
+  const walletIframe = page.frames()[1];
+  await walletIframe.waitForSelector(':root', {hidden: true});
+};
+
 export async function setupLogging(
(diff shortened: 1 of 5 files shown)


packages/e2e-tests/puppeteer/helpers.ts
```diff
@@ -307,8 +317,12 @@ export const waitAndOpenChannel = (usingVirtualFunding: boolean) => async (
 export const waitForNthState = async (page: Page, n = 50): Promise<void> => {
   return doneWhen(page, `parseInt(channelStatus.turnNum) >= ${n}`);
 };
+export const waitForClosedState = async (page: Page): Promise<void> => {
+  return doneWhen(page, `channelStatus.status === 'closed'`);
+};
 
 export async function waitForClosingChannel(page: Page): Promise<void> {
+  await waitForWalletToBeDisplayed(page);
```
andrewgordstewart, line 325: Actually, this expectation should _not_ be here for virtual funding!

lalexgap: I've created #1608 for this. Since this is how the `wallet` currently works I'll leave the check as is.

lalexgap:
> Can this work (including resolving #1605) be done in a different branch than `master`? It would be bad for everyone to have ❌ on their PRs that are unrelated to their work, and not a bonafide flicker.

Yeah, my plan is to leave this open while I resolve the issues that this exposes ( like #1605). I figured there might be some value in putting this PR up now so others can see the additional checks that are failing.

PR #1613, opened 2020-04-30
[e2e] Headful mode fix

spersico:
### Found the issue with the headfull mode
Using the `set -e` and `exec "$@" `changes the context of the shell, making that the `Xvfb` command
to stop being valid.

### Observations: 
  - running `docker run -it dapp` will make the command in line 13 run, without changing the context... and the seeder test **passes**! (ignore the rest :stuck_out_tongue:)
  - running `docker run -it dapp bash` and then running `yarn test:e2e:w3t` will fail
  - running `docker run -it dapp bash` and then running  `Xvfb ... &` and **then** `yarn test:e2e:w3t` **will work!**


packages/e2e-tests/entrypoint.sh
```diff
@@ -1,12 +1,23 @@
...
 
 # Startup Xvfb
 Xvfb -ac :99 -screen 0 1280x800x24 -ac -nolisten tcp -dpi 96 +extension RANDR > /dev/null 2>&1 &
 
-cd packages/e2e-tests
-set -e
-exec "$@"
\ No newline at end of file
+if [[ $@ == *"bash"* ]]
+then
```
kerzhner, line 7: @spersico Nice work! FYI, you can also override entrypoint via a command line flag. So this is not strictly necessary.

PR #1622, opened 2020-04-30
Improve e2e testing

andrewgordstewart:
# What this does
1. Extracts code which starts servers into a separate script. This script can be used to start standalone servers, which developers can use to play around, get started, or re-run integration tests without having to re-start servers (time consuming)
2. Adds timeouts to the wait-on commands -- eliminates circle spending 10 minutes without output before killing test job
3. `package.json` uses the interpreter specified by the shebang. shell scripts are executable in the git repo, so you shouldn't have to `chmod +x bin/start-servers.sh`
4. [This 🔨](https://github.com/statechannels/monorepo/commit/a9940bd7fbd522b96e16f49ee4648d44f27cbd38#diff-70897b729216d823de8976df8749c6ed)  was removed. (It seems bad to kill all node processes on some random developer's computer.)
5. Adds `process.exit()` in the `simple-hub`. This might take care of https://github.com/statechannels/monorepo/issues/1580 and https://github.com/statechannels/monorepo/issues/1566 ?
6. More `$VARIABLES` in the shell script(s) -> more readable/maintainable?
7. Integration tests are `--runInBand` (Maybe running tests in parallel is causing some issues.)
8. `resource_class: large` in rps integration test (one test failed due to Out Of Memory)

Locally, I checked:
A. what happens when I exit the script prematurely? (Child processes eventually die, but not instantly)
B. what happens when the jest tests pass? (It says "Exiting with status 0", and child processes die.)
C. what happens when the [jest tests fail](https://app.circleci.com/pipelines/github/statechannels/monorepo/5736/workflows/dc1fff4e-7d0e-44c2-a9b9-f1dc66e41204/jobs/22413/steps)? (It says "Exiting with status $SOME_NON_ZERO_THING", child processes die.)

# Why did I do this?

i) I was investigating https://github.com/statechannels/monorepo/issues/1593.
ii) The test script seemed great, but rough around the edges
iii) On circle, if the servers failed to start, circle would not time out for 10 minutes.
iv) Locally, trying to reproduce it by running `run-e2e.sh`, it would `killall node`, which I considered to be mildly "bad" towards my teammates.
v) If we are attracting attention to our codebase, this is progress towards a one-command "Here's how to start everything". In this case, (iv) is **really bad**.

# Where it fails
- I still don't fully understand `kill`, `wait`, and `trap`. So I can't guarantee that this works reliably, and I can't figure out how to wait for child processes to die before exiting myself.


packages/e2e-tests/bin/run-e2e.sh
```diff
@@ -0,0 +1,60 @@
...
+cleanup() {
+  # Ignore signals while shutting down
+  trap '' INT TERM EXIT
+  if [[ -z services ]]
+  then
+    echo Warning: services undefined: $services
+  else
+    echo Killing $services
+    # I'm not 100% sure that this will always work.
+    # However, having a background process running shouldn'g cause a circle failure
```
snario, line 20:
```suggestion
    # However, having a background process running shouldn't cause a circle failure
```

packages/e2e-tests/bin/start-servers.sh
```diff
@@ -0,0 +1,48 @@
...
+  exit 1
+fi
+
+cd $PACKAGES/devtools
+yarn start:shared-ganache | tee $E2E_ROOT/shared-ganache.log & 
+
+yarn run wait-on -t $WAIT_ON_TIMEOUT -i $WAIT_ON_INTERVAL $MONOREPO_ROOT/.ganache-deployments/ganache-deployments-8545.json
+
+cd $PACKAGES/$APP 
+yarn start | tee $E2E_ROOT/app.log &
```
snario, line 40: Would it be a good idea to use `$APP` for the filename?

andrewgordstewart:
It "works" with the latest master (fails with https://github.com/statechannels/monorepo/issues/1525)

PR #1629, opened 2020-05-01
Persistent seeder deployment

spersico:
**This branch is where I'm approaching the persistent-seeder deployment.**

### TODO list
Mi first step is deploying manually the docker instance in heroku.
After I get that to work (**currently deployed, but with an error on the script itself**), I plan to automate it with circleci.

### DIFF
--- a/packages/e2e-tests/persistent-seeder-deployment/build-push-release.sh
+++ b/packages/e2e-tests/persistent-seeder-deployment/build-push-release.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -euf -o pipefail
+BASEDIR=$(dirname "$0")
+BACKUP_DOCKERIGNORE="$BASEDIR"/../../../bk.dockerignore
+GLOBAL_DOCKERIGNORE="$BASEDIR"/../../../.dockerignore
+
+switchDockerIgnore () {
+  # necessary for as long as circleci wont support docker 19.03, which adds multiple .dockerignore files support
+  # https://ideas.circleci.com/ideas/CCI-I-1265
+  # https://stackoverflow.com/a/57774684/6569950
+  echo "Switching dockerignore"
+  cp "$BASEDIR"/../../../.dockerignore "$BACKUP_DOCKERIGNORE"
+  cp persistent-seeder.dockerfile.dockerignore "$GLOBAL_DOCKERIGNORE"
+}
+
+restoreDockerIgnore() { #resets, if necessary, the .dockerignore file to it's original state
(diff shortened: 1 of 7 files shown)


spersico:
@geoknee the current issue that I'm running into is that the persistent-seeder script itself doesn't seem to work, in this branch nor in master.

The error that gets shown in the console is 
image

On the heroku console, I can see the same error:
image

And, when running it outside of a docker container, I see:
image

geoknee:
In my view, this is not a problem with the persistent seeder script. If you go to https://web3torrent.statechannels.org/upload/ in your browser now, you will see that the "Select File" button is disabled. This might be an incorrect CHAIN_NETWORK_ID, or some other issue preventing that page from being ready.

geoknee:
I think we ought to have more verbose logging for the persistent-seeder. Something along the lines of https://github.com/statechannels/monorepo/blob/9bc541022d5dc56ba88fd89d7ddc4854cb4c8ab2/packages/e2e-tests/puppeteer/scripts/persistent-seeder.ts#L12-L15

Or there might be a better way?

geoknee:
I reproduced your error locally, then redeployed web3torrent + xstate-wallet, and it went away 🤷‍♂️

packages/e2e-tests/persistent-seeder-deployment/persistent-seeder.dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM node:10.16.3
```
geoknee, line 1: On circle, we use the `circleci/node:10.16.3-browsers` image and then skip the chrome download commands below. Could we pull that trick here, too? Might speed things up. 

Also, can we now delete the existing Dockerfile in this package?

packages/e2e-tests/persistent-seeder-deployment/build-push-release.sh
```diff
@@ -0,0 +1,30 @@
...
+restoreDockerIgnore() { #resets, if necessary, the .dockerignore file to it's original state
+  if test -f "$BACKUP_DOCKERIGNORE"; then
+    echo "Restoring dockerignore backup"
+    mv "$BACKUP_DOCKERIGNORE" "$GLOBAL_DOCKERIGNORE"
+  fi
+}
+
+
+# MAIN SCRIPT
+switchDockerIgnore
```
kerzhner, line 26: I don't think we need this switching dance. I think it's ok for `simple-hub` and this to have the same docker context. The context size will be bigger, but the slowdown is likely not significant enough for added code complexity.

My proposal is here: https://github.com/statechannels/monorepo/pull/1632

PR #1631, opened 2020-05-01
Add conclude-channel idempotency tests for direct and virtual funding

junjizhi:
Addresses: https://github.com/statechannels/monorepo/issues/1602

TODO
- [x] Fix the setup to simulate a directly funded channel between PlayerA and PlayerB
- [x] Fix the failing test
- ~~[ ] Refactor the test setup code (currently it is copied from `challenge.test.ts`)~~

Update: Latest commit also adds a test for the virtual funding case, which addresses: https://github.com/statechannels/monorepo/issues/1603

### DIFF
--- a/packages/xstate-wallet/src/workflows/tests/conclude-channel.test.ts
+++ b/packages/xstate-wallet/src/workflows/tests/conclude-channel.test.ts
@@ -76,7 +76,7 @@ const context: Init = {channelId: targetChannelId, funding: 'Direct'};
 let aStore: TestStore;
 let bStore: TestStore;
 
-const allSignState = (state: State) => ({
+const allSignedState = (state: State) => ({
   ...state,
   signatures: [wallet1, wallet2].map(({privateKey}) => createSignatureEntry(state, privateKey))
 });
@@ -194,12 +194,12 @@ beforeEach(async () => {
   const hubStore = new SimpleHub(wallet3.privateKey);
 
   [aStore, bStore].forEach(async (store: TestStore) => {
-    await store.createEntry(allSignState(firstState(allocation, targetChannel)), {
+    await store.createEntry(allSignedState(firstState(allocation, targetChannel)), {
       applicationSite: TEST_SITE
     });
 
     const ledgerEntry = await store.createEntry(
-      allSignState(firstState(allocation, ledgerChannel))
+      allSignedState(firstState(allocation, ledgerChannel))
     );
     await store.setLedgerByEntry(ledgerEntry);
   });
--- a/packages/xstate-wallet/src/workflows/tests/conclude-channel.test.ts
+++ b/packages/xstate-wallet/src/workflows/tests/conclude-channel.test.ts
@@ -1,7 +1,8 @@
 import {interpret} from 'xstate';
 import waitForExpect from 'wait-for-expect';
 
-import {Init, machine as createMachine} from '../create-and-fund';
+import {Init, machine as createChannel} from '../create-and-fund';
+import {machine as concludeChannel} from '../conclude-channel';
 
 import {Store} from '../../store';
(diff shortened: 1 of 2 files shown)


packages/xstate-wallet/src/workflows/tests/conclude-channel.test.ts
```diff
@@ -0,0 +1,218 @@
...
+  const runMachine = (store: Store) => interpret(machine(store).withContext(context)).start();
+  const [aService, bService] = [aStore, bStore].map(runMachine);
+
+  const targetState = direct ? 'success' : {virtualDefunding: 'asLeaf'};
+
+  await waitForExpect(async () => {
+    expect(bService.state.value).toEqual(targetState);
+    expect(aService.state.value).toEqual(targetState);
+  }, EXPECT_TIMEOUT);
+};
```
andrewgordstewart, line 104: An issue with this approach is that the `waitForExpect` can "miss" the target state. `waitForExpect` polls the callback until it passes, so if you transition past the target state during the polling interval, it might never pass.

You can do something like
```suggestion
  const services = [aStore, bStore].map(runMachine);
  const targetState = direct ? 'success' : {virtualDefunding: 'asLeaf'};

  await Promise.all(services.map(service => 
    new Promise(resolve => service.onTransition(state => state.matches(targetState) && resolve()))
  );
};
```

(I was not aware of this technique when I introduced `waitForExpect`)

packages/xstate-wallet/src/workflows/tests/conclude-channel.test.ts
```diff
@@ -0,0 +1,218 @@
...
+let bStore: TestStore;
+
+const allSignState = (state: State) => ({
+  ...state,
+  signatures: [wallet1, wallet2].map(({privateKey}) => createSignatureEntry(state, privateKey))
+});
+
+let chain: FakeChain;
+
+const runUntilSuccess = async (machine, direct = true) => {
```
andrewgordstewart, line 94: nit: use string-valued funding types rather than booleans
```
const runUntilSuccess = async (machine, fundingType: 'direct' | 'virtual' = 'direct') => {
```

junjizhi: done.

packages/xstate-wallet/src/workflows/tests/conclude-channel.test.ts
```diff
@@ -0,0 +1,248 @@
...
+  const amountA2 = (await aStore.chain.getChainInfo(targetChannelId)).amount;
+  const amountB2 = (await bStore.chain.getChainInfo(targetChannelId)).amount;
+
+  const entryA2 = await aStore.getEntry(targetChannelId);
+  const entryB2 = await bStore.getEntry(targetChannelId);
+
+  expect(amountA2).toMatchObject(amountA1);
+  expect(amountB2).toMatchObject(amountB1);
+
+  expect(outcomesEqual(entryA1.latestState.outcome, entryA2.latestState.outcome)).toBe(true);
```
junjizhi, line 246: We should be able to expect the entry doesn't change.

andrewgordstewart: The issue is that the virtual-defunding-as-leaf workflow _also_ [concludes the target](https://github.com/statechannels/monorepo/blob/master/packages/xstate-wallet/src/workflows/virtual-defunding-as-leaf.ts#L38-L52) (in a less robust fashion)

junjizhi: done.

snario:
Looks good to me, will let @andrewgordstewart approve though

PR #1637, opened 2020-05-01
[w3t, e2e] Wait for input button to not be disabled

geoknee:
I am seeing this issue on production right now. I want to add some message to disambiguate this problem from other "we are stuck waiting on something" puppeteer error messages.

### DIFF
--- a/packages/e2e-tests/puppeteer/scripts/web3torrent.ts
+++ b/packages/e2e-tests/puppeteer/scripts/web3torrent.ts
@@ -32,15 +32,9 @@ export async function uploadFile(
   filePath?: string
 ): Promise<string> {
   // https://pub.dev/documentation/puppeteer/latest/puppeteer/FileChooser-class.html
-  // Not clear why puppeteer FileChooser won't work out of box. We are doing it manually for now.
-  const inputUploadHandle = await page.waitForSelector('input[type=file]');
-  if (inputUploadHandle.getProperty('disabled')) {
-    console.warn('Input button is disabled!!');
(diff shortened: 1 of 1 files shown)


packages/e2e-tests/puppeteer/scripts/web3torrent.ts
```diff
@@ -31,16 +31,21 @@ export async function uploadFile(
...
   filePath?: string
 ): Promise<string> {
-  await page.waitForSelector('input[type=file]');
-
+  // https://pub.dev/documentation/puppeteer/latest/puppeteer/FileChooser-class.html
+  // Not clear why puppeteer FileChooser won't work out of box. We are doing it manually for now.
+  const inputUploadHandle = await page.waitForSelector('input[type=file]');
+  if (inputUploadHandle.getProperty('disabled')) {
+    console.warn('Input button is disabled!!');
+    await page.waitFor(4000);
```
snario, line 39: Couldn't we use `[disabled=false]` in the selector to wait for it being enabled?

geoknee: Maybe, I will try. It might have to be `disabled=undefined`

geoknee: 6a4086a

geoknee: Tested with https://chrome.google.com/webstore/detail/css-selector-tester/bbklnaodgoocmcdejoalmbjihhdkbfon

PR #1647, opened 2020-05-02
Add persistent-seeder deployment to CI

spersico:
This PR adds the persistent-seeder deployment to CircleCI, to be deployed every day.

The flow can be seen [here](https://app.circleci.com/pipelines/github/statechannels/monorepo/5825/workflows/e3668928-d122-4295-8a3e-5fc3f04a59b8) and the job [here](https://app.circleci.com/pipelines/github/statechannels/monorepo/5825/workflows/e3668928-d122-4295-8a3e-5fc3f04a59b8/jobs/23009)

image
_this was actually me, but the loaded Heroku credentials belong to mike :stuck_out_tongue:_


packages/e2e-tests/persistent-seeder-deployment/persistent-seeder.dockerfile
```diff
@@ -1,15 +1,8 @@
-FROM node:10.16.3
-
+FROM circleci/node:10.16.3-browsers
+USER root
```
snario, line 2: I remember seeing that this is generally bad practice. Probably not a big deal since this is all open source, but if we can avoid it perhaps we should?

.circleci/config.yml
```diff
@@ -335,7 +335,32 @@ jobs:
...
+      - checkout
+      - <<: *restore_dep
+      - <<: *restore_lib
+      - run: rm -rf ./patches && cd ./packages/ && find . -maxdepth 1 ! -path "./channel-provider" ! -path "./client-api-schema" ! -path "./e2e-tests" ! -path "./devtools" -exec rm -rf {} ';'
+      - run: yarn --frozen-lockfile
+      - run: curl https://cli-assets.heroku.com/install.sh | sh
+      - run: heroku container:login
+      - docker/build:
+          cache_from: registry.heroku.com/persistent-seeder-staging/persistent-seeder:latest
+          dockerfile: packages/e2e-tests/persistent-seeder-deployment/persistent-seeder.dockerfile
```
snario, line 355: As a general comment, it seems odd to me that `persistent-seeder` is defined inside of `packages/e2e-tests`.

.circleci/config.yml
```diff
@@ -335,7 +335,32 @@ jobs:
...
+      - image: circleci/node:10.16.3-browsers
+    steps:
+      - setup_remote_docker
+      - checkout
+      - <<: *restore_dep
+      - <<: *restore_lib
+      - run: rm -rf ./patches && cd ./packages/ && find . -maxdepth 1 ! -path "./channel-provider" ! -path "./client-api-schema" ! -path "./e2e-tests" ! -path "./devtools" -exec rm -rf {} ';'
+      - run: yarn --frozen-lockfile
+      - run: curl https://cli-assets.heroku.com/install.sh | sh
+      - run: heroku container:login
```
snario, line 352: Can any of these things be defined once and reused since they're partially similar to the hub deployment config?

.circleci/config.yml
```diff
@@ -335,7 +335,32 @@ jobs:
...
+      - docker/build:
+          cache_from: registry.heroku.com/persistent-seeder-staging/persistent-seeder:latest
+          dockerfile: packages/e2e-tests/persistent-seeder-deployment/persistent-seeder.dockerfile
+          image: registry.heroku.com/persistent-seeder-staging/persistent-seeder
+          tag: latest
+      - docker/push:
+          image: persistent-seeder-staging/persistent-seeder
+          registry: registry.heroku.com
+          tag: latest
+      - run: heroku container:release -a persistent-seeder-staging persistent-seeder
```
snario, line 362: We don't _really_ have a staging environment. We just have one production environment. Maybe we shouldn't keep using "staging" in the names of things given this as it might cause confusion.

kerzhner: My original thinking was staging == ropsten and production == mainnet.

packages/e2e-tests/persistent-seeder-deployment/persistent-seeder.dockerfile
spersico: Yeah, I tried not using that, but the image wouldn't let me install xvfb for some reason.
Then I tried using a less privileged user for the rest of things, and ran into even more problems. On Monday I'll give it another look. I'm interested in that issue, I didn't like it either

.circleci/config.yml
snario: Should we consider labelling all other things as staging too then, e.g., the web3torrent and xstate-wallet deployments? I suppose the wallet is intended to support multiple chains... worth a discussion

.circleci/config.yml
geoknee: The reason for that is that it has the same dependencies as, and leans on helper functions defined in that package. It is functionally extremely similar to our e2e-tests (only it does not run as an actual test). Alternatives include:
- moving to an entirely new package
- moving under web3torrent, but that involves either
  - making web3torrent depend on e2e-tests 🤮 
  - duplicating the e2e-tests code 🤮 

I'm not sure it is worth it at this point.

.circleci/config.yml
kerzhner: On second thought, should we get rid of staging/production terminology and instead use ropsten/mainnet terminology? So the heroku application name becomes `persistent-seeder-ropsten`

snario: Yeah I think that makes more sense. It's not clear that we need separate environments or deployments for both chains though. A single service might be able to work in production against multiple chains.

.circleci/config.yml
spersico: I'll take a look right now

.circleci/config.yml
spersico: I don't know the specifics, but I think that's out of scope for this PR.
To start with, to support multiple chains, we need to change code to do so, right now the chain is a single string, not an array of options. 
...At least in web3torrent, I don't think this suggestion would be a one-line change.

packages/e2e-tests/persistent-seeder/build-push-release.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -euf -o pipefail
+BASEDIR=$(dirname "$0")
+
+# MAIN SCRIPT
+docker build -t registry.heroku.com/persistent-seeder-production/persistent-seeder -f persistent-seeder.dockerfile "$BASEDIR"/../../.. || restoreDockerIgnore
```
kerzhner, line 7:
```suggestion
docker build -t registry.heroku.com/persistent-seeder-production/persistent-seeder -f persistent-seeder.dockerfile "$BASEDIR"/../../..
```

spersico: good catch!

packages/e2e-tests/persistent-seeder-deployment/persistent-seeder.dockerfile
spersico: https://github.com/statechannels/monorepo/pull/1647/commits/6a08eccbff4bc9252c99657d75650cad004aad20

.circleci/config.yml
spersico: Done in https://github.com/statechannels/monorepo/pull/1647/commits/ec540f5477405ad9ec9ba8ee88dd6488016ea072 and https://github.com/statechannels/monorepo/pull/1647/commits/fe450db18bfb451b4c72f5eae1dbb54b2ce2f5e7

.circleci/config.yml
```diff
@@ -111,6 +111,30 @@ commands:
       - run:
           command: Xvfb -ac :99 -screen 0 1280x800x24 -ac -nolisten tcp -dpi 96 +extension RANDR > /dev/null 2>&1
           background: true
+  
+  build_push_release:
+    description: 'Build a docker image, push it to the heroku registry, and make a release in heroku'
+    parameters:
+      dockerfilePath:
```
kerzhner, line 118:
```suggestion
      dockerfile_path:
```

.circleci/config.yml
```diff
@@ -111,6 +111,30 @@ commands:
...
+      pipeline:
+        type: string
+      app:
+        type: string
+    steps:
+      - run: curl https://cli-assets.heroku.com/install.sh | sh
+      - run: heroku container:login
+      - docker/build:
+          cache_from: registry.heroku.com/<< parameters.app >>/<< parameters.pipeline >>:latest
+          dockerfile: << parameters.dockerfilePath >>
```
kerzhner, line 129:
```suggestion
          dockerfile: << parameters.dockerfile_path >>
```

.circleci/config.yml
```diff
@@ -327,19 +351,27 @@ jobs:
...
-          image: registry.heroku.com/simple-hub-staging/simple-hub
-          tag: latest
-      - docker/push:
-          image: simple-hub-staging/simple-hub
-          registry: registry.heroku.com
-          tag: latest
-      - run: heroku container:release -a simple-hub-staging simple-hub
-
+      - build_push_release:
+          dockerfilePath: packages/simple-hub/docker/simple-hub.dockerfile
```
kerzhner, line 355:
```suggestion
          dockerfile_path: packages/simple-hub/docker/simple-hub.dockerfile
```

.circleci/config.yml
```diff
@@ -327,19 +351,27 @@ jobs:
...
+    docker:
+      - image: circleci/node:10.16.3-browsers
+    steps:
+      - setup_remote_docker
+      - checkout
+      - <<: *restore_dep
+      - <<: *restore_lib
+      - run: rm -rf ./patches && cd ./packages/ && find . -maxdepth 1 ! -path "./channel-provider" ! -path "./client-api-schema" ! -path "./e2e-tests" ! -path "./devtools" -exec rm -rf {} ';'
+      - build_push_release:
+          dockerfilePath: packages/e2e-tests/persistent-seeder/persistent-seeder.dockerfile
```
spersico, line 371:
```suggestion
          dockerfile_path: packages/e2e-tests/persistent-seeder/persistent-seeder.dockerfile
```

PR #1654, opened 2020-05-04
Upgrade to stable dexie version

snario:


packages/xstate-wallet/package.json
```diff
@@ -11,7 +11,7 @@
     "@statechannels/wire-format": "0.0.1",
     "@xstate/react": "1.0.0-rc.3",
     "async-lock": "1.2.2",
-    "dexie": "^3.0.0-rc.7",
```
kerzhner, line 14: Should an exact version get used here instead of a range?

snario: Yes

packages/xstate-wallet/package.json
```diff
@@ -11,7 +11,7 @@
...
+    "dexie": "^3.0.0",
```
kerzhner, line 14:
```suggestion
    "dexie": "3.0.0",
```

PR #1659, opened 2020-05-04
Don't run tests on deploy branch

snario:
Not sure if the other jobs which require it will automatically inherit this

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -359,35 +359,35 @@ workflows:
             - prepare
           filters:
             branches:
-              ignore: liam/noprepare
+              ignore: deploy
 
       - test:
           requires:
             - prepare
           filters:
             branches:
-              ignore: liam/noprepare
+              ignore: deploy
 
       - integration-test-rps:
           requires:
             - prepare
           filters:
             branches:
-              ignore: liam/noprepare
+              ignore: deploy
 
       - integration-test-web3torrent:
           requires:
             - prepare
           filters:
             branches:
-              ignore: liam/noprepare
+              ignore: deploy
 
       - integration-test-web3torrent-with-hub:
           requires:
             - prepare
           filters:
             branches:
-              ignore: liam/noprepare
+              ignore: deploy
 
       - release-pull-request:
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -352,7 +352,10 @@ jobs:
 workflows:
   statechannels:
     jobs:
-      - prepare
+      - prepare:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -352,7 +352,10 @@ jobs:
 workflows:
   statechannels:
     jobs:
-      - prepare
+      - prepare:
+          filters:
+            branches:
+              ignore: liam/noprepare
```
snario, line 358: Will change to `deploy` if tests pass

kerzhner:
@snario I might be missing something, but don't you need to build monorepo packages that the xstate-wallet and w3t depend on (like nitro-protocol and wire-format) before building those packages? That is basically what `prepare` job does.

snario:
I think that is happening here right now

https://github.com/statechannels/monorepo/blob/fb276ae49838a773d9cf3ccc2129f04f274c36da/.circleci/config.yml#L328-L328

Also, `deploy` runs in parallel to `prepare`

snario:
`release-netlify-production` definitely need `prepare` actually, but not `release-hub-production`

kerzhner:
Check out https://app.circleci.com/pipelines/github/statechannels/monorepo/5865/workflows/075b8ee1-459d-42ce-ab78-429ab198da8f to see the job depencdy graph of a `deploy` workflow. `release-hub-production` does not depend on prepare for the reason you pointed out. release-`netlify-production` depends on `prepare` for the reason I pointed out.

snario:
You're right  -- the spirit of this PR was to not re-run the tests on `deploy`, let me change it to do that correctly

kerzhner:
> You're right -- the spirit of this PR was to not re-run the tests on `deploy`, let me change it to do that correctly

Gotcha. I am of the opinion that we actually want to run test on the deploy workflow. What if the deploy branch contains a hotfix that was never included in master? In that case, the tests would never get run for that code change.

Also, the test are non-blocking for deploys, so the only downside is that we are burning circlci credits. But how often do we expect to deploy?

snario:
> What if the deploy branch contains a hotfix that was never included in master?
I think we should open PRs against master in addition to committing hotfixes

kerzhner:
So then master might contain commits A->B-H where H is a hotfix. `deploy` might contain commits A->H because B is not ready for deployment. What if H doesn't work without B? Tests might help us catch that.

kerzhner:
My overall points are:
- It is not a given that tests have been run on the code being deployed.
- Tests are not blocking, so why not run them?

PR #1660, opened 2020-05-04
Remove --production flag from yarn

geoknee:
(we need to install devtools, which is a dev dep)


kerzhner:
@geoknee The hub currently should not rely on `devtools` when not using Ganache. Let me briefly investigate the devtools error on heroku.

kerzhner:
Looks like this change brought `devtools` into the `simple-hub` production environment: https://github.com/statechannels/monorepo/pull/1660. I am not sure convinced that was intentional?

kerzhner:
Looks like that PR did not make an intentional decision to introduce devtools to the `simple-hub` production environment.

kerzhner:
Closing in favor of https://github.com/statechannels/monorepo/pull/1666

PR #1664, opened 2020-05-04
[web3t] Run e2e test against deployed apps

### LINKED ISSUE #1620: [w3t, e2e] Run e2e tests against production urls
We should be able to run `SC_ENV=production-ropsten yarn test:e2e:w3t-with-hub` to automate testing of our production code (using dappeteer). 

We could tack this on to the end of our production-deployment pipeline on circle (the one that runs on the preffered github branch, currently named `deploy-web3torrent-xstate-wallet`). 

This step would not spin up any servers at all, and should therefore be pretty fast. It should wait on other production steps successfully completing.

It could provide rapid feedback in the case where we deployed code that won't work.

geoknee:
- converts e2e tests for w3t to read in target app url via new env var 
- new circle job `e2e-test-w3t-production`
  - only runs on deploy branch, and *after* apps have been redeployed
  - does not start any servers of its own
  - runs e2e tests (headful, dappeteer enabled) against specified environment (assumes servers are running)

Closes #1620 .

Tested against *this* branch (temporarily). Test runs and fails due to known issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,7 +417,8 @@ workflows:
 
       - e2e-test-w3t-production:
           # for testing only
-          # requires:
+          requires:
+            - prepare
           #   - release-hub-production
           #   - release-netlify-production
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -342,6 +342,25 @@ jobs:
           tag: latest
       - run: heroku container:release -a simple-hub-staging simple-hub
 
+  e2e-test-w3t-production:
+    resource_class: large
+    working_directory: /home/circleci/project
+    docker:
+      - image: circleci/node:10.16.3-browsers
+    environment:
+      SC_ENV: production-ropsten
+    steps:
+      - install_xvfb
+      - run_xvfb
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -339,6 +339,25 @@ jobs:
           tag: latest
       - run: heroku container:release -a simple-hub-staging simple-hub
 
+  e2e-test-w3t-production:
```
snario, line 342: `staging`? cc @kerzhner

snario:
Is there any way we can produce some sort of "report" with referenced transaction IDs somewhere? i.e., etherscan links, maybe in the CI artifacts, that I can click on to view successful transactions?

.circleci/config.yml
geoknee: I think out of scope for this PR, which follows conventions that exist on the base branch. If we were to go with 'staging' we should also make changes in quite a few other places (best done in a separate PR).

geoknee:
@snario I'm sure something like that would be possible. For what purposes would you want to see that? Debugging?

PR #1666, opened 2020-05-04
[simple-hub] production fixes

kerzhner:
- do not use `devtools` in production.
- do not use `pino-pretty` in production.
- add `lodash` to production dependencies.


.circleci/config.yml
```diff
@@ -323,7 +323,7 @@ jobs:
       - checkout
       - <<: *restore_dep
       - <<: *restore_lib
-      - run: rm -rf app-wallet-interface channel-provider channel-client \
+      - run: cd packages && rm -rf app-wallet-interface channel-provider channel-client \
```
snario, line 326: 😱

kerzhner: All credit goes to @spersico for catching this one.

packages/simple-hub/env.ts
```diff
@@ -1 +1,77 @@
-require('@statechannels/devtools').configureEnvVariables();
\ No newline at end of file
+// FIXME: To remove @statechannels/hub's dependency on @statechannels/devtools,
+// this file is a carbon copy of the file that exports configureEnvVariables in devtools.
+
+// const {configureEnvVariables} = require('@statechannels/devtools');
```
andrewgordstewart, line 4: Can you clarify this comment?

As the culprit, I removed this carbon copy because it seemed like:
- There's a `FIXME` here, so the intent was to change this at some point in the future
- `@statechannels/hub` did not want to depend on `@statechannels/devtools`, for some reason
- this package is `@statechannels/simple-hub`
- `@statechannels/simple-hub` _already_ depends on `@statechannels/devtools`, so why not use this?
```
const {configureEnvVariables} = require('@statechannels/devtools');
```

I think the answer is more subtle, that you
- want to use `@statechannels/devtools` as a _development_ dependency
- don't want to depend on `@statechannels/devtools` as a _dependency_
- _do_ want to configure environment variables in production (without using _development_ dependencies)

kerzhner:
> LGTM. We should add an issue to remove environment variable file-fetching from being a runtime dependency in production (perhaps by injecting them into the dockerfile at build time).

Good call: https://github.com/statechannels/monorepo/issues/1667

packages/simple-hub/env.ts
kerzhner: Good call: https://github.com/statechannels/monorepo/issues/1667

PR #1668, opened 2020-05-04
Add Idempotency test for virtual funding

junjizhi:
https://github.com/statechannels/monorepo/issues/1603


junjizhi:
included in https://github.com/statechannels/monorepo/pull/1631

PR #1682, opened 2020-05-05
[e2e] Introduce TX_WAIT_TIMEOUT constant

geoknee:
...and use it wherever we are waiting for a blockchain tx.

We have two e2e tests for web3torrent. Currently `withdraw.test.ts` fails when we target production https://app.circleci.com/pipelines/github/statechannels/monorepo/5927/workflows/72f3061a-0f91-47ae-8e3f-fa3739542874/jobs/23677/steps . However, this seems to be (only) because we time out waiting for ropsten. 

This PR is a hotfix for that which should get the test passing. 

`seed-download.test.ts` will still fail, this PR does nothing to help with that. 

We should probably introduce a more elegant and robust long term fix for this: currently it is difficult to tell when we are waiting for a blockchain transaction. One idea would be to wrap dappeteer's `metamask.confirmTransaction()` method with something like `metamask.confirmTransactionAndWait(numBlocks: number)`.

### DIFF
--- a/packages/e2e-tests/puppeteer/constants.ts
+++ b/packages/e2e-tests/puppeteer/constants.ts
@@ -8,3 +8,4 @@ export const TARGET_NETWORK = process.env.TARGET_NETWORK ? process.env.TARGET_NE
 export const WEB3TORRENT_URL = process.env.WEB3TORRENT_URL
   ? process.env.WEB3TORRENT_URL
   : 'http://localhost:3000';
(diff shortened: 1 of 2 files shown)


geoknee:
After this change got to the `deploy` branch, we see one pass and one fail, as expected https://app.circleci.com/pipelines/github/statechannels/monorepo/5939/workflows/a117f806-9273-491f-bd61-f15b1e0176de/jobs/23759/steps

PR #1685, opened 2020-05-05
Configure tracker on subdomain of statechannels.org

snario:
I have a suspicion that this might have something to do with messages not being exchanged between browser tabs on production. I set the DNS record using [Netlify DNS](https://app.netlify.com/teams/statechannels/dns/statechannels.org) and provisioned the certificate using the [Heroku dashboard](https://dashboard.heroku.com/apps/web3torrent-tracker-staging/settings).


snario:
I think I classify the production issue as unlikely to have anything to do with this.

However, I also think this is a harmless change.

PR #1689, opened 2020-05-05
simple-hub: simplify eth-asset-holder

kerzhner:
This work probably won't fix https://github.com/statechannels/monorepo/issues/1577 but will hopefully make it simpler to think about the hub funding logic.

`ethAssetHolder` Contract object is created on the initial `fund` call and is protected by the `depositing` lock to ensure no race conditions.


packages/simple-hub/src/blockchain/eth-asset-holder.ts
```diff
@@ -23,55 +24,41 @@ export async function makeDeposits(
...
+    log.info(
+      `submitting deposit transaction to eth asset holder with value: ${value
+        .sub(expectedHeld)
+        .toHexString()}`
+    );
+    const tx = await ethAssetHolder.deposit(channelID, expectedHeld.toHexString(), value, {
+      value: value.sub(expectedHeld)
     });
-  }
+    log.info(`waiting for tx to be mined hash=${tx.hash}`);
```
andrewgordstewart, line 55:
```suggestion
    log.info(
    {transaction: {hash: tx.hash, nonce: tx.nonce}},
    'waiting for tx to be mined');
```

packages/simple-hub/src/blockchain/eth-asset-holder.ts
```diff
@@ -23,55 +24,41 @@ export async function makeDeposits(
...
       release();
-      return holdings;
+      return;
+    }
+
+    log.info(
+      `submitting deposit transaction to eth asset holder with value: ${value
+        .sub(expectedHeld)
+        .toHexString()}`
+    );
```
andrewgordstewart, line 51:
```suggestion
    log.info(
      {value: value.sub(expectedHeld).toHexString()},
      'submitting deposit transaction to eth asset holder'
    );
```

PR #1696, opened 2020-05-05
Use string literal instead of function name pointer for xstate entry

snario:


snario:
Reproduced locally by building a production webpack build and running it with a local python server at port 3055 in place of a webpack dev server for the wallet. This change fixed the issue we're seeing on production.

@andrewgordstewart said there are other usages of `<function>.name`, we should change those too.

snario:
To be honest, I think we'd be better off just turning minification off. It gets us negligible performance benefits at the cost of your above points and hobbyist dev debugging in their browser (which we want to encourage, not prohibit).

andrewgordstewart:
> To be honest, I think we'd be better off just turning minification off. It gets us negligible performance benefits at the cost of your above points and preventing hobbyist devs from debugging in their browser (which we want to encourage, not prohibit).

I'm on board.

snario:
Closing in favor of https://github.com/statechannels/monorepo/pull/1699

PR #1707, opened 2020-05-06
[circle] notify slack on e2e-prod failure

geoknee:
We may want to wait to merge this, until this job settles down a bit (to avoid spamming slack). 

Having said that, we don't push to deploy *all* that often, so maybe it is ok.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,7 +94,7 @@ commands:
     description: 'Notify slack '
     steps:
       - slack/status:
-          failure_message: 'A $CIRCLE_JOB job has failed on "${CIRCLE_BRANCH}"!'
+          failure_message: 'A $CIRCLE_JOB job has failed on $CIRCLE_BRANCH!'
           fail_only: true
           mentions: 'SRHGGRGS0' # Group ID for ActiveDevs
           only_for_branches: 'master'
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -94,7 +94,7 @@ commands:
     description: 'Notify slack '
     steps:
       - slack/status:
-          failure_message: 'A $CIRCLE_JOB job has failed on master!'
+          failure_message: 'A $CIRCLE_JOB job has failed on $CIRCLE_BRANCH!'
           fail_only: true
           mentions: 'SRHGGRGS0' # Group ID for ActiveDevs
           only_for_branches: 'master'
```
kerzhner, line 100: Not sure that this will work since the step has a branch filter for `master`

geoknee: @kerzhner  Good spot: any objection to removing that filter? Or might be nicer to just add 'deploy' to the filter?

kerzhner: In general, the idea with that filter is that we do not want to send slack notifications for PR circle runs. We can either add `deploy` to the filter here or move the filtering outside of the `notify_slack` command definition.

geoknee: b7132fd

PR #1709, opened 2020-05-06
Eject Web3Torrent for CRA

lalexgap:
Eject Web3Torrent from CRA to allow us to customize our config as we see fit.


packages/web3torrent/package.json
```diff
@@ -9,26 +9,66 @@
...
   "dependencies": {
+    "@babel/core": "7.6.0",
     "@rimble/connection-banner": "1.2.3",
     "@statechannels/channel-client": "0.0.1",
     "@statechannels/devtools": "0.1.23",
+    "@svgr/webpack": "4.3.2",
+    "@typescript-eslint/eslint-plugin": "^2.2.0",
+    "@typescript-eslint/parser": "^2.2.0",
+    "babel-eslint": "10.0.3",
+    "babel-jest": "^24.9.0",
```
andrewgordstewart, line 20: Lots of these dependencies look like dev dependencies. (Are they even all used?)

lalexgap: Yeah, I'm not sure why `CRA` decides to add them all to `dependencies` instead of `devDependencies`. We can probably clean these up as we go.

andrewgordstewart: https://www.npmjs.com/package/yarn-check might give some insight

lalexgap: Unfortunately `yarn-check` doesn't seem to report correct information about `not used`. IE: It flags `typescript`, `rxjs`, `webtorrent` as not used.

lalexgap: Also a lot of these dependencies (like `babel`) are baked into CRA so I think to remove them we need to update various things (like the `webpack.config`, jest configuration in `package.json`) etc..

.circleci/config.yml
```diff
@@ -3,6 +3,7 @@ version: 2.1
 orbs:
   docker: circleci/docker@1.0.1
   slack: circleci/slack@3.4.1
+  puppeteer: threetreeslight/puppeteer@0.1.2
```
lalexgap, line 6: This gets around this error: https://app.circleci.com/pipelines/github/statechannels/monorepo/6062/workflows/28924b87-cb72-497c-930c-a5eacd549a11/jobs/24564

I'm not sure if there is a better solution.

packages/web3torrent/config/jest/fileTransform.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
```
snario, line 1: Can you add some plain english description to explain why we need this file?

packages/web3torrent/config/modules.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
```
snario, line 1: Can you add some plain english description to explain why we need this file?

packages/web3torrent/config/paths.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
```
snario, line 1: Can you add some plain english description to explain why we need this file?

packages/web3torrent/config/pnpTs.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
```
snario, line 1: Can you add some plain english description to explain why we need this file?

packages/web3torrent/package.json
```diff
@@ -122,5 +180,58 @@
     "test": "jest -c ./config/jest/jest.config.js",
     "test:ci": "CI=true yarn test --runInBand --ci --all --forceExit",
     "test:coverage": "yarn test --coverage"
+  },
+  "jest": {
```
snario, line 184: Split out into separate jest config?

packages/web3torrent/src/index.tsx
```diff
@@ -4,7 +4,7 @@ import Drift from 'react-driftjs';
 
 import App from './App';
 import {web3torrent, Web3TorrentContext} from './clients/web3torrent-client';
-
+console.log(process.env);
```
snario, line 7: Remove log

packages/web3torrent/config/jest/fileTransform.js
lalexgap: All these files are created by running `yarn eject`

snario:
I'm not so opinionated about this that we should not merge it, I don't mind. I'm just a little traumatized from subtle code compilation bugs.

lalexgap:
> The main criticism I have is that this introduces an enormous amount of extra boilerplate and configuration to the codebase, most of which I suspect is copied from somewhere else. I doubt we need every single one of these files checked in, and there probably exists some sane defaults in most places where new code has been introduced that we could use.
> 
> There exists this thing called [`react-app-rewired`](https://github.com/timarney/react-app-rewired) that seems to be the intended technology for this scenario of wanting to edit webpack configs without ejecting. It is probably frustrating that I just found out this exists right now.
> 
> Overall I would just be strongly in favour of trying to remove as much extra boilerplate as possible.

This is all the code hidden in `CreateReactApp` that gets generated by `yarn eject`. I originally looked at `react-app-rewired` but decided  that doing a project `eject` would be nicer as we could finally own the various configs and we would avoid some hacks like writing out to a `.env.local` to set contract addresses.

I agree that all the boilerplate is a bit awful (and a reason I don't love using CRA, there's a lot of complexity in there). I think we should work to reduce that but I'd prefer to do that outside the scope of this PR.

PR #1713, opened 2020-05-06
[e2e] Wait-on reads from env vars

geoknee:
- defaults to previous localhost 3000 and 3055

Currently there is an unnecessary wait-until-timeout situation when running e2e tests against non localhost servers.

https://app.circleci.com/pipelines/github/statechannels/monorepo/6004/workflows/d3324ad3-f0ef-4c50-ac8a-c9837e3fb706/jobs/24213/steps

### DIFF
--- a/packages/e2e-tests/puppeteer/constants.ts
+++ b/packages/e2e-tests/puppeteer/constants.ts
@@ -5,7 +5,10 @@ export const USE_DAPPETEER = getEnvBool('USE_DAPPETEER');
 export const USES_VIRTUAL_FUNDING = process.env.REACT_APP_FUNDING_STRATEGY === 'Virtual';
 export const JEST_TIMEOUT = HEADLESS ? 200_000 : 1_000_000;
 export const TARGET_NETWORK = process.env.TARGET_NETWORK ? process.env.TARGET_NETWORK : 'localhost';
-export const WEB3TORRENT_URL = process.env.WEB3TORRENT_URL
+export const APP_URL = process.env.WEB3TORRENT_URL
   ? process.env.WEB3TORRENT_URL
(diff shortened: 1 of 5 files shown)


packages/e2e-tests/puppeteer/constants.ts
```diff
@@ -5,7 +5,10 @@ export const USE_DAPPETEER = getEnvBool('USE_DAPPETEER');
 export const USES_VIRTUAL_FUNDING = process.env.REACT_APP_FUNDING_STRATEGY === 'Virtual';
 export const JEST_TIMEOUT = HEADLESS ? 200_000 : 1_000_000;
 export const TARGET_NETWORK = process.env.TARGET_NETWORK ? process.env.TARGET_NETWORK : 'localhost';
-export const WEB3TORRENT_URL = process.env.WEB3TORRENT_URL
+export const APP_URL = process.env.WEB3TORRENT_URL
   ? process.env.WEB3TORRENT_URL
   : 'http://localhost:3000';
```
andrewgordstewart, line 10: It might be slightly confusing that `APP_URL` uses `process.env.WEB3TORRENT_URL`, rather than `process.env.APP_URL`.

geoknee: That's fair, but this is how I rationalise it: Previously we had `APP_URL = http://localhost:3000` as an implicit default. Now that is an explicit default but with an override if `WEB3TORRENT_URL` is defined.

If we have cause to introduce, e.g.,  `RPS_URL`, (which maybe we should do soon) we could alter this line to do something sensible.

Open to other suggestions, though. Might be worth merging this PR and revisiting when we get to that particular bridge.

PR #1714, opened 2020-05-06
[simple-hub] switch default chain private key

kerzhner:
Right now, a participant and the hub share a private key for transaction signing during the integration tests. This PR is the immediate remedy for that.

We should probably do something more robust to avoid this conflict in the future.

Along the journey ⛵ of this PR, the following were also addressed:
- Cleanups for env file formatting and removal of stale env files.
- `xstate-wallet` was not reading in env vars for tests.
- `simple-hub` was not correctly generating the chain address.


snario:
I think you need to change the default on the `xstate-wallet`, too:

https://github.com/statechannels/monorepo/blob/53975f591fe23ebfe4d741a3d5c418f7d11ec0af/packages/xstate-wallet/src/config.ts#L30-L31

andrewgordstewart:
> I think you need to change the default on the `xstate-wallet`, too:
> 
> https://github.com/statechannels/monorepo/blob/53975f591fe23ebfe4d741a3d5c418f7d11ec0af/packages/xstate-wallet/src/config.ts#L30-L31

I think it's ok:
- `HUB_ADDRESS` is the (state channel) "signing address"
- `cHubChainPK` is the private key used for signing blockchain transactions (the xstate wallet doesn't need to care about that address)

In practice, a smart hub might use `cHubChainPK`'s address as its `destination` in outcomes.

kerzhner:
@andrewgordstewart 389ddf4b2 so that the hub is able to withdraw funds if it wants to.

packages/xstate-wallet/src/config.ts
```diff
@@ -30,6 +30,8 @@ export const ETH_ASSET_HOLDER_ADDRESS: string = process.env.ETH_ASSET_HOLDER_ADD
 export const HUB_ADDRESS: string =
   process.env.HUB_ADDRESS || '0xaaaa84838319627Fa056fC3FC29ab94d479B8502';
 
+// This account corresponds to 0x187bb12e927c1652377405f81d93ce948a593f7d66cfba383ee761858b05921a
+// That key is provided eth in @statechannels/devtools/utils/startGanache.js
 export const HUB_DESTINATION = (process.env.HUB_DESTINATION ||
   hexZeroPad('0x8199de05654e9afa5c081bce38f140082c9a7733', 32)) as Destination;
```
andrewgordstewart, line 36: Could this address instead be this?
- the address corresponding to `HUB_CHAIN_PK`, if that environment variable is defined

kerzhner: Not a bad idea. What if `HUB_CHAIN_PK` is not defined?

andrewgordstewart: Then we could default to `HUB_ADDRESS` (like now).

Or, we could throw?

kerzhner: 10f36f589

kerzhner: I ended up reverting the implementation of this suggestion. `HUB_DESTINATION` constant/env var has to be defined for `web3torrent`. Of course, `web3torrent` can calculate the destination from the private key. But this seems wrong since `web3torrent` is unaware of private keys at the moment.

If `HUB_DESTINATION` is defined for `web3torrent`, might as well use the env var for the wallet as well.

packages/xstate-wallet/src/integration-tests/running.test.ts
```diff
@@ -30,7 +30,7 @@ test('accepts states when running', async () => {
   ]);
 
   hookUpMessaging(playerA, playerB);
-  const channelId = '0x1823994d6d3b53b82f499c1aca2095b94108ba3ff59f55c6e765da1e24874ab2';
+  const channelId = '0x440b56b6c5b0adca1ee99e3926d1b123fd867566cfab5150479f9d5e9317fa1e';
```
snario, line 33: How are these things calculated? Maybe add a note

kerzhner: I actually don't really know 🤷. The test started failing when we started properly reading in `.env.test` env vars, which changed the chain network id.

PR #1717, opened 2020-05-06
Add notes on when it's safe to change outcomes

andrewgordstewart:


packages/xstate-wallet/notes/outcomes.md
```diff
@@ -0,0 +1,62 @@
...
+}
+```
+
+The outcome determines what happens to funds in a channel, when the channel is finalized.
+This is explained in detail in the [nitro paper](https://magmo.com/nitro-protocol.pdf).
+
+- For simple allocations, `amount` is _transferred_ to `destination`
+  - In case the channel is _underfunded_ -- the holdings are less than the sum of `amount` over the allocation items -- allocation items are prioritized by their order
+- For simple guarantees, funds are transferred according to the outcome of the `targetChannelId`
+  - :warning: If the target channel does not have a simple allocation outcome, things will will break.
```
andrewgordstewart, line 26:
```suggestion
  - :warning: If the target channel does not have a simple allocation outcome, things will break.
```

packages/xstate-wallet/notes/outcomes.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+The outcome determines what happens to funds in a channel, when the channel is finalized.
+This is explained in detail in the [nitro paper](https://magmo.com/nitro-protocol.pdf).
+
+- For simple allocations, `amount` is _transferred_ to `destination`
+  - In case the channel is _underfunded_ -- the holdings are less than the sum of `amount` over the allocation items -- allocation items are prioritized by their order
+- For simple guarantees, funds are transferred according to the outcome of the `targetChannelId`
+  - :warning: If the target channel does not have a simple allocation outcome, things will break.
+  - In case the guarantor channell is _underfunded_ -- the holdings are less than the sum of `amount` over the allocation items of `guarantee.targetChannelId` -- allocation items are prioritized by the order in `guarantee.destinations`
```
geoknee, line 27:
```suggestion
  - In case the guarantor channel is _underfunded_ -- the holdings are less than the sum of `amount` over the allocation items of `guarantee.targetChannelId` -- allocation items are prioritized by the order in `guarantee.destinations`
```

packages/xstate-wallet/notes/outcomes.md
```diff
@@ -0,0 +1,62 @@
...
+   Case (3) _only_ happens in "funding" channels (ledger, joint, and guarantor channels).
+4. a) The wallet has signed a (latest) state `s` for that channel.
+   b) The app requested to update the outcome to `s'.outcome`
+   c) `validTransition(s, s')` is true
+   Case (4) _only_ happens for application channels, and _only_ when in the application workflow's `'running'` state.
+
+### Notes
+
+- The wallet should _never_ change the outcome from a guarantee to an allocation.
+- Currently, the wallet onyl supports eth.
```
geoknee, line 53:
```suggestion
- Currently, the wallet only supports eth.
```

packages/xstate-wallet/notes/outcomes.md
```diff
@@ -0,0 +1,62 @@
...
+1. a) the wallet has not yet signed a state for that channel.
+   b) outcome on `s'` is "suitable" for the channel's purpose.
+   Cases (1) happens for both app channels and funding channels, when setting up the channel.
+2. a) The wallet has signed a (latest) state `s` for that channel.
+   b) `s.outcome === s'.outcome`
+   Case (2) happens for app channels when setting up the channel (the "post-fund-setup" state) and when concluding the channel.
+   Case (2) also happens for ledger channels, but only when closing the channel.
+3. a) The wallet has signed a (latest) state `s` for that channel.
+   b) `s.outcome !== s'.outcome`, but due to the state of other channels in the store, the user's funds remain invariant
+   Case (3) _only_ happens in "funding" channels (ledger, joint, and guarantor channels).
```
geoknee, line 44: Nit: these conditions apply to case (4), so I don't think case (3) only happens in "funding "channels.

packages/xstate-wallet/notes/outcomes.md
```diff
@@ -0,0 +1,62 @@
...
+4. a) The wallet has signed a (latest) state `s` for that channel.
+   b) The app requested to update the outcome to `s'.outcome`
```
geoknee, line 46: When I receive a state for an application channel from my counterparty's wallet, the app is just acting as a blind message relayer. Would that count as the app requesting to update the outcome? Perhaps the condition could be just about whether the state is for an application channel (there is nontrivial appData ?). Might be nice to not condition on how the state got in to the wallet.

packages/xstate-wallet/notes/outcomes.md
andrewgordstewart: Can you give an example?

geoknee: ✋ maybe I didn't read this carefully enough earlier. Ignore me!

packages/xstate-wallet/notes/outcomes.md
andrewgordstewart: > When I receive a state for an application channel from my counterparty's wallet, the app is just acting as a blind message relayer. Would that count as the app requesting to update the outcome?

In this case, the wallet isn't signing any state, but it is learning new information about what outcomes are now finalizable.

That's a different question. So, two questions the wallet has to answer are:
1. When is it safe to sign a state?
2. When is it safe to discard a state?

These questions are related, but this document was meant to be just about question (1). I'll make a remark to clarify this.

> Perhaps the condition could be just about whether the state is for an application channel (there is nontrivial appData ?).

Just a note that nontrivial `appData` is not equivalent to being an application channel. (Applications are free to use the null app, or any app that has no appData, such as payment channels.)

PR #1731, opened 2020-05-07
[web3torrent] ChannelsList component shows peerAccount

geoknee:
Pull property directly off the wire.

Should fix the issue where currently what is shown as a 'peer' in the UI, is actually me...
Screenshot 2020-05-07 at 18 22 24

An extension to this PR would be to replace the trimmed peer addresses with blockies

### DIFF
--- a/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
+++ b/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
@@ -47,7 +47,7 @@ function channelIdToTableRow(
   }
 
   let dataTransferred: string;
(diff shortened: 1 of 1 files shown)


packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
```diff
@@ -47,7 +47,7 @@ function channelIdToTableRow(
   }
 
   let dataTransferred: string;
-  const peerAccount = channel[participantType];
```
geoknee, line 50: The RHS is always me, and never `peerAccount` (although it isn't super clear from the code).

Issue #1732, opened 2020-05-07
Integration test actors should sign blockchain transactions according to selectedAddress


andrewgordstewart:
Currently, all integration test actors sign all blockchain transactions from address `0xD9995BAE12FEe327256FFec1e3184d492bD94C31` (corresponds to [this private key)](https://github.com/statechannels/monorepo/blob/master/packages/devtools/src/constants.ts#L4). See https://github.com/statechannels/monorepo/issues/1577#issuecomment-625399027 for why this is bad.

This is true even if they have different `selectedAddress` values (which you can [set up](https://github.com/statechannels/monorepo/blob/master/packages/e2e-tests/puppeteer/helpers.ts#L133) if you are not using dappeteer)

I think if you _are_ using dappeteer, you just end up using address `0xD9995BAE12FEe327256FFec1e3184d492bD94C31` anyway.

So, this issue potentially requires code changes:
- [ ] Change [getProvider](https://github.com/statechannels/monorepo/blob/master/packages/xstate-wallet/src/utils/contract-utils.ts#L25) to use the right signer according to `window.eth.selectedAddress`
- [x] Change the dappeteer setup to use unique signinig keys

geoknee:
When using a fake web3, it seems to me that we don't care about addresses at all, since we never interact with a chain. 

When using dappeteer, I was expecting this step to avoid this problem: 
https://github.com/statechannels/monorepo/blob/af9d16b0fda4f1379b7066bb24cacf0d1b60b5ea/packages/e2e-tests/puppeteer/helpers.ts#L198-L202

any idea why it does not?

andrewgordstewart:
> When using a fake web3, it seems to me that we don't care about addresses at all, since we never interact with a chain.

If I set `USE_DAPPETEER=false`, then I _do_ set up a fake web3, but I _still_ interact with a chain during integration tests, right?

> When using dappeteer, I was expecting this step to avoid this problem:
> https://github.com/statechannels/monorepo/blob/af9d16b0fda4f1379b7066bb24cacf0d1b60b5ea/packages/e2e-tests/puppeteer/helpers.ts#L198-L202
> 
> any idea why it does not?

:+1: nice!

[...]

PR #1739, opened 2020-05-07
Add provider and (proper) signer to chain

andrewgordstewart:
- The provider can be fetched on initialization
- The signer should only be available after ethereum is enabled.
  Before then, we don't know what signing address we should use.
- The signer uses the selected address

Resolves https://github.com/statechannels/monorepo/issues/1732

### DIFF
--- a/packages/xstate-wallet/src/chain.ts
+++ b/packages/xstate-wallet/src/chain.ts
@@ -238,9 +238,9 @@ export class ChainWatcher implements Chain {
   private mySelectedAddress: string | null;
   private provider: ReturnType<typeof getProvider>;
   private get signer() {
-    if (typeof this.selectedAddress !== 'string') throw 'Ethereum not enabled';
+    if (!this.ethereumIsEnabled) throw new Error('Ethereum not enabled');
 
-    return this.provider.getSigner(this.selectedAddress);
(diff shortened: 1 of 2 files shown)


andrewgordstewart:
Closing, as my attempt to resolve #1732 seems to be incompatible with dappeteer. This is frustrating, considering that it works when I test things manually with chrome windows.

packages/xstate-wallet/src/chain.ts
```diff
@@ -237,39 +236,57 @@ export class ChainWatcher implements Chain {
...
+  private get signer() {
+    if (typeof this.selectedAddress !== 'string') throw 'Ethereum not enabled';
+
+    return this.provider.getSigner(this.selectedAddress);
+  }
 
   public async initialize() {
-    const provider = getProvider();
-    const signer = provider.getSigner();
+    this.provider = getProvider();
```
snario, line 247: I think we should make this a constructor argument of `Chain`

andrewgordstewart:
(Was using the wrong SC_ENV locally. will try again)

packages/xstate-wallet/src/chain.ts
andrewgordstewart: https://github.com/statechannels/monorepo/issues/1394

PR #1742, opened 2020-05-08
Add REACT_APP_HUB_DESTINATION to production-ropsten env

snario:
Was missing from #1714. We may want some sort of "required env vars" check for these files before CI passes? cc @kerzhner


.env.production-ropsten
```diff
@@ -46,6 +46,7 @@ REACT_APP_FIREBASE_API_KEY = $FIREBASE_API_KEY
 REACT_APP_FIREBASE_PREFIX = $FIREBASE_PREFIX
 REACT_APP_FIREBASE_URL = $FIREBASE_URL
 REACT_APP_FUNDING_STRATEGY = 'Virtual'
+REACT_APP_HUB_DESTINATION = $HUB_DESTINATION
```
kerzhner, line 49: Very much looking forward to getting rid of `REACT_APP_` prefixes.. 🗑️

kerzhner:
It's a good question of where runtime configuration should be validated. Right before a release is probably a good time.

PR #1744, opened 2020-05-08
[xstate-wallet] Set mySelectedAddress to ethereum.selectedAddress on initialization

snario:
After #1739 the wallet now asks the user to "Connect to MetaMask" every time even if they're already connected. This fixes that.

### DIFF
--- a/packages/xstate-wallet/src/chain.ts
+++ b/packages/xstate-wallet/src/chain.ts
@@ -235,7 +235,7 @@ const chainLogger = logger.child({module: 'chain'});
 export class ChainWatcher implements Chain {
   private _adjudicator?: Contract;
   private _assetHolders: Contract[];
-  private mySelectedAddress: string | null;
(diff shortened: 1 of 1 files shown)


packages/xstate-wallet/src/chain.ts
```diff
@@ -235,7 +235,7 @@ const chainLogger = logger.child({module: 'chain'});
 export class ChainWatcher implements Chain {
   private _adjudicator?: Contract;
   private _assetHolders: Contract[];
-  private mySelectedAddress: string | null;
+  private mySelectedAddress: string | null = (window.ethereum || {}).selectedAddress;
```
andrewgordstewart, line 238:
```suggestion
  private mySelectedAddress: string | null = window.ethereum?.selectedAddress ?? null;
```

I think your code change would be a runtime `TypeError`  when `window.ethereum` is falsey.

Typescript does not catch this because `window.ethereum` has type `any`, therefore so does `window.ethereum || {}`.

PR #1747, opened 2020-05-08
Add Sentry to web3torrent

junjizhi:
https://github.com/statechannels/monorepo/issues/1519

Was able to deliberately throw an error and catch that on Sentry: https://sentry.io/organizations/statechannels/issues/1654999040/?project=5228838&query=is%3Aunresolved

### DIFF
--- a/packages/web3torrent/package.json
+++ b/packages/web3torrent/package.json
@@ -11,7 +11,7 @@
   "dependencies": {
     "@babel/core": "7.6.0",
     "@rimble/connection-banner": "1.2.3",
-    "@sentry/browser": "^5.15.5",
+    "@sentry/browser": "^5.12.1",
     "@statechannels/channel-client": "0.0.1",
     "@statechannels/devtools": "0.1.23",
     "@svgr/webpack": "4.3.2",
--- a/packages/web3torrent/package.json
+++ b/packages/web3torrent/package.json
@@ -11,7 +11,7 @@
   "dependencies": {
     "@babel/core": "7.6.0",
     "@rimble/connection-banner": "1.2.3",
-    "@sentry/browser": "^5.12.1",
+    "@sentry/browser": "5.12.1",
     "@statechannels/channel-client": "0.0.1",
     "@statechannels/devtools": "0.1.23",
     "@svgr/webpack": "4.3.2",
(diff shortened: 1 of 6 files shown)


packages/web3torrent/package.json
```diff
@@ -10,6 +10,7 @@
   ],
   "dependencies": {
     "@rimble/connection-banner": "1.2.3",
+    "@sentry/browser": "^5.15.5",
```
kerzhner, line 13: Let's pin this to an exact version and use the same version for rps so that only one version of the package is installed in the monorepo.

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -83,6 +84,9 @@ export class PaymentChannelClient {
   async initialize() {
     await this.channelClient.provider.mountWalletComponent(process.env.WALLET_URL);
     await this.initializeHubComms();
+    if (process.env.NODE_ENV === 'production') {
+      Sentry.setUser({id: this.channelClient.signingAddress});
+    }
```
junjizhi, line 89: @snario According to https://docs.sentry.io/enriching-error-data/additional-data/?platform=javascript

> Each of these is optional, but one *must* be present for Sentry to collect data about the user:

So I pass the address as the user `id`

packages/web3torrent/package.json
junjizhi: done.

packages/web3torrent/package.json
```diff
@@ -11,6 +11,7 @@
   "dependencies": {
     "@babel/core": "7.6.0",
     "@rimble/connection-banner": "1.2.3",
+    "@sentry/browser": "^5.12.1",
```
kerzhner, line 14:
```suggestion
    "@sentry/browser": "5.12.1",
```

packages/xstate-wallet/package.json
```diff
@@ -4,6 +4,7 @@
   "version": "0.0.1",
   "author": "Alex Gap",
   "dependencies": {
+    "@sentry/browser": "^5.12.1",
```
kerzhner, line 7:
```suggestion
    "@sentry/browser": "5.12.1",
```

packages/web3torrent/package.json
kerzhner: Do you have a commit where this is fixed? I caret in front of the version number means that the version is not fixed.

packages/web3torrent/package.json
junjizhi: 🤣

Issue #1729, opened 2020-05-07
Channels left open when the client changes seeder


lalexgap:
The downloader should close all channels related to a file when complete.

lalexgap:
This may be related to the issues around closing a channel.

It looks like a `ChannelUpdated` with a closed status never gets emitted. If I disable the distinct turnnum check in `store.channelUpdatedFeed` then I see two updates with the same turnNum but different statuses.
```
{"time":1588883887693,"level":30,"jsonRpcResponse":{"jsonrpc":"2.0","method":"ChannelUpdated","params":{"participants":[{"participantId":"0x0Be56446B48A73bDa9eCD86541D730cF1264E18D","signingAddress":"0x0Be56446B48A73bDa9eCD86541D730cF1264E18D","destination":"0x0000000000000000000000008199de05654e9afa5C081BcE38F140082C9a7733"},{"participantId":"0x0AeCAF662a8D6EeABD559202b0E45F10f1724fbf","signingAddress":"0x0AeCAF662a8D6EeABD559202b0E45F10f1724fbf","destination":"0x0000000000000000000000008199de05654e9afa5C081BcE38F140082C9a7733"}],"allocations":[{"allocationItems":[{"destination":"0x0000000000000000000000008199de05654e9afa5C081BcE38F140082C9a7733","amount":"0x000000000000000000000000000000000000000000000000000000000037215f"},{"destination":"0x0000000000000000000000008199de05654e9afa5C081BcE38F140082C9a7733","amount":"0x0000000000000000000000000000000000000000000000000000000000000000"}],"token":"0x0000000000000000000000000000000000000000"}],"appDefinition":"0x430869383d611bBB1ce7Ca207024E7901bC26b40","appData":"0x0000000000000000000000000000000000000000000000000000000000000000","status":"running","turnNum":"0x0000000000000000000000000000000000000000000000000000000000000032","channelId":"0x76b02aff5d97151a5eb38c1024c98fecba25ef12958c83497ed11ad14dd9e057"}},"msg":"OUTGOING JSONRPC REQUEST:","name":"xstate-wallet"}
[...]

PR #1748, opened 2020-05-08
Check for duplicate turn numbers in store invariant check

### LINKED ISSUE #1554: Store should only store one state per turn number
https://github.com/statechannels/monorepo/issues/1555 asserts that the store should only support one state per turn number, within a given channel.

The store has no control over what state a peer sends for turn `n`. If one of my workflows subsequently want to support a different state at turn `n`, what should the store do?

Right now, some store logic depends (might depend?) on the fact that there's only one state per turn number.

See https://github.com/statechannels/monorepo/issues/1456#issuecomment-615520101

lalexgap:
Fixes #1554 by throwing an error when we have multiple states with the same turn number.

This would of caught #1729

### DIFF
--- a/packages/xstate-wallet/src/store/channel-store-entry.ts
+++ b/packages/xstate-wallet/src/store/channel-store-entry.ts
@@ -259,7 +259,10 @@ export class ChannelStoreEntry {
 
     const duplicateTurnNums = turnNums.some((t, i) => turnNums.indexOf(t) != i);
     if (duplicateTurnNums) {
-      logger.error({signedStates: _.map(signedStates, s => s.turnNum.toHexString())});
+      logger.error(
+        {signedStates: _.map(signedStates, s => s.turnNum.toHexString())},
+        Errors.duplicateTurnNums
+      );
       throw Error(Errors.duplicateTurnNums);
     }
     if (!isReverseSorted(turnNums)) {
--- a/packages/xstate-wallet/src/store/channel-store-entry.ts
+++ b/packages/xstate-wallet/src/store/channel-store-entry.ts
@@ -257,6 +257,11 @@ export class ChannelStoreEntry {
     const {signedStates} = this;
     const turnNums = _.map(signedStates, s => parseInt(s.turnNum.toHexString(), 16));
 
+    const duplicateTurnNums = turnNums.some((t, i) => turnNums.indexOf(t) != i);
+    if (duplicateTurnNums) {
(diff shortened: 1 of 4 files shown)


snario:
Adding a state with the same store number is a legitimate use case though, for example in doing a unidirectional payment channel where the receiver can conclude at any time.

andrewgordstewart:
> Adding a state with the same store number is a legitimate use case though,

More than that, if we are going to _store_ every state that a peer sends us, this would prevent us from adding our own states with those turn numbers.

> for example in doing a unidirectional payment channel where the receiver can conclude at any time.

I don't understand this example. Can you expand?

snario:
> I don't understand this example. Can you expand?

- Sender wants to send at most 5 coins to Receiver in increments of 1
- Sender signs `{ turnNum: 0, isFinal: true, outcome: [5, 0] }`
- Sender signs `{ turnNum: k, isFinal: true, outcome: [5 - k, k] }`
- Sender goes offline for some `k` (e.g., 3)
- Receiver wants on-chain cash _now_ (no timeout)
- Receiver concludes with `{ turnNum: k + 1, isFinal: true, outcome: [5 -j, j] }` for max `k`

andrewgordstewart:
In this particular case, sender can sign `{ turnNum: k, isFinal: true, outcome: [5-k, k] }`

lalexgap:
> Adding a state with the same store number is a legitimate use case though, for example in doing a unidirectional payment channel where the receiver can conclude at any time.

Currently our `store`/`api` doesn't properly support the case of multiple states per a turn number. I think we should support this eventually but I think there's enough details there that we shouldn't try and do it now.

packages/xstate-wallet/src/store/channel-store-entry.ts
```diff
@@ -257,6 +257,11 @@ export class ChannelStoreEntry {
     const {signedStates} = this;
     const turnNums = _.map(signedStates, s => parseInt(s.turnNum.toHexString(), 16));
 
+    const duplicateTurnNums = turnNums.some((t, i) => turnNums.indexOf(t) != i);
+    if (duplicateTurnNums) {
+      logger.error({signedStates: _.map(signedStates, s => s.turnNum.toHexString())});
```
andrewgordstewart, line 262:
```suggestion
      logger.error(
        {signedStates: _.map(signedStates, s => s.turnNum.toHexString())},
        Errors.duplicateTurnNums
      );
```

Nice to have a string to filter by in the browser

PR #1753, opened 2020-05-08
[web3torrent] Handle user declining budget

snario:
Allows a user to decline the approve budget and fund prompt without causing errors.

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -342,6 +342,7 @@ export class PaymentChannelClient {
       );
     } catch (e) {
       if (e.message === 'User declined') {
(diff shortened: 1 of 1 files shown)


packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -327,12 +327,20 @@ export class PaymentChannelClient {
...
+    try {
+      this.budgetCache = await this.channelClient.approveBudgetAndFund(
+        amount,
+        amount,
+        HUB.signingAddress,
+        HUB.outcomeAddress
+      );
+    } catch (e) {
+      if (e.message === 'User declined') {
+        return;
```
lalexgap, line 339: May a log here to indicate that we handled it?

PR #1761, opened 2020-05-08
Remove REACT_APP_ prefixes everywhere

snario:


packages/web3torrent/.env.development
```diff
@@ -2,10 +2,10 @@ GANACHE_CACHE_FOLDER=../../.ganache-deployments
 GANACHE_PORT=8545
```
kerzhner, line 2: I think it might be time to removet his file.

PR #1762, opened 2020-05-08
Disable logging in persistent-seeder

andrewgordstewart:
This hopefully deals with https://github.com/statechannels/monorepo/issues/1755

### DIFF
--- a/.env.persistent-seeder
+++ b/.env.persistent-seeder
@@ -1,5 +1,6 @@
 ## all
 NODE_ENV = 'production'
+LOG_DESTINATION = 
 
 ## web3torrent, xstate-wallet, rps, simple-hub
 CHAIN_NETWORK_ID = '3'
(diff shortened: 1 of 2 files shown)


kerzhner:
Does this mean there will be no logs for the deployed persistent seeder?

Also, I don't think the persistent seeder is being used by anything at the moment. Given that, do we still believe that logs might be the culprit? If so, would disabling `New Block` log lines be a less drastic intervention?

The reason I ask these questions is that it would be nice to not fly fully blind without logs on the persistent seeder. 🛸

andrewgordstewart:
The persistent seeder logs new blocks every time a new block is mined on Ropsten, at the very least.

Is there a way to access the logs of the persistent seeder? If yes, we can address this more accurately. If no (this was my assumption), then we should just disable all logging.

kerzhner:
> Is there a way to access the logs of the persistent seeder? If yes, we can address this more accurately. If no (this was my assumption), then we should just disable all logging.

I am not sure that there is a way to do so at the moment.

PR #1763, opened 2020-05-08
Don't use dappeteer in circle

andrewgordstewart:
As per our discussion in standup, we should remove sources of failure until the circle integration tests pass reliably.


packages/e2e-tests/puppeteer/__tests__/web3torrent/seed-download.test.ts
```diff
@@ -63,7 +63,7 @@ describe('Web3-Torrent Integration Tests', () => {
     console.log('Loading dapps');
     await forEachTab(async (tab, idx) => {
       await setupLogging(tab, idx, 'seed-download', true);
-      if (!USE_DAPPETEER) await setupFakeWeb3(web3tTabB, idx);
```
kerzhner, line 66: 👀

andrewgordstewart:
(Test failure appears not to be due to the fake web3provider.)

PR #1765, opened 2020-05-10
Revert "W3T Restart Fixing"

andrewgordstewart:
Reverts statechannels/monorepo#1533

After #1533, wires were _destroyed_ when the leecher hit 'Cancel'. This has the consequence of severing communication between the two wallets, meaning the channels cannot close.


andrewgordstewart:
Hopefully we don't have to revert this.  I think https://github.com/statechannels/monorepo/commit/8a89c37660d4adb6c76daab205a296a2eff468ac is enough to fix the issue.

lalexgap:
@andrewgordstewart can this be closed?

PR #1774, opened 2020-05-11
Fix env filtering in w3t webpack config

snario:


spersico:
The problem that I see with this fix, is that when an environment variable gets renamed or added, this other file also need to be updated.

snario:
> The problem that I see with this fix, is that when an environment variable gets renamed or added, this other file also need to be updated.

This is annoying, but consistent with how the other packages work at the moment.

Issue #1721, opened 2020-05-07
Remove challenge button from web3torrent


lalexgap:
Currently challenging is not working for virtual channels, so we should remove the challenge button in the web3torrent UI.

PR #1778, opened 2020-05-11
fix: Remove Challenge Channel button from web3torrent

Alonski:
Virtual channels don't currently support challenging

Closes: #1721


Alonski:
@lalexgap Is this what you had in mind?

PR #1779, opened 2020-05-11
Fix failing nightly deploy of master branch

kerzhner:


snario:
Are multiple `run`s different than `run` with many `&&`?

kerzhner:
> Are multiple `run`s different than `run` with many `&&`?

Every `run` step uses a new shell. On second thought, I think that https://github.com/statechannels/monorepo/pull/1779/commits/3d21284eed79a770a0f103bc47ce6e097e48398d is cleaner syntax.

PR #1781, opened 2020-05-11
Separate cancelling test from download test in E2E tests

snario:
cc @spersico


snario:
I saw `Error: Attempting to sign a stale state` flicker running locally

snario:
Direct funding conclusion implementation leads to a no-op when trying to close channel. As a temporary measure we should implement functionality to support a conclusion state, send out the `closing...` status (so this test passes) and push that code. Related to https://github.com/statechannels/monorepo/issues/1649

andrewgordstewart:
The issue is that the test is saying:
1. Wait for the 10th state
2. Wait a while longer (1500ms)
3. expect that paid in leecher's browser is equal to earned in seeder's browser

3 will not work if they do not have the same current state, so it will _flicker_ (sometimes they will have the same current state)

The test that you want is:
1. Wait for the file to be downloaded
2. expect that paid in leecher's browser is equal to earned in seeder's browser

We can (and probably should) do
1. Wait for the channel to be closed
2. expect that paid in leecher's browser is equal to earned in seeder's browser

If we don't _cancel_ the torrent, then the only way that the channels close is for the download to finish. So, waiting for the channel to be closed will also wait for the download to finish.

geoknee:
@snario Be good for the `e2e-test-w3t-production` job to only run the required tests.
Or at least only `notify_slack` on the required tests.

snario:
> @snario Be good for the `e2e-test-w3t-production` job to only run the required tests.
> Or at least only `notify_slack` on the required tests.

7ec72dbde

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -205,9 +205,13 @@ export class PaymentChannelClient {
...
-    const channelResult = await this.channelClient.closeChannel(channelId);
-    this.updateChannelCache(convertToChannelState(channelResult));
-    return convertToChannelState(channelResult);
+    const isClosed = (channelState: ChannelState) =>
+      channelState.channelId === channelId && channelState.status === 'closed';
+    this.channelClient.closeChannel(channelId);
+
+    return this.channelClient.channelState
+      .pipe(map(convertToChannelState), filter(isClosed), first())
+      .toPromise();
```
spersico, line 214: this fixes the closing channel issue. I stopped waiting for the closeChannel function to resolve, and just listen for the channel to be closed

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -139,7 +139,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
       await this.closeDownloadingChannels(torrent);
-      torrent.destroy(callback);
+      torrent.destroy(() => this.emitTorrentUpdated(infoHash));
```
spersico, line 142: this emits an update for the UI to know when the torrent got removed

PR #1784, opened 2020-05-11
[xstate-wallet] :pig: Add gap when creating final state

andrewgordstewart:
This is a quick fix to get the wallet to wait until there's a conclusion
proof.


lalexgap:
This might still be failing due to the fact that we only dispatch a `CHANNEL_UPDATED` when the supported turn number changes https://github.com/statechannels/monorepo/blob/b1e3a9de5d9227766da370f2c5d4f782297f7839/packages/xstate-wallet/src/workflows/application.ts#L248

PR #1785, opened 2020-05-11
Persistent Seeder Fix

### LINKED ISSUE #1548: [e2e] Persistent seeder
- [x] Write a script that spins up a headful puppeteer/ dappeteer instance, uploads an interesting file (nitro-protocol.pdf, or something), funds a hub connection and sits there earning test ETH (or test wei). 

- [x] Ensure this file is tracked by our tracker, and 

- [ ] appears on the front page of anyone visiting our app. 

- [x] deploy this on heroku or similar

spersico:
**This PR:**
1. Sets the pre-defined torrent to show in the welcome screen to be the same that is shared by the persistent seeder.
2. It also fixes a regression that I found, where the file that was shared, wasn't the correct file (because it would be replaced by the stub file for the tests).
3. It gives more time for [some](https://github.com/statechannels/monorepo/compare/persistent-seeder-set?expand=1#diff-d745dc9ef530a2204bfa4ffd6363d00cR255) [parts](https://github.com/statechannels/monorepo/compare/persistent-seeder-set?expand=1#diff-9ed3cedce43d8747b5dc9942f0b66cedR55) of the seeding process that could take more time and fail.
3. I decided to remove the check in the welcome screen for the persistent seeder torrent because the result wasn't to be trusted. 
The check is done, then, in the **File** page, which simplifies/reduces the number of things the web3torrent does (I've seen some memory leaks in the welcome screen when the torrent is found in the welcome screen).

this PR includes @snario's fix #1741, and closes #1548

### DIFF
--- a/packages/e2e-tests/puppeteer/helpers.ts
+++ b/packages/e2e-tests/puppeteer/helpers.ts
@@ -250,7 +250,11 @@ export async function waitAndApproveDepositWithHub(
 ): Promise<void> {
   console.log('Making deposit with hub');
   const walletIFrame = page.frames()[1];
-  await walletIFrame.waitForSelector('#please-approve-transaction', {timeout: TX_WAIT_TIMEOUT}); // longer timeout here because blockchain is slow
+  console.log('Waiting for #please-approve-transaction');
+  console.time('#please-approve-transaction');
+  await walletIFrame.waitForSelector('#please-approve-transaction', {timeout: TX_WAIT_TIMEOUT * 3}); // longer timeout here because blockchain is slow
+  console.timeEnd('#please-approve-transaction');
+
   await metamask.confirmTransaction({gas: 20, gasLimit: 50000});
   await page.bringToFront();
 }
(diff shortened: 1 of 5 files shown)


packages/e2e-tests/puppeteer/helpers.ts
```diff
@@ -250,7 +250,11 @@ export async function waitAndApproveDepositWithHub(
 ): Promise<void> {
   console.log('Making deposit with hub');
   const walletIFrame = page.frames()[1];
-  await walletIFrame.waitForSelector('#please-approve-transaction', {timeout: TX_WAIT_TIMEOUT}); // longer timeout here because blockchain is slow
+  console.log('Waiting for #please-approve-transaction');
+  console.time('#please-approve-transaction');
+  await walletIFrame.waitForSelector('#please-approve-transaction', {timeout: TX_WAIT_TIMEOUT * 3}); // longer timeout here because blockchain is slow
```
geoknee, line 255: Why would this transaction take 3x longer than any others? Would a change to `TX_WAIT_TIMEOUT` be more appropriate?

PR #1788, opened 2020-05-12
Conclude on my turn

andrewgordstewart:
Resolves https://github.com/statechannels/monorepo/issues/1777
Resolves https://github.com/statechannels/monorepo/issues/1787

Changes
1. When the player instructs the wallet to conclude the application channel, the application workflow
  a. waits for the player's turn
  b. signs a final state with outcome equal to the supported state's outcome
  c. After a (arbitrarily chosen) 30s delay, launches a challenge.
2. The `conclude-channel` workflow _requires_ that the supported state is final
   - This simplifies the implementation of the concluding workflow
   - This also means the logic is the same for all participants
3. The `isFinalized` boolean is fixed (it previously checked that there was a supported final state, but that is not the same as finalized)
   - This change was needed to make the `conclude-channel` idempotency tests to pass
4. Integration tests can opt-in to the dexie backend
  a. Uses [this idea](https://github.com/statechannels/monorepo/issues/1526#issuecomment-618644850)
  b. This was needed because of a weird bug where the state variables were not in reverse-sorted order when fetching the `supported` state.

TODO: 
- [ ] In the idempotency tests, we should be able to easily set up stores with given conditions, rather than running through the `createChannel` workflow.
- [ ] Investigate (4.b) on commit https://github.com/statechannels/monorepo/pull/1788/commits/68599b45907be783415c949c35adf911f2bb1bf4
- [ ] The `channelIsClosing` guard should be changed to check if `supported.isFinal`, not `latest.isFinal`. I made that change, and it broke some more tests, so I would rather fix it in a separate PR.


packages/xstate-wallet/src/workflows/conclude-channel.ts
```diff
@@ -1,41 +1,31 @@
...
-  if (latestSignedByMe.isFinal) {
-    return {state: latestSignedByMe};
-  }
-
-  // Otherwise create a new final state
-  return {state: {...latestSignedByMe, turnNum: latest.turnNum.add(10), isFinal: true}};
+const signFinalState = (store: Store) => async ({channelId}: Init): Promise<void> => {
+  const {supported, latestSignedByMe} = await store.getEntry(channelId);
+  if (!supported.isFinal) throw new Error('Supported state not final');
+  if (latestSignedByMe.turnNum.eq(supported.turnNum)) return; // already signed
```
lalexgap, line 16: Might be worth logging this out.

andrewgordstewart: I thought about this and:
- the person who called `conclude-channel` would _always_ log (unless you didn't log if the supported state was your turn)
- if my wallet crashes after signing the finial state, and restarts, it's _expected_ that it shouldn't sign another state (idempotency)

PR #1792, opened 2020-05-12
Rename Site to Domain across repo

snario:
https://statechannels.slack.com/archives/CMPHW40R0/p1585143349003200


snario:
I'm curious how this will work with existing data in users' IndexedDB. Probably requires a migration.

PR #1793, opened 2020-05-12
Run production tests in band

geoknee:
We do this on `master`, and it should improve the reliability of the test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -404,7 +404,7 @@ jobs:
           file: e2e-test-w3t-production-stats
       - attach_workspace:
           at: /home/circleci/project
-      - run: (cd packages/e2e-tests && yarn jest web3torrent --runInBand)
+      - run: (cd packages/e2e-tests && yarn jest web3torrent --runInBand --bail)
       - upload_logs:
           file: e2e-test-w3t-production-stats
       - notify_slack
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -404,7 +404,7 @@ jobs:
           file: e2e-test-w3t-production-stats
       - attach_workspace:
           at: /home/circleci/project
-      - run: (cd packages/e2e-tests && yarn jest web3torrent)
+      - run: (cd packages/e2e-tests && yarn jest web3torrent --runInBand)
```
snario, line 407: `--bail` too?

snario:
@geoknee why did the test get stuck waiting for "Please Approve Transaction" when `USE_DAPPETEER` is `false`?

geoknee:
That is a prompt in the wallet, not metamask. So it should appeart independent of whether we use Dappeteer.

.circleci/config.yml
geoknee, line 407:
```suggestion
      - run: (cd packages/e2e-tests && yarn jest web3torrent --runInBand --bail)
```

andrewgordstewart:
> @geoknee why did the test get stuck waiting for "Please Approve Transaction" when `USE_DAPPETEER` is `false`?

The hub made a deposit, and neither the hub nor the wallet noticed the transaction. There are no ganache logs, so it's hard to say what happened to it.

PR #1798, opened 2020-05-12
Set CLOSE_BROWSERS to true in virtual-funding

snario:


snario:
When using it I noticed it was causing some issues where the test process wouldn't clean up the Chromium browser and upon testing it a second time I hit bugs because, I suspect, the old Chromium instance (hidden and obscured within the old test process) is still seeding / downloading.

PR #1807, opened 2020-05-12
[web3torrent] Do not rely on a timer for UI updates

kerzhner:
Fixes https://github.com/statechannels/monorepo/issues/1805.

The upsides are:
- More responsive UI instead. With a timer, the UI state lags by at most 1 seconds behind the data state.
- Finegrain control over what events trigger UI updates.

### DIFF
--- a/packages/client-api-schema/src/generated-schema.json
+++ b/packages/client-api-schema/src/generated-schema.json
@@ -592,6 +592,29 @@
       ],
       "type": "object"
     },
+    "DomainBudget": {
+      "additionalProperties": false,
+      "properties": {
+        "budgets": {
+          "items": {
+            "$ref": "#/definitions/TokenBudget"
+          },
+          "type": "array"
+        },
+        "domain": {
+          "type": "string"
+        },
+        "hubAddress": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "domain",
+        "hubAddress",
+        "budgets"
+      ],
+      "type": "object"
+    },
     "EnableEthereumError": {
       "additionalProperties": false,
       "properties": {
@@ -1371,29 +1394,6 @@
           "$ref": "#/definitions/GetChannelsResponse"
         }
       ]
-    },
-    "DomainBudget": {
-      "additionalProperties": false,
-      "properties": {
-        "budgets": {
-          "items": {
-            "$ref": "#/definitions/TokenBudget"
-          },
-          "type": "array"
-        },
-        "domain": {
-          "type": "string"
-        },
-        "hubAddress": {
-          "type": "string"
-        }
-      },
-      "required": [
-        "domain",
-        "hubAddress",
-        "budgets"
-      ],
-      "type": "object"
     },
     "TokenBudget": {
       "additionalProperties": false,
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -417,16 +417,13 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
 
       this.emit(ClientEvents.TORRENT_DONE, {torrent});
       await this.closeDownloadingChannels(torrent);
-<<<<<<< HEAD
       track('Torrent Finished Downloading', {
         infoHash: torrent.infoHash,
         magnetURI: torrent.magnetURI,
         filename: torrent.name,
         filesize: torrent.length
       });
-=======
       this.informListeners();
->>>>>>> web3torrent: remove the update timer
     });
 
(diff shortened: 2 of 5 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -186,6 +217,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       if (!torrent.done && wire.amChoking) {
         wire._clearTimeout();
       }
+      this.informListeners();
```
kerzhner, line 220: I am not sure if this is needed. Does the UI need to be updated on these events?

geoknee: I would say safest to say 'yes', especially given (as we discusssed) as updating the react tree is cheap, and react will decide whether anything actually changed. 

This event *might* impact on the "Number of Peers connected" UI, and it might also be something we want to display in future as it indicates that our peer is still live (an important concept for us).

kerzhner:
> 2. Could the react components simply bind to the underlying webtorrent events directly? You could have a single helper function `onWebTorrentUIEvents(callback: ()=>void)` which will accept a callback from a react component, and then register it to each underlying webtorrent event that we want to trigger a refresh in the UI (it would also return a method to unregister from all of those events). To me, that seems like less indirection overall.

🙏 for the great freedback! Isn't this (roughly) the current implementation. `addListenerForUpdates` accepts a callback from a react component and triggers refreshes. Or am I missing something?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -44,13 +45,35 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+  informListeners() {
+    _.mapValues(this.listenersForUpdates, l => l());
+  }
+
+  addListenerForUpdates(key: string, listener: () => void) {
+    this.listenersForUpdates = {
+      ...this.listenersForUpdates,
+      [key]: listener
+    };
+  }
```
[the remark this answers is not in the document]
kerzhner: yep, that's the idea. Not sure if that is desirable behaviour. Can you think of reasons not to adopt this strategy besides the danger of key sharing between independent components?

geoknee:
I think it is very close, yes, but would not involve the `informListeners()` function. In other words you would not be introducing any new events or keeping track of callbacks at quite the low level that you did so far.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -44,13 +45,35 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+
+  removeListenerForUpdates(key: string) {
+    const listenersForUpdatesCopy = {
+      ...this.listenersForUpdates
+    };
+
+    delete listenersForUpdatesCopy[key];
+    this.listenersForUpdates = listenersForUpdatesCopy;
```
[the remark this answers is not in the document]
kerzhner: This is pretty spot on. React components get an event that indicates that something about the torrent has changed. Each React component can choose what to do with that event. But I would imagine that most will decide to refresh on that event.

packages/web3torrent/src/library/web3torrent-lib.ts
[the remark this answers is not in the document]
geoknee: I think that in some EventEmitters, they use the reference to the listener itself as the key, so the behaviour is different. Besides the danger of overwriting keys, nothing specific occurs to me just now. But there might be something else that EventEmitter library code treats differently. And I think there is some value in having the behaviour here match other event emitters in the app.

kerzhner:
@geoknee, check out 2cba6876fa66615599425519433aa4edf06e811f and 7eeeb2dca5152c60f084fb961950fae0227e84c2.

One thing to note is that, now, the web3torrent client adds and removes per-torrent listeners. Perhaps a better approach is to add listeners to torrents instead. However, my understanding is that a torrent might not exist or (or has not been set up yet?) when the File component is rendered.

packages/client-api-schema/src/generated-schema.json
```diff
@@ -592,6 +592,29 @@
       ],
       "type": "object"
     },
+    "DomainBudget": {
```
geoknee, line 595: Did you mean to check this in?

geoknee:
May-13-2020 18-55-58

packages/client-api-schema/src/generated-schema.json
kerzhner: I ran yarn at the root of the monorepo, and I think this caused the change?

PR #1809, opened 2020-05-13
Use locks when defunding guarantor channels

andrewgordstewart:
The reason that locks weren't working during this morning's live test is ... locks weren't used.

So, we were witnessing what happens when you don't lock the ledger.

This PR acquires a lock when defunding the guarantor.

When testing this out with the multiple leecher test, the locks seemed brittle. (The "stress" test sometimes passed, but sometimes failed.) So, since we are committing to _not_ persisting locks in level 0, I swapped it out for AsyncLock.


packages/xstate-wallet/src/workflows/virtual-defunding-as-leaf.ts
```diff
@@ -108,15 +126,47 @@ export const defundGuarantorInLedger = (store: Store) => async ({
   ]);
   return {state: nextState(ledgerState, outcome)};
 };
+export {defundGuarantorInLedger};
+
+const acquireLock = (store: Store) => async ({ledgerId}: ChannelsSet): Promise<ChannelLock> => {
```
andrewgordstewart: This is copied from ledger-funding, and simplified in https://github.com/statechannels/monorepo/pull/1809/commits/2a557630c9f8a759402329614fc61ff94db01e58

PR #1810, opened 2020-05-13
Enable stress test on circle

andrewgordstewart:
Branched off of https://github.com/statechannels/monorepo/pull/1809.

For some reason, the stress test does not want to work with `SC_ENV=circle-integration-w3t`. I haven't looked into why locally.


snario:
One thing that could cause a failure is that in theory some channels might not reach `turnNum >= 10`. Assuming say D starts downloading and A, B, and C all become seeders before D even starts, then three channels could start and end at `turnNum` 9. I'm not sure what the expected number of turns required is for the 12 MB demo file we use.

snario:
Applying this diff (removing cancelling) seems to get the test to pass when I SSH in.

```diff
diff --git a/packages/e2e-tests/puppeteer/__tests__/web3torrent/stress.test.ts b/packages/e2e-tests/puppeteer/__tests__/web3torrent/stress.test.ts
index 5cf3e4d9f..99b88f589 100644
--- a/packages/e2e-tests/puppeteer/__tests__/web3torrent/stress.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/web3torrent/stress.test.ts
@@ -88,10 +88,10 @@ describe('One file, three leechers, one seeder', () => {
     await forEachActor(({tab}) => waitAndOpenChannel(USES_VIRTUAL_FUNDING)(tab));
 
     console.log('Downloading');
-    await forEachActor(({tab}) => waitForNthState(tab, 10));
+    //await forEachActor(({tab}) => waitForNthState(tab, 10));
 
-    console.log('C cancels download');
-    await cancelDownload(actors.C.tab);
+    //console.log('C cancels download');
+    //await cancelDownload(actors.C.tab);
 
     console.log('Waiting for channels to close');
     await forEachActor(({tab}) => waitForClosedState(tab));
```

andrewgordstewart:
> One thing that could cause a failure is that in theory some channels might not reach `turnNum >= 10`. Assuming say D starts downloading and A, B, and C all become seeders before D even starts, then three channels could start and end at `turnNum` 9.
> 
> I'm not sure what the expected number of turns required is for the 12 MB demo file we use, so I could be wrong. Maybe we need like 300 turns.

Good point, I had forgotten about this. An abandoned change made the file 120 MB for this test.

A good starting point for this test on circle might be to just let them all run to completion, and then   add some assertions to check for [this bug](https://github.com/statechannels/monorepo/issues/1797)

snario:
@andrewgordstewart rebased on `master` and commented out `cancelDownload` (causes puppeteer to crash) and the test passes.

andrewgordstewart:
Done: https://github.com/statechannels/monorepo/pull/2022/commits/3c7120c78c50168f70b682089b81a3eb58afd3f4

PR #1812, opened 2020-05-13
Point save_lib to correct directories

snario:
Some of these paths did not exist, some were missing.


snario:
@andrewgordstewart was this meant to also cache things in `build` dirs?

andrewgordstewart:
> @andrewgordstewart was this meant to also cache things in `build` dirs?

It could. I think if we were to put some work into this, we could save some time on the prepare step (maybe 30-60s?).

PR #1818, opened 2020-05-13
Add --ignore-optional to yarn on CircleCI

snario:


kerzhner:
I don't think this has any effect. `--ignore-optional` is only functional for `yarn add`, not `yarn install`

PR #1822, opened 2020-05-13
Add separate source map file for web3torrent

snario:


snario:
`SyntaxError: Unexpected token ':'` wat

PR #1825, opened 2020-05-13
Update Sentry and add to Wallet

snario:


packages/rps/package.json
```diff
@@ -65,7 +66,7 @@
   },
   "devDependencies": {
     "@babel/runtime": "7.8.3",
-    "@sentry/browser": "5.12.1",
+    "@sentry/browser": "5.15.5",
```
andrewgordstewart, line 69:
```suggestion
```
It's added as a dependency above

PR #1827, opened 2020-05-13
[web3torrent] simplify use-budget

kerzhner:
- The hook will only run when `ready` or `paymentChannel` clients change instead of after every render.
- We are able to eliminate `doInitialFetch` state.


kerzhner:
Closing since https://github.com/statechannels/monorepo/pull/1830 is approved and includes these changes.

PR #1829, opened 2020-05-13
[web3torrent] Expose downloadable pino logs

andrewgordstewart:
If I call `window.saveWeb3torrentLogs()`, my browser downloads a file like this:
[web3torrent.Wed, 13 May 2020 18_52_55 GMT.f4cb9bbd3.pino.log](https://github.com/statechannels/monorepo/files/4624057/web3torrent.Wed.13.May.2020.18_52_55.GMT.f4cb9bbd3.pino.log)

## Future improvements:
- [ ] Extract this pattern to a devtools utility
- [ ] Use this pattern in channel-provider and xstate-wallet
- [ ] Use this utility to download artifacts in integration tests

### DIFF
--- a/packages/web3torrent/src/logger.ts
+++ b/packages/web3torrent/src/logger.ts
@@ -1,5 +1,6 @@
 import pino from 'pino';
-import {LOG_DESTINATION, ADD_LOGS, LOG_LEVEL} from './constants';
+import {LOG_DESTINATION, ADD_LOGS, LOG_LEVEL, VERSION} from './constants';
+import _ from 'lodash';
 
 // TODO: Is there a better way to determine if we're in a browser context?
 const IS_BROWSER_CONTEXT = process.env.NODE_ENV !== 'test';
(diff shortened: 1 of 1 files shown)


packages/web3torrent/src/logger.ts
```diff
@@ -28,15 +29,72 @@ const getCircularReplacer = () => {
...
+  }
+}
+
+const logBlob = new LogBlob();
+
+const serializeLogEvent = o => JSON.stringify({...o, name}, getCircularReplacer());
+
+let browser: any = IS_BROWSER_CONTEXT
+  ? {
+      transmit: {
```
kerzhner, line 54: What calls `transmit`?

andrewgordstewart: Transmit is a configuration option of pino's browser API. The `send` callback is called each time the Pino logger is writing a log.

If you were to hook Pino up to data dog or something, you could fire off an HTTP request in this callback as well.

PR #1543, opened 2020-04-24
[xstate-wallet] Allow directly funded channels to be withdrawn

geoknee:
...and test this in both `rps` and `w3t` e2e tests.

PR #1834, opened 2020-05-13
Add withdrawing to Direct Funding Conclude Channel

lalexgap:
Based on @geoknee's work in #1543. Adds withdrawal to conclude channel workflow. To prevent both players from submitting `finalizeAndWithdraw` transaction and have it fail for one player, only player A currently triggers the transaction.

TODO:
- [x] Update UI to show transactions status similar to approve budget and fund (separated to #1840)
- [x] Look into flickering `w3t` integration test (This may have been resolved by using indexedDb for the `w3t` test)

### DIFF
--- a/.env.circle-integration-w3t
+++ b/.env.circle-integration-w3t
@@ -16,7 +16,6 @@ TARGET_NETWORK = 'localhost'
 WEB3TORRENT_URL = 'http://localhost:3000'
 SCREENSHOT_DIR = 'screenshots'
 
-HUB_CHAIN_PK='0x0'
 
 ## xstate-wallet
 # This account corresponds to private key 0x187bb12e927c1652377405f81d93ce948a593f7d66cfba383ee761858b05921a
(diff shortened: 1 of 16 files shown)


.env.circle-integration-w3t
```diff
@@ -16,12 +16,15 @@ TARGET_NETWORK = 'localhost'
...
 SCREENSHOT_DIR = 'screenshots'
 
+
 ## xstate-wallet
 # This account corresponds to private key 0x187bb12e927c1652377405f81d93ce948a593f7d66cfba383ee761858b05921a
 # That key is provided eth in @statechannels/devtools/src/constants.ts
 HUB_DESTINATION = '0x0000000000000000000000008199de05654e9afa5C081BcE38F140082C9a7733'
 XSTATE_WALLET_DEPLOYER_ACCOUNT_INDEX = '3'
 LOG_DESTINATION = 'pino.log'
+USE_INDEXED_DB = 'true'
```
snario, line 26: We did consciously omit this previously so we had at least one integration test using the memory backend, but it's probably not that important

geoknee: I think that's probably not the right way to test the memory backend (I'm just thinking about an exponential blowup of test configurations). We could discuss at standup.

PR #1838, opened 2020-05-14
Use CRA templates for xstate-wallet

snario:
I keep wanting to make tweaks to the webpack config of the `xstate-wallet` and there tend to be weird differences between how it is set up and how create react app sets things up. Also CRA has many optimizations we can take advantage of for setting up a production app (e.g., code splitting, forked typescript processes, standard setups that are google-able easily, explicit configuration settings).

This is 99% copy pasta from web3torrent with the following changes:

- A plugin to use `const enum`
- eslint commented out (btw web3torrent does not use _our_ .eslintrc anymore I noticed)
- Change `baseUrl` of `xstate-wallet` `tsconfig`
- Put `GIT_*` vars in `process.env` instead of `DefinePlugin`
- Kept `keep_fnames`


lalexgap:
I personally prefer the simpler `webpack.config` where we don't have a lot of configuration going on that we don't understand.

However if people prefer to use the CRA `webpack.config` I don't feel too strongly about it.

snario:
I also prefer the simpler one, but the discrepancy across packages has been a bit annoying

kerzhner:
> * eslint commented out (btw web3torrent does not use _our_ .eslintrc anymore I noticed)

Is this something that we should change?

snario:
> > * eslint commented out (btw web3torrent does not use _our_ .eslintrc anymore I noticed)
> 
> Is this something that we should change?

Correction: the webpack dev server does not check our config, but `lint:check` does

PR #1848, opened 2020-05-14
Turn off e2e-test-w3t-production alert

kerzhner:
The job has never passed. Once the job is passing, turn on the alert back on.


kerzhner:
https://github.com/statechannels/monorepo/issues/1849 to remind us to turn the alert back on.

PR #1857, opened 2020-05-14
Add valid app transition code in wallet using synchronous EVM WASM

snario:
Notes:
- Previous usage of `pure-evm` here: https://github.com/statechannels/monorepo/commit/f2d2e3e7e2e0a13e6dad75950617236658176fa7
- Issue on `ethereumjs-vm` for synchronous `runCode` https://github.com/ethereumjs/ethereumjs-vm/issues/677


andrewgordstewart:
Screen Shot 2020-05-14 at 3 53 28 PM
Screen Shot 2020-05-14 at 3 54 42 PM

True to his word...

packages/xstate-wallet/src/store/channel-store-entry.ts
```diff
@@ -127,7 +136,28 @@ export class ChannelStoreEntry {
...
       );
     }
-    return true;
+
+    // JS implementation of App check
+    const fromVariablePart = getVariablePart(toNitroSignedState(firstState)[0].state);
+    const toVariablePart = getVariablePart(toNitroSignedState(secondState)[0].state);
+    const turnNumB = secondState.turnNum;
+
+    const ConsensusAppArtifact = require('@statechannels/nitro-protocol/build/contracts/ConsensusApp.json');
```
andrewgordstewart, line 145: Is there a way to make this channel-agnostic? Is it crazy for the app to provide the artifact to the wallet?

snario: Previously I had envisioned that app definitions would be like chrome extensions. A user must specifically accept / acknowledge adding a new one to their wallet, but the wallet might also come pre-packaged with some trusted ones

andrewgordstewart: So, the user can accept rules that are:
- provided by some URL, or the app, or whatever
- deployed at address `0x123`

The wallet fetches the contract from the URL or whatever, and verifies the bytecode of the contract matches the bytecode deployed at `0x123` in the selected blockchain?

snario: Yeah

packages/xstate-wallet/src/store/channel-store-entry.ts
```diff
@@ -9,18 +9,29 @@ import {
...
 import {Errors} from '.';
 import {logger} from '../logger';
+import {getVariablePart} from '@statechannels/nitro-protocol';
 export interface SignatureEntry {
   signature: string;
   signer: string;
 }
 
+// TODO: Pull out hard-coded app and store app artifacts somewhere by appDefinition
+import ConsensusAppArtifact from '@statechannels/nitro-protocol/build/contracts/ConsensusApp.json';
```
lalexgap, line 32: Are we still using the consensus app? I thought we had switched to null channels?

snario: Well, it's actually worse than that. RPS uses RPS, Web3Torrent uses SingleAssetPayment

lalexgap: Yeah. I think we could probably split that up into two chunks of work. This PR for validating null channels (which requires no external contracts) and then a PR to do something like discussed above (https://github.com/statechannels/monorepo/pull/1857#discussion_r425494473) (or you could do it all in one PR).

Either way I think this reference to ConsensusApp can go :)

packages/xstate-wallet/src/store/channel-store-entry.ts
```diff
@@ -115,19 +126,49 @@ export class ChannelStoreEntry {
...
+        Uint8Array.from(Buffer.from(txData.substr(2), 'hex'))
       );
-    }
-    return true;
+
+      return defaultAbiCoder.decode(['bool'], result)[0] as boolean;
+    };
+
+    const isValidNitroTransition = () => {
+      // JS implementation of Nitro check
```
geoknee, line 151: Do you think we ought to use `PureEVM` to do this check, too?

I don't think we currently expose `_requireValidTransition` as a public function in `ForceMove.sol`. But I think probably we should?

snario: It would be nice to do that, but less important since it's hard-coded into the protocol and the lines of code required are quite small.

snario:
Given the tests are passing and ConsensusApp is hard-coded, I'm somewhat skeptical this is working as expected right now.

packages/xstate-wallet/src/store/channel-store-entry.ts
geoknee: I agree. But: 
- it would actually simplify this code since you would only ever hit the adjudicator and it would call out to the app contract. 
- it means we are using `PureEVM` for as much of this checking as possible,  do not have to do any of our own transcription in javascript, and do not have to worry about keeping that in sync with the contracts (should they change).

snario: I'm not 100% certain that `PureEVM` is capable of doing that `STATICCALL` synchronously. I'd need to figure that out ... @armaniferrante do you know if it would work? We'd need to be passing in the bytecode of the `NitroAdjudicator` _and_ the app being operated on somehow.

snario: It seems possible but `PureEVM` might not do it. If it doesn't then maybe we need to inherit the library and add that feature. Maybe make an issue @geoknee?

armaniferrante: The short answer: yes, I would expect pure-evm to execute the desired codepath for a `STATTICCALL`. However, it doesn't do any nuanced checks differentiating `STATTICCALL` from non stattic calls. Rather, any regular calls that try to access state should just fail.

Some background info incase it's helpful:

`PureEVM` is takes parity's [evm package](https://github.com/openethereum/openethereum/tree/234a2877be91a9beecde89e4177558c6c6689231/ethcore/evm), which executes evm bytecode, and swaps its [implementation](https://github.com/openethereum/openethereum/blob/1ace3c7731a9bb64994fdf357cef39417c33f4e0/ethcore/machine/src/externalities.rs#L130) of the [`Ext` trait](https://github.com/openethereum/openethereum/blob/master/ethcore/vm/src/ext.rs#L67) for the  dumb (pure one)[https://github.com/armaniferrante/pure-evm/blob/master/src/pure_ext.rs#L20].

The `Ext` trait is the interface between the ethereum bytecode interpreter and its host environment. This is how syscall-like opcodes like access to storage, calls, creates, blockchash, chain_id, etc are implemented. Given that `PureEVM` is meant to only execute pure function calls, the host environment is effectively null, as you can see by the implementation.

When executing calls/static calls, it looks like the parity client does checks both inside the bytecode interpreter and also inside the host environment, failing if the STATTICCALL semantics are not upheld (i.e., if the tx changes the state). For example, inside the interpreter see [here](https://github.com/openethereum/openethereum/blob/1ace3c7731a9bb64994fdf357cef39417c33f4e0/ethcore/machine/src/externalities.rs#L147). And inside the host environment, see [here](https://github.com/openethereum/openethereum/blob/234a2877be91a9beecde89e4177558c6c6689231/ethcore/machine/src/executive.rs#L423).

If you need nuanced checks like those, you'll want to plugin your own externalities (`Ext` trait).

armaniferrante: Mind giving some background on the check you're trying to put into `PureEVM`, I don't think I fully understand what you're trying to accomplish.

snario: @armaniferrante there is a function called [`_requireValidTransition`](https://github.com/statechannels/monorepo/blob/da1a4acecf919cdfcaaa8a689d9cdaab3336d35e/packages/nitro-protocol/contracts/ForceMove.sol#L477-L488) that makes an external call, that is guaranteed to be a `STATICCALL`, that @geoknee is recommending we provide to `pure-evm`. This function does some basic checks that apply to _all_ possible applications, _and_ it makes the external app-specific call.

armaniferrante: I see, so you want to enforce the checks provided by `_requireValidTransition` by executing that function inside `PureEVM` and then making an external call within the interpreter. 

That's not going to work. `PureEVM` has no knowledge of any bytecode at any address. It only knows the bytecode and data you give it. You'd have to provide a database/mapping from address to bytecode by implementing the required externalities (`Ext` trait).

armaniferrante: Do you need access to the bytecode of all addresses in the ethereum state? If you do, you should probably just run an actual parity client and call out to it (though then you're dealing with async calls). If the bytecode to be called is known beforehand, you could hack it into pure-evm to have it all synchronous/embedded.

snario: No, we can assume all relevant app's bytecode will be pre-loaded into the wallet.

snario: @geoknee we could also just make two calls, one for the generic Nitro rules and one for the app rules.

armaniferrante: Rougly, here's what you'd need to do:

In [`PureExt`](https://github.com/armaniferrante/pure-evm/blob/master/src/pure_ext.rs#L76), access the bytecode of the contract your calling (probably by calling out from Rust to JavaScript, which then accesses the bytecode from whatever storage you have and returns it to Rust) and then make a recursive call to pure-evm (in Rust) with your bytecode and data.

tomclose:
Closing due to inactivity.

PR #1858, opened 2020-05-14
web3torrent: fix PEER_TRUST

kerzhner:
1. At the moment, the leecher always pays for `wire.requests.length` blocks, instead of `PEER_TRUST` blocks. See proof below.
1. There is an assumption that the last piece is the first to be downloaded with the `torrent.downloaded === 0 ` condition.

This PR fixes both of those.

Proof of point 1.
```
// Current logic
let amountToPay = BUFFER_REFILL_RATE.sub(
      WEI_PER_BYTE.mul(BLOCK_LENGTH).mul(PEER_TRUST - wire.requests.length)
    );

// Substitute BUFFER_REFILL_RATE definition
const amountToPay2 = utils
      .bigNumberify(WEI_PER_BYTE.mul(BLOCK_LENGTH * PEER_TRUST))
      .sub(WEI_PER_BYTE.mul(BLOCK_LENGTH).mul(PEER_TRUST - wire.requests.length));

// What is the above in bytes (instead of wei)
const bytesToPayFor = utils
      .bigNumberify(BLOCK_LENGTH * PEER_TRUST)
      .sub(BLOCK_LENGTH * (PEER_TRUST - wire.requests.length));

// What is the above in blocks, instead of bytes
const blocksToPayFor = utils.bigNumberify(PEER_TRUST).sub(PEER_TRUST - wire.requests.length);

// Big numbers are confusing! And not needed here :)
const blocksToPayFor2 = PEER_TRUST - (PEER_TRUST - wire.requests.length);

// Algebra is fun without a bunch of wrappers!
const blocksToPayFor3 = wire.requests.length;
```


spersico:
I would add in the [DONE listener](https://github.com/statechannels/monorepo/blob/693658ab9e55f3bb5208c5c6e6a43a6495f4ca36/packages/web3torrent/src/library/web3torrent-lib.ts#L401-L402) ( before closing the channels).
Something that logs the final channels status, and the amount paid to each wire, before closing it: I did this in local, for example:
```
log.info({torrent, channels: this.paymentChannelClient.channelCache}, '<< Torrent DONE!');
      torrent.wires.map(wire => {
        const {peerAccount, peerChannelId} = wire.paidStreamingExtension;
        const channelOfWire = this.paymentChannelClient.channelCache[peerChannelId];
        if (channelOfWire) {
          const balance = bigNumberify(channelOfWire.beneficiaryBalance).toString();
          log.info(`<< ${peerAccount} Final Balance: ${balance} Downloaded ${wire.downloaded}`);
        }
      });
```
This would give us a final status check:
image

But I tested it locally, and it worked okay everytime!. Merge away!

kerzhner:
> I would add in the DONE listener

@spersico good idea. Let's do that in a separate PR.

PR #1868, opened 2020-05-15
[xstate-wallet] log context on transition

andrewgordstewart:


andrewgordstewart:
Merging for posterity.

This change will make the logs bigger, and eventually we will want to pare them down. But in the short term, they are useful for figuring out what's going wrong.

In this case, the `context` and `event` in the failure [here](https://app.circleci.com/pipelines/github/statechannels/monorepo/6518/workflows/7483821c-6bb8-40b8-9353-3482a2bacd54/jobs/27653/steps) made it obvious to me that [this code](https://github.com/statechannels/monorepo/blob/master/packages/xstate-wallet/src/chain.ts#L257-L267) is probably never getting executed, meaning we should probably tackle https://github.com/statechannels/monorepo/issues/1394 and reduce the fragility of the `window.ethereum` related workflows.

PR #1869, opened 2020-05-15
[rps] Extend e2e test to 2 games, each player resigns

geoknee:


packages/rps/src/redux/game/saga.ts
```diff
@@ -129,7 +129,8 @@ function* gameSagaRun(client: RPSChannelClient) {
         !cs.isClosing(channelState) &&
         !cs.isClosed(channelState)
       ) {
-        yield* closeChannel(channelState, client);
+        // TODO do not fork this call (when the wallet is fixed and the call resolves properly)
+        yield fork(closeChannel, channelState, client);
```
geoknee, line 133: On approval, I will make an issue to track unforking these calls.

PR #1871, opened 2020-05-15
[ xstate-wallet] closeChannel promise resolution

geoknee:
Currently the `closeChannel` call appears to not ever resolve. This PR involves some changes to the wallet behaviour around handling those requests.  A side effect is that the call does now resolve if it is your turn and you successfully signed an `isFinal` state. Furthermore, the wallet sends a new `code: 300` error response if you make the call when it is not your turn (which should cause the promise to reject). 

- the `closeChannel` call is no longer forked in RPS (i.e. we block on it)


packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -220,13 +220,13 @@ const generateConfig = (
 
     //This could handled by another workflow instead of the application workflow
     closing: {
-      entry: [actions.assignRequestId],
+      entry: [actions.assignRequestId, actions.sendCloseChannelResponse],
```
geoknee, line 223: This might be a bad idea because (I think) the second action depends on first completing?

andrewgordstewart: I think it works but that doesn't make it a good idea. (I think the `interpreter` currently exported by xstate bumps `assign` actions to first in the queue.)

`assignRequestId` should probably happen as an `exit` action in the transition _into_ `closing`.

That's more clear, because the event is right there, meaning it's obvious that the event has a `requestId` property.

And, it's more robust, because we know that the request ID is assigned to the context before the `sendCloseChannelResponse`

## Alternative

The `sendCloseChannelResponse` is _important_, so maybe it should be a _service_, not an _action_. So, maybe when you enter the `closing` state, the first thing you do is
`{ invoke: { src: 'sendCloseChannelResponse', onDone: 'doTheNextThing' } }`

geoknee: Thanks for this @andrewgordstewart . Pairing with @tomclose, we refactored so that these actions happen either side of signing the `isFinal` state. We also allowed for the request to get an error response if you made the API call when it was not your turn.

packages/client-api-docs/source/index.html.md
```diff
@@ -448,6 +448,10 @@ Possible response to a `Channel Proposed` event.
 
 ## Close Channel
 
+The wallet will respond to this request with an error if it is not your turn. If it is your turn, the wallet will respond as soon as it has signed an `isFinal` state, and the channel is updated to `closing` status.
+
+The channel may later update to `closed` status only when other channel participants have responded in kind: this can be detected by listening to `ChannelUpdated` events and filtering on the channel status.
+
```
geoknee, line 454: 👀

packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -208,29 +206,36 @@ const generateConfig = (
...
     },
 
-    signingFinalState: {
-      invoke: {src: signFinalState.name, onDone: 'closing'},
+    attemptToSignFinalState: {
+      invoke: {
+        src: signFinalStateIfMyTurn.name,
+        onDone: {target: 'closing', actions: [actions.sendCloseChannelResponse]},
+        onError: {target: 'running', actions: [actions.sendCloseChannelRejection]}
+      },
```
geoknee, line 223: 👀

PR #1876, opened 2020-05-16
web3torrent: deadlock when a blocked leecher receives a stop signal

kerzhner:
The initial goal of this PR was to remove asynchronicity in `stop` and `start` in order to make it easier to reason about those methods. After doing that, I came to the following realization. Let's think about how blocking works on the seeder.
1. Let's assume that the seeder has blocked 5 requests.
1. Now, the seeder receives a payment for 4 requests.
1. In the current logic, the seeder tries to service the blocked requests. It sends data for 4 requests. On the 5th requests, the seeder does not service the requests and instead sends a request for a payment aka the stop signal.
1. After that, the seeder sends a `PaidStreamingExtensionNotices.START` aka start signal.

This deadlocks the exchange. This is due to the leecher ignoring the stop signal when already in blocked state: https://github.com/statechannels/monorepo/blob/38092c16ec06b19d7e74c6deff877b0a452e153f/packages/web3torrent/src/library/paid-streaming-extension.ts#L107-L112

In the new logic, the seeder always sends a start signal immediately after processing a payment. The new logic is:
1. Let's assume that the seeder has blocked 5 requests.
1. Now, the seeder receives a payment for 4 requests.
1. The seeder sends a start signal to the leecher.
1. The seeder tries to service the blocked requests. It sends data for 4 requests. On the 5th requests, the seeder does not service the requests and instead sends a request for a payment aka the stop signal. While servicing the blocked requests, more requests could have come in from the leecher.


packages/e2e-tests/puppeteer/helpers.ts
```diff
@@ -217,8 +217,9 @@ export async function waitForBudgetEntry(page: Page): Promise<void> {
   await page.waitForSelector('.domain-budget-table > tbody > tr');
 }
 
+// For local runs, the default timeout of 30 seconds is not sufficient for a test with a 13MB file exchange.
```
snario, line 220: Why 13MB?

kerzhner: That's how large our current `web3torrent-tests-stub` test file is. This comment can probably be better worded...

snario: Oh, I shrunk it in #1877

kerzhner: Thanks for the heads up! 38092c16e

PR #1877, opened 2020-05-16
Decrease size of e2e w3t file

snario:
[This](https://app.circleci.com/pipelines/github/statechannels/monorepo/6548/workflows/c71f67cf-a559-4388-ae32-81a05cb83bc5/jobs/27873/steps) job failed because 30 seconds wasn't enough for the download. I don't think we lose much going from 12 MB to ~1.3 MB.


snario:
#1878 also just failed halfway. I wonder if we slowed it down recently? This seems like a recent issue? 
image

kerzhner:
The downloads have gotten slower since we now [respect the `PEER_TRUST`](https://github.com/statechannels/monorepo/pull/1858) constant in the web3torrent app. The constant is currently set to 4. This means that the leecher pays for at most 4 blocks at a time.

PR #1878, opened 2020-05-16
Remove dappeteer usage for e2e Ropsten tests

snario:


packages/xstate-wallet/src/chain.ts
```diff
@@ -240,6 +240,13 @@ export class ChainWatcher implements Chain {
   private get signer() {
     if (!this.ethereumIsEnabled) throw new Error('Ethereum not enabled');
 
+    if (window.ethereum.isFakeWeb3) {
+      return new Wallet(
+        '0xccb052837ccafb700e34c0e0cc0f3e5fbee8f078f3fe6b4e5950c7c8acaa7bce',
```
andrewgordstewart, line 245: Should this be configured via an ENV variable, with default value `'0xccb052837ccafb700e34c0e0cc0f3e5fbee8f078f3fe6b4e5950c7c8acaa7bce'`?

packages/e2e-tests/puppeteer/helpers.ts
```diff
@@ -117,27 +124,36 @@ class FakeMetaMask implements dappeteer.Dappeteer {
...
   // We need to instantiate a web3 for the wallet so we import the web 3 script
   // and then assign it on the window
   const web3JsFile = fs.readFileSync(path.resolve(__dirname, 'web3/web3.min.js'), 'utf8');
   await page.evaluateOnNewDocument(web3JsFile);
   await page.evaluateOnNewDocument(`
-    window.web3 = new Web3("http://localhost:${ganachePort}");
+    window.web3 = new Web3(new Web3.providers.HttpProvider("${RPC_ENDPOINT}"));
+    ${
+      TARGET_NETWORK === 'ropsten'
+        ? 'window.web3.eth.accounts.wallet.add("0xccb052837ccafb700e34c0e0cc0f3e5fbee8f078f3fe6b4e5950c7c8acaa7bce");'
```
andrewgordstewart, line 136: Should this be configured via an ENV variable, with default value `'0xccb052837ccafb700e34c0e0cc0f3e5fbee8f078f3fe6b4e5950c7c8acaa7bce'`?

packages/xstate-wallet/src/chain.ts
snario: Yeah I was going to add that in a separate PR. We currently use this hard-coded value as it's [the default](https://github.com/decentraland/dappeteer/blob/85a52146942613daa3a8e5a102f3bbe8c6f32f0f/src/index.ts#L64) dappeteer account.

snario: Actually it should be a secret because technically anyone on the internet can drain the ETH in this account.

PR #1884, opened 2020-05-17
Rename app-wallet-interface to client-api-docs

snario:
We have `client-api-schema`, and these are the `client-api-docs`


snario:
By the way, the site is only designed to be built on one branch right now: `george/add-provider-api-package`. Perhaps change this to `deploy`? https://app.netlify.com/sites/app-wallet-interface/settings/deploys

PR #1889, opened 2020-05-18
Remove potential unnecessary call to UpdateChannel

snario:


geoknee:
Do you expect this to improve/solve https://github.com/statechannels/monorepo/issues/1883?

snario:
I don't think so actually. However, it does seem like an improvement.

Issue #1883, opened 2020-05-17
Duplicate JSONRPC requests


andrewgordstewart:
> [Attempting to sign a stale state](https://app.circleci.com/pipelines/github/statechannels/monorepo/6569/workflows/70709aef-b443-4611-90cd-a70aeb7a8b03/jobs/28023/parallel-runs/0/steps/0-107) 👀

```
❯ grep UpdateChannel seed-download.pino.log | jq -c '{msg: .msg, id: .jsonRpcRequest.id, time: .time}' | pino-pretty | pbcopy

[1589673253282]: INCOMING JSONRPC REQUEST:
    id: 7
[1589673253361]: INCOMING JSONRPC REQUEST:
    id: 9
[1589673253381]: INCOMING JSONRPC REQUEST:
    id: 10
[1589673253447]: INCOMING JSONRPC REQUEST:
    id: 9
```

The channel client/provider is making the same request twice.

_Originally posted by @andrewgordstewart in https://github.com/statechannels/monorepo/pull/1882#issuecomment-629840861_

snario:
@andrewgordstewart don't we expect duplicates since those logs are for 2 separate browsers?

andrewgordstewart:
Good point

PR #1890, opened 2020-05-18
[xstate-wallet] Add transaction-based updateChannel to Store

snario:
Possibly might help with #1883, specifically the symptom that led to the creation of the issue: `Attempting to sign a stale state`.

### DIFF
--- a/packages/xstate-wallet/src/workflows/application.ts
+++ b/packages/xstate-wallet/src/workflows/application.ts
@@ -16,7 +16,7 @@ import {MessagingServiceInterface} from '../messaging';
 import {filter, map, distinctUntilChanged, first} from 'rxjs/operators';
 import {createMockGuard, unreachable} from '../utils';
 
-import {Store} from '../store';
+import {Store, SimpleAllocation} from '../store';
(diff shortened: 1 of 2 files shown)


snario:
Copy-pasted `signAndAddState` code because I'm unfamiliar with how wrapping two transaction-based methods with `.then()` chains works.

packages/xstate-wallet/src/workflows/application.ts
```diff
@@ -131,8 +131,12 @@ const signFinalState = (store: Store) => async ({channelId}: ChannelIdExists) =>
     .toPromise();
   // Sign a finalized state
   const {supported} = ourTurnEntry;
-  const finalState = {...supported, turnNum: supported.turnNum.add(1), isFinal: true};
-  return store.signAndAddState(channelId, finalState);
+  const finalState = {
+    outcome: supported.outcome as SimpleAllocation,
```
snario, line 135: Might need some runtime check here. Or, we could disable `MixedAllocation` at the type-level for now since we don't support it.

snario:
Interestingly, seems like data exchange stalls out.

PR #1893, opened 2020-05-18
Use JEST_WORKER_ID to detect IS_BROWSER_CONTEXT

snario:
Logs are missing in [important jobs](https://app.circleci.com/pipelines/github/statechannels/monorepo/6600/workflows/ba81f5a6-137a-4813-b112-5b71fdc7cc68/jobs/28304/artifacts) because of an awkward environment variable.

### DIFF
--- a/packages/channel-provider/src/logger.ts
+++ b/packages/channel-provider/src/logger.ts
@@ -6,7 +6,7 @@ const LOG_TO_CONSOLE = process.env.LOG_DESTINATION === 'console';
 // eslint-disable-next-line no-undef
 const LOG_TO_FILE = process.env.LOG_DESTINATION && !LOG_TO_CONSOLE;
 // eslint-disable-next-line no-undef
-const IS_BROWSER_CONTEXT = process.env.NODE_ENV !== 'test';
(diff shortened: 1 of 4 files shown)


packages/channel-provider/src/logger.ts
```diff
@@ -6,7 +6,7 @@ const LOG_TO_CONSOLE = process.env.LOG_DESTINATION === 'console';
 // eslint-disable-next-line no-undef
 const LOG_TO_FILE = process.env.LOG_DESTINATION && !LOG_TO_CONSOLE;
 // eslint-disable-next-line no-undef
-const IS_BROWSER_CONTEXT = process.env.NODE_ENV !== 'test';
+const IS_BROWSER_CONTEXT = process.env.JEST_WORKER_ID === undefined;
```
kerzhner, line 9: Is this safer?
```suggestion
const IS_BROWSER_CONTEXT = !process.env.JEST_WORKER_ID;
```

kerzhner: I am not sure if there are cases when `JEST_WORKER_ID` can end up as `null`

PR #1894, opened 2020-05-18
Set more sensible defaults for gasLimit

snario:
We should figure out what these should be defaulted as using the gas study @geoknee did. For now though, 3M seems way too high.

### DIFF
--- a/packages/nitro-protocol/src/contract/transaction-creators/force-move.ts
+++ b/packages/nitro-protocol/src/contract/transaction-creators/force-move.ts
@@ -10,7 +10,7 @@ import {getFixedPart, getVariablePart, hashAppPart, State} from '../state';
 // TODO: Currently we are setting some arbitrary gas limit
 // To avoid issues with Ganache sendTransaction and parsing BN.js
 // If we don't set a gas limit some transactions will fail
-const GAS_LIMIT = 3000000;
+const GAS_LIMIT = 100_000;
 
 // @ts-ignore https://github.com/ethers-io/ethers.js/issues/602#issuecomment-574671078
 export const ForceMoveContractInterface = new ethers.utils.Interface(ForceMoveArtifact.abi);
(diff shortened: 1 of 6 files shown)


snario:
e.g., [depositing used 34,384](https://ropsten.etherscan.io/tx/0x8de76717e5ac74b424b2ff7fcc13e27bcc6f409f4d2cbb7202a68b37838b5997) while being provided 3,000,000

packages/nitro-protocol/src/contract/transaction-creators/force-move.ts
```diff
@@ -10,7 +10,7 @@ import {getFixedPart, getVariablePart, hashAppPart, State} from '../state';
 // TODO: Currently we are setting some arbitrary gas limit
 // To avoid issues with Ganache sendTransaction and parsing BN.js
 // If we don't set a gas limit some transactions will fail
-const GAS_LIMIT = 3000000;
+const GAS_LIMIT = 100_000;
```
geoknee, line 13: This seems too low to me. The gas scales with the number of states submitted and the complexity of the `ForceMoveApp.validTransition`.  One our test cases already used 137542 gas.

Questions: 

1. What is the downside of having a conservative `GAS_LIMIT` with a lot of headroom?
2. Can we use something to estimate an appropriate `GAS_LIMIT` and not use a constant? EVM-emulation?

geoknee: https://28593-209829093-gh.circle-artifacts.com/0/nitro-protocol-gas.json

snario: These are just placeholders. We need to replace them with reasonable amounts for each function. I recommend using `max` from the above and filling them in for each. Ideally we could use some sort of gas estimator that depends on the complexity of the app's valid transition function.

geoknee:
See #1950

PR #1900, opened 2020-05-19
Refactor W3T closeChannels()

spersico:
It adds support for closing **paying channels** (channels used to pay to the peer, as opposed to **payment channels** that the peer uses to pay to another peer).
It also adds some logs to follow the closing channel flow.

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -137,7 +137,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     log.info('> Cancelling download. Closing payment channels, and then removing torrent');
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
-      await this.closeDownloadingChannels(torrent);
+      await this.closeChannels(torrent, true);
       torrent.destroy(() => this.emitTorrentUpdated(infoHash));
       track('Torrent Cancelled', {
         infoHash,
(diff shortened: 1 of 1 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -484,25 +484,26 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     );
     return lastPieceIsReservedToThisWire;
   }
-  /**
-   * Close any channels that I am downloading from (that my peer opened)
-   */
-  protected async closeDownloadingChannels(torrent: PaidStreamingTorrent) {
+
+  /**Close any channels that I am downloading from or seeding to  */
```
kerzhner, line 488:
```suggestion
  // Close any channels that I am downloading from or seeding to
```

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -137,7 +137,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     log.info('> Cancelling download. Closing payment channels, and then removing torrent');
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
-      await this.closeDownloadingChannels(torrent);
+      await this.closeChannels(torrent, true);
```
kerzhner, line 140: Just to make sure I understand correctly. When the leecher triggers cancel, the leecher's seeding channels will be cancelled. Is that the behavior that we want?

spersico: Yes. I believe so, because when the torrent is cancelled, it's gets removed from the client, which makes it impossible to seed or leech. 
This way, the peer makes an attempt to close the channel before cancelling the download.

PR #1903, opened 2020-05-19
Fix Optional/Cancel Download test 

snario:
- Browser was pointing to same private key as hub
- Puppeteer crashed strangely (error was `Object`) waiting for something that didn't exist
- Test passed despite the actual condition failing (channel don't close)


snario:
Tests fail now waiting for channels to close, which seems to not work. One tab gets stuck at `Closing...` and the other does not seem to even begin to close.

packages/e2e-tests/puppeteer/__tests__/optional/seed-download-cancel.test.ts
```diff
@@ -110,13 +110,12 @@ describe('Optional Integration Tests', () => {
...
-    console.log('Wait for Wallet to be hidden');
-    // Ensure the wallet is not visible
-    await forEachTab(waitForWalletToBeHidden);
-
     console.log('Wait for the "Restart Download" button to appear');
     await waitForFinishedOrCanceledDownload(web3tTabB);
 
+    console.log('Wait for the ChannelUpdated "closed" state');
+    await forEachTab(tab => waitForClosedState(tab));
+
```
spersico, line 118: The logic for this: `await waitForFinishedOrCanceledDownload(web3tTabB);` is that the button should appear when the channel is closed, and the torrent is deleted. The torrent is deleted once the closeChannel call is finished. So I believe that the `    await forEachTab(tab => waitForClosedState(tab));` it's not necessary.

snario: Will let you + @kerzhner decide and edit this branch as needed. When I ran it locally I observed the channel _not_ closing.

packages/devtools/src/constants.ts
```diff
@@ -7,6 +7,7 @@ export const ETHERLIME_ACCOUNTS: Account[] = [
   {privateKey: '0xf473040b1a83739a9c7cc1f5719fab0f5bf178f83314d98557c58aae1910e03a'},
   {privateKey: '0x823d590ed2cb5e8493bb0efc834771c1cde36f9fc49b9fe3620ebd0754ad6ea2'},
   {privateKey: '0xd6d710943471e4c37ceb787857e7a2b41ca57f9cb4307ee9a9b21436a8e709c3'},
+  // Hub uses this (index 6)
```
kerzhner, line 10: In theory, the `devtools` package could be used independent of the other packages. In which case, this comment might be a bit confusing. That being said, I am ok with having the comment

spersico:
So, I'm going to sleep. I've been tinkering with this branch, @kerzhner feel free to take over:
- So, I made sure that the channel is closing both by [resolving the `closeChannel` promise only on close](https://github.com/statechannels/monorepo/pull/1903/commits/5b735690fb86e60389e0c1fb6fd8120af7c69190) and by [not nullifying the channelId used for comparison, only when the channel is already closed](https://github.com/statechannels/monorepo/pull/1903/commits/b4671a7b4e8b60dad68eead1027ca72270ea2a2d).
- taking a look into the last (failed) test, I found a couple of interesting things:
 - By adding a console log into the channelUpdate listener (instead of a **pino** log which for some reason print `JSHandle@object` everywhere, I've been able to find [between the logs ](https://28909-209829093-gh.circle-artifacts.com/0/tmp/seed-download-cancel.browser.log) the following lines:
```
Browser 0 logged 0xb53f208ab5929a48d7e5f7285d8ae6f58a7cf8d21aa6162ac9398036fb5c8fea status closed
Browser 1 logged 0xb53f208ab5929a48d7e5f7285d8ae6f58a7cf8d21aa6162ac9398036fb5c8fea status closed
```

I've also made the channelList show the ACTUAL channel status, which in the screenshots shows as:




Which points to the seeder for some reason not updating the UI for some reason? I'm not sure I understand what's going on. The logs show me that both browsers got the update, why does the seeder doesn't update it's UI? I'm.... confused

PR #1907, opened 2020-05-19
[ rps ] Challenging

geoknee:
WIP

### DIFF
--- a/packages/nitro-protocol/contracts/ForceMove.sol
+++ b/packages/nitro-protocol/contracts/ForceMove.sol
@@ -48,7 +48,7 @@ contract ForceMove is IForceMove {
         Signature[] memory sigs,
         uint8[] memory whoSignedWhat,
         Signature memory challengerSig
-    ) public {
+    ) public override {
         bytes32 channelId = _getChannelId(fixedPart);
 
         if (_mode(channelId) == ChannelMode.Open) {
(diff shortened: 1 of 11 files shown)


packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -48,7 +48,7 @@ contract ForceMove is IForceMove {
         Signature[] memory sigs,
         uint8[] memory whoSignedWhat,
         Signature memory challengerSig
-    ) public {
```
geoknee: I'm interested to know why this is needed / how we currently get away without it during CI?

lalexgap: I'll take a look. Locally the contracts wouldn't compile without the `override`

lalexgap: It looks the removal of `override` occurred in 723e044. I couldn't find the CI run for that specific commit, but based on this [CI run](https://app.circleci.com/pipelines/github/statechannels/monorepo/6641/workflows/1f6066a1-eee3-4099-9808-3207c0c9980d/jobs/28660) it looks like it was failing on CI as well.

If you look on `master` we do use the `override` modifier https://github.com/statechannels/monorepo/blob/4b6fc9d64d96ee5eaf2dbbf03863dd62e832d21f/packages/nitro-protocol/contracts/ForceMove.sol#L50.

The underlying cause is probably the `solidity` vs code extension removing `override` when formatting, https://github.com/juanfranblanco/vscode-solidity/issues/174

geoknee: Ah thanks for clearing this up

lalexgap:
Closing in favour of #1945

PR #1910, opened 2020-05-19
[web3torrent] index peerList by channel id instead of peer id

kerzhner:
Lays the groundwork for https://github.com/statechannels/monorepo/issues/1735. Specifically, this should address https://github.com/statechannels/monorepo/issues/1735#issuecomment-630911254.

Also renaming `pseChannelId` to `seedingChannelId` and `peerChannelId` to `leechingChannelId`.

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -157,18 +157,18 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       torrentInfoHash,
       peerAccount
     });
-    log.info({from: Object.keys(this.peersList)[0]}, '<< blockedPeer %s', peerAccount);
+    log.info({from: this.peersList[torrentInfoHash], peerAccount}, `<< blockedPeer ${peerAccount}`);
   }
 
-  unblockPeer(torrentInfoHash: string, wire: PaidStreamingWire, peerAccount: string) {
-    log.info({from: Object.keys(this.peersList), peerAccount}, '<< unblockedPeer: start');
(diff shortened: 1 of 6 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -269,12 +291,15 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-      if (channelState.channelId === pseChannelId || channelState.channelId === peerChannelId) {
+      if (
+        channelState.channelId === seedingChannelId ||
+        channelState.channelId === leechingChannelId
+      ) {
         if (channelState.status === 'closed' || channelState.status === 'closing') {
-          wire.paidStreamingExtension.pseChannelId = null;
-          wire.paidStreamingExtension.peerChannelId = null;
+          wire.paidStreamingExtension.seedingChannelId = null;
+          wire.paidStreamingExtension.leechingChannelId = null;
```
spersico, line 302: I just noticed that I made a mistake, in this if statement. 
This might make problems: There is the posibility that a peer has both a seedingChannel and a leechingChannel, and only one of them got closed (for example, if it finished downloading the file). This if statement would remove both ids.

spersico: It's not something that you introduced, I think I introduced it at some point

spersico: I think it might be better to make an issue about this, instead of fixing it in this PR though

spersico: https://github.com/statechannels/monorepo/issues/1911

PR #1920, opened 2020-05-20
[TTT] Setup deployment for Production

Alonski:
This sets up TTT to be ready for deployment from the `deploy` branch.


packages/tic-tac-toe/config/dotenv.js
```diff
@@ -6,6 +6,19 @@ const path = require('path');
 const fs = require('fs');
 
 function configureEnvVariables(env) {
+  // State Channel Environment
+  // Intended usage is a single file in monorepo root defining configuration for multiple packages
+  const SC_ENV = process.env.SC_ENV;
```
geoknee, line 11: Any reason not to import this from @statechannels/devtools?

packages/tic-tac-toe/config/env.js
```diff
@@ -32,7 +32,7 @@ const REACT_APP = /^REACT_APP_/i;
 // These are Magmo specific values that should be injected into the application
 const MAGMO_VALUES = [
```
geoknee, line 33: This seems like an out-of-date name

Alonski: I copy pasted this from somewhere and I don't remember where :)
I don't even think I need this file anymore hmm. :)

packages/tic-tac-toe/config/dotenv.js
Alonski: `devtools` requires the env vars directly in this function.
In Ember I use an Ember Addon to inject the env vars during runtime.
If the devtools function was split up into two functions (one that returns a path and one that requires the env vars) I could utilize it.

PR #1922, opened 2020-05-20
Disable sandbox and dev-shm on Puppeteer

snario:
Sometimes tests would fail like this:

```node
  ...

  console.log puppeteer/__tests__/optional/seed-download-cancel.test.ts:117
    Wait for the "Restart Download" button to appear

(node:4824) UnhandledPromiseRejectionWarning: Error: Caught error after test environment was torn down

Protocol error (Runtime.callFunctionOn): Target closed.
(node:4824) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:4824) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
 FAIL  puppeteer/__tests__/optional/seed-download-cancel.test.ts (22.662s)
  Optional Integration Tests
    ✕ Optional - Torrent a file - Cancelling at State N°10 (17533ms)

  ● Optional Integration Tests › Optional - Torrent a file - Cancelling at State N°10

    Object

      76 |   });
      77 | 
    > 78 |   it('Optional - Torrent a file - Cancelling at State N°10', async () => {
         |   ^
      79 |     await web3tTabA.goto(WEB3TORRENT_URL + '/upload', {waitUntil: 'load'});
      80 |     await web3tTabA.bringToFront();
      81 | 

      at new Spec (../../node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.Object.<anonymous>.describe (puppeteer/__tests__/optional/seed-download-cancel.test.ts:78:3)
      at Object.<anonymous> (puppeteer/__tests__/optional/seed-download-cancel.test.ts:49:1)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        23.195s
Ran all test suites matching /optional/i.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

It was hard to understand what was going on but after some research I came across these threads:

- https://github.com/puppeteer/puppeteer/issues/3683
- https://github.com/puppeteer/puppeteer/issues/1175

Which made it seem like some sort of memory error was causing this bug. After adding `--disable-dev-shm-usage` and `--no-sandbox` (together) the test in #1903 seemed to pass more reliably.

[...]

PR #1923, opened 2020-05-20
Remove page.evaluate and page.exposeFunction from cancel test

snario:
So, I still saw the same errors that #1922 was supposed to fix even after applying the changes.

As an experiment, I removed all usages of the `doneWhen` helper and now rely solely on the UI.

The test still throws this `Object` error alongside `Protocol error (Runtime.callFunctionOn): Target closed.`

However, in an interesting twist, if I run it with `USE_DAPPETEER=true`, it never fails.

### DIFF
--- a/packages/e2e-tests/puppeteer/__tests__/optional/seed-download-cancel.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/optional/seed-download-cancel.test.ts
@@ -17,14 +17,12 @@ import {
   setUpBrowser,
   setupLogging,
   waitAndOpenChannel,
-  waitForNthState,
   waitForFinishedOrCanceledDownload,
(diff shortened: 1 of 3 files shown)


snario:
@geoknee do you have any insight into why `USE_DAPPETEER=true` causes the test to never seem to fail?

geoknee:
@snario not sure, but maybe take a look at the puppeteer launch args, they don't look identical in the case of using or not using dappeteer.

snario:
My best guess right now is something to do with how cancel is implemented inside Web3Torrent. I think that it is crashing the headless Chromium instance sometimes.

snario:
I do think this is a good PR though, removing reliance on events in the `console` and using only the UI seems like a better pattern for a puppeteer test.

PR #1924, opened 2020-05-20
Use pipe: true on puppeteer

snario:
Based on [this comment](https://github.com/puppeteer/puppeteer/issues/2735#issuecomment-470309033) that I found after [my hunch](https://github.com/statechannels/monorepo/pull/1923#issuecomment-631478009) that Web3Torrent's usage of WebSockets was interfering with the puppeteer functionality. Passed locally several times in a row, but then recently stopped working.


snario:
I think the test passing here was probably just good luck, it does not pass consistently for me locally. :(

snario:
5/6 seemed to pass above @geoknee

PR #1927, opened 2020-05-20
Sort channels by status firstly in ChannelsList

snario:


packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
```diff
@@ -79,11 +79,17 @@ function channelIdToTableRow(
...
   const channelsInfo = _.keys(channels)
     .filter(
       id => channels[id].payer === mySigningAddress || channels[id].beneficiary === mySigningAddress
     )
-    .sort((id1, id2) => Number(id1) - Number(id2));
+    .sort(
+      (id1, id2) =>
+        statuses.indexOf(channels[id1].status) - statuses.indexOf(channels[id2].status) ||
+        Number(id1) - Number(id2)
+    );
```
andrewgordstewart, line 92: I remember lots of headaches with `Array.prototype.sort`. Would you consider `_.sortBy`?

lalexgap: @andrewgordstewart do you remember what the issue with `Array.prototype.sort` is?

andrewgordstewart: I think it was just that I was doing it wrong:
```
❯ node
> [1, 2].sort(i => i)
[ 2, 1 ]
> [2, 1].sort(i => i)
[ 1, 2 ]
```

ie. `.sort` accepts a boolean comparison function, not a unary sorting-key function (like `_.sortBy`, or python's sort function)

PR #1928, opened 2020-05-20
Correctly pass wires through components

snario:


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -75,6 +75,10 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     this.outcomeAddress = null;
   }
 
+  get(torrentId: PaidStreamingTorrent | string | Buffer): PaidStreamingTorrent | void {
+    return super.get(torrentId) as PaidStreamingTorrent;
```
kerzhner, line 79: Is `get` overridden to enable this type assertion?

snario: Yes

PR #1929, opened 2020-05-20
[web3torrent] Channels display number of payments

geoknee:
This is a demo app. When it works, it's pretty awesome to consider what is going on under the hood. 

OTH if users just come away with an impression of  "webtorrent clone", they will have largely missed the point. 

I think one way of showing off more of how awesome state channels are is to display the number of "L2 transactions" in real time.

This changeset also introduces a blockie for our peer, which has the full address in a tooltip. Personally I think this looks nicer, and it might help identify peers. 

Screenshot 2020-05-20 at 19 21 29
Screenshot 2020-05-20 at 19 21 31

TODO (feedback welcomed)
---
- [x] potential changes to styling of blockie (we could make it match what's in MM by pulling the outcome address for our peer, and then trying to match the colours with MM)
- [x] display our blockie somewhere
- [x] extend blockies to channel ids
- [x] alignment of the table seems off (I don't think I caused this)

### DIFF
--- a/.env.circle-integration-w3t-with-hub
+++ b/.env.circle-integration-w3t-with-hub
@@ -10,12 +10,12 @@ USE_GANACHE_DEPLOYMENT_CACHE = 'true'
 
 # For e2e-test
 BROWSER_LOG_DESTINATION = 'browser.log'
-HEADLESS = 'false'
+HEADLESS = 'true'
 USE_DAPPETEER = 'false'
 TARGET_NETWORK = 'localhost'
 WEB3TORRENT_URL = 'http://localhost:3000'
 SCREENSHOT_DIR = 'screenshots'
-CLOSE_BROWSERS = 'false'
+CLOSE_BROWSERS = 'true'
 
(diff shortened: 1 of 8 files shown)


geoknee:
We can also change the colour of the badge to show payments sent/ received

geoknee:
Screenshot 2020-05-21 at 15 42 47

Rolled in a few other changes. Your budget is now displayed as a percentage, with a circular progress bar. The budget was lowered to 1GB (from 1 petabyte).

snario:
I think the spent / earned progress indicator seems a bit odd. Maybe because there is no border? Where is 100%?

geoknee:
Screenshot 2020-05-21 at 19 34 27

Channel column removed. Progress bars improved.

geoknee:
Considering that there are many PRs that are/will target these components, I think it's worth merging this now if we are reasonably happy: I don't want to spend ages polishing it if it will likely change soon anyways.

PR #1933, opened 2020-05-20
Update UpdateChannel API

andrewgordstewart:
- store.updateChannel accepts a callback, which executes within a tx scope. This allows us to push logic away from the store, and into the relevant workflows
- application workflow responds with "Not your turn" and "No channel found" error responses

(Based on https://github.com/statechannels/monorepo/pull/1931)

## TODO
Before removing draft status, I'd like to address some of these issues:
- [ ] Refactor away the term "Player"
   - web3torrent is not a game!
- [ ] Refactor running test to more easily test the error responses
   - Currently, there is a lot of setup that could be simplified by the ability to [reload a store](https://github.com/statechannels/monorepo/issues/1711)
- [ ] Handle the `PLAYER_UPDATE` event when not in the `running` state
- [ ] Handle the other [possible errors](https://github.com/statechannels/monorepo/blob/59ade2faf295c555cef9ea3d544a92731f5b5648/packages/client-api-schema/src/error-codes.ts#L12-L18)
- [ ] Define more [UpdateChannel errors](https://github.com/statechannels/monorepo/blob/59ade2faf295c555cef9ea3d544a92731f5b5648/packages/client-api-schema/src/error-codes.ts#L12-L18)
   - What should happen if the workflow is not in the `running` state? (more or less equivalent to: the channel cannot accept an update)

### DIFF
--- a/packages/client-api-schema/src/types.ts
+++ b/packages/client-api-schema/src/types.ts
@@ -70,6 +70,7 @@ export function isError(message: JsonRpcMessage): message is ErrorResponse {
 export * from './notifications';
 export * from './methods';
 export * from './data-types';
+export * from './error-codes';
 
 import {ErrorCodes} from './error-codes';
 
--- a/packages/xstate-wallet/src/integration-tests/running.test.ts
+++ b/packages/xstate-wallet/src/integration-tests/running.test.ts
@@ -30,31 +30,33 @@ test('accepts states when running', async () => {
   ]);
 
   hookUpMessaging(playerA, playerB);
-  const channelId = '0x440b56b6c5b0adca1ee99e3926d1b123fd867566cfab5150479f9d5e9317fa1e';
-
-  const playerBChannelUpdatedPromise = playerB.store
-    .channelUpdatedFeed(channelId)
-    .pipe(first())
-    .toPromise();
 
   const stateVars = {
     outcome,
-    turnNum: bigNumberify(4),
+    turnNum: bigNumberify(5),
     appData: '0x0',
     isFinal: false
   };
-  playerA.store.createChannel(
+  const {channelId, latest} = await playerA.store.createChannel(
     [playerA.participant, playerB.participant],
     bigNumberify(4),
     stateVars
   );
-  await playerBChannelUpdatedPromise;
+
+  await playerB.store.addState(latest);
+  await playerB.store.signAndAddState(channelId, latest);
+
+  await playerB.store
+    .channelUpdatedFeed(channelId)
+    .pipe(first(e => e.isSupported))
+    .toPromise();
 
   const context: any = {channelId, applicationDomain: 'localhost', fundingStrategy: 'Direct'};
   playerA.startAppWorkflow('running', context);
   playerB.startAppWorkflow('running', context);
   playerA.workflowMachine?.send('SPAWN_OBSERVERS');
   playerB.workflowMachine?.send('SPAWN_OBSERVERS');
+
   await playerA.messagingService.receiveRequest(
     generatePlayerUpdate(channelId, playerA.participant, playerB.participant),
     'localhost'
@@ -64,9 +66,9 @@ test('accepts states when running', async () => {
     expect(playerA.workflowState).toEqual('running');
     expect(playerB.workflowState).toEqual('running');
     const playerATurnNum = (await playerA.store.getEntry(channelId)).latest.turnNum.toNumber();
-    expect(playerATurnNum).toBe(5);
+    expect(playerATurnNum).toBe(6);
     const playerBTurnNum = (await playerB.store.getEntry(channelId)).latest.turnNum.toNumber();
-    expect(playerBTurnNum).toBe(5);
+    expect(playerBTurnNum).toBe(6);
   }, 3000);
 
   await playerB.messagingService.receiveRequest(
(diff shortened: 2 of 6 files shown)


snario:
All of these changes are very exciting :D

tomclose:
It looks like this PR is doing two things: (1) add proper error responses when it's not your turn etc., (2) move update logic from the store to the application workflow. When reading it, I began to wonder whether (2) is actually a good thing to do.

Before this change, the workflow had to understand that you passed in the `outcome`, `appData` and `isFinal` that you wanted and the store took care of things. We could write that update logic once in the store, test it well, and have it work reliably across all workflows.

In this case, it seems like that api would still work, and there's no reason that we absolutely need to put any more logic into the transaction. It seems like the main benefit is that it makes it relatively easy to send the messaging service responses - but there are other ways to handle this e.g. using a `try / catch` block around the `await updateChannel`.

After this change it seems like there's a lot more that can go wrong when writing a workflow. For example, you have to know that the callback you write there will be executed within a transaction, and you can therefore only make asynchronous requests to the store. You also take on the responsibility of updating the channel correctly, making it harder to give guarantees about data integrity within the store. Finally your workflow is also somewhat coupled to decisions about how the store represents data e.g. an internal change from `turnNum: BigNumber` to `turnNum: number` would require a change to your workflow.

Are there other examples where we need to be able to pass a callback into `updateChannel`, where the existing api wouldn't work?

andrewgordstewart:
> Are there other examples where we need to be able to pass a callback into `updateChannel`, where the existing api wouldn't work?

I think I agree. The only other case I can think of is that, when concluding, I would want to perform similar (but not identical) checks within a transaction.

This change allows `updateChannel` to be used for both of those updates, but I'm not convinced it's worth it.

andrewgordstewart:
Now that I look at all the calls to `signAndAdd`, it strikes me that they would be better off with the  version of `updateChannel` in this PR:
1. there are currently three calls to `signAndAdd`. The call to `updateChannel` would probably be a fourth.
   a. In [conclude-channel](https://github.com/statechannels/monorepo/blob/bc9bd91/packages/xstate-wallet/src/workflows/conclude-channel.ts#L14-L19)
   b. In [application](https://github.com/statechannels/monorepo/blob/bc9bd9136eb55c3febb0758bdd4ac84f14a02339/packages/xstate-wallet/src/workflows/application.ts#L124-L137)
   c. In [support-state](https://github.com/statechannels/monorepo/blob/bc9bd9136eb55c3febb0758bdd4ac84f14a02339/packages/xstate-wallet/src/workflows/support-state.ts#L42-L46)
2. each of these calls should:
   a. fetch the current entry
   b. check the current entry
   c. determine what state they want to sign
   d. sign and add that state
  And, these calls should happen within a transaction, because the state you sign and add in (d) is only the _correct_ state to add, if the current entry is the same as when you fetched it in (a).

So, it seems like we want to lean one way or another:
A. the store has four separate calls to validate and sign states in bespoke ways
B. the store offers a flexible way (by accepting an argument to an `updateChannel` function) to specify how to validate and sign states, and wraps that function within a transaction.

After evaluating this, I'm inclined to say that (B) is preferable over (A)

tomclose:
Looking at these examples, it seems like the logic is more tied to consistency in the store, and not signing things that you obviously shouldn't, rather than workflow-dependent logic. This pushes me towards thinking that they should be store methods.

What if you were to split these into two types?

1. Operations that add a new state to the store (`updateState` and `signFinalStateIfMyTurn` [b])
    * Could be handled by the pre-existing `updateState({ outcome, appData, isFinal })`
    * .. where for `signFinalStateIfMyTurn` is made from setting `isFinal = true`
    * In both cases this would fail if not your turn
2. Operations that countersign an existing state (`signFinalState` [a] and `signState` [c])
    * These could be handled by a new method `countersign(channelId, stateHash)`
    * This would be a no-op if you'd already signed

It seems that in the case of 2, this would lead to increased certainty over what you're signing, which could be good future proofing, if we ever end up with multiple states at each turn number.

andrewgordstewart:
> Looking at these examples, it seems like the logic is more tied to consistency in the store, and not signing things that you obviously shouldn't, rather than workflow-dependent logic. This pushes me towards thinking that they should be store methods.
> 
> What if you were to split these into two types?
> 
> 1. Operations that add a new state to the store (`updateState` and `signFinalStateIfMyTurn` [b])
>    
>    * Could be handled by the pre-existing `updateState({ outcome, appData, isFinal })`
>    * .. where for `signFinalStateIfMyTurn` is made from setting `isFinal = true`

This seems reasonable, except

>    * In both cases this would fail if not your turn

This starts to look like workflow-specific logic. Thinking forward, we want to support updates when it's not your turn, `isFinal` updates for some special payment channels, and more complicated updates in general. Those sound like

For the moment, we don't have those requirements, and having two methods on the store looks like a good idea
1. Sign the state in channel `0x123` with state hash `0xabc` throw if you don't have it
2. Update channel `0x123`, if it's my turn, to have properties `{outcome, isFinal, appData}`. Throw if either
   a. ...
   b. ...
   c. ...

andrewgordstewart:
(Another benefit of having these two store methods is that workflows don't need to be aware of the intricacies of DB transactions)

andrewgordstewart:
Closing this PR as it has served its purpose. I will, in two separate PRs

1. refactor the store methods according to something like https://github.com/statechannels/monorepo/pull/1933#issuecomment-632332837

2. Refactor the application workflow to adhere to the spec of the `UpdateChannel` method

PR #1934, opened 2020-05-20
Update to ethers v5 in Xstate Wallet

lalexgap:
Update ethers to v5 in xstate wallet


packages/xstate-wallet/src/utils/contract-utils.ts
```diff
@@ -20,9 +19,9 @@ export function tokenAddress(assetHolderAddress: string): string | undefined {
   throw 'TokenAddress not found';
 }
 
-export function getProvider(): Web3Provider | JsonRpcProvider {
+export function getProvider(): providers.Web3Provider | providers.JsonRpcProvider {
```
snario, line 22: Can we maintain directly importing these symbols?

packages/xstate-wallet/src/chain.ts
```diff
@@ -325,18 +320,18 @@ export class ChainWatcher implements Chain {
       to: NITRO_ADJUDICATOR_ADDRESS
     };
 
-    const response = await this.signer.sendTransaction(transactionRequest);
+    const response = await this.signer.sendTransaction(transactionRequest as any); // TODO: Seems to be an ethers typing issue around the nonce;
```
snario, line 323: What's the issue exactly?

lalexgap: https://github.com/ethers-io/ethers.js/issues/844

snario: This may be due to different implementations of `BigNumber` in the repo

packages/xstate-wallet/src/chain.ts
```diff
@@ -151,7 +152,7 @@ export class FakeChain implements Chain {
   }
 
   public depositSync(channelId: string, expectedHeld: string, amount: string) {
-    const current = (this.channelStatus[channelId] || {}).amount || bigNumberify(0);
+    const current = (this.channelStatus[channelId] || {}).amount || BigNumber.from(0);
```
snario, line 155: In general I like to use `Zero` instead of `BigNumber.from(0)`. We should probably stick to one convention.

package.json
```diff
@@ -60,7 +60,12 @@
...
   },
-  "workspaces": [
-    "packages/*"
-  ]
+  "workspaces": {
+    "packages": [
+      "packages/*"
+    ],
+    "nohoist": [
+      "**/ethers"
```
kerzhner, line 68: Is this because a few other packages are not yet upgraded to ethers v5?

packages/xstate-wallet/package.json
```diff
@@ -30,7 +30,7 @@
     "@xstate/react": "1.0.0-rc.3",
     "async-lock": "1.2.2",
     "dexie": "3.0.0",
-    "ethers": "4.0.45",
+    "ethers": "5.0.0-beta.187",
```
kerzhner, line 33: we should probably use the same version between the wallet and the simple-hub to avoid having to download 2 versions of the package. The hub currently uses 5.0.0-beta.184

packages/xstate-wallet/src/chain.ts
```diff
@@ -324,10 +320,13 @@ export class ChainWatcher implements Chain {
       ...Transactions.createConcludePushOutcomeAndTransferAllTransaction(
         finalizationProof.flatMap(toNitroSignedState)
       ),
-      to: NITRO_ADJUDICATOR_ADDRESS
+      to: NITRO_ADJUDICATOR_ADDRESS,
+      nonce: undefined
```
kerzhner, line 324: Curious, why is this needed?

packages/xstate-wallet/src/chain.ts
```diff
@@ -444,3 +447,21 @@ export class ChainWatcher implements Chain {
     );
   }
 }
+
+// Since nitro-protocol is still using v4 of ethers we need to convert any bignumbers the v5 version from ethers
+// TODO: Remove this when nitro protocol is using v5 ethers
```
kerzhner, line 452: We might want to create an issue

packages/xstate-wallet/src/utils/contract-utils.ts
```diff
@@ -24,22 +25,16 @@ export function getProvider(): Web3Provider | JsonRpcProvider {
...
-    return new providers.JsonRpcProvider(`http://localhost:${process.env.GANACHE_PORT}`);
+    return new JsonRpcProvider(`http://localhost:${process.env.GANACHE_PORT}`);
   }
 }
 
 export function getEthAssetHolderContract() {
   const provider = getProvider();
   return new Contract(
+    // eslint-disable-next-line no-process-env
     process.env.ETH_ASSET_HOLDER_ADDRESS || '0x0',
```
kerzhner, line 36: Is there a reason why a constant from the config cannot be used here?

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
```diff
@@ -219,7 +219,7 @@ const createLedger = (store: Store) => async (context: Initial): Promise<LedgerI
 
   const stateVars = {
     outcome: initialOutcome,
-    turnNum: bigNumberify(0),
+    turnNum: BigNumber.from(0),
```
kerzhner, line 222: Can we use Zero from '@ethersproject/constants' for code like this? (basically +1 to @snario's comment)

packages/xstate-wallet/src/utils/contract-utils.ts
snario: To be honest I’m surprised this is even passing linting. I thought I disabled using process.env outside of the config file.

snario: Oh I didn’t even see the disable comment. Yeah let’s not do that :p

package.json
lalexgap: Yup. That meant both versions of `ethers` were available which caused me to run into this [issue](https://github.com/ethers-io/ethers.js/issues/844).

It also helped me catch references to v4 of `ethers` that we hadn't removed from the `xstate-wallet` and `simple-hub`.

packages/xstate-wallet/src/chain.ts
lalexgap: Oops, I think I that was from debugging that I forgot to remove.

packages/xstate-wallet/src/utils/contract-utils.ts
lalexgap: I'm not sure how it survived linting prior to this PR 😕

packages/xstate-wallet/src/workflows/approve-budget-and-fund.ts
lalexgap: [21 files later](https://github.com/statechannels/monorepo/pull/1934/commits/0a24f385a509b09cb070beb23ea1e58bd2fb3f11) 😛

PR #1935, opened 2020-05-21
Handle challenge detection

### LINKED ISSUE #1915: Wire up Challenge Detection
Currently when a challenge is raised we don't fire any event in the application workflow. The `isChallenging` function [isn't implemented ](https://github.com/statechannels/monorepo/blob/2faf00f495e5d7952b49010f98e1280feb65b6a1/packages/xstate-wallet/src/store/channel-store-entry.ts#L100)

It might make sense to remove the `ChallengeDetected` feed and just update the `ChainUpdated` feed to handle detecting challenges as well.

lalexgap:
Fixes #1915

This implements / refactors some plumbing around challenges.

The biggest change is that everything is now accessed via the `ChainUpdatedFeed`. 
Also the feed will now query for existing challenges/cleared challenges on start up.

There is now [work](https://github.com/statechannels/monorepo/issues/1960) to do to properly implement challenging/responding in the application workflow. 

TODO
- [x] Handle challenge cleared
- [x] Clean up interface and types
- [x] Handle challenge expiry

### DIFF
--- a/packages/xstate-wallet/src/chain.ts
+++ b/packages/xstate-wallet/src/chain.ts
@@ -9,7 +9,7 @@ import {
 import {Contract, Wallet, BigNumber, BigNumberish} from 'ethers';
 
 import {Observable, fromEvent, from, merge, combineLatest} from 'rxjs';
-import {filter, map, flatMap, defaultIfEmpty, mergeAll} from 'rxjs/operators';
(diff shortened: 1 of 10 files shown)


lalexgap:
Closing as this is not going to be in scope for the initial release.

Issue #34, opened 2019-09-23
[nitro-protocol] Rename `channelStorageHashes`, considering new storage format.


andrewgordstewart:
I am wondering if we still want to use 'hash' to refer to the data that is stored on chain? It is no longer the output of `keccak256`.

How about something like the following: the mapping is named `storageData`, and `_getHash` becomes `_encodeStorageDatum` etc?

_Originally posted by @geoknee in https://github.com/statechannels/nitro-protocol/pull/94_

snario:
I think we need distinctions between 

- channel storage hash: the "encoding" / concatenation of channel storage
- channel storage: the triple
- channel data: the struct with the hash of the outcome and the appData
- channel state: a "full" struct with access to all data, no hashes

snario:
Started a  branch for this

PR #1936, opened 2020-05-21
Introduce ChannelData naming

snario:
Begins addressing #34


andrewgordstewart:
The codebase still has references to the old naming, but as long as this change is just renaming, it seems good to go once those are fixed.

PR #1943, opened 2020-05-21
Remove Funds column on ChannelList

snario:


snario:
It's actually currently set to a value which is incorrect on `master` but correct in this branch

https://github.com/statechannels/monorepo/blob/35e9538ff333b5f1bc31916a96f9d1b252b8b245/packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.scss#L51-L52

snario:
Superseded by #1951

PR #1944, opened 2020-05-21
Rough refactor to display UI using child services

andrewgordstewart:
Creating a draft PR to enable discussion/references to code

### DIFF
--- a/packages/xstate-wallet/src/ui/approve-budget-and-fund-workflow.tsx
+++ b/packages/xstate-wallet/src/ui/approve-budget-and-fund-workflow.tsx
@@ -4,16 +4,16 @@ import {ApproveBudgetAndFundService} from '../workflows/approve-budget-and-fund'
 import {useService} from '@xstate/react';
 
 import {formatEther} from 'ethers/utils';
-import {Button, Heading, Flex, Text, Box, Link} from 'rimble-ui';
+import {Button, Heading, Flex, Text, Box} from 'rimble-ui';
 import {getAmountsFromBudget} from './selectors';
-import {ETH_ASSET_HOLDER_ADDRESS} from '../config';
+import {Deposit} from './depositing';
 
 interface Props {
   service: ApproveBudgetAndFundService;
 }
 
 export const ApproveBudgetAndFund = (props: Props) => {
-  const [current, send] = useService(props.service);
+  const [current, send]: any[] = useService(props.service);
   const {budget} = current.context;
   const {playerAmount, hubAmount} = getAmountsFromBudget(budget);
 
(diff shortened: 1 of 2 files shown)


packages/xstate-wallet/src/ui/approve-budget-and-fund-workflow.tsx
```diff
@@ -59,76 +59,13 @@ export const ApproveBudgetAndFund = (props: Props) => {
...
-  const depositFullyFunded = (
-    <Flex alignItems="center" flexDirection="column">
-      <Heading>Deposit funds</Heading>
-
-      <Text textAlign="center">Waiting for hub to deposit</Text>
+      <Heading>Depositing funds</Heading>
+      We are currently depositing funds into the ledger channel. We can add some
+      approve-budget-and-fund specific stuff here. Then, we can add an element containing the status
+      of the depositing service:
+      <Deposit service={depositingService}></Deposit>
```
andrewgordstewart, line 68: [This document](https://www.notion.so/State-Channels-4cc770686d05438bb23b3c37f4069d7e?p=2accdba2fade47fa9bf87f24fda110a8) originated due to the difficulty of getting the information from invoked services in order to display some UI elements.

I am not convinced that the `current approve-budget-and-fund` is a good engineering practice. (I'm also not convinced that it is a bad engineering practice, I am just undecided.)

However, it seems pretty feasible to have a `depositing` service that can be used by both the `approve-buget-and-fund` service as well as the `application` service (for directly funded apps)

Here is one way you could do that:  a `depositing` service can have a well-defined boundary around its responsibility:
- Submit a deposit, when it's safe, given its current outcome.
- Abandon if its outcome changes
- Abandon if I'm told to abandon
- Retry if the transaction fails
- Finish when the channel is funded to level X

That depositing service can define its own UI.

## Duplication over abstraction

Over time, I have warmed up @tomclose's philosophy of duplicating code:
- copy-paste once
- if you copy-paste again, then abstract the code

However, consider [this article's](https://medium.com/@kooliahmd/code-duplication-anti-pattern-diagnosis-and-treatment-44f8c1555382) main thesis (emphasis mine):

> Theoretically, a code duplication is defined by:
> Two blocs of code doing the same thing _and having the same set of reasons to change._

Being domain experts, I think we have a reasonable chance at evaluating whether the depositing states in `approve-budget-and-fund` and the depositing states in `application` have the same reasons to change: We might want to update the depositing workflow/UI if:
A. We want to make it bulletproof, considering many different types of chain events
B. We need to react to changes to the contract interface
C. We implement the [robust chain-watcher](https://www.notion.so/State-Channels-4cc770686d05438bb23b3c37f4069d7e?p=1cebbc1ff4024f0facfd4e0cd755e763), that emits "probable" and "almost certain" views of the chain
D. We want to change the styling of the UI element when asking the user if they want to resubmit the transaction with a bigger gas fee in case the transaction times out.
...etc

We, as domain experts, also have a reasonable chance at foreseeing other applications of the same depositing service:
E. we add a `ledger-top-up` service (or add that logic to an existing service) which needs to deposit at a certain funding level X, and finish at a certain funding level Y

lalexgap: Overall I'm a fan of this approach (although I think I'm strongly biased to re-usable code vs duplicating 😛 ). Encapsulating depositing into its own service makes it easier to reason about and has well-defined boundaries. 


> That depositing service can define its own UI.

I think this works for this specific use case but I'm wondering if we are ever going to have cases where we want the parent workflow controlling the UI for a child service? I guess in that case the parent workflow will just have to have its own component that knows how to the handle the child service. 

It probably makes sense to follow the general rule of one react component per a service. Parent workflows then can stitch together components as needed.

snario: > However, it seems pretty feasible to have **a depositing service that can be used by both** the approve-buget-and-fund service as well as the application service (for directly funded apps)

I agree with this conclusion based on the general line of thinking in the bottom part of your post.

tomclose: From slack conversation with Andrew:

Tom Close  9:50 PM
I actually think that’s quite good

Andrew  9:51 PM
Sweet

Tom Close  9:51 PM
that gets rid of one of the considerations about displaying sub-machines
I don’t think it definitively answers the question about whether depositing should be a service or not

Andrew: 👍 

Andrew  9:51 PM
I also wonder if it's that much work to modify the visualizer to display invoked machines

Tom Close  9:52 PM
for example it might be that ‘transaction submission’ is actually the pattern to extract first or something

Andrew  9:53 PM
Yes, I agree

tomclose:
I think this was intended as a proof-of-concept, rather than something that's actively being worked on? Feel free to re-open if not.

PR #1946, opened 2020-05-21
Fix tracker check

### LINKED ISSUE #1597: Investigate why we see a "this torrent doesn't have any seeders" UI sometimes

spersico:
fixes #1597 
image

It is my current understanding, that the main issue was that **Heroku was dropping the webSocket connection between the seeder and the tracker.**

The tracker by default has an interval of 120s (even though the screenshot says 60s).
## So, the flow was essentially 

- The Seeder connects to tracker after seeding
- The Seeder loses connection to the tracker, the seeder immediately sets a timer of like... 40 to 50 seconds to reconnect to the tracker.
- The leecher scrapes the tracker to check for seeders, the tracker depending on the moment might work, or it might not (the code here is complex, it has like... waiting mechanisms and weird stuff).
- finally, the seeder reconnects to the tracker, and if the leecher scrapes again, finds the seeder.

## So, what did I do. 
Well, it took me a bunch of time to figure out, that's for sure:

- First, I updated the tracker dependencies (didn't work)
- Then, I fixed some env vars, (didn't change a thing)
- I fixed the tracker logs, just for fun, I want to be able to see what is the staging tracker doing, now I can. You can see them [**here**](https://dashboard.heroku.com/apps/web3torrent-tracker-staging/logs)
- I looked into the client tracker (the one in the function that does the torrent check), I tried a BUNCH of things. Never got to anything (though I'll make a PR tomorrow with a refactor that I thought about).
- I didn't understand what changed between my local tracker and the deployed one. On the local one, the tracker worked just fine, on the deployed one, the tracker gave me trouble sporadically.
- I reduced the probable cause being: using SSL, being Heroku, or being something weird about networks (that's always an option, for me). [Heroku's article about websockets](https://devcenter.heroku.com/articles/websockets#timeouts)
- When I figured out how to change the interval, that was it, once I deployed it, it seemed to work just fine.
- The hardest thing, above all, was figuring out stuff by reading the code and a metric ton of logs because the tracker docs.... yeah...

**I've already deployed the tracker to test this (the code in the screenshot is hitting the web3torrent-tracker.statechannels.org tracker!).**

Drawbacks of my solution and this code:
- The tracker now actually makes logs. It makes a log about each event, showing the infohash of the torrent, and the peerId (the 2 things it knows about each peer). So I expect more RAM use from the tracker. Maybe we should redirect the logs just the same than for the persistent seeder?.
- The interval for each update (the _"I'm still alive"_ message sent by the seeder) is now 30 seconds. I assume that fact will also impact the tracker. In the client side (meaning, in web3torrent) I don't think this will change too much. Before it made an update every 2 minutes, now it's 30 seconds. It's still a lot of time, and we only have one tracker.


packages/web3torrent/tracker/index.js
```diff
@@ -11,6 +11,7 @@ const server = new Server({
   http: true,
   ws: true,
   stats: true,
+  interval: 150000,
```
spersico, line 14: this does all the magic. 
Ignore the actual number. that 150000 is actually 30 seconds.
I don't understand the logic about how is the actual interval calculated.

PR #1950, opened 2020-05-22
[ nitro protocol ] Remove all GAS_LIMIT instances

geoknee:
Making this change didn't break anything. And it seems to allow ethers to automatically provide an accurate estimate for the gas limit. 

Screenshot 2020-05-22 at 11 45 26


geoknee:
I think this change should increase the probability of our `deploy` branch test (which runs puppeteer against the deployed app) passing. Accurate gas limits should get txs mined more quickly.

PR #1952, opened 2020-05-22
Set default value for onNext and onComplete for simple-hub

snario:
Fixes harmless (I think?) error on console:

```
20-05-22 12:56:36.665 +0000] INFO  (simple-hub/86966 on mhz.local): makeDeposit: attempting to make 0 deposits
[2020-05-22 12:56:36.665 +0000] INFO  (simple-hub/86966 on mhz.local): makeDepost: making 0 deposits
[2020-05-22 12:56:36.665 +0000] ERROR (simple-hub/86966 on mhz.local): onNext is not a function
    TypeError: onNext is not a function
        at SafeSubscriber.observable.pipe.subscribe [as _next] (/Users/liam/Documents/Projects/statechannels/monorepo/packages/simple-hub/src/message/on-message.ts:38:11)
        at process._tickCallback (internal/process/next_tick.js:68:7)
```

### DIFF
--- a/packages/simple-hub/src/message/on-message.ts
+++ b/packages/simple-hub/src/message/on-message.ts
@@ -16,7 +16,7 @@ export function onIncomingMessage(
     message: Either<Error, Message>;
   }>,
   onNext: () => void = noop,
-  onComplete: () => void = noop
+  onComplete?: () => void
 ) {
   return observable
     .pipe(
--- a/packages/simple-hub/src/message/on-message.ts
+++ b/packages/simple-hub/src/message/on-message.ts
@@ -14,8 +14,8 @@ export function onIncomingMessage(
     snapshotKey: string;
     message: Either<Error, Message>;
   }>,
-  onNext?: () => void,
-  onComplete?: () => void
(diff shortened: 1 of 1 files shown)


packages/simple-hub/src/message/on-message.ts
```diff
@@ -14,8 +15,8 @@ export function onIncomingMessage(
     snapshotKey: string;
     message: Either<Error, Message>;
   }>,
-  onNext?: () => void,
-  onComplete?: () => void
+  onNext: () => void = noop,
```
kerzhner, line 18: Good catch!

packages/simple-hub/src/message/on-message.ts
```diff
@@ -14,8 +15,8 @@ export function onIncomingMessage(
...
+  onComplete: () => void = noop
```
kerzhner, line 19: Outside of unit testing, the subscriber should never see completion of the observable. If the subscriber does see a completion of the stream, something has gone wrong.
```suggestion
  onComplete?: () => void
```

PR #1963, opened 2020-05-25
Bizarre experimental changes to cancel test to get it to pass

snario:
Got 10 tests in a row passing locally... trying on CI

### DIFF
--- a/packages/e2e-tests/puppeteer/__tests__/optional/seed-download-cancel.test.ts
+++ b/packages/e2e-tests/puppeteer/__tests__/optional/seed-download-cancel.test.ts
@@ -101,8 +101,18 @@ describe('Optional Integration Tests', () => {
 
     await web3tTabB.waitForSelector('.positive.downloading');
 
-    console.log('B cancels download');
-    await cancelDownload(web3tTabB);
+    // NOTE: For some reason wrapping this in a try{}catch{} makes it work.
+    try {
+      // NOTE: Added some delay here because, for some reason that I cannot figure out,
+      // the entire Puppeteer connection between this test and the Chromium window
+      // gets destroyed with an error message "Object" without the delays.
+      console.log('B cancels download');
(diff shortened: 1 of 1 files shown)


snario:
Noticed this:
https://30220-209829093-gh.circle-artifacts.com/0/tmp/seed-download-cancel.pino.log
```
{"time":1590417603679,"level":30,"jsonRpcResponse":{"id":38,"jsonrpc":"2.0","error":{"code":300,"message":"Not your turn"}},"msg":"OUTGOING JSONRPC REQUEST:","name":"xstate-wallet","browserId":1}
```

snario:
@spersico is it possible there is a race condition between when `getLatestPaymentReceipt` resolves and `closeChannel` is called? For example, `getLatestPaymentReceipt` returns and then the channel is updated by a regular `makePayment` and then `closeChannel` is called despite it not being that user's turn yet?

andrewgordstewart:
> Noticed this:
> https://30220-209829093-gh.circle-artifacts.com/0/tmp/seed-download-cancel.pino.log
> 
> ```json
> {
>    "time":1590417603679,
>    "level":30,
>    "jsonRpcResponse":{
>       "id":38,
>       "jsonrpc":"2.0",
>       "error":{
>          "code":300,
>          "message":"Not your turn"
>       }
>    },
>    "msg":"OUTGOING JSONRPC REQUEST:",
>    "name":"xstate-wallet",
>    "browserId":1
> }
> ```

Should the `ChannelClient` reject when it gets this error response?

snario:
I think it should

snario:
> could it be that tests started passing with this change because the catch statement swallows the error instead of bubbling it up?

I don't think so, because the test failed again.

snario:
It seems like this surface [some kind of error](https://github.com/statechannels/monorepo/pull/1963#issuecomment-633620371) so I'm going to close this PR in favour of us fixing that error.

PR #1964, opened 2020-05-25
Store refactor

andrewgordstewart:
Ticks R1, R3 & R6 of https://github.com/statechannels/monorepo/issues/1909

- Refactors the store according to the discussion on https://github.com/statechannels/monorepo/pull/1933
   - Channel updates now all (?) occur within a transaction
- Adds a generic wallet error with code 500
- Refactors the running test not to use `waitForExpect`, but instead block on responses and notifications
   - running test also tests for one error response, when it is not your turn
- Updates the `application` workflow to send error responses during a `PlayerUpdate`

# TODO:
Will remove draft status when
- [x] FIXME messages are resolved

### DIFF
--- a/packages/client-api-schema/src/generated-schema.json
+++ b/packages/client-api-schema/src/generated-schema.json
@@ -315,6 +315,9 @@
         },
         {
           "$ref": "#/definitions/UpdateChannelError"
+        },
+        {
+          "$ref": "#/definitions/JsonRpcError%3C500%2C%22Wallet%20error%22%3E"
         }
       ]
     },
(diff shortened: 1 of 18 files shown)


packages/client-api-schema/src/generated-schema.json
```diff
@@ -649,7 +652,7 @@
       ],
       "type": "object"
     },
-    "JsonRpcError<402,\"Invalid app data\",structure-1165839259-943-963-1165839259-875-965-1165839259-853-966-1165839259-0-1291>": {
+    "JsonRpcError<402,\"Invalid app data\",structure-1165839259-920-940-1165839259-852-942-1165839259-830-943-1165839259-0-1233>": {
```
snario, line 655: What is this structure stuff?

andrewgordstewart: My guess is it's references to line numbers or something -- the difference `y-x` in `1165839259-x-y` seems to be constant in most of this line change: it's 20, 20, 23, and increased by 4.

The numbers don't seem related to the lines in the file though ...

Issue #1961, opened 2020-05-25
Web3Torrent does not close all channels when download is finished


snario:
If a user has 3 channels for a single file, once the file is downloaded, only the most recently updated gets closed.

geoknee:
As far as I know this is no longer a problem.

Issue #1918, opened 2020-05-20
"Not your Turn" when canceling downloads


spersico:
**Describe the bug :bug:**
Testing manually (in master) using **virtual channels** while downloading a file the cancel button might not work on the first time it's clicked, showing instead an error in the console, telling me that it's not my turn.

Are the `closeChannel` channel updates ruled by the same rules than the other updates (that it must wait for its turn)? If it is, then the `PaymentChannelClient.closeChannel()` should wait for its turn, and only **then** close the channel, not throw an error, making the user click have no feedback at all.

This might be related to the error found in #1903, but it's an error by itself.

andrewgordstewart:
https://github.com/statechannels/monorepo/pull/1871

spersico:
> #1871

I see that George's PR is responsible for this behaviour. Is it the correct behaviour? What should be the approach to have a reliable cancel button, that is able to close channels?

spersico:
> > #1871
> 
> I see that George's PR is responsible for this behaviour. Is it the correct behaviour? What should be the approach to have a reliable cancel button, that is able to close channels?

I think that it should wait 'till the user's turn, and then call closeChannel right?

geoknee:
@spersico yes, we decided that it is the responsibility of the app to wait until it is our turn.  The wallet API specifies an error response if you try to close when not your turn. Apologies for not signalling this change more clearly !

Issue #1811, opened 2020-05-13
Error when closing channels, when cancelling downloads


spersico:
**Describe the bug :bug:**
Cancelling a download in progress fails the first time (with every torrent). Seems like cancelling triggers an invalid change of state in xstate-wallet, which makes the client throw, which blocks the torrent from being truly cancelled.https://github.com/statechannels/monorepo/blob/ad74d8d89de84288617ecfaba9f8c4b715b10b9a/packages/web3torrent/src/library/web3torrent-lib.ts#L134-L135
**Currently blocking #1781. Successfully reproduced (at least with virtual funding) in the current master, and in production.**

This results in that with both Virtual funding (with the hub) and Direct Funding, the cancel button must be pressed twice:
- With **Direct Funding**, an error log appears:
```
Browser 1 logged Missing onError handler for invocation 'signFinalState', error was 'Error: Supported state not final'. Stacktrace was 'Error: Supported state not final
    at http://localhost:3055/bundle.js:264367:15
    at Generator.next (<anonymous>)
    at fulfilled (http://localhost:3055/bundle.js:264352:58)'
```
- With **Virtual Funding**, no error got logged, though, between the logs, a couple of warnings can be seen :
```
utils.js:424 Warning: No implementation found for action type 'deleteGuarantorChannel'
utils.js:424 Warning: No implementation found for action type 'deleteTargetChannel'
utils.js:424 Warning: No implementation found for action type 'deleteJointChannel'
```

**Update**
While testing this on **master**, I noticed that the warnings that I mentioned above with virtual-funding, **also appear at the end of the leecher's download** (when closing the channel after finishing the download), so it's not related to the unexpected cancellation of the torrent.

------------------

In #1781 I added a `console.log('CANCEL BUTTON WAS CLICKED') `so that the leecher's log shows when the cancel button is pressed: The last lines of the [logs in circleCI](https://app.circleci.com/pipelines/github/statechannels/monorepo/6342/workflows/fa6b8e9c-32b6-4b41-8d12-6fbd8b2e376d/jobs/26440/steps), show the error that I was talking about.

**To Reproduce**
Steps to reproduce the behaviour:
1. Run all the things with `DEV_SC=virtual-funding` or the direct funding equivalent.
2. Seed a file in a browser and start downloading it in another
3. After a couple of turns passed, cancel the download
4. See the error (the torrent download bar is not supposed to still be shown, the seeder shouldn't see the leecher as connected (notice the number of peers connected didn't change))

[...]

PR #1969, opened 2020-05-25
W3t close channel

spersico:
This PR:
- Prevents the web3torrentClient to trigger payments while a torrent is being cancelled (by pausing the torrent):
https://github.com/statechannels/monorepo/compare/w3t-close-channel?expand=1#diff-a84153ed9f3ab2bd13c41bca43b584c7R134
https://github.com/statechannels/monorepo/compare/w3t-close-channel?expand=1#diff-a84153ed9f3ab2bd13c41bca43b584c7L401
- moves the logic of waiting for the user's turn before closing to the PaymentChannelClient:
https://github.com/statechannels/monorepo/compare/w3t-close-channel?expand=1#diff-9b0f77e1a02c15a73ad86b3e5d0b6620R212
- Makes channelClose calls independent of each other:
https://github.com/statechannels/monorepo/compare/w3t-close-channel?expand=1#diff-a84153ed9f3ab2bd13c41bca43b584c7R522-R525

related to #1961 and #1918 and #1811

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -204,7 +204,7 @@ export class PaymentChannelClient {
     this.insertIntoChannelCache(convertToChannelState(channelResult));
   }
 
-  async closeChannel(channelId: string, waitForMyTurn = false): Promise<ChannelState> {
+  async closeChannel(channelId: string): Promise<ChannelState> {
     const isClosed = (channelState: ChannelState) =>
       channelState.channelId === channelId && channelState.status === 'closed';
 
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -210,7 +210,6 @@ export class PaymentChannelClient {
 
     if (!['closing', 'closed'].includes(this.channelCache[channelId].status)) {
       await this.getLatestPaymentReceipt(channelId);
-      log.info('My turn, about to close channel');
       this.channelClient.closeChannel(channelId);
     }
     return this.channelClient.channelState
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -210,8 +210,8 @@ export class PaymentChannelClient {
 
     if (!['closing', 'closed'].includes(this.channelCache[channelId].status)) {
       await this.getLatestPaymentReceipt(channelId);
+      log.info('My turn, about to close channel');
       this.channelClient.closeChannel(channelId);
-      log.info('closeChannel() - channelClient.closeChannel() called!');
     }
     return this.channelClient.channelState
       .pipe(map(convertToChannelState), filter(isClosed), first())
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -204,12 +204,18 @@ export class PaymentChannelClient {
     this.insertIntoChannelCache(convertToChannelState(channelResult));
   }
 
-  async closeChannel(channelId: string): Promise<ChannelState> {
+  async closeChannel(channelId: string, waitForMyTurn = false): Promise<ChannelState> {
     const isClosed = (channelState: ChannelState) =>
       channelState.channelId === channelId && channelState.status === 'closed';
 
     if (!['closing', 'closed'].includes(this.channelCache[channelId].status)) {
+      if (waitForMyTurn) {
+        log.info('closeChannel() - Waiting in line');
+        await this.getLatestPaymentReceipt(channelId);
+        log.info('closeChannel() - Ready to Close!');
+      }
       this.channelClient.closeChannel(channelId);
+      log.info('closeChannel() - channelClient.closeChannel() called!');
     }
     return this.channelClient.channelState
       .pipe(map(convertToChannelState), filter(isClosed), first())
@@ -254,7 +260,7 @@ export class PaymentChannelClient {
     return convertToChannelState(channelResult);
   }
 
-  getLatestPaymentReceipt(channelId) {
+  getLatestPaymentReceipt(channelId: string) {
     const readyToPay = (state: ChannelState) =>
       state.channelId === channelId &&
       state.status === 'running' &&
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -209,11 +209,7 @@ export class PaymentChannelClient {
       channelState.channelId === channelId && channelState.status === 'closed';
 
     if (!['closing', 'closed'].includes(this.channelCache[channelId].status)) {
-      if (waitForMyTurn) {
-        log.info('closeChannel() - Waiting in line');
-        await this.getLatestPaymentReceipt(channelId);
-        log.info('closeChannel() - Ready to Close!');
-      }
+      await this.getLatestPaymentReceipt(channelId);
       this.channelClient.closeChannel(channelId);
       log.info('closeChannel() - channelClient.closeChannel() called!');
     }
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -120,6 +120,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
       torrent.pause(); // the paymentChannelClosing is done at the moment of payment
+      this.closeChannels(torrent, true);
       this.emitTorrentUpdated(infoHash);
     } else {
       throw new Error('No torrent found');
@@ -130,6 +131,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     log.info('> Cancelling download. Closing payment channels, and then removing torrent');
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
+      torrent.pause();
       await this.closeChannels(torrent, true);
       torrent.destroy(() => this.emitTorrentUpdated(infoHash));
     } else {
@@ -396,9 +398,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       switch (command) {
         case PaidStreamingExtensionNotices.STOP: // synonymous with a prompt for a payment
           if (torrent.paused) {
-            // We currently treat pausing torrent as canceling downloads
-            log.info({data}, 'Closing torrent');
-            await this.closeChannels(torrent);
+            log.info({data}, 'Torrent Paused');
           } else if (!torrent.done || !torrent.destroyed) {
             log.info({data}, 'Making payment');
             await this.makePayment(torrent, wire);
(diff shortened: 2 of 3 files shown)


snario:
I found this: https://30317-209829093-gh.circle-artifacts.com/0/tmp/seed-download-cancel.pino.log

```json
{
   "time":1590450980252,
   "level":30,
   "jsonRpcResponse":{
      "id":12,
      "jsonrpc":"2.0",
      "error":{
         "code":300,
         "message":"Not your turn"
      }
   },
   "msg":"OUTGOING JSONRPC REQUEST:",
   "name":"xstate-wallet",
   "browserId":1
}
```

spersico:
> I found [this log](https://30317-209829093-gh.circle-artifacts.com/0/tmp/seed-download-cancel.pino.log) from the [failed `integration-test-web3torrent-optional` job](https://app.circleci.com/pipelines/github/statechannels/monorepo/6862/workflows/428be103-e527-44f4-985b-7ce8144bdeb8/jobs/30317/artifacts)
> 
> ```json
> {
>    "time":1590450980252,
>    "level":30,
>    "jsonRpcResponse":{
>       "id":12,
>       "jsonrpc":"2.0",
>       "error":{
>          "code":300,
>          "message":"Not your turn"
>       }
>    },
>    "msg":"OUTGOING JSONRPC REQUEST:",
>    "name":"xstate-wallet",
>    "browserId":1
> }
> ```

yeah, I paired with @andrewgordstewart last night and made with him this PR, that puts the ball in xstate-wallet in regards with this error, so that he can start hacking into it on that side.
What I mean with this is: this PR is not supposed to fix the error, but it solves the issues related to it in web3torrent's side

snario:
@spersico Does this PR _fix_ #1961?

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -209,6 +209,8 @@ export class PaymentChannelClient {
       channelState.channelId === channelId && channelState.status === 'closed';
 
     if (!['closing', 'closed'].includes(this.channelCache[channelId].status)) {
+      await this.getLatestPaymentReceipt(channelId);
+      log.info('My turn, about to close channel');
```
andrewgordstewart, line 213: This doesn't look like it's intended to get into master.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -130,6 +131,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     log.info('> Cancelling download. Closing payment channels, and then removing torrent');
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
+      torrent.pause();
```
andrewgordstewart, line 134: Maybe add a bit of explanation, since the intended effect is a bit subtle?
```suggestion
      // I am only allowed to close the channel on my turn.
      // If I don't pause the torrent, then I will continue to make payments, meaning the call to 
      // this.closeChannels would race the `makePayment` function, and one of them would fail with a
      // "Not my turn" error.
      torrent.pause();
```

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -518,19 +518,19 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-          log.info(`About to close payment channel ${leechingChannelId}`);
-          await this.paymentChannelClient.getLatestPaymentReceipt(leechingChannelId);
-          await this.paymentChannelClient.closeChannel(leechingChannelId);
-          wire.paidStreamingExtension.leechingChannelId = null;
-          log.info(`Channel ${leechingChannelId} closed`);
+          log.info(`SP-> About to close payment channel ${leechingChannelId}`);
+          return this.paymentChannelClient.closeChannel(leechingChannelId, true).then(() => {
+            wire.paidStreamingExtension.leechingChannelId = null;
+            log.info(`SP-> Channel closed: ${leechingChannelId}`);
+          });
```
andrewgordstewart: Why the change from `await` to `.then`?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -519,18 +523,18 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
         const {seedingChannelId, leechingChannelId} = wire.paidStreamingExtension;
         if (leechingChannelId) {
           log.info(`About to close payment channel ${leechingChannelId}`);
-          await this.paymentChannelClient.getLatestPaymentReceipt(leechingChannelId);
-          await this.paymentChannelClient.closeChannel(leechingChannelId);
-          wire.paidStreamingExtension.leechingChannelId = null;
-          log.info(`Channel ${leechingChannelId} closed`);
+          return this.paymentChannelClient.closeChannel(leechingChannelId).then(() => {
```
kerzhner, line 526: I think @andrewgordstewart already asked this, but is there a reason to use `.then` promise syntax here instead of await?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -519,18 +523,18 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+            wire.paidStreamingExtension.leechingChannelId = null;
+            log.info(`Channel closed: ${leechingChannelId}`);
+          });
```
snario, line 529: Why the switch to `.then`?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -130,6 +131,11 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     log.info('> Cancelling download. Closing payment channels, and then removing torrent');
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
+      // I am only allowed to close the channel on my turn.
+      // If I don't pause the torrent, then I will continue to make payments, meaning the call to 
+      // this.closeChannels would race the `makePayment` function, and one of them would fail with a
+      // "Not my turn" error.
+      torrent.pause();
```
kerzhner, line 138: The [documentation for pausing](https://webtorrent.io/docs) says: Temporarily stop connecting to new peers. Note that this does not pause new incoming connections, nor does it pause the streams of existing connections or their wires.

Given this, is it true that the leecher would stop making payments?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -120,6 +120,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     const torrent = this.torrents.find(t => t.infoHash === infoHash);
     if (torrent) {
       torrent.pause(); // the paymentChannelClosing is done at the moment of payment
+      this.closeChannels(torrent, true);
```
kerzhner, line 123: Just to make sure I understand the current application correctly. This `pause` method that includes this line is never called, right?

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: I wanted to chain promises, but I'll make an extra function and use that instead, so that I maintain consistency.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: I wanted to chain promises, but I'll make an extra function and use that instead, so that I maintain consistency.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: I wanted to chain promises, but I'll make an extra function and use that instead, so that I maintain consistency.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: Yes, according to the docs it only makes it stop making connections, but in Web3Torrent it also sets the `torrent.pause` to `true` which, in here: https://github.com/statechannels/monorepo/compare/w3t-close-channel?expand=1#diff-a84153ed9f3ab2bd13c41bca43b584c7R404-R405
it makes the leecher stop making payments.
This has the effect that the leecher doesn't pollutes the payment channel with messages when the closing flow is running, and avoids *some* errors on that flow.

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: The `web3torrent.pause()` method is not being used currently, no.

packages/web3torrent/src/library/web3torrent-lib.ts
kerzhner: Do we need a separate function? Would this work?
```suggestion
          await this.paymentChannelClient.closeChannel(leechingChannelId);
          wire.paidStreamingExtension.leechingChannelId = null;
          log.info(`Channel closed: ${leechingChannelId}`);```

File: packages/web3torrent/src/library/web3torrent-lib.ts (hunk 5 above)
Comment (andrewgordstewart) [reply packages/web3torrent/src/library/web3torrent-lib.ts:529]: I'm curious why you wanted to chain promises?

Comment (spersico) [reply packages/web3torrent/src/library/web3torrent-lib.ts]: https://github.com/statechannels/monorepo/pull/1969/commits/034e40f790457b167989f3792ebeda37eaecdbb4

File: packages/web3torrent/src/library/web3torrent-lib.ts (hunk 4 above)
Comment (spersico) [reply packages/web3torrent/src/library/web3torrent-lib.ts]: https://github.com/statechannels/monorepo/pull/1969/commits/034e40f790457b167989f3792ebeda37eaecdbb4

File: packages/web3torrent/src/library/web3torrent-lib.ts (hunk 3 above)
Comment (spersico) [reply packages/web3torrent/src/library/web3torrent-lib.ts]: https://github.com/statechannels/monorepo/pull/1969/commits/034e40f790457b167989f3792ebeda37eaecdbb4

File: packages/web3torrent/src/clients/payment-channel-client.ts (hunk 1 above)
Comment (spersico) [reply packages/web3torrent/src/clients/payment-channel-client.ts]: https://github.com/statechannels/monorepo/pull/1969/commits/034e40f790457b167989f3792ebeda37eaecdbb4

File: packages/web3torrent/src/library/web3torrent-lib.ts (hunk 5 above)
Comment (spersico) [reply packages/web3torrent/src/library/web3torrent-lib.ts]: > I'm curious why you wanted to chain promises?

Because the seeding or leeching `channelId` should only be closed AFTER the channel has been closed (because it's used by the communication flow). By chaining promises, we can launch `channelClose` for all wires at the same time (with an array and using the Promise.all), which results in a faster torrent cancelling.

File: packages/web3torrent/src/library/web3torrent-lib.ts (hunk 4 above)
Comment (spersico) [reply packages/web3torrent/src/library/web3torrent-lib.ts]: > Do we need a separate function? Would this work?

this would close the channel one by one, unless i'm wrong.

File: packages/web3torrent/src/library/web3torrent-lib.ts (hunk 5 above)
Comment (andrewgordstewart) [reply packages/web3torrent/src/library/web3torrent-lib.ts]: Hmm, I'm not sure I follow. Two awaits in a row is semantically equivalent to one promise plus one `then`, isn't it?

Comment (spersico) [reply packages/web3torrent/src/library/web3torrent-lib.ts]: IIRC no:
```
await Promise()
await Promise()
```
is not equivalent to
```
await Promise().then(()=>Promise())
```
but to
```
async ()=>{
  await Promise()
  await Promise()
}
```

And for a `Promise.all` I understand that you need an array of single promises...
..
...
now that I think about it, in this case, you might be right... well, it's been changed already.

PR #1970, opened 2020-05-26
[devtools] Patch @statechannels/devtools

geoknee:
(add missing devDependency)


geoknee:
@snario spotted another thing to fix f6f63d9

PR #1971, opened 2020-05-26
[nitro-protocol] Tweaks before publishing package

geoknee:
As I write the nitro-tutorial, I am spotting some things that could be nicer about this package (for external consumers).

### DIFF
--- a/packages/nitro-protocol/contracts/NitroAdjudicator.sol
+++ b/packages/nitro-protocol/contracts/NitroAdjudicator.sol
@@ -194,10 +194,18 @@ contract NitroAdjudicator is Adjudicator, ForceMove {
     * @param ab Variable parts of each of the pair of states
     * @param turnNumB turnNum of the later state of the pair.
     * @param appDefinition Address of deployed contract containing application-specific validTransition function.
-    * @return true if the later state is a validTransition from its predecessor, false otherwise.
+    * @return true if the later state is a validTransition from its predecessor, reverts otherwise.
     */
-    function validTransition(nParticipants, isFinalAB, ab, turnNumB, appDefinition) public view {
+    function validTransition(
+        uint256 nParticipants,
+        bool[2] memory isFinalAB, // [a.isFinal, b.isFinal]
+        ForceMoveApp.VariablePart[2] memory ab, // [a,b]
+        uint256 turnNumB,
+        address appDefinition
+        ) public pure returns (bool) {
+
         return _requireValidTransition(nParticipants, isFinalAB, ab, turnNumB, appDefinition);
+
     }
 }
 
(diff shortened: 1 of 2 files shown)


packages/nitro-protocol/contracts/NitroAdjudicator.sol
```diff
@@ -184,6 +184,29 @@ contract NitroAdjudicator is Adjudicator, ForceMove {
...
+        uint256 nParticipants,
+        bool[2] memory isFinalAB, // [a.isFinal, b.isFinal]
+        ForceMoveApp.VariablePart[2] memory ab, // [a,b]
+        uint256 turnNumB,
+        address appDefinition
+        ) public pure returns (bool) {
+
+        return _requireValidTransition(nParticipants, isFinalAB, ab, turnNumB, appDefinition);
+
+    }
```
geoknee, line 209: This will be more expensive (gas wise) than the function it wraps, due to it being a ` public` rather than `internal` function: that is, if this function ever gets called as a part of a transaction (`eth_call` is free). 

All of the transactions that a SC wallet might use still make use only of the wrapped function, so gas consumption of those transactions will be unaffected.

packages/nitro-protocol/contracts/NitroAdjudicator.sol
```diff
@@ -184,6 +184,29 @@ contract NitroAdjudicator is Adjudicator, ForceMove {
...
+    function validTransition(
+        uint256 nParticipants,
+        bool[2] memory isFinalAB, // [a.isFinal, b.isFinal]
+        ForceMoveApp.VariablePart[2] memory ab, // [a,b]
+        uint256 turnNumB,
+        address appDefinition
+        ) public pure returns (bool) {
+
+        return _requireValidTransition(nParticipants, isFinalAB, ab, turnNumB, appDefinition);
+
```
snario, line 208:
```suggestion
```

packages/nitro-protocol/src/index.ts
```diff
@@ -53,7 +53,7 @@ export {
   hashOutcome,
 } from './contract/outcome';
 
-export {State, VariablePart, getVariablePart} from './contract/state';
+export {State, VariablePart, getVariablePart, getFixedPart} from './contract/state';
```
snario, line 56: ?

geoknee: It's a function that  consumers of this package will probably want to use?

snario: Oh I misread this as an import, nevermind.

packages/nitro-protocol/contracts/NitroAdjudicator.sol
geoknee: We should discuss a strategy for tracking deployment addresses for each update to @statechannels/nitro-protocol.

PR #1978, opened 2020-05-26
Make persistent-seeder headless

snario:


geoknee:
I don't think this was a sufficient change to make it run headless. We also need to disable dappeteer in the environment variables.

https://github.com/statechannels/monorepo/blob/328793135811136792ff45de155713bcbd8b4c74/packages/e2e-tests/puppeteer/helpers.ts#L201

packages/persistent-seeder/persistent-seeder.ts
```diff
@@ -1,14 +1,16 @@
-import {setUpBrowser} from '@statechannels/e2e-tests/puppeteer/helpers';
+import {setUpBrowser, setupFakeWeb3} from '@statechannels/e2e-tests/puppeteer/helpers';
 import {uploadFile} from '@statechannels/e2e-tests/puppeteer/scripts/web3torrent';
 import {APP_URL as WEB3TORRENT_URL} from '@statechannels/e2e-tests/puppeteer/constants';
 
 export async function persistentSeeder(): Promise<void> {
   console.log('Opening browser');
-  const {browser, metamask} = await setUpBrowser(false, 0);
+  const {browser, metamask} = await setUpBrowser(true);
```
geoknee, line 7: This argument should probably be read in from the environment variable, which is how we currently handle configuration like this.

PR #1985, opened 2020-05-27
Fix web3torrent badge overlapping wallet modal

### LINKED ISSUE #1975: Badges show over the wallet
Is the z-index too high?

Screen Shot 2020-05-26 at 5 37 46 PM

snario:
Fixes #1975.

Using 1300 because it's Google's Material design convention https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/styles/zIndex.js


snario:
I think a tooltip should only appear if you had a hover event over some element, and that element will be behind the modal

PR #1986, opened 2020-05-27
Update 'peers connected' UI to reflect active channels

### LINKED ISSUE #1973: Number of peers doesn't match
From Mike: this is because the number of peers is taken from webtorrent, and the peers in the table are taken from open channels.

Suggestion: get number of peers to just be a count of the rows in the table.

Screen Shot 2020-05-26 at 5 23 30 PM

snario:
Fixes #1973

### DIFF
--- a/packages/web3torrent/src/components/torrent-info/peer-network-stats/PeerNetworkStats.tsx
+++ b/packages/web3torrent/src/components/torrent-info/peer-network-stats/PeerNetworkStats.tsx
@@ -7,6 +7,10 @@ export type PeerNetworkStatsProps = {torrent: TorrentUI};
 
 const PeerNetworkStats: React.FC<PeerNetworkStatsProps> = ({torrent}: PeerNetworkStatsProps) => {
   const showTimeRemaining = !torrent.originalSeed && torrent.parsedTimeRemaining;
+  const numChannels = torrent.wires.filter(
+    wire =>
+      wire.paidStreamingExtension.leechingChannelId || wire.paidStreamingExtension.seedingChannelId
+  ).length;
   return (
     !IdleStatuses.includes(torrent.status) && (
       <section className="PeerNetworkStats">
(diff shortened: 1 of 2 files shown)


packages/web3torrent/src/components/torrent-info/peer-network-stats/PeerNetworkStats.tsx
```diff
@@ -7,14 +7,19 @@ export type PeerNetworkStatsProps = {torrent: TorrentUI};
...
+  ).length;
   return (
     !IdleStatuses.includes(torrent.status) && (
       <section className="PeerNetworkStats">
         {showTimeRemaining && `${torrent.parsedTimeRemaining}. `}
         {prettier(torrent.done || !torrent.downloadSpeed ? 0 : torrent.downloadSpeed)}
         /s down, {prettier(!torrent.uploadSpeed ? 0 : torrent.uploadSpeed)}/s up
         <br />
-        <strong data-test-selector="numPeers">{torrent.numPeers}</strong> Peers connected
+        <strong data-test-selector="numPeers">{numChannels}</strong> running channel
```
geoknee, line 21:
```suggestion
        <strong data-test-selector="numPeers">{numChannels}</strong> running channels
```

snario:
> What happens when a channel closes? Does numChannels decrement?

Yes

packages/web3torrent/src/components/torrent-info/peer-network-stats/PeerNetworkStats.tsx
snario: There is a statement after this checking for `=== 1` and add an `s` if false

geoknee: Gotcha, sorry I missed it.

PR #1991, opened 2020-05-27
Fixes error due to 'in' operator not working

snario:


packages/channel-provider/src/messaging-service.ts
```diff
@@ -103,7 +103,7 @@ export class MessagingService {
         window.removeEventListener('message', listener);
         logger.info({response: event.data}, 'Received response');
         resolve(event.data.result);
-      } else if (isJsonRpcErrorResponse(event.data)) {
+      } else if (event.data && event.data.jsonrpc && isJsonRpcErrorResponse(event.data)) {
```
andrewgordstewart, line 106: Should this runtime check happen in `isJsonRpcErrorResponse`

kerzhner: Another proposal for this issue: https://github.com/statechannels/monorepo/pull/1994

kerzhner:
@snario should we close this PR?

PR #1994, opened 2020-05-27
[channel-provider] log non jsonrpc responses as errors

kerzhner:


packages/channel-provider/src/messaging-service.ts
```diff
@@ -90,21 +90,20 @@ export class MessagingService {
...
+        if (isJsonRpcResponse(event.data) && event.data.id === message.id) {
+          if (callback) {
+            callback(event.data.result);
+          }
+          this.acknowledge();
+          window.removeEventListener('message', listener);
+          logger.info({response: event.data}, 'Received response');
+          resolve(event.data.result);
+        } else if (isJsonRpcErrorResponse(event.data)) {
+          reject(new Error(`Wallet Error ${event.data.error.code}: ${event.data.error.message}`));
```
andrewgordstewart, line 103: I think this is incorrect: what if it is a `isJsonRpcErrorResponse`, but for a different message (with a different `id`)?

PR #1996, opened 2020-05-27
xstate-wallet: log Deposited and AssetTransferred events

kerzhner:
Hoping that this will help investigate https://github.com/statechannels/monorepo/issues/1995.


andrewgordstewart:
> Fyi, all contract events are currently logged. On mobile right now but the code is in the `chain.ts` file

I might have misremembered, since I'm unable to find the code I though was there.

PR #2002, opened 2020-05-28
Web3Torrent Client and library refactor and docs improvement

spersico:
This PR does 3 things:
- It adds jsdocs for every method in the web3torrent library.
- It cleans up a lot of logs, by deconstructing props and formatting the text.
- It refactors the Web3TorrentClient to clean it up a LOT. Moving logic into the PaymentChannelClient, and to the Web3torrent library.
- It simplifies the TorrentEvents.NOTICE STOP event handler.

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -102,7 +102,7 @@ export class PaymentChannelClient {
 
     if (FUNDING_STRATEGY !== 'Direct' && !(await doesBudgetExist())) {
       // TODO: This only checks if a budget exists, not if we have enough funds in it
-      log.info('Direct Funding - Creating Budget');
+      log.info('Virtual Funding - Creating Budget');
       await this.createBudget(INITIAL_BUDGET_AMOUNT);
     }
   }
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -9,13 +9,16 @@ import {
   HUB,
   FIREBASE_PREFIX,
   fireBaseConfig,
-  FUNDING_STRATEGY
+  FUNDING_STRATEGY,
+  INITIAL_BUDGET_AMOUNT
 } from '../constants';
 import {AddressZero} from 'ethers/constants';
 import * as firebase from 'firebase/app';
 import 'firebase/database';
 import {map, filter, first} from 'rxjs/operators';
 import {logger} from '../logger';
+import _ from 'lodash';
+
 const log = logger.child({module: 'payment-channel-client'});
 const hexZeroPad = utils.hexZeroPad;
 
@@ -91,6 +94,17 @@ export class PaymentChannelClient {
     await this.channelClient.provider.enable();
 
     log.info('payment channel client enabled');
+
+    const doesBudgetExist = async () => {
+      const budget = await this.getBudget();
+      return !!budget && !_.isEmpty(budget);
+    };
+
+    if (FUNDING_STRATEGY !== 'Direct' && !(await doesBudgetExist())) {
+      // TODO: This only checks if a budget exists, not if we have enough funds in it
+      log.info('Direct Funding - Creating Budget');
+      await this.createBudget(INITIAL_BUDGET_AMOUNT);
+    }
   }
 
   private initializeHubComms() {
--- a/packages/web3torrent/src/clients/web3torrent-client.ts
+++ b/packages/web3torrent/src/clients/web3torrent-client.ts
@@ -1,6 +1,6 @@
 import React from 'react';
 import {WebTorrentAddInput, WebTorrentSeedInput, ExtendedTorrent} from '../library/types';
-import WebTorrentLibrary from '../library/web3torrent-lib';
+import Web3TorrentLibrary from '../library/web3torrent-lib';
 import {Status} from '../types';
 import {paymentChannelClient} from './payment-channel-client';
 
--- a/packages/web3torrent/src/clients/web3torrent-client.ts
+++ b/packages/web3torrent/src/clients/web3torrent-client.ts
@@ -4,7 +4,7 @@ import Web3TorrentLibrary from '../library/web3torrent-lib';
 import {Status} from '../types';
 import {paymentChannelClient} from './payment-channel-client';
 
-export const web3TorrentClient = new WebTorrentLibrary({paymentChannelClient});
+export const web3TorrentClient = new Web3TorrentLibrary({paymentChannelClient});
 
 export const Web3TorrentClientContext = React.createContext(web3TorrentClient);
 
--- a/packages/web3torrent/src/clients/web3torrent-client.ts
+++ b/packages/web3torrent/src/clients/web3torrent-client.ts
@@ -1,50 +1,28 @@
 import React from 'react';
 import {WebTorrentAddInput, WebTorrentSeedInput, ExtendedTorrent} from '../library/types';
-import WebTorrentPaidStreamingClient from '../library/web3torrent-lib';
+import WebTorrentLibrary from '../library/web3torrent-lib';
 import {Status} from '../types';
 import {paymentChannelClient} from './payment-channel-client';
-import {defaultTrackers, INITIAL_BUDGET_AMOUNT, FUNDING_STRATEGY} from '../constants';
-import _ from 'lodash';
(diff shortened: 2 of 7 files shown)


spersico:
image

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -91,6 +94,17 @@ export class PaymentChannelClient {
...
     log.info('payment channel client enabled');
+
+    const doesBudgetExist = async () => {
+      const budget = await this.getBudget();
+      return !!budget && !_.isEmpty(budget);
+    };
+
+    if (FUNDING_STRATEGY !== 'Direct' && !(await doesBudgetExist())) {
+      // TODO: This only checks if a budget exists, not if we have enough funds in it
+      log.info('Direct Funding - Creating Budget');
```
kerzhner, line 105: Is this actually virtual funding?
```suggestion
      log.info('Virtual Funding - Creating Budget');
```

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -1,62 +1,29 @@
 import React from 'react';
 import {WebTorrentAddInput, WebTorrentSeedInput, ExtendedTorrent} from '../library/types';
-import WebTorrentPaidStreamingClient from '../library/web3torrent-lib';
+import WebTorrentLibrary from '../library/web3torrent-lib';
```
kerzhner, line 3:
```suggestion
import Web3TorrentLibrary from '../library/web3torrent-lib';
```

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -1,62 +1,29 @@
...
 import {paymentChannelClient} from './payment-channel-client';
-import {defaultTrackers, INITIAL_BUDGET_AMOUNT, FUNDING_STRATEGY} from '../constants';
-import _ from 'lodash';
-import {track} from '../analytics';
 
-export const web3TorrentClient = new WebTorrentPaidStreamingClient({
-  paymentChannelClient: paymentChannelClient,
-  tracker: {announce: defaultTrackers}
-});
+export const web3TorrentClient = new WebTorrentLibrary({paymentChannelClient});
```
kerzhner, line 7:
```suggestion
export const web3TorrentClient = new Web3TorrentLibrary({paymentChannelClient});
```

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -1,62 +1,29 @@
 import React from 'react';
```
kerzhner, line 1: Do we need to have a `web3torrent-client`? Seems like we might be able to get rid of this level of indirection and invoke methods of web3torrent-lib directly.

packages/web3torrent/src/clients/web3torrent-client.ts
```diff
@@ -1,62 +1,29 @@
...
-  return new Promise((resolve, reject) =>
-    web3TorrentClient.seed(input, {...torrentNamer(input)}, (torrent: any) => {
-      resolve({
-        ...torrent,
-        status: Status.Seeding,
-        originalSeed: true
-      });
-    })
+  return new Promise(resolve =>
+    web3TorrentClient.seed(input, (torrent: any) =>
```
kerzhner, line 25: Is there a reason to have callback style functions in `web3torrent-lib`? Perhaps the original intention was to match the signature of the overridden webtorrent functions. But do we think that's important?

At the moment, simplicity and clarity is likely more important than matching method signatures between webtorrent and web3torrent.

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -45,58 +44,68 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
     this.outcomeAddress = null;
   }
 
+  /**
+   * From a string, a file, a buffer, or an array from the previous, it creates a torrent,
+   * adds it to the client to be shared and announces it to the tracker.
+   */
   seed(
     input: WebTorrentSeedInput,
     optionsOrCallback?: TorrentOptions | TorrentCallback,
```
kerzhner, line 85: Should we get rid of the complexity of optionsOrCallback and create a function signature that matches our needs?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -148,52 +162,46 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-        peerAccount: wire.paidStreamingExtension.peerAccount,
-        seedingChannelId: wire.paidStreamingExtension.seedingChannelId
-      },
-      '<< unblockPeer: finish'
-    );
+    log.info({from: torrentPeers, peerAccount, seedingChannelId}, '<< unblockPeer: finish');
   }
 
-  // Note, this is only used by unit tests.
+  /** Toggle between Blocked/Unlocked a channel for a torrent. */
```
kerzhner, line 204: Should we move this to a test utility module since this is only used by tests?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -390,6 +395,10 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     );
   }
 
+  /**
+   * Sets the behaviour of a torrent, once created.
+   * This is, toguether with SetupWire, is what defines the behaviour of each created torrent in Web3Torrent.
```
kerzhner, line 400:
```suggestion
   * This is, together with SetupWire, is what defines the behaviour of each created torrent in Web3Torrent.
```

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -555,25 +568,25 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
   protected jumpStart(torrent: ExtendedTorrent, wire: PaidStreamingWire) {
     if (torrent.done) {
       log.info('<< JUMPSTART: FINISHED');
-      log.trace({torrent, wire});
+      log.trace({torrent, wire: wire.paidStreamingExtension.serialize()});
       return;
     }
     log.info({requests: wire.requests}, `<< START ${wire.paidStreamingExtension.peerAccount}`);
     log.trace({pieces: torrent.pieces});
     (torrent as any)._updateWireWrapper(wire); // TODO: fix this type, if its the solution
```
kerzhner, line 580: Do we know what this does or if it is necessary?

packages/web3torrent/src/clients/web3torrent-client.ts
spersico: I didn't want to change things too much. This PR was really big as is. I could take a look into removing it.

packages/web3torrent/src/clients/web3torrent-client.ts
spersico: I decided to just match the function signature of the original function. Maybe we could refactor those functions in another PR?

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: I decided to just match the function signature of the original function. Maybe we could refactor those functions in another PR?

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: good idea!. Will do

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: I don't know what `log.trace` does, I just know that using the wire object directly has problems (it doesn't appear correctly if the wire is not serialized.

kerzhner: Just to be clear, I am referring to `(torrent as any)._updateWireWrapper(wire);`

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: https://github.com/statechannels/monorepo/pull/2002/commits/8cd21365370688b097a10b0cec78efd544f801e6

packages/web3torrent/src/library/web3torrent-lib.ts
spersico: I don't remember, with today's PR I will check and document it.

Issue #1992, opened 2020-05-27
Active channels ending in the "disconnected" state


snario:

tomclose:
We initially thought this was a deadlock issue, but I'm not so sure. In the list below, there are multiple wires with most counterparties _but not with the original seeder_ (the one that looks like a face, with 196 transactions). George and I noticed this behaviour consistently over multiple browsers.

Hypothesis:
* The seeder always has data that I need, so I keep requesting data from them, the torrent wire stays open, and the channel stays alive
* Other peers sometimes have data for me, and sometimes don't. When they don't have data for me, I don't request data from them, the torrent wire drops, and the channel is left in a disconnected state.

Possible fixes:
1. We should handle closing the channel in the case where someone doesn't have data for you for some time. This should happen before the torrent timeout kicks in. This would lead to the channels ending in a closed state, instead of in a disconnected state
2. In some cases, we might still end up with an open channel without a torrent wire. It's probably worth checking to see if there are open channels with a given peer, in a given direction (send/receive) before setting up a new one. Even if you don't want to reuse these channels, you could use this opportunity to close them. (But we should probably reuse, if we can.)

![Screen Shot 2020-05-28 at 10 46 35 AM](https://user-images.githubusercontent.com/29596/83126351-9912a880-a0d0-11ea-848e-5dc9c8d09979.png)

snario:
>  In some cases, we might still end up with an open channel without a torrent wire. It's probably worth checking to see if there are open channels with a given peer, in a given direction (send/receive) before setting up a new one. Even if you don't want to reuse these channels, you could use this opportunity to close them. (But we should probably reuse, if we can.)

We've decided not to do this in v1, but the work is there in #2004 ([`george/reuse-channels`](https://github.com/statechannels/monorepo/tree/george/reuse-channels))

geoknee:
As far as I know we aren't seeing this problem any more.

PR #2004, opened 2020-05-28
[web3torrent] Reuse channels on new wires

geoknee:
During wire setup, we currently attach a callback, firing on every wire REQUEST received by the seeder, that creates a new channel on the _first request for this wire_. 

This PR makes it so that on that first request, we check to see whether our client's `peersList` contains a channel with the peer making the request. If so, we we will use that. If not, proceed as before. 

I believe this should stop the client getting polluted with new channels in the case of a deadlock or period of inactivity.

See #1992

### DIFF
--- a/.env.circle-integration-w3t-with-hub
+++ b/.env.circle-integration-w3t-with-hub
@@ -10,12 +10,12 @@ USE_GANACHE_DEPLOYMENT_CACHE = 'true'
 
 # For e2e-test
 BROWSER_LOG_DESTINATION = 'browser.log'
-HEADLESS = 'false'
+HEADLESS = 'true'
 USE_DAPPETEER = 'false'
 TARGET_NETWORK = 'localhost'
 WEB3TORRENT_URL = 'http://localhost:3000'
 SCREENSHOT_DIR = 'screenshots'
-CLOSE_BROWSERS = 'false'
+CLOSE_BROWSERS = 'true'
 
 ## simple-hub
 SIMPLE_HUB_DEPLOYER_ACCOUNT_INDEX = '1'
--- a/.env.circle-integration-w3t-with-hub
+++ b/.env.circle-integration-w3t-with-hub
@@ -10,12 +10,12 @@ USE_GANACHE_DEPLOYMENT_CACHE = 'true'
 
 # For e2e-test
 BROWSER_LOG_DESTINATION = 'browser.log'
-HEADLESS = 'true'
+HEADLESS = 'false'
 USE_DAPPETEER = 'false'
 TARGET_NETWORK = 'localhost'
 WEB3TORRENT_URL = 'http://localhost:3000'
 SCREENSHOT_DIR = 'screenshots'
-CLOSE_BROWSERS = 'true'
+CLOSE_BROWSERS = 'false'
 
 ## simple-hub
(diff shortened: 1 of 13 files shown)


packages/web3torrent/src/library/types.ts
```diff
@@ -134,22 +134,24 @@ export type WebTorrentAddInput = string | Buffer | ParseTorrent;
...
   wire: PaidStreamingWire | PeerWire;
   buffer: string;
   beneficiaryBalance: string;
   uploaded: number;
 };
 
-export type TorrentPeers = {
-  [key: string /* ChannelId */]: PeerByTorrent;
+export type PeersByChannel = {
+  [key: string /* ChannelId */]: PeerByChannel;
```
snario, line 146:
```suggestion
  [channelId: string]: PeerByChannel;
```

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -234,31 +234,54 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+        const {peerAccount: peer, isForceChoking, seedingChannelId} = wire.paidStreamingExtension;
+
+        if (!seedingChannelId) {
+          // Check to see if we have a running channel with this peer..
+          wire.paidStreamingExtension.seedingChannelId = _.findKey(
+            this.channelsByInfoHash[torrent.infoHash],
+            {
+              id: peer
+            }
+          );
```
geoknee, line 247: This needs to be modified such that we don't hook into channels that aren't `running`.

andrewgordstewart:
How can you ensure that the channel isn't already in use for some other purpose?

geoknee:
It's a good question. If we want to repurpose an existing channel onto a new wire, we should make sure there is no existing wire that is using the channel?

packages/web3torrent/src/library/types.ts
geoknee: 42a86b9

packages/web3torrent/src/library/web3torrent-lib.ts
geoknee: 8644bcb

.env.circle-integration-w3t-with-hub
```diff
@@ -10,12 +10,12 @@ USE_GANACHE_DEPLOYMENT_CACHE = 'true'
...
 # For e2e-test
 BROWSER_LOG_DESTINATION = 'browser.log'
-HEADLESS = 'true'
+HEADLESS = 'false'
 USE_DAPPETEER = 'false'
 TARGET_NETWORK = 'localhost'
 WEB3TORRENT_URL = 'http://localhost:3000'
 SCREENSHOT_DIR = 'screenshots'
-CLOSE_BROWSERS = 'true'
+CLOSE_BROWSERS = 'false'
```
snario, line 18: Is this just for development?

geoknee: Yeah thanks for spotting 013ec4f

geoknee:
Decided to close this because: 

- we are doing a feature freeze
- it seems quite a risky thing to tinker with
- we don't expect the situation to arise all that often / we don't see it happen as often since some recent fixes went in (the situation where a wire drops)

PR #2006, opened 2020-05-28
🐷 Add delay before starting the close channel logic

andrewgordstewart:
With this delay added, both the `seed-download-cancel` and `stress` tests pass 100% of the time locally.


kerzhner:
Per conversation with @andrewgordstewart, we should not need this PR given https://github.com/statechannels/monorepo/pull/1997

andrewgordstewart:
> Maybe leave a comment that this hack 🪓 should be removed?

https://github.com/statechannels/monorepo/issues/2009

I've already made an issue, so I think that's unnecessary

PR #2010, opened 2020-05-28
Kill channels after two keep-alive pings with zero progress

snario:
Implements @tomclose's [suggestion](https://github.com/statechannels/monorepo/issues/1992#issuecomment-635242533) from #1992:

> We should handle closing the channel in the case where someone doesn't have data for you for some time. This should happen before the torrent timeout kicks in. This would lead to the channels ending in a closed state, instead of in a disconnected state

I tested on [`ags/close-channels-on-my-turn`](https://github.com/statechannels/monorepo/pull/1997) by starting the stress test and killing two of the windows leaving no peer with the full file but two partially downloading from each other and it worked.

Screen Shot 2020-05-28 at 14 32 20

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -228,10 +228,12 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
         wire._clearTimeout();
       }
 
-      if (!wire.amChoking && wire.downloaded === wire['prevDownloaded']) {
-        this.closeChannel(wire, wire.paidStreamingExtension.leechingChannelId);
-      } else {
-        wire['prevDownloaded'] = wire.downloaded;
+      if (wire.paidStreamingExtension.leechingChannelId) {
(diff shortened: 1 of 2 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -227,6 +227,13 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       if (!torrent.done && wire.amChoking) {
         wire._clearTimeout();
       }
+
+      if (!wire.amChoking && wire.downloaded === wire['prevDownloaded']) {
+        this.closeChannel(wire, wire.paidStreamingExtension.leechingChannelId);
+      } else {
+        wire['prevDownloaded'] = wire.downloaded;
+      }
```
snario, line 235: Not sure if this is the best way to retain this tiny bit of local data on the wire but it's convenient and short.

spersico: To be consistent, I would put this data in the paidStreamingExtension. All of web3torrent's wires extended data that we handle is in there already. Seems like a cleaner solution than this.

Issue #2003, opened 2020-05-28
Loads of zero byte disconnected channels


tomclose:
George and I ended in a state where:
* The original seeder (George chrome) had only 5 connections
* Tom's Brave and George's Opera managed to download the file successfully
* Tom's Chrome and Tom's Firefox were both stuck at 39MB/129MB
* And the two browsers that were stuck open channels, but no data is transferred.

![Screen Shot 2020-05-28 at 11 18 34 AM](https://user-images.githubusercontent.com/29596/83129698-19d3a380-a0d5-11ea-9452-607e686d6df2.png)

tomclose:
Unexplained things:
* Even though there doesn't seem have been any data exchanged with the original seeder, there's still an active channel, that hasn't been killed by webtorrent

tomclose:
The zero byte channels are stuck in the pre-fund setup state. Here's an example:

```
key: "0x0095c6fcd3c639d8f9cca26baedb79153298fd691c6d15da4ca1e7aa562e9c9e"
value:
applicationDomain: "web3torrent.statechannels.org"
channelConstants:
appDefinition: "0xde8585b18e89ad34521f87f236dd745b1c4af99b"
chainId: "3"
challengeDuration: "300"
channelNonce: "8"
participants: Array(2)
0: {participantId: "0x0E9f7295BFC5E1783B7629221ACa82BcdF25e0F8", signingAddress: "0x0E9f7295BFC5E1783B7629221ACa82BcdF25e0F8", destination: "0x00000000000000000000000087e0ED760fb316eeb94Bd9cF23D1d2BE87aCe3d8"}
1: {participantId: "0x675221673cE0bd41b351433991967D8DFd3B2622", signingAddress: "0x675221673cE0bd41b351433991967D8DFd3B2622", destination: "0x000000000000000000000000d4Fa489Eacc52BA59438993f37Be9fcC20090E39"}
length: 2
funding: undefined
myIndex: 1
stateVariables: Array(1)
0:
appData: "0x0000000000000000000000000000000000000000000000000000000000000000"
appDefinition: "0xde8585b18e89ad34521f87f236dd745b1c4af99b"
chainId: "3"
challengeDuration: {_hex: "0x012c", _isBigNumber: true}
channelNonce:
_hex: "0x08"
_isBigNumber: true
isFinal: false
outcome:
allocationItems: (2) [{…}, {…}]
[...]

PR #2019, opened 2020-05-29
Use 'id' instead of 'key' for absolute xstate node reference

snario:
It [seems like](https://xstate.js.org/docs/guides/ids.html#identifying-state-nodes) `id` is what you're supposed to use when doing direct references of state nodes. Might help with #2003?

### DIFF
--- a/packages/xstate-wallet/src/workflows/ledger-funding.ts
+++ b/packages/xstate-wallet/src/workflows/ledger-funding.ts
@@ -53,6 +53,7 @@ const checkTarget = (store: Store) => async (ctx: Init) => {
 };
 
 export const config: MachineConfig<any, any, any> = {
+  id: WORKFLOW,
   key: WORKFLOW,
   initial: 'acquiringLock',
   states: {


snario:
Since `onError` is used in the state that we see the workflow gets stuck at:

https://github.com/statechannels/monorepo/blob/10fe97c412306ddcc5de4da8827f1da1848efb37/packages/xstate-wallet/src/workflows/ledger-funding.ts#L45

And `onError` targets a specific state node:

https://github.com/statechannels/monorepo/blob/10fe97c412306ddcc5de4da8827f1da1848efb37/packages/xstate-wallet/src/workflows/ledger-funding.ts#L41-L42

It could be that the state is being set to `#ledger-funding.failure` and not `#failure`, which is the thing that should be either sending out error information or releasing the lock or both:

https://github.com/statechannels/monorepo/blob/10fe97c412306ddcc5de4da8827f1da1848efb37/packages/xstate-wallet/src/workflows/ledger-funding.ts#L67-L69

PR #2020, opened 2020-05-29
Use InfuraProvider when mocking for persistent-seeder and e2e tests

snario:
Should address https://sentry.io/organizations/statechannels/issues/1697515651/?project=5236239&referrer=slack

### DIFF
--- a/packages/xstate-wallet/src/utils/contract-utils.ts
+++ b/packages/xstate-wallet/src/utils/contract-utils.ts
@@ -1,11 +1,11 @@
 import {Contract} from 'ethers';
 import {ContractArtifacts} from '@statechannels/nitro-protocol';
 
-import {ETH_ASSET_HOLDER_ADDRESS} from '../config';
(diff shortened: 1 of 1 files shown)


packages/xstate-wallet/src/utils/contract-utils.ts
```diff
@@ -23,6 +23,9 @@ export function tokenAddress(assetHolderAddress: string): string | undefined {
 
 export function getProvider(): Web3Provider | JsonRpcProvider {
   if (window.ethereum) {
+    if (window.ethereum.mockingInfuraProvider) {
```
kerzhner, line 26: I wish we didn't need to reference mocks in production code 🤷

snario: Yeah I definitely agree. I started on some work to rip out the provider as a configuration option but haven't finished it yet.

PR #2022, opened 2020-05-29
Update/Close race

andrewgordstewart:
- [channel-provider] Make some type-related changes to the channel-provider package to make it possible/more comfortable to catch wallet errors
- [web3torrent] Retry `CloseChannel` when it fails due to a NotYourTurn error
- [web3torrent] Catch NotYourTurn errors when calling `UpdateChannel`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,7 @@ save_built_artifacts: &save_built_artifacts
       - packages/web3torrent/build
       - packages/xstate-wallet/build
 
+
 restore_built_artifacts: &restore_built_artifacts
   restore_cache:
     # Restore from the most recently cached version of lib, using a partial key match
@@ -300,6 +301,36 @@ jobs:
           file: integration-test-web3torrent-stats
       - upload_e2e_screenshots
 
+  integration-test-web3torrent-optional:
+    resource_class: large
+    working_directory: /home/circleci/project
+    docker:
+      - image: circleci/node:10.16.3-browsers
+    environment:
+      SC_ENV: circle-integration-w3t
+    steps:
+      - install_xvfb
+      - run_xvfb
+      - checkout
+      - log_stats:
+          file: integration-test-web3torrent-stats
+      - attach_workspace:
+          at: /home/circleci/project
+      - start_servers:
+          app: web3torrent
+      - run: (cd packages/e2e-tests && yarn test optional)
+      - store_artifacts:
+          path: /tmp/ganache.log
+      - store_artifacts:
+          path: /tmp/seed-download-cancel.pino.log
+      - store_artifacts:
+          path: /tmp/seed-download-cancel.browser.log
+      - store_artifacts:
+          path: packages/simple-hub/pino.log
+      - upload_logs:
+          file: integration-test-web3torrent-stats
+      - upload_e2e_screenshots
+
   integration-test-web3torrent-with-hub:
     resource_class: large
     working_directory: /home/circleci/project
@@ -509,9 +540,6 @@ workflows:
           requires:
             - prepare
       - integration-test-web3torrent:
-          requires:
-            - prepare
-      - integration-test-web3torrent-optional:
           requires:
             - prepare
       - integration-test-web3torrent-with-hub:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,6 @@ save_built_artifacts: &save_built_artifacts
       - packages/web3torrent/build
       - packages/xstate-wallet/build
 
-
 restore_built_artifacts: &restore_built_artifacts
   restore_cache:
     # Restore from the most recently cached version of lib, using a partial key match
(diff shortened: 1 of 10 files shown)


packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -280,31 +282,31 @@ export class PaymentChannelClient {
...
-     * Now, any UpdateChannel RPC should be finished, so when it's my turn, the CloseChannel
-     * call should succeed.
-     */
-
     const closing = this.channelState(channelId)
-      .pipe(first(cs => this.canUpdateChannel(cs)))
-      .subscribe(cs => {
+      .pipe(
+        filter(cs => this.canUpdateChannel(cs)),
+        take(5)
```
snario, line 288: Might warrant a comment as to why 5 is here

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -390,10 +392,18 @@ export class PaymentChannelClient {
...
+    try {
+      await this.updateChannel(channelId, {
+        beneficiary: {...beneficiary, balance: add(beneficiary.balance, amountWillPay)},
+        payer: {...payer, balance: subtract(payer.balance, amountWillPay)}
+      });
+    } catch (error) {
+      if (error.error.code === ErrorCode.UpdateChannel.NotYourTurn) {
+        logger.warn({channelId}, 'Possible race condition detected');
+      } else {
+        logger.error({error}, 'makePamyent: Unexpected error');
```
snario, line 404:
```suggestion
        logger.error({error}, 'makePayment: Unexpected error');
```

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -280,31 +282,31 @@ export class PaymentChannelClient {
...
+        take(5) // Limit to 5 attempts
+      )
+      .subscribe(async cs => {
         logger.info(
           {channelId, cs, me: this.mySigningAddress},
           "It's my turn, closing the channel"
         );
-        this.channelClient.closeChannel(channelId);
+        try {
+          await this.channelClient.closeChannel(channelId);
```
kerzhner, line 296: In the case when the seeder is initiating the channel close, this call always race (payment acceptance) [https://github.com/statechannels/monorepo/blob/master/packages/web3torrent/src/library/web3torrent-lib.ts#L340]

.circleci/config.yml
```diff
@@ -384,9 +384,9 @@ jobs:
       - store_artifacts:
           path: /tmp/ganache.log
       - store_artifacts:
-          path: /tmp/seed-download-cancel.pino.log
+          path: /tmp/multiple-leecher.pino.log
       - store_artifacts:
-          path: /tmp/seed-download-cancel.browser.log
+          path: /tmp/multiple-leecher.browser.log
```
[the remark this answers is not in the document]
andrewgordstewart: This is currently in process in the `ags/fix-artifacts` branch

packages/channel-client/src/types.ts
```diff
@@ -67,3 +68,19 @@ export interface EventsWithArgs {
   // TODO: Is `ChannelResult` the right type to use here?
   ChannelProposed: [ChannelResult];
 }
+
+export const ErrorCode: ErrorCodes = {
+  EnableEthereum: {EthereumNotEnabled: 100},
```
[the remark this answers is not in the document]
andrewgordstewart: They are fairly arbitrary. When organizing the client schema last week, I opted to use 400's for one API, 300's for another, etc.

 The code added here is simply a constant value -- the codes were already in place in the `ErrorCodes` type.

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -280,31 +282,31 @@ export class PaymentChannelClient {
...
-     */
-
     const closing = this.channelState(channelId)
-      .pipe(first(cs => this.canUpdateChannel(cs)))
-      .subscribe(cs => {
+      .pipe(
+        filter(cs => this.canUpdateChannel(cs)),
+        take(5) // Limit to 5 attempts
+      )
+      .subscribe(async cs => {
```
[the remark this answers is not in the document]
andrewgordstewart: I think so. I was thinking of adding it but never got to it

packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -280,31 +282,31 @@ export class PaymentChannelClient {
...
+        try {
+          await this.channelClient.closeChannel(channelId);
+        } catch (error) {
+          if (error.error.code === ErrorCode.CloseChannel.NotYourTurn) {
+            // Perhaps the application calls UpdateChannel around the same time as the CloseChannel call
+            logger.warn(
+              {channelId},
+              'Possible race condition detected -- is a concurrent UpdateChannel occurring?'
+            );
+            return;
```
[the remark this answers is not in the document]
andrewgordstewart: I don't think it's needed but I view it as being explicit that we are ok catching this error

packages/web3torrent/src/clients/payment-channel-client.ts
andrewgordstewart: So, the torrent client should probably switch modes and not accept payments. Then, there would be at most once race.

andrewgordstewart: So, the torrent client should probably switch modes and not accept payments. Then, there would be at most once race.

PR #2028, opened 2020-05-29
Log out version

lalexgap:
Now that we're relying on saved logs a lot more it probably make sense to log out the version so we can tell what changes are/aren't included when the log was generated.


andrewgordstewart:
Just a quick note: the version is already a part of the filename, so we have this data already.

This change is better, anyway. Filenames can change pretty easily

Issue #2033, opened 2020-05-31
[xstate-wallet] Budget Notifications never sent


geoknee:
As far as I can tell, we never call this function in the wallet:

https://github.com/statechannels/monorepo/blob/3c5d73e601166c9b788ad970081c917e60c74e41/packages/xstate-wallet/src/messaging.ts#L103-L110

Which means the app cannot display up to date information about the budget, since it currently expects these notifications so it can cache the budget in the payment-channel-client. 

I wasn't quite sure exactly where we should send this notification: in the application workflow? in the store?

geoknee:
@lalexgap Do you have thoughts?

lalexgap:
> @lalexgap Do you have thoughts?

I think we could create some `budget` related services or actions that we trigger from the workflow that 
- call the budget method in the `store` (aka `reserveFunds`)
- Sends out a budget notification 

I think this would follow a similar pattern that we use for `UpdateChannel` in the application workflow (https://github.com/statechannels/monorepo/blob/3c5d73e601166c9b788ad970081c917e60c74e41/packages/xstate-wallet/src/workflows/application.ts#L322)

It might make sense to create these in `workflow-utils` so we can share them between the `application` and `approve-budget` workflows.

Issue #2026, opened 2020-05-29
[web3torrent] Add additional UI to show ledger funds that are "locked" in app channels


geoknee:
Currently this UI reacts to an aggregate of funds over all payment channels : 

https://github.com/statechannels/monorepo/blob/50e77c3f81b64466584b012081bbc05baf9af5cf/packages/web3torrent/src/components/domain-budget-table/DomainBudgetTable.tsx#L43-L53
 
Spent/received are the numerators, spendBudget and recieveBudget are the denominators, and the ratio controls the position of our little speedometers: 

![Screenshot 2020-05-29 at 18 08 35](https://user-images.githubusercontent.com/1833419/83286200-6c958400-a1d7-11ea-852d-2bb8d6823427.png)

To me the idea of a budget a sort of license granted to wallet by the user, to sign off on changes to channels without prompting via the UI. Those changes in this instance are to my balance in any payment channel managed by web3torrent. I think the UI is doing a reasonable job of reflecting this.

If I run out of budget. Nothing super bad should happen. I just get another prompt from the wallet and increase my budget if I want to. No transactions necessary to recover from that.

I think there is also a need to also display information about the ledger channel with the hub, which is the funding source for all of these channels. It is one step higher in the funding tree, and is not covered by a budget, in the sense that we prompt the user when funding the ledger channel, and we also do not refer to any budget when allocating money from the ledger channel to a w3t payment channel. 

I would suggest that we want an *additional* component to show the following quantities of interest that relate to the ledger channel, namely: 

i) the amount allocated to Me
ii) the amount allocated to the Hub
iii) the sum of the amounts allocated to Payment channels.
iv) the amount I deposited on chain
v) the amount the hub has deposited on chain

If I run out of money in the ledger channel, this either requires a ledger channel top up or defunding of some running payment channels. 

This mockup is one way we could do it:
![Screenshot 2020-05-29 at 18 57 54](https://user-images.githubusercontent.com/1833419/83290287-517a4280-a1de-11ea-9630-eec5369be725.png)

A pot for me, a pot for the hub, and a pot for the payment channels which is divided into money that I should be able to get back into my pot, and money that my counterparties want to get back in their pot (which on my screen would go back into the hub's pot). Some of that (red gray, green gray) might need a challenge to recover.

Overall I think we should move towards thinking of budgets and ledger channels as fairly well independent of each other: up to now I think we have sometimes been using the terms interchangeably. 

But curious to know what others think.

[...]

PR #2034, opened 2020-05-31
[web3torrent] Improve budget UI

geoknee:
Work towards closing #2026 .

The main change is to consider your budget amounts and the running payment channels as fractions of the amount of money in your ledger channel (rather than the current approach which treats the money in your payment channels as fractions of your budget amounts). 

TODO
- [ ] Make it look nice. Colours? Tooltips? Fix styling on withdraw button
- [x] Fix the issue where budget amounts do not update on opening a new payment channel. ~Blocked on #2033~.
~Blocked on #2068~ still not seeing the updates I expect when funding a payment channel 
- [ ] Fix the issue that this UI component doesn't always seem to display itself #2063 
- [x] FIx broken tests `TimeoutError: waiting for selector ".domain-budget-table > tbody > tr" failed: timeout 30000ms exceeded
 `

Red section: hub's allocation ( = my receiving budget)
Orange section: my allocation ( = my spending budget)
Thin light orange / light red : total spent/ received in ("locked") payment channels. 

Screenshot 2020-05-31 at 10 09 59


packages/web3torrent/src/components/domain-budget-table/DomainBudgetTable.tsx
```diff
@@ -52,71 +42,53 @@ export const DomainBudgetTable: React.FC<DomainBudgetTableProps> = props => {
 
   const receiveBudget = bigNumberify(budgetCache.budgets[0].availableReceiveCapacity);
 
-  const inverseSpentFraction = spent.gt(0) ? spendBudget.div(spent).toNumber() : undefined;
```
geoknee, line 55: This was the wrong way to think about budgets.

packages/web3torrent/src/components/domain-budget-table/DomainBudgetTable.tsx
```diff
@@ -52,71 +42,53 @@ export const DomainBudgetTable: React.FC<DomainBudgetTableProps> = props => {
...
-  const inverseRecievedFraction = received.gt(0)
-    ? receiveBudget.div(received).toNumber()
-    : undefined;
-  const receiveFraction = inverseRecievedFraction ? 1 / inverseRecievedFraction : 0;
+  const total = spendBudget
+    .add(receiveBudget)
+    .add(spent)
+    .add(received); // this should be constant
+
+  const spendBudgetPercentage = spendBudget.gt(0) ? 100 / total.div(spendBudget).toNumber() : 0;
```
geoknee, line 50: This seems better to me.

geoknee: Note slightly contorted logic around computing percentages, because we use ethers BigNumbers which are always integers. This could probably be factored out into a helper function.

geoknee:
An improved UI around budget would help debug #2003, where a leading hypothesis is that users run out of budget. Every time we open an uploading payment channel, we should see a decrease in our receiving budget. Every time we open a downloading payment channel, we should see a decrease in our spending budget. The decrease is a constant amount,  even if we only end up upload/download 1 byte. On defunding, we should see a partial reversal of those changes.

packages/web3torrent/src/constants.ts
```diff
@@ -177,10 +177,7 @@ export {SINGLE_ASSET_PAYMENT_CONTRACT_ADDRESS};
 
 export const FUNDING_STRATEGY = process.env.FUNDING_STRATEGY === 'Direct' ? 'Direct' : 'Virtual';
 
-export const INITIAL_BUDGET_AMOUNT = utils.hexZeroPad(
-  utils.parseEther('0.000001').toHexString(),
-  32
-); // 1 Szabo (equivalent to 1000GB at 1 wei per byte)
+export const INITIAL_BUDGET_AMOUNT = utils.hexZeroPad(utils.parseEther('0.001').toHexString(), 32); //  (equivalent to 1GB at 1 wei per byte)
```
lalexgap, line 180: It looks like this larger budget is making the `receivedPercentage`/`spendPercentage` too small to actually show up on the pie chart.

lalexgap: Even at 1 szabo the percentage of funds locked up in a channel starts off really small which seems not to render in the pie chart.

geoknee: Yeah I set it to 1GB total initial budget (1000 times lower than it currently is), and was trying uploading a file of 100MB or so in an attempt to get visual confirmation of the data flow. But even then I don't see what I expect...

lalexgap: > Yeah I set it to 1GB total initial budget (1000 times lower than it currently is), and was trying uploading a file of 100MB or so in an attempt to get visual confirmation of the data flow. But even then I don't see what I expect...

I think the change to `INITIAL_BUDGET_AMOUNT` might be incorrect here. `0.001 Eth` is 1000 szabo or 1000 times larger than the previous budget. If I set it to a `gwei` (`0.000000001 Eth`) I see the locked funds on the pie chart fairly quickly. (Unfortunately I'm running into some issues right now where I get stuck on connecting or I would show a screenshot :) )

lalexgap: Using 
```
export const INITIAL_BUDGET_AMOUNT = utils.hexZeroPad(
  utils.parseEther('0.0000000001').toHexString(),
  32
); //  100 Mwei (or 100 mb) // TODO: Change this back
```
I see the pie chart look like I would expect (3.2Mwei would 3.2% of the budget)
image

lalexgap:
I'm wondering if a pie chart will the easiest way to communicate budget info. Right now with our default budget of 1 szabo the earned/spent slices will be incredibly small to start and I'm guessing for a lot of users will remain that way.  Considering that a 100 MB would take 100Mwei which would only be `0.01%` of the 1 szabo budget, we'll often have to deal with tiny little slices on the pie chart.

 We can explode the pie chart and add percentage labels but I think it might get a little visually busy doing that (and make mousing over sections of the chart more difficult).Although design is not my strong suit so perhaps it wouldn't be too bad!

packages/web3torrent/src/constants.ts
geoknee: Ah yes, it would have helped if I hadn't got confused between multiplication and division 😊

geoknee:
My view on this, is that this component is important enough that we should tune the maximum file size and initial budget so that users can easily what is going on. With the temporary budget things look good, IMO:

Screenshot 2020-06-04 at 14 37 06

geoknee:
I put the identity blockie inside the pie chart 😮 

Screenshot 2020-06-04 at 15 05 58

snario:
Just as a comment, this thing without labels is quite confusing.

I still kind of think it might be a good idea to move the entire budget UI into the "your budget" section so as a user you can avoid conflating the channels and torrenting concepts with the budget concepts.

packages/web3torrent/src/hooks/use-budget.ts
```diff
@@ -16,6 +16,7 @@ export function useBudget({ready}: {ready: boolean}) {
       setLoading(false);
     };
     if (ready) getAndSetBudget();
+    return paymentChannelClient.channelClient.onBudgetUpdated(getAndSetBudget);
```
geoknee, line 19: I did this so that components that use this hook will get re-rendered when a `'BudgetUpdated'` event fires. I had to make the `channelClient` a public property to do this.

geoknee: @kerzhner be good to get your thoughts.

geoknee: I think this solves the problem of the budget not always displaying: you might have a budget but your react components don't know about it yet. #2042

geoknee:
@snario :

Screenshot 2020-06-04 at 15 36 44

PR #2038, opened 2020-06-01
Add basic file size limitations

lalexgap:
Ups the budget size and adds a file size check to avoid running into #2003

### DIFF
--- a/packages/web3torrent/src/constants.ts
+++ b/packages/web3torrent/src/constants.ts
@@ -177,9 +177,11 @@ export {SINGLE_ASSET_PAYMENT_CONTRACT_ADDRESS};
 export const FUNDING_STRATEGY = process.env.FUNDING_STRATEGY === 'Direct' ? 'Direct' : 'Virtual';
 
 export const INITIAL_BUDGET_AMOUNT = utils.hexZeroPad(
-  utils.parseEther('0.000000001').toHexString(),
+  utils.parseEther('0.000001').toHexString(),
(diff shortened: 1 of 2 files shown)


packages/web3torrent/src/constants.ts
```diff
@@ -177,9 +177,11 @@ export {SINGLE_ASSET_PAYMENT_CONTRACT_ADDRESS};
 export const FUNDING_STRATEGY = process.env.FUNDING_STRATEGY === 'Direct' ? 'Direct' : 'Virtual';
 
 export const INITIAL_BUDGET_AMOUNT = utils.hexZeroPad(
-  utils.parseEther('0.000000001').toHexString(),
+  utils.parseEther('0.000001').toHexString(),
   32
-); // 1 Gwei (equivalant to 1GB at 1 wei per byte)
+); // 1 Szabo (equivalent to 1000GB at 1 wei per byte)
```
geoknee, line 182: Might be worth expanding this comment to explain that the budget should include a multiplicative overhead to cover the *number* of payment channels we would want to allow. 

I.e. the budget consumed is (number of payment channels) * (funding for each payment channel) = (number of payment channels) * (size of torrent in bytes * cost of 1 byte in wei). 

So 1 Szabo is equivalent to 1 channel downloading 1000GB at 1 wei per byte, or to 1000 channels downloading 1GB (the maximum filesize now) at 1 wei per byte.

PR #2044, opened 2020-06-01
[web3torrent] Fix typo in drift env var

### LINKED ISSUE #2043: [web3torrent] No drift button
I've not seen this appearing recently.

https://github.com/statechannels/monorepo/blob/921a185ebfe72259e71fd29835a51a70f3251e57/packages/web3torrent/src/index.tsx#L22

geoknee:
Closes #2043

### DIFF
--- a/packages/web3torrent/config/webpack.config.js
+++ b/packages/web3torrent/config/webpack.config.js
@@ -75,7 +75,7 @@ module.exports = function(webpackEnv) {
         [
           'VERSION',
           'CHAIN_NETWORK_ID',
-          'DRIFT_CHATBOX_ID',
+          'DRIFT_CHATBOX_APP_ID',
           'FAKE_CHANNEL_PROVIDER',
           'FIREBASE_API_KEY',
           'FIREBASE_PREFIX',


packages/web3torrent/config/webpack.config.js
```diff
@@ -75,7 +75,7 @@ module.exports = function(webpackEnv) {
         [
           'VERSION',
           'CHAIN_NETWORK_ID',
-          'DRIFT_CHATBOX_ID',
+          'DRIFT_CHATBOX_APP_ID',
```
geoknee, line 78: Do we still want to whitelist environment variables? As is we need to add a new env var in two places (`.env.something` as well as here). Easy to miss this, and it doesn't match the pattern used in our other apps.

snario: I don't think it's that useful to be whitelisting them

geoknee: https://github.com/statechannels/monorepo/pull/2045

geoknee:
In favour of #2045

Issue #2057, opened 2020-06-02
Reduce logging to sane levels


kerzhner:

andrewgordstewart:
A quick win would be to `log.trace` the workflow context in wallet transitions

kerzhner:
It would also be nice to keep the logging level at trace in production for now. This way we have as much debugging information as possible for digging into issues.

andrewgordstewart:
> It would also be nice to keep the logging level at trace in production for now. This way we have as much debugging information as possible for digging into issues.

We can
- save `trace` logs in the LogBlob (so it's downloadable)
- don't print `trace` logs to the console (so it doesn't overwhelm the console logs)

that doesn't save any memory, and doesn't lose any logs, just cleans up visual clutter.

andrewgordstewart:
Another quick win: https://github.com/statechannels/monorepo/issues/2041#issuecomment-637737955

andrewgordstewart:
I took a peek through https://32582-209829093-gh.circle-artifacts.com/0/home/circleci/project/logs/multiple-leecher.A.7.pino.log, which serializes a bunch of the web3torrent stuff away. 

Sorting the biggest file (~2MB) by line length, grabbing the first ~10% (150) lines, and reading the log message:
```
❯ cat multiple-leecher.A.7.pino.log | awk '{ print length(), $0 | "sort -n" }' | tail -n 150 | sed 's/^[0-9]* //' | jq .msg  
"<< unblockPeer: start"
"<< blockPeer"
"<< unblockPeer: finish"
"<< blockPeer"
"<< unblockPeer: finish"
"<< blockPeer"
"<< blockPeer"
"<< unblockPeer: start"
"<< unblockPeer: start"
"<< unblockPeer: finish"
"<< blockPeer"
"<< blockPeer"
"<< blockPeer"
"<< unblockPeer: finish"
"<< unblockPeer: finish"
"<< blockPeer"
...
```

So, to trim the logs size, we'll need to trim those log statements...

That being said ... why are peers blocked and unblocked so frequently? Does that affect torrent performance?

andrewgordstewart:
> So, to trim the logs size, we'll need to trim those log statements...

For the record:
```
~/Downloads via ☕ v11.0.2 on ☁️  us-west-1 
❯ cat multiple-leecher.A.7.pino.log | awk '{ print length(), $0 | "sort -n" }' | tail -n 150 > longLines.log 

[...]

PR #2059, opened 2020-06-02
Switch to log.trace where appropriate

lalexgap:
Related to #2057.

Switchs multiple log.info to log.trace so logging output will be reduced.

Currently the `LOG_LEVEL` is left on `trace` for production but that can easily be flipped when we want to deploy a quieter version.


packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -452,7 +452,7 @@ export class PaymentChannelClient {
       );
     } catch (e) {
       if (e.message === 'User declined') {
-        log.info('User declined budget creation');
+        log.trace('User declined budget creation');
```
andrewgordstewart, line 455: I think it's worth taking some time to think about what should be `info`, `trace`, etc. 

For instance, this is catching an error, and it doesn't happen repeatedly. So I don't think it should be `trace`.

A good rule of thumb seems to be:
A. things that happen repeatedly should tend have log level equal to `debug` or `trace`
B. things we might want to know about, to see what's going on, should have log level equal to `info`, `warn` or `error`

Did you apply some kind of rule of thumb like this?

andrewgordstewart:
I changed the comment, because I don't think we want to close #2057 even with these changes -- we shouldn't have ~100MB log files at the trace level when downloading a ~100MB file

packages/web3torrent/src/clients/payment-channel-client.ts
lalexgap: While this is catching an "error" technically, it's essentially logging out that the user has declined the budget creation by clicking a button.

In my mind this is useful information to have if we're trying to debug what went wrong in some situation. However there doesn't seem to be a huge benefit to logging this out in the normal process, as we're essentially echoing the action the user took in the UI.

My rough rule of thumb is that `info` is used for messages that might be of interest for advanced users and provide insight into what is going on. `trace` should be used for messages that help us trace what was going on when something went wrong. (Although I only did a quick pass through `web3torrent` so the log level might not be right for all of them.

.env.production-ropsten
```diff
@@ -1,7 +1,7 @@
 ## all
 NODE_ENV = 'production'
 LOG_DESTINATION = 'console'
-
+LOG_LEVEL= 'trace' # TODO: This should be set to info before we release
```
andrewgordstewart, line 4: Let's create an issue to track this

PR #2082, opened 2020-06-03
Minor Budget related changes

lalexgap:
Some minor changes around @geoknee's work.

This PR
- restores styling to the withdraw button
- gets the failing test passing
- sets the budget to 100 Mwei (temporarily) to make progress on the pie chart visible.

I've held off on styling the pie-chart  as I think it may be worth discussing first.


geoknee:
Thanks for this @lalexgap . I think there is still an issue where the budget is not decreasing at all when we fund a new channel. I'll keep looking into that.

PR #2091, opened 2020-06-05
[web3torrent] Improve payment counting

geoknee:
This is a little inaccurate as is.


geoknee:
Although this was an improvement, I think it will still get tripped up by state updates that are not payments, such as the concluding update(s).

funds/numPayments seems to give quite varying answer for each row of the table:

Screenshot 2020-06-05 at 18 02 19

PR #2107, opened 2020-06-09
[web3torrent] performance tweaks

kerzhner:
PR description:
- Do not update UI on the most frequent webtorrent events: request, notice, upload, and download.
- Always after `onChannelUpdated`.
- Reduce frames-per-second to at most 10.
- Bump peer trust from 4 to 5.

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -475,8 +475,8 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     torrent.on(TorrentEvents.READY, emitTorrentUpdated(TorrentEvents.READY));
     torrent.on(TorrentEvents.WARNING, emitTorrentUpdated(TorrentEvents.WARNING));
     // These events are too frequent
-    // torrent.on(TorrentEvents.DOWNLOAD, emitTorrentUpdated);
-    // torrent.on(TorrentEvents.UPLOAD, emitTorrentUpdated);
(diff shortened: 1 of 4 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -595,7 +594,8 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
   }
 
   /** Emit an event that triggers a UI re-render */
-  private emitTorrentUpdated(infoHash) {
+  private emitTorrentUpdated(infoHash, trigger) {
+    // log.trace(`emitTorrentUpdate: ${trigger}`);
```
geoknee, line 598: Does this comment belong?

packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -595,7 +594,8 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
   }
 
   /** Emit an event that triggers a UI re-render */
-  private emitTorrentUpdated(infoHash) {
+  private emitTorrentUpdated(infoHash, trigger) {
```
geoknee, line 597: Should `trigger` have a type annotation?

packages/web3torrent/src/library/web3torrent-lib.ts
kerzhner: I would be in favor of leaving this line in here. The intention is to make it easier to "profile" why rerendering occurs.

packages/web3torrent/src/library/web3torrent-lib.ts
kerzhner: Good call, resolved in 3e9eb20d1.

PR #2112, opened 2020-06-09
[nitro-protocol] Trying to upgrade ethers

geoknee:
This is quite a slog, so just putting this here to deter anyone from replicating...

### DIFF
--- a/packages/nitro-protocol/package.json
+++ b/packages/nitro-protocol/package.json
@@ -6,6 +6,7 @@
   "bugs": "https://github.com/statechannels/monorepo/issues",
   "dependencies": {
     "@counterfactual/cf-adjudicator-contracts": "0.0.11",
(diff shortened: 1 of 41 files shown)


kerzhner:
Thanks for tackling this! Let's make sure to remove [this nohoist](https://github.com/statechannels/monorepo/blob/master/package.json#L67-L69) configuration. The benefits are:
- One version of `ethers` in the monorepo root as opposed to one version per package.
- With `ethers` v5, some of the functionality has been moved to `@ethersproject/...` modules. nohoist for `ethers` but not for `@ethersproject/...` modules can result in unexpected behavior.

geoknee:
There's still quite a lot of work to do on this branch to get the tests passing.

geoknee:
In favour of #2542

PR #2114, opened 2020-06-09
[nitro-protocol] Tweak tsconfig.json and package.ison

geoknee:


packages/nitro-protocol/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@statechannels/nitro-protocol",
   "description": "A protocol for state channel networks",
   "version": "0.1.0",
-  "author": "George Knee",
+  "author": "statechannels",
```
snario, line 5: I think it's better to list specific people. It can be an array.

packages/nitro-protocol/package.json
```diff
@@ -71,6 +72,7 @@
     "{src,test}/**/*.{ts,tsx}": "eslint"
   },
   "main": "lib/src/index.js",
+  "private": false,
```
snario, line 75: Why private?

geoknee: !private

packages/nitro-protocol/package.json
geoknee: This matches our other published package (devtools). It's an open source project so I think a list of authors would be a bit of a pain to maintain.

packages/nitro-protocol/package.json
snario: Oh I see. Seems odd to _add_ this key. Usually it's just omitted

PR #2118, opened 2020-06-09
show warning if balance is low

### LINKED ISSUE #2116: Check eth balance before starting deposit process

lalexgap:
Shows a basic warning if your balance is less than the initial budget amount with a link to a faucet.
Fixes #2116 
image

### DIFF
--- a/packages/web3torrent/src/App.tsx
+++ b/packages/web3torrent/src/App.tsx
@@ -5,14 +5,16 @@ import {Route, Switch, BrowserRouter} from 'react-router-dom';
 import './App.scss';
 import {LayoutFooter, LayoutHeader} from './components/layout';
 import Welcome from './pages/welcome/Welcome';
-
+import {Flash, Link} from 'rimble-ui';
 import File from './pages/file/File';
 import Upload from './pages/upload/Upload';
 import {RoutePath} from './routes';
-import {requiredNetwork} from './constants';
+import {requiredNetwork, INITIAL_BUDGET_AMOUNT} from './constants';
 import {Budgets} from './pages/budgets/Budgets';
 import {web3TorrentClient} from './clients/web3torrent-client';
 import {identify} from './analytics';
+import {providers} from 'ethers';
+import {utils} from 'ethers';
 
 const App: React.FC = () => {
   const [currentNetwork, setCurrentNetwork] = useState(
@@ -20,6 +22,7 @@ const App: React.FC = () => {
   );
 
   const [initialized, setInitialized] = useState(false);
+
   useEffect(() => {
     web3TorrentClient.paymentChannelClient.initialize().then(() => {
       setInitialized(true);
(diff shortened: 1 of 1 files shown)


packages/web3torrent/src/App.tsx
```diff
@@ -55,6 +83,16 @@ const App: React.FC = () => {
           requiredNetwork={requiredNetwork}
           onWeb3Fallback={!('ethereum' in window)}
         />
+        {ready && balance && balance.lt(INITIAL_BUDGET_AMOUNT) && (
+          <Flash my={3} variant="warning">
+            You currently have {utils.formatEther(balance)} ETH in your wallet. You'll need at least{' '}
+            {utils.formatEther(INITIAL_BUDGET_AMOUNT)} ETH in your wallet to fund Web3Torrent. You
```
kerzhner, line 89: Not sure I am a fan of how many 0s this outputs for small amounts

packages/web3torrent/src/App.tsx
```diff
@@ -33,20 +36,45 @@ const App: React.FC = () => {
...
+    }
+  }, []);
 
   useEffect(() => {
     if (window.ethereum && typeof window.ethereum.on === 'function') {
-      const listener = chainId => setCurrentNetwork(Number(chainId));
-      window.ethereum.on('networkChanged', listener);
+      const networkChangeListener = chainId => setCurrentNetwork(Number(chainId));
+      const addressChangeListener = accounts => {
+        console.log(accounts[0]);
```
kerzhner, line 50: Should we use the pino logger?

packages/web3torrent/src/App.tsx
```diff
@@ -33,20 +36,45 @@ const App: React.FC = () => {
       }
     });
   }, [initialized]);
+  const [selectedAddress, setSelectedAddress] = useState(undefined);
+  useEffect(() => {
```
kerzhner, line 40: Any particular reason to have 2 useEffect hooks that run after every render (as opposed to merging them into 1)?

kerzhner: Also, would you mind explaining why we need `useEffect` here? My understanding of what this does is:
1. The App component renders.
1. After the render this hook is run.
1. If `window.ethereum.selectedAddress` has changed, the App component is rerendered.

What about just using the `window.ethereum.selectedAddress` during the first render?

packages/web3torrent/src/App.tsx
```diff
@@ -33,20 +36,45 @@ const App: React.FC = () => {
...
+    const getBalance = async () => {
+      if (selectedAddress) {
+        const provider = new providers.Web3Provider(window.ethereum);
+        const balance = await provider.getBalance(selectedAddress);
+        setBalance(balance);
+      }
+    };
+
+    getBalance();
+  }, [ready, selectedAddress]);
```
kerzhner, line 76: Should `ready` get read in the hook?

packages/web3torrent/src/App.tsx
lalexgap: Ah this should be removed, I just added it for some quick debugging.

packages/web3torrent/src/App.tsx
lalexgap: Without `useEffect` we run into an infinite render loop.  I think this is due to `setSelectedAddress` being called, which results in the hook below calling `setBalance` which forces a re-render. By using an empty dependency array we limit this hook to running once after the first render.

I don't think we can rely on `window.ethereum.selectedAddress` changing to force a re-render, as I don't think it's a valid dependency for a hook. (It can be mutated, which wouldn't trigger the hook)

So we rely on attaching the listener that updates a piece of state indicating the address has changed. We could change that so it's just a flag for the hook that the address has changed and and just use `window.ethereum.selectedAddress` in the hook but that seems more convoluted that just storing the address and depending on that. That would allow us to avoid the initial hook though.

packages/web3torrent/src/App.tsx
lalexgap: I was thinking about that too. However two reasons I stuck with it.
- The approve and fund workflow displays the same thing when showing the budget.
- I think more people know `ETH than `Mwei` or other denominations of eth.

packages/web3torrent/src/App.tsx
lalexgap: > Any particular reason to have 2 useEffect hooks that run after every render (as opposed to merging them into 1)?

This is mostly to keep the logic organized. One hook is setting up listeners and one hook is setting the current address.

kerzhner: Thanks for the explanation, that makes a lot of sense! What about the following code to remove the hook
```
  const [selectedAddressState, setSelectedAddress] = useState(undefined);
  const selectedAddress = selectedAddressState ?? window.ethereum?.selectedAddress;
```

While I do like React hooks better than the class alternative, I think my preference is to use hooks as sparingly as possible since they are not super easy to reason about.

PR #2121, opened 2020-06-10
Add some error messages for running out of budget

### LINKED ISSUE #2054: Display warning when budget is low
Figure out mimumum thing that needs to be done to prevent blocking when you're budget runs out.

lalexgap:
Adds some errors when running out of budget for uploading or downloading.

Fixes #2054

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -535,7 +535,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
 
   /** Utility function. Checks if the budget has enough receive capacity to accept a new channel. Pauses the torrent if it does not. **/
   private checkUploadBudget(budget, torrent) {
-    if (budget?.budgets?.length === 1) {
+    if (budget?.budgets?.length) {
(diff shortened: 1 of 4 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -525,6 +533,20 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
+  /** Utility function. Checks if the budget has enough receive capacity to accept a new channel. Pauses the torrent if it does not. **/
+  private checkUploadBudget(budget, torrent) {
+    if (budget?.budgets?.length === 1) {
+      const receiveCapacity = bigNumberify(budget.budgets[0].availableReceiveCapacity);
+      const uploadSize = WEI_PER_BYTE.mul(torrent.length).toHexString();
+      if (receiveCapacity.lt(uploadSize)) {
+        log.warn(
+          `Receive capacity (${receiveCapacity.toHexString()}) is less than file cost (${uploadSize}). Pausing torrent.`
+        );
+        torrent.pause();
```
lalexgap, line 545: I'm not sure if this does what I want it to. I could `destroy` the torrent but that would cause existing connections to fail.

PR #2136, opened 2020-06-10
Add balance check to Approve Budget and Fund

### LINKED ISSUE #2116: Check eth balance before starting deposit process

lalexgap:
Fixes #2116 

Adds a balance check to the approve budget and fund workflow and removes it from the application.

The workflow waits until the user deposits enough funds before proceeding.

image

### DIFF
--- a/packages/xstate-wallet/src/chain.ts
+++ b/packages/xstate-wallet/src/chain.ts
@@ -9,7 +9,7 @@ import {
 import {Contract, Wallet, BigNumber, BigNumberish} from 'ethers';
 
 import {Observable, fromEvent, from, merge, interval} from 'rxjs';
-import {filter, map, flatMap} from 'rxjs/operators';
+import {filter, map, flatMap, distinctUntilChanged} from 'rxjs/operators';
 import {One, Zero} from '@ethersproject/constants';
 import {hexZeroPad} from '@ethersproject/bytes';
 import {TransactionRequest} from '@ethersproject/providers';
@@ -52,6 +52,7 @@ export interface Chain {
   challenge: (support: SignedState[], privateKey: string) => Promise<string | undefined>;
   finalizeAndWithdraw: (finalizationProof: SignedState[]) => Promise<string | undefined>;
   getChainInfo: (channelId: string) => Promise<ChannelChainInfo>;
+  balanceUpdatedFeed(address: string): Observable<BigNumber>;
 }
 
 type Updated = ChannelChainInfo & {channelId: string};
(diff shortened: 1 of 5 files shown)


packages/xstate-wallet/src/chain.ts
```diff
@@ -199,11 +200,18 @@ export class FakeChain implements Chain {
 
     return merge(first, updates);
   }
-
+  public balanceUpdatedFeed(): Observable<BigNumber> {
```
kerzhner, line 203: My understanding of having two implementations of the `balanceUpdatedFeed` is that the address parameter is optional. Is it cleaner to have one implementation with an optional parameter?
```suggestion
  public balanceUpdatedFeed(address?: string): Observable<BigNumber> {
```

packages/xstate-wallet/src/ui/approve-budget-and-fund-workflow.tsx
```diff
@@ -29,6 +29,31 @@ export const ApproveBudgetAndFund = (props: Props) => {
     };
   }, [current]);
 
+  const waitForSufficientFundsInit = (
+    <Flex alignItems="center" flexDirection="column">
+      <Heading>App Budget</Heading>
+
+      <Text textAlign="center">Querying blockchain</Text>
```
kerzhner, line 36:
```suggestion
      <Text textAlign="center">Checking if your Metamask account has sufficient ETH</Text>
```

packages/xstate-wallet/src/ui/approve-budget-and-fund-workflow.tsx
```diff
@@ -29,6 +29,31 @@ export const ApproveBudgetAndFund = (props: Props) => {
...
+  const waitForSufficientFunds = (
+    <Flex alignItems="left" flexDirection="column">
+      <Heading textAlign="center" mb={2}>
+        App Budget
+      </Heading>
+      <Text pb={3} fontSize={1}>
+        You don&#39;t have enough ETH in your wallet!
+      </Text>
+      <Text pb={3} fontSize={1}>
+        You&#39;ll need at least {utils.formatEther(playerAmount)} ETH in your wallet to fund the
```
kerzhner, line 49:
```suggestion
        You&#39;ll need at least {utils.formatEther(playerAmount)} ETH in your Metamask wallet to fund the
```

packages/xstate-wallet/src/chain.ts
lalexgap: There are actually two classes in here `FakeChain` and `ChainWatcher` which both implement the `Chain` interface. So I needed to add the method to both classes.

Outside the scope of this PR but we should probably move `FakeChain` out of `chain.ts`

kerzhner: Ah that makes sense!

PR #2138, opened 2020-06-10
Fix some logging issues

andrewgordstewart:
1. When I tried to set the log level to `trace`, it created an infinite loop:

Screen Shot 2020-06-10 at 4 38 06 PM

Most of the non-Json-RPC message is a `PINO_LOG` message containing the error `"Received a non JsonRpc message"` :P

2. @kerzhner noticed that chrome expands trace calls. Firefox also seems to. So, trace log messages are redirected to `console.debug` instead.

3. `LOG_LEVEL` was not getting through the wallet's webpack config.

Fixes https://github.com/statechannels/monorepo/issues/2134

### DIFF
--- a/packages/channel-provider/src/messaging-service.ts
+++ b/packages/channel-provider/src/messaging-service.ts
@@ -109,8 +109,6 @@ export class MessagingService {
         } else if (isJsonRpcErrorResponse(event.data)) {
           reject(new RpcError(event.data.error));
         }
-      } else {
-        logger.error({event}, 'Received a non JsonRpc response');
       }
     };
 
--- a/packages/channel-provider/src/logger.ts
+++ b/packages/channel-provider/src/logger.ts
@@ -30,7 +30,9 @@ const browser: any = IS_BROWSER_CONTEXT
         warn: postMessageAndCallToConsoleFn(console.warn),
         info: postMessageAndCallToConsoleFn(console.info),
         debug: postMessageAndCallToConsoleFn(console.debug),
-        trace: postMessageAndCallToConsoleFn(console.trace)
+        // Firefox & chrome automatically expand trace calls, which is pretty annoying.
+        // So, we direct trace calls to console.debug instead.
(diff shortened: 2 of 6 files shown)


packages/channel-provider/src/messaging-service.ts
```diff
@@ -109,8 +109,6 @@ export class MessagingService {
         } else if (isJsonRpcErrorResponse(event.data)) {
           reject(new RpcError(event.data.error));
         }
-      } else {
-        logger.error({event}, 'Received a non JsonRpc response');
```
kerzhner, line 113: I am in favor of somehow noting/logging that the event cannot be interpreted so that we don't silently drop events. Maybe not possible given the infinite loop problem at the moment.

andrewgordstewart, line 113: The infinite loop here is:
- when the channel provider triggers a log event, it calls `postMessage` to make web3torrent aware of that log event
- but L113 here will read _that_ message, containing the log event, see that it's not a JsonRpc message, and log an error
- logging that error calls `postMessage` ...

## Alternative 1
1. web3torrent listens for log events only from the channel provider
2. wallet logger calls postMessage with a notification that there was a log event
3. channel-provider reads that event, forwards it to web3torrent

## Alternative 2
0. Add a `LogEvent` notification type to the client-wallet-API
1. channel-provider stores the log events and exposes the `saveLogs` function on the window
   a. channel-provider can expose a window function `addToLogs`, or listen to some event from the app, about new log events
2. web3torrent does whatever's done in `a` when it gets a log event
3. channel-provider stores a log event when it receives a `LogEvent` JsonRpc notification

I'm starting to lean towards 2, though I'd prefer to do either of these in a separate PR (and maybe not urgently?)

andrewgordstewart: https://github.com/statechannels/monorepo/pull/2138#discussion_r438973814

PR #2139, opened 2020-06-10
[web3torrent] Channel list UI

kerzhner:
Fixes https://github.com/statechannels/monorepo/issues/1949.

The goal of the PR is to display the most relevant set of channels for the current torrent. We create an in-memory map of channel id to torrent hash that is updated when a channel is created, joined, or updated. The resulting behavior is:
- After page refresh, no channels are displayed from previous leeching or seeding.
- Only channels for the current file shared are displayed.

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -164,6 +164,7 @@ export type ChannelCache = Record<string, ChannelState | undefined>;
 export class PaymentChannelClient {
   channelCache: ChannelCache = {};
   budgetCache?: DomainBudget;
+  channelIdToTorrentMap: Record<string, string> = {};
 
   get mySigningAddress(): string | undefined {
     return this.channelClient.signingAddress;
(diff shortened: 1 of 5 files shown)


packages/web3torrent/src/clients/payment-channel-client.ts
```diff
@@ -240,7 +243,7 @@ export class PaymentChannelClient {
     }
   }
 
-  async createChannel(peers: Peers): Promise<ChannelState> {
+  async createChannel(peers: Peers, torrentHash: string): Promise<ChannelState> {
```
geoknee, line 246: One downside to this approach is that the payment-channel-client now contains torrent specific concepts. How hard would it be to store this mapping in the web3torrent client?

geoknee: Or, do you think this approach would work? https://github.com/statechannels/monorepo/pull/2161

kerzhner: Totally agreed that this solution is not great. I am open to suggestions on where to store this mapping.

kerzhner: The channel id to torrent info hash mapping has been moved to web3torrent library: 426d77b49.

PR #2140, opened 2020-06-11
Reload page on address change

lalexgap:
Reloads web3torrent when the address changes.

Since the  `accountsChanged` event will only get fired from the wallet iFrame (due to `window.ethereum.enable` only being called from the wallet iframe) the wallet has to be the one responsible for detecting the account change event.

Due to the fact that our iframe has a different domain than the main window, web3torrent needs to be responsible for reloading the whole page.

To accomplish this the wallet now sends a reload message to the main frame where the channel provider handles it and reloads the page.


snario:
Why does the page need to be reloaded?

geoknee:
I believe we decided to close this in favour of other approaches that got merged in.

PR #2143, opened 2020-06-11
Change destination address to be signingAddress

### LINKED ISSUE #2128: Switching address and returning to the app does not work
If you
- Use web3torrent with one address
- Close web3torrent
- Switch address
- navigate to web3torrent and try to use it

it doesn't seem to work

snario:
Fixes #2128.

This makes it so that Web3Torrent provides the _signingAddress_ of its peer and wallet as the destination when creating a channel. Similarly, the wallet uses the signing address in the `ApproveBudgetAndFund` call which then sets up the ledger channel as appropriate. This small changes allows the user of Web3Torrent to use the app, close tabs and switch accounts, and return to the app and use it again.

When withdrawing, the selected address is actually used, so withdrawing works as expected. This change only really becomes slightly confusing if you _challenge_ a channel, but even then it is not that confusing.

It's worth considering whether or not it makes sense for the app to even be providing a destination value to the wallet.

### DIFF
--- a/packages/web3torrent/src/clients/payment-channel-client.ts
+++ b/packages/web3torrent/src/clients/payment-channel-client.ts
@@ -114,16 +114,16 @@ const arrangePeers = ({beneficiary, payer}: Peers): [Peer, Peer] => {
   return peers;
 };
 
-const formatParticipant = ({signingAddress, outcomeAddress}: Peer): Participant => ({
+const formatParticipant = ({signingAddress /*, outcomeAddress */}: Peer): Participant => ({
   participantId: signingAddress,
   signingAddress,
-  destination: outcomeAddress
+  destination: signingAddress // Using signingAddress for destination allows user to switch MetaMask accounts
 });
 const formatParticipants = (peers: Peers) => arrangePeers(peers).map(formatParticipant);
 
 const formatItem = (p: Peer): AllocationItem => ({
   amount: hexZeroPad(bigNumberify(p.balance).toHexString(), 32),
-  destination: p.outcomeAddress
+  destination: p.signingAddress // Using signingAddress for destination allows user to switch MetaMask accounts
 });
 const formatAllocations = (peers: Peers): Allocations => [
   {token: AddressZero, allocationItems: arrangePeers(peers).map(formatItem)}
--- a/packages/xstate-wallet/src/messaging.ts
+++ b/packages/xstate-wallet/src/messaging.ts
@@ -263,7 +263,7 @@ async function convertToInternalEvent(
     case 'ApproveBudgetAndFund':
       const {hub, playerParticipantId} = request.params;
       const signingAddress = await store.getAddress();
-      const destination = store.chain.selectedAddress;
+      const destination = signingAddress; // Using signingAddress for destination allows user to switch MetaMask accounts
       if (!destination) {
         throw new Error('No selected destination');
       }


geoknee:
Unless I have misunderstood, this would mean that when you withdraw from the ledger channel, the funds get paid out to an account that is not known to metamask, and getting control of those funds requires digging into IndexedDB for a private key? 

> When withdrawing, the selected address is actually used, so withdrawing works as expected. This change only really becomes slightly confusing if you challenge a channel, but even then it is not that confusing.

Do you mean that we can basically overwrite the outcome with the one that we really want, as long as we are all collaborating? 

I'm not sure that it's worth doing this just to support the use switching accounts in metamask. Won't #2140 mostly fix that problem, anyway?

snario:
> Unless I have misunderstood, this would mean that when you withdraw from the ledger channel, the funds get paid out to an account that is not known to metamask, and getting control of those funds requires digging into IndexedDB for a private key?

Actually I take back what I said earlier. I thought what was already implemented correctly was that, regardless of ledger channel outcomes, when you call `CloseAndWithdraw` the account selected in MetaMask would be the recipient of the funds. It looks like it _should_ be implemented that way based on this code:

https://github.com/statechannels/monorepo/blob/49b9585cd7b981e9c21fd36994e862539a0fdfa2/packages/xstate-wallet/src/messaging.ts#L252-L262

> Do you mean that we can basically overwrite the outcome with the one that we really want, as long as we are all collaborating?

This is how I thought it worked (see above)

> I'm not sure that it's worth doing this just to support the use switching accounts in metamask. Won't #2140 mostly fix that problem, anyway?

#2140 does not fix the problem of switching accounts outside of the tab and then returning AFAICT because your ledger channel's destination address is of the previous session.

There is a check in the wallet which ensures equality of the target channel destination (provided by the app) and the ledger channel destination:

https://github.com/statechannels/monorepo/blob/49b9585cd7b981e9c21fd36994e862539a0fdfa2/packages/xstate-wallet/src/utils/outcome.ts#L71-L72

snario:
Also the fact that the tests passed gave me the impression that the currently selected MetaMask address would be getting the funds. However maybe instead it's just an indication that this is not tested. I haven't checked just yet.

geoknee:
Yeah I think you have touched on some stuff that we should fix around concluding and withdrawing: 

- We should get the necessary data to conclude the ledger channel from the latest state we have for that channel (not `window.ethereum.selectedAddress`, not even `store.chain.selectedAddress`)
- We should test that funds are returned to whichever account was selected in metamask at the point when ApproveBudgetAndFund was called.

(In a previous iteration of the contracts, IIRC you could specify a new address when withdrawing. That is not currently supported, though).

snario:
We discussed this on standup and the plan is to:

- Rename `selectedAddress` to `idealOutcomeAddress` or something like that which conveys that this is not a selected address but rather an address that, if ETH were sent there, would be easily accessible by the user. It will be set as the selected address in MetaMask the first time any app ever calls `EnableEthereum`.

- Store `idealOutcomeAddress` in IndexedDB and re-use that as the destination address for all channels in the future instead of checking `ethereum.selectedAddress` each time.

- Remove `destination` from being a property that can be provided by the application in the call to `CreateChannel`

andrewgordstewart:
Suggestion: rename `selectedAddress` to `destinationAddress`?

PR #2157, opened 2020-06-12
[ web3torrent ] Set strict: true in tsconfig

geoknee:
This enables `--strictNullChecks` and some other things that I think will be very useful for catching type errors before they hit runtime.


geoknee:
It is also quite concerning that we don't seem to have circle properly checking for typescript errors in this package: right now it is only the `test` job that is picking them up (and I think not all of them).

geoknee:
This is a Pandora's box that I wish I hadn't opened...

snario:
@geoknee this should be moved to `apps` or closed

PR #2159, opened 2020-06-12
Introduce destinationAddress concept

### LINKED ISSUE #2128: Switching address and returning to the app does not work

snario:
Fixes #2128.

Based on https://github.com/statechannels/monorepo/pull/2143#issuecomment-642811267.

> We discussed this on standup and the plan is to:
> 
> * Rename `selectedAddress` to `idealOutcomeAddress` or something like that which conveys that this is not a selected address but rather an address that, if ETH were sent there, would be easily accessible by the user. It will be set as the selected address in MetaMask the first time any app ever calls `EnableEthereum`.
> * Store `idealOutcomeAddress` in IndexedDB and re-use that as the destination address for all channels in the future instead of checking `ethereum.selectedAddress` each time.
> * Remove `destination` from being a property that can be provided by the application in the call to `CreateChannel`

### DIFF
--- a/packages/channel-client/src/channel-client.ts
+++ b/packages/channel-client/src/channel-client.ts
@@ -26,8 +26,8 @@ export class ChannelClient implements ChannelClientInterface {
     return this.provider.signingAddress;
   }
 
-  get selectedAddress(): string | undefined {
-    return this.provider.selectedAddress;
+  get destinationAddress(): string | undefined {
+    return this.provider.destinationAddress;
   }
 
   get walletVersion(): string | undefined {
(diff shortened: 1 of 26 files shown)


snario:
> to change my destinationAddress, I can do a factory reset (or otherwise delete my IndexedDB), and then go through EthereumEnable workflow again? I suppose long term we ought to support changing the destinationAddress in a less destructive manner.

Yes

> The destinationAddress is still provided in the createChannel API call (inside the allocations field).

This is a good point... I didn't notice it to be honest, I need to investigate the implications

snario:
Just reviewed and it seems like it's fine. Both parties just remember what their destination address is.

PR #2161, opened 2020-06-12
[web3torrent] (alternative) Filter channels list by current torrent

geoknee:


packages/web3torrent/src/components/torrent-info/channels-list/ChannelsList.tsx
```diff
@@ -131,13 +132,22 @@ function channelIdToTableRow(
...
   const statuses = ['running', 'closing', 'proposing', 'closed'];
-
+  const web3TorrentClient = useContext(Web3TorrentClientContext);
   const channelsInfo = _.keys(channels)
     .filter(
       id =>
         channels[id].payer.signingAddress === mySigningAddress ||
         channels[id].beneficiary.signingAddress === mySigningAddress
     )
+    .filter(id =>
```
kerzhner, line 142: What if a wire for a torrent is destroyed? Does this work if the wire has completed? I was worried about these scenarios when tying wires to list of channels displayed.

Also, the following might be a bt contrived, but it will not work in the cases of:
1. The user is still on the torrent page, but the torrent has been destroyed for some reason.
1. The user uploads torrent A, then uploads torrent B, then uploads torrent A. For the third upload, the channels from the first upload would not be shown.

That being said, I am totally fine with this solution.

geoknee: Yeah these are good points. One option might be to keep channels that we can't find a wire for, and instead simply  filter _out_ channels that we *know* are for other torrents?

We could even put channels with no wires in a separate table.

kerzhner: I like that idea for certain channels! But, try out this branch with a one leecher, one seeder test. The channel disappears after exchange is completed. I think putting that channel into an unknowns table might be poor UI.

geoknee:
Agreed that we need a mapping between channels and torrents that survives wires (the only current link between them) being dropped.

PR #2170, opened 2020-06-14
Improve E2E tests

andrewgordstewart:
Resolves https://github.com/statechannels/monorepo/issues/2135

Generally switches to the philosophy that:
- We should detect events that indicate that the app/torrent has made "progress"
- We should then make expectations about the UI once that progress has been made.

The multiple-leecher test also has `C` cancel a partially completed download. This did cause a [flicker](https://github.com/statechannels/monorepo/issues/2169) locally.

In addition, one really useful has been added: the page's _content_ (and also a screenshot) is now stored as an [artifact](https://34456-209829093-gh.circle-artifacts.com/0/home/circleci/project/packages/e2e-tests/screenshots/multiple-leecher.A.html). In case a test fails due to an HTML element, you can inspect elements yourself. Given this, the screenshot is not very useful.

I ran the test half a dozen times on https://github.com/statechannels/monorepo/pull/2170/commits/05d81839a8809f386ac6cbd34c426e4a67ba561d with no flickering. Hopefully it stays consistent!


andrewgordstewart:
... six times was enough to make it flicker :/

PR #2171, opened 2020-06-15
[nitro-protocol] Overhaul documentation website

geoknee:
- Minor styling changes
- Splits the _Docs_ into 3 parts targeting a) Dapp devs, b) Wallet devs, c) Contract devs (most existing material moved to c)
- Adds an in-depth tutorial for wallet devs, accompanied by a new repository [`nitro-tutorial`](https://github.com/statechannels/nitro-tutorial)


geoknee:
Current site: https://protocol.statechannels.org/.
Deploy preview for latest commit on this branch: https://5ee746469840ba3fb65ab2f1--nitro-protocol.netlify.app

snario:
@geoknee I agree this is a drastic improvement.

One recommendation I have is adding a subsection under the "Execute state transitions off chain" section that touches on the question of duplicating writing application state transition logic (i.e., once in Solidity, once in the application-native language).

This was the motivation for what is now called the "CounterfactualAdapterApp", but the basic idea is that if the application has an `applyAction(State, Action) -> State` interface, then you can do an `eth_call` to some node, or run EVM code locally, given the prior state and a chosen action to compute the next state. This way the application could avoid needing to be responsible for constructing a new state and literally run the exact code that would be run on-chain if there were an on-chain challenge required.

PR #2174, opened 2020-06-15
Remove mySelectedAddress on ChainWatcher

snario:
Sometimes when I load the Web3Torrent app I saw that I would go through the "EnableEthereumWorkflow" again, despite `window.ethereum.selectedAddress` being defined. I manually attempted to load https://xstate-wallet.statechannels.org and, really quickly, inspect `window.ethereum.selectedAddress` which _sometimes_ was `undefined` if I called it fast enough. I suspect what is happening is that there is a race condition between the initialization of `ChainWatcher` and MetaMask injecting `window.ethereum` into the app. This change should fix it.

### DIFF
--- a/packages/xstate-wallet/src/chain.ts
+++ b/packages/xstate-wallet/src/chain.ts
@@ -321,6 +321,9 @@ export class ChainWatcher implements Chain {
   }
 
   public get selectedAddress(): string | null {
+    if (this.mySelectedAddress === null && window && window.ethereum) {
+      this.mySelectedAddress = window.ethereum.selectedAddress ?? null;
+    }
     return this.mySelectedAddress;
   }
 


snario:
@geoknee yeah good point. I changed this PR to use a different approach. Can you re-check?

packages/xstate-wallet/src/chain.ts
```diff
@@ -321,6 +321,9 @@ export class ChainWatcher implements Chain {
   }
 
   public get selectedAddress(): string | null {
+    if (this.mySelectedAddress === null && window && window.ethereum) {
+      return window.ethereum.selectedAddress ?? null;
+    }
```
geoknee, line 326: This *might* be a small improvement (if this class has no selectedAddress, try to fall back on `window.ethereum.selectedAddress`), but 

i) I don't think it will help solve the problem in the description of this PR (which is when `window.ethereum.selectedAddress` is not yet defined. 
ii) the change has little to do with the title of this PR. 

Also, I am a bit concerned that this introduces a path for the following potentially confusing situation: the private variable `this.mySelectedAddress` is null but `chain.selectedAddress` is a string. 

It might be a bit nicer to put this fallback logic inside the `ethereumIsEnabled()` getter below.

snario: I think it does fix the issue in the PR, because the value is only undefined for a split second right when the wallet loads

snario: I've updated it to _set_ `this. mySelectedAddress`.

PR #2183, opened 2020-06-15
Increase test timeout

### LINKED ISSUE #2178: Test failure
https://app.circleci.com/pipelines/github/statechannels/monorepo/7376/workflows/b9f0bd56-a7a7-4d9a-b819-da54fc19be29/jobs/34218

lalexgap:
Fixes #2178 (hopefully)

Doubles the timeout for the client seeding test as we're occasionally seeing timeouts there.

### DIFF
--- a/packages/web3torrent/src/library/client-seeding.test.ts
+++ b/packages/web3torrent/src/library/client-seeding.test.ts
@@ -84,7 +84,7 @@ describe('Seeding and Leeching', () => {
         done();
       });
     });
-  }, 10000);
+  });
 
   it('should reach a ready-for-leeching, choked state', done => {
     seeder.seed(defaultFile as File, defaultSeedingOptions(), seededTorrent => {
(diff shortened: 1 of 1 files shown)


kerzhner:
It looks like this `client-seeding.test.ts` specifies a timeout for each individual test. Is the intention of this PR to override all of those timeouts? Or are we trying to set a 20 second timeout for just one test?

lalexgap:
> It looks like this `client-seeding.test.ts` specifies a timeout for each individual test. Is the intention of this PR to override all of those timeouts? Or are we trying to set a 20 second timeout for just one test?

Ah good catch. The intention is to override the timeout for all tests, however I missed the `10000` timeout specified at the end of the test. Fixed in 09afb8fe

kerzhner:
I am really surprised that these tests would take 10 seconds each, let alone 20 seconds. The only thing I can think of is that the tracker server takes a while to start.

https://github.com/statechannels/monorepo/pull/2184 waits for the tracker server to start before running any of the unit tests. It also _decreases_ the timeout for the unit tests to the default 5 seconds.

lalexgap:
Closing in favour of #2184

PR #2187, opened 2020-06-15
[web3torrent] clean up useEffect hooks

kerzhner:
Fixes https://github.com/statechannels/monorepo/issues/2090 (mostly).
Fixes https://github.com/statechannels/monorepo/issues/1835 (mostly).

One common pattern in our code base:
1. Use `useEffect` hook to trigger an asynchronous operation after a component render.
1. At the asynchronous operation, update component state.

If the component is unmounted before step 2 completes, we see errors described in the issues above. This PR uses observables (instead of promises) in `useEffect` hooks in order to properly clean up after `useEffect` hooks. There remains one problematic hook: b6f8f0cb6.

Most of the `web3torrent` unit test log warnings should be fixed by this PR.

### DIFF
--- a/packages/web3torrent/package.json
+++ b/packages/web3torrent/package.json
@@ -126,6 +126,7 @@
     "@types/webtorrent": "0.107.0",
     "@types/workbox-webpack-plugin": "5.1.0",
     "bittorrent-tracker": "9.15.0",
+    "canvas": "2.6.1",
     "enzyme": "3.11.0",
     "enzyme-adapter-react-16": "1.15.2",
     "eslint": "6.8.0",
--- a/packages/web3torrent/src/hooks/use-budget.ts
+++ b/packages/web3torrent/src/hooks/use-budget.ts
@@ -27,6 +27,13 @@ export function useBudget({ready}: {ready: boolean}) {
     const onBudgetUpdatedSubscription = paymentChannelClient.channelClient.onBudgetUpdated(() =>
       getBudget().subscribe(setBudget)
     );
+
+    // Mostly works.
+    // Cleanup is not properly done for the following case:
+    // 1. onBudgetUpdated event arrives.
+    // 2. getBudget().subscribe(setBudget) is invoked.
+    // The operation from step 2 is not properly cleaned up.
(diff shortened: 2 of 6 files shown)


packages/web3torrent/src/hooks/use-budget.ts
```diff
@@ -2,21 +2,42 @@ import {web3TorrentClient} from '../clients/web3torrent-client';
...
+    const onBudgetUpdatedSubscription = paymentChannelClient.channelClient.onBudgetUpdated(() =>
+      getBudget().subscribe(setBudgetUI)
+    );
+
+    // Mostly works.
+    // Cleanup is not properly done for the following case:
+    // 1. onBudgetUpdated event arrives.
+    // 2. getBudget().subscribe(setBudget) is invoked.
+    // The operation from step 2 is not properly cleaned up.
+    // To fix this issue, onBudgetUpdated would need to return an observable instead of a subscription.
```
tomclose, line 36: Would something like this work?
```ts
useEffect(() => {
  let cancelled = false;
  const getAndSetBudget = async () => {
    const budget = await paymentChannelClient.getBudget();

    !cancelled && setBudget(budget);
    !cancelled && setLoading(false);
  };
  if (ready) getAndSetBudget();
  const unsubscribe = paymentChannelClient.channelClient.onBudgetUpdated(getAndSetBudget);

  return () => { cancelled = true; unsubscribe() };
})
```

kerzhner: I like that idea. Implemented in 0c0f8af349a0b735adef3e08f27cd2e8bf03d086

packages/web3torrent/package.json
```diff
@@ -126,6 +126,7 @@
     "@types/webtorrent": "0.107.0",
     "@types/workbox-webpack-plugin": "5.1.0",
     "bittorrent-tracker": "9.15.0",
+    "canvas": "2.6.1",
```
tomclose, line 129: Is this actually used?

kerzhner: We call `useContext` in our unit tests (via component instantiation). Without that dev dependency, the unit tests pass but we see a fairly verbose warning.

PR #2195, opened 2020-06-16
[nitro-docs] Add section on indirect (de)funding

geoknee:
- removes tables from the outcomes section
- adds (long) section on indirect (d)efunding
- adds a passage about setup states

TODO 
- [x] write a tutorial in the nitro-tutorial repo to accompany this page
- [ ] extend to virtual (de)funding (wontfix?)

### DIFF
--- a/packages/nitro-protocol/docs/wallet-dev-tutorial/finalize-a-channel-happy.md
+++ b/packages/nitro-protocol/docs/wallet-dev-tutorial/finalize-a-channel-happy.md
@@ -3,11 +3,15 @@ id: finalize-a-channel-happy
 title: Finalize a channel (happy)
 ---
 
-Finalization of a state channel is a necessary step before defunding it (i.e. releasing the assets that were deposited on chain, in the case of direct funding). In the so-called 'happy' case, all participants cooperate to achieve this.
+Finalization of a state channel is a necessary step before defunding it. In the so-called 'happy' case, all participants cooperate to achieve this.
 
-A participant wishing to end the state channel will sign a state with `isFinal = true`. Then, the other participants support that state. Once a full set of `n` such signatures exists \(this set is known as a **finalization proof**\) anyone in possession may use it to finalize the `outcome` on-chain. They would do this by calling `conclude` on the adjudicator.
+A participant wishing to end the state channel will sign a state with `isFinal = true`. Then, the other participants may support that state. Once a full set of `n` such signatures exists \(this set is known as a **finalization proof**\) the channel is said to be 'closed' or 'finalized'.
+
+In most cases, the channel would be finalized and defunded off chain, and no contract calls are necessary.
 
 ## Call `conclude`
+
+In the case where assets were deposited against the channel on chain (the case of direct funding), anyone in possession of a finalization proof may use it to finalize the `outcome` on-chain. They would do this by calling `conclude` on the adjudicator. This enables [assets to be released](./release-assets).
 
 The conclude method allows anyone with sufficient off-chain state to immediately finalize an outcome for a channel without having to wait for a challenge to expire (more on that later).
 
--- a/packages/nitro-protocol/docs/wallet-dev-tutorial/finalize-a-channel-happy.md
+++ b/packages/nitro-protocol/docs/wallet-dev-tutorial/finalize-a-channel-happy.md
@@ -5,9 +5,9 @@ title: Finalize a channel (happy)
 
 Finalization of a state channel is a necessary step before defunding it. In the so-called 'happy' case, all participants cooperate to achieve this.
 
-A participant wishing to end the state channel will sign a state with `isFinal = true`. Then, the other participants may support that state. Once a full set of `n` such signatures exists \(this set is known as a **finalization proof**\) the channel is said to be 'closed' or 'finalized'.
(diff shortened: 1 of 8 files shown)


geoknee:
Deploy preview at https://5ee8d5270e6a998d6a0a850b--nitro-protocol.netlify.app/docs/wallet-dev-tutorial/fund-and-defund-off-chain

packages/nitro-protocol/docs/wallet-dev-tutorial/finalize-a-channel-happy.md
```diff
@@ -3,12 +3,16 @@ id: finalize-a-channel-happy
 title: Finalize a channel (happy)
 ---
 
-Finalization of a state channel is a necessary step before defunding it (i.e. releasing the assets that were deposited on chain, in the case of direct funding). In the so-called 'happy' case, all participants cooperate to achieve this.
+Finalization of a state channel is a necessary step before defunding it. In the so-called 'happy' case, all participants cooperate to achieve this.
 
-A participant wishing to end the state channel will sign a state with `isFinal = true`. Then, the other participants support that state. Once a full set of `n` such signatures exists \(this set is known as a **finalization proof**\) anyone in possession may use it to finalize the `outcome` on-chain. They would do this by calling `conclude` on the adjudicator.
+A participant wishing to end the state channel will sign a state with `isFinal = true`. Then, the other participants may support that state. Once a full set of `n` such signatures exists (this set is known as a **finalization proof**) the channel is said to be 'closed' or 'finalized'.
```
tomclose, line 8: I'm not sure if we should use "finalization proof". It sounds like it should be proof that the channel has finalized - but it hasn't actually finalized at this point. I think I prefer "conclusion proof", but we could just remove the bracketed part altogether.

geoknee: Probably we should decide on the terminology now, and then do an audit of the entire docs site to make sure it is consistent? Off the top of my head (this would probably benefit from some more careful analysis)

Option 1
---
_conclusion-proof_ := support proof for an `isFinal=true` state (seems a bit awkward).
_finalized_ := property of a channel, where a conclusion proof has been successfully submitted to the chain and has caused the appropriate change of state.

Option 2
---
_conclusion-proof_ := support proof for an `isConclude=true` state (requires changes to the contracts 😞 )
_finalized_ := property of a channel, where a conclusion proof has been successfully submitted to the chain and has caused the appropriate change of state.

Option 3 (closest to current)
---
 _finalization proof_ := support proof for an `isFinal=true` state
_finalized_ := property of a channel, where a finalization proof exists or is posessed by a given agent. 
_finalized on chain_ := property of a channel, where a finalization proof has been successfully submitted to the chain and has caused the appropriate change of state.

There is also the issue of the Adjudicator method being called `conclude`. This suggests (assuming we don't want to change the contracts now):

Option 4
---
 _finalization proof_ := support proof for an `isFinal=true` state
_finalized_ := property of a channel, where a finalization proof exists or is posessed by a given agent. 
_concluded_ := property of a channel, where a finalization proof has been successfully submitted to the chain and has caused the appropriate change of state.

andrewgordstewart: I would add
# Option 5
---
_finalization proof_:= support proof where `isFinal==true` for every state
_final_ := property of a (channel, agent), where a finalization proof exists or is possessed by a given agent.
_finalized_ := property of a channel, where a finalization proof has been successfully submitted to the chain and has caused the appropriate change of state.
_finalize_ := contract method to finalize the channel, given a finalization proof

# Option 6
---
_finalization proof_:= support proof where `isFinal==true` for every state
_finalizable_ := property of a (channel, agent), where a finalization proof exists or is possessed by a given agent.
_finalized_ := property of a channel, where a finalization proof has been successfully submitted to the chain and has caused the appropriate change of state.
_finalize_ := contract method to finalize the channel, given a finalization proof

This option makes the most sense to me. All other options have two words that are synonyms, which seems a bit arbitrary.. This option uses one word which means something different: "just about finalized, but not quite".

geoknee: This thread started due to the confusion around _finalization proof_ =/=> channel is _finalized_. How about:

Option 6b
---
_finalizability proof_ := support proof where `isTerminal==true` for every state
_finalizable_  := property of a (channel, agent), where a finalizability proof exists or is possessed by a given agent.
_finalized_ := property of a channel, where **either** a finalizability proof has been successfully submitted to the chain  , or a challenge expired (either way causing the appropriate change of state)
_finalize_ := contract method to finalize the channel, given a finalizability proof

(I don't think either of `isFinal` or `isFinalizable` are great because an `isFinal=false` state can be challenged with and hence lead to the channel finalizing on chain.  Hence something completely different -- `isTerminal`).

In fact, I think we probably should have the terminology do justice to the 2 different ways in which a channel can be finalized on chain. Hence:

Option 7
---
_termination proof_ := support proof where `isTerminal==true` for every state
_terminated_  := property of a (channel, agent), where a termination proof exists or is possessed by a given agent.
_finalized_ := property of a channel, where **either** a termination proof has been successfully submitted to the chain , or a challenge expired (either way causing the appropriate change of state), 
_finalize_ := contract method to finalize the channel, given a termination proof

PR #2208, opened 2020-06-16
Fix factory reset

andrewgordstewart:
Fixes https://github.com/statechannels/monorepo/issues/2198


tomclose:
Going to merge, so we can get into this evening's deploy.

PR #2209, opened 2020-06-16
Only allow withdrawal when no active torrent

lalexgap:
Adds a `canWithdrawFeed` to the web3 torrent client that is used to notify the file component when the user should be allowed to withdraw.

The withdrawal button is disabled with a tooltip when there is an active torrent.

image

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -27,6 +27,7 @@ import {Message} from '@statechannels/client-api-schema';
 import {utils} from 'ethers';
 import {logger} from '../logger';
 import {track} from '../analytics';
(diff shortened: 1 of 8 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -45,12 +46,17 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
   channelIdToTorrentMap: Record<string, string> = {};
   stoppedTorrents: Record<string, boolean> = {};
 
+  private canWithdrawSubject: rxjs.Subject<boolean>;
+  public get canWithdrawFeed() {
+    return this.canWithdrawSubject as rxjs.Observable<boolean>;
```
kerzhner, line 51: Curios, why is the assertion needed here?

PR #2217, opened 2020-06-17
Troubleshooting mutliple leecher test

tomclose:
Fixes https://github.com/statechannels/monorepo/issues/2215

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -169,7 +169,8 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
       (torrent as any).maxWebConns = 0; // don't start downloading from existing upload peers
       this.stopUploading(infoHash); // also stop uploading immediately
 
-      await this.closeTorrentChannels(torrent, true);
+      await this.closeTorrentChannels(torrent, false);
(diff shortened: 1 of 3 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -582,6 +583,26 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     );
   }
 
+  protected async waitForMySeederChannelsToClose(torrent: PaidStreamingTorrent) {
```
geoknee, line 586: are MySeederChannels the same as mySeedingChannels?

andrewgordstewart:
These are the screenshots at the end of the multiple-leecher test on 42eb44e.

Screen Shot 2020-06-17 at 9 29 29 AM

I think React is still reacting to information, because the `.html` files are a bit different.

I am confused about who this is:
Screen Shot 2020-06-17 at 9 49 55 AM

geoknee:
@andrewgordstewart I don't think this branch includes this change https://github.com/statechannels/monorepo/pull/2216, which solved a problem about inconsistent blockies.

andrewgordstewart:
> I am confused about who this is:
> Screen Shot 2020-06-17 at 9 49 55 AM

Actually, none of the _peers_ blockies match up with any of the _who am I_ blockies.

PR #2225, opened 2020-06-17
[web3torrent] for leecher, do not set seeding channel id to null blindly

kerzhner:
In my testing, the following is the usual sequence of events when the leecher cancels download from one seeder:
1. Leeching channel id is set to null in `onChannelUpdated`.
1. Seeding channel id is set to null incorrectly in `closeChannel`.

This race condition might impact wires that are both seed and leech wires.

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -619,7 +619,7 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
         wire.paidStreamingExtension.leechingChannelId = null;
         log.info(`Leechning channel ${channelId} set to null`);
       }
-      if (channelId === wire.paidStreamingExtension.leechingChannelId) {
+      if (channelId === wire.paidStreamingExtension.seedingChannelId) {
         wire.paidStreamingExtension.seedingChannelId = null;
         log.info(`Seeding channel ${channelId} set to null`);
       }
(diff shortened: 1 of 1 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -611,12 +613,15 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
       await this.paymentChannelClient.closeChannel(channelId);
+      // Note: it is possible that onChannelUpdated already set the wire.paidStreamingExtention leeching or seeding id to null
+      // while we were awaiting the above promise
       if (channelId === wire.paidStreamingExtension.leechingChannelId) {
         wire.paidStreamingExtension.leechingChannelId = null;
-        log.info(`Payment Channel closed: ${channelId}`);
-      } else {
+        log.info(`Leechning channel ${channelId} set to null`);
+      }
+      if (channelId === wire.paidStreamingExtension.leechingChannelId) {
```
tomclose, line 622:
```
 if (channelId === wire.paidStreamingExtension.seedingChannelId) {
```

tomclose:
See here: https://github.com/statechannels/monorepo/issues/2220#issuecomment-645602510

> It's possible that multiple requests only arrive once the wire is underway, and the seeder cancelling has somehow removed the channelId suddenly from an active wire?

It's also consistent with another problem that Andrew and I were discussing on slack:

Screen Shot 2020-06-17 at 10 13 13 PM

packages/web3torrent/src/library/web3torrent-lib.ts
kerzhner, line 622:
```suggestion
      if (channelId === wire.paidStreamingExtension.seedingChannelId) {
```

PR #2226, opened 2020-06-17
Stop button clean up

lalexgap:
Some basic logic to hide/disable the stop button when appropriate.


packages/web3torrent/src/components/torrent-info/TorrentInfo.tsx
```diff
@@ -52,22 +63,26 @@ const TorrentInfo: React.FC<TorrentInfoProps> = ({
...
+              infoHash: torrent.infoHash,
+              magnetURI: torrent.magnetURI,
+              filename: torrent.name,
+              filesize: torrent.length
+            });
+            setButtonDisabled(true);
+            return web3TorrentClient.cancel(torrent.infoHash);
+          }}
+        >
+          Stop {torrent && torrent.status === Status.Seeding ? 'Seeding' : 'Downloading'}
```
kerzhner, line 83: Does this work after the leecher finishes the download but is still seeding?

lalexgap: Fixed here 1efdbb2

PR #2229, opened 2020-06-18
Mark torrent as destroyed to cancel. Fixes #2227

### LINKED ISSUE #2227: New channels after stop download as leecher
Repro steps:
1. Run the multiple leecher test.
1. Let one of the leechers progress about half way through the download.
1. Stop the leecher.
1. Observe existing channels conclude nicely, but then see new downloading/uploading/disconnecting channels crop up. None of the new channels actually upload or download any data.

The screenshot shows the stopped leecher after all of the other leechers have finished running (which is a bit after stop was initiated)
Screen Shot 2020-06-17 at 3 54 24 PM
[cancel-broken.txt](https://github.com/statechannels/monorepo/files/4796017/cancel-broken.txt)

part1

tomclose:
I've run the multiple leacher test multiple times, cancelling at random. All seems to work 🎉  🤞 .

### DIFF
--- a/packages/web3torrent/src/library/web3torrent-lib.ts
+++ b/packages/web3torrent/src/library/web3torrent-lib.ts
@@ -170,14 +170,14 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
     log.debug({torrent}, '> Cancelling download');
 
     if (torrent) {
+      // if we don't stop discovery, we'll still try to connect to new peers after we've
+      // turned off torrenting, which will lead to "Error: torrent is destroyed"
+      await new Promise(resolve => (torrent as any).discovery.destroy(resolve));
+
       // We want to stop all torrenting, but don't want to close the wires, as we need them
       // to close the channels. We therefore manually set the destroyed flag on the
       // torrent instead of calling the destroy() method.
       (torrent as any).destroyed = true;
-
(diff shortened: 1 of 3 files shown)


packages/web3torrent/src/library/web3torrent-lib.ts
```diff
@@ -170,12 +170,11 @@ export default class WebTorrentPaidStreamingClient extends WebTorrent {
...
-      // this.closeChannels would race the `makePayment` function, and one of them would fail with a
-      // "Not my turn" error.
-      torrent.pause(); // prevents connections to new peers
-      (torrent as any).maxWebConns = 0; // don't start downloading from existing upload peers
+      // We want to stop all torrenting, but don't want to close the wires, as we need them
+      // to close the channels. We therefore manually set the destroyed flag on the
+      // torrent instead of calling the destroy() method.
+      (torrent as any).destroyed = true;
+
       this.stopUploading(infoHash); // also stop uploading immediately
```
geoknee, line 178: Do we still need this line, then?

tomclose: I think we do, as setting `destroyed` manually doesn't have any effect on existing wires, so we need something to stop the data being piped down them.

PR #2230, opened 2020-06-18
Improve Wallet favicon

geoknee:
- reference in html so that metamask displays it
- use svg so it looks sharp

Current: 
Screenshot 2020-06-18 at 09 18 36
Fixed:
Screenshot 2020-06-18 at 09 35 23
Fixed and sharp:
Screenshot 2020-06-18 at 09 53 29


snario:
These are my favourite kind of PRs 😍

PR #2233, opened 2020-06-18
Button emphasis

geoknee:
~Jun-18-2020 16-57-04~

~Double pulse, 5 seconds of nothing, repeat.~

~Still seductive, but hopefully not _too_ seductive?~ Prefer box shadow option, below. 

Whatever we end up using here, we can roll out to other key moments in the user story. For example, I just finished the download: the "share link" button could pulse this way to nudge me into exploring that.


geoknee:
I think the box shadow is probably less controversial: 
Screenshot 2020-06-18 at 17 09 58

Issue #1801, opened 2020-05-12
Upgrade to Node 12


kerzhner:
Node 10 is moving to maintenance mode very soon on 2020-05-19.

PR #2254, opened 2020-06-24
[monorepo] Try to upgrade to node 12

geoknee:
#1801 

I can get `yarn` to work fine locally, so need to debug the circle configuration to work out why it is failing there. 

TODO 
- [ ] clean up git history
- [ ] reenable native solc compilation


geoknee:
In favour of #2257

PR #2260, opened 2020-06-26
Workflow experiments

tomclose:
Experimenting with different ways of writing workflows.


packages/xstate-wallet/src/workflows/directly-funded-app.ts
```diff
@@ -0,0 +1,360 @@
+import {StateMachine, createMachine, Guard} from 'xstate';
```
tomclose, line 1: Ignore this file

PR #2262, opened 2020-06-29
[nitro protocol] Do not pack code that references node.js builtins

geoknee:


packages/nitro-protocol/test/test-helpers.ts
```diff
@@ -12,7 +12,6 @@ import {
   Allocation,
   AllocationItem,
 } from '../src/contract/outcome';
-import {hashState, State} from '../src/contract/state';
 import fs from 'fs';
```
geoknee, line 15: Importing this module may cause an error if the consuming project tries to bundle this dependency for the browser. We solve this problem in the wallet by telling webpack to mock out certain modules. This changeset should make it nicer for the consuming packages so they do not have to do that (by not exposing any non-browser modules).

geoknee: e.g. 
https://github.com/statechannels/monorepo/blob/78a7ac1a26e7dec0cb9e6ce006c2186a26b1749c/packages/xstate-wallet/config/webpack.config.js#L653-L664

geoknee:
This approach seems a bit heavy handed and also brittle. Pursuing an alternative that builds for node _and_ the browser on a different branch.

PR #2264, opened 2020-06-30
[nitro-protocol] Build for web using webpack

geoknee:
...along with some general tidying up of imports and exports, including modifying _all_ instances of `import {foo} from 'ethers/bar'` to `import {bar} from 'ethers'`.

The aim is for consuming projects to be able to import `@statechannels/nitro-protocol` and run it in the browser without having to worry about shimming their build to mock out nodejs-specific code. It should Just Work.

Under this changeset, we take the _output_ of `tsc`, and bundle it using `webpack`: stubbing out things like `fs` as we do so.

Nonessential side effect: upgrading `ts-loader` in other packages.

PR #2266, opened 2020-07-01
[benchmarking] Add benchmarking package.

geoknee:
Preceded by #2264.

It's a bit rough, but this new package contains new code that can be run in node-js and in the browser, to enable us to time various function calls. 

Running benchmarks in the browser currently requires manually clicking on a button:
Screenshot 2020-07-01 at 17 39 28
It would be possible to automate that with puppeteer, of course. 

Package includes a python script to generate plots: 
benchmark


geoknee:
https://github.com/statechannels/monorepo/blob/fe0fcb88e480e910f573b0babd812b80b9ba7995/packages/benchmarking/readme.md

PR #2267, opened 2020-07-02
Extract core wallet code to separate package

andrewgordstewart:
This extracts a bunch of code that we want to share between the server/browser wallets & hub into a package called `wallet-components`. This is mostly store, types, etc.

This PR attempts to move files with the minimal changes possible. We should perform refactors to deal with issues below in separate PRs, as this change touches so many files.

# TODO (future PRs)
- [ ] Remove `'dom'` environment from components `tsconfig.json`
   - [ ] Remove references to `window`
   * the components package should work in either the node or browser environment
- [ ] Remove `BigNumber` from components
   * I think it provides less utility than it costs
- [ ] `turnNum`, `challengeDuration`, etc. should be type `number`
- [ ] define module exports for @statechannels/wallet-components, as opposed to via file lookup in this PR
- [ ] Resolve `FIXME`s added in this changeset
- [x] rename `components` -> `core`


packages/wallet-components/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@statechannels/wallet-components",
+  "description": "State channel wallet components.",
+  "version": "0.0.1",
+  "author": "Alex Gap",
+  "dependencies": {
+    "@statechannels/client-api-schema": "0.0.1",
+    "@statechannels/nitro-protocol": "0.1.2",
+    "@statechannels/wire-format": "0.0.1",
+    "@statechannels/wallet-components": "0.0.1",
```
geoknee, line 10: It depends on itself?

PR #2269, opened 2020-07-02
Patch devtools

geoknee:
Include module resolution to allow install on node12

### DIFF
--- a/packages/devtools/package.json
+++ b/packages/devtools/package.json
@@ -15,6 +15,7 @@
     "dotenv": "8.2.0",
     "dotenv-expand": "5.1.0",
     "etherlime": "2.3.3",
+    "etherlime-lib": "1.1.5",
     "ethers": "4.0.45",
     "fs-extra": "8.1.0",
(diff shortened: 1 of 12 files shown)


geoknee:
Maybe we should remove the resolution from the top level `package.json`? Not sure

geoknee:
I'm not sure this actually helps anyone trying to add @statechannels/devtools to their project if they are using node12.

Perhaps instead we ought to add a warning/hint in the readme (consuming projects need only add the resolutions line to their package.json), or specify node10 in the engines field 🤢 ?

packages/devtools/package.json
```diff
@@ -15,6 +15,7 @@
     "dotenv": "8.2.0",
     "dotenv-expand": "5.1.0",
     "etherlime": "2.3.3",
+    "etherlime-lib": "1.1.5",
```
geoknee, line 18: 🤦

geoknee: We need this.

PR #2270, opened 2020-07-02
[nitro] Tweak readme

geoknee:
The quoted contract would not actually compile.


snario:
Could these reference snippets of code in the repo in the future? (like embed a github UI)

PR #2273, opened 2020-07-03
[nitro-protocol] Reduce contract artifacts size

geoknee:
Should reduce the size of this package (to roughly 50%, I think).

https://bundlephobia.com/result?p=@statechannels/nitro-protocol@0.1.3

### DIFF
--- a/packages/nitro-protocol/bin/compile.js
+++ b/packages/nitro-protocol/bin/compile.js
@@ -36,7 +36,7 @@ compile.on('close', code => {
 // strip out uneeded entries from artifacts
 function stripArtifacts() {
   const files = fs.readdirSync(path.resolve(__dirname, '../build/contracts'));
-  console.log('Stripping uneeded entried from the following artifacts: ', files);
+  console.log('Stripping uneeded entries from the following artifacts: ', files);
   for (file in files) {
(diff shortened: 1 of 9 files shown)


packages/nitro-protocol/bin/compile.js
```diff
@@ -26,5 +29,22 @@ compile.stderr.on('data', data => {
...
+
+// strip out uneeded entries from artifacts
+function stripArtifacts() {
+  const files = fs.readdirSync(path.resolve(__dirname, '../build/contracts'));
+  console.log('Stripping uneeded entried from the following artifacts: ', files);
+  for (file in files) {
+    let artifact = require(path.resolve(__dirname, '../build/contracts/' + files[file]));
+    delete artifact['ast'];
+    delete artifact['legacyAst'];
+    delete artifact['source'];
```
geoknee, line 44: AFAICT we don't need these fields. There might be others we don't need.

geoknee: Looks like we do need the `ast` in order to generate documentation https://app.circleci.com/pipelines/github/statechannels/monorepo/7717/workflows/b06863cb-d388-42da-b932-224ae689ab98/jobs/36629/steps

geoknee:
Included in #2271

Issue #1652, opened 2020-05-04
Harmonize types across xstate-wallet and nitro-protocol


geoknee:
I wanted to record the places where we do conversion of objects between the types in either package, and kick off a discussion about whether we should try and consolidate those types and remove the need for the conversion. This is spurred by the upcoming release of our protocol contracts, documentation and accompanying typescript files. 

For example, these imports give an idea of what we currently convert between: 
https://github.com/statechannels/monorepo/blob/06180ee6f377393c737a2b49586ae03492802c60/packages/xstate-wallet/src/store/state-utils.ts#L1-L21

Let's look at one example in more detail: 
https://github.com/statechannels/monorepo/blob/06180ee6f377393c737a2b49586ae03492802c60/packages/xstate-wallet/src/store/state-utils.ts#L28-L45

If we changed the type of `challengeDuration` and `turnNum` in `nitro-protocol` to `ethers.BigNumber`, we wouldn't have to do so much translation inside the xstate-wallet, and neither would other (3rd party) consumers of nitro-protocol who also plan on using `ethers.js`.

On the other hand, it seems we made the choice to use `number` in nitro-protocol [deliberately](https://github.com/statechannels/monorepo/commit/457da6abd028b3b4a60ca1b3c22d2950e43994d6), likely because these fields are actually only `uint48` in the contracts and not `uint256`. So it might make more sense to make the change in the other direction, and have these fields stored as `numbers` in all xstate-wallet code. Javascript can [safely store](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) 48 bit integers as numbers.

geoknee:
Places in the monorepo that are relevant to this dicussion: 

Types imported into state-utils (i.e. the alternatives to the nitro-protocol types):
https://github.com/statechannels/monorepo/blob/06180ee6f377393c737a2b49586ae03492802c60/packages/xstate-wallet/src/store/types.ts#L36-L39

Replicated state-utils inside the `simple-hub`:
https://github.com/statechannels/monorepo/blob/88bd1b4b49788c657eef33182a080860af515a2c/packages/simple-hub/src/wallet/xstate-wallet-internals/index.ts#L2-L18

Helper functions that *might* want to live in nitro-protocol:
[...]

PR #2277, opened 2020-07-04
Remove unnecessary use of `BigNumber` classes

andrewgordstewart:
This is an attempt to remove some of our technical debt:
- we use `BigNumber` values when we don't need to
- it is _annoying_ to have to use `BigNumber` when we don't need to
   - @kerzhner described this as "death by a thousand paper cuts"
- we **do not** currently use `BigNumber` well enough to ensure that values which _should_ be a `BigNumber` instance _are_ a `BigNumber` instance

The following changes are made:
- `turnNum`, `finalizesAt`, and `challengeDuration` are all `uint48` types in solidity
- `turnNum`, `finalizesAt`, `challengeDuration` and `blockNumber` are all `number` types in javascript

I'll leave a couple of notes here:
1. after changing some `BigNumber` values to `number`, I ran into the _same bug_ three times, where the compiler believed a variable had type `BigNumber`, but its value had type `Object`. This lead to a runtime error `amount.toHexString is not a function`
2. Most of the contract changes were automatic formatting triggered when I hit save. I did not notice this at the time.
3. For some random reason, the create-and-fund test fails with the memory backend, which I believe is due to concurrent "transactions"

(1) is an example of the reason so many of us don't want to use `BigNumber`s. Considering that I had already done 80% of the work of removing the use of `BigNumber` for amounts, I would have spent less time removing the possibility of this bug than finding the source of this bug.


geoknee:
#1652

andrewgordstewart:
Converted to draft, as I realized this requires a migration.

PR #2278, opened 2020-07-06
[nitro-docs] Index redirects to docs entry

geoknee:
home splash screen refactored into a separate file so it isn't lost forever
remove some dead wood


geoknee:
Preview https://5f030df374983ff19b9f4bc1--nitro-protocol.netlify.app/

PR #2279, opened 2020-07-08
Resurrect `hub` package (as `server-wallet`)

andrewgordstewart:
My goal was to create a package that has some seed data that I can put into a postgresql database. We used to have a "hub", which ran a "server wallet" that had a postgresql database and some seed data. So,

- I reverted the commit 043cbe3, which removed that package
- I then
  1. removed firebase related code
  2. removed consensus-app related code
  3. removed code pertaining to acting as a hub
  4. Removed the old migrations (we should use a different schema)
  5. some more odds and ends (dependencies, readme, env vars, etc.)
- Then, I copied over some code from `wallet-core`, and edited it.
   - The relevant code are `store-types`, `state-utils`, and `bn`
   - This is probably contentious, and it's probably worth reverting that, and depending on `wallet-core`
- Then, I copied over the logic from `ChannelStoreEntry` onto the `Channel` model, with the new schema
  - The new schema uses a couple of `jsonb` columns to avoid creating relations to other tables
- Then, I created some seed data that I could insert

## Test plan
I think that we can avoid hitting the database in most tests.

We intend to write some "workflow" code which
1. accepts a set of channel states
2. determines what "Protocol step" we're in
3. declares some side effects

In step 1, we can just pass a `ChannelModel` class instance. -- we don't have to fetch the data from the database. So, our tests can look like
```typescript
test('sending prefund setup', () => {
  const app = Channel.fromJson(fixtures.app.gotPrefundSetup);
  const guarantor = Channel.fromJson(fixtures.unknown);
  const joint = Channel.fromJson(fixtures.unknown);
  const ledger = '0xabc';

  expect(appWorkflow({app, guarantor, joint, ledger})).toMatchObject({
    //
  });
});
```

## TODO:
- [ ] Switch to using `SC_ENV` (or better) environment variable setting
- [ ] Get a test running that tests something useful, and passes


packages/server-wallet/src/bn.ts
```diff
@@ -0,0 +1,30 @@
...
+      [method](b)
+      .toHexString();
+  } else {
+    return EthersBigNumber.from(a)
+      [method](b)
+      .toNumber();
+  }
+};
+
+export const BigNumber = {
```
andrewgordstewart, line 20: This was partially playing around with the idea of not using `BigNumber` class instances.

I'd remove this, and depend on the types defined in `wallet-core`, with the goal of not adding yet another `State` type.

packages/server-wallet/Procfile
```diff
@@ -0,0 +1 @@
+hub: node ./lib/hub/server.js
```
kerzhner, line 1: Let's get rid of this file. We are using docker containers now (as opposed to Heroku slugs).

kerzhner:
I haven't looked closely at this PR, but this seems like a reasonable approach.

PR #2281, opened 2020-07-10
Publish wallet-core and client-api-schema

andrewgordstewart:
This PR prepares the `wallet-core` package and `client-api-schema` package for publishing:
- updating their `package.json` files so that they publish the right things (up-to-date, compiled typescript files)
- updating their `tsconfig.json` files so that they compile the right things (no tests)

I've taken the liberty of publishing these two packages as of 2c7f551. You can explore the files that got published:
- https://www.npmjs.com/package/@statechannels/wallet-core
- https://www.npmjs.com/package/@statechannels/client-api-schema
(The "Explore" tab will let you browse the files.)

### DIFF
--- a/packages/wallet-core/package.json
+++ b/packages/wallet-core/package.json
@@ -27,7 +27,7 @@
   },
   "license": "MIT",
   "scripts": {
-    "prepare": "yarn build",
+    "prepare": "rm -rf lib; yarn build",
     "build": "yarn tsc -b",
     "build:ci": "yarn build",
     "lint:check": "eslint . --ext .ts --cache",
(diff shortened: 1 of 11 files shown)


packages/wallet-core/package.json
```diff
@@ -23,7 +27,7 @@
   },
   "license": "MIT",
   "scripts": {
-    "prepare": "yarn build",
+    "prepare": "rm -rf lib; yarn build",
```
lalexgap, line 30: Just curious, why do we need to remove the `lib` folder?

andrewgordstewart: I don't think it's guaranteed that running the compiler will remove old, compiled files. When preparing, a precursor to publishing, we want to make sure the contents of lib are exactly what we intended.

PR #2282, opened 2020-07-11
Add BigNumber utility class

andrewgordstewart:
Amounts cannot be safely stored in a Javascript number.

There is some contention over whether amounts should be stored in a string, or in a BigNumber class. Things
like storing and fetching amounts from the database, and serializing and deserializing are more complicated
with BigNumber classes. Constructing test data requires constantly importing BigNumber, and calling
`BigNumber.from(1)`, etc.

These problems are certainly not impossible to overcome, but they do add complexity, and can possibly add
bugs. Furthermore, saying that the type of an amount is BigNumber means your project has to commit to
using ethers, which is opinionated. Storing big numbers in string Javascript values is less opinionated.

This changeset is quite small, even after adding the (somewhat opinionated) Uint256 type.

Strings are not assignable to Uint256. The programmer must assert that strings are of type Uint256.
I claim that this is less mental overhead than importing BigNumber from ethers, than calling
`BigNumber.from(1)` If preferred, you may import BigNumber from `'./bignumber'`, and call
`BigNumber.from(1)` -- it will return a value of type `Uint256`.

### DIFF
--- a/packages/wallet-core/src/bignumber.ts
+++ b/packages/wallet-core/src/bignumber.ts
@@ -3,7 +3,10 @@ import {Uint256} from './types';
 
 type T = BigNumberish;
 
-const binaryOperator = (name: keyof EthersBigNumber) => (a: T, b: T): Uint256 => {
+const binaryOperator = <S extends any = Uint256>(name: keyof EthersBigNumber) => (
+  a: T,
+  b: T
+): S => {
   if (typeof EthersBigNumber.from(a)[name] !== 'function') throw Error(`Invalid method ${name}`);
 
   const result = (EthersBigNumber.from(a)[name] as any)(b);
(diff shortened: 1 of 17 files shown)


packages/wallet-core/src/bignumber.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * This is a convenience class that looks like instances of ether's BigNumber class.
+ * This way, people can call
+ * ```
+ * BN.eq(2, '3') // false
+ * BN.add('0x123', 4) // '0x661'
+ * BN.add('0xabc', 10) // '0xac6'
+ */
+
+export class BigNumber {
```
lalexgap, line 30: Naming this `BigNumber` may confuse it with ether's `BigNumber` (especially when vscode adds the import statement). Maybe we could go with something slightly different so we don't run into that?

andrewgordstewart: What about `BN`?

lalexgap: There is a `BN.js` [library](https://github.com/indutny/bn.js/) but that's probably fine since we don't use it, so I'd be good with `BN`

PR #2294, opened 2020-07-16
[docs, channel-client] Add autogenerated API docs to our website

geoknee:
I spent quite a bit of time trying TypeDoc + `typedoc-plugin-markdown` with docusarus theme. The output was not great, so I spent some more time trying to write my own theme. I gave up on this. 

I also briefly considered JSDoc. 

Then I tried api-extractor, which worked pretty well!

~This branch mostly a POC at the moment. It has thrown up a few questions about our overall approach.~


geoknee:
Preview here https://5f106017597266d269ed975d--nitro-protocol.netlify.app/docs/next/channel-client-api/channel-client.channelclient

packages/channel-client/src/channel-client.ts
```diff
@@ -94,10 +146,28 @@ export class ChannelClient implements ChannelClientInterface {
...
+   */
   async joinChannel(channelId: string): Promise<ChannelResult> {
     return this.provider.send({method: 'JoinChannel', params: {channelId}});
   }
 
+  /**
+   * Updates the state of a channel
+   *
+   * @param channelId id for the state channel
+   * @param participants Array of Participants for this channel TODO not clear why this is here, as it must be constant once the channel has its id.
```
geoknee, line 165: ❓

packages/channel-client/src/channel-client.ts
```diff
@@ -159,10 +272,25 @@ export class ChannelClient implements ChannelClientInterface {
...
   async getBudget(hubAddress: string): Promise<DomainBudget> {
     return this.provider.send({method: 'GetBudget', params: {hubAddress}});
   }
 
+  /**
+   * Requests the funds to be withdrawn from this site's ledger channel
+   *
+   * @param hubAddress The address of a state channel hub
+   * @param hubOutcomeAddress An ethereum account that the hub's funds will be paid out to TODO this doesn't make sense
+   * @returns A promise that resolves to a DomainBudget.
```
geoknee, line 291: ❓ 
(for `closeAndWithdraw`, I don't understand why I should get to say anything about the hub's outcome address)

geoknee: This probably also does not belong, but we should double check to see what (if anything) the wallet does with it.

packages/channel-client/src/channel-client.ts
geoknee: Agreed that this does not belong here.

packages/channel-client/src/channel-client.ts
geoknee: TODO

packages/channel-client/src/channel-client.ts
geoknee: #2296

snario:
It would be nice if the ways an API call can fail were visible:

https://5f106017597266d269ed975d--nitro-protocol.netlify.app/docs/next/channel-client-api/channel-client.channelclient.updatechannel

packages/channel-client/src/channel-client.ts
geoknee: #2306

geoknee:
> We could instead type `ChannelResult | UpdateChannelError` and type those errors.

Agreed, but this is way out of scope for this PR, right?

snario:
> Agreed, but this is way out of scope for this PR, right?

Yeah, but the existing docs manually add the errors. e.g., https://api.statechannels.org/#join-channel

geoknee:
> Yeah, but the existing docs manually add the errors. e.g., https://api.statechannels.org/#join-channel

Ah, those docs are for the JSON-RPC interface, which these docs are not necessarily going to replace on their own. 

Regardless,  I think it should be possible to thread those errors through into these docs. 

In fact they are in there already, after a fashion https://5f106017597266d269ed975d--nitro-protocol.netlify.app/docs/next/channel-client-api/channel-client.userdeclinederrorcode

I'll see what I can do to make that nicer.

geoknee:
I think we should merge this now (establishes the docgen pipeline) and I'll keep improving the annotations and integration into the website.

geoknee:
Latest preview: https://5f1801eb24378972643f1372--nitro-protocol.netlify.app/docs/next/channel-client-api/channel-client.channelclient

PR #2299, opened 2020-07-17
[channel-client] Include src and tests directories in npm package

kerzhner:


packages/channel-client/package.json
```diff
@@ -26,6 +26,10 @@
     "ts-jest": "25.0.0",
     "typescript": "3.7.5"
   },
+  "files": [
+    "lib/src",
+    "lib/tests"
```
andrewgordstewart, line 31: Why would you include the tests?

kerzhner: https://github.com/statechannels/monorepo/blob/39d32bd65bd2aabd27d831eddb25035dc368fc5a/packages/channel-client/src/index.ts#L6

kerzhner: Tests are imported in `index.ts` which is the package entrypoint.

andrewgordstewart: Can we include just that file?

It's besides the point of this PR, but it's strange to me that the `tests` folder of the `channel-client` package must export a mock `ChannelProvider`.

kerzhner: We probably can include just that fille.

kerzhner: 58fec193a

PR #2303, opened 2020-07-20
Publish @statechannels/channel-provider 0.0.5

snario:
The previous version did not include the required types files.

### DIFF
--- a/packages/channel-provider/package.json
+++ b/packages/channel-provider/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@statechannels/channel-provider",
-  "version": "0.0.3",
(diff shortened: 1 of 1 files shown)


packages/channel-provider/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@statechannels/channel-provider",
-  "version": "0.0.3",
+  "version": "0.0.5",
```
geoknee, line 3: what happened to 0.0.4?

snario: @lalexgap published it but did not bump the code

PR #2305, opened 2020-07-20
Remove apps

andrewgordstewart:
When we get the apps monorepo running where we want it, we can pull this bandaid!

Getting things to work required many iterations of `yarn && yarn build`, since dev dependencies used by `xstate-wallet` were not specified in its `package.json` file.


snario:
My hope was to get a ✅ on CircleCI on `apps` and then do this

geoknee:
Does this remove `simple-hub`? Should it?

andrewgordstewart:
> Does this remove `simple-hub`? Should it?

It does not. Maybe it should! Should I?

geoknee:
Well, I spotted it over in the apps repo. So unless there is something I am misunderstanding, it should go. Could be done in a separate PR, though.

snario:
Yes, `simple-hub` should only be in `apps`.

PR #2306, opened 2020-07-20
[channel-client] Clean up API and types

geoknee:
This PR

- simplifies `closeAndWithdraw` parameters: the only param needed is now the hub participant ID (the wallet should be able to infer the other properties from this). 

The current implementation is probably unclear due to our baking in hub parameters into the wallet and into the app. 

- Clears up types in the channel provider

The current implementation suffers from some union types being "infected" with with an `any`. This meant we did not have type safety. 

TODO 
- consider a similar change to  `approveBudgetAndFund` parameters


packages/channel-provider/src/types.ts
```diff
@@ -138,7 +156,10 @@ export interface ChannelProviderInterface {
   off: OffType;
   mountWalletComponent(url?: string): Promise<void>;
   enable(): Promise<void>;
-  send(request: MethodRequestType): Promise<MethodResponseType[MethodRequestType['method']]>;
+  send<M extends Method = Method>(
+    method: M,
+    params: MethodType[M]['request']['params']
+  ): Promise<MethodType[M]['response']['result']>;
```
geoknee, line 162: I think this is better.

packages/channel-client/src/channel-client.ts
```diff
@@ -140,38 +131,26 @@ export class ChannelClient implements ChannelClientInterface {
...
-      method: 'CloseAndWithdraw',
-      params: {
-        playerParticipantId: this.signingAddress as string,
-        hub: {
-          participantId: HUB.participantId,
-          signingAddress: hubAddress,
-          destination: hubOutcomeAddress
-        }
-      }
+  async closeAndWithdraw(hubParticipantId: string): Promise<DomainBudget | {}> {
```
geoknee, line 151: This change to the function signature is the main driver of this PR.

geoknee:
I agree the current API makes such an assumption, and I haven't done anything to remedy it here. 

We may wish to support multiple ledgers with the same hub at some point in the future, but I guess it is not necessary for the API to cover that now. This PR is more aiming to remove stuff that I don't think belongs in any future API.

(Recall that we can support multiple assets in one ledger channel, so it is not necessary to have one ledger channel for each. That's not to say that it wouldn't be more convenient for some other reason, though).

geoknee:
Perhaps `closeAndWithdraw` ought to accept the ledger channel Id as a single parameter?

andrewgordstewart:
> We may wish to support multiple ledgers with the same hub at some point in the future, but I guess it is not necessary for the API to cover that now. This PR is more aiming to remove stuff that I don't think belongs in any future API.

:+1:

> I agree the current API makes such an assumption, and I haven't done anything to remedy it here.

To be clear, the changed API makes _two_ assumptions:
- I never want to open more than one ledger with the hub
- I never want to use more than one `participantId`

The current API only makes the first assumption, which seems like a much bigger assumption (at least in the foreseeable future).

packages/client-api-schema/src/methods/CloseAndWithdraw.ts
```diff
@@ -1,8 +1,7 @@
-import {Participant} from '../data-types';
 import {JsonRpcRequest, JsonRpcResponse, JsonRpcError} from '../utils';
 import {ErrorCodes} from '../error-codes';
 
-export type CloseAndWithdrawParams = {hub: Participant; playerParticipantId: string};
+export type CloseAndWithdrawParams = {hubParticipantId: string};
```
geoknee, line 4: Here's the change to the JSON-RPC schema. It brings it into line with the channel client API. Advantage being, the channel client does not have to fill in the remaining fields as the request comes across. 

It also brings this schema entry close to the `getBudget` params (but with the remaining difference being `hubAddress` vs `hubParticipantId`. We may want to bring those in line with each other.

geoknee:
This PR makes some breaking changes to the API, which reflect our _current_ wallet functionality a bit more closely. It will not be the final API by any stretch. 

More important, most likely, is the tightening up of type-checking which I fixed along the way.

PR #2307, opened 2020-07-20
Add server-wallet package

andrewgordstewart:
This PR is porting over work done externally. Most of the work was thought out, roughly up to https://github.com/statechannels/monorepo/commit/7ad029135a52debd77e4571123c652a8c1d807b3)

Some of it -- the "`protocolEngine`" -- was merely sketched out to see how it worked.

Due to the nature of getting the code building, and the tests running, within this monorepo, it is awkward to split up this PR into two.


packages/server-wallet/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "module": "commonjs",
+    "target": "es2017",
+    "noImplicitAny": false,
+    "rootDir": "src",
+    "outDir": "lib",
+    "esModuleInterop": true,
+    "strict": false
```
snario, line 11: If we can avoid having to set this to `false` that would be nice.

packages/server-wallet/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "module": "commonjs",
+    "target": "es2017",
+    "noImplicitAny": false,
```
snario, line 7: If we can avoid having to set this to `false` that would be nice.

packages/server-wallet/.eslintrc.js
```diff
@@ -0,0 +1,38 @@
+// From the tslint.json we used previously
```
snario, line 1: I think you can strip this file down quite a bit. It might also be a good time to introduce some more opinionated rules like grouped and sorted imports.

.circleci/config.yml
```diff
@@ -3,15 +3,19 @@ version: 2.1
...
 defaults: &defaults
   environment:
     USE_NATIVE_SOLC: false
+    PSQL_ARGS: -p 5432 -h localhost
+    NODE_ENV: test
   working_directory: /home/circleci/project
   resource_class: medium
   docker:
     - image: geoknee/statechannels:node12.16-solc0.6.2 # Fast contract compilation with solc installed
+    - image: circleci/postgres:12-alpine-ram
```
snario, line 16: Worth adding a comment as to why we use this

packages/server-wallet/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
snario, line 1: We have this in the monorepo root already, so probably no need to deduplicate

packages/server-wallet/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+
+Copyright (c) 2019
```
snario, line 3: Wrong year

packages/server-wallet/.eslintrc.js
snario: Done

packages/server-wallet/src/type-aliases.d.ts
```diff
@@ -0,0 +1,6 @@
+declare type ParticipantId = string;
```
snario, line 1: Why do we need this file?

packages/server-wallet/src/errors.ts
```diff
@@ -0,0 +1,19 @@
+const CHANNEL_EXISTS = new Error(
+  'Attempting to open a channel that already exists -- use a different nonce'
+);
+const CHANNEL_MISSING = new Error('Channel does not exist');
+const STATE_NOT_SIGNED = new Error('State not signed by mover');
+const INVALID_TRANSITION = new Error('Invalid transition');
+const INVALID_STATE_UNKNOWN_REASON = new Error('State is not valid, but th reason is not known');
+const NOT_OUR_TURN = new Error('Not our turn to create a state');
+const VALUE_LOST = new Error('Value not preserved');
```
snario, line 9: Maybe exceptions should be `CapitalCamelCase`? Also shouldn't these values include some kind of parameters e.g., `channelId` or "error codes" associated with them (i.e., based on client-api-schema)

packages/server-wallet/src/env.ts
```diff
@@ -0,0 +1,77 @@
+// FIXME: To remove @statechannels/hub's dependency on @statechannels/devtools,
+// this file is a carbon copy of the file that exports configureEnvVariables in devtools.
+
+// const {configureEnvVariables} = require('@statechannels/devtools');
```
snario: I'm going to use `configureEnvVariables` since we don't have the issue that required this hack

snario: 80a2499

packages/server-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,238 @@
...
+  ids: Bytes32[],
+  outbox: Outgoing[] = []
+): Promise<ExecutionResult> => {
+  if (ids.length === 0) return {ids, outbox};
+
+  const [channelId, ...nextIds] = ids;
+
+  const channel = await Channel.forId(channelId, undefined);
+  const actions = await executionLoop(channel);
+  const todos = actions.filter(isInternal);
```
tomclose, line 201: I'm a bit unsure about this approach. It seems that it's pretty restrictive to have one piece of code output actions for another part to execute - e.g. it's hard to make later actions conditional on the outcome of previous actions. An alternative would be for `executionLoop` to return a list of `messages`, to add to the outbox, and `channelIds` to add to the list to process.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,238 @@
...
+    });
+
+    // TODO: Push the channelId back, once we actually trigger these actions
+    // nextIds.push(channelId);
+  }
+
+  outbox = outbox.concat(actions.filter(isOutgoing));
+
+  if (nextIds.length) {
+    return await protocolEngine(nextIds, outbox);
```
tomclose, line 234: I think it might be marginally more descriptive to write this in a loop format instead of using recursion. Something like the following seems to track the algorithm we're following pretty closely:
```ts
let outbox = [];
let channelIdQueue = [initialChannelId];
while (channelIdQueue.length > 0) {
    const [channelId, ...restOfQueue] = channelIdQueue;
    const [messages, updatedChannelIds] = executeLoop(channelId);
    outbox = [...outbox, ...messages];
    channelIdQueue = [...restOfQueue, ...updatedChannelIds];    //TODO: remove duplicates
}
```

tomclose: Also wondering if `executeLoop` should be `executeLoopStep` or `attemptToAdvanceState` or `takeNextAction` or something.

packages/server-wallet/src/wallet/index.ts
snario: Can a message be a transaction that needs to be sent to the chain?

tomclose: In what I wrote above, `message` means "protocol message" - the thing that the app receives via a `MessageQueued` and forwards on to the opponents.

I think the example of the transaction is a good example here. The code inside a workflow for submitting a transaction needs to be something like:
```ts
try {
  const transactionId = await chain.sendEthDeposit(channelId, amt, ...etc);
  await store.recordTransactionId(transactionId);
} catch e {
  await store.recordFailedTransaction(e);
}
```
This is because we now need to store things like `transactionId` in the store, in order for the protocol to be restartable.

The logic above would be difficult to write using actions, as it isn't a "fire-and-forget" scenario; the next step is dependent on the outcome of the previous step. I anticipate that a lot of the actions that we're taking will be like this.

andrewgordstewart: > I'm a bit unsure about this approach. It seems that it's pretty restrictive to have one piece of code output actions for another part to execute - e.g. it's hard to make later actions conditional on the outcome of previous actions. An alternative would be for executionLoop to return a list of messages to add to the outbox, and channelIds to add to the processing queue.

You could argue that you can make later actions conditional on earlier actions by calling `takeNextAction` again.

```
try {
  const transactionId = await chain.sendEthDeposit(channelId, amt, ...etc);
  await store.recordTransactionId(channelId, transactionId);
} catch e {
  await store.recordFailedTransaction(channelId, e); // or retry in some cases etc.
}
```

I would think of this as a single action: `SubmitTransaction`: every time we submit a transaction, we should record its hash.

It seems difficult to write a system where you can be sure it works the same way when you restart it, unless you design the system to have some properties. One property that would do this for us is:
- the `takeNextAction` function returns zero or one actions
- if it returns zero actions, we remove that channelID from the channels queue
- if it returns one actions, we do it, and call `takeNextAction` again

If we execute multiple actions within `takeNextAction`, how do we ensure that if the system crashes after executing the first, then the next time we run it, it would execute the second?

tomclose: I don't think grouping some sections of code into actions and running them in another process actually helps with that either though - it just makes the logic harder to follow. In the example above, what happens if the `SubmitTransaction` action fails in between `sendEthDeposit` and `recordTransactionId`?

In general it seems very hard to get the property that "wherever the system crashes, even mid-operation, it will be fine". I think the property we're going for is more that if you shut your wallet down during an idle period, and start it up later, and feed in any events you missed in the mean-time, then the protocols will continue to run.

I think a lot of the loop functions will be of the form "query some state, write some state, send some messages", but where the messages will need some information from back from the store after the write (e.g. waiting to see if the write was successful / including the id of a created guarantor channel etc.). It seems natural to write this in the loop in that form, rather than writing it as "query some state" -> "[action] write some state, send some messages", or as "query some state" -> "[action] write some state" -> "[next loop] read some state" -> "[action] send some messages".

packages/server-wallet/src/errors.ts
andrewgordstewart: https://github.com/statechannels/monorepo/issues/2325

packages/server-wallet/tsconfig.json
andrewgordstewart: https://github.com/statechannels/monorepo/issues/2326

packages/server-wallet/tsconfig.json
andrewgordstewart: https://github.com/statechannels/monorepo/issues/2327

andrewgordstewart:
@snario I created issues for non-trivial suggestions, to not increase the scope of this PR.

PR #2329, opened 2020-07-21
Update tsconfig.json

andrewgordstewart:
- Sets `noImplicitAny: true`
- Sets `strict: true`

Resolves https://github.com/statechannels/monorepo/issues/2327
Resolves https://github.com/statechannels/monorepo/issues/2326

### DIFF
--- a/packages/server-wallet/.eslintrc.js
+++ b/packages/server-wallet/.eslintrc.js
@@ -15,15 +15,19 @@ module.exports = {
     'plugin:import/typescript',
   ],
   rules: {
-    'import/order': [1, {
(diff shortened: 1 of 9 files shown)


packages/server-wallet/.eslintrc.js
```diff
@@ -15,15 +15,19 @@ module.exports = {
...
-      'groups': ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
-      'newlines-between': 'always'
-    }],
+    'import/order': [
+      1,
+      {
+        groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
+        'newlines-between': 'always',
+      },
+    ],
```
snario, line 24: I think it can inherit from the root

PR #2331, opened 2020-07-21
Initial protocol interface

andrewgordstewart:
This PR begins to define a framework for protocols 

It takes the opinion that protocols should be pure functions that either resolve to zero or one actions, or resolve to a "protocol error". This is something discussed with @kerzhner and @lalexgap. We decided that a pure function is an attractive option, and it is not hard to abandon that constraint and have protocols with side effects, if it turns out to be the wrong choice.

The wallet should call a protocol, and:
- if there is no action, remove the `channelId` from the queue
- if there is an action, execute it, and remove the `channelId` from the queue
- if there is an error, abort the "request cycle"

Protocols in general don't need to specify that a message should get sent. When the wallet signs a state, that state should be sent, and the app should be notified. Likewise, when an objective is created, the objective should be sent.

Future work:
- Define the `ProtocolError` type
- Add a `CreateObjective` action type


packages/server-wallet/src/fp.ts
```diff
@@ -0,0 +1,13 @@
+export type Left<E extends Error> = {readonly _tag: 'left'; error: E};
```
lalexgap, line 1: Using https://www.npmjs.com/package/either-ts might save some of this boiler plate? Although it may be simpler to just use our own implementation

packages/server-wallet/src/models/channel.ts
```diff
@@ -138,6 +139,18 @@ export class Channel extends Model implements RequiredColumns {
     });
   }
 
+  get protocolState(): ChannelData {
```
lalexgap, line 142: Will this eventually return transaction info as well?

andrewgordstewart: Yes! As well as adjudicator and asset holder info.

packages/server-wallet/src/fp.ts
andrewgordstewart: I thought about that, but figured that implementing the boilerplate was less effort than evaluating which option we should use.

Plus, using our own implementation means that we can have more strictly typed errors (rather than just the `Error` type)

packages/server-wallet/src/protocols/state.ts.ts
```diff
@@ -0,0 +1,46 @@
+import {SignedStateWithHash} from '@statechannels/wallet-core';
```
lalexgap, line 1: Should be `state.ts` instead of `state.ts.ts`

packages/server-wallet/src/fp.ts
kerzhner: I had a good experience using `fp-ts` in the `simple-hub` package FWIW.

andrewgordstewart: I am happy to try that!

packages/server-wallet/src/protocols/state.ts
```diff
@@ -0,0 +1,46 @@
...
+*/
+export type ChannelData = {
+  channelId: string;
+  myIndex: 0 | 1;
+  supported?: SignedStateWithHash;
+  latest?: SignedStateWithHash;
+  latestSignedByMe?: SignedStateWithHash;
+};
+
+export type Stage = 'Missing' | 'PrefundSetup' | 'PostfundSetup' | 'Running' | 'Final';
```
snario, line 18: "Closing" might be a stage?

snario: Also, "challenging" or "responding", unless it's  included within Running

andrewgordstewart: >"Closing" might be a stage?

A channel might be closing, but a state is either final or not final.

Likewise,
> Also, "challenging" or "responding", unless it's included within Running

A channel might be "challenging" or "responding", but that is not a property of a state. It is a function of the stage of `latest`, `latestSignedByMe`, and the adjudicator state (which is not yet recorded).

packages/server-wallet/package.json
```diff
@@ -8,6 +8,7 @@
     "@statechannels/nitro-protocol": "0.1.5",
     "@statechannels/wallet-core": "0.1.4",
     "ethers": "5.0.7",
+    "fp-ts": "^2.7.0",
```
kerzhner, line 11: Let's avoid dependency version ranges when possible.

packages/server-wallet/src/fp.ts
```diff
@@ -0,0 +1,2 @@
+export {Either, left, right} from 'fp-ts/lib/Either';
```
kerzhner, line 1: My vote is to avoid this indirection and import directly from `fp-ts` in the module that needs the import.

packages/server-wallet/src/protocols/state.ts
```diff
@@ -0,0 +1,46 @@
+import {SignedStateWithHash} from '@statechannels/wallet-core';
+
+import {Either, Option} from '../fp';
+
+import {ProtocolAction} from './actions';
+
+/*
+The ChannelState type is the data that protocols need about a given channel to decide what to do next.
+*/
+export type ChannelState = {
```
kerzhner, line 10: Would this data be computed from the database? Or stored in the database?

andrewgordstewart: Computed from the database. So, the protocols do not need to care how states are stored

packages/server-wallet/src/fp.ts
kerzhner: Resolved via da96814f9

packages/server-wallet/package.json
kerzhner: Resolved via da96814f9

Issue #2311, opened 2020-07-20
Server Wallet API


andrewgordstewart:
The draft wallet interface is sketched out [here](https://github.com/statechannels/monorepo/blob/ags/server-wallet-take-2/packages/server-wallet/src/wallet/index.ts#L41-L54).

So far, `createChannel` and `pushMessage` have been partially implemented. That is, they are implemented enough that:
- you can create a new channel
- you can save channels & states in a postgresql database

A "`protocolEngine`" function has been defined which
- runs protocols, which should be pure functions that declare actions
- executes the actions declared by the protocols

Some of us share the opinion that the API calls should never reject. Instead, the wallet class should `try / catch` any errors that are thrown, and resolve with either a result or a typed error.

This task involves:
- [ ] Decide if the API should return typed errors. If so, enumerate what errors you can think of
- [x] Implement a "proper" `protocolEngine` (and think about the naming)
- [x] Implement the `updateChannel` method

andrewgordstewart:
stale

PR #2332, opened 2020-07-21
Direct Funding Protocol

lalexgap:
Adds a basic direct funding protocol and some simple tests.

The protocol assumes that the prefund setup will always be signed before running (based on discussions with @andrewgordstewart )


packages/server-wallet/src/protocols/direct-funding.ts
```diff
@@ -1,6 +1,88 @@
...
+} from '@statechannels/wallet-core';
 
-import {Protocol, ChannelState} from './state';
+import {match} from '../match';
 
-export const protocol: Protocol<ChannelState> = (_ps: ChannelState) => Promise.resolve(right(none));
+import {Protocol, ChannelState, stage, ProtocolResult} from './state';
+import {protocol as fundingProtocol} from './depositing';
+
+export type DirectFundingState = ChannelState & {minimalOutcome: SimpleAllocation};
```
lalexgap: This assumes that the  `minimalOutcome` would come from the application request or an Objective shared from the opponent.

packages/server-wallet/src/protocols/actions.ts
```diff
@@ -9,7 +9,7 @@ Actions that protocols can declare.
 
 type UpdateOpts = Partial<StateVariables>;
 type UpdateChannel = {type: 'UpdateChannel'; channelId: Bytes32} & UpdateOpts;
-type SignState = {type: 'SignState'; channelId: Bytes32; hash: Bytes32};
+export type SignState = {type: 'SignState'; channelId: Bytes32} & UpdateOpts;
```
lalexgap, line 12: I'm assuming that `SignState` will
- Sign the state
- Update the store with the state
- Communicate the state to opponents.

I'm not sure we need an `UpdateChannel` and `SignState`?

packages/server-wallet/src/protocols/state.ts
```diff
@@ -13,8 +15,8 @@ export type ChannelState = {
   supported?: SignedStateWithHash;
   latest?: SignedStateWithHash;
   latestSignedByMe?: SignedStateWithHash;
+  funding: Record<Address, Uint256>;
```
lalexgap, line 18: This would be the funds deposited on chain, indexed by `AssetHolderAddress`

packages/server-wallet/src/protocols/actions.ts
andrewgordstewart: It might be a worthwhile separation:
- `UpdateChannel` will create a _new_ state, and sign it.
- `SignState` will sign an _existing_ state

packages/server-wallet/src/protocols/direct-funding.ts
andrewgordstewart: Should it be a `MixedAllocation`? IE. should we prepare for multi-token channels?

packages/server-wallet/src/protocols/__test__/direct-funding.test.ts
```diff
@@ -1,12 +1,18 @@
...
 const c = channelWithVars();
-
+const protocolState = {
+  ...c.protocolState,
+  minimalOutcome: {
+    type: 'SimpleAllocation' as 'SimpleAllocation',
+    allocationItems: [],
+    assetHolderAddress: '0x0',
+  },
+};
```
andrewgordstewart, line 12: I'd like to try and keep the test data clean. So, this can be created in a fixture file for `direct-funding-state`.

packages/server-wallet/src/protocols/depositing.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import {Protocol, ChannelState, ProtocolResult} from './state';
+
+export type ProtocolState = ChannelState & {
+  assetHolderAddress: Address;
+  amountToDeposit: Uint256;
+  safeToDepositAmount: Uint256;
+  fundedAt: Uint256;
+  depositTransactionId: string | undefined;
+};
```
andrewgordstewart, line 16: I'd rather if the `Depositing` protocol state looked like
```
{
  app: ChannelState,
  assetHolderAddress: Address;
  amountToDeposit: Uint256;
  safeToDepositAmount: Uint256;
  fundedAt: Uint256;
  depositTransactionId: string | undefined;
}
```

We have been burned a few times by defining flat data types like this, and then using the spread operator to create them.

packages/server-wallet/src/match.ts
```diff
@@ -0,0 +1,27 @@
...
+  handlers: Record<T, Protocol<PS>>
+) => (ps: PS): ProtocolResult => handlers[partitioner(ps)](ps);
+
+export const match = <PS extends ChannelState, T extends string>(
+  partitioner: Partitioner<T, PS>,
+  handlers: (Partial<Record<T, Protocol<PS>>> & {Default: Protocol<PS>}) | Record<T, Protocol<PS>>
+) => (ps: PS): ProtocolResult => {
+  const handler = handlers[partitioner(ps)];
+  if (handler) {
+    return handler(ps);
```
andrewgordstewart, line 19: It'd be nice to allow the handlers to be sync or async functions. With this change, then `match` will still return a `Promise`
```suggestion
    return await handler(ps);
```

packages/server-wallet/src/protocols/actions.ts
lalexgap: > It might be a worthwhile separation:
> 
> * `UpdateChannel` will create a _new_ state, and sign it.
> * `SignState` will sign an _existing_ state

How would `SignState` specify the existing state? I don't think just signing whatever the latest state happens to be would be safe.

If we plan to specify them by using `Partial<StateVariables>` then I'm not sure if the distinction between new or existing states matter?  We just want a signed state with a certain turnnumber, outcome, etc.. It doesn't matter if we're creating it from scratch or supporting an existing state (I think?)

packages/server-wallet/src/protocols/direct-funding.ts
lalexgap: > Could we get a direct funding protocol checked in that is partially implemented? EG. it can ignore depositing, and just sign pre- and post-fund-setup states?

I think it would be pretty easy to get a partially implemented version checked in. However it would be dependent on implementation of `Action` and the basic runloop ( would that be in scope of #2311 ?)

packages/server-wallet/src/protocols/depositing.ts
lalexgap: This might be a good pattern to follow across all our `ProtocolState`

packages/server-wallet/src/match.ts
lalexgap: With the change it looks like `match` is not returning a `Promise` which is what I would expect?

lalexgap:
> Could we get a direct funding protocol checked in that is partially implemented? EG. it can ignore depositing, and just sign pre- and post-fund-setup states?

I think it would be pretty easy to get a partially implemented version checked in. I think the main thing missing right now is tests. However it would be dependent on implementation of Action and the basic runloop ( would that be in scope of #2311 ?) to actually work.

andrewgordstewart:
>I think it would be pretty easy to get a partially implemented version checked in. I think the main thing missing right now is tests. However it would be dependent on implementation of Action and the basic runloop ( would that be in scope of #2311 ?) to actually work.

Tests can just assert on the returned value! Eg. if the protocol is expected to return a given action, for a given "protocol state", then we can expect that action.

We can test the execution of the actions separately.

packages/server-wallet/src/protocols/actions.ts
andrewgordstewart: > > It might be a worthwhile separation:
> > 
> > * `UpdateChannel` will create a _new_ state, and sign it.
> > * `SignState` will sign an _existing_ state
> 
> How would `SignState` specify the existing state? I don't think just signing whatever the latest state happens to be would be safe.

The current `Channel` model implements [this](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/models/channel.ts#L155), which is a change from the `ChannelStoreEntry`.

> If we plan to specify them by using `Partial<StateVariables>` then I'm not sure if the distinction between new or existing states matter? We just want a signed state with a certain turnnumber, outcome, etc.. It doesn't matter if we're creating it from scratch or supporting an existing state (I think?)

I'm actually thinking it's beneficial to specify _all_ `StateVariables` in the `UpdateChannel` action, to help deal with race conditions. In that case, there's no difference between `SignState` and `UpdateChannel` -- they can both be `SignState`, with some specific `StateVariables`.

packages/server-wallet/src/match.ts
andrewgordstewart: https://github.com/statechannels/statechannels/commit/3801967b8cec16dfd2f87eb62c118c0ea6ad6e29

Maybe we should do something like that?

lalexgap:
> > I think it would be pretty easy to get a partially implemented version checked in. I think the main thing missing right now is tests. However it would be dependent on implementation of Action and the basic runloop ( would that be in scope of #2311 ?) to actually work.
> 
> Tests can just assert on the returned value! Eg. if the protocol is expected to return a given action, for a given "protocol state", then we can expect that action.
> 
> We can test the execution of the actions separately.

Yup! Sorry I didn't mean to imply that actions /runloop are needed for testing, they're just needed to get things actually working (ie: with an integration test)

packages/server-wallet/src/protocols/state.ts
```diff
@@ -11,10 +13,10 @@ export type ChannelState = {
   channelId: string;
   myIndex: 0 | 1;
   supported?: SignedStateWithHash;
-  latest?: SignedStateWithHash;
+  latest: SignedStateWithHash;
   latestSignedByMe?: SignedStateWithHash;
+  funding: Record<Address, Uint256>;
```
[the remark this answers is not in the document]
lalexgap: I like this idea, it would make things a lot cleaner.

packages/server-wallet/src/protocols/__test__/direct-funding.test.ts
```diff
@@ -4,7 +4,7 @@ import {some, none} from 'fp-ts/lib/Option';
 
 import {protocol} from '../direct-funding';
 import {alice} from '../../wallet/__test__/fixtures/signingWallets';
-import {Address, Uint256} from '../../type-aliases';
+import {Uint256} from '../../type-aliases';
```
andrewgordstewart, line 7: Hmm, I think this should use the `Uint256` from `wallet-core`. Casting to that seems fine in tests, instead of `BN.from(5)`, but we want to make sure that things that are a `Uint256` have passed some kind of a data validator, which is the point of the "branded" `Uint256` type

packages/server-wallet/src/protocols/__test__/application.test.ts
```diff
@@ -0,0 +1,43 @@
+import {simpleEthAllocation, makeDestination, BN} from '@statechannels/wallet-core';
+import matchers from '@pacote/jest-either';
+import {some, none} from 'fp-ts/lib/Option';
+
+import {protocol} from '../application';
+import {alice} from '../../wallet/__test__/fixtures/signingWallets';
+import {Uint256} from '../../type-aliases';
+
+import {
+  directFundingProtocolState,
```
andrewgordstewart, line 9: Should be renamed

packages/server-wallet/src/protocols/__test__/direct-funding.test.ts
```diff
@@ -2,14 +2,14 @@ import {simpleEthAllocation, makeDestination, BN} from '@statechannels/wallet-co
 import matchers from '@pacote/jest-either';
```
andrewgordstewart, line 2: This file should be renamed?

lalexgap: Oops, accidentally added a duplicate test file somehow when renaming

packages/server-wallet/src/protocols/application.ts
```diff
@@ -0,0 +1,57 @@
...
+  'Not Funded': () => right(none), // TODO: Start depositing
+});
+
+const prefundSetupSigned = match(ps => stage(ps.app.supported), {
+  PrefundSetup: prefundSetupStateSupported,
+  Default: () => right(none), // Still waiting for the opponent to sign
+});
+
+export const protocol: Protocol<ProtocolState> = match(ps => stage(ps.app.latestSignedByMe), {
+  Missing: () => left(new Error(`The direct funding protocol requires a signed prefund setup`)),
```
andrewgordstewart, line 54:
```suggestion
  Missing: () => left(new Error(`The application protocol requires a signed prefund setup`)),
```

PR #2334, opened 2020-07-22
Run syncpack and yarn

kerzhner:
1. `npx  syncpack fix-mismatches`
2. Then manually downgrade packages that use ethers v4 due to https://github.com/statechannels/monorepo/issues/2333.
3. `npx syncpack format`
4. `npx syncpack set-semver-ranges`
5. `yarn`


kerzhner:
Added a `syncpack` script. We can even run this script as part of CI or a commit hook if we would like.

PR #2335, opened 2020-07-22
[doc] Reorganize sidebar, rewrite overview page and rename sections

geoknee:
I'm following some of @snario's suggestions.

The idea is to:

- define our target audience better
- expand the scope of the documentation from nitro-protocol to the whole statechannels stack

Once this stuff is settled, it should be easier to see where the holes are, that need to be filled in before we push this out / publicise it.


geoknee:
Preview https://5f181ca6e8517e019d88d909--nitro-protocol.netlify.app

geoknee:
Suggest that this changeset be accompanied by 

- URL change to docs.statechannels.org
- link from statechannels.org is named 'docs' instead of 'protocol docs'

packages/nitro-protocol/docs/app-devs/intro.md
```diff
@@ -1,8 +1,8 @@
 ---
-id: dapp-devs-intro
+id: app-devs-intro
 title: Introduction
 ---
 
-As a Dapp dev you only need to author and deploy a single smart contract. These docs will help you do that.
+As a application developer you will only need to author and deploy a single smart contract. These docs will help you do that.
```
snario, line 6:
```suggestion
As an application developer you will only need to author and deploy a single smart contract. These docs will help you do that.
```

PR #2341, opened 2020-07-22
Add a simple HTTP Client->Server end-to-end test harness

snario:
This PR adds an end to end test harness for the use case of a `server-wallet` communicating to another `server-wallet` over HTTP request response cycles only. For simplicity, the parties involved are `ping` and `pong`. See the [README](https://github.com/statechannels/statechannels/blob/8a10c06d5941f144bd53099d86f71f2da4337a44/packages/server-wallet/e2e-test/README.md) for a description of how it works.

One wallet runs inside the Jest test process and another is spawned and run within a bespoke process running an Express server. Both re-use the same PostgreSQL service, but using different databases.


snario:
Hoping I can re-use `ChannelClient` by pointing it against the HTTP server and writing an `HTTPChannelProvider`...

packages/server-wallet/test/e2e.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    env: {...process.env, NODE_ENV: 'test', PORT: port},
+  });
+
+  server.stdout.on('data', data => console.log(data.toString()));
+  server.stderr.on('data', data => console.error({error: data.toString()}, `Server threw error`));
+
+  return server;
+}
+
+const aliceHttp = startServerWallet('5001');
```
andrewgordstewart, line 14: Minor point: I don't know that we want Alice to start an HTTP server.

We definitely want to build a virtual funding protocol that works given the constraint that
- Alice can spontaneously message Bob
- Bob can only respond to messages from Alice
where this constraint comes from the fact that Bob is serving HTTP requests, and Alice is making HTTP requests.

snario: Okay; so it would be overkill for this test, and messaged can be provided to Alice directly by calling Wallet.pushMessage then

snario:
Going to add the e2e test separately...

andrewgordstewart:
> Hoping I can re-use `ChannelClient` by pointing it against the HTTP server and writing an `HTTPChannelProvider`...

What is the value of defining an `HTTPChannelProvider`, as opposed to directly using the server-wallet API?

snario:
> > Hoping I can re-use `ChannelClient` by pointing it against the HTTP server and writing an `HTTPChannelProvider`...
> 
> What is the value of defining an `HTTPChannelProvider`, as opposed to directly using the server-wallet API?

For tests which test the `Wallet` they can use the `Wallet` API, but for tests that test the HTTP interface, they should send API requests over HTTP. This is the case with the end-to-end test where one client is sending HTTP requests to the server.

andrewgordstewart:
> > > Hoping I can re-use `ChannelClient` by pointing it against the HTTP server and writing an `HTTPChannelProvider`...
> > 
> > 
> > What is the value of defining an `HTTPChannelProvider`, as opposed to directly using the server-wallet API?
> 
> For tests which test the `Wallet` they can use the `Wallet` API, but for tests that test the HTTP interface, they should send API requests over HTTP. This is the case with the end-to-end test where one client is sending HTTP requests to the server.

I'm not sure I follow. I didn't think it was the case that Bob would send wallet API requests to Bob's wallet via HTTP.

Alice will send _messages_ to Bob via HTTP, but Alice should not have an `HTTPChannelProvider` that connects Alice directly to Bob's wallet!

snario:
No, you're right about that, let me adjust the PR

packages/server-wallet/e2e-test/pong/controller.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  public async acceptMessageAndReturnReplies(message: Message): Promise<Message> {
+    const {
+      recipient: to,
+      sender: from,
+      data: {signedStates},
+    } = message;
+
+    // FIXME: server-wallet is using wallet-core, not wire-format for
+    // types of messages between parties. e2e-test uses wire-format
```
andrewgordstewart, line 17: So, the wallet should call `serializeMessage` on outgoing messages, right?

Make an issue out of this comment?

packages/server-wallet/e2e-test/pong/controller.ts
```diff
@@ -0,0 +1,45 @@
+import {Message} from '@statechannels/wire-format';
+import {SignedState} from '@statechannels/wallet-core';
+
+import {Wallet} from '../../src/wallet';
+
+export default class PongController {
+  private readonly wallet: Wallet = new Wallet();
+
+  public async acceptMessageAndReturnReplies(message: Message): Promise<Message> {
```
andrewgordstewart, line 9: Either in this PR or shortly after, can we have this so that `pong`

1. calls `pushMessage`
2. looks at the channel result. if its turn number is at least 4, it calls `updateChannel`
3. it then responds with the new outgoing messages
?

This
A. doesn't seem like a lot of work
B. should work with the existing server-wallet
C. would add a lot of value in terms of knowing that "things are working end-to-end"
   - In particular, this is essentially the stage 0 goal!

snario: I'm going to do it right after, been waiting on `ChannelUpdated` being emitted by the call to `pushMessage`

PR #2342, opened 2020-07-22
Narrower naming of iframe-channel-provider

snario:


snario:
> Q: Is this still valid https://5f1955c6fc7cfffc1858c11e--nitro-protocol.netlify.app/docs/next/app-devs/quick-start-dapp#enabling-the-channel-provider, namely does requiring the package attach the iframe-channel-provider instance as window.channelProvider? or window.iframeChannelProvider ?

Still valid!

PR #2350, opened 2020-07-23
[docs] Move to its own package

geoknee:
This PR 

- moves the doc website out from under nitro-protocol. 
- establishes a new workflow for our Typescript APIs: as a part of their `prepare` scripts, they run `api-extractor` and copy the JSON output to the new `docs-website` package. This package then converts the JSON into lovely markdown which lives in a new section of the website: 
Screenshot 2020-07-28 at 15 08 34
- adds `iframe-channel-provider` package to this workflow
Screenshot 2020-07-28 at 15 08 40

In future PRs: 
-  we can add `nitro-protocol`, `client-api-schema` and `wire-protocol`. This will  ensure all relevant types are exposed in the documentation.
- We can fill in the TSDoc annotations to generate more verbose and useful docs.


geoknee:
Suggestion for the reviewer: filter out  `.md` files (there are lots of them, but they are autogenerated).

geoknee:
Preview here https://5f2031a3a8da0cca7639efbe--nitro-protocol.netlify.app/docs/next/typescript-api/

geoknee:
One particularly nice thing about this approach is that `api-documenter` can document several package APIs holistically. So we get hyperlinks from one package to another within the docs. (Say when the ChannelClient class has a property of type ChannelProviderInterface, you can just click through to that).

PR #2358, opened 2020-07-24
[WIP] updateChannel handler

kerzhner:
Would love feedback on:
- Functional programming approach to error handlers.
- Open questions listed.

### DIFF
--- a/packages/server-wallet/src/wallet/handlers.ts
+++ b/packages/server-wallet/src/wallet/handlers.ts
@@ -1,0 +1,33 @@
+import {Either, left, right, chain, ap, map} from 'fp-ts/lib/Either';
+import {Option} from 'fp-ts/lib/Option';
+import {SignedStateWithHash} from '@statechannels/wallet-core';
+import {pipe} from 'fp-ts/lib/function';
+
+import {ProtocolAction} from '../protocols/actions';
+import {ChannelState} from '../protocols/state';
+
+import {UpdateChannelParams} from '.';
+
+type HandlerResult = Either<Error, Option<ProtocolAction>>;
+
+// Check that the channel is:
+// - Is funded.
+// - No challenge exists.
+// - In the running state.
+// - It is my turn to update
+
+// Then:
+// - Create and sign new state.
+// - Check if any other actions need to be taken.
+
+export function updateChannel(
+  args: UpdateChannelParams,
+  channelState: ChannelState
+): HandlerResult {
+  // TODO: check if the channel is funded and that no challenge exists once that data is part of the ChannelState
+  const latestIfExists = (cs: ChannelState): Either<Error, SignedStateWithHash> =>
+    cs.latest ? right(cs.latest) : left(new Error('updateChannel: must have latest state'));
+  const runningTurnNumber = (ss: SignedStateWithHash): Either<Error, SignedStateWithHash> =>
+    ss.turnNum < 3 ? left(new Error('updateChannel: channel must be in running state')) : right(ss);
+  pipe(channelState, latestIfExists, chain(runningTurnNumber));
+}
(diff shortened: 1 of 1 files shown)


packages/server-wallet/src/handlers/fixtures/updateChannel.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const defaultChannelState: ChannelState = {
+  channelId: '0x1234',
+  myIndex: 0,
+  supported: stateWithHashSignedBy()({turnNum: 3}),
+  latest: stateWithHashSignedBy()({turnNum: 3}),
+  funding: (): Uint256 => '0x0',
+};
+
+export const channelStateFixture = fixture(defaultChannelState);
```
andrewgordstewart, line 29: This probably belongs in its own `fixtures/channel-state` file?

packages/server-wallet/src/handlers/updateChannel.ts
```diff
@@ -0,0 +1,72 @@
...
+): HandlerResult {
+  // todo: check if the channel is funded and that no challenge exists once that data is part of the ChannelState
+  const ensureSupportedStateExists = (
+    cs: ChannelState
+  ): Either<UpdateChannelError, ChannelStateWithSupported> =>
+    hasSupportedState(cs) ? right(cs) : left(new UpdateChannelError(Errors.invalidLatestState));
+  const hasRunningTurnNumber: ValidateState = cs =>
+    cs.supported.turnNum < 3 ? left(new UpdateChannelError(Errors.notInRunningStage)) : right(cs);
+  const isMyTurn: ValidateState = ss =>
+    ss.supported.turnNum % channelState.myIndex
```
tomclose, line 48: This doesn't look correct to me. I think it should be
```ts
 ss.supported.turnNum % ss.supported.participants.length === ss.myIndex;
```

kerzhner: 👍 The handler signs the state that follows the supporting state. Maybe the correct logic is:
`ss.supported.turnNum + 1 % ss.supported.participants.length === ss.myIndex;`

packages/server-wallet/src/handlers/updateChannel.ts
```diff
@@ -0,0 +1,84 @@
...
+import {SignedStateWithHash, StateVariables, Outcome} from '@statechannels/wallet-core';
+import {pipe} from 'fp-ts/lib/function';
+import {ChannelId} from '@statechannels/client-api-schema';
+import {curry} from 'lodash';
+
+import {ProtocolAction, SignState} from '../protocols/actions';
+import {ChannelState} from '../protocols/state';
+
+type ChannelStateWithSupported = ChannelState & {
+  supported: SignedStateWithHash;
```
andrewgordstewart, line 11: We might as well make `latestSignedByMe` present

```suggestion
  supported: SignedStateWithHash;
  latestSignedByMe: SignedStateWithHash;
```

packages/server-wallet/src/handlers/updateChannel.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+const incrementTurnNumber = (
+  args: UpdateChannelHandlerParams,
+  cs: ChannelStateWithSupported
+): StateVariables => ({
+  ...args,
+  turnNum: cs.supported.turnNum + 1,
+  isFinal: false,
+});
```
andrewgordstewart, line 61: If we defined `incrementTurnNumber` this way, we wouldn't have to import `curry`.
```suggestion
const incrementTurnNumber = (args: UpdateChannelHandlerParams) => (
  cs: ChannelStateWithSupported
): StateVariables => ({
  ...args,
  turnNum: cs.supported.turnNum + 1,
  isFinal: false,
});
```

packages/server-wallet/src/handlers/updateChannel.ts
```diff
@@ -0,0 +1,83 @@
+import {Either, left, right, chain, map} from 'fp-ts/lib/Either';
```
lalexgap, line 1: Minor but most of our files names are using kebab-casing instead of camel casing, so we should probably rename this to `update-channel`?

andrewgordstewart: Let's make an issue out of this, and make a lint rule to enforce it?

PR #2360, opened 2020-07-26
Next actions

andrewgordstewart:
This PR connects the wallet API calls to the "application" protocol, such that a `pushMessage` call will
- store states contained in the message
- take some actions via calling `takeActions`
- return a `ChannelResult` and a `MessageQueued` notification

## Open questions
There are still some open questions about the "request cycle" approach.

1. What type should the API return? Given that we will `takeNextActions` after handling the initial API call, we might end up wanting to return multiple `ChannelResults` and multiple notifications. There could potentially be multiple `ChannelResults` for a single channel. In that case, which one do we return? Or do we return all of them? See (3) for an explicit example.

2. Are `ChannelResults` "different" from other notifications? As in, should the wallet just return a list of notifications, some of which might be a `ChannelUpdated` notice?

3. Should API calls for specific channels always return just a single `ChannelResult`? The app API methods here are: `joinChannel`, `createChannel`, `closeChannel`, and `updateChannel`. I can't think of a scenario where they should, but the `requestCycle` pattern makes this specific scenario a bit complicated:
  a. Alice calls `pushMessage` with a message containing a signed state for channel `0xabc`
  b. Her wallet stores the states, then calls `takeActions(['0xabc'])`
  c. `takeActions` might cause some _other_ channel `'0x123'` to be updated
  In this scenario, the wallet needs to decide whether to return a `ChannelResult` to Alice for channel `'0x123'`. In this PR, the wallet informs Alice of all channel results (which currently happen to only be app channels).

4. Should the wallet bother merging messages? Or should that be left up to the client? EG. the wallet might, in a single request cycle, want to sync two states for two different channels, to a single recipient. The wallet should probably merge those into a single `MessageQueued` notification.

## Technical work
The `takeActions` loop is trying to be functional, but the combination of `Option.fold`, `Either.fold` and awaiting promises is a bit nasty -- the result is that I `try/catch` [here](https://github.com/statechannels/statechannels/compare/ags/handle-actions?expand=1#diff-3f1750fddf87d507315aa4cb72ae8a20R216). Perhaps in the short term, I should abandon the functional approach and handle the `nextAction` more imperatively.


snario:
Regarding (4), it seems like the wallet _should_ do the merging. I was anticipating there would be some kind of "message builder" pattern which is perhaps an aggregation in the context of a single request-response cycle of the wallet. So, for multiple actions this object would be passed into each one, like an aggregator in a fold function.

packages/server-wallet/src/models/channel.ts
```diff
@@ -96,11 +103,15 @@ export class Channel extends Model implements RequiredColumns {
 
   static jsonAttributes = ['vars', 'participants'];
 
-  static forId(channelId: Bytes32, tx?: Transaction): Promise<Channel> {
-    return Channel.query(tx)
+  static async forId(channelId: Bytes32, tx: Transaction | undefined): Promise<Channel> {
```
snario, line 106: Could this method instead return an `Promise<Either<ChannelError, Channel>>`? Should it?

packages/server-wallet/src/models/channel.ts
```diff
@@ -152,6 +163,24 @@ export class Channel extends Model implements RequiredColumns {
     };
   }
 
+  get channelResult(): ChannelResult {
+    const {channelId, participants, appDefinition, supported} = this;
+    if (!supported) throw 'not supported';
```
snario, line 168: We should introduce a typed error for this

packages/server-wallet/src/models/channel.ts
```diff
@@ -152,6 +163,24 @@ export class Channel extends Model implements RequiredColumns {
...
+
+    const allocations = serializeAllocation(checkThat(outcome, isAllocation));
+
+    return {
+      channelId,
+      participants,
+      appData,
+      allocations,
+      appDefinition,
+      status: 'funding', // FIXME
```
snario, line 179: Do we have a specification written somewhere for what defines these status types? I actually forget what `funding` means but I'm guessing it's that this channel is not yet funded, but there exists some objective the wallet is carrying out, whether direct or virtual, to fund it.

packages/server-wallet/src/models/channel.ts
```diff
@@ -162,7 +191,7 @@ export class Channel extends Model implements RequiredColumns {
     }
   }
 
-  signAndAdd(stateVars: StateVariables, privateKey: string): SignedState {
+  signAndAdd(stateVars: StateVariables): SyncState {
```
snario, line 194: It seems odd to me that a model should need to return a data type related to messaging. Maybe it should still return a `SignedState` and the caller should handle deciding what messages should be queued up?

packages/server-wallet/src/protocols/__test__/application.test.ts
```diff
@@ -1,17 +1,15 @@
-import {simpleEthAllocation, makeDestination, BN, Uint256} from '@statechannels/wallet-core';
+import {simpleEthAllocation, BN, Uint256} from '@statechannels/wallet-core';
 import matchers from '@pacote/jest-either';
 import {some, none} from 'fp-ts/lib/Option';
 
 import {protocol} from '../application';
-import {alice} from '../../wallet/__test__/fixtures/signingWallets';
+import {alice} from '../../wallet/__test__/fixtures/participants';
```
snario, line 6: Maybe we should create some kind of `shared` folder for re-used test fixtures across unit tests, seems odd for one test file, nested under `protocols` to reach for fixtures from another context designed for `wallet`.

packages/server-wallet/src/protocols/actions.ts
```diff
@@ -7,9 +7,9 @@ import {Bytes32} from '../type-aliases';
 Actions that protocols can declare.
 */
 
+export type Notice = Omit<Notification, 'jsonrpc'>;
```
snario, line 10: What's the point of the omission?

packages/server-wallet/src/protocols/actions.ts
```diff
@@ -7,9 +7,9 @@ import {Bytes32} from '../type-aliases';
 Actions that protocols can declare.
```
snario, line 7: In general the types in this file seem important for the wallet and not specific to "protocol actions", maybe there should be a `src/types.ts` file which includes only these types to make it easier to find things.

packages/server-wallet/src/protocols/application.ts
```diff
@@ -29,7 +29,7 @@ const getFundingStatus = (ps: ProtocolState): FundingStatus => {
   const allocation = checkThat(ps.app.supported?.outcome, isSimpleAllocation);
 
   const currentFunding = ps.app.funding(allocation.assetHolderAddress);
-  const targetFunding = allocation.allocationItems.map(a => a.amount).reduce(BN.add);
+  const targetFunding = allocation.allocationItems.map(a => a.amount).reduce(BN.add, BN.from(0));
```
snario, line 32: Should we use `BN.from(0)` or `Zero`? Minor nit but we should be consistent

packages/server-wallet/src/wallet/__test__/fixtures/states.ts
```diff
@@ -34,28 +34,30 @@ const defaultState: State = {
...
 // TODO: Persist these signatures between tests
-const signatureCache: Record<string, string> = {};
-const _signState = (s: State, pk: string): string => {
-  const key = `${pk}-${hashState(s)}`;
-  return (signatureCache[key] = signatureCache[key] || signState(s, pk));
+const signatureCache: Record<string, SignatureEntry> = {};
+const _signState = (s: State, sw: SigningWallet): SignatureEntry => {
+  const key = `${sw.privateKey}-${hashState(s)}`;
+  return (signatureCache[key] = signatureCache[key] || sw.signState(s));
 };
```
snario, line 41: If we're adding some caching, should we explore using `redis` instead of implementing our own strategy?

snario: Perhaps if this is just for local testing it doesn't matter.

packages/server-wallet/src/wallet/__test__/sign-state.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import knex from '../../db/connection';
+import {Channel} from '../../models/channel';
+
+import {stateWithHashSignedBy} from './fixtures/states';
+import {bob} from './fixtures/signingWallets';
+
+let tx: Objection.Transaction;
+beforeEach(async () => {
+  // Make sure alice's PK is in the DB
+  await seed(knex);
```
snario, line 16: One thing I noticed is that since we re-use these fixtures and clear the DB on every test, it makes it hard to parallelize the jest tests. Do we want to parallelize the jest tests? Could we do something so that instead of running this global seed there is some per-test unique seed? Out of scope for the PR

packages/server-wallet/src/wallet/index.ts
```diff
@@ -165,34 +158,70 @@ export class Wallet implements WalletInterface {
...
+    };
+
+    const handleAction = async (action: ProtocolAction): Promise<any> => {
+      switch (action.type) {
+        case 'SignState': {
+          const notices = await Store.signState(action, tx);
+          notices.map(n => outbox.push(n.notice));
+
+          const {channelResult} = await Channel.forId(action.channelId, tx);
+          channelResults.push(channelResult);
```
snario, line 198: This just generally feel unintuitive. Maybe I'm thinking about it wrong, but it seems like there should be a separation of interfaces between the Wallet, the Store, and the Channel model. Like
```
Wallet (update store, decides what messages / notifications should be)
|__> Store (makes DB-specific optimizations and interacts with model e.g., transactions)
       |__> Channel (implements object in DB-specific domain e.g., PostgreSQL)
```
It seems here we have
```
Wallet (update store, re-uses what store tells it to use as notifications, queries Models directly for response)
|__> Store (makes DB-specific optimizations and interacts with model, relays notifications from model)
       |__> Channel (implements object in DB-specific domain and compiles notifications to be sent)
|__> Channel
```

packages/server-wallet/src/models/channel.ts
andrewgordstewart: I'm unaware of one, which is why this is hard coded. There definitely needs to be one.

packages/server-wallet/src/models/channel.ts
andrewgordstewart: Fair point. This method can be extracted into a pure function that transforms a signed state into a notification. That probably makes more sense

packages/server-wallet/src/wallet/__test__/fixtures/states.ts
andrewgordstewart: It was written at the time for local testing. I think we should even persist them between test runs -- it would mean a test file could run in 3s instead of 10s, which seems like a big win for developer experience

packages/server-wallet/src/wallet/__test__/sign-state.test.ts
andrewgordstewart: One large advantage of writing protocols and "handlers" as pure functions that decide what to do, but don't do it, is that it's trivial to run those tests in parallel.

So, we can have one test job of these protocol level decisions, which test the complicated logic, and run in parallel.

We can have another test job that tests the integration of those protocol decisions into our database store, which runs in serial.

If we are unhappy running those tests in serial, we can explore some techniques for running them in parallel, like executing each test case inside a transaction.

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: Yup. The store was introduced near the end of this piece of work. I believe that the `Channel` model should live entirely behind a store interface. This decouples the wallet from the database.

packages/server-wallet/src/protocols/__test__/application.test.ts
andrewgordstewart: My reasoning at the time is that fixtures should be defined next to the data structure they return. That approach seems reasonable, and so does a centralized approach that defines all the fixtures in one place.

packages/server-wallet/src/protocols/actions.ts
andrewgordstewart: The server wallet isn't really a json-rpc server, so I'd flip the question around on its head: what's the point of including it? 😛

packages/server-wallet/src/protocols/actions.ts
andrewgordstewart: Yup, definitely worth thinking about where these actions should be defined.

My reasoning is that these protocols are actually code that should be shared between the server wallet and the browser wallet. So, the actions that protocols declare should be separate from the server wallet's own types.

packages/server-wallet/src/models/channel.ts
andrewgordstewart: Yes, and I think so!

packages/server-wallet/src/protocols/actions.ts
snario: Makes sense, we should probably export & import something other than `Notification`

packages/server-wallet/src/protocols/__test__/application.test.ts
snario: I think my point then becomes that we should have a shared folder for data types re-used in multiple contexts

snario: For example 
```
.
├── protocols
├── shared
│   ├── __test__
│   │   └── fixtures
│   └── types.ts
├── store
│   └── models
└── wallet

```

Issue #2363, opened 2020-07-27
This probably belongs in its own `fixtures/channel-state` file?


kerzhner:
This probably belongs in its own `fixtures/channel-state` file?

_Originally posted by @andrewgordstewart in https://github.com/statechannels/statechannels/pull/2358#discussion_r460470786_

PR #2368, opened 2020-07-27
Implement `updateChannel`

andrewgordstewart:
Also, resolves https://github.com/statechannels/statechannels/issues/2363


packages/server-wallet/src/wallet/index.ts
```diff
@@ -99,8 +100,25 @@ export class Wallet implements WalletInterface {
...
-  async updateChannel(_args: UpdateChannelParams): Result {
-    throw 'Unimplemented';
+  async updateChannel({channelId, appData, allocations}: UpdateChannelParams): Result {
+    return Channel.transaction(async tx => {
+      const channel = await Store.getChannel(channelId, undefined);
+
+      const outcome = deserializeAllocations(allocations);
+      const decision = updateChannel({channelId, appData, outcome}, channel);
+
+      switch (decision._tag) {
```
kerzhner, line 110: https://gcanti.github.io/fp-ts/modules/Either.ts.html#isleft

Would this be better than inspecting an internal property?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -99,8 +100,21 @@ export class Wallet implements WalletInterface {
...
+
+      const decision = updateChannel({channelId, appData, outcome}, channel);
+      if (Either.isLeft(decision)) {
+        throw decision.left;
+      } else {
+        const {outgoing, channelResult} = await Store.signState(decision.right, tx);
+
+        return {outbox: outgoing.map(n => n.notice), channelResults: [channelResult]};
+      }
+    });
```
snario, line 117: Is there some clean way of encapsulating the bit that says "ensure this stuff happens in a single transaction on the 'channels' table" inside the `Store`? This is a similar point as before, where it seems like the locking mechanism of the database in question shouldn't need to be known to the `Wallet`, but could be an implementation detail by a `Store`.

andrewgordstewart: Yes, there is, and we can work on that next.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -99,8 +100,21 @@ export class Wallet implements WalletInterface {
...
-    throw 'Unimplemented';
+  async updateChannel({channelId, appData, allocations}: UpdateChannelParams): Result {
+    return Channel.transaction(async tx => {
+      const channel = await Store.getChannel(channelId, undefined);
+
+      const outcome = deserializeAllocations(allocations);
+
+      const decision = updateChannel({channelId, appData, outcome}, channel);
+      if (Either.isLeft(decision)) {
+        throw decision.left;
```
snario, line 111: This is an example where we could make the code slightly more readable with naming:

```ts
const { left: err, right: newState } = ...;

if (err) {
    throw new Error(err);
}

Store.signState(newState);
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,14 +1,29 @@
...
-    return notifications;
+    const {channelResult} = channel;
+
+    return {outgoing, channelResult};
+  },
+
+  getChannel: async function(
+    channelId: Bytes32,
+    tx: Objection.Transaction | undefined
+  ): Promise<ChannelState> {
```
snario, line 26: This could be an `Option`

snario:
```ts
const dbResponse = await Channel.forId(channelId, tx);
if (!dbResponse) return none;
return some(dbResponse.protocolState);
```

packages/server-wallet/src/wallet/index.ts
```diff
@@ -99,8 +100,21 @@ export class Wallet implements WalletInterface {
   async joinChannel(_args: JoinChannelParams): Result {
     throw 'Unimplemented';
   }
-  async updateChannel(_args: UpdateChannelParams): Result {
-    throw 'Unimplemented';
+  async updateChannel({channelId, appData, allocations}: UpdateChannelParams): Result {
+    return Channel.transaction(async tx => {
+      const channel = await Store.getChannel(channelId, undefined);
```
snario, line 105: See [this comment](https://github.com/statechannels/statechannels/pull/2368/files#r461213698). We could have:
```ts
const query = await Store.getChannel(channelId, undefined);
if (isNone(query)) return UpdateChannelError.channelNotFound(channelId);
const {value: channel} = query;
```

snario, line 105: Why do we pass in `undefined`?

andrewgordstewart: It should probably be `tx` :P

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: I am not really convinced that `Option` is worth the overhead, though you did point out a bug in the code: the return type (without option) should be
A. `Promise<ChannelState | undefined>`; or
B. `Promise<Either<ChannelNotFoundError, ChannelState>>`

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: That would be nice, but I think the `Either.Either` type does not have _both_ a `left` and a `right` property: it has either the `left` property or the `right` property.

We can, of course, rename `Either.left` as `isError` and `Either.getLeft` somewhere, and write it like
```
if (isError(result) {
    throw getError(result);
}

Store.signState(getValue(result))
```

Here's another option:
```
      // This can be extracted as a helper
      const throwError = (error: Error) => {
        throw error;
      };
      const nextState = Either.getOrElseW(throwError)(
        updateChannel({channelId, appData, outcome}, channel)
      );
      const {outgoing, channelResult} = await Store.signState(nextState, tx);

      return {outbox: outgoing.map(n => n.notice), channelResults: [channelResult]};
```

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: Resolved in 572a3ec

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: Resolved in 572a3ec. I opted to return `ChannelState | undefined` for now.

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: Fixed in 572a3ec.

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: I tried the second strategy in 572a3ec.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -241,3 +257,11 @@ const takeActions = async (channels: Bytes32[]): Promise<ExecutionResult> => {
...
   return {outbox, error, channelResults};
 };
+
+function getOrThrow<E, T>(result: Either.Either<E, T>): T {
+  return Either.getOrElseW<E, T>(
+    (err: E): T => {
+      throw err;
+    }
+  )(result);
+}
```
snario, line 267: My only hesitation with the existence of this helper is that we don't really want to encourage using it. It might be simpler to implement some kind of catch all error response type out of the wallet, like `WalletError.unspecifiedError({ code: 500, msg: ... })` and log the exception with stack trace using `logger.error`?

I guess I'm just saying it's more valuable to return an unspecified left value to the wallet API caller than it is to throw it, since it's not like this is an _unexpected error_, those of which we do want to see things get thrown, this is expected but not handled. Dunno

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,14 +1,29 @@
...
+    const {channelResult} = channel;
+
+    return {outgoing, channelResult};
+  },
+
+  getChannel: async function(
+    channelId: Bytes32,
+    tx: Objection.Transaction | undefined
+  ): Promise<ChannelState | undefined> {
+    return (await Channel.forId(channelId, tx))?.protocolState;
```
snario, line 27: Wouldn't this throw an exception if `(await Channel.forId(channelId, tx))` was `undefined`?

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: So, we probably _don't_ want to throw an error. Instead, we _want_ to return a strictly typed error in an error case.

This helper method is simply a fill-in until we decide what the wallet API should be, considering the changes that the run-loop architecture and the HTTP-friendly protocols force on us.

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: No, it returns `Promise.resolve(undefined)`!
```
❯ ts-node
> const foo: {bar: undefined | {baz: 'baz'}} = {bar: undefined}
'use strict'
> foo.bar?.baz
undefined
```

packages/server-wallet/src/wallet/index.ts
snario: Yeah my point is just that a slightly better fill-in could be to return some super generic typed error, as a value, to the API caller. As opposed to throwing the error. So, by throwing (using the current code), we would have:

```ts
try {
 {outgoing, channelResults} = wallet.doSomething()
} catch (e) {
  // e is either a known error, like UpdateChannelError or some unknown unknown bug
}
```

As opposed to with a generic error:

```ts
try {
 {errors, outgoing, channelResults} = wallet.doSomething()
} catch (e) {
  // e is some unknown unknown bug
}
```

packages/server-wallet/src/wallet/store.ts
snario: Ah, it's a TypeScript 3.7 [feature](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: > Maybe out of scope for the PR, but I saw `getOrThrow` and was worried if we add this then within the next 7 days we might see it littered in 5 other places out of convenience.

I think that's a legitimate worry, but I believe we are definitely committed to having API methods return typed errors, and never throw. So it's fair to say we will remove this snippet within 7 days (ideally, by tomorrow!).

I'll add a note that it should not be used elsewhere.

Issue #2369, opened 2020-07-27
Investigate Atomicity and PushMessage


lalexgap:
Right now we're making the assumption that `PushMessage` is atomic. That is if any `SignedState` or `Objective` is invalid then nothing gets stored in the wallet.

However we need to consider the case where a message is containing multiple unrelated states or objectives. Does it make sense to allow `PushMessage` to partially suceed?

From @andrewgordstewart 
 alice calls pushMessage with some states related to app1 that trigger a ledger update
2. alice concurrently calls pushMessage with some states related to app2 that trigger a ledger update
then what can happen? her wallet can
1. handle the app1-related states, which queues up a ledger update
2. concurrently, handle the app2-related states, which queues up a ledger update
3. the 1st handler now calls nextAction(ledgerId). there are two queued updates, the wallet is smart, and batches them into one update
4. the 1st handler now has two dirty app channels, and runs nextAction(app1) followed by nextAction(app2)
that would result in the first handler putting outgoing messages for both app1 and app2 (which might not work, if alice is the server end of the HTTP funding protocol)

PR #2370, opened 2020-07-27
Implement PushMessage in the Store

lalexgap:
Implements `PushMessage` in the `Store` and moves a bunch of logic out of the `Channels` mode into the `Store`

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -217,7 +217,7 @@ const takeActions = async (channels: Bytes32[]): Promise<ExecutionResult> => {
 
 // TODO: This should be removed, and not used externally.
 // It is a fill-in until the wallet API is specced out.
-function getOrThrow<E, T>(result: Either.Either<E, T>): T {
+export function getOrThrow<E, T>(result: Either.Either<E, T>): T {
   return Either.getOrElseW<E, T>(
     (err: E): T => {
       throw err;
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -144,10 +144,8 @@ export class Wallet implements WalletInterface {
 
   async pushMessage(message: AddressedMessage): Result {
     const channelIds = await Channel.transaction(async tx => {
-      return Store.pushMessage(message, tx);
+      return await Store.pushMessage(message, tx);
     });
-
-    // TODO: Generate channelIds from objectives
 
     const {channelResults, outbox} = await takeActions(channelIds);
 
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -14,18 +14,16 @@ import {
   Message,
   Outcome,
   SignedStateVarsWithHash,
-  calculateChannelId,
   convertToParticipant,
 } from '@statechannels/wallet-core';
 import * as Either from 'fp-ts/lib/Either';
 import * as Option from 'fp-ts/lib/Option';
 
 import {Bytes32} from '../type-aliases';
-import {Channel, RequiredColumns} from '../models/channel';
+import {Channel} from '../models/channel';
 import {Nonce} from '../models/nonce';
 import {Outgoing, ProtocolAction} from '../protocols/actions';
 import {SigningWallet} from '../models/signing-wallet';
-import {addHash} from '../state-utils';
 import {logger} from '../logger';
 import * as Application from '../protocols/application';
 import knex from '../db/connection';
@@ -145,57 +143,11 @@ export class Wallet implements WalletInterface {
   }
 
   async pushMessage(message: AddressedMessage): Result {
-    const channelIds: Bytes32[] = [];
+    const channelIds = await Channel.transaction(async tx => {
+      return Store.pushMessage(message, tx);
+    });
 
-    try {
-      await Channel.transaction(async tx => {
-        for (const ss of message.signedStates || []) {
-          // We ignore unsigned states
-          if (!ss.signatures?.length) {
-            logger.info(`pushMessage received unsigned state`);
-            return;
-          }
-
-          const channelId = calculateChannelId(ss);
-          let channel = await Channel.query(tx)
-            .where('channelId', channelId)
-            .first();
-
-          if (!channel) {
-            const addresses = ss.participants.map(p => p.signingAddress);
-            const signingWallet = await SigningWallet.query(tx)
-              .whereIn('address', addresses)
-              .first();
-
-            if (!signingWallet) {
-              logger.error(
-                {
-                  knownWallets: await SigningWallet.query(tx).select(),
-                  addresses,
-                },
-                'Not in channel'
-              );
-              throw Error('Not in channel');
-            }
-
-            const {address: signingAddress} = signingWallet;
-            const cols: RequiredColumns = {...ss, vars: [addHash(ss)], signingAddress};
-
-            channel = Channel.fromJson(cols);
-
-            const {channelId} = await Channel.query(tx).insert(channel);
-            channelIds.push(channelId);
-          } else {
-            ss.signatures?.map(sig => channel.addState(ss, sig));
-            await Channel.query(tx).update(channel);
-            channelIds.push(channel.channelId);
-          }
-        }
-      });
-    } catch (err) {
-      logger.error({err}, 'Could not push message');
-      throw err;
-    }
+    // TODO: Generate channelIds from objectives
 
     const {channelResults, outbox} = await takeActions(channelIds);
 
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -92,10 +92,10 @@ export const Store = {
     const sigValidationResult = validateSignatures(signedState);
     if (isLeft(sigValidationResult)) return sigValidationResult;
 
-    const singingWalletResult = await getSigningWallet(signedState, tx);
-    if (isLeft(singingWalletResult)) return singingWalletResult;
+    const signingWalletResult = await getSigningWallet(signedState, tx);
+    if (isLeft(signingWalletResult)) return signingWalletResult;
 
-    const {address: signingAddress} = singingWalletResult.right;
+    const {address: signingAddress} = signingWalletResult.right;
 
     const cols: RequiredColumns = {
       ...signedState,
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -9,6 +9,7 @@ import {
   State,
   calculateChannelId,
   StateVariables,
+  ChannelConstants,
 } from '@statechannels/wallet-core';
 import _ from 'lodash';
 import {Either, left, right, isLeft} from 'fp-ts/lib/Either';
@@ -92,32 +93,18 @@ export const Store = {
     signedState: SignedState,
     tx: Objection.Transaction
   ): Promise<Either<StoreError, undefined>> {
-    const sigValidationResult = validateSignatures(signedState);
-    if (isLeft(sigValidationResult)) return sigValidationResult;
+    getOrThrow(validateSignatures(signedState));
 
-    const signingWalletResult = await getSigningWallet(signedState, tx);
-    if (isLeft(signingWalletResult)) return signingWalletResult;
+    const {address: signingAddress} = getOrThrow(await getSigningWallet(signedState, tx));
 
-    const {address: signingAddress} = signingWalletResult.right;
-
-    const channelId = calculateChannelId(signedState);
-    let channel = await Channel.query(tx)
-      .where('channelId', channelId)
-      .first();
-
-    if (!channel) {
-      const cols: RequiredColumns = {...signedState, vars: [], signingAddress};
-
-      channel = Channel.fromJson(cols);
-
-      await Channel.query(tx).insert(channel);
-    }
-
+    const channel = await getOrCreateChannel(signedState, signingAddress, tx);
     let channelVars = channel.vars;
 
     channelVars = getOrThrow(addState(channelVars, signedState));
 
-    channelVars = clearOldStates(channelVars, channel.isSupported ? channel.support : undefined);
+    channelVars = getOrThrow(
+      clearOldStates(channelVars, channel.isSupported ? channel.support : undefined)
+    );
 
     const invariantValidationResult = validateInvariants(channelVars, channel.myAddress);
     if (isLeft(invariantValidationResult)) return invariantValidationResult;
@@ -145,18 +132,23 @@ enum StoreErrors {
   staleState = 'Stale state',
 }
 
-function validateSignatures(signedState: SignedState): Either<StoreError, undefined> {
-  const {participants} = signedState;
+async function getOrCreateChannel(
+  constants: ChannelConstants,
+  signingAddress: string,
+  tx: Objection.Transaction
+): Promise<Channel> {
+  const channelId = calculateChannelId(constants);
+  let channel = await Channel.query(tx)
+    .where('channelId', channelId)
+    .first();
 
-  for (const sig of signedState.signatures) {
-    const signerIndex = participants.findIndex(p => p.signingAddress === sig.signer);
-    if (signerIndex === -1) {
-      return left(new StoreError(StoreErrors.invalidSignature, {signedState, signature: sig}));
-    }
+  if (!channel) {
+    const cols: RequiredColumns = {...constants, vars: [], signingAddress};
+    channel = Channel.fromJson(cols);
+    await Channel.query(tx).insert(channel);
   }
-  return right(undefined);
+  return channel;
 }
-
 async function getSigningWallet(
   signedState: SignedState,
   tx: Objection.Transaction
@@ -170,6 +162,21 @@ async function getSigningWallet(
     return left(new StoreError(StoreErrors.notInChannel));
   }
   return right(signingWallet);
+}
+/*
+ * Validator functions
+ */
+
+function validateSignatures(signedState: SignedState): Either<StoreError, undefined> {
+  const {participants} = signedState;
+
+  for (const sig of signedState.signatures) {
+    const signerIndex = participants.findIndex(p => p.signingAddress === sig.signer);
+    if (signerIndex === -1) {
+      return left(new StoreError(StoreErrors.invalidSignature, {signedState, signature: sig}));
+    }
+  }
+  return right(undefined);
 }
 
 function validateStateFreshness(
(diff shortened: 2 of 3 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -126,53 +126,22 @@ export class Wallet implements WalletInterface {
...
-        }
-      });
-    } catch (err) {
-      logger.error({err}, 'Could not push message');
-      throw err;
-    }
+    await Store.transaction(async () => {
+      if (message.signedStates) {
+        message.signedStates.forEach(async ss => {
+          await Store.addSignedState(ss);
```
snario, line 132: If one signature is invalid, should the entire API call return an error? Should it return one successful `ChannelUpdated` outbox notification and one `Error` notification? Also, since it is within a transaction, if one fails would they all fail to persist to disk?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -126,53 +126,22 @@ export class Wallet implements WalletInterface {
...
-            await Channel.query(tx).update(channel);
-            channelIds.push(channel.channelId);
-          }
-        }
-      });
-    } catch (err) {
-      logger.error({err}, 'Could not push message');
-      throw err;
-    }
+    await Store.transaction(async () => {
```
snario, line 129: Do all objectives and all signed states need to be a part of the same transaction?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,14 +1,66 @@
...
+import {logger} from '../logger';
+import {addHash} from '../state-utils';
+import knex from '../db/connection';
 
 export const Store = {
+  transaction: async function(callback: () => Promise<void>): Promise<void> {
+    return knex.transaction(async () => {
+      await callback();
+    });
+  },
```
snario, line 16:
```ts
transaction: (callback: () => Promise<void>): Promise<void> => knex.transaction(callback),
```
Maybe more succinct? Changes the return type I think. Why do we want this; is it necessary to lock the _entire_ store for any use case?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,14 +1,66 @@
...
+    return Promise.resolve();
+  },
+  addSignedState: async function(signedState: SignedState): Promise<void> {
+    if (!signedState.signatures?.length) {
+      throw new Error('No signature for signed state');
+    }
+
+    const channelId = calculateChannelId(signedState);
+    let channel = await Channel.query()
+      .where('channelId', channelId)
```
snario, line 36:
```suggestion
      .where({channelId})
```

snario, line 36: There is a helper function in Andrew's PR (see [here](https://github.com/statechannels/statechannels/pull/2368/files#diff-ddee9e09686c5ac512163bd25f5122acR23-R28)) you could re-use

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,14 +1,66 @@
...
+
+      if (!signingWallet) {
+        logger.error(
+          {
+            knownWallets: await SigningWallet.query().select(),
+            addresses,
+          },
+          'Not in channel'
+        );
+        throw Error('Not in channel');
```
snario, line 53: I'd type this error and place it somewhere. This is a good example of a weird edge case we don't really have documented anywhere yet. `PushMessageError.receivedIrrelevantState` or something like this.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,14 +1,66 @@
...
+      }
+
+      const {address: signingAddress} = signingWallet;
+      const cols: RequiredColumns = {...signedState, vars: [addHash(signedState)], signingAddress};
+
+      channel = Channel.fromJson(cols);
+      await Channel.query().insert(channel);
```
snario, line 60: This seems like it should be a method on `Channel` I think?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,14 +1,66 @@
...
     await Channel.query(tx).update(channel);
 
     return notifications;
   },
+
+  addObjective: async function(_objective: Objective): Promise<void> {
+    // TODO: Implement this
+    return Promise.resolve();
+  },
+  addSignedState: async function(signedState: SignedState): Promise<void> {
```
snario, line 29: Seems like in general this function should be more like:

```ts
const ret = validation(ss);

if (ret is invalid) return <whatever the error was>;

else Channel.upsertSignedState(ss);
```

At the moment the validation and the upsert logic is all in one function.

snario: Probably `vet1` should happen _outside` the DB transaction context as well, to minimize time with the lock.

packages/server-wallet/src/wallet/index.ts
lalexgap: Currently the assumption is that if a signature is invalid or a state is malformed the whole `PushMessage` call fails and nothing is persisted. I created #2369 for further discussion around this.

I think a common case is receiving a message containing `SignedState`s and `Objective`s for a common task (ie: virtual funding), so it seems to make sense to either accept them all or reject them all instead of partially updating the store. It also makes it an easier API call for the `app` to understand, as it would either fail or succeed instead of partially succeeding/failing.

That said there may be some performance considerations that we might run into or other things to consider!

packages/server-wallet/src/wallet/store.ts
lalexgap: Currently `Channel` has a lot of code that looked like it was copied from the previous `store`  that should probably be moved to the current `store` so we can structure the code this way.

lalexgap:
> Non-blocking general comment... shouldn't we persist the message to disk before we do anything at all? This way if something happens mid-execution we could restart the client and read any new unhandled messages? This might warrant a new DB model, `Messages` or something, I guess? Thoughts @andrewgordstewart? It could just be a "Level 1" feature, but worth documenting somewhere.

Would this apply for all API calls? It seems like other API calls could run into the same problem of crashing mid-execution as well, (although I guess other API calls are easy for the `app` to reconstruct)

packages/server-wallet/src/wallet/store.ts
lalexgap: Current errors have just been copy and pasted from existing code, so I still need to go thru and clean them up.

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: FYI: SQL transactions do not lock the entire store. Postgresql will acquire locks as necessary during the transaction, such as row-level write locks if on multiple rows, if updating multiple channels within a single transaction.

This is different from indexedDB transactions, which are scoped to (named) stores.
- write transactions acquire a read/write lock on the entirety of all stores that it's scoped to
- read transactions acquire a read lock, which do not prevent others from reading

andrewgordstewart:
> Non-blocking general comment... shouldn't we persist the message to disk before we do anything at all? This way if something happens mid-execution we could restart the client and read any new unhandled messages? This might warrant a new DB model, `Messages` or something, I guess? Thoughts @andrewgordstewart? It could just be a "Level 1" feature, but worth documenting somewhere.

I guess this depends on who is responsible for retrying. If the wallet crashes mid-execution, then the app did as well. Will the app call `pushMessage` again?

packages/server-wallet/src/wallet/index.ts
snario: It seems to me like if we have valid data that may at some point be useful, we should probably persist it just to be safe. Maybe this should be true for `signedStates` and not `objectives`?

packages/server-wallet/src/wallet/store.ts
snario: Neat. Is there any reason to use `Channel.transaction` ever if we can just use this general purpose `Store.transaction`? 

Also, it seems like `Store.transaction` is not "backend-agnostic" then, since it doesn't make sense in the context of an IndexedDB backend _(or, maybe it does, if you implement some logic to check which tables should be modified in the IndexedDB case?)_. Maybe we should just be careful about what methods we tag on to the `Store` since a goal is to figure out how much of this can be re-used in the browser wallet.

snario:
> I guess this depends on who is responsible for retrying. If the wallet crashes mid-execution, then the app did as well. Will the app call `pushMessage` again?

This isn't a guarantee, for example xstate-wallet.statechannels.org could crash for some reason but the application tab might remain open. Or, if the server-wallet were wrapped in by an HTTP server and the "app" was making network calls within a trusted private network, the microservice could go down for any number of reasons but the app would remain alive.

I think that if the wallet has new data, and has a database connection, it should have a responsibility to ensure it does whatever it takes to handle that message. So, persisting it such that it can continue operation if the process crashes would be reasonable.

...

Upon re-reading this though I think that would mean by "app" is basically "whatever client is responsible for making direct calls to the `Wallet` object. In the `server-wallet` this means whatever instantiates a `new Wallet` but in the browser wallet it actually maps more logically to the message handler which we would probably usually consider as "part of the wallet".

If we refactored the browser wallet based on things we've learned here so far, it's possible that we might implement some kind of persistence layer inside the bit of code that watches for incoming postMessage messages (i.e., the `ChannelWallet.messagingService`), but _not_ inside the `Store`. 

In either case though, these things would probably be re-using the same database, which might make it more beneficial to implement it inside the `Store`.

I'm not sure what I think.

snario:
I think this is missing the functionality that should emit a "ChannelProposed" event?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -126,18 +126,17 @@ export class Wallet implements WalletInterface {
...
+    if (message.signedStates) {
+      message.signedStates.forEach(async ss => {
+        await Store.addSignedState(ss);
+      });
+    }
+    if (message.objectives) {
+      message.objectives.forEach(async o => {
+        await Store.addObjective(o);
+      });
+    }
```
snario: I think any of these could _return_ an error, not throw an error, so shouldn't this be inside some kind of a loop which checks for errors responses and if it sees one, returns it to the caller?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,66 +1,225 @@
...
-  },
   signState: async function(action: SignState, tx: Objection.Transaction): Promise<SyncState> {
     const channel = await Channel.forId(action.channelId, tx);
-    const notifications = channel.signAndAdd(action);
-    await Channel.query(tx).update(channel);
+    const state: State = {...channel.channelConstants, ...action};
+
+    const validationResult = validateStateFreshness(state, channel);
+
+    if (isLeft(validationResult)) throw validationResult.left;
```
snario, line 28: Could re-use the `getOrThrow` helper by @andrewgordstewart here, and the renamed `isLeft` to `isError`.

Also I mentioned this [in a previous PR comment](https://github.com/statechannels/statechannels/pull/2368#discussion_r461212099), but we can clean up the responses using response value destructuring.

I still think we should do this... here is proof it works:

```ts
> let test = fp.either.right("abc")
undefined
> let { left: error, right : val } = example;
undefined
> if (error) throw Error('');
undefined
> val
'abc'
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,66 +1,225 @@
...
-      const {address: signingAddress} = signingWallet;
-      const cols: RequiredColumns = {...signedState, vars: [addHash(signedState)], signingAddress};
-
-      channel = Channel.fromJson(cols);
-      await Channel.query().insert(channel);
-    } else {
-      signedState.signatures?.map(sig => channel.addState(signedState, sig));
-      await Channel.query().update(channel);
-    }
+    const singingWalletResult = await getSigningWallet(signedState);
```
snario, line 71: Typo "`sing`"`ingWallet`

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,66 +1,225 @@
...
+}
+
+enum StoreErrors {
+  duplicateTurnNums = 'multiple states with same turn number',
+  notSorted = 'states not sorted',
+  multipleSignedStates = 'Store signed multiple states for a single turn',
+  invalidSignature = 'Invalid signature',
+  notInChannel = 'Not in channel',
+  staleState = 'Stale state',
+}
```
snario, line 109: Is there some way we can make these things be parametrizable functions by default? i.e., when defining a new exception I can define both the exception type and the arguments that led to it? e.g., `multipleSignedStates(ss1: SignedState, ...sss: SignedState[])`?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,66 +1,225 @@
...
+
+function validateSignatures(signedState: SignedState): Either<StoreError, undefined> {
+  const {participants} = signedState;
+
+  for (const sig of signedState.signatures) {
+    const signerIndex = participants.findIndex(p => p.signingAddress === sig.signer);
+    if (signerIndex === -1) {
```
snario, line 116:
```ts
if (!_.find(participants, {signingAddress: sig.signer})) {
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,66 +1,225 @@
...
+  return right(undefined);
+}
+
+async function getSigningWallet(
+  signedState: SignedState
+): Promise<Either<StoreError, SigningWallet>> {
+  const addresses = signedState.participants.map(p => p.signingAddress);
+  const signingWallet = await SigningWallet.query()
+    .whereIn('address', addresses)
+    .first();
```
snario, line 129: I've seen this a few times before, probably we can extract into a utility method on `SigningWallet`?

snario: `await SigningWallet.for(signedState)`

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,66 +1,225 @@
...
+
+  if (!signingWallet) {
+    return left(new StoreError(StoreErrors.notInChannel));
```
snario, line 132: I think in another PR the pattern is to define a class per error, e.g., `new StoreErrors.NoSigningWalletForChannel({channelId})`

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,66 +1,225 @@
...
+  channel: Channel
+): Either<StoreError, undefined> {
+  if (
+    channel.isSupportedByMe &&
+    channel.latestSignedByMe &&
+    channel.latestSignedByMe.turnNum >= signedState.turnNum
+  ) {
+    return left(new StoreError(StoreErrors.staleState));
+  }
+  return right(undefined);
```
snario, line 148: Seems more intuitive to return `true`, maybe?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -1,66 +1,225 @@
...
+}
+
+function validateInvariants(
+  stateVars: SignedStateVarsWithHash[],
+  myAddress: string
+): Either<StoreError, undefined> {
+  const signedByMe = stateVars.filter(s => s.signatures.some(sig => sig.signer === myAddress));
+  const groupedByTurnNum = _.groupBy(signedByMe, s => s.turnNum.toString());
+  const multipleSignedByMe = _.map(groupedByTurnNum, s => s.length)?.find(num => num > 1);
```
snario, line 157: I think these sorts of things should be extracted as general purpose functional utilities and then composed.

e.g.,

```ts
const signatureCheckerFor = (address: string) => (({signer}): Signature) => signer === address;

const stateSignedBy = ({signatures}: SignedState, address: string) =>
    _.any(signatures, signatureChecker(address)); 

const multipleSignedBy = (signedStates: SignedState[], address: string): boolean =>
    !!_.chain(signedStates)
     .groupBy(signedStates, 'turnNum')
     .filter(state => stateSignedBy(state, address))
     .any(group => group.length > 1)

```

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: `Channel.transaction` is just a convenience method for starting a database transaction. It's not special to the `Channel` model in any way.

It's not really possible to have a db-agnostic transaction, since database management systems all implement transactions differently. The store needs to be up front about what properties of transactions it is using. The important property is the ability to roll back.

The browser wallet also assumed that transactions could be scoped to "object stores", and that write locks would Lock the entire object store. That is something you can implement in Postgres, but it is not the default behaviour (for good reason!)

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: Is that interpreted by node or ts-node? I would be surprised if that code compiles without type errors

andrewgordstewart:
```
> import * as fp from 'fp-ts'
undefined
> let test = fp.either.right("abc")
undefined
> let { left: error, right : val } = test;
[eval].ts:7:7 - error TS2339: Property 'left' does not exist on type 'Either<never, string>'.

7 let { left: error, right : val } = test;
        ~~~~
[eval].ts:7:20 - error TS2339: Property 'right' does not exist on type 'Either<never, string>'.

7 let { left: error, right : val } = test;
```

It does not compile

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: Which PR?

packages/server-wallet/src/wallet/store.ts
snario: Sad :(

packages/server-wallet/src/wallet/store.ts
snario: Maybe I imagined it, I though this is how I remembered you doing it

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: It's not really that sad. We initially added `fp-ts` because it defined an `Either` type, and @kerzhner had good experience with it in the `simple-hub` package. We are free to define our own `Either` type that can have the property you'd like.

If we are not going to use `pipe`, `chain`, etc. from `fp-ts`, then maybe we just remove the dependency and define
```
type Either<E extends Error, T> = {error: E; value: undefined, ok: false} | {error: undefined, value: T, ok: true}

const test: Either<Error, 'test'> = {error: undefined, value: 'test', ok: true}
if (test.ok) {
  // test.value is of type `'test'`
}

if (test.error) {
  // test.error is of type `Error`
}

if (test.value) {
  // test.value is of type `'test'`
}
```

If we are going to use things like `pipe`, `chain`, etc. then maybe it helps to keep using `fp-ts`, and its `Either` type Or, even in that case, implement our own that extends `Either` from `fp`, but has the property that you want:
```
type Either<E, T> = {_tag: 'left'; left: E; right: undefined} | {_tag: 'right'; left: undefined; right: T}
```

packages/server-wallet/src/wallet/store.ts
lalexgap: Using a `boolean` value seems slightly strange with `Either` since you would never have a `right` value of `false`. Using `undefined` seems to fit better with the pattern of calling the `validate` method and doing something only if there's an error.

Although using undefined is still slightly weird...

lalexgap:
> Is there a way to checkpoint some of this work, and have it merged into master partially finished?

I think it's in a good enough state to get merged in. If someone wants to review and flag any flagrant issues I'll get this merged in.

packages/server-wallet/src/wallet/store.ts
kerzhner: Do we want to throw an error in this case? It seems like we are mixing `Either`s and `throw`s. It looks like customers of the `signState` api ([one example](https://github.com/statechannels/statechannels/blob/ag/push-message/packages/server-wallet/src/wallet/index.ts#L116)), do not catch the error.

kerzhner: Per https://statechannels.slack.com/archives/CPX2BU1FA/p1596063316027800, I think we should get rid of `Either` and use `throw`

PR #2375, opened 2020-07-28
Join channel

andrewgordstewart:
- Adds a join-channel handler, similar to the existing update-channel handler
- Implements `joinChannel` on the wallet

## Open questions
- [ ] (A) Sometimes, two messages get sent to the same recipient. Should those messages be merged?
   - What if the messages are for unrelated channels?
- [ ] (B) What is the definition of a `ChannelResult`?
   - What is the definition of `status`? 
   - What `turnNum` should be reported? That of `latest`? Or of `supported`?
- [ ] (C) Depending on the answer to (B), there might be multiple `ChannelResult`s for a single channel as a result of the API call.
   - Should they be merged into a single, "maximal" result?
   - If so, how do you order results?

### DIFF
--- a/packages/server-wallet/src/handlers/__test__/join-channel.test.ts
+++ b/packages/server-wallet/src/handlers/__test__/join-channel.test.ts
@@ -1,32 +1,31 @@
 import matchers from '@pacote/jest-either';
+import {left, right} from 'fp-ts/lib/Either';
 
 import {joinChannel} from '../join-channel';
 import {joinChannelFixture} from '../fixtures/join-channel';
 import {channelStateFixture} from '../../protocols/__test__/fixtures/channel-state';
+import {channel} from '../../models/__test__/fixtures/channel';
+import {stateSignedBy} from '../../wallet/__test__/fixtures/states';
+import {bob, alice} from '../../wallet/__test__/fixtures/signing-wallets';
 expect.extend(matchers);
 
-const args = joinChannelFixture();
+const prefundVars = {turnNum: 0, appData: '0xf00'};
+const runningVars = {turnNum: 7, appData: '0xf00'};
 
-test('validJoin', () => {
-  expect(
-    joinChannel(args, channelStateFixture({latest: {turnNum: 0, appData: 'foo'}}))
-  ).toMatchRight({
-    type: 'SignState',
-    turnNum: 0,
-    appData: 'foo',
-  });
+test.each`
+  input                                                                 | result
+  ${channel({vars: [stateSignedBy(bob())(prefundVars)]}).protocolState} | ${{type: 'SignState', ...prefundVars}}
+  ${channelStateFixture({latest: prefundVars})}                         | ${{type: 'SignState', ...prefundVars}}
+`('happy path', ({input, result}) => {
+  expect(joinChannel(joinChannelFixture(), input)).toMatchObject(right(result));
 });
 
-describe('invalid join', () => {
-  test('when the latest state is not turn 0', () =>
-    expect(joinChannel(args, channelStateFixture({latest: {turnNum: 3}}))).toMatchObject({
-      left: new Error('latest state must be turn 0'),
-    }));
-
-  test('when I have signed a state', () =>
-    expect(
-      joinChannel(args, channelStateFixture({latest: {turnNum: 0}, latestSignedByMe: {turnNum: 0}}))
-    ).toMatchObject({
-      left: new Error('already signed prefund setup'),
-    }));
+test.each`
+  input                                                                          | result
+  ${channelStateFixture({latestSignedByMe: prefundVars})}                        | ${new Error('already signed prefund setup')}
+  ${channel({vars: [stateSignedBy(alice())(prefundVars)]}).protocolState}        | ${new Error('already signed prefund setup')}
+  ${channel({vars: [stateSignedBy(alice(), bob())(prefundVars)]}).protocolState} | ${new Error('already signed prefund setup')}
+  ${channelStateFixture({latest: runningVars})}                                  | ${new Error('latest state must be turn 0')}
+`('error cases', ({input, result}) => {
+  expect(joinChannel(joinChannelFixture(), input)).toMatchObject(left(result));
 });
--- a/packages/server-wallet/src/handlers/__test__/join-channel.test.ts
+++ b/packages/server-wallet/src/handlers/__test__/join-channel.test.ts
@@ -1,0 +1,32 @@
+import matchers from '@pacote/jest-either';
+
+import {joinChannel} from '../join-channel';
+import {joinChannelFixture} from '../fixtures/join-channel';
+import {channelStateFixture} from '../../protocols/__test__/fixtures/channel-state';
+expect.extend(matchers);
+
+const args = joinChannelFixture();
+
+test('validJoin', () => {
+  expect(
+    joinChannel(args, channelStateFixture({latest: {turnNum: 0, appData: 'foo'}}))
+  ).toMatchRight({
+    type: 'SignState',
+    turnNum: 0,
+    appData: 'foo',
+  });
+});
+
+describe('invalid join', () => {
+  test('when the latest state is not turn 0', () =>
+    expect(joinChannel(args, channelStateFixture({latest: {turnNum: 3}}))).toMatchObject({
+      left: new Error('latest state must be turn 0'),
+    }));
+
+  test('when I have signed a state', () =>
+    expect(
+      joinChannel(args, channelStateFixture({latest: {turnNum: 0}, latestSignedByMe: {turnNum: 0}}))
+    ).toMatchObject({
+      left: new Error('already signed prefund setup'),
+    }));
+});
--- a/packages/server-wallet/src/handlers/join-channel.ts
+++ b/packages/server-wallet/src/handlers/join-channel.ts
@@ -1,0 +1,62 @@
+import {Either, left, right, chain, map} from 'fp-ts/lib/Either';
+import {StateVariables} from '@statechannels/wallet-core';
+import {pipe} from 'fp-ts/lib/function';
+import {ChannelId} from '@statechannels/client-api-schema';
+
+import {SignState} from '../protocols/actions';
+import {ChannelState} from '../protocols/state';
+
+type HandlerResult = Either<JoinChannelError, SignState>;
+type StepResult = Either<JoinChannelError, ChannelState>;
+export interface JoinChannelHandlerParams {
+  channelId: ChannelId;
+}
+
+export enum Errors {
+  channelNotFound = 'channel not found',
+  invalidLatestState = 'latest state must be turn 0',
+  alreadySignedByMe = 'already signed prefund setup',
+}
+
+export class JoinChannelError extends Error {
+  readonly type = 'JoinChannelError';
+  constructor(reason: Errors, public readonly data: any = undefined) {
+    super(reason);
+  }
+}
+
+// The helper functions should be factored out, tested, and reusable
+const hasStateSignedByMe = (cs: ChannelState): boolean => !!cs.latestSignedByMe;
+
+const ensureNotSignedByMe = (cs: ChannelState): Either<JoinChannelError, ChannelState> =>
+  hasStateSignedByMe(cs) ? left(new JoinChannelError(Errors.alreadySignedByMe)) : right(cs);
+
+function ensureLatestStateIsPrefundSetup(cs: ChannelState): StepResult {
+  if (cs.latest.turnNum === 0) return right(cs);
+  return left(new JoinChannelError(Errors.invalidLatestState));
+}
+
+function latest(cs: ChannelState): StateVariables {
+  return cs.latest;
+}
+// End helpers
+
+export function joinChannel(
+  args: JoinChannelHandlerParams,
+  channelState: ChannelState
+): HandlerResult {
+  // TODO: use Action creator from another branch
+  const signStateAction = (sv: StateVariables): SignState => ({
+    type: 'SignState',
+    channelId: args.channelId,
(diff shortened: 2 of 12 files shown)


packages/server-wallet/src/handlers/join-channel.ts
```diff
@@ -0,0 +1,62 @@
...
+
+type HandlerResult = Either<JoinChannelError, SignState>;
+type StepResult = Either<JoinChannelError, ChannelState>;
+export interface JoinChannelHandlerParams {
+  channelId: ChannelId;
+}
+
+export enum Errors {
+  channelNotFound = 'channel not found',
+  invalidLatestState = 'latest state must be turn 0',
```
snario, line 17: Maybe `invalidTurnNum` to be more descriptive?

packages/server-wallet/src/handlers/join-channel.ts
```diff
@@ -0,0 +1,62 @@
...
+// The helper functions should be factored out, tested, and reusable
+const hasStateSignedByMe = (cs: ChannelState): boolean => !!cs.latestSignedByMe;
+
+const ensureNotSignedByMe = (cs: ChannelState): Either<JoinChannelError, ChannelState> =>
+  hasStateSignedByMe(cs) ? left(new JoinChannelError(Errors.alreadySignedByMe)) : right(cs);
+
+function ensureLatestStateIsPrefundSetup(cs: ChannelState): StepResult {
+  if (cs.latest.turnNum === 0) return right(cs);
+  return left(new JoinChannelError(Errors.invalidLatestState));
+}
```
snario: Seems to me like these could be defined in some a group to be more easily readable, like...

```ts
const joinChannelValidators = [ notSignedByMe, latestStateIsPrefundSetup ];
```

packages/server-wallet/src/handlers/join-channel.ts
```diff
@@ -0,0 +1,62 @@
...
+  readonly type = 'JoinChannelError';
+  constructor(reason: Errors, public readonly data: any = undefined) {
+    super(reason);
+  }
+}
+
+// The helper functions should be factored out, tested, and reusable
+const hasStateSignedByMe = (cs: ChannelState): boolean => !!cs.latestSignedByMe;
+
+const ensureNotSignedByMe = (cs: ChannelState): Either<JoinChannelError, ChannelState> =>
```
snario, line 31:
```suggestion
const ensureNotSignedByMe = (cs: ChannelState): StepResult =>
```

packages/server-wallet/src/wallet/index.ts
```diff
@@ -81,23 +82,36 @@ export class Wallet implements WalletInterface {
...
+    const {outbox} = await knex.transaction(
+      async (tx): Promise<{outbox: any}> => {
+        const channel = await Store.getChannel(channelId, tx);
+
+        if (!channel)
+          throw new JoinChannel.JoinChannelError(JoinChannel.Errors.channelNotFound, {
+            channelId,
+          });
+
+        const nextState = getOrThrow(JoinChannel.joinChannel({channelId}, channel));
```
snario, line 103: `getOrThrow ` 👀

packages/server-wallet/src/wallet/index.ts
```diff
@@ -210,47 +224,52 @@ const takeActions = async (channels: Bytes32[]): Promise<ExecutionResult> => {
...
-    } catch (err) {
-      // FIXME
-      logger.error({err}, 'Error handling action');
-      await setError(err);
-    }
+      if (!app) {
+        setError(new Error('Channel not found'));
+
+        throw 'unreachable';
+      }
```
snario, line 259: I'm a bit confused by this block, why does it exist?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -210,47 +224,52 @@ const takeActions = async (channels: Bytes32[]): Promise<ExecutionResult> => {
...
+
+      try {
+        const nextAction = await Application.protocol({app});
+        // TODO: handleAction might also throw an error.
+        // It would be nice for handleAction to return an Either type, pipe the right values,
+        // and handle the left values with setError
+        await Either.fold(setError, Option.fold(markChannelAsDone, handleAction))(nextAction);
+      } catch (err) {
+        // FIXME
```
snario, line 268: Could add a note about what specifically should be fixed

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: There is a big, outstanding question is: what is the wallet API?
- When should you return exactly one `ChannelResult`? When should you return an array of `ChannelResult[]`?
- What do you do when you have some successful results but some errors?

`getOrThrow` is a fill in until we answer questions like that. It is not meant to stay, and has an explicit note as such.

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: You should handle the case if there is no app channel. This block sets the error, and calls `tx.rollback()`, which ends up throwing an error as well.

However, typescript doesn't know that, and if you don't either throw or return, then the compiler thinks `app` may still be undefined when you leave this block.

The reality is that we should _return_  an error, and then this throw is unnecessary.

packages/server-wallet/src/handlers/join-channel.ts
andrewgordstewart: I'm not sure if there's much added value in that, since the implementation of the `joinChannel` handler _is_ that group.

Do you think there's much more added value in doing something like
```
const joinChannelValidators = [ notSignedByMe, latestStateIsPrefundSetup ];
const joinChannelHandler = flow(...joinChannelValidators)
```
vs something like
```
const joinChannelHandler = flow(notSignedByMe, latestStateIsPrefundSetup)
```
?

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: https://github.com/statechannels/statechannels/issues/2379

packages/server-wallet/src/handlers/join-channel.ts
tomclose: My immediate thought when reading both of these is "what is `flow`?".

tomclose: Do you mean `pipe`?

packages/server-wallet/src/handlers/__test__/join-channel.test.ts
```diff
@@ -0,0 +1,31 @@
+import matchers from '@pacote/jest-either';
```
snario, line 1: Didn't see this in the PR, was this installed to dev dependencies?

packages/server-wallet/src/handlers/join-channel.ts
```diff
@@ -0,0 +1,62 @@
...
+  const signStateAction = (sv: StateVariables): SignState => ({
+    type: 'SignState',
+    channelId: args.channelId,
+    ...sv,
+  });
+
+  return pipe(
+    channelState,
+    ensureNotSignedByMe,
+    chain(ensureLatestStateIsPrefundSetup),
```
snario, line 58: Why are one of these functions wrapped in `chain` and the other not? They seem to have the same signature

packages/server-wallet/src/wallet/__test__/integration/join-channel.test.ts
```diff
@@ -0,0 +1,68 @@
+import {simpleEthAllocation} from '@statechannels/wallet-core';
+
+import {Channel} from '../../../models/channel';
+import {Wallet} from '../..';
+import {seed} from '../../../db/seeds/1_signing_wallet_seeds';
```
snario, line 5: Maybe do `seed as seedSigningWallets` to be more more explicit?

packages/server-wallet/src/wallet/__test__/integration/join-channel.test.ts
```diff
@@ -0,0 +1,68 @@
...
+let w: Wallet;
+beforeEach(async () => {
+  await truncate(knex);
+  w = new Wallet();
+});
+
+// Make sure alice's PK is in the DB
+beforeEach(async () => {
+  await seed(knex);
+});
```
snario, line 21: Any reason not to wrap all of this and the `it`s in a `describe`?

packages/server-wallet/src/wallet/__test__/integration/join-channel.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      {params: {recipient: 'bob', sender: 'alice', data: {signedStates: [postFS]}}},
+    ],
+    // TODO: channelResults is not calculated correctly: see the Channel model's channelResult
+    // implementation
+    // channelResults: [{channelId, turnNum: 3, outcome, status: 'funding'}],
+  });
+
+  const updated = await Channel.forId(channelId, undefined);
+  expect(updated.protocolState).toMatchObject({latest: postFS, supported: preFS});
+});
```
snario, line 68: It might be worthwhile to dream up scenarios of tests that will eventually need to pass based on our experience with previous wallets and create empty `it.skip` tests that describe what we eventually need to implement. Since we haven't written down every possible scenario, we probably need to do lots of incremental things to ensure we don't forget about edge cases to handle

packages/server-wallet/src/wallet/index.ts
```diff
@@ -210,47 +224,54 @@ const takeActions = async (channels: Bytes32[]): Promise<ExecutionResult> => {
...
-          channelResults.push(channelResult);
-          return;
+    await knex.transaction(async tx => {
+      const setError = async (e: Error): Promise<void> => {
+        error = e;
+        await tx.rollback(error);
+      };
+      const markChannelAsDone = async (): Promise<any> => {
+        channels.shift() as string;
+      };
```
snario, line 234: This is neither async and returns `void`, maybe we can update the signature?

packages/server-wallet/src/wallet/index.ts
snario: This would indicate a corrupt DB essentially right?

andrewgordstewart: What do you mean by "indicate"? Do you mean "result in"?

packages/server-wallet/src/wallet/__test__/integration/join-channel.test.ts
andrewgordstewart: Is there a reason to do that?

packages/server-wallet/src/handlers/join-channel.ts
andrewgordstewart: https://gcanti.github.io/fp-ts/modules/function.ts.html#flow

I don't know why `flow` is not called `compose`...

packages/server-wallet/src/handlers/__test__/join-channel.test.ts
andrewgordstewart: It's already a dev dependency!

packages/server-wallet/src/handlers/join-channel.ts
andrewgordstewart: While true, they do not return a compatible type as input. If I have `f: (a: T1)=> T2` and `g: (a: T1) => T2` then I cannot compose them (unless `T2` extends `T1`)

`chain` changes a function to accept an `Either`, and either
- propagate the error, if given a `left`
- call `ensureLatestStateIsPrefundSetup` on the value, if given a `right`

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: It has the specific signature for use in `Option.fold`. We should simply not use `Option.fold`, which would correct this.

Can be done with https://github.com/statechannels/statechannels/issues/2379.

packages/server-wallet/src/wallet/index.ts
snario: If `!app` but the wallet is executing this code, it means that some internal bit of code was run inside the wallet that thought an app existed (i.e., somehow any checks for existence of a `Channel` in the DB passed, or no check was done), and then passed it to this internal function. So... if `!app` then it's either a bug in the code, or the DB was corrupted during execution?

Maybe this is a pointless question. Just reasoning about why this code exists and we can't just assume `app` or somehow have a guarantee before the function executes.

packages/server-wallet/src/handlers/join-channel.ts
snario: My comment was strictly written with code readability in mind... for the random passer-by of this file on Github, what small changes could we make to the way its written to maximize the likelihood they understand what's going on

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: `takeActions` is a generic function that accepts arbitrary channel IDs, so it should gracefully handle the case where the channel you're given doesn't exist.

Something, somewhere, has to convince the typescript compiler that app is not undefined at a point where you access properties of it. That is the purpose of this block of code.

We can also assert that `app` is not undefined?

packages/server-wallet/src/handlers/join-channel.ts
andrewgordstewart: I'm not exactly sure what change you're proposing though. Are you suggesting something like the snippets above? Or just defining an unused constant `joinChannelValidators` for documentation purposes?

packages/server-wallet/src/wallet/__test__/integration/join-channel.test.ts
andrewgordstewart: https://github.com/statechannels/statechannels/pull/2375/commits/dff9037f149ec351d758f508dc939b0bfa1794ec

PR #2380, opened 2020-07-28
server-wallet: update-channel handler cleanups +more unit tests

kerzhner:


packages/server-wallet/src/handlers/fixtures/update-channel.ts
```diff
@@ -1,16 +1,18 @@
...
-  ]),
+  ...stateVars(),
+  channelId,
 };
 
 export const updateChannelFixture = fixture(defaultVars);
+
+export const signStateFixture = fixture(
+  signState({channelId, ...stateVars({turnNum: lastPostFundTurnNum + 1})})
+);
```
andrewgordstewart, line 18: Should this be in `protocols/__tests__/fixtures/actions.ts`?

packages/server-wallet/src/handlers/update-channel.ts
```diff
@@ -3,32 +3,37 @@ import {SignedStateWithHash, StateVariables, Outcome} from '@statechannels/walle
...
+
+  static readonly reasons = {
+    channelNotFound: 'channel not found',
+    invalidLatestState: 'must have latest state',
+    notInRunningStage: 'channel must be in running state',
+    notMyTurn: 'it is not my turn',
+  } as const;
+
+  constructor(
+    reason: typeof UpdateChannelError.reasons[keyof typeof UpdateChannelError.reasons],
```
andrewgordstewart, line 34: FYI I played around with making an error-class-factory for about 5 or 10 minutes, and then gave up.

If we use this pattern often, then it would be nice to define `type Values<E> = E[keyof E];`.
then, this can be
```suggestion
    reason: Values<typeof UpdateChannelError.reasons>,
```

packages/server-wallet/src/handlers/update-channel.ts
```diff
@@ -39,16 +44,19 @@ const hasSupportedState = (cs: ChannelState): cs is ChannelStateWithSupported =>
 const ensureSupportedStateExists = (
   cs: ChannelState
 ): Either<UpdateChannelError, ChannelStateWithSupported> =>
-  hasSupportedState(cs) ? right(cs) : left(new UpdateChannelError(Errors.invalidLatestState));
+  hasSupportedState(cs)
+    ? right(cs)
+    : left(new UpdateChannelError(UpdateChannelError.reasons.invalidLatestState));
```
andrewgordstewart, line 49: :idea: Maybe on another iteration of the construction of errors, the `UpdateChannelError` class could have static constructors for error reason. Then, this could be refactored as:
```suggestion
    : left(UpdateChannelError.invalidLatestState());
```

packages/server-wallet/src/handlers/fixtures/update-channel.ts
kerzhner: Good call: a8861313d

packages/server-wallet/src/handlers/update-channel.ts
kerzhner: That does look cleaner: 81cfd0442

packages/server-wallet/src/handlers/update-channel.ts
kerzhner: I thought about a similar idea, but then decided against it since, for every new error reason, we would have to add a static function. That being said, unit tests need to construct error instances. So, a static factory would be nice for that.

PR #2383, opened 2020-07-29
[docs] Add client-api-schema to Typescript API docs

geoknee:
This is kind of important, IMO: it allows application developers to inspect the types defined in this package: e.g. `ChannelResult`. Difficult to use the channel client without this. 

We could consider re-exporting type such as those from the `channel-client`. Otherwise (I think?) folks need to import a third(!) package in order to easily write their application.

TODO

- [ ] Add package readmes for all published packages #2389 
- [x] Make sure everything is ported over from `client-api-docs` and then delete that package

**Note:** Large number of autogenerated files added in this PR. We could decide to not check them into the repo: however, when we generate a new docs version they would all get copied over and checked in anyways. So we may as well have them from the get-go.


geoknee:
Preview here https://5f22f13d65cf9339d59f474b--nitro-protocol.netlify.app/docs/next/typescript-api/

This section of our site is designed to replace https://app-wallet-interface.netlify.app/, which I suggest we wind down when we merge this PR. I have tried to maintain all of the information that was in that package (`client-api-docs`, now deleted with this changeset). 

I think it is fair to say 

1. that information is now tightly coupled to the underlying code via TSDoc annotations (and these don't seem to conflict with the json-schema generaor annotations, we just get a warning). ✅ 
2. The information is not that easy to consume, being a bit of a rabbit-hole of generic types and aliases. 🚫 
3. This way of doing this, along with the api report, _does_ highlight quite a few places where we can clean up the way we define, export and share types across our packages. ✅  (to be addressed in future PRs).

PR #2384, opened 2020-07-29
Update Ping Pong clients and get test to furthest point

snario:
Just waiting on https://github.com/statechannels/statechannels/pull/2370 and then can continue to make progress


packages/server-wallet/e2e-test/e2e-utils.ts
```diff
@@ -56,4 +59,21 @@ export const waitForServerToStartAndResetDatabase = (
...
+ */
+export const getPongsParticipantInfo = async (pongServer: PongServer): Promise<Participant> => {
+  const {data: participant} = await axios.get<Participant>(`${pongServer.url}/participant`);
+  return participant;
+};
+
+/**
+ * Seeds Pong's database with a channel.
+ */
+export const seedPongWithChannel = async (
```
andrewgordstewart, line 74: I'm not sure if the indirection of adding and hitting HTTP endpoints to seed a channel is that valuable.

One alternative is to explicitly seed the database: You can have two `knex` objects: `knexPing`, connected to the `ping` database and `knexPong`, connected to the `pong` database. You can then do
```
import {seedChannels} from './path/to/seeds/seed_channels'
import {channel} from './path/to/fixtures/channels

await seedChannels(knexPong, myChannels)
```

snario: Yeah I think this is better too, I'll do this.

packages/server-wallet/e2e-test/e2e-utils.ts
```diff
@@ -56,4 +59,21 @@ export const waitForServerToStartAndResetDatabase = (
     }, pingInterval);
   });
 
+/**
+ * Asks the Pong Server to identify itself so that Ping may use this info in the creation
+ * of a new channel with Pong.
+ */
+export const getPongsParticipantInfo = async (pongServer: PongServer): Promise<Participant> => {
```
andrewgordstewart, line 66: This seems like a legitimate method that belongs on the `PingClient` -- the client would need to find out the peer's participant ID.

packages/server-wallet/e2e-test/ping/client.ts
```diff
@@ -1,14 +1,26 @@
...
 
 export default class PingClient {
-  private readonly wallet: Wallet = new Wallet();
+  private readonly wallet: ServerWallet = new ServerWallet();
 
-  constructor(private readonly pongHttpServerURL: string) {}
+  constructor(private readonly pk: Bytes32, private readonly pongHttpServerURL: string) {}
+
+  public get me(): Participant {
+    const {address: signingAddress} = new Wallet(this.pk);
```
andrewgordstewart, line 17: Should `PingClient` have a `get address()` method?

packages/server-wallet/e2e-test/pong/controller.ts
```diff
@@ -1,11 +1,40 @@
...
+    await truncate(knex);
+    await SigningWallet.query().insert(bob());
+  }
+
+  public async seedWith(channelJson: object): Promise<void> {
+    await Channel.query().insert({
+      ...channelJson,
+      // Test fixture used in e2e.test.ts hard-codes Alice atm
+      signingAddress: this.participantInfo.signingAddress,
+    });
```
andrewgordstewart, line 35: These probably don't belong on the `PongClient`.

packages/server-wallet/e2e-test/pong/controller.ts
```diff
@@ -17,29 +46,28 @@ export default class PongController {
...
+      channelResults: [pingsUpdate],
+    } = await this.wallet.pushMessage({
       signedStates: convertedSignedStates,
       to,
       from,
     });
 
-    if (!channelResults) throw Error('sanity check');
+    // FIXME: At this point, pingsUpdate is undefined, but should be a result
+    // waiting on https://github.com/statechannels/statechannels/pull/2370
```
andrewgordstewart, line 58: Can you call `this.wallet.getChannel` to get around that for now? And maybe `logger.warn("pingsUpdate is undefined")` or leave a `TODO` that you shouldn't do that?

packages/server-wallet/e2e-test/pong/controller.ts
```diff
@@ -17,29 +46,28 @@ export default class PongController {
...
-      if (notification.method === 'ChannelProposed') {
-        console.log('Observed a ChannelProposed event');
-      }
-    }
+    if (!pingsUpdate)
+      return {
+        recipient: from,
+        sender: this.myParticipantID,
+        data: {signedStates: [], objectives: []},
+      };
```
andrewgordstewart, line 65: Is this meant to be placeholder code for something?

packages/server-wallet/src/models/__test__/fixtures/channel.ts
```diff
@@ -31,5 +31,10 @@ export const channelWithVars: Fixture<Channel> = fixture<Channel>(
...
 
-export const withSupportedState = (stateVars: Partial<StateVariables>): Fixture<Channel> =>
-  fixture(channel({vars: [addHash(stateSignedBy(alice(), bob())({...stateVars}))]}));
+export const withSupportedState = (
+  stateVars: Partial<StateVariables>,
+  channelProps?: DeepPartial<RequiredColumns>
+): Fixture<Channel> =>
+  fixture(
+    channel({vars: [addHash(stateSignedBy(alice(), bob())({...stateVars}))], ...channelProps})
+  );
```
andrewgordstewart, line 40: I don't understand this change. `withSupportedState(someState)` already accepts `channelProps`, since its type is `Fixture<Channel>`.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -187,7 +187,7 @@ export class Wallet implements WalletInterface {
             channelIds.push(channelId);
           } else {
             ss.signatures?.map(sig => channel.addState(ss, sig));
-            await Channel.query(tx).update(channel);
+            await Channel.query(tx).upsertGraph(channel);
```
andrewgordstewart, line 190: I don't understand the context of this change, since it's made in a single, large commit. Could you comment on why it's necessary?

snario: I think Alex identified it in #2370 but `update` causes a duplicate channel_id error in Postgres

packages/server-wallet/src/models/__test__/fixtures/channel.ts
snario: Oh I see. I didn't know how `Fixture` worked

snario: There is still an issue though because this fixture has computed properties computed by the defaults, e.g., `channelId`

packages/server-wallet/e2e-test/pong/controller.ts
snario: Yes, for `joinChannel`

packages/server-wallet/src/models/__test__/fixtures/channel.ts
andrewgordstewart: Yeah, the fixture pattern is "nice", but it is still rough around the edges. I think my suggestion does not work in this case.

PR #2387, opened 2020-07-30
Store.signState returns Either

kerzhner:
Updates the `signState` api to return an `Either` instead of throwing errors. This work can be contrasted with https://github.com/statechannels/statechannels/pull/2386.

I must admit that working with typed error return values using imperative programming is tedious. We are repeatedly inspecting if a return value is an error or a valid return and making decisions based on that.

Perhaps the tedium is worth it to have explicit, typed error handling?


kerzhner:
Also, [`toMatchRight`](https://www.npmjs.com/package/@pacote/jest-either) doesn't seem to work as expected...

kerzhner:
Closing this as we are no longer using error return values.

PR #2393, opened 2020-07-30
Add transaction methods to Store

### LINKED ISSUE #2372: Add `startTransaction` method to the store
- [ ] Remove uses of `knex.startTransaction`
- [ ] Remove uses of `knex.transaction`
- [ ] Remove uses of `Channel.transaction`

lalexgap:
Adds some basic transaction methods to Store and changes existing code to use those to keep things consistent.

Fixes #2372

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -13,7 +13,7 @@ import {
   serializeAllocation,
   Outcome,
 } from '@statechannels/wallet-core';
-import {JSONSchema, Model, Pojo, QueryContext, Transaction, ModelOptions} from 'objection';
+import {JSONSchema, Model, Pojo, QueryContext, ModelOptions} from 'objection';
 import _ from 'lodash';
 import {ChannelResult} from '@statechannels/client-api-schema';
 
@@ -99,8 +99,8 @@ export class Channel extends Model implements RequiredColumns {
 
   static jsonAttributes = ['vars', 'participants'];
 
-  static async forId(channelId: Bytes32, tx: Transaction | undefined): Promise<Channel> {
-    const result = Channel.query(tx)
+  static async forId(channelId: Bytes32): Promise<Channel> {
+    const result = Channel.query()
       .where({channelId})
       .withGraphFetched('signingWallet')
       .first();
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -26,7 +26,6 @@ import {Outgoing, ProtocolAction} from '../protocols/actions';
 import {SigningWallet} from '../models/signing-wallet';
 import {logger} from '../logger';
 import * as Application from '../protocols/application';
-import knex from '../db/connection';
 import * as UpdateChannel from '../handlers/update-channel';
 import * as JoinChannel from '../handlers/join-channel';
 
@@ -60,7 +59,7 @@ export type WalletInterface = {
 
 export class Wallet implements WalletInterface {
   async createChannel(args: CreateChannelParams): SingleChannelResult {
-    return Channel.transaction(async tx => {
+    return Store.transaction(async () => {
       const {participants, appDefinition, appData, allocations} = args;
       const outcome: Outcome = deserializeAllocations(allocations);
       // TODO: How do we pick a signing address?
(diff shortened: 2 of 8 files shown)


packages/server-wallet/src/wallet/store.ts
```diff
@@ -14,19 +13,23 @@ import {
...
 
 import {Channel, SyncState, RequiredColumns} from '../models/channel';
 import {SigningWallet} from '../models/signing-wallet';
 import {addHash} from '../state-utils';
 import {ChannelState} from '../protocols/state';
+import knex from '../db/connection';
 
 export const Store = {
+  transaction: <T>(callback: (transaction: Objection.Transaction) => Promise<T>): Promise<T> =>
+    knex.transaction(callback),
```
kerzhner, line 26: If I am reading this code correctly, this is a single parameter function that calls another single parameter function without modifying the parameter. But perhaps the syntax above a more readable type signature?
```suggestion
  transaction: knex.transaction,
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -14,19 +13,23 @@ import {
...
+  startTransaction: (): Promise<Objection.Transaction> => knex.transaction(),
```
kerzhner, line 27:
```suggestion
  startTransaction: knex.transaction,
```

packages/server-wallet/src/wallet/store.ts
lalexgap: That's a good point, I think we just need ```transaction: knex.transaction```

packages/server-wallet/src/wallet/store.ts
lalexgap: Removed `startTransaction` as it would be identical to `transaction`

packages/server-wallet/src/models/channel.ts
```diff
@@ -99,8 +99,8 @@ export class Channel extends Model implements RequiredColumns {
 
   static jsonAttributes = ['vars', 'participants'];
 
-  static async forId(channelId: Bytes32, tx: Transaction | undefined): Promise<Channel> {
-    const result = Channel.query(tx)
+  static async forId(channelId: Bytes32): Promise<Channel> {
+    const result = Channel.query()
```
andrewgordstewart, line 103: Why was the `tx` argument removed from this function?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -160,8 +161,8 @@ export class Wallet implements WalletInterface {
   }
 
   async pushMessage(message: AddressedMessage): MultipleChannelResult {
-    const channelIds = await Channel.transaction(async tx => {
-      return await Store.pushMessage(message, tx);
+    const channelIds = await Channel.transaction(async () => {
+      return await Store.pushMessage(message);
     });
```
andrewgordstewart, line 166: The transaction here is not meaningful, since `Store.pushMessage` would execute in separate transactions.

This might be confusing coming from the browser wallet, but Dexie uses some magical promise-local data to keep track of _the_ current transaction within the execution of some async code.

Objection doesn't do this, so if you want to execute some queries within the context of a transaction, you need a handle on that transaction. IE. you are allowed to execute a concurrent transaction, if you wish. You can see an example of that [here](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/models/__test__/channel.test.ts#L32-L43).

packages/server-wallet/src/wallet/store.ts
```diff
@@ -19,14 +18,15 @@ import {Channel, SyncState, RequiredColumns} from '../models/channel';
 import {SigningWallet} from '../models/signing-wallet';
 import {addHash} from '../state-utils';
 import {ChannelState} from '../protocols/state';
+import knex from '../db/connection';
 
 export const Store = {
```
snario, line 23: As a minor comment, we should perhaps rename this `Store` to `PostgreSQLStore` if we're trying to be deliberate about defining and reusing this abstraction in other contexts like the browser.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -78,30 +77,32 @@ export class Wallet implements WalletInterface {
...
       return {outbox: outgoing.map(n => n.notice), channelResult};
     });
   }
 
   async joinChannel({channelId}: JoinChannelParams): SingleChannelResult {
-    const {outbox, channelResult} = await knex.transaction(
-      async (tx): Promise<{outbox: any; channelResult: ChannelResult}> => {
-        const channel = await Store.getChannel(channelId, tx);
+    const {outbox, channelResult} = await Store.transaction(
+      async (): Promise<{outbox: any; channelResult: ChannelResult}> => {
```
snario, line 96: I believe `{outbox: any; channelResult: ChannelResult}` is defined elsewhere

lalexgap:
> Correct me if I'm wrong, but I think these callbacks are not actually using the transaction that gets started. I think the `..., tx)` is actually important. (This is one reason why the `Store` in the browser wallet is not very portable to a non-browser context.)

I think you're correct. In my mind I was assuming that queries would use whatever ambient transaction was started but it looks like `knex` doesn't follow that pattern.

PR #2395, opened 2020-07-31
Standardize internal wallet errors

kerzhner:
When a client calls a wallet api (such as `updateChannel`), either:
1. the call succeeds
2. the call fails for a reason the wallet understands. The wallet can then return an error to the api caller explaining why the call failed.
3. the call fails but the wallet expected the call to succeed.

The idea is to always capture errors at the outermost layer of the wallet. It can look something like: 
``` typescript
try {
  // perform a few operations that might throw an error
} catch (err) {
 if(!isWalletError(err)) throw err;
 // check for known error cases. Repackage the error to return a `client-api-schema` error.
 // if error does not match know errors, throw err
}
```

Example of an error the wallet understands: `updateChannel` api is called with a `channelId` that is not in the store.
Example of an unexpected error: `updateChannel` api is called but the execution fails somewhere with something like a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError)


packages/server-wallet/src/models/channel.ts
```diff
@@ -325,39 +326,40 @@ export class Channel extends Model implements RequiredColumns {
...
-  notMyTurn = "Cannot update channel unless it's your turn",
-}
-
-class ChannelError extends Error {
-  readonly type = 'ChannelError';
-  constructor(reason: Errors, public readonly data: any = undefined) {
+export class ChannelError extends WalletError {
+  readonly type = WalletError.errors.ChannelError;
+
+  static readonly reasons = {
```
andrewgordstewart, line 332: Should we take the opportunity to prune these at all?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -106,22 +107,22 @@ export const Store = {
   },
 };
 
-class StoreError extends Error {
-  readonly type = 'InvariantError';
-  constructor(reason: StoreErrors, public readonly data: any = undefined) {
+class StoreError extends WalletError {
+  readonly type = WalletError.errors.InvariantError;
```
andrewgordstewart, line 111: A bit off topic, but `InvariantError` doesn't quite seem right. Is "not being a participant in the channel" an invariant error?

packages/server-wallet/src/handlers/join-channel.ts
```diff
@@ -29,11 +34,13 @@ export class JoinChannelError extends Error {
...
 const ensureNotSignedByMe = (cs: ChannelState): StepResult =>
-  hasStateSignedByMe(cs) ? left(new JoinChannelError(Errors.alreadySignedByMe)) : right(cs);
+  hasStateSignedByMe(cs)
+    ? left(new JoinChannelError(JoinChannelError.reasons.alreadySignedByMe))
+    : right(cs);
 
 function ensureLatestStateIsPrefundSetup(cs: ChannelState): StepResult {
   if (cs.latest.turnNum === 0) return right(cs);
-  return left(new JoinChannelError(Errors.invalidTurnNum));
+  return left(new JoinChannelError(JoinChannelError.reasons.invalidTurnNum));
```
andrewgordstewart, line 43: The only thing that would make this better is if you could do
```suggestion
  return left(JoinChannelError.invalidTurnNum());
```

kerzhner: Agreed. We can definitely create these factories at least for often instantiated errors.

packages/server-wallet/src/models/channel.ts
kerzhner: Good call: 3a8f3a169

packages/server-wallet/src/wallet/store.ts
kerzhner: Renaming to `StoreError` for now: 2c7e4b77a

PR #2402, opened 2020-07-31
Improve lockApp error handling

andrewgordstewart:
It seems like the default behaviour when a channel is missing should be
to throw a generic ChannelMissing error. This attempts to minimize the
chance that a missing channel goes unhandled.

This allows you to pass a custom missing channel handler, in case you
want to throw a different error.

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -93,15 +93,11 @@ export class Wallet implements WalletInterface {
   async joinChannel({channelId}: JoinChannelParams): SingleChannelResult {
     const {outbox, channelResult} = await Store.lockApp(
       channelId,
-      async (tx): Promise<{outbox: any; channelResult: ChannelResult}> => {
-        const channel = await Store.getChannel(channelId, tx);
-
+      async (tx, channel): Promise<{outbox: any; channelResult: ChannelResult}> => {
         if (!channel)
           throw new JoinChannel.JoinChannelError(
             JoinChannel.JoinChannelError.reasons.channelNotFound,
-            {
-              channelId,
-            }
+            {channelId}
           );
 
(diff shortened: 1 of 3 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -91,36 +91,37 @@ export class Wallet implements WalletInterface {
...
+    const criticalCode: AppHandler<SingleChannelResult> = async (tx, channel) => {
+      const nextState = getOrThrow(JoinChannel.joinChannel({channelId}, channel));
+      const {outgoing, channelResult} = await Store.signState(channelId, nextState, tx);
+      return {outbox: outgoing.map(n => n.notice), channelResult};
+    };
 
-        const nextState = getOrThrow(JoinChannel.joinChannel({channelId}, channel));
-        const {outgoing, channelResult} = await Store.signState(channelId, nextState, tx);
+    const handleMissingChannel: MissingAppHandler<SingleChannelResult> = () => {
+      throw new JoinChannel.JoinChannelError(JoinChannel.JoinChannelError.reasons.channelNotFound, {
```
kerzhner, line 101: Should `JoinChannelError.reasons.channelNotFound` and `UpdateChannelError.reasons.channelNotFound` be distinct errors? Or should there be a canonical `channelNotFound` error?

Also, it looks like the `Wallet` will throw an error when the `joinChannel` api is called. That's probably fine for now. But in the long term, we should probably return error responses from wallet apis?

andrewgordstewart: I think you're in charge of that ;)

While the answer is unclear, this approach seems ok. It seems like there should be a distinct join channel error for the `joinChannel` _api call_, but there probably doesn't need to be a distinct, internal, channel missing error for the `joinChannel` handler.

PR #2409, opened 2020-08-03
Add getChannels implementation to server-wallet

snario:


packages/server-wallet/src/wallet/store.ts
```diff
@@ -94,6 +94,10 @@ export const Store = {
     return (await Channel.forId(channelId, tx))?.protocolState;
   },
 
+  getChannels: async function(tx: Transaction | undefined): Promise<Channel[]> {
```
andrewgordstewart, line 97: Maybe this should return a `Promise<ChannelResult[]>`, if the sole intention is to get a list of channels for the app.

A better signature would be `Promise<ChannelState[]>`, since there should be a function that converts a `ChannelState` to a `ChannelResult`.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -138,8 +138,13 @@ export class Wallet implements WalletInterface {
...
     throw 'Unimplemented';
   }
+
   async getChannels(): MultipleChannelResult {
-    throw 'Unimplemented';
+    const channels = await Channel.transaction(async tx => await Store.getChannels(tx));
+    return {
+      channelResults: channels.map(channel => channel.channelResult),
+      outbox: [],
+    };
```
andrewgordstewart, line 147: A transaction isn't needed here, since `getChannels` only involves a single query.

If you make the change suggested below, this can then be

```suggestion
    return { channelResults: await Store.getChannels(), outbox: [] };
```

packages/server-wallet/src/wallet/store.ts
snario: Yes you're right

snario: 2729767ed

packages/server-wallet/src/wallet/index.ts
snario: 90498f815

PR #2413, opened 2020-08-04
[client-api-schema, iframe-channel-provider] Cleanup types

geoknee:
One thing I find particularly confusing about the types in this package is not knowing when types are entirely generic (on the one hand), or aware of our specific state channels methods and parameters (on the other). This PR tries to improve that, as well as deduplicating types that were also in `iframe-channel-provider`. 

Changes required follow-on work in `xstate-wallet`, `server-wallet`, too. 

- JsonRpc header types remain generic and ignorant of our application-specific types, and now live in schema package exclusively (not in channel provider anymore)
- Our application-specific types are all prepended with StateChannels, e.g. StateChannelsRequest
- parseNotification implemented and used in iframe-channel-provider
- request ids made mandatory at the JSONRPC level, optional at the `channelprovider.request` level
- Listener callbacks narrowed to `StateChannelsResponse` type


packages/iframe-channel-provider/src/postmessage-service.ts
```diff
@@ -58,21 +63,24 @@ export class PostMessageService {
...
     // message ids.
-    if (message.id) logger.error('message id should not be defined');
+    if (messageWithOptionalId.id) logger.error('message id should not be defined');
 
     // message IDs should be unique
-    message.id = message.id || this.requestNumber++;
+    const message: JsonRpcRequest = {
+      ...messageWithOptionalId,
+      id: messageWithOptionalId.id || this.requestNumber++
+    };
```
geoknee, line 78: This approach allows the `JsonRpcRequest` type to insist on the `id` field being specified.

packages/iframe-channel-provider/src/types/index.ts
```diff
@@ -1,4 +1,3 @@
 export * from './providers';
 export * from './events';
-export * from './jsonrpc';
```
geoknee, line 3: 👋 bye bye

packages/client-api-schema/src/jsonrpc-header-types.ts
```diff
@@ -95,18 +110,31 @@ export interface JsonRpcNotification<NotificationName = string, NotificationPara
...
 /**
- * Specifies error headers as per {@link https://www.jsonrpc.org/specification | JSON-RPC 2.0 Specification }
+ * Type guard for {@link JsonRpcRequest | JsonRpcRequest}
+ *
+ * @returns true if the message is a JSON-RPC request, false otherwise
  * @beta
  */
-export type JsonRpcErrorResponse = ErrorResponse;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function isJsonRpcRequest(message: any): message is JsonRpcRequest {
```
snario, line 119: If this really is `any` maybe we should check for `undefined`? Unless we can expect it is at least `object`

PR #2416, opened 2020-08-04
Add joinChannel call to Ping Pong test

snario:


packages/server-wallet/e2e-test/e2e.test.ts
```diff
@@ -10,7 +10,7 @@ import knexPing from '../src/db/connection';
 import PingClient from './ping/client';
 import {killServer, startPongServer, waitForServerToStart, PongServer, knexPong} from './e2e-utils';
 
-jest.setTimeout(10_000); // Starting up Pong server can take ~5 seconds
+jest.setTimeout(20_000); // Starting up Pong server can take ~5 seconds
```
snario, line 13: I was coding this while on a Zoom call so I bumped this.

packages/server-wallet/e2e-test/pong/controller.ts
```diff
@@ -49,18 +49,11 @@ export default class PongController {
...
-        recipient: from,
-        sender: this.myParticipantID,
-        data: {signedStates: [], objectives: []},
-      };
-    }
-
     const {
       outbox: [messageToSendToPing],
-    } = await this.wallet.updateChannel(channelResult);
+    } = await (channelResult.turnNum < 4 ? this.wallet.joinChannel : this.wallet.updateChannel)(
```
andrewgordstewart, line 54: Does `joinChannel` actually get called in `e2e-test.ts` as a result of this change?

snario: Yeah, but the `turnNum` does not increase

andrewgordstewart: You should still be able to assert that the channel now has `{supported: {turnNum: 0}}`, instead of `{supported: undefined}`

snario: Good point, I'll add this.

packages/server-wallet/e2e-test/e2e.test.ts
```diff
@@ -55,15 +55,15 @@ describe('e2e', () => {
...
     expect(channel.participants).toStrictEqual([ping, pong]);
-    expect(channel.status).toBe('opening');
+    expect(channel.status).toBe('funding');
     expect(channel.turnNum).toBe(0);
+
+    const inspectedChannel = await Channel.query()
+      .where({channelId: channel.channelId})
+      .first();
+
+    expect(inspectedChannel.supported).toBeDefined();
```
andrewgordstewart, line 66: Something like this is a bit more precise?

```suggestion
    expect((await Channel.forId(channel.channelId)).protocolState).toMatchObject({ supported: { turnNum: 0 } });
```

packages/server-wallet/src/models/channel.ts
```diff
@@ -96,7 +96,7 @@ export class Channel extends Model implements RequiredColumns {
 
   static jsonAttributes = ['vars', 'participants'];
 
-  static async forId(channelId: Bytes32, tx: Transaction | undefined): Promise<Channel> {
+  static async forId(channelId: Bytes32, tx?: Transaction): Promise<Channel> {
```
andrewgordstewart, line 99: I had hoped that making `tx` a required argument would defend against developers unknowingly querying for data outside the context of a transaction. If you are performing a transaction, and you wrote to channel `'0xf00'`, then `Channel.forId('0xf00', undefined)` would return different data than `Channel.forId('0xf00', ongoingTx)`

snario: I see, okay will revert

PR #2417, opened 2020-08-04
[server-wallet/e2e-test] Run ping client in a separate process

andrewgordstewart:
- Creates a `ping.ts` script which accepts command line arguments, using yargs
- `e2e.test.ts` spawns a `ping` process, rather than creating a ping client within the same process as the jest test runnier

This makes it easier to create v1 & v2 stress tests: we can spawn multiple `ping` scripts to represent different actors. They can have different databases (different indexers) or the same database (simulating horizontal scaling).

### DIFF
--- a/packages/server-wallet/e2e-test/e2e.test.ts
+++ b/packages/server-wallet/e2e-test/e2e.test.ts
@@ -1,6 +1,7 @@
 import {Participant} from '@statechannels/client-api-schema';
 
 import {alice, bob} from '../src/wallet/__test__/fixtures/signing-wallets';
+import {alice as aliceP, bob as bobP} from '../src/wallet/__test__/fixtures/participants';
 import {Channel} from '../src/models/channel';
 import {withSupportedState} from '../src/models/__test__/fixtures/channel';
 import {SigningWallet} from '../src/models/signing-wallet';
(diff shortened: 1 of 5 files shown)


packages/server-wallet/e2e-test/e2e.test.ts
```diff
@@ -65,31 +78,60 @@ describe('e2e', () => {
...
+      latest: {turnNum},
+    });
 
-    const {channelId} = seed;
+  await expectSupportedState(ChannelPing, 3);
+  await expectSupportedState(ChannelPong, 3);
+  // END SETUP
+
+  // SCRIPT
+  const pingScript = childProcess.spawn(`yarn`, [
```
lalexgap, line 109: This could probably be moved to a function in `e2e-utils`

snario:
General comment, I don't think we should access the database of either client directly except for seeding in the tests if possible. To better emulate how these clients would work in a real use case they should call the wallet API to inspect its state.

andrewgordstewart:
> General comment, I don't think we should access the database of either client directly except for seeding in the tests if possible. To better emulate how these clients would work in a real use case they should call the wallet API to inspect its state.

While I agree, it is not very feasible right now, since models are globally bound to a single knex within the context of the jest suite.

packages/server-wallet/e2e-test/e2e.test.ts
andrewgordstewart: Done: https://github.com/statechannels/statechannels/pull/2417/commits/69f358db00b84196bd84518ac143a2fffd6c4065

packages/server-wallet/e2e-test/e2e.test.ts
```diff
@@ -65,8 +79,40 @@ describe('e2e', () => {
...
+    const pingScript = childProcess.spawn(`yarn`, args);
+    pingScript.on('error', logger.error);
+
+    await new Promise(resolve => {
+      pingScript.on('exit', _data => resolve());
+
+      // FIXME: I would expect `.on('exit', resolve)` to work, but it does not,
+      // jest warns 'Jest did not exit one second after the test run has completed.'
+      // This indicates to me that the spawned process does not exit, even after logging 'DONE'
+      pingScript.stdout.on('data', data => data.toString() === 'DONE\n' && resolve());
```
lalexgap, line 108: :( If we have to use `DONE` as a magic keyword could we make it a constant? I'm hoping we can figure out a better way to handle this.

andrewgordstewart: Absolutely. The best thing is to sort out what's going on, though.

andrewgordstewart: I think this will resolve it. (It does locally)

https://github.com/statechannels/statechannels/pull/2417/commits/36041ad2a62198a17db2691555f93b6d4a06626a

PR #2420, opened 2020-08-04
Rename ping to payer, pong to receiver

lalexgap:
Renames ping and pong to payer and receiver to keep them things clear.


lalexgap:
> The following files still contain `ping` and `pong`:
> 
> * e2e-test/e2e.test.ts
> * e2e-test/README.md

Fixed!

.circleci/config.yml
```diff
@@ -250,10 +250,10 @@ jobs:
           working_directory: packages/server-wallet
           command: |
             sudo apt install -y postgresql-client || true
-            createdb ping $PSQL_ARGS
-            SERVER_DB_NAME=ping yarn db:migrate
-            createdb pong $PSQL_ARGS
-            SERVER_DB_NAME=pong yarn db:migrate
+            createdb payer $PSQL_ARGS
```
andrewgordstewart, line 253: Outside the scope of this PR, but this seems like a bad name for a database. It's plausible that I might already have a database called `payer`. I think it's sufficiently unlikely that I have one named `sc_e2e_payer`.

packages/server-wallet/e2e-test/README.md
```diff
@@ -11,35 +11,35 @@ Channel | |    | |              Channel | |    | |
         | v    | v                      | v    | v
       +-+-+----+-+--+                 +-+-+----+-+--+
       |             +-------2-------->+             |
-      | Ping Client |  POST /inbox    | Pong Client |
+      | Payer Client |  POST /inbox    | Receiver Client |
```
andrewgordstewart, line 14: I don't know how these diagrams were generated, but now they're misaligned :P

snario:
I think the outcomes are still set as 0, so there aren't really any payments being made or received yet?

packages/server-wallet/e2e-test/README.md
lalexgap: Manually fixed the diagram for now

lalexgap:
> I think the outcomes are still set as 0, so there aren't really any payments being made or received yet? For context, this is why I named them ping and pong originally, as all they do is sign and send states.

Yup, you're right! Although I think `makePayment` makes it clearer what we hope to do (and the fact that we're not updating the outcome probably doesn't matter too much at the moment)

PR #2421, opened 2020-08-05
Refactor application protocol

andrewgordstewart:
This refactors the application protocol to
- not use `match`, but instead find an exact match
- not to use `Either` or `Option`, but instead throw errors and return `undefined` in place of a `None`

Then, changes were made to the application protocol & pushMessage in order to send out a `ChannelProposed` notification on receiving the first app prefund setup state.


packages/server-wallet/src/protocols/actions.ts
```diff
@@ -25,25 +21,14 @@ export type SubmitTransaction = {
...
-});
-
-export const signStateProtocolResult = (props: Omit<SignState, 'type'>): ProtocolResult =>
-  right(some(signState(props)));
+export const noAction = undefined;
+
+const actionConstructor = <A extends ProtocolAction = ProtocolAction>(type: A['type']) => (
+  props: Omit<A, 'type'>
+): A => ({...props, type} as A);
+export const submitTransaction = actionConstructor<SubmitTransaction>('SubmitTransaction');
```
kerzhner, line 29: That's kind of verbose... oh well.

andrewgordstewart: It's type safe and has auto completion!

PR #2422, opened 2020-08-05
[monorepo] Move to lockstep versioning

geoknee:
`npx lerna-dependency-graph -o graph.png -f png` produces:
graph

I propose that locking versions across all the packages in this repo will make controlling versions about 10x easier. 

This changeset also 

- allows the documentation website to have a version which matches the "monorepo version". When we run `lerna version`, the docs on `master` are snapshotted, and checked into git as historical files which should not ordinarily be changed
- allows for automatically generated git tags https://github.com/statechannels/statechannels/releases/tag/v0.2.0


readme.md
```diff
@@ -97,6 +97,24 @@ To run all tests:
...
+to increase the version in all of them (this works for the docs website, too).
+
+Use
+
+```shell
+yarn lerna publish from-git
+```
+
+to publish the packages to npm (you will need to be logged in with `npm login`), as well as pushing the version update to git.
+
```
geoknee, line 117: Suggested publishing flow.

package.json
```diff
@@ -63,9 +63,6 @@
     "packages": [
       "packages/*",
       "packages/docs-website/*"
-    ],
-    "nohoist": [
-      "**/ethers"
```
geoknee, line 68: I had to remove this in order to prevent an error in `yarn`. 

https://github.com/yarnpkg/yarn/issues/7807

I didn't see any obvious problems arising from removing these lines...

geoknee: Rolled back in 9430c05 ; the problem stems from having a different version of devtools specified at the top level package.json (not ideal to have to manually change this...)

packages/docs-website/website/versioned_docs/version-0.2.0/app-devs/hook-up-messaging.md
```diff
@@ -0,0 +1,48 @@
+---
```
andrewgordstewart, line 1: I'm curious about this changeset, since the LOC added in this PR is very large, and it seems like we might have a similar changeset when we bump the version.

Suppose we bump to version `0.2.1`. Does this file get deleted? Or is there a new folder `version-0.2.1` added?

Do we need to check in these `.md` files? It seems like they are _built_ from the existing code, similar to the `js` files that get distributed. But we don't check in the transpiled `.js` files.

andrewgordstewart: I guess that the docs that are _hosted_ are using the files checked into this repository. Is that true? If so, that would explain why they should be checked in. But our repo is already 78mb, which is _large_ for a git repo, and this appears to worsen the problem. In that case, perhaps we could look into using "[git large file storage](https://git-lfs.github.com/)"

geoknee: > I'm curious about this changeset, since the LOC added in this PR is very large, and it seems like we might have a similar changeset when we bump the version.

This changeset is particularly large because it stores a version of the docs that include many changes, mostly introducing the typescript API section of the website. Future version bumps will likely result in much smaller changesets  https://docusaurus.io/docs/en/versioning#fallback-functionality. 

> Suppose we bump to version 0.2.1. Does this file get deleted? Or is there a new folder version-0.2.1 added?

A new folder will be added. 

> Do we need to check in these .md files? It seems like they are built from the existing code, similar to the js files that get distributed. But we don't check in the transpiled .js files.

The versioned_docs folder contains historical docs that cannot easily be generated from the existing code (i.e. at the existing commit). The current docs folder is not checked in, as that _is_ generated from the existing code. 

> I guess that the docs that are hosted are using the files checked into this repository. Is that true?

Yes.

> But our repo is already 78mb, which is large for a git repo, and this appears to worsen the problem. 

The docs seem to be less than 0.5mb, (which is less than 1% of 75MB) so I think this is probably tolerable for now. It will obviously increase with the number of versions we create, but if it starts becoming a real issue we can drop some of the old versions. 

>   perhaps we could look into using "git large file storage"

I don't think we have particularly large files. We have a large number of small files.

andrewgordstewart: > The docs seem to be less than 0.5mb, (which is less than 1% of 75MB) so I think this is probably tolerable for now. It will obviously increase with the number of versions we create, but if it starts becoming a real issue we can drop some of the old versions.

Dropping the old versions don't remove them from the _repository_, just from the _codebase_. For instance, we have much less than 78mb of code checked into the codebase. The repository includes enough information to reproduce the codebase at any point in time, so dropping the old versions won't reduce the size of the repo. Thus, I was worrying out loud that each version bump might monotonically increase the size of the repo ~0.5-1%, which is not a great habit.

> I don't think we have particularly large files. We have a large number of small files.

_Technically_ [everything is a file](https://en.wikipedia.org/wiki/Everything_is_a_file) :P But pendantry aside, I think it is still simple to point git large file storage at a single directory.

PR #2434, opened 2020-08-05
Move stress test to jest test

### LINKED ISSUE #2433: Move stress test to jest test and add assertions
Running the stress test in the jest test will allow us to assert things and make sure nothing breaks

lalexgap:
Fixes #2433 

Moves the stress test into a jest test instead of running it as a node script. This allows us to easily assert things.


packages/server-wallet/e2e-test/e2e-utils.ts
```diff
@@ -1,16 +1,36 @@
...
+): Promise<void> => {
+  let args = ['start', '--database', 'payer', '--channels', ...channelIds];
+
+  if (numPayments) args = args.concat(['--numPayments', numPayments.toString()]);
+
+  const payerScript = fork(join(__dirname, './payer/index.ts'), args, {
+    execArgv: ['-r', 'ts-node/register'],
+  });
+  payerScript.on('message', message =>
+    console.log(PerformanceTimer.formatResults(JSON.parse(message as any)))
```
andrewgordstewart, line 29: Future work: in addition to logging to the console, store it as an artifact. They can then be tracked over time, or referenced eg. as a status check.

PR #2436, opened 2020-08-06
[xstate-wallet] Allow yarn start to work

geoknee:
Currently, `yarn start` fails citing

- env vars not set
-  `webpack-dev-server` not found. 
 
Also: 

- cannot use 'in' operator to search for `x` in `y`


.env.virtual-funding
```diff
@@ -0,0 +1,18 @@
+## shared between multiple packages
```
snario, line 1: Why add a new monorepo env file?

packages/xstate-wallet/package.json
```diff
@@ -72,6 +72,7 @@
     "url-loader": "3.0.0",
     "url-parse": "1.4.7",
     "webpack": "4.41.5",
+    "webpack-dev-server": "^3.11.0",
```
snario, line 75: Surprised this wasn't here 😳

.env.virtual-funding
geoknee: It allows us to do `SC_ENV=virtual-funding yarn start` (there are no env files checked into this repo at present).

packages/xstate-wallet/package.json
geoknee: I'm thinking we need an integration test in this repo to check for this stuff before we publish it. There are other things that are currently broken.

.env.virtual-funding
snario: What's the objective though? I'm not aware of any integration test which uses multiple services in the repo atm

geoknee: I propose that we _should_ have an integration test in this repo (see my other comment).  If you like I can save the addition of this file until I make that PR?

package.json
```diff
@@ -33,7 +33,10 @@
     "@types/react": "16.9.19",
     "broccoli": "3.3.3",
     "ethereum-transaction-debugger": "https://registry.yarnpkg.com/@favware/skip-dependency/-/skip-dependency-1.0.2.tgz",
-    "scrypt": "https://registry.yarnpkg.com/@favware/skip-dependency/-/skip-dependency-1.0.2.tgz"
+    "scrypt": "https://registry.yarnpkg.com/@favware/skip-dependency/-/skip-dependency-1.0.2.tgz",
+    "**/react": "16.8.0",
+    "**/react-dom": "16.8.0",
+    "@babel/core": "7.8.3"
```
geoknee, line 39: I think maybe hoisting our docs website up to be a new package has led to multiple versions of some of these packages in our node_modules. This deduplicates them.

.env.virtual-funding
geoknee: Progress on such a test over at #2437

geoknee:
Subsumed in #2437

PR #2437, opened 2020-08-06
[integration-tests*] Add integration-tests package

geoknee:
This package is a test suite that runs in jest's jsdom environment. It allows us to checksif the channel client, provider and wallet all work together.

### Advantages
- runs in jest in 6 seconds, limited (I think) by downloading the wallet bundle
- no puppeteer necessary
- no explicit app necessary (no extra server, html etc) it can just be the `channel-client` or some class that wraps it
- allowed me to catch some things that were unwittingly broken recently (fixed on this branch)
- should be able to call `channel-client` methods, and make assertions about the responses / state of the` channel-client.`

### Hacks
- injecting web3 (similar to what we did in puppeteer)
- polyfill for window.crypto (does not exist in jsdom, but ethers needs it).
- ~needed a way to bypass the first user click to enable ethereum~
- I couldn't find a way to easily do these hacks inside the test script, so I had to conditionally modify the wallet itself 😟 

### Limitations
- ~difficult to see how to test any workflow that involves _user mouse clicks_ can be encapsulated by a test like this. The cross-origin security properties we enjoy mean we would probably need to go back to puppeteer for that.~
- ~OTOH: 💡  what if we  use the _same origin_ for app and wallet, in the tests? i.e. embed the wallet in an iframe on the page where the wallet lives?~
- difficult to see how to integrate multiple apps and wallets that communicate (again we could go back to puppeteer). It's probably doable if we can put multiple iframes on one page (we might need to take a hacksaw to the `channel-provider` to do that)

### TODO
- [x] integrate into a circle job (using something like `start-servers`)
- [x] expand coverage to `channel-client.createChannel`
- [x] fix problem with `jest-canvas-mock`


packages/channel-client/jest.config.js
```diff
@@ -5,5 +5,7 @@ module.exports = {
     }
   },
   preset: 'ts-jest',
-  testMatch: ['**/__tests__/**/*.ts?(x)', '**/?(*.)+(spec|test).ts?(x)']
+  testMatch: ['**/__tests__/**/*.ts?(x)', '**/?(*.)+(spec|test).ts?(x)'],
+  testEnvironment: 'jsdom',
+  testEnvironmentOptions: {url: 'http://bar.com/', resources: 'usable'}
```
snario, line 10: ?

packages/channel-client/tests/client-provider-wallet.test.ts
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable no-undef */
+
+import {IFrameChannelProviderInterface} from '@statechannels/iframe-channel-provider';
+
+jest.setTimeout(30000);
+require('@statechannels/iframe-channel-provider');
```
snario, line 6: ?

packages/xstate-wallet/public/index.html
```diff
@@ -1,7 +1,31 @@
...
+          web3.eth.getAccounts().then(lst => {
+            web3.eth.defaultAccount = lst[0];
+            console.log('Using address ' + web3.eth.defaultAccount);
+            window.ethereum.selectedAddress = web3.eth.defaultAccount;
+            r([window.ethereum.selectedAddress]);
+          });
+        });
+      window.ethereum.networkVersion = 9001;
+      window.ethereum.on = () => {};
+      window.ethereum.enable();
```
snario, line 26: Maybe it should have its own configuration option, and not re-use `WINDOW_CRYPTO_POLYFILL`? e.g., `INJECT_GANACHE_WEB3_POLYFILL`?

packages/channel-client/jest.config.js
geoknee: This is necessary to have jsdom load iframes. We can also set the domain of the 'app' here.

packages/channel-client/tests/client-provider-wallet.test.ts
geoknee: This is the standard way we expect apps to load the wallet?

geoknee:
Turns out you can click on things in the iframe, without needing to match the origin of the windows 🤷 . 

Bit confused as to why this is possible, as apparently jsdom respects CORS https://github.com/jsdom/jsdom/issues/1941.

package.json
```diff
@@ -33,7 +33,10 @@
     "@types/react": "16.9.19",
     "broccoli": "3.3.3",
     "ethereum-transaction-debugger": "https://registry.yarnpkg.com/@favware/skip-dependency/-/skip-dependency-1.0.2.tgz",
-    "scrypt": "https://registry.yarnpkg.com/@favware/skip-dependency/-/skip-dependency-1.0.2.tgz"
+    "scrypt": "https://registry.yarnpkg.com/@favware/skip-dependency/-/skip-dependency-1.0.2.tgz",
+    "**/react": "16.8.0",
+    "**/react-dom": "16.8.0",
+    "@babel/core": "7.8.3"
```
tomclose, line 39: Wondering whether these should be in the integration-test package?

geoknee: No, but we might want to move them inside some of the other packages, e.g. `xstate-wallet`. 

Currently if we run `yarn` from the top level, these resolutions will do their job. But if we `yarn add` from an individual package, I think they will not 😟

PR #2439, opened 2020-08-06
[channel-provider] Restore early return to onMessage

### LINKED ISSUE #2438: [channel-provider] WalletReady message fails parseNotification check
This check was added recently. The WalletReady message is not part of the api schema. 

Solution: add it to the schema.

geoknee:
We expect a certain number of messages which will not be json-rpc related.

Closes #2438

### DIFF
--- a/packages/iframe-channel-provider/src/channel-provider.ts
+++ b/packages/iframe-channel-provider/src/channel-provider.ts
@@ -182,6 +182,9 @@ export class IFrameChannelProvider implements IFrameChannelProviderInterface {
   off: OffType = (method, params) => this.events.off(method, params);
 
   protected async onMessage(event: MessageEvent) {
+    if (!event.data.jsonrpc) {
+      return;
+    }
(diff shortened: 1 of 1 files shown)


packages/iframe-channel-provider/src/channel-provider.ts
```diff
@@ -182,6 +182,9 @@ export class IFrameChannelProvider implements IFrameChannelProviderInterface {
   off: OffType = (method, params) => this.events.off(method, params);
 
   protected async onMessage(event: MessageEvent) {
+    if (!event.data.jsonrpc) {
+      return;
+    }
     const message = parseNotification(event.data); // Narrows type, throws if it does not fit the schema
```
geoknee, line 188: We don't want to reach L188 if the message is not json-rpc, since it might throw an error when trying to parse the message as a notification. 

Non json-rpc messages included some of our logging, and the `Wallet Ready` message.

PR #2443, opened 2020-08-07
Add getParticipant method on server wallet

snario:
We do not know yet what the participant ID is, and the destination should probably not default to the signing address?

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -10,8 +10,11 @@ import {
   calculateChannelId,
   StateVariables,
   ChannelConstants,
+  Participant,
+  makeDestination,
 } from '@statechannels/wallet-core';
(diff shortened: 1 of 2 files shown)


packages/server-wallet/src/wallet/store.ts
```diff
@@ -31,6 +34,16 @@ const throwMissingChannel: MissingAppHandler<any> = (channelId: string) => {
 };
 
 export const Store = {
+  getFirstParticipant: async function(): Promise<Participant> {
+    const signingKey = await SigningWallet.query().first();
+    if (!signingKey) throw new StoreError(StoreError.reasons.missingSigningKey);
+    return {
+      participantId: 'unspecified',
```
andrewgordstewart, line 41: For the time being, we should perhaps continue with making the participant ID equal to the signing address?
```suggestion
      participantId: signingKey.address,
```

PR #2449, opened 2020-08-11
Publish @statechannels/server-wallet

snario:


packages/server-wallet/package.json
```diff
@@ -57,7 +57,6 @@
   },
   "main": "lib/src/wallet/index.js",
   "types": "lib/src/wallet/index.d.ts",
-  "private": true,
```
geoknee, line 60: Why remove this @snario ?

PR #2453, opened 2020-08-11
[server-wallet] Add getSigningAddress to Wallet api

kerzhner:
`getSigningAddress` has the following behavior:
- if at least one signing address exists, return it.
- otherwise, create a signing address (and private key), store it, and return it.

### DIFF
--- a/packages/server-wallet/src/wallet/__test__/signing-address.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/signing-address.test.ts
@@ -1,4 +1,5 @@
 import {ethers} from 'ethers';
+import {Transaction} from 'objection';
 
 import {truncate} from '../../db-admin/db-admin-connection';
 import knex from '../../db/connection';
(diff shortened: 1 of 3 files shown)


kerzhner:
On second thought, I don't think the transactions (as used) give protection against generating duplicate signing addresses given the [default Postgres isolation level of Read Committed](https://www.postgresql.org/docs/9.5/transaction-iso.html#:~:text=Read%20Committed%20is%20the%20default,query%20execution%20by%20concurrent%20transactions.)

kerzhner:
> On second thought, I don't think the transactions (as used) give protection against generating duplicate signing addresses given the [default Postgres isolation level of Read Committed](https://www.postgresql.org/docs/9.5/transaction-iso.html#:~:text=Read%20Committed%20is%20the%20default,query%20execution%20by%20concurrent%20transactions.)

Fixed in fb1c07f14.

packages/server-wallet/src/wallet/__test__/signing-address.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import {seedAlicesSigningWallet} from '../../db/seeds/1_signing_wallet_seeds';
+import {SigningWallet} from '../../models/signing-wallet';
+
+import {alice} from './fixtures/participants';
+
+beforeEach(async () => {
+  await truncate(knex);
+});
+
+describe('signingAddress', () => {
```
andrewgordstewart, line 15: Can you add a test, which might flicker, but almost certainly won't, that makes sure that it can't add two signing addresses?

Perhaps even better: what about adding a DB constraint that ensures there is at most one signing address? One easy way is to add a column whose value must be `0`, and add a uniqueness constraint on that column. With a tactic like that, we don't need to use transactions -- you can just try-catch inserting a random wallet, and on that constraint violation, fetch the wallet that you now know exists.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -74,6 +74,12 @@ export class Wallet implements WalletInterface {
     return participant;
   }
 
+  public async getSigningAddress(): Promise<string> {
```
andrewgordstewart, line 77: I noticed that the wallet already has a `getParticipant` method. I think that method should just create a signing wallet if none exists. That removes the need to expose `getSigningAddress`.

Even better: the wallet needs to know what my participant Id is. So, the application should call something like `setParticipant({participantId: string, destination: string})` and it should:
- error if that participant is already set (uniqueness constraint on `participantId`
- create a random signing key for that participant, save that in a `SigningWallet`, and also save the participantId/destination somewhere. (Maybe the `SigningWallet` model is actually a `Participant` model?)

PR #2459, opened 2020-08-13
Fetch nonce before creating channel transaction

andrewgordstewart:
It is dangerous to start a query _outside_ a transaction, while
executing a transaction.

What happens if I call `createChannel` 10 times in a row?
1. I max out the connection pool, consuming 8 connections
2. Every ongoing transaction tries to get a new connection, to start
   a new query outside of that transaction to get the next nonce
3. No transaction finishes, since no connections open up

We've reached a deadlock.

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -81,21 +81,23 @@ export class Wallet implements WalletInterface {
   }
 
   async createChannel(args: CreateChannelParams): SingleChannelResult {
+    const {participants, appDefinition, appData, allocations} = args;
+    const outcome: Outcome = deserializeAllocations(allocations);
+
+    const channelNonce = await Nonce.next(participants.map(p => p.signingAddress));
+    const channelConstants: ChannelConstants = {
+      channelNonce,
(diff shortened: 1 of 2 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -81,21 +81,23 @@ export class Wallet implements WalletInterface {
   }
 
   async createChannel(args: CreateChannelParams): SingleChannelResult {
+    const {participants, appDefinition, appData, allocations} = args;
+    const outcome: Outcome = deserializeAllocations(allocations);
+
+    const channelNonce = await Nonce.next(participants.map(p => p.signingAddress));
```
tomclose, line 87: Just so I understand: by fetching the nonce outside of the transaction, it could be that the transaction later will fail (e.g. if another thread successfully created a channel with that nonce in the meantime?).

andrewgordstewart: `Nonce.next` is guaranteed to return a unique nonce, for the same set of participants! Another thread that called `Nonce.next` with the same participants would get a different nonce, so there is no collision like you describe.

Issue #2460, opened 2020-08-13
Create integration test with two wallets communicating back and forth


lalexgap:
Currently we don't have a test for testing the back and forth communication between wallets. It would probably be a good idea to have a test like that.

andrewgordstewart:
Doesn't the e2e test do that?

lalexgap:
> Doesn't the e2e test do that?

That's a good point, I forgot about that. Let me investigate.

lalexgap:
> > Doesn't the e2e test do that?
> 
> That's a good point, I forgot about that. Let me investigate.

You're right, it looks like it's an issue in our `validateSignatures` function

PR #2461, opened 2020-08-13
Check signature in sign state test

lalexgap:
Fixes https://github.com/statechannels/the-graph/issues/52

Changes the `SignState` test to check for a valid signature. 

Currently we don't have an integration test that tests back and forth between two wallets which is why none of our other tests catch this. I've created #2460 for that.


lalexgap:
Closing for now as I was slightly incorrect in opening #2460

PR #2463, opened 2020-08-13
Fix ValidateSignatures

lalexgap:
Fixes https://github.com/statechannels/the-graph/issues/52

`ValidateSignatures` was relying on the `signer` address on the signature object instead of actually verifying the signature. This fixes `validateSignature` to actually validate the signature provided.


packages/server-wallet/src/wallet/__test__/lock-app.test.ts
```diff
@@ -110,10 +112,7 @@ describe('concurrency', () => {
       );
       const t2 = Date.now();
 
-      // Roughly asserts that the `signState` calls are interwoven
```
andrewgordstewart, line 113: Probably would have been good to leave the reasoning for this timeout :P

PR #2465, opened 2020-08-14
Implement transition validation with Pure Evm

lalexgap:
Implements a transition validator that uses pure EVM. Bytecode is fetched from a new table, and if no bytecode is present validation is skipped.


packages/server-wallet/src/wallet/constants.ts
```diff
@@ -0,0 +1 @@
+export const CHAIN_ID = '0x0';
```
snario, line 1: Why is this a constant?

snario:
I feel like it may be useful to make this a configuration option when instantiating a `Wallet`, e.g., `skipValidationCheck: true`

packages/server-wallet/src/wallet/constants.ts
lalexgap: Eventually the `CHAIN_ID` will be compiled in or fetched at runtime, but for now I think it makes sense to point to a common constant.

snario: It may be good to put in the config file

lalexgap: > I feel like it may be useful to make this a configuration option when instantiating a `Wallet`, e.g., `skipValidationCheck: true`

Good idea, did that here b3e11aa

lalexgap:
> I feel like it may be useful to make this a configuration option when instantiating a `Wallet`, e.g., `skipValidationCheck: true`

Good idea! Did that here b3e11aa

packages/server-wallet/src/wallet/constants.ts
lalexgap: I think that probably makes sense but I think that can be outside the scope of this PR unless anyone has strong feelings about it.

packages/server-wallet/src/__test__/evm-validator.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+it('skips validating when no byte code exists for the app definition', async () => {
+  // Sanity check that the bytecode doesn't exist
+  expect(
+    await AppBytecode.getBytecode(CHAIN_ID, UNDEFINED_APP_DEFINITION, undefined)
+  ).toBeUndefined();
+  const state = createState({
+    appDefinition: UNDEFINED_APP_DEFINITION,
+  });
+  expect(await validateTransitionWithEVM(state, state)).toBe(true);
```
andrewgordstewart, line 54: Should it skip it? Or reject?

packages/server-wallet/src/db/migrations/20200814010138_app_bytecode.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const appBytecode = 'app_bytecode';
+
+export async function up(knex: Knex): Promise<any> {
+  await knex.schema.createTable(appBytecode, function(table) {
+    table.string('chain_id').notNullable();
+
+    table.string('app_definition').notNullable();
+    table.text('app_bytecode').notNullable();
+    table.primary(['chain_id', 'app_definition']);
```
andrewgordstewart, line 11: In my experience, I have only used an auto-incrementing integer, provided by postgres, for the primary key. I'm not 100% sure about mixing primary key strategies. What's the benefit given here? Does it automatically create a uniqueness constraint on `['chain_id', 'app_definition']`?

Not saying we shouldn't do this, it's just something I'm unfamiliar with.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -60,7 +60,20 @@ export type WalletInterface = {
...
 export class Wallet implements WalletInterface {
+  options: WalletOptions = {
+    skipEVMValidation: false,
+  };
+
+  constructor(options?: Partial<WalletOptions>) {
+    if (options) {
+      this.options = {...this.options, ...options};
+    }
+  }
```
andrewgordstewart, line 76:
```suggestion
  constructor(options: Partial<WalletOptions> = {}) {
    this.options = {...this.options, ...options};
  }
```

packages/server-wallet/src/wallet/index.ts
```diff
@@ -104,7 +117,8 @@ export class Wallet implements WalletInterface {
       const {outgoing, channelResult} = await Store.signState(
         channelId,
         {...channelConstants, turnNum: 0, isFinal: false, appData, outcome},
-        tx
+        tx,
+        this && this.options.skipEVMValidation
```
andrewgordstewart, line 121: It seems like the option is only used in one place, and that place could use the global `config` object.

Is it necessary (preferable?) to pass around a `Wallet` option to the store, modifying the signature of many functions?

lalexgap: I think it probably makes sense to just use the global `config` object (I wasn't aware that that was there)

packages/server-wallet/src/db/migrations/20200814010138_app_bytecode.ts
lalexgap: > In my experience, I have only used an auto-incrementing integer, provided by postgres, for the primary key. I'm not 100% sure about mixing primary key strategies. What's the benefit given here? Does it automatically create a uniqueness constraint on `['chain_id', 'app_definition']`?
> 
> Not saying we shouldn't do this, it's just something I'm unfamiliar with.

Yup, it automatically creates a uniqueness constraint on `['chain_id', 'app_definition']` . I think this is the "right" way as the combination of `chain_id` and `app_definition` should always be a unique bytecode entry.

We could use the auto-incrementing integer for the primary key but I think the only benefit is that it's the same thing other tables do.

packages/server-wallet/src/__test__/evm-validator.test.ts
lalexgap: I'm not sure. Right now I've gone with the more permissive option of skipping it but I think the argument could be made that we should reject and rely on setting `skipEVMValidation` in tests and other places.

packages/server-wallet/src/wallet/constants.ts
lalexgap: It looks like there is already a `config` class with ` chainNetworkId` so I should use that instead I think

packages/server-wallet/src/wallet/index.ts
lalexgap: resolved 42f6f58

packages/server-wallet/src/wallet/constants.ts
lalexgap: resolved 6d64209

PR #2466, opened 2020-08-17
Allow node 12.18

geoknee:
Adds node >12.18 to `engines` specification, and uses `12.18` on circle.


package.json
```diff
@@ -17,7 +17,7 @@
   },
   "engines": {
     "yarn": "^1.17.0",
-    "node": "~12.16.0"
+    "node": "^12.16.3 || >12.18"
```
lalexgap, line 20: Minor but could this just be `>=12.16.0`

geoknee, line 20:
```suggestion
    "node": ">=12.16.0"
```

packages/nitro-protocol/package.json
```diff
@@ -41,7 +41,7 @@
   },
   "engines": {
     "yarn": "^1.17.0",
-    "node": "~12.16.0"
+    "node": "^12.16.3 || >12.18"
```
geoknee, line 44:
```suggestion
    "node": ">=12.16.0"
```

packages/wallet-core/package.json
```diff
@@ -20,7 +20,7 @@
     "typescript": "3.7.5"
   },
   "engines": {
-    "node": "~12.16.0"
+    "node": "^12.16.3 || >12.18"
```
geoknee, line 23:
```suggestion
    "node": ">=12.16.0"
```

packages/xstate-wallet/package.json
```diff
@@ -117,7 +117,7 @@
     "wait-for-expect": "3.0.2"
   },
   "engines": {
-    "node": "~12.16.0"
+    "node": "^12.16.3 || >12.18"
```
geoknee, line 120:
```suggestion
    "node": ">=12.16.0"
```

andrewgordstewart:
Is there a benefit to using version 12.18?

geoknee:
@andrewgordstewart yes, some projects we might want to integrate with could be using newer versions of node.

PR #2472, opened 2020-08-18
Accept a funding event in the server wallet

lalexgap:
Adds basic support for sending a funding event to the wallet.

### DIFF
--- a/packages/server-wallet/src/models/funding.ts
+++ b/packages/server-wallet/src/models/funding.ts
@@ -1,0 +1,71 @@
+import {Model, Transaction} from 'objection';
+import {BigNumberish} from 'ethers';
+import {Zero} from '@statechannels/wallet-core';
+
+import {Uint256, Bytes32, Address} from '../type-aliases';
+
+import {Channel} from './channel';
+
+export const REQUIRED_COLUMNS = {
+  channelId: 'channelId',
+  amount: 'amount',
+  assetHolder: 'assetHolder',
+};
+export interface RequiredColumns {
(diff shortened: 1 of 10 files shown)


packages/server-wallet/src/models/funding.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+  readonly channelId!: Bytes32;
+  readonly amount!: Uint256;
+  readonly assetHolder!: Address;
+
+  static async getFundingAmount(
+    channelId: Bytes32,
+    assetHolder: Address,
+    tx: Transaction | undefined
+  ): Promise<BigNumberish> {
```
andrewgordstewart: I don't think we should use `BigNumberish` -- we should return a `Uint256`, which anyone can consume as a `string`.

The only runtime type returned by `getFundingAmount` is a `string`.

packages/server-wallet/src/wallet/__test__/integration/update-funding.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    outbox: [
+      {
+        params: {
+          recipient: 'bob',
+          sender: 'alice',
+          data: {signedStates: [{turnNum: 3}]},
+        },
+      },
+    ],
+    channelResult: {channelId: c.channelId, turnNum: 0}, // The turnNum is coming from the supported state so we expect it be 0 still
```
andrewgordstewart, line 46: It seems like a channel result should have a `funding` property.

PR #2474, opened 2020-08-19
[channel-provider] Ignore messages that don't come from the wallet, and enhance test environment to handle postMessage origin

geoknee:


geoknee:
Tests now failing due to (I think) this issue https://github.com/jsdom/jsdom/issues/2745 😞

andrewgordstewart:
> Tests now failing due to (I think) this issue [jsdom/jsdom#2745](https://github.com/jsdom/jsdom/issues/2745) 😞

In the integration test, is it possible to intercept the event and add `event.origin`? It seems hacky, but it might work?

geoknee:
I'll try!

packages/integration-tests/mount-and-enable.test.ts
```diff
@@ -8,18 +8,52 @@ require('@statechannels/iframe-channel-provider');
...
+  window.addEventListener('message', (event: MessageEvent) => {
+    if (event.origin === '') {
+      event.stopImmediatePropagation();
+      const eventWithOrigin: MessageEvent = new MessageEvent('message', {
+        data: event.data,
+        origin: WALLET_URL
+      });
+      window.dispatchEvent(eventWithOrigin);
+    }
+  });
```
geoknee, line 27: @andrewgordstewart this seems to work.

packages/integration-tests/test-utils.ts
```diff
@@ -0,0 +1,14 @@
...
+    if (event.origin === '') {
+      event.stopImmediatePropagation();
+      const eventWithOrigin: MessageEvent = new MessageEvent('message', {
+        data: event.data,
+        origin
+      });
+      window.dispatchEvent(eventWithOrigin);
+    }
+  });
+}
```
geoknee, line 14: For now this will need to be called in each test file's `beforeAll` block. A nicer way to hide this shim from test authors would be to integrate it into a test environment for jest. https://jestjs.io/docs/en/configuration.html#testenvironment-string

PR #2475, opened 2020-08-19
Speed up signing states

kerzhner:
Fixes https://github.com/statechannels/the-graph/issues/69.

Note: `@bitauth/libauth` breaks `nitro-protocol` webpack due to **javascript** optional chaining. It looks like we'll need to add Babel. Already tried unsuccessfully:
- Upgrade webpack to v5.
- Use `package.json` `resolutions` to force `acorn` to the latest version.


packages/nitro-protocol/src/signatures.ts
```diff
@@ -48,6 +52,21 @@ export function signState(state: State, privateKey: string): SignedState {
   return {state, signature};
 }
 
+export async function fastSignState(
+  state: State,
+  privateKey: string
+): Promise<{state: State; signature: string}> {
+  const wallet = new Wallet(privateKey);
```
andrewgordstewart, line 59: It is also worth investigating how slow this operation is: constructing a `Wallet` will use an `'elliptic'` elliptic curve to get the `address`.

We could use the equivalent function from `libauth` instead.

kerzhner:
> Note: `@bitauth/libauth` breaks `nitro-protocol` webpack due to **javascript** optional chaining. It looks like we'll need to add Babel. Already tried unsuccessfully:
> 
> * Upgrade webpack to v5.
> * Use `package.json` `resolutions` to force `acorn` to the latest version.

cc @lalexgap, the webpack/babel master in case you have spare cycles today/tomorrow.

kerzhner:
> We should definitely rip out `new ethers.Wallet` from `signState`, since it is also a bottleneck.

Agreed! Let's do that in a follow up PR.

packages/nitro-protocol/benchmark/signState.ts
```diff
@@ -0,0 +1,31 @@
+import {Wallet, constants} from 'ethers';
+import {AddressZero} from 'ethers/constants';
+import _ from 'lodash';
+
+import {signState, fastSignState} from '../src/signatures';
+import {State} from '../src';
+
+async function benchmark() {
```
andrewgordstewart, line 8: It's be nice to run these benchmarks in circle. We'd then have a historical reference, and a more consistent environment.

Issue #2432, opened 2020-08-05
Figure out ChannelResults/ChannelNotifications and make them consistent


lalexgap:
[This code](https://github.com/statechannels/statechannels/blob/master/packages/client-api-schema/src/data-types.ts#L106-L125) is not documented, and it's not 100% clear what each status means.

andrewgordstewart:
As far as I can tell, we actually want a data structure more like this:

```
type Variables = {
  allocations: Allocation[];
  appData: string;
  turnNum: Uint48;
};
export type ChannelStatus = {
  type:
    | 'proposed' // MEANS: prefund setup not signed by me
    | 'opening' // MEANS: prefund setup signed by me, but not supported
    | 'funding' // MEANS: prefund setup supported, postfund setup not supported
    | 'running' // MEANS: postfund setup supported
    | 'closing' // MEANS: supported final state
    | 'closed'; // MEANS: conclusion proof
} & Variables;
export type AdjudicatorStatus =
  | {type: 'open'; turnNum: Uint48}
  | ({
      type:
        | 'challenging' // MEANS a challenge is ongoing, and it's not my turn
        | 'responding'; // MEANS a challenge is ongoing, and it's my turn
      challengeExpirationTime: number;
    } & Variables)
  | ({type: 'closed'} & Variables);
export interface ChannelResult {
  channelId: ChannelId;
  participants: Participant[];
  appDefinition: Address;
  channelStatus: ChannelStatus;
  adjudicatorStatus: AdjudicatorStatus;
}
```
## Reasoning
1) When there is a challenge registered on chain, the state might be unrelated to the latest supported state we have (or just the latest state, if we have no supported state)
2) There is a distinction between
  A. "Someone (maybe me) wants to close the channel, because they signed a final state" 
  B. "I have the ability to instantly close the channel, because I have a conclusion proof"
  C. "The channel is finalized on chain. Nothing more can be done"
3) If I receive two results for a channel, one where the turn number is 10 and one where it's 12, I might not actually know which one takes precedence with the current `ChannelResult` interface -- if the turn 10 result has a `'challenging'` status, then it might be more recent than the turn 12 result

All of these issues are dealt with by having orthogonal two orthogonal `ChannelStatus` and `AdjudicatorStatus` properties.

(3) is resolved because this separation allows you to _merge_ results if need be by taking the maximum turn number. This is possible (in the `ChannelStatus`) when the wallet maintains the invariant that there is at most one supported state.

## Open questions
[...]

PR #2489, opened 2020-08-25
Client api docs

LayneHaber:
Provides manual documentation of the JSON RPC schema defined in the `@statechannels/client-api-schema` package.


packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+
+In order to create a virtual or direct channel, you must have first created and funded a ledger channel between client and hub.
+
+A `Participant` contains identifying information about channel members:
+
+| Name           | Type      | Description                                              |
+| -------------- | --------- | -------------------------------------------------------- |
+| participantId  | string    | Unique identifier of channel participant                 |
+| signingAddress | `Address` | Address used to sign channel state updates               |
+| destination    | `Address` | Address to receive funds in case of disputes/withdrawals |
```
snario, line 27: `destination` is of type `Destination`

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| token           | `Address`          | Address of asset on chain                                |
+| allocationItems | `AllocationItem[]` | Assigns some asset to a given user by address and amount |
+
+### Parameters
+
+| Name            | Type                            | Description                                |
+| --------------- | ------------------------------- | ------------------------------------------ |
+| participants    | `Participant[]`                 | Identifying information members of channel |
+| allocations     | `Allocation[]`                  | Array of funding amounts for participants  |
+| appDefinition   | string                          | Address of contract governing the channel  |
```
snario, line 42:
```suggestion
| appDefinition   | Address                          | Address of contract governing the channel  |
```

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| participants            | `Participant[]`   | Identifying information members of channel |
+| allocations             | `Allocation[]`    | Array of funding amounts for participants  |
+| appData                 | string            | Encoded initial state of app               |
+| appDefinition           | string            | Address of contract governing the channel  |
+| channelId               | string            | Unique channel identifier                  |
+| status                  | `ChannelStatus`   | Current status of channel                  |
+| turnNum                 | number            | Channel nonce                              |
+| challengeExpirationTime | number (optional) | Time current challenge on channel elapses  |
+
+where the channel status can be one of:
```
snario, line 63: ref: https://github.com/statechannels/statechannels/issues/2432#issuecomment-669335975

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+
+### Errors
+
+| Code | Message                | Description                           |
+| ---- | ---------------------- | ------------------------------------- |
+| 1000 | SigningAddressNotFound | Unable to find expected ephemeral key |
+| 1001 | InvalidAppDefinition   | App definition not valid address      |
+| 1002 | UnsupportedToken       | Token in allocations not supported    |
+
+TODO: (HIGH) Verify these are all of the expected errors (what if request times out? what if some participant refuses to join the channel?)
```
snario, line 86: Note the above list is probably incomplete. More kinds of errors could exist and do not have codes yet.

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+
+## ApproveBudgetAndFund
```
snario, line 88: Note this method is browser specific, and should exist on top of a browser provider, not as a core wallet API

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| challengeExpirationTime | number (optional) | Time current challenge on channel elapses  |
+
+### Errors
+
+| Code | Message            | Description                                 |
+| ---- | ------------------ | ------------------------------------------- |
+| 1100 | Channel not found  | Could not find channel to update in storage |
+| 1101 | Invalid transition | Channel cannot be joined                    |
+
+## CloseAndWithdraw
```
snario, line 183: This is also probably the wrong API and should be split up into `close`, `defund`, and `withdraw`

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| turnNum                 | number            | Channel nonce                              |
+| challengeExpirationTime | number (optional) | Time current challenge on channel elapses  |
+
+### Errors
+
+| Code | Message           | Description                                    |
+| ---- | ----------------- | ---------------------------------------------- |
+| 1300 | Channel not found | Could not find channel to challenge in storage |
+
+## GetBudget
```
snario, line 266: Browser specific

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| ---------- | ------------- | ---------------------------------------------------------------------- |
+| domain     | string        | The channel the funds are attributed to                                |
+| hubAddress | string        | Signing address of channel intermediary                                |
+| budgets    | TokenBudget[] | An array of approved send and received capacity for a list of channels |
+
+### Errors
+
+TODO: (HIGH) Define errors
+
+## GetChannels
```
snario, line 291: Might make sense to include some query language

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+
+## CreateChannel
+
+This method is used to create a channel between a set of participants with given budgets. It contains information regarding type of channel and initial state of that channel.
+
+There are three different types of channels:
+
+- _Direct Channel_: channels that support applications between ledger channel counterparties
+- _Ledger Channel_: channels that manage the budgets and allocations between direct and virtual channels
+- _Virtual Channel_: channels that support applications routed over ledger channels
```
geoknee, line 17: This is pretty close, but I think it is better to think of channels as either being __application__ channels, or __ledger__ channels. Namely, they either run an application, or they [don't](https://docs.statechannels.org/docs/implementation-notes/consensus-app#implicit-or-null-app).

Separately, there are three main ways of funding a channel. 
1. Directly, with an on-chain deposit
2. Indirectly (sometimes called 'ledger funding'), via a ledger channel allocating funds to this one. That ledger channel might well be directly funded. 
3. Virtually, also via a ledger channel allocating funds to this one, but in a more complicated way*

* Namely, by a certain construction of several channels that fund each other and are participated in by a hub (in addition to one or the other or all of the participants of this channel). This is not yet documented nicely, and would be a nice thing to add (following the nitro paper but looking something like [this](https://docs.statechannels.org/docs/protocol-tutorial/off-chain-funding#indirect-funding)). 

I would also probably keep budgets out of this section, for clarity.

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| ---- | ---------------------- | ------------------------------------- |
+| 1000 | SigningAddressNotFound | Unable to find expected ephemeral key |
+| 1001 | InvalidAppDefinition   | App definition not valid address      |
+| 1002 | UnsupportedToken       | Token in allocations not supported    |
+
+TODO: (HIGH) Verify these are all of the expected errors (what if request times out? what if some participant refuses to join the channel?)
+
+## ApproveBudgetAndFund
+
+TODO: (HIGH) Better description of the actual usage of this rpc call. Is this essentially a collateral request? How much of an option do people have to reject these requests? Is this rejection exposed through the wallet, or would it have to be "undone" by some other series of wallet calls? Is this how you install apps in a ledger channel?
```
geoknee, line 90: Budgets currently serve a dual purpose (I would say).

1. They are essentially a permission, set by the user, allowing the wallet to silently open, _virtually fund_ , and close channels for a certain domain (without prompting the user), up to the given amounts (one for sending, one for receiving). The user has the chance to reject the budget via a wallet popup. If the budget runs out, they would need to re-approve another one. 
2. The budget is typically serviced by a ledger channel with the hub, so this _will_ require collateral from the hub. The ledger channel allocations typically match the budget amounts (my maximum send is my balance in the ledger channel, my maximum receive is the hub's balance in the ledger channel). 

See also [here](https://web3torrent.statechannels.org/budgets) and [here](https://www.notion.so/web3Torrent-FAQ-9f384d9dbadc4828aa81e14fcc360466).

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| hubAddress | string          | Signing address of channel intermediary                                |
+| budgets    | `TokenBudget[]` | An array of approved send and received capacity for a list of channels |
+
+### Errors
+
+TODO: (HIGH) Define errors in typescript
+
+## UpdateChannel
+
+Used to take a turn in your channel, or propose an update to a channel or application state. The application update must be a valid transition, and properly signed. The updated `ChannelResult` is returned.
```
geoknee, line 118: The wallet will do the signing, so there is no 'check' _here_ for that. Of course the contract will check the transition if it gets to chain.

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| appData                 | string            | Encoded initial state of app               |
+| appDefinition           | string            | Address of contract governing the channel  |
+| channelId               | string            | Unique channel identifier                  |
+| status                  | `ChannelStatus`   | Current status of channel                  |
+| turnNum                 | number            | Channel nonce                              |
+| challengeExpirationTime | number (optional) | Time current challenge on channel elapses  |
+
+### Errors
+
+TODO: (HIGH) Clean up error message types and codes
```
geoknee, line 143: Yes!

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,420 @@
...
+| ---- | ------------------ | ------------------------------------------- |
+| 400  | Channel not found  | Could not find channel to update in storage |
+| 401  | Invalid transition | Illegal state transition proposed           |
+| 402  | Invalid app data   | Incorrect encoded app information           |
+| 403  | Not your turn      | Cannot update channel                       |
+| 404  | Channel closed     | Channel no longer accepting updates         |
+
+## JoinChannel
+
+Called when you would are joining a channel that has been created. Generally, creating a channel is done, and a notification is sent to the desired counterparty, who then dispatches the `JoinChannel` request.
```
geoknee, line 155: When `createChannel` is called, that wallet will queue messages for the other participants. Once these are delivered to the other wallets, they will emit a `ChannelProposed` notification. The application itself should decide what to do (most likely, call `joinChannel`). 

Hopefully it is clear [here](https://docs.statechannels.org/docs/app-devs/make-api-calls).

geoknee (commented):
Thanks for this @LayneHaber , and also @snario for chipping in.

I think this is great as an exercise to understand this package and the API in general. As it happens we are in the process of migrating to an [autogenerated documentation ](https://docs.statechannels.org/docs/typescript-api/client-api-schema)for this stuff, and have just wound-down a [similar looking document ](https://app-wallet-interface.netlify.app/#introduction) that we had been working with.

If possible, I think we need to stick with the autogenerated route (it should be easier to maintain, version and so on). But we should also recognise that something like this is just plain easier to read and understand. I think we can have the best of both worlds though, it's just a case of either:

i) managing the documentation generation pipeline to organise things nicely.
ii) possibly altering the way the types are actually organised themselves: currently one has to dig around for e.g. `CreateChannelParams` in a massive list of type aliases, which is horrible. We want things organised by method as you have done here, and that might be possible by arranging the types in a smarter way (example of the awkwardness [here](https://docs.statechannels.org/docs/typescript-api/client-api-schema.closechannelrequest)). 

or both. Keen to hear your thoughts.

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
LayneHaber: It seems like it's an `Address` type in the `data-types` [file](https://github.com/connext/statechannels/blob/6dbfe6c96a9973fb8d1e48cac4a6e640d7bd7e99/packages/client-api-schema/src/data-types.ts#L37-L50):

```typescript
/**
 * Container for data specific to a single state channel participant
 */
export interface Participant {
  /**
   * App allocated id, used for relaying messages to the participant
   */
  participantId: string;
  /**
   * Address used to sign channel updates
   */
  signingAddress: Address;
  /**
   * Address of EOA to receive channel proceeds (the account that'll get the funds).
   */
  destination: Address;
}
```

What would the `Destination` type look like?

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
LayneHaber: I'm still confused as to what the `funding` status is, if there is a better pattern to handle the channel statuses, we should implement it in the finalized schema rather than having irrelevant ones included.

Also the comments mention `these statuses` does that mean opening is also irrelevant or they are just duplicates?

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
LayneHaber: We can try to implement them as completely as possible in the new spec, this would probably be a good place for me to start today

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
LayneHaber: Okay, that makes more sense. Can we remove it or add a comment in the types file explaining it? I would, but I'm still a bit unclear around the full context. Will remove from this file though

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
LayneHaber: What type of query language? Like just example code usage?

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
LayneHaber: Removed

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
LayneHaber: I've removed the documentation here, since it is a browser-specific method

LayneHaber: But thanks for the detailed explainer!

LayneHaber:
@geoknee going to respond with some thoughts

> I think we need to stick with the autogenerated route (it should be easier to maintain, version and so on)

I agree 100% that autogenerated documentation is easy to maintain. If you want to go down that route, we can keep this file alive (we can move it if need be) as a spec for the finalized core RPC interface (including errors, params, etc).

> currently one has to dig around for e.g. CreateChannelParams in a massive list of type aliases

Yeah, there is an interesting tradeoff between using fancy typescript and defining everything explicitly so your documentation can be easily autogenerated. In general we've tended towards maximizing type reuse and defaulting to manual documentation, but we have had a history of out of date documentation.

Happy to reorganize some of the type definitions to make it easier to catch by the documentation tool, but not keen to setup + debug the documentation generation pipeline. Still think the doc should be kept alive as a spec, and the PR opened to collect comments (though can move to issues if thats easier)

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
snario: It is a `bytes32` which may or may not be a converted left padded `address` (i.e., it is either a `ChannelId` or an `Address`)

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
snario: I'd comment on the linked issue — this is still open for discussion

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
snario: Not sure, but it seems inefficient to have a general purpose API to "get everything" instead of something like... "get everything that is running" or "get everything with this appDefinition". A `Store` might implement this differently depending on the size and situation of the database.

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
geoknee: Little bit of extra context: We can (and do) add annotations to the declarations in that file, such that we can actually enforce stricter checks at run time. Example (left padded address): 

https://github.com/statechannels/statechannels/blob/c831b2f14ac086567989db3691a6a1f0adcb3298/packages/client-api-schema/src/data-types.ts#L14-L18

[Here ](https://docs.statechannels.org/docs/protocol-tutorial/deposit-assets#destinations)is the corresponding docs entry.

As you can see from the code comment, this work is a little incomplete.

I think in this case `destination` ought to remain an `Address`, since here we are sure we want an ethereum address. In other places, `Destination` is right.

geoknee:
@snario I am not totally clear why we want to strip out all of the browser specific stuff. We still want to support that, right? The schema is currently used by the browser wallet and accompanying packages, and not yet used by the server wallet. So it makes sense to me to continue to document all of the browser-specific stuff. 

I would be in favour of having a core set of types, and a core schema to go with, that is extended by some browser specific stuff and a browser specific schema (which is essentially a bit more permissive when you validate requests against it). But that can all live in this package.

snario:
> @snario I am not totally clear why we want to strip out all of the browser specific stuff. We still want to support that, right? The schema is currently used by the browser wallet and accompanying packages, and not yet used by the server wallet. So it makes sense to me to continue to document all of the browser-specific stuff. 
> 
> 
> 
> I would be in favour of having a core set of types, and a core schema to go with, that is extended by some browser specific stuff and a browser specific schema (which is essentially a bit more permissive when you validate requests against it). But that can all live in this package. 

I agree with your suggestion. I just want to make sure we are all on the same page that these browser specific methods should not be assumed to be fundamental wallet requirements. 

In addition to being browser specific, ApproveBudgetAndFund is even virtual channel specific in its current form. These are shortcuts / shorthands.

geoknee:
I am on your page 💯 .

LayneHaber:
I'm going to keep this doc linked to only the core types then.

To regurgitate what was in slack, I propose the following changes to the error types:

We should keep the general HTTP *xx error code mapping, but prefix them with some other number to indicate the error was from the channel (e.g. 404 becomes 1404), then we can follow the general error mappings:

```bash
1400 -- invalid data supplied
1401 -- unauthorized action
1403 -- forbidden action
1404 -- * not found in storage
```

Which means (by method) we would see the following errors (roughly):
```
CreateChannel
Signing address not found (1404)
Invalid participants (1400)
Insufficient funds (1401)
Unsupported/invalid token (1400)
Invalid app definition (1400)
Invalid app data (1400)

JoinChannel
Channel not found (1404)
Invalid transition (1403)
Channel has invalid status (1403)

UpdateChannel
Channel not found (1404)
Invalid transition (1403)
Invalid app data (1400)
Not your turn (1403)
Channel has invalid status (1403)

CloseChannel
Channel not found (1404)
Not your turn (1403)

DefundChannel
Channel not found (1404)
Not your turn (1403)
Channel has invalid status (1403)

Withdraw
Channel not found (1404)
Not your turn (1403)
Channel has invalid status (1403)

ChallengeChannel
Channel not found (1404)

GetChannels

GetState
Channel not found (1404)

GetWalletInformation
Signing wallet not found (1404)

PushMessage
Participant not found (1404)
Unauthorized (1403)
```

I'm still not clear if this is 100% of the methods we would use though. Wondering specifically about `deposit` and `clearChallenge`, but there could be other core methods I'm missing

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
LayneHaber: That is true, but there are a lot of different `getters` we could settle on. We can look at some of the ones that we have defined in our node, but it's hard to envision a complete set of performant getters needed by every application.

We should definitely think more about the ones that are needed though (i.e. get all by definition, or get all by participant)

LayneHaber: What about what I just added to the doc in the parameters?

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,343 @@
...
+};
+```
+
+so that if nothing is supplied, it will return all channels and if a participant is supplied, it will return all channels with that participant, etc.
+
+### Response
+
+Returns a `ChannelResult[]`
+
+## GetState
```
LayneHaber, line 278: rename get channel

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,343 @@
...
+
+Accepts an empty object as the JSON RPC Request parameters.
+
+### Response
+
+| Name               | Type                | Description                           |
+| ------------------ | ------------------- | ------------------------------------- |
+| signingAddress     | string              | Eth address used to sign from wallet  |
+| destinationAddress | string \| undefined | Where funds should go after a dispute |
+| walletVersion      | string              | Wallet version number                 |
```
LayneHaber, line 312: separate into `GetVersion`

packages/docs-website/docs/protocol-docs/client-specification/json-rpc-api.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Parameters
+
+| Name      | Type    | Description            |
+| --------- | ------- | ---------------------- |
+| recipient | string  | Recipient identifier   |
+| sender    | string  | Sender identifier      |
+| data      | unknown | Arbitrary message data |
+
+### Response
```
LayneHaber, line 332: different behavior in browser and server (make issue)

andrewgordstewart:
Closing as this is stale, and the tests are failing.

PR #2491, opened 2020-08-25
server-wallet: Remove requirement on NODE_ENV

snario:


snario:
It defaults to `development` in `devtools` which the migrations use

PR #2492, opened 2020-08-25
Sync channel

andrewgordstewart:


snario:
Updated based on https://www.notion.so/elfour/Channel-syncing-224c505e440c4056b5e4f072f0bab631

packages/server-wallet/src/wallet/index.ts
```diff
@@ -80,6 +85,35 @@ export type WalletInterface = {
 };
 
 export class Wallet implements WalletInterface {
+  public async syncChannel({channelId}: SyncChannelParams): SingleChannelResult {
+    return Store.lockApp(
```
tomclose, line 89: Do we need to lock the app here? It's only reading data, so it looks like it might be fine not to?

tomclose (approved):

packages/server-wallet/src/wallet/index.ts
snario: We do not — this is a relic of the previous API. Let me remove...

PR #2495, opened 2020-08-25
Upgrade ethers to 4.0.48

snario:
This upgrades `ethers` to 4.0.48, which solves a security vulnerability with `elliptic`.

It also bumps `nitro-protocol` and `server-wallet` versions, so we can distribute those out to consumers.


andrewgordstewart:
I confirmed that `server-wallet` resolves `elliptic` to version `6.5.3` by running `yarn clean && yarn` on this branch, followed by
```
monorepo/packages/server-wallet on  liam/ethers-4.0.48 [$] is 📦 v0.3.3 via ⬢ v12.16.3 on ☁️  us-west-1
❯ node -e "console.log(require('elliptic').version)"
6.5.3
```

PR #2501, opened 2020-08-26
Upgrade and patch docusaurus

geoknee:
This is an attempt to prevent the current build failure on master. 

My hypothesis is that `docusaurus-build` is really slow, and also doesn't output any information as it works. The solution I propose here is to output some status info, so that circle doesn't time out the job. 

`docusaurus-build` generates 1623 html files in a big loop, each go round does some blocking I/O.


geoknee:
It took 13 minutes to build the docs.

andrewgordstewart:
I'm a bit wary about this:
- `release-pull-request` is now a required check. Why is it required? Should the docs instead be generated and deployed when things are merged into master? Or perhaps when a new version is published?
  - Having the deployed docs change from under my feet for a given version seems a bit bad.
- [this step](https://app.circleci.com/pipelines/github/statechannels/statechannels/8746/workflows/27618253-d0ee-46b3-9490-9460076f9685/jobs/41264) took 25 minutes. Why is it so slow?

Even without this, the CI seems to grow and grow until someone puts a bunch of effort into making it fast. This seems to blow everything else out of the water :P

geoknee:
> I'm a bit wary about this:
> 
> * `release-pull-request` is now a required check. Why is it required? Should the docs instead be generated and deployed when things are merged into master? Or perhaps when a new version is published?

Currently what happens is, if you make a change that impacts the docs then a deploy preview for the documentation site is generated. That's what the `release-pull-request` job does. 

It is _required_ because it is the thing that guarantees `release-master` (which publishes the actual docs) will not fail on the master branch, which is a situation we are in right now. Every pull request save this one has a 🚫 . It's pretty bad if the docs do not get published. 


>   * Having the deployed docs change from under my feet for a given version seems a bit bad.

I don't understand what you mean. There is a `next` version of the docs that stays in sync with the master branch. All other versions that have a version number are stable and should not change (apart from in cases where we want to, say fix a typo in a previous version). 

[It is possible to _not_ use the `next` version .](https://docusaurus.io/docs/en/commands#docusaurus-build). That might save us a little bit of time, but probably not much. 

> * [this step](https://app.circleci.com/pipelines/github/statechannels/statechannels/8746/workflows/27618253-d0ee-46b3-9490-9460076f9685/jobs/41264) took 25 minutes. Why is it so slow?
> 

If you read my comment above, you'll see why I think `docusaurus-build` is particularly slow. It's largely to do with having many files that need to be generated. There's also a lot of slowness from pushing those files to netlify. 

> Even without this, the CI seems to grow and grow until someone puts a bunch of effort into making it fast. This seems to blow everything else out of the water :P

Yes you're right this is now by far the limiting factor _when these steps are triggered_, which is not always.

 I will spend some time to see what can be done to bring the time down. Meanwhile, this PR (I believe) is making our process strictly better so should be merged (even if it is revealing some other problems that weren't as obvious to us before).

geoknee:
See #2519 : upshot being that the job only takes 3 minutes to run instead of 25 now. And we don't version the docs anymore so large number of .md files have been removed.

andrewgordstewart:
:racehorse:

PR #2503, opened 2020-08-26
[client-api-schema] Experiment with organisation of types

geoknee:
⚠️ this branch currently totally breaks everything. It's just an example of some tweaks we could make to improve the docs that get autogenerated. 

I created these screenshots by building the docs locally -- usually we would get a deploy preview on this PR.


packages/client-api-schema/src/index.ts
```diff
@@ -27,6 +27,9 @@
  *   "result": {"success": true}
  * }
  * ```
+ *
+ * # Methods
+ * {@link CreateChannel}
  */
```
geoknee, line 33: We maintain a list of links on the front page of the API docs
Screenshot 2020-08-26 at 18 42 32

Just one method shown for now, but we would add all of them (possibly pinching off a browser-only list)

packages/client-api-schema/src/methods/CreateChannel.ts
```diff
@@ -1,8 +1,23 @@
...
+
+/**
+ * Request a new channel
+ */
+export interface CreateChannel {
+  request: JsonRpcRequest<'CreateChannel', CreateChannelParams>;
+  response: JsonRpcResponse<ChannelResult>;
+  errorResponse: JsonRpcErrorResponse<CreateChannelError>;
+}
+
```
geoknee, line 20: We export interfaces like this to bundle things that belong together in a single 'method'
Screenshot 2020-08-26 at 18 42 40

geoknee: and we also ensure all relevant types are exported, which makes the generic type parameters hyperlinks that you can click on to drill down to the basic types:
Screenshot 2020-08-26 at 18 42 51

The `api-generator` tool warns us if we forgot to export a type.

packages/client-api-schema/src/index.ts
LayneHaber: These previews look super clean!

packages/client-api-schema/src/methods/CreateChannel.ts
LayneHaber: So I think what I mentioned earlier about fancy types and autogenerated documentation tradeoff is rearing its head here.

It may be a good convention to just bite the bullet and do at *most* two layers of hyperlinks. If we start to do things that have super good code reuse these type docs will get really ugly and hard to use, i.e. the example type in #2507 :

```typescript
// Say you want to define an object that is just a `ChannelResult` but all fields are optional
// and the `status` could be a `ChannelStatus` array

// You can define the object very verbosely:
type GetChannelsRequestParameters = {
  participants?: Participant[]; // string[] for participantIds?
  channelId?: Bytes32;
  status?: ChannelStatus[];
  appDefinition?: Address;
  ...
};

// or with some fancy typescript
type GetChannelsRequestParameters = Partial<Omit<ChannelResult, "status"> & { status?: ChannelStatus[] }>
```

In this package we would probably want to default to the more verbose definition. The upside is the docs will be way better, and your typedefs themselves will be more readable (it will even help with vscode extensions). The downside is that you have to make sure that the `ChannelResult` and the `GetChannelsRequestParameters` are constantly in sync. 

The tradeoff here is largely dependent on how often you think you'll be changing these types. If it's fairly rarely (which it should be, rpc interface changes are usually breaking) then I think it's worth it to use more verbose typedefs

geoknee: One thing you get from the autogenerated docs is a similar level of complexity / indirection in the docs as are in the source code. 

Totally agree about the tradeoffs between verbosity and reuse of types. Having many deep levels of hyperlinks to click through is not great, and it mirrors the hops a developer has to go through in their mind to understand the types themselves. 

In terms of how frequently we change the RPC interface, it seems likely that in the short term there will be quite a lot of changes. In that case i would prefer keeping the docs tightly coupled to the code itself, even if that makes them less readable. When things settle down a bit, we could revisit this and see if increasing verbosity is worth it then.

Just one more piece of context I wanted to make you aware of. Our documentation is generated holistically across several packages. So if we look at the [`createChannel` method on the `ChannelClient` class ](https://docs.statechannels.org/docs/typescript-api/channel-client.channelclient.createchannel)(which is in a different package in this monorepo), the return type is `ChannelResult` which is a hyperlink to the docs for `client-api-schema`. 

This brings me on to the question of the target audience of the documentation. Up to now we have been focussing a lot on third party application developers, who will likely _not_ be concerned with the `@statechannels/client-api-schema` package, because they can use the `@statechannels/channel-client` package and _that_ will serve as the API for them.  That API  and its docs can be much cleaner, because the "methods" are _actually_  javascript methods.

LayneHaber: > In terms of how frequently we change the RPC interface, it seems likely that in the short term there will be quite a lot of changes. In that case i would prefer keeping the docs tightly coupled to the code itself, even if that makes them less readable. When things settle down a bit, we could revisit this and see if increasing verbosity is worth it then.

Cool, that sounds good!

> target audience of the documentation

I think that you should definitely keep the majority of the documentation targeted towards application developers. However, you should probably include this in an `Advanced` section for developers who want to implement a channel wallet in a new environment. There may be very few of them, but there's nothing in your docs to help them out at the moment.

geoknee:
Just a quick update, I have carried on with this idea locally, expanding this approach to the other methods. I have also started folding in some of the ideas in #2504 . 

Stumbling blocks that I am hitting: 
- For annoying reasons it is not easy to get a list or table of links on the front page of the docs. They all render in line at the moment 🙄 
Screenshot 2020-08-27 at 18 14 30

- I think union types are going to be extremely useful for us, but they do not get documented nicely:
Screenshot 2020-08-27 at 18 17 07
Interfaces at least have their properties enumerated in a table with hyperlinks. For type aliases, you can't drill down and have to go back to the main menu 👎 . 

I might find solutions to both problems by tinkering with the `api-documenter` tool.

PR #2519, opened 2020-08-28
Upgrade to docusaurus 2

geoknee:
Motivation:
- `docusaurus-build` on 1.x is very slow. I spent some time trying to monkeypatch it to be faster, but ultimately failed at this.
- `docusaurus build` on 2.x seems much faster. It looks better, and there are [some nice features ](https://v2.docusaurus.io/docs/migrating-from-v1-to-v2)

Take a look yourself: 
https://5f493ec0004db05da087dfce--nitro-protocol.netlify.app/

To note: 
- Docusaurus 2 stores the entire snapshot of the docs when you cut a new version. when of versions are saved. So the problem of having lots of .md files is exacerbated. In the end I just nuked all versioned_docs on this branch as this was annoying me. Sadly I don't think [git LFS](https://git-lfs.github.com/) would be much help if we wanted to keep using versioning.
- I spent some time fixing mermaid diagrams, which got broken as a part of the migration.
- The build job needs more memory on circle
- the combination of upgrade and nuking versioned docs (but mostly nuking versioned docs) means it builds a lot faster. About 1 minute to build, and 2 minutes to deploy to netlify. 

TODO
- [x] Take a careful look through the site to see if anything isn't as good as it was
- [x] Decide on what to do about versioning
- [x] fix all the broken links that are highlighted in the output of `docusaurus build`
- [x] fix solidity syntax highlighting


geoknee:
https://v2.docusaurus.io/docs/versioning

It seems that docusaurus themselves would not recommend using versioning for our site, and certainly not to cut versions as often as we do. So I think versioning is out.

andrewgordstewart:
:racehorse: :scissors: this looks like an awesome change set. It's unfortunate about the versioned docs, but perhaps we can revisit that later with a different approach.

Issue #2502, opened 2020-08-26
`server-wallet` config through constructor instead of environment variables


rhlsthrm:
**Problem description :confused:**
We are mocking out the `server-wallet`, and we need a way to inject the config through the constructor as opposed to through environment variables. This allows us to more easily test using dependency injection.

**Describe the solution you'd like :rocket:**
Accept constructor arguments to configure `server-wallet`. It's fine to use the env vars as a fallback.

**Describe alternatives you've considered**
Alternative is to inject the env in the test framework itself, but that's messier.

**Additional context**

snario:
> We are mocking out the `server-wallet`

Do you have a link to this? I'm in favour of taking the [`ServerWalletConfig`](https://github.com/statechannels/monorepo/blob/5cc9ad70b90e833941f947082d9a6e65ec2b6e3a/packages/server-wallet/src/config.ts#L5-L21) and making it a part of the constructor. However maybe first we want to think more carefully about what should be in here?

Also does your use case actually _require_ this or is it just for cleaner code? AFAICT there isn't a use case supported where you'd want more than one `Wallet` per process, so using `process.env` is tenable (i.e., this is a non blocking issue).

rhlsthrm:
> Do you have a link to this?
https://github.com/connext/temp-indra-v8/blob/master/modules/isomorphic-node/src/test/mocks/wallet.ts

Temporary location for this work. As you mention it can be worked around for now.

bohendo:
> I'm in favour of taking the ServerWalletConfig and making it a part of the constructor. However maybe first we want to think more carefully about what should be in here?

I see no reason why we need to finalize constructor params first. I do want us to think more deeply about what the simplest-yet-complete set of constructor params are but in the meantime it'd be really quick & easy on your end to throw the existing ServerWalletConfig into the constructor & immediately make our side's testing env run a lot more smoothly.. Hmm if it'd be so quick & easy maybe I should just submit a PR :thinking: lol

PR #2527, opened 2020-08-31
Onchain Service V0

LayneHaber:
Contains:
- channel registration with a given set of asset holder addresses + asset ids
- executes `updateChannelFunding` on attached channel wallet if `Deposited` event is detected
- submits + retries transactions


packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+            destinationHoldings,
+          },
+        };
+        this.store.set(destination, updated);
+        // Call the appropriate callback on the wallet
+        this.channelWallet.updateChannelFunding({
+          channelId: destination,
+          amount: amountDeposited.toString(),
+          token: this.assetHolders.get(assetHolder.address)?.assetId,
+        });
```
snario, line 224: You may want to de-couple the chain event monitoring from the post-event hook. Perhaps this `OnchainService` has some [observables](https://rxjs-dev.firebaseapp.com/guide/observable) it maintains, and you can bind this hook function to each observable. e.g.,
```
[ ethers event monitors ] x 50
                    | sends event to
  [ global deposited observable]
                    |
[ hook attached to 'Deposit' ]
                    |
[ hook calls `updateChannelFunding` ]
```

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+
+import {Wallet as ChannelWallet} from '../wallet';
+import {Address, Bytes32} from '../type-aliases';
+
+const BAD_NONCE = "the tx doesn't have the correct nonce";
+const INVALID_NONCE = 'Invalid nonce';
+const NO_TX_HASH = 'no transaction hash found in tx response';
+const UNDERPRICED_REPLACEMENT = 'replacement transaction underpriced';
+const DEFAULT_MAX_RETRIES = 5;
+const KNOWN_ERRORS = [BAD_NONCE, NO_TX_HASH, UNDERPRICED_REPLACEMENT, INVALID_NONCE];
```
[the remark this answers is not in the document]
andrewgordstewart: https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/wallet/store.ts#L227-L243

That is our current convention

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+// arguments are not known at the time of creating a transaction.
+type MinimalTransaction = {
+  to: Address;
+  value: string;
+  data: string;
+};
+
+type AssetHolderInformation = {
+  assetId: Address;
+  contractAddress: Address;
```
andrewgordstewart, line 32:
```suggestion
  tokenAddress: Address;
  assetHolderAddress: Address;
```

ERC20 asset holders already stores the token information, so this type probably isn't even necessary: you should be able to just provide the asset holder address.

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+  private wallet: Wallet;
+  private retries: number;
+
+  // NOTE: V0 has an in-memory store, which will eventually be
+  // shifted to more permanent storage
+  // Holds latest channel event by type (i.e. { "fundsDeposited": {...} })
+  private store: Map<string, object> = new Map();
+
+  // Stores references to all contracts in memory
+  private assetHolders: Map<string, AssetHolderInformation & {contract: Contract}> = new Map();
```
andrewgordstewart, line 61: Per my comment above, if the key is the asset holder address, then you only need to store the contract -- the other asset holder information is already on the contract.

```suggestion
  private assetHolders: Map<string, Contract> = new Map();
```

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+// requires the nonce and chain ID to be specified, when sometimes those
+// arguments are not known at the time of creating a transaction.
+type MinimalTransaction = {
+  to: Address;
+  value: string;
+  data: string;
+};
+
+type AssetHolderInformation = {
+  assetId: Address;
```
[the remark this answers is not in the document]
LayneHaber: Address of the token for the asset holder

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+  private channelWallet: ChannelWallet | undefined = undefined;
+
+  constructor(env: OnchainServiceEnvironment) {
+    this.retries = env.transactionRetries || DEFAULT_MAX_RETRIES;
+    this.provider =
+      typeof env.provider === 'string' ? new providers.JsonRpcProvider(env.provider) : env.provider;
+    this.wallet =
+      typeof env.wallet === 'string'
+        ? new Wallet(env.wallet, this.provider)
+        : (env.wallet as Wallet).connect(this.provider);
```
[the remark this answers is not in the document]
LayneHaber: Calling `connect` on an `undefined` will fail (but obviously not with a nice error). I don't even really need to add the typecasting, I think I just wanted an easier clickthrough at the moment and forgot to remove it. We can do more validation, kind of your call but it wouldn't be horribly antipattern to just rely on typescript types

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+   */
+  public registerChannel(
+    channelId: Bytes32,
+    assetHolders: AssetHolderInformation[]
+  ): Promise<string | undefined> {
+    if (!this.channelWallet) {
+      throw new Error(
+        'Must attach channel wallet before registering channel, this will go away in v1'
+      );
+    }
```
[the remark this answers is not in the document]
LayneHaber: What does it mean to register a channel if you don't have a wallet you can take action on or call a function on once the event is fired? I'm not parroting events from this service in any way, so it would only update storage (which may be desirable, I'm not sure).

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    // Create and store new contracts if we don't have record of
+    // required assetHolders
+    assetHolders.forEach(assetHolder => {
+      if (this.assetHolders.has(assetHolder.contractAddress)) {
+        return;
+      }
+      // TODO: Use the AssetHolder abi?
+      const contract = new Contract(
+        assetHolder.contractAddress,
+        ContractArtifacts.EthAssetHolderArtifact.abi,
```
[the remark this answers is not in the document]
LayneHaber: I'll add the export from the `nitro-protocol` package

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  public async submitTransaction(
+    channelId: Bytes32,
+    tx: MinimalTransaction
+  ): Promise<providers.TransactionResponse> {
+    if (!this.channelWallet) {
+      throw new Error(
+        'Must attach channel wallet before registering channel, this will go away in v1'
+      );
+    }
```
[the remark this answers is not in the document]
LayneHaber: Have the same questions for what it means if you call `submitTransaction` without any guarantee that it will have an appropriate effect on the channel itself, but agree that these checks aren't strictly _needed_ to make the functions work (moreso here than in `registerChannel`)

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    // Use higher of chain or memory nonce
+    const nonce = this.memoryNonce > chainNonce ? this.memoryNonce : chainNonce;
+
+    // TODO: should we set gas minimums or preferences?
+
+    // Send the transaction
+    const response = await this.wallet.sendTransaction(tx);
+    if (!response.hash) {
+      throw new Error(NO_TX_HASH);
+    }
```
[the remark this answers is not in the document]
LayneHaber: So, I'm not sure. This is an error check I pulled from our existing production onchain transaction service. Unfortunately there's no comment also explaining under what conditions it happened, or with which version of ethers, so it's probably not going to happen.

That being said, the fact that it is present in our code in production indicates that it is something to watch for and it seems relatively inexpensive to assert. Open to removing it though

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    this.memoryNonce = nonce + 1;
+    return response;
+  }
+
+  // NOTE: could also do this in the `wait` callback in `submitTransaction`
+  // but would need more information about what event to parse from the
+  // receipt via the API
+  private _registerAssetHolderCallbacks(assetHolder: Contract): void {
+    assetHolder.on(
+      'Deposited',
```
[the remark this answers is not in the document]
LayneHaber: > We also have helper functions inside of @statechannels/nitro-protocol to filter these events I think.

I'll double check but it seemed like those helpers were for parsing events from transaction logs. 

> filtering API to restrict events to specific subsets of emitted data based on the contained fields

I'll try to add the restricted event filters, but my concern is around the filter dynamically updating when I register a new channel. I don't want to turn the listeners off and on to update the filter (seems like it could get weirdly racy) every time I add a new channel, and I don't want a registered channel to have a 1:1 relationship with listeners on the contract. This seems like something that I'm going to need tests to clear up though, but if we can do it with the ethers filter easily that would definitely be better

packages/server-wallet/src/onchain-service/index.ts
LayneHaber: The generic `AssetHolder` contract doesn't have a reference to this property, how would I access it? Definitely would love to drop the type.

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+  public attachChannelWallet(wallet: ChannelWallet): void {
+    this.channelWallet = wallet;
+  }
+
+  /**
+   * Adds a channel to watch events and submit transactions for
+   *
+   * @param channelId Unique channel identifier to register chain listeners for
+   * @param assetHolders Onchain addresses of the channel chain context
+   * @returns String if there is an error, undefined if successful
```
kerzhner, line 92: Does this function ever return a string? Also, this function is not `async` but returns a Promise. Should the function return a Promise?

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+        }
+        const updated = {
+          ...record,
+          ['Deposited']: {
+            destination,
+            amountDeposited,
+            destinationHoldings,
+          },
+        };
+        this.store.set(destination, updated);
```
kerzhner, line 218: Do you think there might be issues with events arriving out of order? What if the second `Deposited` event arrives before the first?

kerzhner: One option is to compare `destinationHoldings` from the event to `destinationHoldings` stored.

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    for (let attempt = 1; attempt < this.retries; attempt += 1) {
+      try {
+        const response = await this.queue.add(() => this._sendTransaction(tx));
+        return response;
+      } catch (e) {
+        // Store error
+        errors[attempt] = e.message;
+
+        // Only retry if it is a known error
+        const knownErr = KNOWN_ERRORS.find(err => e.message.includes(err));
```
kerzhner, line 157:
```suggestion
        const knownErr = KNOWN_ERRORS.find(e.message.includes);
```

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    // If the channel is not registered, do not send transactions
+    if (!this.store.has(channelId)) {
+      throw new Error(`Channel is not registered ${channelId}`);
+    }
+
+    // Create the errors object
+    const errors: {[k: number]: string} = {};
+    for (let attempt = 1; attempt < this.retries; attempt += 1) {
```
kerzhner, line 148: Might be a touch cleaner
```suggestion
    for (const attempt in _.range(retries - 1)) {
```

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+        return response;
+      } catch (e) {
+        // Store error
+        errors[attempt] = e.message;
+
+        // Only retry if it is a known error
+        const knownErr = KNOWN_ERRORS.find(err => e.message.includes(err));
+        if (!knownErr) {
+          throw new Error(
+            `Transaction failed to send with unknown error on attempt ${attempt}/5: ${JSON.stringify(
```
kerzhner, line 160:
```suggestion
            `Transaction failed to send with unknown error on attempt ${attempt}/${this.retries}: ${JSON.stringify(
```

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+              errors
+            )}`
+          );
+        }
+      }
+    }
+    throw new Error(`Unable to send transaction after 5 retries: ${JSON.stringify(errors)}`);
```
kerzhner, line 167:
```suggestion
    throw new Error(`Unable to send transaction after ${this.retries} retries: ${JSON.stringify(errors)}`);
```

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
+import {ContractArtifacts} from '@statechannels/nitro-protocol';
+import {providers, Wallet, Contract, BigNumber} from 'ethers';
+import PriorityQueue from 'p-queue';
+
+import {Wallet as ChannelWallet} from '../wallet';
+import {Address, Bytes32} from '../type-aliases';
+
+const BAD_NONCE = "the tx doesn't have the correct nonce";
```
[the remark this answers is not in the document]
LayneHaber: Yes it is, not really sure how it got through the pre-commit hooks. I'll change my settings back to format on save 👍

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+const INVALID_NONCE = 'Invalid nonce';
```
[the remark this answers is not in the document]
LayneHaber: There isn't, but sometimes the error will appear both ways. These error strings were pulled from the ones we've manually added in our production environment

packages/server-wallet/src/onchain-service/index.ts
LayneHaber: There's a note on the string vs. void in the interface, it was a bit of a hack but I've removed it.

Also, I know right now it's synchronous, but I have a feeling that as soon as we start adding storage calls it will need the asynchronous interface, so I return a `Promise.resolve()` in most places

packages/server-wallet/src/onchain-service/index.ts
```diff
@@ -0,0 +1,228 @@
...
+  private store: Map<string, object> = new Map();
+
+  // Stores references to all contracts in memory
+  private assetHolders: Map<string, AssetHolderInformation & {contract: Contract}> = new Map();
+
+  // Transaction queue
+  private queue = new PriorityQueue({concurrency: 1});
+
+  // Memory tracking wallet nonce to assist with retries
+  private memoryNonce = 0;
```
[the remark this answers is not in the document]
LayneHaber: It's the experimental packages, is it okay to use? Have you guys used it before? cc @andrewgordstewart I think you were in the issue comments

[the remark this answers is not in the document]
andrewgordstewart: The worst that can happen is it causes a transaction to fail due to a duplicate nonce, so I think it's fine!

packages/server-wallet/src/onchain-service/transaction-submission.ts
```diff
@@ -0,0 +1,120 @@
...
+} from './types';
+
+// FIXME: replace with
+// import {WalletError as ChannelWalletError} from '@statechannels/server-wallet';
+import {WalletError as ChannelWalletError} from '..';
+
+type Values<E> = E[keyof E];
+
+export class TransactionSubmissionError extends ChannelWalletError {
+  readonly type = ChannelWalletError.errors.OnchainError;
```
snario, line 18: TransactionSubmissionError?

packages/server-wallet/src/onchain-service/transaction-submission.ts
```diff
@@ -0,0 +1,120 @@
...
+        const response = await this.queue.add(() => this._sendTransaction(tx));
+        return response;
+      } catch (e) {
+        // Store the error in memory
+        indexedErrors[attempt.toString()] = e.message;
+
+        // Retry IFF known error
+        const knownErr = Object.values(TransactionSubmissionError.knownErrors).find(
+          e.message.includes
+        );
```
snario, line 89: This seems like a helpful generic function on the base error class

packages/server-wallet/tsconfig.json
```diff
@@ -16,5 +16,5 @@
     {"path": "../nitro-protocol"},
     {"path": "../wallet-core"}
   ],
-  "include": ["src", "e2e-test"]
+  "include": ["src", "e2e-test", "deployment"]
```
snario, line 19: I'm not sure if we want this in here — we probably need a second config file...`tsconfig.integration.json` or something. Not that important really, but flagging it for now.

packages/server-wallet/src/onchain-service/transaction-submission.ts
```diff
@@ -0,0 +1,120 @@
...
+  private provider: providers.JsonRpcProvider;
+
+  // Wallet to send transactions
+  private wallet: Wallet;
+
+  // Queues to handle transactions
+  private queue = new PriorityQueue({concurrency: 1});
+
+  // Memory tracking wallet nonce to assist with retries
+  // TODO: maybe replace with NonceManager?
```
snario, line 54: See [`NonceManager`](https://docs.ethers.io/v5/api/experimental/#experimental-noncemanager). I think you would replace the `wallet` private property on this class with that

snario:
I think it would be helpful to add that diagram to this PR

packages/server-wallet/src/onchain-service/index.ts
LayneHaber: Looks like you will have to change the contract to make this an explicitly public property before I am able to access it this way.

andrewgordstewart: Ahhh. That's too bad -- duplicating the data is not the worst thing in the world, but it would sure be nice if we could just get the token address from the asset holder contract.

packages/server-wallet/src/onchain-service/transaction-submission.ts
LayneHaber: Have you used it before? A little spooked by the fact it's in `experimental` packages

packages/server-wallet/src/onchain-service/transaction-submission.ts
LayneHaber: Sure, was going to reuse the type for all onchain services. Since this extends the `WalletError`, which has a static list of types, and the wallet is *only* concerned with an onchain transaction service my thinking was both services should use the same error type

packages/server-wallet/src/onchain-service/transaction-submission.ts
kerzhner: We have used the `NonceManager` before for an older state channels hub, and, from what I remember, it worked well!

packages/server-wallet/src/onchain-service/onchain-service.ts
```diff
@@ -0,0 +1,173 @@
...
+  public registerChannel(
+    channelId: Bytes32,
+    assetHolders: AssetHolderInformation[]
+  ): Promise<void> {
+    if (!this.channelWallet) {
+      throw new OnchainServiceError(OnchainServiceError.reasons.noChannelWallet);
+    }
+
+    // If the channel has already been registered, return
+    if (this.store.has(channelId)) {
```
andrewgordstewart, line 90: What if the channel has already been registered, but for different asset holder addresses?

packages/server-wallet/src/onchain-service/onchain-service.ts
```diff
@@ -0,0 +1,173 @@
...
+   * Adds a channel to watch events and submit transactions for
+   *
+   * @param channelId Unique channel identifier to register chain listeners for
+   * @param assetHolders Onchain addresses of the channel chain context
+   * @returns an empty promise that will resolve once the channel is registered
+   */
+  public registerChannel(
+    channelId: Bytes32,
+    assetHolders: AssetHolderInformation[]
+  ): Promise<void> {
```
andrewgordstewart, line 84: I don't see any async methods in this function's body. Does it need to be async?

(It seems like it _should_ be async, because eg. the asset holders need to be connected to an ethereum provider at a certain address, doesn't it?)

packages/server-wallet/src/onchain-service/transaction-submission.ts
```diff
@@ -0,0 +1,120 @@
...
+        ? new Wallet(wallet, this.provider)
+        : wallet.connect(this.provider);
+  }
+
+  public async submitTransaction(
+    tx: MinimalTransaction,
+    options: TransactionSubmissionOptions = {}
+  ): Promise<providers.TransactionResponse> {
+    // Check via casting to avoid 0 returning falsy values
+    const attempts = typeof options.maxSendAttempts === 'number' ? options.maxSendAttempts : 1;
```
andrewgordstewart, line 70: Since we're using typescript, we could use the [null coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). (The compiled javascript will be somewhat similar to this code)

```suggestion
    const attempts = options.maxSendAttempts ?? 1;
```

packages/server-wallet/src/wallet/index.ts
```diff
@@ -240,6 +244,11 @@ export class Wallet implements WalletInterface {
   onNotification(_cb: (notice: StateChannelsNotification) => void): {unsubscribe: () => void} {
     throw 'Unimplemented';
   }
+
+  // Should be called after wallet creation
```
andrewgordstewart, line 248: At first, I didn't understand this comment. This is a method defined on an instance of a wallet, so it can only be called after the wallet is created.

Then I thought it means "this _must_ be called after wallet creation". If so, why isn't it just passed to the wallet's constructor?

packages/server-wallet/src/onchain-service/onchain-service.ts
LayneHaber: It's async now, typed it like this because I knew it *would* be async eventually

packages/server-wallet/src/onchain-service/onchain-service.ts
LayneHaber: I had not considered that possibility, and I rearranged some things to address this. As it stands:

- all registered channels are saved by channel id in the mapping with their events as the value (no longer any reference to asset holder info stored there)
- the asset holders are saved by asset holder address in the mapping, with an ethers Contract as the value

When a new channel is registered:
- IF it does not exist in the channels mapping, it is added (function does not return here anymore)
- IF the asset holder is not in the mapping, listeners for the contract events are registered

packages/server-wallet/src/onchain-service/index.ts
LayneHaber: I don't want to continually read the token address from the chain, so I'll keep the mapping type as is, but the input types have changed!

andrewgordstewart: ah fair, that sounds good

packages/server-wallet/src/wallet/index.ts
LayneHaber: We could pass it into the wallets constructor, there's just nothing else in there right now (though we should also make the environment config injectable see #2502 )

packages/server-wallet/src/onchain-service/index.ts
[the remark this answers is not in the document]
LayneHaber: Check out the EVT changes I made, there are filters I included at that level, each asset holder only has the event registered once.

We can try to add the ethers filters, but I think the `pipe` functions have a nicer API

LayneHaber:
> What happens if submitTransaction gets called twice? This might happen if a server-wallet crashes but the on-chain service does not crash. It should probably not send the same thing twice.

So I handle the case where we don't error if the same channel is registered twice* in the onchain service, but I'm not sure how to handle the double submission problem.. Open to any and all strategies here, I'm not sure the best way to handle that from the onchain services side.

*NOTE: We don't do anything to track the relation between asset holder and channel. I don't think we do, but this is an edgecase to pay attention to.

packages/server-wallet/jest/chain-setup.ts
```diff
@@ -0,0 +1,22 @@
...
+  const account = {privateKey: config.serverPrivateKey, amount: utils.parseEther('100').toString()};
+  const ganacheServer = new GanacheServer(8545, 1337, [account]);
+  await ganacheServer.ready();
+
+  const deployedArtifacts = await deploy();
+
+  // TODO: is this the best way to add these addresses?
+  process.env.ETH_ASSET_HOLDER_ADDRESS = deployedArtifacts.ethAssetHolder;
+  process.env.ERC20_ADDRESS = deployedArtifacts.token;
+  process.env.ERC20_ASSET_HOLDER_ADDRESS = deployedArtifacts.tokenAssetHolder;
```
snario, line 18: You could name the artifacts in all-caps in the `deploy.ts` file, then:

`process.env = {...process.env, ...deployedArtifacts};`

packages/server-wallet/src/onchain-service/transaction-submission.ts
```diff
@@ -0,0 +1,124 @@
...
+    noHash: 'no transaction hash found in tx response',
+    underpricedReplacement: 'replacement transaction underpriced',
+  };
+
+  static isKnownErr(errorMessage: string): boolean {
+    const idx = Object.values(TransactionSubmissionError.knownErrors).findIndex(err =>
+      errorMessage.includes(err)
+    );
+    return idx !== -1;
+  }
```
snario, line 32: I think I meant that this seems generic enough to be on `ChannelWalletError`

packages/server-wallet/src/onchain-service/transaction-submission.ts
snario: I think we may want to distinguish them as this code will not be inside `server-wallet` soon

packages/server-wallet/src/onchain-service/transaction-submission.ts
```diff
@@ -0,0 +1,124 @@
...
+
+import {
+  TransactionSubmissionServiceInterface,
+  MinimalTransaction,
+  TransactionSubmissionOptions,
+} from './types';
+
+// FIXME: replace with
+// import {WalletError as ChannelWalletError} from '@statechannels/server-wallet';
+import {WalletError as ChannelWalletError} from '..';
```
snario, line 12: I think this will actually work now based on my knowledge of yarn workspaces; no need to postpone

packages/server-wallet/src/onchain-service/onchain-service.ts
```diff
@@ -0,0 +1,297 @@
...
+          token: info.tokenAddress,
+        });
+    });
+
+    // Post to evt on every onchain deposit event
+    const contract = new Contract(
+      info.assetHolderAddress,
+      info.tokenAddress === constants.AddressZero
+        ? ContractArtifacts.EthAssetHolderArtifact.abi
+        : ContractArtifacts.Erc20AssetHolderArtifact.abi,
```
snario, line 273: I think this is a pretty critical mapping, and we may want to organize it somewhere else in the folder

packages/server-wallet/src/onchain-service/onchain-service.ts
```diff
@@ -0,0 +1,297 @@
...
+    const depositEvt = Evt.create<DepositedEvent & {block: number}>();
+
+    // Setup deposit evt so it will emit properly formatted event IFF
+    // - belongs to a registered channel
+    // - is for a subsequent deposit event
+    // And execute channel wallet and storage callbacks (wallet callback
+    // executed IFF a channel wallet has been attached to the service)
+    depositEvt
+      .pipe(e => {
+        return this.store.has(e.destination);
```
snario, line 229: Maybe transform the value into `this.store.get(e.destination)` after do you don't need to call it in subsequently piped functions

packages/server-wallet/src/onchain-service/onchain-service.ts
```diff
@@ -0,0 +1,297 @@
...
+      })
+      .pipe(e => {
+        const record = this.store.get(e.destination);
+        const prevHoldings = record?.Deposited?.destinationHoldings || BigNumber.from(0);
+        return prevHoldings.lt(e.destinationHoldings);
```
snario, line 234: We try to use `BN` where possible (a custom class which wraps `string`) to manage BigNumbers

packages/server-wallet/src/onchain-service/onchain-service.ts
```diff
@@ -0,0 +1,297 @@
...
+  }
+}
+
+export class OnchainService implements OnchainServiceInterface {
+  private provider: providers.JsonRpcProvider;
+  private attempts: number;
+
+  // NOTE: V0 has an in-memory store, which will eventually be
+  // shifted to more permanent storage
+  private store: Map<string, ChannelEventRecord> = new Map();
```
snario, line 71: You may want to consider making this "store" an external component with an in-memory implementation to start. Then you can add methods for storing events, transaction IDs (for in progress transactions), private keys, and things like this.

packages/server-wallet/src/onchain-service/onchain-service.ts
```diff
@@ -0,0 +1,297 @@
...
+    reason: Values<typeof OnchainServiceError.reasons>,
+    public readonly data: any = undefined
+  ) {
+    super(reason);
+  }
+}
+
+export class OnchainService implements OnchainServiceInterface {
+  private provider: providers.JsonRpcProvider;
+  private attempts: number;
```
snario, line 67: I think this concept should be internal to the `TransactionSubmissionServiceInterface`

snario (changes requested):
1. We'll need to bump the version of `nitro-protocol`

2. I think you should add some `pino` logging throughout this for key operations (e.g., callback registered for channel, event detected, call to wallet made)

3. You may want to make some kind of system architecture diagram, and attach it to the PR / put it in a README to explain what the core loop of this piece of software is. I think it's important to be considering which components can fail, which should or should not be "retried", how to reboot from a crashed process, how to have a clean and explicit "state" representation of the system, etc. Not all of this is required for V0, but we can make some early decisions _now_ that will simplify things in V1, V2, etc.

packages/server-wallet/src/onchain-service/transaction-submission.ts
LayneHaber: I'm thinking we may not really even want to have these errors extend the `WalletError` in that case, since changing the type means I always have to change the base class.

Is there anywhere that the code *expects* the error to be a `WalletError`?

packages/server-wallet/src/onchain-service/onchain-service.ts
LayneHaber: Hm, I'm not so sure. I like that you can specify a number to retry at the function level in the transaction submission service. I also feel like that is something you would just want to set on default when you set up your onchain service. I don't feel super strongly, but the retries at the API level is nice

packages/server-wallet/src/onchain-service/transaction-submission.ts
LayneHaber: Other wallet errors don't have the `knownErr`s fields, but can add it to the class. Is there some other instance in the wallet where this type of error message checking would be useful?

packages/server-wallet/src/onchain-service/onchain-service.ts
LayneHaber: I'm not sure what you mean here, can you explain a bit more?

LayneHaber:
> I think you should add some pino logging throughout this for key operations (e.g., callback registered for channel, event detected, call to wallet made)

Agree, but not sure how your logging is setup. Kind of wanted to wait to add that until it was in its final living place, since there may be some difference in logging interfaces. Is there some logger service interface you guys use or do you just use the `pino` package directly?

packages/server-wallet/src/onchain-service/transaction-submission.ts
LayneHaber: It keeps failing the build for me, so I'll just leave the comments for now

snario:
We can just import `pino` directly and use it — the most important thing is making decisions about what logs we want to make at what log levels. The destination of those logs can be configured later. Generally we like to do `"Verb happened", { detailName: "details..." }`

LayneHaber:
Okay. I was talking more about interfaces that wrap pino, but we can deal with that later when we move it back into our stuff. 

Last time we went off the deep end in the logging, and it ended up getting out of hand pretty quickly. Are there any general logging strategies/levels you have already been using within the repo?

packages/server-wallet/src/onchain-service/onchain-service.ts
LayneHaber: Going to take this out of the onchain service completely and put it in the `submitTransaction` API with a default value

snario:
> Okay. I was talking more about interfaces that wrap pino, but we can deal with that later when we move it back into our stuff.
> 
> Last time we went off the deep end in the logging, and it ended up getting out of hand pretty quickly. Are there any general logging strategies/levels you have already been using within the repo?

I don't think we have a philosophy around logging. It seems like a developer responsibility to think about what kinds of logs are important to see. I'd say probably `INFO` should show one line per channel registered, event observed, and wallet function called (i.e., "start monitoring for events", "observed an event", "handled event"). Maybe `DEBUG` should have more info. I'd say make best guesses for now and we can always adjust as needed.

packages/server-wallet/src/onchain-service/store.ts
```diff
@@ -0,0 +1,183 @@
...
+export class TransactionSubmissionStore implements TransactionSubmissionStoreInterface {
+  // NOTE: V0 has an in-memory store, which will eventually be
+  // shifted to more permanent storage
+  private readonly transactions: Map<
+    string,
+    TransactionModel<keyof typeof TransactionStatuses>[]
+  > = new Map();
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  constructor() {}
```
snario, line 32: Is this necessary?

packages/server-wallet/src/onchain-service/store.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  // Used to save a transaction *before* it is sent to mempool
+  saveTransactionRequest(channelId: Bytes32, tx: NoncedMinimalTransaction): Promise<void> {
+    const existing = this.transactions.get(channelId) || [];
+    const idx = existing.findIndex(
+      e => e.status === 'pending' && (e.data as NoncedMinimalTransaction).nonce === tx.nonce
+    );
+    return Promise.resolve(this._updateTransactions(channelId, idx, {status: 'pending', data: tx}));
```
snario, line 40: I don't think it's necessary to wrap return values in `Promise.resolve` really. Why not use the `async` function modifier and `await`?

packages/server-wallet/src/onchain-service/store.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    // Update record
+    if (previousRecordIndex === -1) {
+      // No record of pending transaction found
+      existing.push(data);
+    } else {
+      // Update existing record
+      existing[previousRecordIndex] = data;
+    }
+    this.transactions.set(channelId, existing);
```
snario, line 103: I think the call to `existing.push` would make this redundant?

packages/server-wallet/src/onchain-service/store.ts
```diff
@@ -0,0 +1,183 @@
...
+    return latest;
+  }
+
+  public saveEvent<T extends ContractEventName>(
+    channelId: Bytes32,
+    data: ChannelEventRecordMap[T]
+  ): Promise<void> {
+    const existing = this.events.get(channelId) || [];
+    const idx = existing.findIndex(e => {
+      return JSON.stringify(e) === JSON.stringify(data);
```
snario, line 163: Do events not have any other identifiers?

packages/server-wallet/src/onchain-service/store.ts
```diff
@@ -0,0 +1,183 @@
...
+    });
+    if (idx === -1) {
+      const updated = [...existing, data];
+      this.events.set(channelId, updated);
+    }
+    return Promise.resolve();
+  }
+
+  public createChannel(channelId: Bytes32): Promise<void> {
```
snario, line 172: Maybe rename to `registerChannel` to avoid confusion

packages/server-wallet/src/onchain-service/store.ts
```diff
@@ -0,0 +1,183 @@
...
+      existing.push(data);
+    } else {
+      // Update existing record
+      existing[previousRecordIndex] = data;
+    }
+    this.transactions.set(channelId, existing);
+  }
+}
+
+export class OnchainServiceStore implements OnchainServiceStoreInterface {
```
snario, line 107: Maybe we should call this `ChainMonitor` and the other one `TransactionSender` and completely decouple them? They don't need to share any state, I don't think.

packages/server-wallet/src/onchain-service/transaction-submission.ts
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+
+    throw new TransactionSubmissionError(TransactionSubmissionError.reasons.failedAllAttempts, {
+      attempts,
+      indexedErrors,
+    });
+  }
+
+  private async _sendTransaction(
```
snario, line 122: I think either this function, or the `submitTransaction` should probably be tolerant to being called more than once and thus checking storage first before re-submitting

packages/server-wallet/src/onchain-service/onchain-service.ts
snario: The mapping from `tokenAddress` to asset holder ABI might not be as simple as this. I can think of one example where you might have an asset holder for a token that is using neither `EthAssetHolderArtifact` nor `Erc20AssetHolderArtifact`. I'm not sure where else we might want to put this, but it probably shouldn't be inline.

packages/server-wallet/src/onchain-service/store.ts
LayneHaber: Depends on how they want to be managed, do you now need to have both a chain monitor and transaction sender within the app/somewhere the wallet can access?

But overall, don't have a problem with decoupling these, especially in v0 when nailing down the wallet <--> chain service flow is more of a v1 scope

packages/server-wallet/src/onchain-service/store.ts
LayneHaber: I went back and forth on this 😆 

Will change to `register`

packages/server-wallet/src/onchain-service/store.ts
LayneHaber: Atm no, but @andrewgordstewart and I talked about restructuring the events and that was one of the things that came up. We're going to try to use the transaction hash.

One question I have is if transactions can be bundled and emit multiple events, since then we may want to use uuid as the identifier. Kind of took a shortcut here because I knew this would change pretty completely when its v1 with permanent storage

packages/server-wallet/src/onchain-service/store.ts
```diff
@@ -0,0 +1,183 @@
...
+    channelId: Bytes32,
+    tx: NoncedMinimalTransaction,
+    reason: string
+  ): Promise<void> {
+    // Find the submitted transaction
+    const existing = this.transactions.get(channelId) || [];
+    // The transaction can fail at any time before it is sent or during
+    // mining, so look by nonce
+    const idx = existing.findIndex(
+      e => (e.data as providers.TransactionResponse | NoncedMinimalTransaction).nonce === tx.nonce
```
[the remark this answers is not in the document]
LayneHaber: Wen't back and forth on this one too, will change it back

packages/server-wallet/src/onchain-service/store.ts
LayneHaber: It's not async since it's all just mapping functions.

I just used `Promise.resolve` instead of typing the function as `async` when there's nothing to `await` yet. I figure may as well do this, since in v1 the calls to storage will definitely be `async`

packages/server-wallet/src/onchain-service/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+  callback: (event: any) => void | Promise<void>,
+  filter?: (event: any) => boolean,
+  timeout?: number
+): Evt<any> | Promise<any> => {
+  if (filter && timeout) {
+    return evt.attach(filter, timeout, callback);
+  } else if (filter) {
+    return evt.attach(filter, callback);
+  }
+  return evt.attach(callback);
```
andrewgordstewart, line 43: This looks a bit nasty, partially due to the signature of `evt.attach`. In particular, it seems to be missing the casae where `timeout` is truthy, and `filter` is not.

I can think of two solutions:
1. overwrite default values with the provided values, if they are truthy (preferable, if that makes sense)
2. the following suggestion
```suggestion
  const attachArgs = [];
  if (filter) attachArgs.push(filter);
  if (timeout) attachArgs.push(timeout);
  attachArgs.push(callback)
  
  return evt.attach(...attachArgs);
```

packages/server-wallet/src/onchain-service/onchain-service.ts
LayneHaber: Okay, we can separate it into a function but doing anything more than that is just premature since we only have an `EthAssetHolderArtifact` and an `Erc20AssetHolderArtifact`

packages/server-wallet/src/onchain-service/transaction-submission.ts
LayneHaber: I'm going to leave it as is until it's moved out of `experimental` or until the packages import the `@ethersproject` dependencies.

Just came out of a bit of a nasty ethers upgrade and the versioning was mentioned in the documentation for the experimental packages. Open to using this for v1 if people don't think that will be a problem though!

packages/server-wallet/src/onchain-service/transaction-submission.ts
LayneHaber: So how would you differentiate between the a transaction being re-submitted vs a transaction with the same call data being submitted intentionally while another is in mempool? Is there an assumption we can make that the call data would never be the same? Would we want to make people wait until that transaction has been mined before they are allowed to send a second?

packages/server-wallet/src/onchain-service/store.ts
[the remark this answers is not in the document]
LayneHaber: Also simplified a lot of the tx store -- really this should be done in v1 with more permanent storage.

packages/server-wallet/src/wallet/index.ts
LayneHaber: Wallet and chain service interaction improvements should be done in v1, resolving comment.

packages/server-wallet/src/onchain-service/transaction-submission.ts
LayneHaber: Also, this seems more like a v1 issue, where we can do some more creative unique constraints using systems like postgres.

LayneHaber: (And also store-level transactions)

packages/server-wallet/src/onchain-service/onchain-service.ts
snario: For TG we have a unique asset holder which is neither of these two

packages/server-wallet/src/onchain-service/transaction-submission.ts
snario: I can't think of any example in the nitro protocol where a transaction would have the exact same calldata of a previous transaction

packages/server-wallet/package.json
```diff
@@ -88,6 +90,7 @@
     "test:ci": "yarn test --bail",
     "test:e2e": "SERVER_DB_NAME=payer yarn jest -c ./jest/jest.e2e.config.js",
     "test:stress": "SERVER_DB_NAME=payer yarn jest -c ./jest/jest.stress.config.js --runInBand",
+    "test:chain": "RPC_ENDPOINT='http://localhost:8545' jest --runInBand -c ./jest/jest.chain.config.js",
```
snario, line 93: This `RPC_ENDPOINT` may as well be configured in `jest.chain.config.js`?

packages/server-wallet/src/config.ts
```diff
@@ -39,6 +41,8 @@ const config: ServerWalletConfig = {
   rpcEndpoint: process.env.RPC_ENDPOINT,
   chainNetworkID: process.env.CHAIN_NETWORK_ID || '0x0',
   ethAssetHolderAddress: process.env.ETH_ASSET_HOLDER_ADDRESS,
+  erc20Address: process.env.ERC20_ADDRESS,
```
snario, line 44: I don't think it makes sense to have something called `erc20Address` / `ERC20_Address`. This information will probably need to be a mapping of some kind? For example, DAI, MKR, and USDC all have different addresses.

snario:
Next step will be ensuring the tests pass on CI

packages/server-wallet/src/onchain-service/README.md
```diff
@@ -0,0 +1,9 @@
+# Onchain Service
+
+## Background
+
+This is an implementation of connext [IRC-2](https://github.com/connext/IRCs/blob/master/2-onchain-interactions.md)
+
+## Service Architecture
+
+![alt text](https://github.com/connext/IRCs/blob/02-onchain-service/assets/IRC-2-architecture.png)
```
andrewgordstewart, line 9: Broken link

packages/server-wallet/src/onchain-service/__chain-test__/onchain-service.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    const receipt = await response.wait();
+    expect(receipt.transactionHash).toBe(response.hash);
+
+    // Wait for event
+    const emitted = await p;
+    expect(emitted).toMatchObject({
+      amountDeposited: value,
+      destination: channelId,
+      destinationHoldings: value,
+    });
```
andrewgordstewart, line 94: If you rename the promise `p` to be `emitted` above, you can rewrite this as

```suggestion
    await expect(emitted).resolves.toMatchObject({
      amountDeposited: value,
      destination: channelId,
      destinationHoldings: value,
    });
```

packages/server-wallet/src/onchain-service/__chain-test__/transaction-submission.test.ts
```diff
@@ -0,0 +1,111 @@
...
+import {TransactionSubmissionStore} from '../store';
+
+import {MinimalTransaction} from '..';
+
+const verifyResponse = async (
+  response: providers.TransactionResponse,
+  tx: MinimalTransaction
+): Promise<void> => {
+  expect(response.to).toEqual(tx.to);
+  expect(response.value?.toString()).toEqual(tx.value?.toString());
```
andrewgordstewart, line 15: Is it acceptable for both the expected and the received to be `undefined`?

packages/server-wallet/src/onchain-service/__chain-test__/transaction-submission.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    const responses = await Promise.all([
+      service.submitTransaction(channelId, tx),
+      service.submitTransaction(channelId, tx),
+    ]);
+    await Promise.all(responses.map(response => verifyResponse(response, tx)));
+  });
+
+  it('should retry transactions with known errors', async () => {
+    // Create a "wallet" that forces an error
+    const mock = getFailingWallet(TransactionSubmissionError.knownErrors.badNonce);
```
andrewgordstewart, line 84: I like the `knownErrors` pattern, but it might deserve a better name and place.

When I first saw it, I was confused, thinking that what you were calling `knownErrors` is what we call `reasons`: the types of errors that this thing is expected to **throw**.

Instead, they are the types of errors that this thing internally **catches**, and that is a bit of an implementation detail. Therefore, `knownErrors` doesn't seem like it belongs to the `TransactionSubmissionError` class.

I [can't think](https://martinfowler.com/bliki/TwoHardThings.html) of a better name off the top of my head. Something like `EthersErrorsThatWeHandle` is not good.

packages/server-wallet/src/onchain-service/__chain-test__/transaction-submission.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    // Recreate service
+    service = new TransactionSubmissionService(provider, mock, new TransactionSubmissionStore());
+
+    const tx = getTransaction();
+    await expect(service.submitTransaction(channelId, tx, {maxSendAttempts: 2})).rejects.toEqual(
+      new TransactionSubmissionError(TransactionSubmissionError.reasons.unknownError)
+    );
+  });
+});
```
andrewgordstewart, line 111: This comment is for the [`PNG` file](https://github.com/statechannels/statechannels/blob/5d5b5947cf75e7e32f5c962763fa9209bc74c354/packages/server-wallet/src/onchain-service/assets/onchain-architecture.png) that I see below. I can't comment on it directly.

I'm not sure about including that architecture diagram here, because
A. it is quite different from the way the browser-wallet works
B. it is quite different from the way the server-wallet will work (which is a bit different from A)
C. it uses different terminology than the rest of this codebase

packages/server-wallet/src/onchain-service/onchain-service.ts
```diff
@@ -0,0 +1,278 @@
...
+  } as const;
+  constructor(
+    reason: Values<typeof OnchainServiceError.reasons>,
+    public readonly data: any = undefined
+  ) {
+    super(reason, data);
+    logger.error(reason, {type: this.type, ...(data || {})});
+  }
+}
+ */
```
andrewgordstewart, line 77: ✂️ can this be deleted?

packages/server-wallet/src/onchain-service/types.ts
```diff
@@ -0,0 +1,156 @@
...
+  [K in keyof ChannelEventRecordMap]: Evt<ChannelEventRecordMap[K]>;
+};
+
+// Defines the interface for the service that is responsible for handling
+// all onchain interactions (i.e. events and transactions)
+export interface OnchainServiceInterface {
+  registerChannel(channelId: Bytes32, assetHolders: Address[]): Promise<void>;
+
+  // TODO: remove in v1
+  attachChannelWallet(wallet: ChannelWallet): void;
```
andrewgordstewart, line 69: Let's make an issue for this.

packages/server-wallet/src/onchain-service/types.ts
```diff
@@ -0,0 +1,156 @@
+import {Address, Bytes32} from '@statechannels/client-api-schema';
```
andrewgordstewart, line 1: Feedback for this file.

I came to this file to try and see what is the interface that the wallet will use. But this file contains a mix of internal types (ie. implementation details) as well as external interfaces.

It might still make sense for this file to contain both of these types, so that all the types are imported from the same file. If that's the case, perhaps they could be organized (eg. with some comment headers) so that the external interface is clear.
```
// BEGIN EXTERNAL INTERFACE
 export type OnchainServiceInterface = { ... }
...
export type TransactionSubmissionStoreInterface = { ... } // Is this an external interface?
// END EXTERNAL INTERFACE

// BEGIN INTERNAL TYPES

// END INTERNAL TYPES
```

I might prefer two separate files, maybe `types/internal.ts` and `types/external.ts`

andrewgordstewart (approved):
I would prefer to
1. skip the test on circle (so that our CI passes) and merge this as is
2. @LayneHaber can deal with any of mine or @snario 's comments that are _quick_
3. Plan out the outstanding changes in a visible, "resilient" location, _external to this PR_.
  - That could be a notion document, a github project, or simply a github issue.
  - We can sync on what needs to next be done

Some general thoughts:
## The good
- This PR has two clear tests that I can read, and I can see that it is working as intended
- [This doc](https://github.com/connext/IRCs/blob/master/2-onchain-interactions.md) is a great resource that I can use to glean context.
- The diff in `yarn.lock` is acceptable, considering that an entire new feature set is included.
  - We should strive to include heavyweight dependencies only if they add enough added value
  - It's a big pet peeve of mine that `yarn install` burns 5% of my battery, and (more importantly) slows down CI

## The bad
- The tests don't pass on circle
- There are two planning docs for this project.
  1. https://github.com/connext/IRCs/blob/master/2-onchain-interactions.md 
     - This doc is now a MD file that I can't comment on, and does not live nearby to the implementation (it's a different GH organization)
  2. [notion doc](https://www.notion.so/534b015f85554b4dac5122946e656bdb?v=17b508d14f344ff3ba5955805262e7c4&p=fcaccef737404520b41280e97ee0ae2e)
    - This doc is not publicly available and now appears out of sync.
- This PR seems to contain two independent bits of work: transaction submission, and event monitoring
  - These probably should have been self-contained PRs. I think the narrower scope would have improved the process, and made it easier to get something high-quality, with tests that pass on circle, quickly.
- `TODO`s and `FIXME`s are a bat habit that we all make. We've got loads of them in our codebase. I was hoping that since this was going to be a _simple_ and _self-contained_ project, that it wouldn't need as many as there are here.

## The ugly
- The amount of conversation on this PR is a bit overwhelming, so much so that I can't really follow it.
- The project did not appear to be sufficiently well-defined. As @LayneHaber's first contribution to the @statechannels project, you were missing a load of internalized knowledge that made your job unnecessarily difficult in getting this merged quickly.

PR #2531, opened 2020-09-01
server-wallet: Change main to top level index

kerzhner:
Also, export the `Outgoing` type.

### DIFF
--- a/packages/server-wallet/package.json
+++ b/packages/server-wallet/package.json
@@ -58,7 +58,7 @@
   "lint-staged": {
     "src/**/*.ts": "eslint --max-warnings=0"
   },
-  "main": "lib/src/wallet/index.js",
+  "main": "lib/src/index.js",
   "private": false,
   "resolutions": {
     "broccoli": "3.3.3",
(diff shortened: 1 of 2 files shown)


snario:
What are the best practices for exporting types versus classes in TypeScript? Maybe it makes sense to only export non-types from `src/index.ts` but then separately have some `src/types.ts` which is exporting _all_ types? You can use `"types": "lib/src/types.d.ts"` in `package.json` and I think it will achieve the same effect as this PR but may be more cleanly organized?

kerzhner:
A couple thoughts:
- I think we should point the main key of `package.json` to the top level `index.ts` as opposed to an `index.ts` deeper in the package.
- I am not sure if exporting types from `index.ts` is an antipattern, but I am open to having a dedicated `types.ts` file if we think including types in `index.ts` is a bad idea.

kerzhner:
A cursory look at documentation suggests that types in index.ts are maybe fine: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

snario (approved):

PR #2533, opened 2020-09-03
Export configured knex

kerzhner:
Builds on https://github.com/statechannels/statechannels/pull/2531.

Exporting configured knex is intended to help packages that use `server-wallet` and need to migrate the `server-wallet` database.

### DIFF
--- a/packages/server-wallet/src/index.ts
+++ b/packages/server-wallet/src/index.ts
@@ -2,5 +2,6 @@ import {Message} from '@statechannels/wallet-core';
 
 import {Wallet} from './wallet';
 import {Outgoing} from './protocols/actions';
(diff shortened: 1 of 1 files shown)


andrewgordstewart (approved):
I don't know if I agree with the name that it gets exported as. Why not export it as adminKnex?

kerzhner:
> I don't know if I agree with the name that it gets exported as. Why not export it as adminKnex?

We could do that. Would we ever export non-`adminKnex` from the `server-wallet`? To me, that seems unlikely since consumers of `server-wallet` should not form queries to the `server-wallet` DB (although `adminKnex` allows consumers to do just that unfortunately).

Should the export be named `knex`?

PR #2540, opened 2020-09-07
[nitro-protocol] Emit more information in events

### LINKED ISSUE #2526: [nitro-protocol] Insufficient event information in `respond`
**Describe the bug :bug:**
In the `respond` function of `ForceMove.sol` the challenge is cleared and the data emitted from the `ChallengeCleared` event is not verbose enough to create a new state in the application (as opposed to `ChallengeRegistered`). Consider the following scenario:

- A and B have a channel together, where B is the turn taker at nonce `n` and A has already sent a signed `n-1` to B
- B becomes malicious and stops responding, A wants to get their funds out of the channel
- A starts the challenge by calling `forceMove` with state `n-1`, the `ChallengeRegistered` event is emitted
- B calls `respond` to clear the challenge with some state `n`, the `ChallengeCleared` event is emitted

A could determine the data sent to `respond` by parsing the calldata, but this is much less intuitive than using existing events to capture channel information.

Instead, the `respond` should also emit a `ChallengeRegistered` event instead of only a `ChallengeCleared` event.

geoknee:
Closes #2526 .

I did this mostly as an experiment to see how much extra gas would be needed.

Contract |              Deployment |  Method |                             Calls |  Min |    Avg |     Max |   
--- |                   --- |         --- |                                --- |    --- |    --- |     --- |   
TESTForceMove BEFORE |         2218375 |     * |                                  * |      * |      * |       * |     
-* |                     * |           forceMove |                          9 |      90003 |  109503 |  147964 |
-* |                     * |           checkpoint |                         3 |      65780 |  70798 |   80792 | 
-* |                     * |           respond |                            7 |      58520 |  60740 |   66254 | 
TESTForceMove AFTER |         2248392 |     * |                                  * |      * |      * |       * |     
-* |                     * |           forceMove |                          9 |      86885 |  106459 |  144852 |
-* |                     * |           checkpoint |                         3 |      78198 |  83208 |   93198 | 
-* |                     * |           respond |                            7 |      65686 |  67874 |   73327 | 

 I think it is definitely debatable about whether we should be emitting redundant data in events to the sole end of a nicer dev experience. This changeset is towards the "emit everything" end of the tradeoff scale, with the notable exception that I have removed the [`FixedPart`](https://docs.statechannels.org/protocol-tutorial/execute-state-transitions#fixed-and-variable-parts) -- this is easily reversed, but it just seemed like an obvious thing for a wallet to store against the channelId. Otherwise:

- `ChallengeRegistered` now emits `isFinalCount` instead of `isFinal` for the challenge state
- `ChallengeCleared` is now two events (overloaded). In the case of a checkpoint, it emits the entire support proof that was submitted. In the case of a `respond`, it emits the response state and signature (clients can construct a support proof by inspecting the corresponding and prior `ChallengeRegistered` event.)


packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -85,8 +82,7 @@ contract ForceMove is IForceMove {
             largestTurnNum,
             uint48(now) + fixedPart.challengeDuration,
             challenger,
-            isFinalCount > 0,
-            fixedPart,
+            isFinalCount,
```
geoknee, line 85: `isFinalCount > 0` is not sufficient information for a client, who only consumes these events and not `CALLDATA`, to maintain a support for the latest state. They might need that to launch a challenge. This change remedies that and makes it more consistent with emitting _all_ of the `variableParts` as we currently do.

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -781,21 +824,4 @@ contract ForceMove is IForceMove {
...
-        address challenger,
-        bool isFinal,
-        FixedPart fixedPart,
-        ForceMoveApp.VariablePart[] variableParts,
-        Signature[] sigs,
-        uint8[] whoSignedWhat
-    );
-
-    event ChallengeCleared(bytes32 indexed channelId, uint48 newTurnNumRecord);
-    event Concluded(bytes32 indexed channelId);
```
geoknee, line 800: It is not necessary to overwrite these events here, this contract can simply reference those in `IForceMove` which this contract extends / implements.

packages/nitro-protocol/test/contracts/ForceMove/forceMove.test.ts
```diff
@@ -190,39 +190,30 @@ describe('forceMove', () => {
...
-          variableParts: eventVariableParts,
-        } = event.args;
+        // Check this information is enough to respond or checkpoint
+        // (fixed part is assumed known)
+
+        const formatVariablePartForEthers = (variablePart: VariablePart) => [
+          variablePart.outcome,
+          variablePart.appData,
+        ];
+        const formatSignatureForEthers = (sig: Signature) => [sig.v, sig.r, sig.s];
```
geoknee, line 204: This is quite a bit neater and transparent IMO. Likely to change when we upgrade ethers.

PR #2542, opened 2020-09-08
Upgrade all packages to ethers 5.0.12

geoknee:


geoknee:
This has gone mostly smoothly with one exception. A test failure on ERC20 deposits that I can't reproduce locally 😞 .

Specifically, the check that the ERC20 balance of the depositor changes correctly:

https://github.com/statechannels/statechannels/blob/c7a796b788122775ee497c9741e8ead31a7efb94/packages/nitro-protocol/test/contracts/ERC20AssetHolder/deposit.test.ts#L99-L103

(test failure suggests balance is not decremented).

geoknee:
Might be a race condition exposed by some small timing adjustments? I might try `--runInBand` as a diagnostic...

Pretty strange though, as I think this is the only test that targets the `Token` contract.

geoknee:
OH

packages/nitro-protocol/test/contracts/ERC20AssetHolder/deposit.test.ts
```diff
@@ -75,12 +77,13 @@ describe('deposit', () => {
       const {data: amountTransferred} = getTransferEvent(events);
       expect(held.eq(amountTransferred)).toBe(true);
     }
+
+    const balanceBefore = BigNumber.from(await Token.balanceOf(signer0Address));
     const tx = ERC20AssetHolder.deposit(destination, expectedHeld, amount);
```
geoknee, line 82: The test is now passing, I think because I put the check on `balanceBefore` _ahead_ of forming the `deposit` transaction. This avoids a race.

geoknee: (if the `deposit` transaction gets mined before the `balanceOf` call, we actually store `balanceAfter` instead of `balanceBefore`, which explains the test failure).

packages/nitro-protocol/test/contracts/ConsensusApp/validTransition.test.ts
```diff
@@ -27,15 +25,15 @@ beforeAll(async () => {
...
-    ${true}  | ${[0, 2]} | ${['0x1', '0x1']} | ${['0x1', '0x2']} | ${'valid consensus -> propose'}
-    ${false} | ${[0, 1]} | ${['0x1', '0x1']} | ${['0x1', '0x2']} | ${'invalid consensus -> propose: votesReqd too low'}
-    ${true}  | ${[2, 1]} | ${['0x1', '0x1']} | ${['0x2', '0x2']} | ${'valid vote'}
-    ${false} | ${[1, 1]} | ${['0x1', '0x1']} | ${['0x2', '0x2']} | ${'invalid vote: votesReqd not decreased'}
-    ${true}  | ${[1, 2]} | ${['0x1', '0x1']} | ${['0x2', '0x1']} | ${'valid veto'}
-    ${true}  | ${[2, 2]} | ${['0x1', '0x1']} | ${['0x2', '0x2']} | ${'valid pass'}
-    ${true}  | ${[1, 0]} | ${['0x1', '0x2']} | ${['0x2', '0x']}  | ${'valid finalVote'}
-    ${false} | ${[1, 0]} | ${['0x1', '0x3']} | ${['0x2', '0x']}  | ${'invalid finalVote: proposedOutcome1 ≠ currentOutcome2'}
+    isValid  | votesReqd | outcomes            | proposedOutcomes    | description
+    ${true}  | ${[0, 2]} | ${['0x01', '0x01']} | ${['0x01', '0x02']} | ${'valid consensus -> propose'}
```
andrewgordstewart, line 29: This change suggests that these should be
```suggestion
    ${true}  | ${[0, 2]} | ${[one, one]} | ${[one, two]} | ${'valid consensus -> propose'}
```

I don't think this PR is the right time to do that :P

andrewgordstewart (approved):
I looked at every line, hopefully long enough to see that there are no controversial changes here.

The size of the change set suggests that we over-use magic constants.

packages/docs-website/docs/protocol-tutorial/finalize-a-channel-happy.md
```diff
@@ -45,14 +45,14 @@ for (let i = 0; i < 3; i++) {
   participants[i] = wallets[i].address;
 }
 const chainId = '0x1234';
-const channelNonce = bigNumberify(0).toHexString();
+const channelNonce = BigNumber.from(0).toHexString();
```
geoknee, line 48: We should probably _not_ be making changes to any .md files in `docs-website/docs/protocol-tutorial` since the accompanying repo is still using ethers 4 

https://github.com/statechannels/nitro-tutorial/blob/2cf51f00ddf9b47d83685b3b14d9b58d24e6585d/package.json#L19

packages/nitro-protocol/src/contract/consensus-app.ts
```diff
@@ -30,8 +26,8 @@ export async function validTransition(
   const toVariablePart = getVariablePart(toConsensusData, toOutcome);
   const turnNumB = 0; // This isn't actually used by the contract so any value works
 
-  const contract = new Contract(contractAddress, ConsensusAppContractInterface.abi, signer);
-  return await contract.functions.validTransition(
+  const contract = new Contract(contractAddress, ConsensusAppArtifact.abi, signer);
```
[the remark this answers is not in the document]
geoknee: Perhaps! As it is now, it is [documented](https://docs.statechannels.org/implementation-notes/consensus-app) in a way that suggests it is deprecated. I'd be up for a separate PR that rips it out (contract and all).

packages/docs-website/docs/protocol-tutorial/finalize-a-channel-happy.md
geoknee: Resolved in 4ba1f42

PR #2548, opened 2020-09-10
[nitro-protocol et al] Delete ConsensusApp.sol and related code, docs

### LINKED ISSUE #2543: delete the `consensus-app` code
Unrelated: Should we just delete the `consensus-app` code?

_Originally posted by @andrewgordstewart in https://github.com/statechannels/statechannels/pull/2542#discussion_r485099946_

geoknee:
Closes #2543


packages/integration-tests/create-channel.test.ts
```diff
@@ -101,7 +101,7 @@ describe('Client-Provider-Wallet', () => {
                       '0x00000000000000000000000063E3FB11830c01ac7C9C64091c14Bb6CbAaC9Ac7'
                   }
                 ],
-                assetHolderAddress: '0x4ad3F07BEFDC54511449A1f553E36A653c82eA57'
+                assetHolderAddress: expect.stringMatching(/^0x([a-fA-F0-9]{40})|0$/) // Matches any ethereum address
```
geoknee, line 104: The only real change that wasn't deletion. Prior to this change we were using a magic string for this contract address, and it changed under this changeset because the order of contract deployment changed (because I deleted one). Propose that we just match any address here, since the test isn't really concerned with the exact address.

andrewgordstewart (approved):
✂️

PR #2549, opened 2020-09-10
[nitro-protocol] Use @ethereum-waffle/compiler for contract compilation

geoknee:
I chose waffle since they seem to be among the best maintained dapp dev framework, and in particular [should allow us to get back to native solc compilation](https://github.com/ethereum/solidity/issues/8133), as well as avoiding the annoyingly long pause we get when running etherlime.

However, it wasn't clear at the outset that our doc generation tool, `solidoc` as well as our `jest-gas-reporter` tool, depends on the contract artifacts being organised a certain way (i.e. on the way that the framework transforms the output of `solc` and stores it in files). And using waffle breaks that. I had to `patch-package` my way out of this. 

This changeset does seem to shave about 1-2 minutes off of the `prepare` step on circle. 

TODO 
- [x] reinstate switchability of native compilation (let's not force devs to install native solc). We can do this by upgrading `waffle.json` to `waffle.js` and switching the config based on the env vars
- [x] actually enable native solc compilation
- [x] Fix any errors arising

Other notes
---
Waffle, buidler, truffle, etherlime and co are all frameworks that allow for deployment and testing. RN we have rolled our own testing with jest, ethers and some of our own devtools. We can probably(?) do deployment with just ethers, and then we will have no need of etherlime at all. 

Looks like we may be able to use jest and waffle together in interesting ways:
https://medium.com/ethworks/introducing-waffle-support-for-jest-experimental-331f39d630bd


patches/@ethereum-waffle+compiler+3.1.0.patch
```diff
@@ -0,0 +1,32 @@
...
+@@ -41,10 +41,10 @@ function getMinimalContent(contractJson, config) {
+     return JSON.stringify({ abi, bytecode }, null, 2);
+ }
+ async function saveOutputSingletons(output, config, filesystem = fsOps) {
+-    for (const [, file] of Object.entries(output.contracts)) {
++    for (const [source_path, file] of Object.entries(output.contracts)) {
+         for (const [contractName, contractJson] of Object.entries(file)) {
+             const filePath = path_1.join(config.outputDirectory, `${contractName}.json`);
+-            filesystem.writeFile(filePath, getContent(contractJson, config));
++            filesystem.writeFile(filePath, getContent(contractJson, output.sources[source_path] ? output.sources[source_path].ast : undefined, source_path, contractName, 'unknown', config));
```
geoknee, line 14: There is a small feature regression here, I didn't figure out how to get the compiler version into the artifact so it just emits 'unknown'.

yarn.lock
```diff
@@ -1900,6 +1900,87 @@
...
+  integrity sha512-qFwy9OHbbvltns4tz12mpbSryg8rCpATsdjM+PtSSXSJ7s04fcDlKWJgFERE3jsxE7GMGH8mfr4e7Jqncxxh7w==
+  dependencies:
+    "@resolver-engine/imports" "^0.3.3"
+    "@resolver-engine/imports-fs" "^0.3.3"
+    "@types/mkdirp" "^0.5.2"
+    "@types/node-fetch" "^2.5.5"
+    ethers "^5.0.1"
+    mkdirp "^0.5.1"
+    node-fetch "^2.6.0"
+    solc "^0.6.3"
```
geoknee, line 1939: AFAIK all of the frameworks that do compilation , including waffle, also do a bunch of other stuff (e.g. testing). An ideal lightweight solution would _only_ use this dependency https://www.npmjs.com/package/@ethereum-waffle/compiler

geoknee: 9d2d7b3

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
 
 defaults: &defaults
   environment:
-    USE_NATIVE_SOLC: false
+    USE_NATIVE_SOLC: true
```
snario, line 9: 🎉

snario (approved):

PR #2552, opened 2020-09-11
[server-wallet] Configure Wallet class via constructor and adopt knex passing pattern

geoknee:
https://vincit.github.io/objection.js/recipes/multitenancy-using-multiple-databases.html#knex-passing-pattern

Previously, things like `.query()` worked due to a globally bound `knex` instance.

Now, we need to do `query(knex)` where the `knex` object is instantiated in the `Wallet` constructor and passed down / threaded where necessary. 

The history on this branch involves a detour where I went too far and replaced `.query(tx)` with `.query(knex)`. As long as `tx` was already bound to `knex`, this was unnecessary. Have since rolled that back. The thing that unlocked my confusion in the end was for the various methods to accept a `TransactionOrKnex` type, which gives fine grained control, e.g. for tests to run a query without using a transaction. 

TODO
- [x] unit tests passing
- [x] e2e test passing
- [x] stress test passing
- [x] fix "jest did not exit 1 second..." warnings


andrewgordstewart:
I am curious about the tradeoffs involved in this approach.

A `Model` class that is globally bound to a DB connection has these (known) drawbacks:
1. You cannot have two wallet instances in the same process connected to separate DBs
2. The DB configuration must be done with environment variables.

I am not even sure if (2) is real. You could instead try globally binding the `Model` class when the wallet is initializing. (We can enforce that the connection is never changed, once it is globally bound to `Model`)

Which of these drawbacks are important enough to
A. Ask developers to import a `knex` object in any file that they interact with the DB
B. (Potentially?) make it easier to introduce deadlock bugs

I'm not sure that (B) is an issue, since tests should catch these bugs by restricting the connection pool to one connection.

(A) seems like more overhead compared to (1).

I realize that the work here has already been done, but I'd like to be clear on exactly what problem we're trying to solve, and make sure this is the best solution.

geoknee:
@andrewgordstewart thanks for the feedback. Perhaps we can discuss this at standup today.

packages/server-wallet/src/models/__test__/channel.test.ts
```diff
@@ -1,23 +1,28 @@
...
+
 import {Channel, ChannelError} from '../channel';
 import {seedAlicesSigningWallet} from '../../db/seeds/1_signing_wallet_seeds';
+import {extractDBConfigFromServerWalletConfig, defaultConfig} from '../../config';
 import {stateWithHashSignedBy} from '../../wallet/__test__/fixtures/states';
-import knex from '../../db/connection';
 
 import {channel} from './fixtures/channel';
 
+const knex: Knex = Knex(extractDBConfigFromServerWalletConfig(defaultConfig));
```
tomclose, line 10: Could this use `testKnex` too?

packages/server-wallet/src/wallet/__test__/integration/close-channel.test.ts
```diff
@@ -1,38 +1,50 @@
...
 import {Wallet} from '../..';
 import {seedAlicesSigningWallet} from '../../../db/seeds/1_signing_wallet_seeds';
 import {truncate} from '../../../db-admin/db-admin-connection';
-import knex from '../../../db/connection';
 import {stateWithHashSignedBy} from '../fixtures/states';
 import {alice, bob} from '../fixtures/signing-wallets';
 import {channel} from '../../../models/__test__/fixtures/channel';
+import {defaultConfig, extractDBConfigFromServerWalletConfig} from '../../../config';
+
+const knex: Knex = Knex(extractDBConfigFromServerWalletConfig(defaultConfig));
```
tomclose, line 12: Could this use `testKnex`?

tomclose (approved):

packages/server-wallet/src/config.ts
```diff
@@ -49,4 +52,21 @@ const config: ServerWalletConfig = {
   metricsOutputFile: process.env.METRICS_OUTPUT_FILE,
 };
 
-export default config;
+export const payerConfig: ServerWalletConfig = {...defaultConfig, postgresDBName: 'payer'};
+export const receiverConfig: ServerWalletConfig = {...defaultConfig, postgresDBName: 'receiver'};
```
andrewgordstewart, line 56: These seem like test constructs. Do we want to export them from a `src` file?

tomclose:
@andrewgordstewart it seems like your first point might actually be:

1. If the wallet runs in the same process as any other software that uses a global knex configuration, then they must use the same db.

I agree that you could pass the db config in via the constructor, but that seems like it might be misleading and lead to surprising behaviour:
```ts
const wallet1 = new Wallet(db1config);
const wallet2 = new Wallet(db2config);
// ooops wallet1 and wallet2 both point at db2...
```
It almost seems better to stick with the ENV variables if ☝️  is the behaviour, as there the expectation is that the db configuration is per process.

I think there are several advantages to genuinely having per-wallet config, including easier wallet-to-wallet testing (the only reason we need to spin up payment/receipt manager servers at the moment is because of this wallet limitation), and generally being a good citizen in other people's code bases.

packages/server-wallet/src/evm-validator.ts
```diff
@@ -18,7 +18,7 @@ const bytecodeCache: Record<string, string | undefined> = {};
 export const validateTransitionWithEVM = async (
   from: State,
   to: State,
-  tx: Objection.Transaction
+  txOrKnex: TransactionOrKnex
```
andrewgordstewart, line 21: The type `Objection.Transaction` was intended to help prevent a deadlock bug: `validateTransitionWithEVM` should _only_ be run from within a `lockApp` transaction, and so the type here means the compiler can help prevent deadlocks.

From a test, we can pass `undefined as any` for the value of `tx`.

packages/server-wallet/src/models/__test__/channel.test.ts
```diff
@@ -1,23 +1,28 @@
...
+
 beforeEach(async () => seedAlicesSigningWallet(knex));
+afterAll(async () => await knex.destroy());
 
 it('can insert Channel instances to, and fetch them from, the database', async () => {
   const vars = [stateWithHashSignedBy()()];
   const c1 = channel({channelNonce: 1234, vars});
 
-  await Channel.query()
+  await Channel.query(knex)
```
andrewgordstewart, line 19: These changes are quite noisy. An alternative strategy could be
```
import {Channel as UnboundChannel} from 'path/to/channel/model'
const Channel = UnboundChannel.bindKnex(testKnex)
```

If we had a "test" `Channel` model that was bound to the test knex, then the change can be even simpler:
```
import {TestChannel as Channel} from 'path/to/test/channel/model'
```

tomclose: Is this:
```ts
import {Channel as UnboundChannel} from 'path/to/channel/model'
const Channel = UnboundChannel.bindKnex(testKnex)
```
any different to this?
```ts
import {Channel} from 'path/to/channel/model'
Channel.bindKnex(testKnex)
```

andrewgordstewart:
> @andrewgordstewart it seems like your first point might actually be:
> 
> 1. If the wallet runs in the same process as any other software that uses a global knex configuration, then they must use the same db.
> 
> I agree that you could pass the db config in via the constructor, but that seems like it might be misleading and lead to surprising behaviour:
> 
> ```ts
> const wallet1 = new Wallet(db1config);
> const wallet2 = new Wallet(db2config);
> // ooops wallet1 and wallet2 both point at db2...
> ```

Another possible behaviour is
> ```ts
> const wallet1 = new Wallet(db1config);
> const wallet2 = new Wallet(db2config); // ERROR: trying to bind `Model` to a different connection.
> ```

I do not think this alternative is particularly pleasant.

> It seems better to stick with the ENV variables if ☝️ is the behaviour, as there the expectation is that the db configuration is per process.

I sort of agree, but it depends on the problem we're trying to solve. Is it that the DB is configured via environment variables? Is it that we can only connect to one DB per process?

> I think there are several advantages to genuinely having per-wallet config, including easier wallet-to-wallet testing (the only reason we need to spin up payment/receipt manager servers at the moment is because of this wallet limitation), and generally being a good citizen in other people's code bases.

There _are_ other reasons to spin up Payment/Receipt manager servers, such as:
- we want to profile just the `PaymentManager`
- we want to benchmark usage that is closer to production usage

>generally being a good citizen in other people's code bases.

Could you clarify this point?

packages/server-wallet/src/models/__test__/channel.test.ts
andrewgordstewart: I don't think that `Channel.bindKnex(testKnex)` has any side effects on the `Channel` class. Instead, it returns a class that is bound to `testKnex`.

https://vincit.github.io/objection.js/api/model/static-methods.html#static-bindknex

```
const BoundModel1 = SomeModel.bindKnex(knex1);
const BoundModel2 = SomeModel.bindKnex(knex2);

// Throws since the knex instance is null.
await SomeModel.query();
```

andrewgordstewart:
> I think there are several advantages to genuinely having per-wallet config, including easier wallet-to-wallet testing (the only reason we need to spin up payment/receipt manager servers at the moment is because of this wallet limitation), and generally being a good citizen in other people's code bases.

A further benefit of this approach is the ease in which we can parallelize tests that hit the database.

We have to run most of our tests in serial right now, because they interact with the same database.

We could try running tests within a transaction, but this might be complicated (impossible?), because we use DB transactions to enforce "application" invariants, such as "`updateChannel` must run sequentially among all wallet instances".

With this change, we could, in principle
- have the following databases: `server_wallet_test_1`, `server_wallet_test_2`, `server_wallet_test_3`, `server_wallet_test_4`
- maintain a pool of `knex` objects attached to these DBs
- each test file waits for a `knex` object, and binds its `Wallet` instances to that object

allowing us to run 4 test files in parallel without interference

However, I don't know how to share that resource pool across the separate jest workers.

packages/server-wallet/src/evm-validator.ts
geoknee: Ok -- that probably deserves a code comment to explain. And it might have to be `knex as any` now?

packages/server-wallet/src/wallet/__test__/integration/close-channel.test.ts
geoknee: This spawns a `new Wallet`, so it could use that wallet's `knex` exclusively.

packages/server-wallet/src/models/__test__/channel.test.ts
geoknee: b7d7eb7

packages/server-wallet/src/wallet/__test__/integration/close-channel.test.ts
geoknee: b7d7eb7

packages/server-wallet/src/config.ts
geoknee: dc74ccd

packages/server-wallet/src/evm-validator.ts
geoknee: 4d7af33

packages/server-wallet/src/models/__test__/channel.test.ts
geoknee: I think the alternative sounds fine, and that pattern is already in use in the e2e test:

https://github.com/statechannels/statechannels/blob/ac5330b40c66c56008e1dae769bdc7359fe1fa60/packages/server-wallet/e2e-test/e2e.test.ts#L35

However, I'm not sure it's worth doing just so we can use `.query()` rather than `.query(knex)`. It also means we have more of a mixture of
i) [`knex` passing](https://vincit.github.io/objection.js/recipes/multitenancy-using-multiple-databases.html#knex-passing-pattern) and
 ii) [model binding]( https://vincit.github.io/objection.js/recipes/multitenancy-using-multiple-databases.html#model-binding-pattern) 
going on: I prefer to keep the same patterns across the codebase if possible, and to have the tests mirror the behaviour of the actual code, too.

packages/server-wallet/src/metrics.ts
```diff
@@ -47,26 +46,27 @@ async function time<T>(label: string, cb: () => Promise<T>): Promise<T> {
   }
 }
 
+// GK TODO: unbreak this
```
geoknee, line 49: 🔧

packages/server-wallet/src/wallet/store.ts
```diff
@@ -196,7 +197,8 @@ export const Store = recordFunctionMetrics({
     const channel =
       maybeChannel || (await timer('get channel', async () => getOrCreateChannel(signedState, tx)));
 
-    if (!config.skipEvmValidation && channel.supported) {
+    if (!defaultConfig.skipEvmValidation && channel.supported) {
+      // TODO be better to inspect Wallet.walletConfig in case the defaultConfig was not used or overridden
```
geoknee, line 201: 🔧

geoknee: f873d82

packages/server-wallet/src/wallet/store.ts
```diff
@@ -43,13 +43,33 @@ const throwMissingChannel: MissingAppHandler<any> = (channelId: string) => {
...
 };
 
-export const Store = recordFunctionMetrics({
-  getFirstParticipant: async function(knex: Knex): Promise<Participant> {
-    const signingAddress = await Store.getOrCreateSigningAddress(knex);
-    return {participantId: signingAddress, signingAddress, destination: makeDestination(HashZero)};
-  },
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export class Store {
+  constructor(readonly timingMetrics: boolean) {
```
andrewgordstewart: Now that Store is a class, it can be constructed with a knex attribute, an, meaning we wouldn't need to pass around the knex object.

packages/server-wallet/src/metrics.ts
geoknee: https://github.com/statechannels/statechannels/pull/2552/commits/43da1b6fe821a32b1eecb5ce43a69cd7c0b11866

packages/server-wallet/src/wallet/store.ts
geoknee: Yeah that's a possible route -- you wouldn't need to pass around the `knex` object inside the `Store` class, but you would still need to accept transactions in the `Store` methods, and have them fall back to `this.knex` if no `tx` is provided.

Also the `Wallet` class does stuff with the `knex` object that totally bypasses the `Store`. So it seems right that the `Wallet` owns the `knex` object. I guess it seems marginal to me whether it should be passed in to the `Store` constructor, or each method. We could consider it in a future PR?

PR #2554, opened 2020-09-14
[monorepo] Minor tweaks to Circle config

geoknee:


.circleci/config.yml
```diff
@@ -32,16 +32,19 @@ save_dep: &save_dep
     key: v6-dependency-cache-{{ checksum "yarn.lock" }}
     paths:
       - node_modules
+      - packages/benchmarking/node_modules
```
kerzhner, line 35: It is unfortunate that we are unable to use globs: https://circleci.com/docs/2.0/caching/#basic-example-of-dependency-caching

.circleci/config.yml
```diff
@@ -179,7 +182,7 @@ jobs:
       - log_stats:
           file: prepare-stats
       - <<: *restore_dep
-      - run: yarn --frozen-lockfile
+      - run: yarn --frozen-lockfile --cache-folder ~.cache/yarn
```
kerzhner, line 185: Not sure I understand the reason for specifying the `cache-folder`

kerzhner (approved):

.circleci/config.yml
geoknee: There's an explanation here: https://circleci.com/docs/2.0/caching/#yarn-node. 

I thought it was worth putting in, in case it made caching of dependencies work better. It seems to me that caching is not working great at the moment, since I don't think I have ever seen `success Already up-to-date.` on circle, which I would have expected?

kerzhner: Is the `~.cache/yarn` directory persisted between workflow runs?

geoknee: No! good catch

geoknee: https://github.com/statechannels/statechannels/pull/2558 💥

PR #2560, opened 2020-09-16
[docs-website] Point to latest deployed wallet

geoknee:
We should point app developers at the most recent version of the wallet that we have deployed. In particular, v0.3.0 includes a number of fixes and passes the `integration-test` (so is known to work with `channel-client`, `iframe-channel-provider` and `client-api-schema` all @ 0.3.0). 

AFAIK more recent versions of those packages should also work with the 0.3.0 browser wallet, but we do not yet have a rigorous sem ver workflow in this repo so that is by no means guaranteed.

### DIFF
--- a/packages/docs-website/docs/app-devs/quick-start-dapp.md
+++ b/packages/docs-website/docs/app-devs/quick-start-dapp.md
@@ -30,7 +30,7 @@ The channel provider needs to be pointed at our hosted wallet:
 
 ```typescript
 await window.channelProvider.mountWalletComponent(
-  'https://xstate-wallet-v-0-3-0.statechannels.org/'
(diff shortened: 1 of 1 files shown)


snario (commented):
Maybe it should be the inverse, so https://xstate-wallet-v-0-2-0.statechannels.org is used by Web3Torrent on production and  https://xstate-wallet.statechannels.org is "latest".

geoknee:
Seems bad to have the code changing underneath https://xstate-wallet.statechannels.org. I think it's better to have predictable and immutable thing at each endpoint?

geoknee:
@snario I think we ought to merge this ASAP and then continue the discussion.

tomclose (approved):

PR #2561, opened 2020-09-16
[server-wallet] Groundwork for directly funded channels

kerzhner:
This PR implements:
1. CreateChannel objective as an explicit signal for channel creation.
1. Wallet emits a dummy `submitTransaction` when a directly funded channel needs funding.

Remaining work:
- [ ] Migrate to the new `fundChannel` chain service api from the `submitTransaction` api. This should probably be done in a follow on PR as this PR is getting large.

### DIFF
--- a/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
+++ b/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
@@ -5,7 +5,7 @@ const chainServiceRequests = 'chain_service_requests';
 
 export async function up(knex: Knex): Promise<any> {
   await knex.schema.alterTable(channels, table =>
-    table.specificType(chainServiceRequests, 'text[]')
+    table.specificType(chainServiceRequests, 'text[]').notNullable()
   );
 }
 
--- a/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
+++ b/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
@@ -7,7 +7,7 @@ const fundingStrategy = 'funding_strategy';
 export async function up(knex: Knex): Promise<any> {
   await knex.schema.alterTable(channels, table => {
     table.specificType(chainServiceRequests, 'text[]').notNullable();
-    table.string(fundingStrategy);
+    table.string(fundingStrategy).notNullable();
   });
 }
 
--- a/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
+++ b/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
@@ -4,7 +4,9 @@ const channels = 'channels';
 const chainServiceRequests = 'chain_service_requests';
 
 export async function up(knex: Knex): Promise<any> {
-  await knex.schema.alterTable(channels, table => table.jsonb(chainServiceRequests));
+  await knex.schema.alterTable(channels, table =>
+    table.specificType(chainServiceRequests, 'text[]')
+  );
 }
 
 export async function down(knex: Knex): Promise<any> {
--- a/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
+++ b/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
@@ -2,13 +2,18 @@ import * as Knex from 'knex';
 
 const channels = 'channels';
 const chainServiceRequests = 'chain_service_requests';
+const fundingStrategy = 'funding_strategy';
 
 export async function up(knex: Knex): Promise<any> {
-  await knex.schema.alterTable(channels, table =>
-    table.specificType(chainServiceRequests, 'text[]').notNullable()
-  );
+  await knex.schema.alterTable(channels, table => {
+    table.specificType(chainServiceRequests, 'text[]').notNullable();
+    table.string(fundingStrategy);
+  });
 }
 
 export async function down(knex: Knex): Promise<any> {
-  await knex.schema.alterTable(channels, table => table.dropColumn(chainServiceRequests));
+  await knex.schema.alterTable(channels, table => {
+    table.dropColumn(chainServiceRequests);
+    table.dropColumn(fundingStrategy);
+  });
 }
--- a/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
+++ b/packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
@@ -1,0 +1,12 @@
+import * as Knex from 'knex';
+
+const channels = 'channels';
+const chainServiceRequests = 'chain_service_requests';
+
+export async function up(knex: Knex): Promise<any> {
+  await knex.schema.alterTable(channels, table => table.jsonb(chainServiceRequests));
+}
+
+export async function down(knex: Knex): Promise<any> {
+  await knex.schema.alterTable(channels, table => table.dropColumn(chainServiceRequests));
+}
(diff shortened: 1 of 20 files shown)


snario:
What's the purpose of storing these chain requests in the database? Surely the chain service has a responsibility to remember if it has been tasked with something anyway, and can handle receiving duplicate requests?

kerzhner:
> What's the purpose of storing these chain requests in the database? Surely the chain service has a responsibility to remember if it has been tasked with something anyway, and can handle receiving duplicate requests?

One way to think about how the wallet functions is:
1. The wallet is notified of new information. For example: a new state channel state, application's intention to create a new state channel state, new chain event.
1. The wallet stores the new information in permanent storage.
1. The wallet starts the run loop. The run loop inspects the database state and decides if new actions need to be taken.

The consequence of the above architecture is that some wallet state needs to be modified to prevent an infinite loop in step 3. Ideally, this state would be in-memory. If I understand the wallet code correctly, there is currently no way to store state like this in memory as the wallet was designed to be a stateless reducer on top of a database store.

tomclose (approved):

packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
```diff
@@ -0,0 +1,19 @@
+import * as Knex from 'knex';
+
+const channels = 'channels';
+const chainServiceRequests = 'chain_service_requests';
+const fundingStrategy = 'funding_strategy';
+
+export async function up(knex: Knex): Promise<any> {
+  await knex.schema.alterTable(channels, table => {
+    table.specificType(chainServiceRequests, 'text[]').notNullable();
+    table.string(fundingStrategy).notNullable();
```
andrewgordstewart, line 10: I think we will have to edit this migration so that it can run in existing wallets.

As is, the migration will fail if there are any existing rows in the `channels` table, since this migration
1. requires a non-null `funding_strategy` value
2. does not provide a default value

If we provided a default value, then we would have to either
A. take down existing wallets for maintenance; or
B. accept that existing wallets will experience a pause while this migration runs, considering that it will require a write-lock on the `channels` table.

packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
```diff
@@ -0,0 +1,19 @@
+import * as Knex from 'knex';
+
+const channels = 'channels';
+const chainServiceRequests = 'chain_service_requests';
+const fundingStrategy = 'funding_strategy';
+
+export async function up(knex: Knex): Promise<any> {
+  await knex.schema.alterTable(channels, table => {
+    table.specificType(chainServiceRequests, 'text[]').notNullable();
```
andrewgordstewart, line 9: Should we add some DB constraints on the values of these columns?

packages/server-wallet/src/protocols/application.ts
```diff
@@ -28,6 +29,31 @@ const isFunded = ({app: {funding, supported}}: ProtocolState): boolean => {
...
+  if (app.chainServiceRequests.indexOf('fund') > -1) return false;
+
+  const myDestination = app.participants[app.myIndex].destination;
+  const allocation = checkThat(app.supported?.outcome, isSimpleAllocation);
+  const currentFunding = app.funding(allocation.assetHolderAddress);
+  const allocationsBeforeMe = _.takeWhile(
+    allocation.allocationItems,
+    a => a.destination !== myDestination
+  );
+  const targetFunding = allocationsBeforeMe.map(a => a.amount).reduce(BN.add, BN.from(0));
```
andrewgordstewart, line 43: Should we make it clear that this assumes these facts?

1. each participant cares about a unique destination
2. each destination appears at most once

One reason to drop (2), for instance, is to support ledger top-ups with as few state updates as possible.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -159,6 +167,24 @@ export class Store {
     return {outgoing, channelResult};
   }
 
+  async addChainServiceRequest(
+    channelId: Bytes32,
+    type: ChainServiceApi,
+    tx: Transaction
+  ): Promise<void> {
+    const channel = await Channel.query(tx)
+      .where('channelId', channelId)
```
andrewgordstewart, line 176: Should this be `Channel.forId(channelId, tx)`?

tomclose:
Looking at this again, I'm confused about why we need the `CreateChannel` objective, as it looks like `OpenChannel` already serves as the signal for creating a channel. Could you elaborate?

andrewgordstewart:
@tomclose there is an incoming document, but `OpenChannel` does not have a signed state on it.

On discussion with @kerzhner, that seems like a bad pattern:
- the objective has to assume that it was sent with the first signed state, else the objective is useless
- if you want to eg. enforce database constraints to ensure your models are valid, you need to combine the objective with the state to construct the channel model
- you have to consider the case that you don't have the channel, when you include a signedState in a message
- the message format thus doesn't allow you to send only the state variables, and instead requires you send the channel constants (which everyone always had, almost all the time)

tomclose:
The wire messages _do_ already include the channel constants though: 
https://github.com/statechannels/statechannels/blob/bf972bde98abf4bd70dcd5ca60f951034b6ead82/packages/wire-format/src/types.ts#L64-L76
Are you saying this change is partly so that we can implement an optimization where we remove the channel constants from the wire messages?

Part of the intention with the messaging format is that the objective and states could arrive separately, and it should all just work. The reason for that was to make the syncing/bootstrapping when you restart as easy as it could be. The case where you receive a state that you don't have the channel for is probably one we need to be able to handle.

packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
kerzhner: Any specific ones you are thinking about? Maybe restrict the set of strings that can be stored?

packages/server-wallet/src/protocols/application.ts
kerzhner: 👍 Is your thought to alter the logic to potentially drop these assumptions? Or to leave a comment stating the assumptions?

andrewgordstewart:
I don't know if I follow. If the state is with the objective, then it is not possible for them to arrive separately, so by sending them separately, you seem to have created an extra scenario that you have to

packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
andrewgordstewart: In the future, wallets may want to upgrade in a rolling fashion, so they can still serve requests during the upgrade. At that point, we can explore more advanced migration strategies (they are not that hard, they just require care).

packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
andrewgordstewart: Yeah.

I personally prefer DB constraints, though it has the downside of "now, your validation can live in two places".

packages/server-wallet/src/protocols/application.ts
andrewgordstewart: I think a comment would do for now, since partial top ups are not soon on the roadmap

tomclose:
The case I'm thinking about is one where you want to bootstrap your channels without going all the way back to the `createChannel` objective. Or an example when you're in a multi-party channel and the update from B arrives before the `createChannel` objective from A. In those cases, you'd want to be able to accept states that you don't already have a channel for, without needing a `createChannel` objective. Therefore it has to be ok for the `createChannel` objective to arrive after some states for the channel. 

It might be ok to drop objectives that you don't have a channel for, provided that the wallet always processes the `signedStates` from a message before the `objectives`. A 'correct' wallet will always provide the first state in the same packet as the objective.

andrewgordstewart:
Mike's proposal will hopefully clear this up :)

PR #2562, opened 2020-09-16
[nitro-protocol, wallet-core] Consolidate Destination and Uint* types

geoknee:
The overall aim here is to thin out `wallet-core` as much as possible, moving code into `nitro-protocol` where necessary, with a view to removing `wallet-core` completely if possible.

Here I have picked a number of types that are duplicated, i.e. exist in both packages. It happens that the types are incompatible. The ones in `wallet-core` are *richer* in the sense of being [branded ](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d) types.  I decided to preserve the richer ones (with some tweaks) and make any necessary changes to get things compiling. 

This has meant doing a lot of `as Destination` and `as Uint256` throughout several packages. Along the way, I had to resolve conflicts with types exported by `wire-format`. I chose to revert to the types exported by `nitro-protocol`.


packages/nitro-protocol/src/contract/types.ts
```diff
@@ -1,4 +1,8 @@
+//https: medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d
+type Brand<T, B> = T & {_brand: B};
```
geoknee, line 2: Establishes a convention for pulling these kinds of tricks.

packages/nitro-protocol/src/contract/types.ts
```diff
@@ -13,7 +17,10 @@ export type Uint32 = number;
...
-export type Uint64 = string;
-export type Uint128 = string;
-export type Uint256 = string;
+// const a: Uint256 = '0xa' as Uint256;
+// const b: Uint256 = '0xb' as Uint256;
+// const s: Uint256 = a + b; // Type error: Type 'string' is not assignable to type 'Uint256'
+export type Uint56 = Brand<string, 'Uint56'>;
+export type Uint64 = Brand<string, 'Uint64'>;
+export type Uint128 = Brand<string, 'Uint128'>;
+export type Uint256 = Brand<string, 'Uint256'>;
```
geoknee, line 26: I think we ought to put something about this in the readme and maybe docs, since these types are _exported_ and therefore form part of the TS API. Consumers may need to know to do ` as Destination`. Do we want to expose this (fairly exotic) set of types?

packages/wallet-core/src/types.ts
```diff
@@ -1,10 +1,5 @@
 import {FundingStrategy} from '@statechannels/client-api-schema';
-
-export type Uint256 = string & {_isUint256: void};
```
geoknee, line 3: Deleted. Yay.

packages/wallet-core/src/types.ts
```diff
@@ -35,7 +30,7 @@ export interface StateVariables {
   isFinal: boolean;
 }
 export type StateVariablesWithHash = StateVariables & Hashed;
-export type Destination = string & {_isDestination: void};
```
geoknee, line 38: Deleted. Yay.

packages/nitro-protocol/readme.md
```diff
@@ -65,7 +65,7 @@ const {
 import {Channel} from '@statechannels/nitro-protocol';
 
 const channel: Channel = {
-  chainId: '0x1',
+  chainId: '0x1' as Uint256,
```
geoknee, line 68: Perhaps we should have a `toUint256()` function that pads with zeros and does the type assertion.

andrewgordstewart: Where does the requirement on the length of the string come from?

Can our `BN.from` be the `toUint256` you mention?

geoknee: I wasn't sure about the length of the string. On the one hand, it doesn't seem necessary and ethers will probably take care of the final encoding before anything gets submitted to the chain. On the other hand, we do validate the format of RPC requests and wallet-wallet messages, and we do have types such as `Uint48` and `Uint56`. So it seems like an opportunity to establish a convention about the format of the strings that exist at runtime, and one that reflects the solidity types.

This is easily changed though, so I'm interested to know what you think the pros/cons might be. 

> Can our BN.from be the toUint256 you mention?

Yes probably, although I am doing these changes step by step and we might pass through an intermediate stage where we use `ethers.BigNumber.from` before I think about pushing our `BN` down into `nitro-protocol`.

andrewgordstewart: > I wasn't sure about the length of the string. On the one hand, it doesn't seem necessary and ethers will probably take care of the final encoding before anything gets submitted to the chain. On the other hand, we do validate the format of RPC requests and wallet-wallet messages, and we do have types such as `Uint48` and `Uint56`. So it seems like an opportunity to establish a convention about the format of the strings that exist at runtime, and one that reflects the solidity types.
> 
> This is easily changed though, so I'm interested to know what you think the pros/cons might be.

The only important thing, from this point of view, is that it is easily changed.

It seems to me like there are two requirements:
1. I am able to encode our representation of a state and arrive at the same encoding as you do (and obviously, that gets decoded to the correct values in solidity).
2. I am able to compare two values of properties

One solution to both is to use a canonical representation of each value (eg. fixed-length, lower-cased hex string representation). 

Another solution is to use a careful set of unary and binary operators that don't care as much about the particular representation of the values, such as `BN.eq`, `BN.isZero`, etc.

> > Can our BN.from be the toUint256 you mention?
> 
> Yes probably, although I am doing these changes step by step and we might pass through an intermediate stage where we use `ethers.BigNumber.from` before I think about pushing our `BN` down into `nitro-protocol`.

Makes sense.

Issue #2565, opened 2020-09-17
`server-wallet-profiling` step is broken.


andrewgordstewart:
https://app.circleci.com/pipelines/github/statechannels/statechannels/8937/workflows/808fecce-ef8e-4a82-ad10-334fa1c1e767/jobs/42506/parallel-runs/0/steps/0-105

This logs `TypeError: builder.knex(...).queryBuilder is not a function`. Likely related to https://github.com/statechannels/statechannels/pull/2552.

It would be good for `server-wallet-profiling` to return exit code `1` when the child process prints an error to `STDERR`.

andrewgordstewart:
It now appears to be totally broken, which I think is because https://github.com/statechannels/statechannels/pull/2580 introduced a `knex` instance (and thus a database pool) that does not get destroyed at the end of the profile script.

PR #2566, opened 2020-09-17
[server-wallet] Fail profile job on error

geoknee:
- [x] ~Expose problem with a nonzero exit code~
~Find a fix - suspect something to do with using `recordFunctionMetrics()` (I get a different error if I don't use that)  metrics~
 ~Apply fix~

#2565 Has been now been solved, but this PR:

-  inserts a change which makes the test properly fail if there is an error (which could save us in the future). 
- changes the `PayerClient` to accept configuration in it's constructor
- a little bit of refactoring here and there

### DIFF
--- a/packages/server-wallet/e2e-test/e2e-utils.ts
+++ b/packages/server-wallet/e2e-test/e2e-utils.ts
@@ -41,7 +41,7 @@ export const triggerPayments = async (
   );
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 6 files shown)


geoknee:
I think maybe there still is a residual error, although the profiling job seems to successfully complete with expected output.

geoknee:
There's no residual error, just output to `stderr`, which was causing a `process.exit(1)` (but no longer).

snario (approved):

PR #2568, opened 2020-09-18
Ags/worker threads

andrewgordstewart:
By putting the entire call to updateChannel in one of your workers, I can get a 2x speedup on my 4 core machine.
master
```
monorepo/packages/server-wallet on  master [$!] is :package: v0.3.6 via ⬢ v12.16.3 on :cloud:  us-west-1 took 33s 
❯ ts-node benchmarks/update-channel.benchmark.ts
[@statechannels/devtools] NODE_ENV is undefined — setting to "development"
serial x 100: 1111.306ms
concurrent x 100: 690.045ms
^C
```

concurrent updateChannel
```
monorepo/packages/server-wallet on  ags/worker-threads [$] is :package: v0.3.6 via ⬢ v12.16.3 on :cloud:  us-west-1 
❯ ts-node benchmarks/update-channel.benchmark.ts
[@statechannels/devtools] NODE_ENV is undefined — setting to "development"
serial x 100: 1306.597ms
concurrent x 100: 370.224ms
```

### DIFF
--- a/packages/server-wallet/src/utilities/workers/loader.js
+++ b/packages/server-wallet/src/utilities/workers/loader.js
@@ -1,4 +1,4 @@
 const path = require('path');
 
-require('ts-node').register();
+require('ts-node').register({typeCheck: false});
 require(path.resolve(__dirname, './worker.ts'));
--- a/packages/server-wallet/src/utilities/workers/manager.ts
+++ b/packages/server-wallet/src/utilities/workers/manager.ts
@@ -3,6 +3,7 @@ import path from 'path';
 
 import {Pool} from 'tarn';
 import {State, StateWithHash} from '@statechannels/wallet-core';
+import {UpdateChannelParams} from '@statechannels/client-api-schema';
 
 import {StateChannelWorkerData} from './worker-message';
 
@@ -10,7 +11,7 @@ export class WorkerManager {
   pool = new Pool({
     create: (): Worker => new Worker(path.resolve(__dirname, './loader.js')),
     destroy: (worker: Worker): Promise<number> => worker.terminate(),
-    min: 1,
+    min: 5,
     max: 5,
     reapIntervalMillis: 2147483647,
   });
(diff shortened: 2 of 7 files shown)


andrewgordstewart:
This strategy has a few pitfalls:
1. the main thread's wallet might be configured differently than the default config. Thus, the main thread should send its config to the worker threads when they spin up their wallet.
2. the worker thread might now throw a `NotYourTurn` error, and the main thread won't learn about it. Instead, the `manager.updateChannel` promise won't resolve.

packages/server-wallet/src/utilities/workers/loader.js
```diff
@@ -1,4 +1,4 @@
 const path = require('path');
 
-require('ts-node').register();
+require('ts-node').register({typeCheck: false});
```
snario, line 3: For speed here we may want to use something like this https://github.com/alangpierce/sucrase

packages/server-wallet/src/utilities/workers/manager.ts
```diff
@@ -3,14 +3,15 @@ import path from 'path';
...
 
 import {StateChannelWorkerData} from './worker-message';
 
 export class WorkerManager {
   pool = new Pool({
     create: (): Worker => new Worker(path.resolve(__dirname, './loader.js')),
     destroy: (worker: Worker): Promise<number> => worker.terminate(),
-    min: 1,
+    min: 5,
     max: 5,
```
snario, line 15: Maybe these should be in the config?

packages/server-wallet/src/utilities/workers/loader.js
andrewgordstewart: FYI this was introduced not for speed, but because `ts-node` was complaining about a type error that it doesn't flag when compiling normally. (Possibly it is not using the project's tsconfig?)


I am reluctant to have the published library require any kind of transpilation at runtime in any case. It seems like it should be all JavaScript.

andrewgordstewart:
BTW this is not a PR meant to be reviewed. It just served as a reference for an alternative worker-thread strategy, and a placeholder for discussions, such as
> the worker thread might now throw a NotYourTurn error, and the main thread won't learn about it. Instead, the manager.updateChannel promise won't resolve.

PR #2572, opened 2020-09-20
Rename wallet-core Message -> Payload

tomclose:
Before this PR we had `Message` defined in multiple places
```ts
// in wire-protocol
export interface Message {
  recipient: string; // Identifier of user that the message should be relayed to
  sender: string; // Identifier of user that the message is from
  data: {
    signedStates?: SignedState[];
    objectives?: Objective[];
  };
}

// in client-api-schema
export interface Message {
    data: unknown;
    recipient: string;
    sender: string;
}

// in wallet-core
export interface Message {
  signedStates?: SignedState[];
  objectives?: Objective[];
}
```
The wallet core type was totally incompatible with the other two :/. 

This PR renames the wallet-core `Message` to `Payload` - consistent with being the type of `Message['data']` from wire-protocol.

### DIFF
--- a/packages/server-wallet/e2e-test/payer/client.ts
+++ b/packages/server-wallet/e2e-test/payer/client.ts
@@ -2,7 +2,7 @@ import axios from 'axios';
 import {ChannelResult, Participant} from '@statechannels/client-api-schema';
 import {Wallet, constants} from 'ethers';
 const {AddressZero} = constants;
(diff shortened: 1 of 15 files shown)


andrewgordstewart (approved):
I skimmed the changes and it looks like it's simply a (well needed) rename.

PR #2573, opened 2020-09-20
Server wallet uses wire format

tomclose:
This PR contains a bunch of fixes required to get the server-wallet to communicate using the wire-format:

* Updates wire-format to include the new `requests` property
* Updates wallet-core serde to not drop the `requests` property
* Changes the type of the message handed to `pushMessage` to `unknown`, and then validates the correct format immediately
* Ensures that the messages being sent from the wallet are the right format

The change to `unknown` is really important, as it will prevent users of the wallet from having to cast messages to the right type, as is happening all over the `payment-manager` and `receipt-manager`.

There are potential performance implementations, because converting from the wire-message format and back can involve hashing/signature recovery. In theory, this isn't necessary, as we have the properties already (or need to calculate them anyway), but that would require more changes than I wanted to make in one PR. In any case, we should probably merge, profile and then re-optimize if it turns out to be a problem.

### DIFF
--- a/packages/server-wallet/readme.md
+++ b/packages/server-wallet/readme.md
@@ -29,6 +29,7 @@ SERVER_DB_PORT=5432
 For any environment variables specific to local setup, such as postgres host or port, do not modify `.env` files checked into the repository. Instead, add the variables to `.env.test.local` (or to other local `.env` files). Specifically, you might want to override `SERVER_DB_USER`, `SERVER_DB_HOST` and `SERVER_DB_PORT`.
 
 ```
+echo "SERVER_DB_USER=postgres" > .env.test.local # you probably need to do at least this
 yarn install
 NODE_ENV=test yarn db:create
 NODE_ENV=test yarn db:migrate
--- a/packages/server-wallet/readme.md
+++ b/packages/server-wallet/readme.md
@@ -29,7 +29,7 @@ SERVER_DB_PORT=5432
 For any environment variables specific to local setup, such as postgres host or port, do not modify `.env` files checked into the repository. Instead, add the variables to `.env.test.local` (or to other local `.env` files). Specifically, you might want to override `SERVER_DB_USER`, `SERVER_DB_HOST` and `SERVER_DB_PORT`.
 
 ```
-echo "SERVER_DB_USER=postgres" > .env.test.local # you probably need to do at least this
+echo "SERVER_DB_USER=postgres" > .env.test.local # you probably need to do at least this (if you're on a mac)
 yarn install
 NODE_ENV=test yarn db:create
 NODE_ENV=test yarn db:migrate
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -23,6 +23,7 @@ import {
   BN,
   Zero,
   SignedState,
+  serializeMessage,
 } from '@statechannels/wallet-core';
 import * as Either from 'fp-ts/lib/Either';
 import {ETH_ASSET_HOLDER_ADDRESS} from '@statechannels/wallet-core/lib/src/config';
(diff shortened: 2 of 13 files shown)


packages/server-wallet/readme.md
```diff
@@ -29,6 +29,7 @@ SERVER_DB_PORT=5432
 For any environment variables specific to local setup, such as postgres host or port, do not modify `.env` files checked into the repository. Instead, add the variables to `.env.test.local` (or to other local `.env` files). Specifically, you might want to override `SERVER_DB_USER`, `SERVER_DB_HOST` and `SERVER_DB_PORT`.
 
 ```
+echo "SERVER_DB_USER=postgres" > .env.test.local # you probably need to do at least this
```
snario, line 32: Might be worth mentioning this is a Mac-specific thing, since the default installation instructions for `brew` and `Postgres.app` set up the DB with `postgres` user not `root` on Mac

snario (approved):

andrewgordstewart:
Hmm. This code is already benchmarked by our CI suite, and it seems to slow down the wallet by nearly 50%. This should probably not be published, since we are putting significant effort into wallet performance right now.

Do you have a plan to recover the lost throughput?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -131,14 +132,14 @@ export class Wallet implements WalletInterface {
     return {
       outbox: peers.map(recipient => ({
         method: 'MessageQueued',
-        params: {
-          recipient,
-          sender,
-          data: {
+        params: serializeMessage(
```
tomclose, line 135: This call to `serializeMessage` calls `serializeState`, which calculates the `channelId` for each channel.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -366,11 +369,7 @@ export class Wallet implements WalletInterface {
         peers.map(recipient => {
           outbox.push({
             method: 'MessageQueued',
-            params: {
-              recipient,
-              sender,
-              data: {signedStates},
-            },
+            params: serializeMessage({signedStates}, recipient, sender),
```
tomclose, line 372: Another call to `serializeMessage`

packages/server-wallet/src/wallet/store.ts
```diff
@@ -158,7 +159,7 @@ export class Store {
       type: 'NotifyApp' as 'NotifyApp',
       notice: {
         method: 'MessageQueued' as 'MessageQueued',
-        params: {sender, recipient, data},
+        params: serializeMessage(data, recipient, sender),
```
tomclose, line 162: `serializeMessage` again. This one is in the store, so the `channelId` should be readily available.

packages/wallet-core/src/serde/wire-format/deserialize.ts
```diff
@@ -29,13 +30,20 @@ export function convertToInternalParticipant(participant: {
   return {...participant, destination: makeDestination(participant.destination)};
 }
 
+export function validatePayload(rawPayload: unknown): Payload {
+  // todo: wire-format should export a validator specially for the payload
+  return deserializeMessage(validateMessage({recipient: '', sender: '', data: rawPayload}));
```
tomclose, line 35: This call to `deserializeMessage` calls `deserializeState`, which calls `getAddress` on the signatures.

tomclose: We have to do this later anyway in the wallet, when we check the signature.

tomclose:
The problems in performance come from converting between the wire-message format and the internal format used by the server-wallet. There are two problems:

1. on the way in `deserializeMessage` calls `getAddress` on each signature
2. on the way out `serializeMessage` calls `calculateChannelId` on each state

For 1, the wallet always needs to verify this information, which it does in `validateSignatures` inside `addSignedState`. We could probably safely feed the wire-state through to that point to avoid having to do the conversion before that point.

For 2, the wallet recalculates this information anyway, so option A would be to remove `channelId` from the wire-format. In theory, the wallet doesn't always have to do this though - if it already has the channel stored, it could use the provided `channelId` to  identify the channel and just check the constants match (or just ignore them), and then use the `channelId` to calculate the `stateHash`. Option B would be to pass the channelId through.

Overall, this should probably all happen in the store: all outbox states/messages should be in the wire format, and it should expect all states provided to pushMessage to be in that format.

PR #2575, opened 2020-09-21
[server-wallet] Add E2E test for channel syncing

snario:


packages/server-wallet/e2e-test/payer/client.ts
```diff
@@ -101,21 +101,35 @@ export default class PayerClient {
...
     );
 
     await this.time(`push message ${channelId}`, async () => this.wallet.pushMessage(reply));
   }
 
+  public async syncChannel(channelId: string): Promise<void> {
+    const {
+      outbox: [{params}],
+    } = await this.wallet.syncChannel({channelId});
+    const reply = await this.messageReceiverAndExpectReply((params as WireMessage).data as Payload);
```
andrewgordstewart, line 121: When I began reviewing this PR yesterday, I got to this line, became puzzled, and stopped.

[This](https://github.com/statechannels/statechannels/pull/2573) was merged yesterday. Could it be used to clarify what this line does?

snario: I didn't change any behaviour, just copied the pattern from [before](https://github.com/statechannels/statechannels/blob/30bc18df58ede924d39fb49e121a26b68b4dedd4/packages/server-wallet/e2e-test/payer/client.ts#L105)

andrewgordstewart: Given that the `server-wallet` now speaks the wire format, I believe that we don't need to (and I assert that we **shouldn't**) copy the pattern from before.

I'm not sure about that though, are you able to clarify?

packages/server-wallet/e2e-test/e2e.test.ts
```diff
@@ -127,6 +127,51 @@ describe('payments', () => {
...
+      await payerClient.createPayment(channelId); // Forget to send
+
+      await expectSupportedState(ChannelPayer, 4);
+      await expectSupportedState(ChannelReceiver, 3);
+
+      await payerClient.syncChannel(channelId);
+
+      await expectSupportedState(ChannelPayer, 5);
+      await expectSupportedState(ChannelReceiver, 5);
+    });
```
andrewgordstewart, line 172: I like these tests. They're very easy to read, and understand what they're trying to test.

packages/server-wallet/e2e-test/payer/client.ts
andrewgordstewart: I think, given that the `PaymentManager` is being rewritten, it is safe to ignore my worry here.

packages/server-wallet/e2e-test/receiver/controller.ts
```diff
@@ -25,18 +24,23 @@ export default class ReceiverController {
...
   public async acceptMessageAndReturnReplies(message: Payload): Promise<Payload> {
-    const {signedStates, objectives} = message;
+    const reply: Payload = {
+      signedStates: [],
+      objectives: [],
+    };
 
     const {
       channelResults: [channelResult],
+      outbox: [maybeSyncStateResponse],
```
andrewgordstewart, line 34: I'm not so sure about the naming here: the receiver controller shouldn't care the contents of the message -- it should just send the message.

That's what this code looks to do, but the name might be a bit deceptive. It also obscures the fact that there may be _multiple_ outgoing messages, and they should all be send to the relevant peer.

packages/server-wallet/e2e-test/receiver/controller.ts
```diff
@@ -25,18 +24,23 @@ export default class ReceiverController {
...
-        objectives: [],
-      };
-    } else {
+    if (maybeSyncStateResponse) {
+      const syncResponse = (maybeSyncStateResponse.params as WireMessage).data as Payload;
+      reply.signedStates = reply.signedStates?.concat(syncResponse.signedStates || []);
+      reply.objectives = reply.objectives?.concat(syncResponse.objectives || []);
+    }
+
+    if (channelResult && channelResult.turnNum % 2 === 0) {
```
andrewgordstewart, line 43: I don't follow the need for this condition.

andrewgordstewart (approved):
This generally looks ok, but I did have a couple of concerns. 

It's pretty awesome that the syncing protocol now has a test!

packages/server-wallet/e2e-test/receiver/controller.ts
snario: If the channel result had an odd turn number, then it would mean that it is _not_ the turn of the receiver, which would mean that the receiver should not try to update the channel (it would see a "not your turn" error). The case where it might be an odd turn number is introduced in this PR and would be the result of a channel syncing done by the payer where the payer tries to sync after having ignored the previous reply from the receiver (i.e., the 'can update pre-existing channel, send state, but ignore reply, and sync later' case).

packages/server-wallet/e2e-test/receiver/controller.ts
snario: Right now the receiver client is built for the end to end test and the only case where there would be anything in the outbox after a `pushMessage` call in the tests is the case where the payer sends a request for channel data (i.e., syncing). If we decided to add more features to the receiver then we would have to remove some of the hard coded elements of it (like this example where it's hard coded to expect a sync response in index 0 of the outbox, or nothing at all). However, then we start getting into the territory of writing a receipt manager. I guess a question would be how generalized should this test harness be?

PR #2577, opened 2020-09-21
Return false and log warning if bytecode missing on transition validation

snario:
Seems like an oversight after we added the flag to `ServerWalletConfig`

### DIFF
--- a/packages/server-wallet/src/evm-validator.ts
+++ b/packages/server-wallet/src/evm-validator.ts
@@ -27,16 +27,16 @@ export const validateTransitionWithEVM = async (
     });
     return false;
   }
+
   const bytecode =
     bytecodeCache[from.appDefinition] ??
     (bytecodeCache[from.appDefinition] =
       (await AppBytecode.getBytecode(defaultConfig.chainNetworkID, from.appDefinition, tx)) ||
       MISSING);
+
   if (bytecode === MISSING) {
-    // logger.warn(
-    //   `No bytecode found for appDefinition ${from.appDefinition} and chain id ${config.chainNetworkID}. Skipping valid transition check`
-    // );
-    return true;
+    logger.warn(`No bytecode found for appDefinition ${from.appDefinition}.`);
+    return false;
   }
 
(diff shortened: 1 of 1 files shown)


snario:
Seems like right now it is defaulted to _not_ skip EVM validation, which is why this errors

packages/server-wallet/src/evm-validator.ts
```diff
@@ -27,16 +27,16 @@ export const validateTransitionWithEVM = async (
...
     (bytecodeCache[from.appDefinition] =
       (await AppBytecode.getBytecode(defaultConfig.chainNetworkID, from.appDefinition, tx)) ||
       MISSING);
+
   if (bytecode === MISSING) {
-    // logger.warn(
-    //   `No bytecode found for appDefinition ${from.appDefinition} and chain id ${config.chainNetworkID}. Skipping valid transition check`
-    // );
-    return true;
+    logger.warn(`No bytecode found for appDefinition ${from.appDefinition}.`);
```
andrewgordstewart, line 38: This means a warning gets logged in every state update. That should not be the behaviour here -- it should get logged at most once.

snario: The state update would fail, too. The interface for this function is expecting a `boolean` response though, so we should log the reason why it returned `false` somewhere I think

andrewgordstewart: Oh I see. You're suggesting, then, that the _behaviour_ of the validator should change so that it fails when the bytecode is not known?

In that case, I'd probably expect it to throw an error, instead of returning `false`. Returning `false`, to me, should indicate that it is _known_ to be an invalid transition.

andrewgordstewart: It seems like either behaviour is ok, but that the package is not ready for this changeset (eg. since the tests are failing!)

snario: I'll leave this PR open for now as we might be in a position to make this change soon... and certainly will have to for mainnet :)

andrewgordstewart:
Closing as this is stale, and is easy to reproduce with the correct behaviour.

PR #2578, opened 2020-09-21
[server-wallet] Remove pre-database query data validation

snario:
`update` validates and updates every column of the table for a particular query which seems unnecessary (@andrewgordstewart ?). `patch` should be faster since it is a simpler query.

Tradeoff:
- Faster queries (I think?)
- Less validation

### DIFF
--- a/packages/server-wallet/src/models/funding.ts
+++ b/packages/server-wallet/src/models/funding.ts
@@ -50,7 +50,7 @@ export class Funding extends Model implements RequiredColumns {
       return await Funding.query(knex).insert({channelId, amount, assetHolder});
     } else {
       return await Funding.query(knex)
-        .update({channelId, amount, assetHolder})
+        .patch({channelId, amount, assetHolder})
         .where({channelId, assetHolder})
         .returning('*')
         .first();
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -175,14 +175,9 @@ export class Store {
   ): Promise<void> {
     const channel = await Channel.forId(channelId, tx);
 
-    const cols: RequiredColumns = {
-      ...channel,
-      chainServiceRequests: [...channel.chainServiceRequests, type],
-    };
-
     await Channel.query(tx)
       .where({channelId: channel.channelId})
-      .update(cols);
+      .update({chainServiceRequests: [...channel.chainServiceRequests, type]});
   }
 
   async getChannel(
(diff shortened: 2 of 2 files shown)


andrewgordstewart:
Observation:
- the stress test on this branch took `70.828` seconds to make 2500 payments
- The most recent run on master took `95.0241` seconds to make 2500 payments

Note that there is significant variation in the stress test performance (that I don't know how to explain). However, this is a large enough difference to suggest that this is a real optimization.

> Less validation

It's not clear whether this tradeoff is worth it, since we need to be really careful about _not_ validating database input. Database entries are hard to change, and allowing bad data in could be a disaster.

That said, most of our validations (these in particular) should not have a large performance overhead, and it is becoming more clear that `knex` + `Objection` does in fact have a non-trivial overhead that we are paying dearly for. Thus, we will probably need to find a reliable way of validating our data without using `Objection` anyway.

andrewgordstewart:
> Tradeoff:
> 
> * Faster queries (I think?)
> * Less validation

- ✂️ ✂️  less code

snario:
>  it is becoming more clear that `knex` + `Objection` does in fact have a non-trivial overhead that we are paying dearly for.

I think in this case it's not that `knex` + `Objection` are adding overhead; we chose this overhead (perhaps unknowingly) by using `required` on the `JSONSchema` object and the `.update` method which does validation.

andrewgordstewart:
> > it is becoming more clear that `knex` + `Objection` does in fact have a non-trivial overhead that we are paying dearly for.
> 
> I think in this case it's not that `knex` + `Objection` are adding overhead; we chose this overhead (perhaps unknowingly) by using `required` on the `JSONSchema` object and the `.update` method which does validation.

That is overhead though? Should it take a real amount of CPU cycles to check that a javascript object has a few properties?

In any case, it may have simply been the variance: the [latest run](https://app.circleci.com/pipelines/github/statechannels/statechannels/9035/workflows/f9456116-65c4-4fa3-996f-e68878a38733/jobs/43121) on master ran in 80s. Without re-running the stress test on this branch, I'd be hesitant to say for certain that this is an optimization. (It is _likely_ an optimization.)

andrewgordstewart:
I seem to have lost the context for this change.

The PR changes the code from
- validating data
to
- not validating data

This kind of change should have some justification. I thought I saw some justification when I clicked on some failing circle run, but since you have force-pushed, it is hard to find :/

andrewgordstewart:
Oh, circle's got my back:
- you aimed to just update _selected_ columns, not all _required_ columns
- if you do that, you get this error
```
ValidationError: appDefinition: is a required property, channelNonce: is a required property, challengeDuration: is a required property, ...
```
(from [this run](https://app.circleci.com/pipelines/github/statechannels/statechannels/9012/workflows/a83461cc-2b71-41f2-aa62-625d82189cfa/jobs/42987))

andrewgordstewart (approved):
> patch ignores the required properties and only validates the ones that are found.

Ok, I misunderstood what patch did. 

> it is a simpler query

this lead me to believe that `patch` does not perform any validation. But, the difference between `patch` and `update` is not that: it's that `update` should only be used when you want to update all columns, and therefore it validates it _as if it were a new row_ (and checks that all `required` columns are there).

> Use update if you update the whole row with all its columns. Otherwise, using the patch method is recommended.

Quotes are taken from [here](https://vincit.github.io/objection.js/api/query-builder/mutate-methods.html#update).

PR #2580, opened 2020-09-21
Knex should be implementation detail of Store

snario:
I think we agree that Knex is an implementation detail of a PostgreSQL specific implementation of a Store, thus it should not be known by a Wallet, especially if we intend on re-using that code in the browser.

### DIFF
--- a/packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
@@ -80,6 +80,4 @@ describe('happy path', () => {
 });
 
 it("doesn't create a channel if it doesn't have a signing wallet", () =>
-  expect(w.createChannel(createChannelArgs())).rejects.toThrow(
-    'null value in column "signing_address"'
-  ));
+  expect(w.createChannel(createChannelArgs())).rejects.toThrow('Not in channel'));
--- a/packages/server-wallet/src/wallet/__test__/lock-app.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/lock-app.test.ts
@@ -14,7 +14,11 @@ import {stateVars} from './fixtures/state-vars';
 
 jest.setTimeout(10_000);
 
-const store = new Store(knex, defaultConfig.timingMetrics, defaultConfig.skipEvmValidation);
+let store: Store;
+
+beforeAll(async () => {
+  store = new Store(knex, defaultConfig.timingMetrics, defaultConfig.skipEvmValidation);
+});
 
 it('works', async () => {
   await seedAlicesSigningWallet(knex);
--- a/packages/server-wallet/src/wallet/__test__/lock-app.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/lock-app.test.ts
@@ -20,7 +20,7 @@ const store = new Store(
   defaultConfig.skipEvmValidation
 );
 
-afterAll(async () => store.knex.destroy());
+afterAll(async () => store.closeDatabaseConnection());
 
 it('works', async () => {
   await seedAlicesSigningWallet(knex);
--- a/packages/server-wallet/src/wallet/__test__/lock-app.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/lock-app.test.ts
@@ -27,7 +27,7 @@ it('works', async () => {
 
   const {channelId, latest} = c;
   await expect(
-    store.lockApp(knex, channelId, async tx =>
+    store.lockApp(channelId, async tx =>
       store.signState(channelId, {...latest, turnNum: latest.turnNum + 1}, tx)
     )
   ).resolves.toMatchObject({channelResult: {turnNum: 6}});
@@ -76,7 +76,7 @@ describe('concurrency', () => {
     await Promise.all(
       _.range(numAttempts).map(() =>
         store
-          .lockApp(knex, channelId, async tx => store.signState(channelId, next(c.latest), tx))
+          .lockApp(channelId, async tx => store.signState(channelId, next(c.latest), tx))
           .then(countResolvedPromise)
           .catch(countRejectedPromise)
           .finally(countSettledPromise)
@@ -117,9 +117,7 @@ describe('concurrency', () => {
       await Promise.all(
         channelIds.map(channelId =>
           store
-            .lockApp(knex, channelId, async (tx, c) =>
-              store.signState(channelId, next(c.latest), tx)
-            )
+            .lockApp(channelId, async (tx, c) => store.signState(channelId, next(c.latest), tx))
             .then(countResolvedPromise)
             .finally(countSettledPromise)
         )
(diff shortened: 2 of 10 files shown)


andrewgordstewart:
I think this should go even further: it should have models that are bound to its knex.

snario:
Something introduced an unhandled promise in the server wallet tests... not sure what.

andrewgordstewart:
> Something introduced an unhandled promise in the server wallet tests... not sure what.

Each time you construct a store, you construct a knex that must be destroyed for jest to exit.

You could require a knex instance by the store's constructor.

snario:
> I think this should go even further: the store should have models that are bound to its knex.

Ah I see! I think the store should probably create its own connection

andrewgordstewart:
> > I think this should go even further: the store should have models that are bound to its knex.
> 
> Ah I see! I think the store should probably create its own connection

I sort of agree, but then every store that is constructed during a test suite needs to be destroyed. I don't particularly like that tradeoff.

snario:
> I sort of agree, but then every store that is constructed during a test suite needs to be destroyed. I don't particularly like that tradeoff.

Since a store is basically a wrapper around a database connection, I don't think it's that unreasonable really. And presently there are only [five occurrences](https://github.com/statechannels/statechannels/pull/2580/commits/d6bdde2c9517cca54a5bbbf0efd7887ad2a4de92), which could be bundled up in a setup and teardown script if we desired to simplify it, or just a single jest test.

snario:
**ci/circleci: server-wallet-profiling** failure related to https://github.com/statechannels/statechannels/pull/2566

packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
```diff
@@ -80,6 +80,4 @@ describe('happy path', () => {
 });
 
 it("doesn't create a channel if it doesn't have a signing wallet", () =>
-  expect(w.createChannel(createChannelArgs())).rejects.toThrow(
-    'null value in column "signing_address"'
-  ));
+  expect(w.createChannel(createChannelArgs())).rejects.toThrow('Not in channel'));
```
geoknee, line 83: Why this change?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -162,95 +164,66 @@ export class Wallet implements WalletInterface {
...
-              objectives: [
-                {
-                  participants,
-                  type: 'CreateChannel',
-                  data: {
-                    signedState: params.data.signedStates[0],
-                    fundingStrategy: 'Direct',
-                  },
+      return {
+        method: 'MessageQueued' as const,
```
geoknee, line 207: I see `as const` sprinkled around a fair bit. Do you know what is it doing for us?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -299,6 +314,40 @@ export class Store {
...
+    channelId: string,
+    fromAmount: Uint256,
+    assetHolderAddress: Address
+  ): Promise<void> {
+    await Funding.updateFunding(this.knex, channelId, fromAmount, assetHolderAddress);
+  }
+
+  async nextNonce(signingAddresses: Address[]): Promise<number> {
+    return await Nonce.next(this.knex, signingAddresses);
+  }
```
geoknee, line 350: Why are these additions in scope for this PR?

packages/server-wallet/src/wallet/__test__/lock-app.test.ts
```diff
@@ -8,13 +8,19 @@ import {Store} from '../store';
...
 jest.setTimeout(10_000);
 
-const store = new Store(defaultConfig.timingMetrics, defaultConfig.skipEvmValidation);
+const store = new Store(
+  extractDBConfigFromServerWalletConfig(defaultConfig),
+  defaultConfig.timingMetrics,
+  defaultConfig.skipEvmValidation
+);
+
+afterAll(async () => store.knex.destroy());
```
geoknee, line 23: Suggestion: the `store` could have a `.destroy()` method that destroys its `knex`? The pattern being, a `wallet` destroys its direct descendants (`store`) and a `store` destorys its direct descendants (`knex`).

geoknee (approved):

packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
snario: The create channel method throws a `StoreError` now instead of raising a Knex error

packages/server-wallet/src/wallet/index.ts
snario: No idea 🤷 I think @kerzhner has been adding these

packages/server-wallet/src/wallet/store.ts
snario: The wallet was in control of database models, which relied on a database connection, which is now contained within the store, so now the database models are also inside the store.

packages/server-wallet/src/wallet/__test__/lock-app.test.ts
snario: SGTM, will update.

snario: 520a43d68

packages/server-wallet/src/wallet/store.ts
snario: e.g., 
image

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: It narrows the type from `string` to a specific string type

packages/server-wallet/src/wallet/store.ts
```diff
@@ -94,6 +94,10 @@ export class Store {
     }
   }
 
+  async closeDatabaseConnection(): Promise<void> {
```
andrewgordstewart, line 97: This could probably just be named `destroy`:
- there may be multiple things that the store needs to clean up
- in a world where we have multiple implementations of a `Store` interface, we might not have a `Database`

Also, (extremely small) nit:
- database connections are ephemeral. They are created on demand, and the number can fluctuate depending on the `knex` config as well as both the wallet load (when not under stress, connections are closed after a timeout) and the DB load (the DB may forcibly close a connection for a variety of reasons)
- knex (well, `tarn`, under the hood) maintains a dynamic pool of connections
- `knex.destroy()` destroys the _pool_ of connections, and signals to `knex` not to start any new ones.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -62,7 +65,19 @@ const throwMissingChannel: MissingAppHandler<any> = (channelId: string) => {
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export class Store {
-  constructor(readonly timingMetrics: boolean, readonly skipEvmValidation: boolean) {
+  knex: Knex;
+
+  constructor(
+    readonly knexOrConfig: Knex | Knex.Config,
```
andrewgordstewart, line 71: I really think this should _insist_ on accepting a `Knex`.

We are now living in a world with a new "gotcha":
1. Say I create a `new Store` with a config object instead of a `Knex` instance.
2. Then, when I run my test, I will notice that jest doesn't exit.
3. I have to then learn that `Store` constructed a connection pool, and that pool may contain open connections, and that prevents the jest process from exiting.

There is a significant cost to (3).
A. Multiple people on this team have had to learn this the hard way, so this issue has already cost at least a full developer day.
B. When they do learn it, they need to maintain some additional overhead: "should I pass in a config? do I need to destroy the pool?"

I fail to see any benefit in the short-term of doing (1). Any benefit should be clearly stated, and (IMO) planned into our work towards consolidating the server and browser wallets.

packages/server-wallet/src/wallet/store.ts
snario: https://github.com/statechannels/statechannels/pull/2602

packages/server-wallet/src/wallet/store.ts
snario: https://github.com/statechannels/statechannels/pull/2602

PR #2582, opened 2020-09-21
WIP - re-optimize wallet

tomclose:
Rough plan:

Tackle `deserializeMessage` first (half done). Do this by pushing the `WirePayload` right down into the wallet store, and rewrite `validateSignatures` to actually return the addresses, instead of checking them against what's there.

Blocked in this by the fact that `pushMessage` isn't the only caller of `addSignedState` - `signState` calls it too, although it definitely shouldn't, as it involves recalculating the `channelId`, that we already know, and validating the signature that we've just created 😳 . Rough plan:

* Factor out the bit that actually adds the state to the store
* Rewrite signState to (1) check the state transition, (2) sign the state, (3) add to the store, using function from point above
* Rewrite `addSignedState` to take a `WireState`

Then remove the `serializeMessage` calls identified [here](https://github.com/statechannels/statechannels/pull/2573). Rough plan:

* Change the `Outbox` format to only ever include wire-messages
* This should remove the need to call `serializeMessage` at the last moment.
* Will also need to address the `CreateChannel` objective, which is constructed in a slightly hacky way from the outbox, and includes a signed state.

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -354,7 +354,7 @@ export class Wallet implements WalletInterface {
     const knex = this.knex;
     const store = this.store;
 
-    const message = validatePayload(rawPayload);
+    const wirePayload = validatePayload(rawPayload);
 
     // TODO: Move into utility somewhere?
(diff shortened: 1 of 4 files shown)


geoknee:
In favour of #2590

PR #2586, opened 2020-09-21
Perform UpdateChannel and PushMessage using worker threads

lalexgap:
Introduces worker threads to handle processing of expensive `wallet` actions.

A new `manager` class has been introduced that spawns threads and delegates `pushMessage` and `updateChannel` to these threads.

The amount of threads is controlled by the environment variable `process.env.AMOUNT_OF_WORKER_THREADS` (or  `workerThreadAmount` on the config object passed into the wallet constructor). This defaults to `0` which disables threading.

The `loader.js` file is responsible for finding the built `worker.js`  and falls back to using `ts-node` if no built workers can be found.


andrewgordstewart (commented):
Ignore the content of my comments. I had a cat on my lap while reviewing this, and just wanted to bookmark points to later comment on.

Submitting this review so that I don't lose those bookmarks.

andrewgordstewart (approved):
Let's merge this in and address future concerns separately!

Could you update the pr description with an overview of what changes are included, and how it all works?

PR #2588, opened 2020-09-22
[server-wallet] Use lodash instead of custom helper pick function

snario:


packages/server-wallet/src/utilities/helpers.ts
```diff
@@ -1,7 +0,0 @@
-export function pick<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K> {
```
andrewgordstewart, line 1: There's a reason this utility is used, and perhaps @kerzhner should have documented it: `_.pick` is not type safe, and returns `Partial<T>`, not `Pick<T, K>`.

andrewgordstewart (changes requested):
Before approving this, we should get @kerzhner 's input.

packages/server-wallet/src/utilities/helpers.ts
snario: I think it does return `Pick<T, K>`? https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ff2359e74ce1c539097e47dc586d49d348a94587/types/lodash/object/pick.d.ts#L22

kerzhner: It seems that the return type of `_.pick` depends on whether the first argument includes all of the properties in later arguments. For example:
```
  const a = _.pick(
    {
      ...constants,
      vars: [],
      chainServiceRequests: [],
      signingAddress,
      fundingStrategy,
    },
    ...CHANNEL_COLUMNS
  );
```
Type of `a` is:
```
const a: Partial<{
    vars: never[];
    chainServiceRequests: never[];
    signingAddress: string;
    fundingStrategy: FundingStrategy;
    chainId: string;
    participants: Participant[];
    channelNonce: number;
    appDefinition: string;
    challengeDuration: number;
}>
```

kerzhner: So the following code has no typescript errors, but will result in a runtime error:
```
  const cols = _.pick(
    {
      ...constants,
    },
    ...CHANNEL_COLUMNS
  );
  const channel = Channel.fromJson(cols);
```

kerzhner: Whereas this code results in a typescript error:
```
  const cols = pick(
    {
      ...constants,
    },
    ...CHANNEL_COLUMNS
  );
  const channel = Channel.fromJson(cols);
```
The error is:
```
Argument of type '"channelId" | "participants" | "appDefinition" | "vars" | "chainId" | "channelNonce" | "challengeDuration" | "signingAddress" | "chainServiceRequests" | "fundingStrategy"' is not assignable to parameter of type '"participants" | "appDefinition" | "chainId" | "channelNonce" | "challengeDuration"'.
  Type '"channelId"' is not assignable to type '"participants" | "appDefinition" | "chainId" | "channelNonce" | "challengeDuration"'.ts(2345)
```

kerzhner: All of that said, I agree with @snario that it is not ideal to have a custom pick function. It's easy to forget to use it.

andrewgordstewart: For clarity, `_.pick` has an overloaded signature: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ff2359e74ce1c539097e47dc586d49d348a94587/types/lodash/object/pick.d.ts#L30

So Mike identified a case where it could have prevented a runtime error if not for this overloading.

kerzhner:
We have 3 options with this PR:
1. Close it.
1. Merge it and live with less type safety.
1. Come up with an alternative solution to using custom `pick`.

Thoughts?

PR #2590, opened 2020-09-22
[server-wallet] Optimize type conversion

geoknee:
Builds off #2582 .

According to the stress test, `Individual makePayment call`:

Before #2573: [1585](https://app.circleci.com/pipelines/github/statechannels/statechannels/8936/workflows/b0a56521-7e7b-49bf-be8c-8a4a3c2cc604/jobs/42502) ms 

After #2573: [2080](https://app.circleci.com/pipelines/github/statechannels/statechannels/9003/workflows/852632ee-b3a8-416f-99f0-de9bb0708dcb/jobs/42926) ms

On this branch: [1376](https://app.circleci.com/pipelines/github/statechannels/statechannels/9062/workflows/d6234e3f-8e5c-4e02-b6ce-8db79d2398b5/jobs/43312) ms


packages/server-wallet/src/utilities/signatures.ts
```diff
@@ -36,17 +36,6 @@ export function fastRecoverAddress(state: State, signature: string, stateHash: s
...
-
-  const signingAddresses = participants.map(p => p.signingAddress);
-
-  if (signingAddresses.indexOf(recoveredAddress) < 0) {
-    throw new Error(
-      `Recovered address ${recoveredAddress} is not a participant in channel ${calculateChannelId(
-        state
-      )}`
-    );
-  }
```
geoknee, line 49: We were doing this twice.

packages/server-wallet/src/wallet/__test__/add-signed-state.test.ts
```diff
@@ -22,14 +19,32 @@ describe('addSignedState', () => {
...
+          ],
+        },
+      ],
+      participants: [],
+      channelNonce: 1,
+      channelId,
+      chainId: '0x01',
+      challengeDuration: 9001,
+      signatures: [BOB_SIGNATURE],
+    };
```
geoknee, line 44: Notes: 

- this test asserts something a bit different now, since the behaviour has changed
- this mock is a bit ugly.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -133,23 +141,50 @@ export class Store {
   async signState(
     channelId: Bytes32,
     vars: StateVariables,
-    tx: Transaction // Insist on a transaction since assSignedState requires it
```
geoknee, line 136: 😊

packages/server-wallet/src/wallet/store.ts
```diff
@@ -232,14 +300,26 @@ export class Store {
         data: {signedState, fundingStrategy},
       } = objective;
       const channelId = calculateChannelId(signedState);
-      const singedStateWithHash = {...signedState, stateHash: hashState(signedState)};
-      validateSignatures(singedStateWithHash);
+      const stateHash = hashState(signedState);
+      const signedStateWithHash = {...signedState, stateHash};
+      // ensure signatures are
```
geoknee, line 305: ✂️

geoknee: 107e625

andrewgordstewart:
🐎 Nice work!!

packages/server-wallet/src/wallet/store.ts
```diff
@@ -168,6 +203,36 @@ export class Store {
...
+
+    const cols: RequiredColumns = {
+      ...channel,
+      vars: channel.vars,
+    };
+
+    const result = await timer('updating', async () =>
+      Channel.query(tx)
+        .where({channelId: channel.channelId})
+        .update(cols)
```
snario, line 228: Should use `.patch` (see #2578)

packages/server-wallet/src/wallet/store.ts
```diff
@@ -258,28 +262,43 @@ export class Store {
...
+      recoverParticipantSigners(wireSignedState, channelId)
+    );
 
     const channel =
-      maybeChannel || (await timer('get channel', async () => getChannel(signedState, tx)));
+      maybeChannel || (await timer('get channel', async () => getChannel(channelId, tx)));
 
     if (!this.skipEvmValidation && channel.supported) {
       const {supported} = channel;
+      const supportedNitroState = {
```
andrewgordstewart, line 360: Why not use `wireStateToNitroState` here?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -367,7 +367,13 @@ export class Store {
     }
 
     const sswh: SignedStateWithHash = {
-      ...wireSignedState,
+      chainId: wireSignedState.chainId,
```
andrewgordstewart, line 386: 👍 👍  (Down with the spread operator!)

packages/server-wallet/src/evm-validator.ts
```diff
@@ -16,8 +15,8 @@ const bytecodeCache: Record<string, string | undefined> = {};
  * Returns a promise that resolves to true if the validateTransition returns true false otherwise
  */
 export const validateTransitionWithEVM = async (
-  from: State,
-  to: State,
+  from: NitroState,
```
andrewgordstewart, line 18: Perhaps the PR's description could eg. explain the reasoning for this change? When I look through the commit history, I can't discern the reason for this change.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -168,6 +203,36 @@ export class Store {
     return {outgoing, channelResult};
   }
 
+  async addMyState(
+    channel: Channel,
+    signedState: SignedStateWithHash,
+    tx: Transaction
+  ): Promise<Channel> {
+    const timer = timerFactory(this.timingMetrics, `addMyState ${channel.channelId}`);
```
andrewgordstewart, line 211: Should we be extra safe here and ensure that I am the one who actually signed this state? ie. something like
```
if (signedState.signature.signer !== channel.myAddress) throw 'Whoops!'
```

This would prevent a security issue where we accidentally add an untrusted state. (I hope that the name `addMyState` would be sufficient to prevent that, but I can imagine that `addMyState` eventually being renamed to something different.)

andrewgordstewart (approved):
Nice work!

I might have appreciated a more descriptive PR, since there are several seemingly unrelated changes to the codebase.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -236,7 +236,7 @@ export class Store {
       } = objective;
       const channelId = calculateChannelId(signedState);
       const singedStateWithHash = {...signedState, stateHash: hashState(signedState)};
-      validateSignatures(singedStateWithHash);
+      // validateSignatures(singedStateWithHash); // TODO do we need this?
```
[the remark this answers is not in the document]
geoknee: I have put the validation back in inside `addObjective`. I don't see a reason why we _shouldn't_ validate the signatures here when we are elsewhere. There is likely an attack vector on the wallet otherwise, where I can put forged states (with incorrect signatures) in your wallet by bundling them inside an objective.

packages/server-wallet/src/wallet/store.ts
geoknee: Because `supported` is not a `WireState`. But your point still stands for `toNitroState`, which I should have used. 17f06b8

packages/server-wallet/src/evm-validator.ts
geoknee: The reason is that I found it more convenient to hoist the call to `toNitroState` up out of this function, because then I could also leverage `wireStateToNitroState` in the outer context if I wanted to, and get away with fewer `thisTypetoThatType` helpers.

 Also, I have an aspiration for this validation code to live in `nitro-protocol`, and this change would make that move easier to accomplish.

packages/server-wallet/src/wallet/store.ts
geoknee: 36134b2

packages/server-wallet/src/wallet/store.ts
geoknee: Sounds very sensible. It's not entirely foolproof, as we could accidentally add a state that has `{signature: bogus, signer: me}`. But that seems much more unlikely. 

6b5d133

geoknee:
Performance of post-review commits is a bit puzzling:

17f06b8 [2441ms](https://app.circleci.com/pipelines/github/statechannels/statechannels/9095/workflows/1f874b42-690b-43b3-813d-a99780962fec/jobs/43551)

bbecf7d [1231ms](https://app.circleci.com/pipelines/github/statechannels/statechannels/9097/workflows/ecadedec-33f1-4614-99ef-a6fb2abdf7bb/jobs/43570)

(the numbers for 25 `makePayment` calls are *very* close to 25x the single `makePayment` call in both cases)

PR #2591, opened 2020-09-22
[server-wallet] Use @statechannels/wasm-utils

andrewgordstewart:
I'd like to see if this works on circle.

### DIFF
--- a/packages/server-wallet/src/utilities/signatures.ts
+++ b/packages/server-wallet/src/utilities/signatures.ts
@@ -1,15 +1,12 @@
-import {instantiateSecp256k1, Secp256k1, RecoveryId} from '@bitauth/libauth';
-import {utils, Wallet} from 'ethers';
+import {Wallet} from 'ethers';
 import {State, StateWithHash} from '@statechannels/wallet-core';
-
-let secp256k1: Secp256k1;
-export const initialized: Promise<any> = instantiateSecp256k1().then(m => (secp256k1 = m));
+import * as wasmUtils from '@statechannels/wasm-utils';
 
 const knownWallets: Record<string, string> = {};
 const cachedAddress = (privateKey: string): string =>
   knownWallets[privateKey] || (knownWallets[privateKey] = new Wallet(privateKey).address);
 
-export async function fastSignState(
+export async function signState(
   state: StateWithHash,
   privateKey: string
 ): Promise<{state: State; signature: string}> {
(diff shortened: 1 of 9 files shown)


geoknee:
Seeing 
```
@statechannels/server-wallet:     RuntimeError: unreachable
@statechannels/server-wallet:         at wasm-function[243]:0x45426
@statechannels/server-wallet:         at wasm-function[320]:0x478ab
@statechannels/server-wallet:         at wasm-function[337]:0x47c3f
@statechannels/server-wallet:         at wasm-function[279]:0x46a97
@statechannels/server-wallet:         at wasm-function[122]:0x3be4d
@statechannels/server-wallet:         at wasm-function[312]:0x4766d
```

geoknee:
The errors went away when I realised I was passing in `server-wallet.State` instead of `nitro-protocol.State`.

geoknee:
@andrewgordstewart would it make more sense to switch to these utils in `nitro-protocol`, rather than `server-wallet` (or even `wallet-core`)?

andrewgordstewart:
Good idea.

The package also has types, so I'm surprised it didn't catch that, but perhaps it's not properly configured.

geoknee:
> The package also has types, so I'm surprised it didn't catch that, but perhaps it's not properly configured.

Yes, see this [discussion](https://github.com/statechannels/native-utils/pull/4#issuecomment-698298326).

geoknee:
I propose that we tackle converting our entire stack to use this wasm stuff in a separate PR. That should

- make it much easier to avoid accidentally using some slower utils 
- share the speed advantages with other applications (e.g. browser wallet)
- cut down on `thisTypeToThatType` gymnastics

and my rationale for separate PR is that there are some questions:
- will there be a circular dependency relationship between `native-utils` and `nitro-protocol`?
- should `native-utils` actually just be part of `nitro-protocol`?
- will it seamlessly run in the browser?

geoknee:
100 makePayment calls taking 
b4cddc8:  [86948 ms](https://app.circleci.com/pipelines/github/statechannels/statechannels/9184/workflows/f5008bea-d152-4b39-a947-4e3ea2e073e4/jobs/44191)
(master) 849f9b6 : [138799 ms ](https://app.circleci.com/pipelines/github/statechannels/statechannels/9179/workflows/f2eaaecf-0428-4258-a068-42e63357b5d3/jobs/44161)

Which looks to be a pretty nice speedup, notwithstanding issues with trusting these numbers as discussed in #2606

geoknee:
Rebased, and snuck in a version bump for `wasm-utils` (which has better typings).

packages/server-wallet/src/utilities/signatures.ts
```diff
@@ -1,59 +1,25 @@
...
 ): Promise<{state: State; signature: string}> {
   const address = cachedAddress(privateKey);
   if (state.participants.map(p => p.signingAddress).indexOf(address) < 0) {
     throw new Error("The state must be signed with a participant's private key");
   }
 
-  const {stateHash} = state;
-
-  const signature = await fastSignData(stateHash, privateKey);
+  const signature = await wasmUtils.signState(toNitroState(state), privateKey).signature;
```
andrewgordstewart, line 19: I think this is now synchronous.

With bitauth, we needed to await the EC to be constructed, and thus this needed to be async.

But the wasm-utils are synchronous -- I think it probably constructs a EC on each call, which is suboptimal, but it still "fast enough".

andrewgordstewart (commented):
GitHub says I'm not allowed to approve this pr, but I approve it anyway!

geoknee (approved):
Great work ;-)

packages/server-wallet/src/utilities/signatures.ts
geoknee: #2619

PR #2592, opened 2020-09-22
[server-wallet] Serialize queued message objective output of createChannel

snario:


kerzhner (approved):
Looks reasonable to me. Looks like e2e tests are failing.

snario:
Nevermind I ran into some kind of serializing hell. Without `serializeObjective` this payload is _already_ serialized since the code serializing it is using a serialized value from the outbox.

Issue #2541, opened 2020-09-07
[integration-tests] Flickering


geoknee:
I have not yet been able to reproduce this locally.

```shell
fails=0
for i in {1..100}
do
  yarn integration-tests > /dev/null 2>&1
  if [ $? -eq 1 ]
  then
    fails++
  fi
  echo "${fails} fails out of ${i} runs"
done
```
produced 0 fails out of 100 for me. 

https://app.circleci.com/pipelines/github/statechannels/statechannels/8847/workflows/2419b50b-c426-4dc2-a8fb-9e085c30333c/jobs/41938

https://app.circleci.com/pipelines/github/statechannels/statechannels/8753/workflows/e32a1582-127b-4fd7-8f57-b6976c7f210c/jobs/41300

andrewgordstewart:
Another example: https://app.circleci.com/pipelines/github/statechannels/statechannels/8866/workflows/09d44d95-0174-44a9-85da-45c5df702de1/jobs/42056

geoknee:
https://app.circleci.com/pipelines/github/statechannels/statechannels/8887/workflows/a8724eeb-ad1c-49e5-8af1-aff9d7226ad2/jobs/42179

geoknee:
Observation: failures tend to have `create-channel.test.ts `passing and then `mount-and-enable.test.ts` failing.  Overall passing tests tend to have `mount-and-enable.test.ts` passing and then `create-channel.test.ts` passing too. 

Although I can't pinpoint exactly why a race would matter here, it does seem fairly plausible that the `enable-ethereum` workflow behaves differently when it has already been run (i.e. as a part of `create-channel.test.ts`). But this seems pretty strange as each test ought to be sandboxed...

andrewgordstewart:
stale

PR #2593, opened 2020-09-23
[integration-tests] Run in band

geoknee:
Might prevent flickering #2541 .


geoknee:
Well maybe it does prevent flickering. But not in the direction I had hoped 😆 . Still, may be a useful data point.

geoknee:
This definitely does not prevent flickering. It might change the frequency, but more investigation is needed.

PR #2595, opened 2020-09-23
fix(wallet-core): Add a check on serializeOutcome to avoid confusion

snario:


geoknee:
Could this problem be solved by removing some type assertions, instead? I think we ought to be relying on the typescript compiler to protect us here, instead of runtime checks.

PR #2601, opened 2020-09-23
[server-wallet]: React to setFunding ChainService calls

kerzhner:
The current strategy is to use a mock chain service to mimic the chain service v2 api.


kerzhner:
No longer needed now that https://github.com/statechannels/statechannels/pull/2633 has been merged

PR #2603, opened 2020-09-23
WIP [server-wallet] Use native utils

geoknee:
Curious results so far
```
➜  server-wallet git:(natives) ✗ ts-node benchmarks/fast-signature.benchmark.ts
signState: 3152.816ms
fastSignState: 203.003ms
nativeSignState: 279.176ms
```

I note that this is not exactly a fair test, since `fastSignState` has the has passed in.

### DIFF
--- a/packages/server-wallet/benchmarks/fast-signature.benchmark.ts
+++ b/packages/server-wallet/benchmarks/fast-signature.benchmark.ts
@@ -1,9 +1,11 @@
 import {ethers, Wallet} from 'ethers';
 import _ from 'lodash';
-import {signState, State, simpleEthAllocation} from '@statechannels/wallet-core';
+import {signState, State, simpleEthAllocation, toNitroState} from '@statechannels/wallet-core';
+import {signState as nativeSignState} from '@statechannels/native-utils';
 
 import {participant} from '../src/wallet/__test__/fixtures/participants';
 import {fastSignState} from '../src/utilities/signatures';
+import {addHash} from '../src/state-utils';
 
 async function benchmark(): Promise<void> {
   const wallet = Wallet.createRandom();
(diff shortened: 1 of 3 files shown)


packages/server-wallet/benchmarks/fast-signature.benchmark.ts
```diff
@@ -24,10 +26,18 @@ async function benchmark(): Promise<void> {
   iter.map(() => signState(state, wallet.privateKey));
   console.timeEnd('signState');
 
+  const stateWithHash = addHash(state);
+
   console.time('fastSignState');
-  const result = iter.map(async () => fastSignState(state, wallet.privateKey));
+  const result = iter.map(async () => fastSignState(stateWithHash, wallet.privateKey));
```
geoknee, line 32: This was a TS error not caught by `yarn build`.

packages/server-wallet/benchmarks/fast-signature.benchmark.ts
```diff
@@ -24,10 +26,18 @@ async function benchmark(): Promise<void> {
...
   await Promise.all(result);
   console.timeEnd('fastSignState');
+
+  const nitroState = toNitroState(state);
+
+  console.time('nativeSignState');
+  iter.map(() => nativeSignState(nitroState, wallet.privateKey));
+  console.timeEnd('nativeSignState');
```
andrewgordstewart, line 40: This might be a more fair comparison.

```suggestion
  console.time('nativeSignState');
  iter.map(() => nativeSignState(toNitroState(state), wallet.privateKey));
  console.timeEnd('nativeSignState');
```

packages/server-wallet/benchmarks/fast-signature.benchmark.ts
andrewgordstewart, line 32: This would make the benchmark a fair comparison. (If done, you should delete `stateWithHash`, but the GH UI did not let me suggest that.)

```suggestion
  const result = iter.map(async () => fastSignState(addHash(state), wallet.privateKey));
```

packages/server-wallet/package.json
```diff
@@ -8,6 +8,7 @@
     "@statechannels/client-api-schema": "^0.3.7",
     "@statechannels/nitro-protocol": "^0.3.7",
     "@statechannels/wallet-core": "^0.3.7",
+    "@statechannels/native-utils": "^0.1.4",
```
andrewgordstewart, line 11: Can you instead install `wasm-utils`?
- it appears like it will work on any (most?) systems
- `native-utils` does not install on my system

geoknee:
I'm going to take over #2591 instead.

PR #2604, opened 2020-09-23
fix(server-wallet): Restrict outbox object type to MessageQueuedNotification

snario:


packages/server-wallet/src/protocols/actions.ts
```diff
@@ -7,7 +7,7 @@ import {Bytes32, Uint256} from '../type-aliases';
 Actions that protocols can declare.
 */
 
-export type Notice = Omit<StateChannelsNotification, 'jsonrpc'>;
+export type Notice = Omit<MessageQueuedNotification, 'jsonrpc'>;
```
snario, line 10: ☝️ Fundamental change of this PR

packages/server-wallet/e2e-test/receiver/controller.ts
```diff
@@ -35,7 +35,7 @@ export default class ReceiverController {
     } = await this.time('push message', async () => this.wallet.pushMessage(message));
 
     if (maybeSyncStateResponse) {
-      const syncResponse = (maybeSyncStateResponse.params as WireMessage).data as Payload;
+      const syncResponse = maybeSyncStateResponse.params.data as Payload;
```
tomclose, line 38: It might be worth using `validatePayload` here instead of casting.

The use-case of wanting to merge payloads is interesting, as this is something a consumer could very reasonably want to do, precisely in this sort of scenario, where they push a message, react to an update, and call some method. We should think about how we should handle this. A couple of possibilities that spring to mind:

1. exposing a `mergePayload` method on the wallet, or perhaps better, a `mergeOutboxes` method
2. the wallet accepts `unknown`s, but emits `Payload`s, allowing the application to e.g. merge them using a utility
3. allowing any wallet operation to accept some messages to be merged into the result

tomclose (approved):
Looks good to me. Feel free to merge. We do need to figure out how to handle the message merging though, so we should schedule that for a discussion.

PR #2607, opened 2020-09-23
Remove CreateChannel objective

kerzhner:
For background, reference https://www.notion.so/Channel-creation-9f16c02442704490bde6203284f18026#b37bf705e2f742ada29eb1f005fdccd1.

The wallet:
- Sends an `OpenChannel` objective after signing state with turn 0.
- Creates a new channel in the database upon receiving a new state for which a channel does not exist.
- Updates `fundingStrategy` upon receiving the `OpenChannel` objective.

### DIFF
--- a/packages/client-api-schema/client-api-schema.api.md
+++ b/packages/client-api-schema/client-api-schema.api.md
@@ -259,6 +259,9 @@ export type ErrorCodes = {
         NotYourTurn: 403;
         ChannelClosed: 404;
     };
+    SyncChannel: {
+        ChannelNotFound: 1400;
+    };
     PushMessage: {
         WrongParticipant: 900;
     };
@@ -283,7 +286,7 @@ export type ErrorCodes = {
 export type ExternalDestination = string;
 
 // @public (undocumented)
-export type FundingStrategy = 'Direct' | 'Ledger' | 'Virtual';
+export type FundingStrategy = 'Direct' | 'Ledger' | 'Virtual' | 'Unfunded' | 'Unknown';
 
 // @public (undocumented)
 export interface Funds {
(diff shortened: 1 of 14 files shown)


packages/client-api-schema/client-api-schema.api.md
```diff
@@ -259,6 +259,9 @@ export type ErrorCodes = {
         NotYourTurn: 403;
```
geoknee, line 259: This file (as well as the duplicate under `/temp/` are autogenerated: I'm not sure why they changed given no changes were otherwise made to `client-api-schema`?

geoknee (approved):

packages/server-wallet/src/wallet/store.ts
```diff
@@ -199,7 +201,30 @@ export class Store {
     channel = await timer('adding MY state', async () => this.addMyState(channel, signedState, tx));
 
     const sender = channel.participants[channel.myIndex].participantId;
-    const data = await timer('adding hash', async () => ({signedStates: [signedState]}));
+    let data: Payload = await timer('adding hash', async () => ({
+      signedStates: [signedState],
+      objectives: [],
+    }));
```
geoknee, line 207: General question (applies before this PR): 
- why is this being timed? 
- why is it labelled 'adding hash'? 
- why is it `async`?

packages/client-api-schema/client-api-schema.api.md
kerzhner: I wonder if my local environment is different from that of the last machine that built these files? If you rebuild these files on your machines, are there changes?

geoknee: > If you rebuild these files on your machines, are there changes?

Before this PR, yes, after this PR, no. 

So I think that means (although unrelated to this PR) we needed to check these changes in. Or we could just gitignore them for now, since they are only used as a part of the docs generation pipeline at present. And it's weird that there is a duplicate. 

On the other hand, they are potentially useful canaries in our coal mine, as they highlight when we make changes to the API of one of these packages. So we could think about leveraging that somehow.

geoknee: #2614

PR #2617, opened 2020-09-25
ci: Add CI job to validate commits against conventional commits standard

snario:


andrewgordstewart:
Three questions:
1. What does this do?
1. Have we decided to use conventional commits in this repo, at this point in time?
2. Are we sure we want to validate every commit? Does this do so? Is there a way to opt out of commit validation?

My vague understanding is that "conventional commits" allows for unconventional commits, in that they are simply ignored. Perhaps we want to maintain the possibility of including an unconventional commit, but we want to make the developer specifically ask for that.

An analogous situation is I can include `[skip ci]` in a commit header, and circle will not run the CI suite.

andrewgordstewart:
Another question, which is sort of a personal request: is there a clear set of tools we could suggest (ie. in the README) that will enable developers to satisfy the validation, without a lot of mental overhead?

I don't think we need to put a lot of effort into this, but eg. maybe there is a [VS-Code extension](https://marketplace.visualstudio.com/items?itemName=vivaxy.vscode-conventional-commits) we can recommend.

tomclose:
@snario I think this would be useful in statechannels/the-graph. My understanding is we've decided to do conventional commits there and will later decide whether to do it in the monorepo.

geoknee:
This alternative seems to me to have clearer documentation: https://github.com/wagoid/commitlint-github-action. The default config checks for things listed here: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional

I'm not sure about a VSCode extension, but there is at least an npm package command line tool https://commitlint.js.org/#/ which we could even choose to run in a precommit hook.

geoknee:
There's also this tool  http://commitizen.github.io/cz-cli/ to author compliant commits with prompts.

PR #2618, opened 2020-09-25
style(server-wallet): Add linting exception for loader.js

snario:


geoknee:
is `style(package-name)` conventional? Should we be doing that instead of `[package-name] fix:` (say) ?

snario:
Based on this: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

tomclose:
You also don't need to put the package name, if you don't want to. It can figure out which packages were affected by looking at the code that changed.

tomclose (approved):

PR #2619, opened 2020-09-25
chore: remove reference to wallet-core signState

geoknee:
- And all signing is now sync (not async)
- SigningWallet.signState can take a State or StateWithHash, adds the hash if necessary


geoknee:
I'm not sure what caused the test failure on the first commit. I'm sure adding a test didn't fix it 😕

tomclose (approved):

PR #2625, opened 2020-09-28
chore: Remove build step from circle

geoknee:
The `build` command is largely duplicating work done during the `prepare` step. It is not one of the `npm` lifecycle [scripts](https://docs.npmjs.com/misc/scripts). It runs on every push and takes 2-10 minutes (ish). 

It does some useful things for us that aren't currently included in `prepare`. It 

- builds packages that are not depended on (which we don't "need" to do in `prepare`, but we probably should*) such as the xstate-wallet
- runs `lint:check`
- builds storybook in `xstate-wallet`. 

I consider these all useful to do on every push, to prevent us breaking these things without realising. So I have ensured they run during `prepare` instead. From what I can tell this hasn't significantly impacted the typical running time for `prepare` on circle.

> *`prepare`: Run both BEFORE the package is packed and published, on local npm install without any arguments,

This PR also introduces cleaning the `lib` folder before running `tsc`, as that seems like a good pattern to adopt everywhere.


geoknee:
Post review we can make the `build` job non-required on github. That would make it mergeable.

tomclose (approved):

PR #2626, opened 2020-09-28
fix(devtools): Spawn ganache command in a shell

### LINKED ISSUE #2622: Ganache does not start on Windows
When trying to start xstate-wallet on Windows, after installing from NPM, ganache fails to startup because of a problem in NodeJS' spawn() function.

I tried to make a PR for you but did not have the permissions to push a branch. The fix is pretty easy. In statechannels/packages/devtools/src/ganache/server.ts, on line 153, change:

`this.server = spawn('npx', ['-c', cmd], {stdio: 'pipe'});`

to 

`this.server = spawn('npx', ['-c', cmd], {stdio: 'pipe', shell: true});`

geoknee:
Closes #2622

### DIFF
--- a/packages/devtools/src/ganache/server.ts
+++ b/packages/devtools/src/ganache/server.ts
@@ -150,7 +150,7 @@ export class GanacheServer {
 
     const cmd = `ganache-cli ${opts}`;
 
-    this.server = spawn('npx', ['-c', cmd], {stdio: 'pipe'});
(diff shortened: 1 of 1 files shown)


ditchfieldcaleb:
FYI, this seems to break things on Linux - env variables don't seem to be carried over to the ganache process.

tomclose (approved):

PR #2629, opened 2020-09-29
fix(xstate-wallet): Specify correct dependencies

### LINKED ISSUE #2627: [xstate-wallet] Missing dependencies
There are apparently still some missing dependencies in `package.json` that means these instructions do not work at present. At least that should be fixed before merging this.

_Originally posted by @geoknee in https://github.com/statechannels/statechannels/pull/2598#issuecomment-699009277_

geoknee:
Closes #2627 

This changeset is somewhat large since I chose to redirect imports from `@ethersproject/...` to `ethers`, meaning that `@ethersproject/...` need not be depended on directly.


packages/xstate-wallet/package.json
```diff
@@ -115,6 +115,7 @@
     "@types/webpack-merge": "4.1.5",
     "@types/workbox-webpack-plugin": "5.0.0",
     "babel-loader": "8.1.0",
+    "depcheck": "^1.2.0",
```
geoknee, line 122: This should be cut out, but I'll do that in a separate PR. 

(I had planned on integrating the tool into CI, but that didn't prove straightforward).

andrewgordstewart (approved):

PR #2632, opened 2020-09-29
Merge Outgoing Messages to same recipient and duplicate channel results

### LINKED ISSUE #2623: Server wallet: pushMessage should only return one channelResult per channel
### Expected behaviour
Calling `pushMessage` on the server wallet should never contain more than one `ChannelResult` for each channel. This `ChannelResult` should represent the state of the channel at the end of the `pushMessage` operation i.e. after all actions have been taken and the run loop has terminated.

### Actual behaviour
Currently `pushMessage` sometimes returns more than one `ChannelResult` per channel. In the `PaymentManager` I'm running into a case where the `ReceiptManager` provides a payload that contains states 0 and 3 in a given channel. The response from `pushMessage` contains three `ChannelResults`.

lalexgap:
This PR implements two new functions that get called in the `wallet`:

- `mergeOutgoing`
- `mergeChannelResults`

`mergeOutgoing` merges multiple messages to the same `receipient` into one message. This allows one client to send all the messages to another in one request/response cycle.

`mergeMultipleChannelResults` merges `ChannelResult[]`s by only allowing one entry per a `channelId`. If there are multiple entries it takes the one with the highest `turnNum` . Fixes #2623 .

### DIFF
--- a/packages/server-wallet/src/utilities/__test__/messaging.test.ts
+++ b/packages/server-wallet/src/utilities/__test__/messaging.test.ts
@@ -45,10 +45,14 @@ describe('mergeOutgoing', () => {
       params: {recipient: USER1, sender: USER2, data: {signedStates: [state2]}},
     };
 
-    const merged = mergeOutgoing([message1, message2]);
+    const merged = mergeOutgoing([message2, message1]);
     expect(merged).toHaveLength(1);
 
     expect((merged[0] as any).params.data.signedStates).toHaveLength(2);
+    // Merging should sort the states by ascending turnNum
+    expect(
+      (merged[0] as any).params.data.signedStates.map((s: {turnNum: number}) => s.turnNum)
+    ).toEqual([1, 2]);
   });
 
   test('it handles duplicate states', () => {
--- a/packages/server-wallet/src/utilities/__test__/messaging.test.ts
+++ b/packages/server-wallet/src/utilities/__test__/messaging.test.ts
@@ -17,7 +17,7 @@ describe('mergeChannelResults', () => {
     expect(result[0]).toMatchObject({turnNum: 2});
   });
 
-  test("it doesn't merge  different channels", () => {
+  test("it doesn't merge different channels", () => {
     const duplicateChannelResult: ChannelResult[] = [
       channelWithVars({channelId: '0x123'}).channelResult,
       channelWithVars({channelId: '0xabc'}).channelResult,
@@ -52,6 +52,8 @@ describe('mergeOutgoing', () => {
   });
 
   test('it handles duplicate states', () => {
+    // We use two seperate state objects that are equivalent to avoid things passing
+    // due to object references being the same
     const state1 = stateSignedBy()({turnNum: 1});
     const state2 = stateSignedBy()({turnNum: 1});
     const message1: Notice = {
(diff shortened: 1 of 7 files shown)


packages/server-wallet/src/utilities/__test__/messaging.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    };
+    const message2: Notice = {
+      method: 'MessageQueued',
+      params: {recipient: USER1, sender: USER2, data: {signedStates: [state2]}},
+    };
+
+    const merged = mergeOutgoing([message1, message2]);
+    expect(merged).toHaveLength(1);
+
+    expect((merged[0] as any).params.data.signedStates).toHaveLength(2);
```
andrewgordstewart, line 51: Should we expect the states to be sorted by turn number, for a given channel?

packages/server-wallet/src/utilities/__test__/messaging.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  test('it handles duplicate states', () => {
+    const state1 = stateSignedBy()({turnNum: 1});
+    const state2 = stateSignedBy()({turnNum: 1});
```
andrewgordstewart, line 56: This test would be more clear to me if there were no state 2 or message 2, but you instead just called `mergeOutgoing` with `[message1, message1]`

packages/server-wallet/src/utilities/messaging.ts
```diff
@@ -0,0 +1,58 @@
+import {ChannelResult} from '@statechannels/client-api-schema';
+import {Payload} from '@statechannels/wallet-core';
+import _ from 'lodash';
+
+import {Notice} from '../protocols/actions';
+
+// Merges any messages to the same recipient to make
```
andrewgordstewart, line 7: To make what?

packages/server-wallet/src/utilities/messaging.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (differentSender) {
+    throw new Error(`Trying to merge outgoing messages with multiple recipients ${senders}`);
+  }
+
+  const {sender} = outgoing[0].params;
+
+  const mergedOutgoing: Record<string, Payload> = {};
+
+  for (const notice of outgoing) {
+    const {recipient, data} = notice.params;
```
andrewgordstewart, line 23: Maybe you can assert on the type of data here so you don't have to repeatedly do it below?

andrewgordstewart (approved):
LGTM!

packages/server-wallet/src/utilities/__test__/messaging.test.ts
lalexgap: One possible pitfall is that if we're not doing a deep equality check in the `merge` method we may miss that with `[message1, message1]` since it would pass due to the object references being the same.

 Perhaps I should be more explicit about that in a comment

andrewgordstewart: Good point. An intermediate would be to clone message1 to construct message2.

packages/server-wallet/src/utilities/messaging.ts
lalexgap: Oops!

packages/server-wallet/src/utilities/__test__/messaging.test.ts
lalexgap: We can now!

Issue #2534, opened 2020-09-03
[server-wallet(ish)] Onchain Service Discussion


LayneHaber:
There are still several items to discuss regarding the behavior of the `OnchainService`. Here are a few (listed in no particular order):

- What is the best way to make this usable in a multichain environment? Multiple instances of the class? Configuring the service to handle multiple chains and providers?
- What's the best chain service <--> wallet <--> app architecture and division of responsibilities?
- How should the onchain service handle different types of asset holders? (think `ZKSyncDaiAssetHolder` or something)
- How should we handle historical event syncing?

This is by no means a comprehensive list, but it is a start!

snario:
I like the idea that it is the responsibility of a service external to the `server-wallet` to **set** definitively the on-chain funding state of the system. This has nice properties like externalizing the entirety of event interpretation, handling of orphaned blocks, handling confirmation windows, and other kinds of things like that to a separate service.

LayneHaber:
I agree, I think the part I'm hung up on is the best way to get that information to the wallet (I'm assuming that the chain store and the channel store are completely separate). We can allow an accessor method on the service interface, but another idea @andrewgordstewart and I threw around a bit was to have the service emit essentially the same event twice, once when its emitted by the contract, and once when it is finalized. The "app" (which I consider to be the thing governing the wallet, and already responsible for routing messages to the wallet), can then register a handler to the events, and push that information via some `pushEvent`, `setAdjudicatorState`, etc. method on the wallet.

I think I'm lacking some context into what the wallet will need to get from the chain and when that may be messing with my mental models slightly on this front, so this may not be the most ideal way for the data to flow

snario:
The wallet needs to know:

- When its user has made a deposit
- When a counterparty has made a deposit
- When a withdrawal has been made
- When a challenge has been initiated
- When a challenge has been responded to
- When a checkpoint has been made

Specifically it needs to know when it can be 100% confident in this information in order to make a decision based on it. For example, it may be waiting to sign a state until after it observes some funds deposited on-chain. Or, it may not sign any new states until it sees a challenge responded to on-chain. Or, if it initiates a challenge, it may lock that channel preventing any future states being signed for it until that challenge has been responded to (i.e., because otherwise it is committing an attributable fault).

[...]

PR #2633, opened 2020-09-29
Chain Service v1

kerzhner:
Background: @LayneHaber [recently built ChainService v0](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/onchain-service/index.ts). After that, we implemented directly funded channels in the server wallet and realized that Chain Service v0 API necessitated that the server wallet constructed chain transactions and understood chain events.

This PR implements `registerChannel` and `fundChannel` methods that abstract away chain specific constructs from the wallet as recommended by @snario [here ](https://github.com/statechannels/statechannels/issues/2534#issuecomment-687112052) and documented [here](https://www.notion.so/v1-Spec-4aeae35305b14a45bccf594de4291fe3)


packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -0,0 +1,120 @@
...
+      subscriber.onFundingChanged({
+        channelId,
+        assetHolderAddress: contractAddress,
+        amount: BN.from(holding),
+      });
+    });
+
+    return obs.pipe(share());
+  }
+}
```
kerzhner, line 120: This is probably the densest/most complex part of the PR. It would be good to have at least one reviewer fully understand this code.

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -0,0 +1,120 @@
...
+    channelId: Bytes32,
+    subscriber: ChainEventSubscriberInterface
+  ): Observable<FundingChangedArg> {
+    const contract: Contract = new Contract(
+      contractAddress,
+      ContractArtifacts.EthAssetHolderArtifact.abi
+    ).connect(this.provider);
+
+    // Create an observable that emits events on contract events
+    const obs = new Observable<FundingChangedArg>(subscriber => {
```
lalexgap, line 98: Having `subscriber` shadowed here makes it slightly harder to follow the code. Maybe we could rename it to `s`?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -0,0 +1,120 @@
...
+        subscriber.next({
+          channelId: destination,
+          assetHolderAddress: contractAddress,
+          amount: BN.from(destinationHoldings),
+        })
+      );
+    });
+
+    // Fetch the current contract holding, and emit as an event
+    contract.holdings(channelId).then((holding: any) => {
```
lalexgap, line 110: Is this to fetch the current holdings when first subscribing to the observable? If so could we use  `merge`  like we do [here](https://github.com/statechannels/statechannels/blob/348e627ae94c1dac10d14a028fe31434bce9db8c/packages/xstate-wallet/src/chain.ts#L205)? That way the `createContractObservable` method wouldn't need to take in a `subscriber`

packages/server-wallet/src/mock-chain-service/index.ts
```diff
@@ -9,9 +17,42 @@ type FundChannelArg = {
...
   static fundChannel(_arg: FundChannelArg): Promise<SubmissionStatus> {
     const submissionStatus: SubmissionStatus = 'Success';
     return Promise.resolve(submissionStatus);
   }
+
+  registerChannel(_channelId: Bytes32, _assetHolders: Address[]): Promise<void> {
+    return Promise.resolve();
+  }
+
+  // TODO: remove in v1
```
lalexgap, line 39: Should this be removed?

packages/server-wallet/.env.test
```diff
@@ -6,6 +6,6 @@ SERVER_DB_USER=root
 HUB_DESTINATION='hub'
 
 GANACHE_HOST=0.0.0.0
-GANACHE_PORT=8547
+GANACHE_PORT=8545
```
lalexgap, line 9: Is this to avoid the polling issue? Maybe we could leave a comment to clarify :)

packages/server-wallet/src/mock-chain-service/index.ts
kerzhner: Chain Service v1 eliminates the need for both `mock-chain-service` and `onchain-service` (which is chain service v0). My thought was to remove those in a separate PR in order to reduce the change set of this PR and make it easier to roll back changes if that needs to be done.

packages/server-wallet/.env.test
kerzhner: The reason I made this change is because other places in the `server-wallet` reference port `8545` like [here](https://github.com/statechannels/statechannels/blob/chain-service/packages/server-wallet/package.json#L95) and [here](https://github.com/statechannels/statechannels/blob/chain-service/packages/server-wallet/deployment/deploy.ts#L18). We should probably change both of those to reference an environment variable instead of hardcoding `8545`.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Good catch! 23fe7a2c9

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: That's good feedback, and I'll play around with that. I would like to keep `registerChannel` as a synchronous method, so let me see if I can make that work.

andrewgordstewart: You could fetch it asynchronously and concat instead of merge if you want to guarantee the order

kerzhner: `concat` worked well! b1279f5f2

lalexgap (approved):
This looks good to me

PR #2635, opened 2020-09-30
migrate to yarn 2

geoknee:
Notes: 

Module resolution is handled in quite a radically different way (no more node_modules). This means 

- installation can be super fast (shaving 3 minutes off circle)
- it is _much_ more difficult for code to run if you forgot to specify a dependency in your manifest. Or, if one of your dependencies forgot to specify a dependency in _their_ manifest. These problems must be manually fixed. 

Side effects on this branch
- no more `etherlime` which means no more `scrypt` 🎉 
- typescript upgraded to v4
- lerna partially deprecated in favour of `yarn`
- devtools does not spawn `ganache-cli` in a separate process

https://yarnpkg.com/advanced/migration


.gitignore
```diff
@@ -62,3 +62,12 @@ storybook-static
 ganache-deployments.json
 ganache-deployments*.json
 *.lock
+
+# yarn 2
+# https://yarnpkg.com/advanced/qa#which-files-should-be-gitignored
+.yarn/*
+!.yarn/cache
```
andrewgordstewart, line 69: It looks like these files are checked in down below.

It would be nice to not have those files as a part of the history. Otherwise, the repo will permanently be `X` megabytes larger.

andrewgordstewart: Here is a relevant list of tradeoffs: https://github.com/yarnpkg/berry/issues/180#issuecomment-493789642

I would prefer using `git-lfs` if we choose to do zero-installs, so that we don't end up in this kind of a situation: https://github.com/yarnpkg/berry/issues/180#issuecomment-506022504

geoknee: > Do you intend to check in a cache of each dependency in the repository?

Yes, as per the recommendations for zero-installs.

> I would prefer using git-lfs if we choose to do zero-installs

I looked into this before and I recall that `git-lfs` is not well suited to large numbers of files. It is designed for things like `.psd` files. It is not possible to point `git-lfs` at an entire directory.

andrewgordstewart: Then I would advocate for _not_ using zero-installs, and evaluating that choice separately.

Using zero installs has a very high cost: the size of the repo will jump by hundreds of megabytes initially, and by tens to hundreds of megabytes any time we make a large change to our dependencies. This is difficult to change after the fact, as it could require re-writing git history on the `master` branch.

geoknee: > Then I would advocate for _not_ using zero-installs, and evaluating that choice separately.

Noted that zero-installs has disadvantages. I'll try and compile a list of pros and cons specific to our project and we can see whether the tradeoff is acceptable or not.

.circleci/config.yml
```diff
@@ -195,9 +181,8 @@ jobs:
             - packages/*/dist
             - packages/*/build
             - packages/*/lib
-            - packages/*/node_modules
             - packages/client-api-schema/types
-            - node_modules
+            - .yarn
```
geoknee, line 185: Thinking about this some more, I think the right thing to do is to completely remove circle caching `save_cache` and `restore_cache`, since the dependencies are pulled down in the `checkout_code` step.

geoknee:
Problems encountered (besides large number of files issue): 

- incompatibility with lerna (it complains about the new [`workspace:` range syntax](https://yarnpkg.com/features/workspaces#workspace-ranges-workspace) in updated manifests)
- `yarn workspaces foreach` feels inferior to `lerna run`. There is not option to `--bail` or filter using `--since`, for example, and I hit this (closed?) issue https://github.com/yarnpkg/berry/issues/977. I also found a certain test would pass when run in its own package directory but fail when run using this command (although this could be for some other reason). 
- couldn't get eslint to work (yet)
- incompatible with [skip dependency](https://www.npmjs.com/package/@favware/skip-dependency)
- incompatible with patch-package, although I was able to workaround this using yarn's own `patch` feature.

geoknee:
TODO: cleanup the repo so the history does not include all of these files in the yarn cache. There's [another branch](https://github.com/statechannels/statechannels/tree/yarn2-b-node-modules) that keeps yarn 2 but uses good old node_modules. That branch's history should also be rewritten.

PR #2637, opened 2020-09-30
Add joinChannels and updateFundingForChannels method

lalexgap:
- Adds a `JoinChannels` method that allows for joining multiple channels in one call to the wallet.
- Adds a `updateFundingForChannels` method that allows up for updating the funding for multiple channels in one call to the wallet.


lalexgap:
Closing in favour of #2638

PR #2638, opened 2020-10-01
Add methods to wallet for batch processing

lalexgap:
This PR introduces three new methods to the wallet API:
- JoinChannels - Allows the joining of multiple channels in one call to the wallet.
- UpdateFundingForChannels - Allows the updating of funding for multiple channels in one call to the wallet.
- MergeMessages - Merges multiple messages into one message. This is useful for merging the result of different API calls in one request/response cycle.

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -49,8 +49,13 @@ import {Store, AppHandler, MissingAppHandler} from './store';
 // TODO: The client-api does not currently allow for outgoing messages to be
 // declared as the result of a wallet API call.
 // Nor does it allow for multiple channel results
-export type SingleChannelResult = Promise<{outbox: Outgoing[]; channelResult: ChannelResult}>;
-export type MultipleChannelResult = Promise<{outbox: Outgoing[]; channelResults: ChannelResult[]}>;
+export type SingleChannelResult = Promise<SingleChannelMessage>;
+export type MultipleChannelResult = Promise<MultipleChannelMessage>;
+type SingleChannelMessage = {outbox: Outgoing[]; channelResult: ChannelResult};
+type MultipleChannelMessage = {outbox: Outgoing[]; channelResults: ChannelResult[]};
+type Message = SingleChannelMessage | MultipleChannelMessage;
+const isSingleChannelMessage = (message: Message): message is SingleChannelMessage =>
+  'channelResult' in message;
 
 export interface UpdateChannelFundingParams {
   channelId: ChannelId;
@@ -83,6 +88,8 @@ export type WalletInterface = {
 
   // Register chain <-> Wallet communication
   attachChainService(provider: OnchainServiceInterface): void;
+
+  mergeMessages(messages: Message[]): Message;
 };
 
 export class Wallet implements WalletInterface, ChainEventListener {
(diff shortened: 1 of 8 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -66,14 +71,15 @@ export type WalletInterface = {
...
   updateChannel(args: UpdateChannelParams): SingleChannelResult;
   closeChannel(args: CloseChannelParams): SingleChannelResult;
   getChannels(): MultipleChannelResult;
   getState(args: GetStateParams): SingleChannelResult;
   syncChannel(args: SyncChannelParams): SingleChannelResult;
 
-  updateChannelFunding(args: UpdateChannelFundingParams): void;
-
+  updateChannelFunding(args: UpdateChannelFundingParams): SingleChannelResult;
+  updateFundingForChannels(args: UpdateChannelFundingParams[]): MultipleChannelResult;
```
kerzhner, line 82: Could be nice for the naming to be consistent. So maybe `updateChannelFunding`/`updateChannelsFunding` or `updateFundingForChannel`/`updateFundingForChannels`

kerzhner: Another option is to have just one api call `updateChannelsFunding` that works for 1 or more channels.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -240,6 +273,19 @@ export class Wallet implements WalletInterface {
     );
     return {outbox: mergeOutgoing(outgoing.map(n => n.notice)), channelResult};
   }
+
+  async joinChannels(channelIds: ChannelId[]): MultipleChannelResult {
```
kerzhner, line 277: Do we need `joinChannel` if we have a `joinChannels` that already works for a single element `channelIds` array?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -102,11 +110,13 @@ export class Wallet implements WalletInterface {
     // Bind methods to class instance
     this.getParticipant = this.getParticipant.bind(this);
```
kerzhner, line 111: Probably doesn't belong in this PR, but it might be time to refactor this code to:
```
const methods = [this.getParticipants, this.updateChannelFunding, ...];
methods.map(m => m.bind(this))
```

kerzhner (approved):

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: It would also be nice to have an explanation of why this is necessary.

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: There's a question of maintaining backwards compatibility. But I think that the server-wallet's API is closer to the One True API than the `client-api-schema`'s opinion.

lalexgap: I've removed `joinChannel`, `createChannel` and `updateChannelFunding` in 
1684ff4

tomclose: The backwards compatibility struggle is real. Especially when `createChannel` would guarantee you a `SingleChannelResult`, but `createChannels` doesn't. I suggest that we put these methods back in - it costs very little to maintain them, and we reduce the hassle burden for users of the wallet.

tomclose: For reference, this is [the diff](https://github.com/statechannels/the-graph/pull/180/files#diff-290c3918607a033549623003e44452b1) for the payment manager:

Screen Shot 2020-10-05 at 3 38 45 PM

PR #2641, opened 2020-10-01
Mock Chain Service implements v1 interfaces

kerzhner:
In this PR:
- Mock Chain Service (which is currently used by the wallet) implements the v1, rich chain service api. The mock chain service is also moved under the `chain-service` directory.
- Chain service v0 (formerly known as onchain-service) is moved under the `chain-service` directory. We could delete chain service v0, but it might be nice to keep it around to have a reference on how to implement features lacking from chain service v1. These features include transaction resubmission on known errors and erc20 asset holder support.


packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -24,19 +24,21 @@ export interface ChainEventSubscriberInterface {
   onHoldingUpdated(arg: HoldingUpdatedArg): void;
 }
 
-interface ChainEventEmitterInterface {
+export interface ChainEventEmitterInterface {
```
tomclose, line 27: This interface doesn't seem to be used elsewhere. Could we get away without exporting it?

I'm also wondering it would be simpler to just define a single `ChainServiceInterface`, instead of these two separate interfaces which are then combined. Do you envisage a situation soon where we'll have the two interfaces implemented by separate objects?

tomclose (approved):

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Good catch on not needing to export the interfaces used to compose the `ChainServiceInterface`. Fixed in d9cf5b610.

My initial reasoning for separating the interfaces was to allow a separate implementation of transaction submission and event listening services:
- `ChainModifierInterface` submits transactions. It can be a standalone service where a process is spun up on request and spun down after a transaction is successfully mined.
- `ChainEventEmitterInterface` monitors the chain. It has to be a long lived process or a some sort of an event listener.

In practice, we are no longer thinking of splitting out the chain service as a separate process (or processes) in the medium term. So the two interfaces can be combined.

PR #2645, opened 2020-10-02
Add erc20 support to the chain service

kerzhner:
This PR is extends https://github.com/statechannels/statechannels/pull/2641.

### DIFF
--- a/packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
+++ b/packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
@@ -1,19 +1,23 @@
-import {randomChannelId} from '@statechannels/nitro-protocol';
+import {ContractArtifacts, randomChannelId} from '@statechannels/nitro-protocol';
 import {BN} from '@statechannels/wallet-core';
-import {providers} from 'ethers';
+import {BigNumber, Contract, providers, Wallet} from 'ethers';
 
 import {defaultConfig} from '../../config';
+import {Address} from '../../type-aliases';
 import {ChainService} from '../chain-service';
 
-/* eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion */
+/* eslint-disable no-process-env, @typescript-eslint/no-non-null-assertion */
 const ethAssetHolderAddress = process.env.ETH_ASSET_HOLDER_ADDRESS!;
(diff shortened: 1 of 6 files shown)


packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -59,6 +65,24 @@ describe('fundChannel', () => {
...
 });
 
+it('Fund erc20', async () => {
+  const channelId = randomChannelId();
+  const tokenContract: Contract = new Contract(
+    erc20Address,
+    ContractArtifacts.TokenArtifact.abi,
+    ethWallet
+  );
+  await (await tokenContract.increaseAllowance(erc20AssetHolderAddress, BN.from(5))).wait();
```
tomclose, line 75: Should increasing the allowance be part of the `fundChannel` method?

tomclose (approved):

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
kerzhner: Good question. Probably yes, at least in the short term. There might be use cases where a wallet wants to approve many tokens for use in multiple channels for an asset holder.

snario: It's not uncommon to approve an infinite amount, and in our case it may be reasonable considering the contract would be the ERC20 asset holder which has safe code that we know can't be manipulated by anyone except the caller.

snario:
How do you think we move from here to support a third or fourth kind of asset type? For example, an application may have its own custom asset type it wants its users to use. Will it need to edit the code of the chain service? Perhaps the methods to provide an asset holder address and ABI should be public on the service?

kerzhner:
> How do you think we move from here to support a third or fourth kind of asset type? For example, an application may have its own custom asset type it wants its users to use. Will it need to edit the code of the chain service? Perhaps the methods to provide an asset holder address and ABI should be public on the service?

The asset holder address (or addresses) is an argument in both [fund channel](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L26) and [register channel](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L38) APIs. It's a good question on how to specify an ABI for those addresses. To be honest, I haven't given that much thought yet...

PR #2646, opened 2020-10-05
Upgrade and cleanup dependencies

geoknee:
- Removes @ethersproject/foo (with one exception)
- upgrades typescript and @typescript/eslint in all packages <- this flagged a lot of warnings and errors that I have (mostly) just ignored for now
- upgrades xstate
- upgrades api-extractor
- regenerates yarn.lock

This is mostly a cherry pick from #2635, and should help fix many "missing dependency" problems


packages/channel-client/src/channel-client.ts
```diff
@@ -343,6 +344,7 @@ export class ChannelClient implements BrowserChannelClientInterface {
    * @returns A promise that resolves to a DomainBudget.
    *
    */
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
tomclose, line 347: What's this doing?

packages/devtools/.eslintrc.js
```diff
@@ -6,6 +6,8 @@ module.exports = {
   extends: ['../../.eslintrc.js'],
   rules: {
     '@typescript-eslint/no-explicit-any': 'off',
-    '@typescript-eslint/explicit-function-return-type': 'off'
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/explicit-module-boundary-types': 'off'
+    // TODO remove these ^
```
tomclose, line 11: Can you remove these now?

packages/channel-client/src/channel-client.ts
geoknee: It's a new rule introduced in the upgrade https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md . It objects to our use of `{}` as a type. 

Rather than just globally ignoring this rule, I thought it better to ignore in specific places, giving us a greater chance to remedy in the future.

packages/devtools/.eslintrc.js
geoknee:
```suggestion
    // TODO remove these and fix resulting issues ^
```

geoknee: This is outside the scope of this PR.

tomclose (approved):

PR #2648, opened 2020-10-05
[devtools] Do not spawn ganache server in a child process

geoknee:
I believe this will be more robust (the current way failed during my yarn 2 experiment). It may also make it easier to collect logs. It should also enable tools like `depcheck` to correctly infer that `ganache-cli` should be a dependency.

We could also think about using `ganache-core` instead of `ganache-cli`.


kerzhner:
Thanks for looking into this @geoknee! This work would also be helpful as I am seeing the following behavior:
- When `yarn` is used to start unit tests, `ganache-cli` executable is successfully located (even when the executable is in the monorepo root `node_modules/.bin` as opposed to the package `node_modules/.bin`
- When `node jest` is use to start unit tests, the binary is not located. vs code starts unit tests this way.

packages/devtools/src/ganache/server.ts
```diff
@@ -137,29 +134,29 @@ export class GanacheServer {
...
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      network_id: this.chainId,
+      accounts: accounts.map(a => {
+        return {balance: oneMillion, secretKey: a.privateKey};
+      }),
+      gasLimit,
+      gasPrice: BigNumber.from(gasPrice).toHexString(),
+      logger: {
+        log: x => {
+          extractLogsFromVerboseGanacheOutput(x);
```
geoknee, line 151: I wasn't 100% sure of the value of a) piping through this function as opposed to b) piping straight to `logger.info`.

a produces logs like [this](https://gist.github.com/geoknee/765ff7b0f6c1935879dbe9bd9d7ee5c6). 
b produces logs like [this](https://gist.github.com/geoknee/1b902a552588ebc10036ae8ebe4093e8).

geoknee: (I went with 'a' so far)

packages/devtools/src/ganache/server.ts
```diff
@@ -137,29 +134,28 @@ export class GanacheServer {
...
-        logger.info(data.toString());
+    const oneMillion = ethers.utils.parseEther('1000000').toHexString();
+
+    // ganache core exports a very permissive object[] type for accounts
+    // it should be {balance: HexString, secretKey: string}[]
+    const serverOptions: ganache.IServerOptions = {
+      network_id: this.chainId,
+      accounts: accounts.map(a => {
+        return {balance: oneMillion, secretKey: a.privateKey};
+      }),
```
kerzhner, line 145:
```suggestion
      accounts: accounts.map(a => ({balance: oneMillion, secretKey: a.privateKey}),
```

packages/devtools/src/ganache/server.ts
kerzhner: What about when verbose logging is off?

kerzhner (approved):

geoknee:
> What about when verbose logging is off?

AFAICT there is no verbosity switch when you run ganache this way: it seems to just default to being verbose.

packages/devtools/src/ganache/server.ts
geoknee: See my other comment: I think the logging is always verbose and [no way to switch it of](https://www.npmjs.com/package/ganache-core#options)f. Should probably clean up some environment variables in light of that.

kerzhner:
> > What about when verbose logging is off?
> 
> AFAICT there is no verbosity switch when you run ganache this way: it seems to just default to being verbose.

I am not 100% sure, but I think we turned off verbose logging by default as the ganache logs were getting quite large. I wonder if that will be an issue with this implementation.

packages/devtools/src/ganache/server.ts
geoknee: I'll revise that to saying there is no way to switch verbosity, but whether this means verbose is ON or OFF or not applicable, I do not know.

geoknee:
There's a bit of investigation to do to figure out what level of verbosity we are getting from ganache. Hopefully it is full verbosity, and that it still works well with the `extractLogsFromVerboseGanacheOutput()` function. I think we can still have some control of log size by setting the `LOG_LEVEL`.

geoknee:
So the answer is actually pretty simple. `verbose` is in fact an option that can be passed to the ganache server (unsurprisingly), it just isn't listed in the readme. I've reinstated switching that option based on environment variables, and also conditionally piping the output to @andrewgordstewart 's helper function which extracts better logs.  So the behaviour should be the same as what we had before.

PR #2650, opened 2020-10-05
Enforce max connection pool size of 1 for threads

lalexgap:
This ensures that worker threads only use one `postgres` connection. (This seems to be the default behaviour but this enforces it)

### DIFF
--- a/packages/server-wallet/src/utilities/workers/worker.ts
+++ b/packages/server-wallet/src/utilities/workers/worker.ts
@@ -10,7 +10,10 @@ import {ServerWalletConfig} from '../../config';
 import {isStateChannelWorkerData} from './worker-data';
 
 // We only expect a worker thread to use one postgres connection but we enforce it just to make sure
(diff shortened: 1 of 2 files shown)


packages/server-wallet/src/utilities/workers/worker.ts
```diff
@@ -9,7 +9,9 @@ import {ServerWalletConfig} from '../../config';
 
 import {isStateChannelWorkerData} from './worker-data';
 
-const walletConfig = workerData as ServerWalletConfig;
+// We only expect a worker thread to use one postgres connection but we enforce it just to make sure
+const walletConfig = {...(workerData as ServerWalletConfig), postgresPoolSize: {min: 0, max: 1}};
```
andrewgordstewart, line 13: Can we utilize the type system to ensure this remains a valid manipulation of the wallet config?

This will probably need to be prettied up.
```suggestion
const walletConfig: ServerWalletConfig = {...(workerData as ServerWalletConfig), postgresPoolSize: {min: 0, max: 1}};
```

andrewgordstewart (approved):

PR #2651, opened 2020-10-05
Chain service calls increaseAllowance for erc20 asset holders

kerzhner:
For erc20 asset holders, the chain service calls `increaseAllowance` before depositing to address https://github.com/statechannels/statechannels/pull/2645#discussion_r499456033. The alternative (which is the implementation prior to this PR) is for the chain service to assume that `increaseAllowance` has already been called prior to the `fundChannel` api call.

This PR also:
- adds a unit test for depositing into a channel with 2 asset holder allocations.
- a stress test that creates 100 deposits (mostly to try to reproduce transaction nonce errors).
- adds a transaction queue to avoid nonce errors.

### DIFF
--- a/packages/server-wallet/package.json
+++ b/packages/server-wallet/package.json
@@ -75,7 +75,6 @@
   },
   "main": "lib/src/index.js",
   "scripts": {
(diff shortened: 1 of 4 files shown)


lalexgap (approved):
Looks good to me

PR #2654, opened 2020-10-06
ci: Add publish and publish-canary workflows

geoknee:
Mostly copy-paste from the graph monorepo. 

Questions:
---
- We are not doing conventional commits very strictly in this repo. #2617 would help with that?
- Do we keep with lockstep versioning? Will these actions work with that?

### DIFF
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -1,0 +1,39 @@
+name: Publish Packages (canary)
+
+on:
+  workflow_dispatch:
+
+jobs:
+  canary-publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2
+        with:
+          # pulls all commits and tags (needed for lerna / semantic release to correctly version)
+          fetch-depth: 0
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.16.3
+
+      - uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-yarn-cache-v0-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-cache-v0-
+
+      - name: yarn install
+        # skip the prepare step here, as lerna publish will run prepare before publishing
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
+          SKIP_PREPARE=true yarn install
+
+      - name: Configure CI Git User
+        run: |
+          git config --global user.name "github-actions"
+          git config --global user.email "github-actions@users.noreply.github.com"
+
+      - name: publish packages to npm
+        run: yarn run lerna publish --canary --force-publish --preid next.$(git rev-parse --short HEAD) --dist-tag next --yes
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,55 @@
+name: Publish Packages
+
+on:
+  workflow_dispatch:
+
+jobs:
+  publish-to-npm:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2
+        with:
+          # pulls all commits and tags (needed for lerna / semantic release to correctly version)
+          fetch-depth: 0
+
+      - name: Check that we're on master
+        if: github.ref != 'refs/heads/master'
+        run: exit 1
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.16.3
+
+      - uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-yarn-cache-v0-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-canary.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-yarn-cache-v0-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-cache-v0-
+
+      - name: yarn install
+        # skip the prepare step here, as lerna publish will run prepare before publishing
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
```
geoknee, line 30: I guess this needs to be configured as a GH secret.

.github/workflows/publish-canary.yml
```diff
@@ -0,0 +1,39 @@
...
+          SKIP_PREPARE=true yarn install
+
+      - name: Configure CI Git User
+        run: |
+          git config --global user.name "github-actions"
+          git config --global user.email "github-actions@users.noreply.github.com"
+
+      - name: publish packages to npm
+        run: yarn run lerna publish --canary --force-publish --preid next.$(git rev-parse --short HEAD) --dist-tag next --yes
```
geoknee, line 39: We never used the `next` tag, and only used `alpha` or `beta` in the `server-wallet` package.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,55 @@
...
+          SKIP_PREPARE=true yarn install
+
+      - name: Configure CI Git User
+        run: |
+          git config --global user.name "github-actions"
+          git config --global user.email "github-actions@users.noreply.github.com"
+
+      - name: publish packages to npm
+        run: |
+          yarn run lerna version --conventional-commits --yes --exact --no-push --message "Publish [skip ci]"
```
geoknee, line 44: This is untested so far. Do the flags make sense for this repo?

tomclose: 1. We'd have to start using conventional commits if we want this to work properly
2. We either want `exact` or we want to do proper semver (otherwise you risk yarn upgrading to a version that should be compatible but isn't). My feeling is that sticking with `exact` might be the right move while things are still changing a lot.

geoknee: I did 
```shell
yarn run lerna version --conventional-commits --yes --exact --no-push --no-git-tag-version --message "Publish [skip ci]"
```
as a sort of "[dry run](https://github.com/lerna/lerna/issues/51)" test.  It generated: 
```
 - benchmarking: 0.3.10 => 0.3.11 (private)
 - @statechannels/channel-client: 0.3.10 => 0.3.11
 - @statechannels/client-api-schema: 0.3.10 => 0.3.11
 - @statechannels/devtools: 0.3.10 => 0.3.11
 - docs-website: 0.3.10 => 0.3.11
 - @statechannels/iframe-channel-provider: 0.3.10 => 0.3.11
 - @statechannels/integration-tests: 0.3.10 => 0.3.11 (private)
 - @statechannels/jest-gas-reporter: 0.3.10 => 0.3.11
 - @statechannels/nitro-protocol: 0.3.10 => 0.3.11
 - @statechannels/server-wallet: 0.3.10 => 0.3.11
 - @statechannels/wallet-core: 0.3.10 => 0.3.11
 - @statechannels/wire-format: 0.3.10 => 0.3.11
 - @statechannels/xstate-wallet: 0.3.10 => 0.3.11
```
and pinned exact versions as expected. It also gave me a changelog in each package that all looked like this 

```txt
# Change Log

All notable changes to this project will be documented in this file.
See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.

## [0.3.11](https://github.com/statechannels/monorepo/compare/v0.3.10...v0.3.11) (2020-10-06)

**Note:** Version bump only for package @statechannels/nitro-protocol

```

I then added a commit like `fix!: blah` which should register a breaking change. I ran the command above again and no difference. I also tried `fix!(iframe-channel-provider): blah`. Am I missing something @tomclose ?

I also tried changing version to 'independent' in `lerna.json`. That resulted in many more sections in each changelog although they were mostly empty (I guess because we don't do a lot of conventional commits in this repo yet). I still didn't see my breaking change, though.

geoknee:
#2655

.github/workflows/publish.yml
tomclose: The `!` isn't actually how you do a breaking change. To record a breaking change you need to put `BREAKING_CHANGE: ... ` in the commit message (see [here](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer)). The `!` is just an optional convenience to draw attention to the breaking change.

tomclose (approved):

.github/workflows/publish.yml
geoknee: From https://www.conventionalcommits.org/en/v1.0.0/: 
> BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with MAJOR in semantic versioning). A BREAKING CHANGE can be part of commits of any type.

🤷‍♂️

Issue #2659, opened 2020-10-08
[server-wallet] wasm error on createChannel


geoknee:
If I pass `0x0` or `0x000`as `token` or `appData` in `createChannel`, i get:

```
    RuntimeError: unreachable
        at wasm-function[241]:0x44a8e
        at wasm-function[318]:0x46f69
        at wasm-function[336]:0x4731f
        at wasm-function[277]:0x46155
        at wasm-function[181]:0x40af2
        at wasm-function[346]:0x47495
```

If I pass `0x00` it works ok.

geoknee:
https://github.com/statechannels/native-utils/pull/8

PR #2660, opened 2020-10-08
[server-wallet] Add a 2x Wallet createChannel test

geoknee:
Preparatory work towards ledger-funding. Test can cover some or all funding strategies.

- Test fails because 'status' does not update as expected
- Apparently no difference in behaviour when passing a different funding strategy
- test is (almost) self contained with the help of a bit of `execSync` and a new `DBAdmin().migrate()` helper. It just needs a postgres server running
- generated #2659


packages/server-wallet/src/__test__/create-channel-2-wallets.test.ts
```diff
@@ -0,0 +1,125 @@
...
+});
+afterAll(async () => {
+  await Promise.all([a.destroy(), b.destroy()]);
+  execSync('dropdb TEST_A $PSQL_ARGS');
+  execSync('dropdb TEST_B $PSQL_ARGS');
+});
+
+const testCases: FundingStrategy[] = ['Unfunded', 'Direct', 'Ledger', 'Virtual'];
+
+it.each(testCases)(
```
geoknee, line 33: We may (or may not) wish to assert differently depending on the funding strategy.

geoknee: @tomclose we discussed a 'fake' funding strategy. Perhaps this is already covered by 'Unfunded'?

snario:
Added 0329ee7 based on [this Notion strategy](https://www.notion.so/elfour/Server-wallet-thoughts-ba7c8177dae44544987e5956e8233901#3fd8d1a7b2064d328eb8efb10c2a963f). Doesn't include _rejecting_ objectives, and just uses an in-memory objectives store, but gets the test to pass.

snario:
Going to close this in favour of new smaller PRs

PR #2667, opened 2020-10-09
Remove Chain Service v0

kerzhner:
The code was kept around so that it could be referenced when implementing v1. Removing Chain Service v0 allows us to clean up `p-queue` and `evt` dependencies.


packages/server-wallet/package.json
```diff
@@ -92,7 +90,7 @@
     "start:shared-ganache": "NODE_ENV=development npx start-shared-ganache",
     "start:watch": "nodemon --watch 'src/**/*' -e ts,tsx --exec ts-node ./src/start-dev-server.ts | npx pino-pretty",
     "test": "yarn jest -c ./jest/jest.config.js --runInBand",
-    "test:chain": "RPC_ENDPOINT='http://localhost:8545' jest -c ./jest/jest.chain.config.js",
```
kerzhner, line 95: I don't believe there is a reason to hardcode `RPC_ENDPOINT` for this command.

andrewgordstewart (approved):

PR #2669, opened 2020-10-12
[server-wallet] Add Objectives to Wallet Store & Refactor Run Loop

snario:
This pull request adds an in-memory representation of "objectives" inside the store. Before this PR, an objective was an ephemeral object that was created and sent by one party to another and used loosely in handling by `pushMessage`. Now, an objective is stored in the creators store and then sent to the receiver who also stores a copy.

Because objectives are now stored, they can be useful. This PR changes the fundamental run-loop from:

- this: "iterate over a set of channel IDs and run them in the Application protocol"
- to this: "iterate over all approved objectives and run them in their respective protocol"

There are two objectives implemented; `"OpenChannel"` and `"CloseChannel"`.

An objective can be:
- pending: meaning it has not yet been approved
- approved: meaning it has not yet been fully executed
- rejected: meaning it won't be executed (unimplemented in this PR)
- succeeded: meaning it was fully executed
- failed: meaning it failed in execution after approval (unimplemented in this PR)

A protocol takes an approved objective and tries to make it succeed. It might take several iterations (e.g., when opening a channel you need to sign multiple states before it is "running" and you can consider the channel as "opened"). Because of this there is a new protocol action called `"CompleteObjective"` that the run loop treats as the marker of a successful objective execution.

**Some nuances:**
- `pushMessage` would have a  channel result added to `channelResults` by the run loop if there was a new state added to the store but no action necessary to take on it in the run loop. Because we are now objectives based, if there is no objective for a new state received, the run loop won't add it to the result, and so the `pushMessage` handler does it manually.
- Closing a channel used to be implicit; if you had an `isFinal` state then the Application protocol would simply close it automatically. Now, an objective is required. So, the closer will send a `CloseChannel` objective and the receiver's wallet will auto-approve it, thus now both wallets will try to close the channel.
- There is now an `approveObjective` method which is basically a  wrapper around `joinChannel` but slightly higher level. You could also call it "approve open channel request" perhaps.

**Note on protocol state**
Since there are more protocols now (open and close, not just "application"), there can be unique protocol states per objective. So for example ledger funding will be a new protocol and objective pair with its own state. Both open and close use the same protocol state so there is no change in this PR.

The run loop has some filter function in it to check "what channels are related to this objective?" which just checks `obj.data.targetChanelId` in this PR since both `OpenChannel` and `CloseChannel` use that same property name on the objective's data field. In the future we probably will need some kind of `getRelatedChannelsForObjective(o: Objective): ProtocolState` method like alluded to above.

## Major TODO Item

The objectives store is **in memory**. It should be a SQL table. Maybe an [associated entity](https://en.wikipedia.org/wiki/Associative_entity)?


packages/server-wallet/src/protocols/__test__/close-channel.test.ts
```diff
@@ -0,0 +1,44 @@
...
+test.each`
+  supported        | latestSignedByMe | latest           | action                      | cond                                                                  | result
+  ${closingState}  | ${postFundState} | ${closingState}  | ${signState(closingState)}  | ${'when the postfund state is supported, and the channel is closing'} | ${'signs the final state'}
+  ${closingState2} | ${runningState}  | ${closingState2} | ${signState(closingState2)} | ${'when the postfund state is supported, and the channel is closing'} | ${'signs the final state'}
+`('$result $cond', ({supported, latest, latestSignedByMe, action}) => {
+  const ps = applicationProtocolState({app: {supported, latest, latestSignedByMe}});
+  expect(protocol(ps)).toMatchObject(action);
+});
+
+// todo: add the following test case once https://github.com/statechannels/the-graph/issues/80 is resolved
```
geoknee, line 33: That issue is closed -- does this mean that it is resolved?

snario: #2671

packages/server-wallet/src/wallet/index.ts
```diff
@@ -451,13 +518,35 @@ export class Wallet implements WalletInterface, ChainEventSubscriberInterface {
...
+    const objectives = Object.values(this.store.objectives).filter(
+      objective =>
+        // Only supports these two
+        (objective.type === 'OpenChannel' || objective.type === 'CloseChannel') &&
+        // Only runs on those with relevant channels
+        _.includes(channels, objective.data.targetChannelId) &&
+        // Only runs on pending or approved
+        (objective.status === 'approved' || // Need approved b.c. next action to take
+          objective.status === 'pending') /* Need pending because you want new channel result */
+    );
```
snario, line 534: This is the bit of code which asks the store "which objectives in the store are relevant for the run loop?"

packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
```diff
@@ -205,6 +216,15 @@ describe('when the application protocol returns an action', () => {
     const finalState = {...state, isFinal: true, turnNum: turnNum + 1};
     const p = wallet.pushMessage({
       signedStates: [serializeState(stateSignedBy([bob()])(finalState))],
+      objectives: [
+        {
+          type: 'CloseChannel',
+          participants: [],
```
snario, line 222: For some reason `participants` is required in `Objective` but I don't think it needs to be

packages/server-wallet/src/wallet/index.ts
```diff
@@ -300,6 +335,15 @@ export class Wallet implements WalletInterface, ChainEventSubscriberInterface {
       });
     };
 
+    // FIXME: This is just to get existing joinChannel API pattern to keep working
+    /* eslint-disable-next-line */
+    const {objectiveId} = _.find(
+      this.store.objectives,
+      o => o.type === 'OpenChannel' && o.data.targetChannelId === channelId
+    )!;
```
snario, line 343: We expect an objective to exist because `pushMessage` would have stored it earlier. In the test cases which do not use the public API we manually insert the objective into the store, too.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -364,14 +408,29 @@ export class Wallet implements WalletInterface, ChainEventSubscriberInterface {
...
     };
-    const criticalCode: AppHandler<SingleChannelResult> = async (tx, channel) => {
-      const nextState = getOrThrow(CloseChannel.closeChannel(channel));
-      const {outgoing, channelResult} = await this.store.signState(channelId, nextState, tx);
 
-      return {outbox: outgoing.map(n => n.notice), channelResult};
+    const criticalCode: AppHandler<void> = async (tx, channel) => {
+      // TODO: (Objectives Rewrite) Keeping this here b/c we need to do input validation
+      // and check if its our turn and throw an error as existing tests expect
+      getOrThrow(CloseChannel.closeChannel(channel));
```
snario, line 415: This thing generates a "sign state action" which is unnecessary. We just need it for request validation cc @kerzhner

packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
geoknee: That field is used in the xstate wallet, so removing it from `wire-format` would be breaking.

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -0,0 +1,40 @@
...
+      isFinal: true,
+    })) ||
+    // Or, countersign a given final state
+    (isFinal(ps.app.latest) &&
+      !isFinal(ps.app.latestSignedByMe) &&
+      signState({
+        channelId: ps.app.channelId,
+        ...ps.app.supported,
+        turnNum: ps.app.latest.turnNum,
+      })));
```
snario, line 32: I think this logic could probably be simpler

snario: https://github.com/statechannels/statechannels/pull/2718

tomclose (approved):
Have verified that this works with the the-graph: https://github.com/statechannels/the-graph/pull/240

andrewgordstewart:
🎉

PR #2670, opened 2020-10-12
[server-wallet] Add 'Unfunded' funding strategy support

snario:


packages/server-wallet/src/__test__/create-channel/fake-funding.test.ts
```diff
@@ -49,7 +49,7 @@ it('Create a fake-funded channel between two wallets ', async () => {
     allocations: [allocation],
     appDefinition: ethers.constants.AddressZero,
     appData: '0x00', // must be even length
-    fundingStrategy: 'Direct',
+    fundingStrategy: 'Unfunded',
```
geoknee, line 52: The name of the test file  is 'fake funding'. I think we decided that 'unfunded' should be deprecated terminology?

packages/server-wallet/src/__test__/create-channel/fake-funding.test.ts
```diff
@@ -81,46 +81,21 @@ it('Create a fake-funded channel between two wallets ', async () => {
     turnNum: 0,
   });
 
-  // B sends countersigned PreFund0 to A
+  // B sends countersigned PreFund0 to A _and_ PostFund03(since it is unfunded)
```
geoknee, line 84: Is this right? I think A hasn't signed `PostFund3` yet? Can we keep the fake funded case as close as possible to the direct funded case?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -200,37 +200,22 @@ export class Store {
...
-        objectives: [
-          {
-            participants: channel.participants,
-            type: 'OpenChannel',
-            data: {
-              targetChannelId: channel.channelId,
-              fundingStrategy: 'Direct',
-            },
-          },
-        ],
```
geoknee, line 224: What's the reason for removing this objective here?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -454,6 +439,17 @@ export class Store {
...
+      const objective: Objective = {
+        participants: constants.participants,
+        type: 'OpenChannel',
+        data: {
+          targetChannelId: channelId,
+          fundingStrategy,
+        },
+      };
+      // TODO: Message serialization should not be inside store.signState
+      ((outgoing[0].notice.params.data as unknown) as Payload).objectives = [objective];
```
geoknee, line 451: Do we need 2x `as` in this line?

packages/server-wallet/src/__test__/create-channel/fake-funding.test.ts
snario: Yeah I figure we could rename it to FakeFunded in a separate PR?

packages/server-wallet/src/__test__/create-channel/fake-funding.test.ts
snario: I think it is right, yes.

> Can we keep the fake funded case as close as possible to the direct funded case?

This depends, why would we do that? Direct funding requires the channel to be funded, and so you need to call `updateFunding` on the wallet. Fake funding does not require the channel to be funded, so you can automatically sign the post fund setup without delay. This seems like the expected behaviour

geoknee: I think we might do so if fake funding is really a sort of developer tool that allows funding to be skipped, but otherwise makes the wallet behave as similarly as possible, with the idea that fake funded channels most likely get upgraded to funded channels with minimal fuss later on. I don't know of any other use case for a fake funded channel.

packages/server-wallet/src/wallet/store.ts
snario: We do not have access to `fundingStrategy` within the `signState` method, but we do within `createChannel`. Also, this bit of code here is within a conditional expression that checks if `turnNum === 0`, which is hyper specific and is only needed within the context of `createChannel`. So the objective gets created and sent within `createChannel`.

packages/server-wallet/src/__test__/create-channel/fake-funding.test.ts
snario: I guess I don't understand what `Unfunded` means, then. Using `Direct` and calling `updateChannelFunding` without actually doing any funding, or running a blockchain at all, would be the same thing as what you're describing.

geoknee: I guess that is the current way we do "fake funding". I think we ought to deprecate `updateChannelFunding` entirely (eventually), since presumably the chain watcher will provide this information. Then, we can keep fake funding functionality by passing in "strategy:fake" and have the wallet auto-skip waiting for funding events. Which is exactly what you have here, up to the actual name you're using. And the other test in this dir can eventually have the `updateChannelFunding` calls removed when we are ready to switch to the chain watcher?

snario: > since presumably the chain watcher will provide this information

The chain watcher will / already provides the information by calling `updateChannelFunding`, right @kerzhner ?

kerzhner: The chain service does *not* use `updateChanneFunding` at the moment. When the wallet registers for chain events, the wallet passes in an object that implements the [`ChainEventSubscriberInterface`](https://github.com/statechannels/statechannels/blob/e7f4098be07fce36cc6f8115f25e32d4d1668a80/packages/server-wallet/src/chain-service/chain-service.ts#L42). Currently, the wallet implements that interface. The chain service will then call `onHoldingUpdated` method on the object when a `Deposit` event should be relayed to the wallet. `onHoldingUpdated` has a slightly different signature than `updateChannelFunding`.

kerzhner: Though it looks like `onHoldingUpdated` calls `updateChannelFunding` (incorrectly) at the moment: https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/wallet/index.ts#L519

packages/server-wallet/src/__test__/create-channel/fake-funding.test.ts
tomclose: I agree that “fake-funded” is better in this case. To me “unfunded” sounds a lot like, “not funded (yet/anymore)” rather than “it’s fine to progress, as we’re never going to fund it”.

geoknee (approved):

PR #2674, opened 2020-10-13
style: Remove type which is an alias for a Promise

snario:


geoknee (approved):
👍 👍 👍

tomclose (approved):
🎉 

Wondering if `MultipleChannelOutput` might be a better name than `MultipleChannelMessage`?

PR #2683, opened 2020-10-13
server-wallet: increase turn numbers for postfund setup

kerzhner:
Before this PR, every participant signs postfund state with turn number 3. The challenges with this are:
- https://www.notion.so/Server-wallet-application-Bob-Alice-communication-c0127c3196694e14bd0cf4858955fa96#032978ba91394e87a9774c94fb1222a0.
- Only works for a 2 party channel. That's ok for now.
- Until every participant signs the postfund state, there is no supported postfund state. Maybe that's not a big deal.

With this PR, turn numbers increase for postfund states.
- Alice signs postfund turn 2
- Bob signs postfund turn 3.

### DIFF
--- a/packages/server-wallet/src/__test__/create-channel/direct-funding.test.ts
+++ b/packages/server-wallet/src/__test__/create-channel/direct-funding.test.ts
@@ -119,7 +119,6 @@ it('Create a fake-funded channel between two wallets ', async () => {
 
   // < PostFund3B
   const resultA2 = await a.updateFundingForChannels([depositByB]);
-  // PostFund3A >
   const resultB2 = await b.updateFundingForChannels([depositByB]);
 
   expect(getChannelResultFor(channelId, resultA2.channelResults)).toMatchObject({
--- a/packages/server-wallet/src/__test__/create-channel/direct-funding.test.ts
+++ b/packages/server-wallet/src/__test__/create-channel/direct-funding.test.ts
@@ -38,8 +38,14 @@ it('Create a fake-funded channel between two wallets ', async () => {
 
   const allocation: Allocation = {
     allocationItems: [
-      {destination: participantA.destination, amount: BigNumber.from(1).toHexString()},
-      {destination: participantB.destination, amount: BigNumber.from(1).toHexString()},
+      {
+        destination: participantA.destination,
+        amount: BigNumber.from(1).toHexString(),
+      },
+      {
+        destination: participantB.destination,
+        amount: BigNumber.from(1).toHexString(),
+      },
     ],
     token: '0x00', // must be even length
   };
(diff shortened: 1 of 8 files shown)


packages/server-wallet/src/__test__/create-channel/direct-funding.test.ts
```diff
@@ -38,8 +38,14 @@ it('Create a fake-funded channel between two wallets ', async () => {
 
   const allocation: Allocation = {
     allocationItems: [
-      {destination: participantA.destination, amount: BigNumber.from(1).toHexString()},
```
kerzhner, line 41: There seem to be prettier configuration differences between my local setup and @snario's. And both pass CI?

andrewgordstewart: Either before or after is valid when opening this file from VSCode on my machine, both when opening from the `monorepo` root as well as from the `packages/server-wallet` package.

This is also valid:
```
      {destination: participantA.destination, amount: BigNumber.from(1).toHexString()},
      {
        destination: participantB.destination,
        amount: BigNumber.from(1).toHexString(),
      },
```

This is one of the only times where prettier doesn't choose a unique representation of a bit of code. As a sanity check, you didn't add any newline characters and then formatted, did you?

packages/server-wallet/src/protocols/application.ts
```diff
@@ -29,6 +29,25 @@ const isFunded = ({app: {funding, supported}}: ProtocolState): boolean => {
...
+// At the time of implementation, all particiapants sign turn 0 as prefund state
+// This function should also work with prefund state with increasing turn numbers.
+function myTurnToPostfund({app}: ProtocolState): boolean {
+  // I am the first participant
+  if (isPrefundSetup(app.supported) && isPrefundSetup(app.latestSignedByMe) && app.myIndex === 0) {
+    return true;
+  }
+
+  // I am NOT the first participant
+  // todo: this is not correct when there are more than 2 participants as we do not check that EVERY participant
```
andrewgordstewart, line 41: Create an issue for this?

andrewgordstewart (approved):

packages/server-wallet/src/protocols/application.ts
```diff
@@ -72,10 +91,15 @@ const fundChannel = (ps: ProtocolState): ProtocolResult | false =>
   requestFundChannelIfMyTurn(ps);
 
 const signPostFundSetup = (ps: ProtocolState): ProtocolResult | false =>
-  isPrefundSetup(ps.app.supported) &&
-  isPrefundSetup(ps.app.latestSignedByMe) &&
+  myTurnToPostfund(ps) &&
```
snario, line 94: Capitalize "f"?

snario:
This is a breaking change I believe, so we probably ought to use a conventional commit to indicate that.

packages/server-wallet/src/__test__/create-channel/direct-funding.test.ts
kerzhner: That is good to know. There is a chance I:
1. edited the value of `allocation` and saved.
1. edited the value of `allocation to match original and saved.

packages/server-wallet/src/protocols/application.ts
kerzhner: https://github.com/statechannels/statechannels/issues/2684

packages/server-wallet/src/protocols/application.ts
kerzhner: It seems like the file follows the convention of lower case "f" in prefund and postfund naming. So I did this instead a06a80693

kerzhner:
> This is a breaking change I believe, so we probably ought to use a conventional commit to indicate that.

It _might_ be a breaking change, though I am not sure. But given that, I agree with you.

That being said, it is a bit unfortunate that changing the internal wallet implementation of how channels are opened might result in a breaking change.

PR #2685, opened 2020-10-14
Direct funding works e2e

kerzhner:
This PR:
- Creates an e2e test for opening a directly funded channel.
- Wires up the chain service in the wallet so that the test passes.


packages/server-wallet/jest/chain-setup.ts
```diff
@@ -6,6 +6,10 @@ import {defaultConfig} from '../src/config';
 import {deploy} from '../deployment/deploy';
 
 export default async function setup(): Promise<void> {
+  process.env['GANACHE_HOST'] = '0.0.0.0';
+  process.env['GANACHE_PORT'] = '8545';
+  process.env['RPC_ENDPOINT'] = `http://${process.env['GANACHE_HOST']}:${process.env['GANACHE_PORT']}`;
```
kerzhner, line 11: Why are these environment variables defined here, a reviewer such as yourself might ask.

The wallet instantiates a chain service if an rpc endpoint and a chain private key are defined. For non-chain unit tests, ganache is not running. So the chain service instantiation will fail if these environment variables are defined (since default wallet configuration is populated from environment variables).

Some alternative approaches are be:
- Use different env var files for tests vs chain tests.
- Add a configuration to the wallet that determines whether or not a chain service should be instantiated.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -61,6 +62,10 @@ type Message = SingleChannelOutput | MultipleChannelOutput;
 const isSingleChannelMessage = (message: Message): message is SingleChannelOutput =>
   'channelResult' in message;
 
+export interface NotificationReceiver {
+  onWalletNotification(message: SingleChannelOutput): void;
+}
```
tomclose, line 67: This looks suspiciously like an event emitter. I'm wondering if it would be worth using an event emitter (e.g eventemitter3, like we do in the channel provider and xstate wallet). We could then expose `onNotification` (or event `onFundingUpdated`) on the wallet api, and any receivers could subscribe to it themselves. It seems like that would give us a load of goodies that we might need in the future (unsubscribe, multiple-subscribers etc.) out-of-the-box.

packages/server-wallet/src/__chain-test__/direct-funding.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    const aNotificationReceiver: NotificationReceiver = {
+      onWalletNotification(message: SingleChannelOutput) {
+        if (counter > 0) {
+          resolve(message);
+        }
+        counter++;
+      },
+    };
+    a.notificationReceiver = aNotificationReceiver;
+  });
```
tomclose, line 74: If you decided to go with the event emitter approach below, this would become a lot simpler. You could just do something like:
```ts
await new Promise(resolve => a.onNotification((channelResult) => resolve(channelResult))) // first notification
await new Promise(resolve => a.onNotification((channelResult) => resolve(channelResult))) // second notification
```

packages/server-wallet/src/__chain-test__/direct-funding.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  await a.dbAdmin().createDB();
+  await b.dbAdmin().createDB();
+  await Promise.all([a.dbAdmin().migrateDB(), b.dbAdmin().migrateDB()]);
+});
+afterAll(async () => {
+  await Promise.all([a.destroy(), b.destroy()]);
+  await a.dbAdmin().dropDB();
+  await b.dbAdmin().dropDB();
+  a.destroy();
+  b.destroy();
```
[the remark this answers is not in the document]
kerzhner: Good catch d35daf5ce

packages/server-wallet/src/__chain-test__/direct-funding.test.ts
kerzhner: I wonder if it is possible to miss the second notification with this approach. I am not sure.

Is it a concern that the second notification would fire before the second event listener is registered?

kerzhner: Maybe that's not a concern as ["The EventEmitter calls all listeners synchronously"](https://nodejs.org/api/events.html#events_asynchronous_vs_synchronous)

packages/server-wallet/src/wallet/index.ts
kerzhner: I like the event emitter idea. Check out 2826a7021

geoknee:
Just a heads up @kerzhner if you really want a 🥕 you'll need to have shorter body lines in your commits 😉 

> body's lines must not be longer than 100 characters [body-max-line-length]

tomclose (approved):

kerzhner:
> Just a heads up @kerzhner if you really want a 🥕 you'll need to have shorter body lines in your commits 😉
> 
> > body's lines must not be longer than 100 characters [body-max-line-length]

I wouldn't mind if we increased the line length limit to longer than 100 characters if others are not opposed to that.

tomclose:
I'm guessing that the 100 character line limit exists for the purpose of making changelogs look good?

PR #2689, opened 2020-10-14
[server-wallet] Allow registering of bytecode

lalexgap:
Exposes a `RegisterAppDefinition` method on the `wallet` that registers the `bytecode` for that `AppDefiniton`. 

This new method uses the `chainService` to fetch the bytecode and then registers it in the database.

Currently the bytecode is overriden if `RegisterAppDefinition` is called twice to play nice with `ganache`. Once we start actually using `ChainId` we may want to throw an error if there already exists bytecode for a given definition.

### DIFF
--- a/packages/server-wallet/src/models/app-bytecode.ts
+++ b/packages/server-wallet/src/models/app-bytecode.ts
@@ -23,18 +23,17 @@ export class AppBytecode extends Model implements RequiredColumns {
     appDefinition: Address,
     appBytecode: Bytes32,
     txOrKnex: TransactionOrKnex
-  ): Promise<AppBytecode> {
-    const insert = (
-      await txOrKnex(this.tableName).insert({chainId, appDefinition, appBytecode})
-    ).toString();
-    const update = txOrKnex(this.tableName)
-      .update({appBytecode})
-      .where({chainId, appDefinition});
-
-    return txOrKnex.raw(`? ON CONFLICT (chain_id,app_definition) DO ? returning *`, [
-      insert,
-      update,
-    ]);
+  ): Promise<void> {
+    await txOrKnex.raw(
+      `INSERT INTO ${this.tableName} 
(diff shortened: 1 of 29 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -142,6 +147,15 @@ export class Wallet implements WalletInterface, ChainEventSubscriberInterface {
     this.chainService = new MockChainService();
   }
 
+  public async registerAppDefintion(appDefinition: string): Promise<void> {
```
kerzhner, line 150: Is it worth writing a chain unit test for this function?

packages/server-wallet/src/models/app-bytecode.ts
```diff
@@ -18,6 +18,25 @@ export class AppBytecode extends Model implements RequiredColumns {
...
+    txOrKnex: TransactionOrKnex
+  ): Promise<AppBytecode> {
+    const insert = (
+      await txOrKnex(this.tableName).insert({chainId, appDefinition, appBytecode})
+    ).toString();
+    const update = txOrKnex(this.tableName)
+      .update({appBytecode})
+      .where({chainId, appDefinition});
+
+    return txOrKnex.raw(`? ON CONFLICT (chain_id,app_definition) DO ? returning *`, [
```
kerzhner, line 34: Is it worth having a unit test verify that the on conflict clause works as intended?

kerzhner (approved):

packages/server-wallet/src/wallet/index.ts
lalexgap: Good point! Done here: f36ac33

packages/server-wallet/src/models/app-bytecode.ts
lalexgap: Good point! Done here: 334155f

PR #2692, opened 2020-10-15
refactor: defaultConfig -> processEnvConfig

geoknee:
A more accurate name?

⚠️  base not master


geoknee:
May reopen and rebase when things are quieter!

PR #2694, opened 2020-10-15
Store objectives in the DB and introduce crash tolerance test

geoknee:
- Adds two new database tables, to support open and close channel objectives (separately). 

- Adds a test that would fail on the base branch, where one wallet crashes and restarts (and passes on this branch):
Screenshot 2020-10-16 at 11 22 10

TODO (future PRs?) 
- [ ] optimize / DRY out new db schemas
- [ ] review db constraints
- [ ] review objective id scheme / make it unique

### DIFF
--- a/packages/server-wallet/src/__test__/crash-tolerance/crash-tolerance.test.ts
+++ b/packages/server-wallet/src/__test__/crash-tolerance/crash-tolerance.test.ts
@@ -1,0 +1,194 @@
+import {
+  CreateChannelParams,
+  Participant,
+  Allocation,
+  CloseChannelParams,
+} from '@statechannels/client-api-schema';
+import {makeDestination} from '@statechannels/wallet-core';
+import {BigNumber, ethers} from 'ethers';
+
+import {defaultConfig as processEnvConfig} from '../../config';
+import {Wallet} from '../../wallet';
+import {getChannelResultFor, getPayloadFor} from '../test-helpers';
+
+const a = new Wallet({...processEnvConfig, postgresDBName: 'TEST_A'});
+const b = new Wallet({...processEnvConfig, postgresDBName: 'TEST_B'}); // Wallet that will "crash"
+const b2 = new Wallet({...processEnvConfig, postgresDBName: 'TEST_B'}); // Wallet that will "restart" (same db)
+// TODO needs to have same signing address as b? Pass in thru config
+
+let channelId: string;
+let participantA: Participant;
+let participantB: Participant;
+
+beforeAll(async () => {
+  await a.dbAdmin().createDB();
+  await b.dbAdmin().createDB();
+  await Promise.all([a.dbAdmin().migrateDB(), b.dbAdmin().migrateDB()]);
+
+  participantA = {
+    signingAddress: await a.getSigningAddress(),
+    participantId: 'a',
+    destination: makeDestination(
+      '0xaaaa000000000000000000000000000000000000000000000000000000000001'
+    ),
+  };
+  participantB = {
+    signingAddress: await b.getSigningAddress(),
+    participantId: 'b',
+    destination: makeDestination(
+      '0xbbbb000000000000000000000000000000000000000000000000000000000002'
+    ),
+  };
+});
+afterAll(async () => {
+  // Don't destory b here, it get's destroyed midway through test
+  await Promise.all([a.destroy(), b2.destroy()]);
+  await a.dbAdmin().dropDB();
+  await b2.dbAdmin().dropDB(); // Only need to drop the db once as it is shared with b
+});
+
+it.only('Create a fake-funded channel between two wallets, of which one crashes midway through ', async () => {
+  const allocation: Allocation = {
+    allocationItems: [
+      {destination: participantA.destination, amount: BigNumber.from(1).toHexString()},
+      {destination: participantB.destination, amount: BigNumber.from(1).toHexString()},
+    ],
+    token: '0x00', // must be even length
+  };
+
+  const createChannelParams: CreateChannelParams = {
+    participants: [participantA, participantB],
+    allocations: [allocation],
+    appDefinition: ethers.constants.AddressZero,
+    appData: '0x00', // must be even length
+    fundingStrategy: 'Direct',
+  };
+
+  //        A <> B
+  // PreFund0
+  const resultA0 = await a.createChannel(createChannelParams);
+
+  // TODO compute the channelId for a better test
+  channelId = resultA0.channelResults[0].channelId;
+
+  expect(getChannelResultFor(channelId, resultA0.channelResults)).toMatchObject({
+    status: 'opening',
+    turnNum: 0,
+  });
+
+  //    > PreFund0A
+  const resultB0 = await b.pushMessage(getPayloadFor(participantB.participantId, resultA0.outbox));
+
+  expect(getChannelResultFor(channelId, resultB0.channelResults)).toMatchObject({
+    status: 'proposed',
+    turnNum: 0,
+  });
+
+  // Destory Wallet b and restart Wallet b2
+  await b.destroy();
+
+  //      PreFund0B
+  const resultB1 = await b2.joinChannel({channelId});
+  expect(getChannelResultFor(channelId, [resultB1.channelResult])).toMatchObject({
+    status: 'opening',
+    turnNum: 0,
+  });
+
+  //  PreFund0B <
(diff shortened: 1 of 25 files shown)


packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
```diff
@@ -0,0 +1,17 @@
+import * as Knex from 'knex';
+
+const tableName = 'open-channel-objectives';
+
+export async function up(knex: Knex): Promise<any> {
+  return knex.schema.createTable(tableName, function(table) {
+    table.integer('objectiveId');
```
andrewgordstewart, line 7: This should be a primary key.

packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
```diff
@@ -0,0 +1,17 @@
...
+    table.string('status');
+    table.string('type');
+    table.string('targetChannelId');
```
andrewgordstewart, line 10: This should be a foreign key to the `channels` table

packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
```diff
@@ -0,0 +1,17 @@
+import * as Knex from 'knex';
+
+const tableName = 'open-channel-objectives';
+
+export async function up(knex: Knex): Promise<any> {
+  return knex.schema.createTable(tableName, function(table) {
```
andrewgordstewart, line 6: It seems like none of these columns should be allowed to have a `NULL` value? If so, they should be have a `NOT NULL` constraint.

packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
andrewgordstewart, line 10: It looks like this table should have a unique constraint on `targetChannelId`?

packages/server-wallet/src/db/migrations/20201014174040_close-channel-objectives.ts
```diff
@@ -0,0 +1,17 @@
+import * as Knex from 'knex';
+
+const tableName = 'close-channel-objectives';
+
+export async function up(knex: Knex): Promise<any> {
+  return knex.schema.createTable(tableName, function(table) {
```
andrewgordstewart, line 6: This table has identical schema to `open-channel-objectives`. What's the purpose of having two separate tables?

andrewgordstewart (commented):
Left some feedback on the schema change.

packages/server-wallet/src/db/migrations/20201014174040_close-channel-objectives.ts
geoknee: It's not exactly the same, although it is very similar. In general, the different objective types have `data` properties of differing shape.

packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
geoknee: I made that change, but rolled it back. We need to first make sure `objectiveId` is unique.

packages/server-wallet/src/__test__/crash-tolerance/crash-tolerance.test.ts
```diff
@@ -0,0 +1,194 @@
...
+  const resultB0 = await b.pushMessage(getPayloadFor(participantB.participantId, resultA0.outbox));
+
+  expect(getChannelResultFor(channelId, resultB0.channelResults)).toMatchObject({
+    status: 'proposed',
+    turnNum: 0,
+  });
+
+  // Destory Wallet b and restart Wallet b2
+  await b.destroy();
+  b = new Wallet({...processEnvConfig, postgresDBName: 'TEST_B'}); // Wallet that will "restart" (same db)
```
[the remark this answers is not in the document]
snario: Might be ambitious for this PR, but would it be possible to have the test designed such that it tests crashing at every possible interval between API method calls?

packages/server-wallet/src/__test__/create-and-close-channel/fake-funding.test.ts
```diff
@@ -86,7 +86,9 @@ it('Create a fake-funded channel between two wallets ', async () => {
   // after joinChannel, B double-signs PreFund0
   const bJoinChannelOutput = await b.approveObjective(
     // eslint-disable-next-line
-    bProposeChannelPushOutput.objectivesToApprove![0].objectiveId
+    bProposeChannelPushOutput.objectivesToApprove![0].objectiveId,
+    // eslint-disable-next-line
+    bProposeChannelPushOutput.objectivesToApprove![0].type
```
snario, line 91: Hm we may want to do some assertion on `objectivesToApprove` to avoid using `!`

geoknee: We could have a `getOpenChannelObjectiveFor(targetChannelId:string, singleChannelOutput: SingleChannelOutput)` test utility that throws if there isn't exactly 1 objective of the correct type. We then assert on the output of this function.

packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
andrewgordstewart: What makes it not unique? That suggests an error in the client code to me, but I don't know if I have the right context

packages/server-wallet/src/db/migrations/20201014174040_close-channel-objectives.ts
andrewgordstewart: That's on the client side though. The DB schema appears identical to me.

packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
geoknee: Yeah it's definitely a problem in the client code. We can probably just have the db choose an incrementing id for us (currently it is just a placeholder: the `channelNonce`, in fact).

packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
geoknee: Most likely yes, as long as the objective is never inserted before the target channel (which could feasibly happen during opening a channel). It could be a nullable foreign key.

packages/server-wallet/src/db/migrations/20201014174040_close-channel-objectives.ts
geoknee: Yeah you're right, but it was a typo 🥴 . The close channel objective shouldn't have a `fundingStrategy`.

packages/server-wallet/src/db/migrations/20201014174039_open-channel-objectives.ts
andrewgordstewart: That sounds like a client bug to me. What happens if the client crashes before inserting the channel? What happens if the client decides to delete a closed channel but forgot to delete the associated objective?

packages/server-wallet/src/db/migrations/20201014174040_close-channel-objectives.ts
andrewgordstewart: The funding strategy sounds like it belongs in the field to me.

geoknee: Hmm maybe. I was going off of the type recently added to `wallet-core` on the base branch 
https://github.com/statechannels/statechannels/blob/e0f75ef9236548f0fcc78db05bd58472c27292ea/packages/wallet-core/src/types.ts#L102-L106

packages/server-wallet/src/__test__/crash-tolerance/crash-tolerance.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    appData: '0x00', // must be even length
+    fundingStrategy: 'Direct',
+  };
+
+  //        A <> B
+  // PreFund0
+  const resultA0 = await a.createChannel(createChannelParams);
+
+  // TODO compute the channelId for a better test
+  channelId = resultA0.channelResults[0].channelId;
```
snario, line 70: We have this TODO in 3 or 4 files now... maybe we ought to either remove it, add an issue, or do it? :p

geoknee:
in favour of #2703

PR #2698, opened 2020-10-16
refactor: remove db-admin knex and truncate method

geoknee:
It is currently a bit confusing to have both a db-admin file and a db-admin-connection file, with each featuring a `truncate` method/function, as well as one exporting a Knex instance. 

In scenarios where there is no `Wallet` instance (in our tests, for example), we should use the `testKnex` instance which is currently configured via environment variables.  Otherwise we can use the `knex` instance that lives inside each `Wallet`, which is configured via the `Wallet` constructor.


packages/server-wallet/src/db-admin/db-admin.ts
```diff
@@ -24,7 +28,7 @@ export class DBAdmin {
   }
 
   async truncateDB(): Promise<void> {
-    const tables = ['signing_wallets', 'channels', 'nonces'];
+    const tables = [SigningWallet.tableName, Channel.tableName, Nonce.tableName];
```
geoknee, line 31: We also need to keep this list in sync with the files in the migrations directory. I think it is already behind those.

snario (approved):

PR #2701, opened 2020-10-19
[nitro-protocol] Enhance testing of ERC20 deposits and withdrawals

geoknee:
Headlines: 

`ERC20AssetHolder.deposit`:
```
Deposits Tokens (expectedHeld = 0):
76024 gas

Deposits Tokens (expectedHeld = 1):
46024 gas
```

Most expensive for first person to deposit, since they must create (instead of modify) a storage slot. Perhaps some interesting game theory there!

`concludePushOutcomeAndTransferAll`:
```
{ERC20: {A: 1, B: 1}}:
140121 gas

{ERC20: {At: 1, Bt: 1}} (At and Bt already have some TOK, so it's cheaper to pay out-- this is probably the norm, though):
110109 gas

10 TOK payouts:
406,543 gas

50 TOK payouts:
1,589,293 gas

100 TOK payouts:
2,656,645 gas

```

Context: that block gas limit of 8million.

### DIFF
--- a/packages/nitro-protocol/contracts/test/TestErc20AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/test/TestErc20AssetHolder.sol
@@ -1,6 +1,7 @@
 pragma solidity ^0.6.0;
 pragma experimental ABIEncoderV2;
 import '../ERC20AssetHolder.sol';
+import './TESTAssetHolder.sol';
 
 /**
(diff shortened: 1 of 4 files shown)


geoknee:
Might be more realistic / interesting to test the case where 98 of the 100 destinations are actually _internal_ (i.e. channel ids).

tomclose (approved):

PR #2702, opened 2020-10-19
refactor: log level of debug for mock chain service warning

kerzhner:
Most unit tests use a mock chain service. This reduces logging noise for unit tests.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -26769,7 +26769,7 @@ rxjs@6.5.5:
   dependencies:
     tslib "^1.9.0"
 
-rxjs@^6.3.3, rxjs@^6.4.0, rxjs@^6.5.3, rxjs@^6.5.4, rxjs@^6.6.0, rxjs@^6.6.2, rxjs@^6.6.3:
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -26769,7 +26769,7 @@ rxjs@6.5.5:
   dependencies:
     tslib "^1.9.0"
 
-rxjs@^6.3.3, rxjs@^6.4.0, rxjs@^6.5.3, rxjs@^6.5.4, rxjs@^6.6.0, rxjs@^6.6.2, rxjs@^6.6.3:
+rxjs@6.6.3, rxjs@^6.3.3, rxjs@^6.4.0, rxjs@^6.5.3, rxjs@^6.5.4, rxjs@^6.6.0, rxjs@^6.6.2:
```
geoknee, line 26772: Does this change belong?

geoknee (approved):

yarn.lock
kerzhner: I ran yarn, and that was the result. So I think yes.

Issue #2570, opened 2020-09-19
Ensure that chain_service migration can run on non-empty DBs


andrewgordstewart:
I think we will have to edit this migration so that it can run in existing wallets.

As is, the migration will fail if there are any existing rows in the `channels` table, since this migration
1. requires a non-null `funding_strategy` value
2. does not provide a default value

If we provided a default value, then we would have to either
A. take down existing wallets for maintenance; or
B. accept that existing wallets will experience a pause while this migration runs, considering that it will require a write-lock on the `channels` table.

_Originally posted by @andrewgordstewart in https://github.com/statechannels/statechannels/pull/2561#r491503249_

andrewgordstewart:
Related: https://github.com/statechannels/statechannels/pull/2605#discussion_r493756218

PR #2703, opened 2020-10-19
Store objectives in the DB and introduce crash tolerance test

geoknee:
Rewritten history version of #2694 

- single db table for all objective types
- objective ids (db table primary keys) computed (on insert) deterministically by `type`, and `data`. Implied that you can't reinsert the same objective for the same channel (say)


geoknee:
The crash tolerance test is now failing (after rebase) because no bytecode can be found after a wallet crash.

geoknee:
> The crash tolerance test is now failing (after rebase) because no bytecode can be found after a wallet crash.

Gotcha! 
https://github.com/statechannels/statechannels/blob/b592c01d01f9a9c0422474f6b9dcf1e1a034ea8b/packages/server-wallet/src/config.ts#L64-L65

Is this behaviour that we want? Might it be better to not fail the evm check if there is no bytecode (maybe just a warning)? We should fail *if* the app provides bytecode and the state transition check fails?

tomclose (approved):

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
```diff
@@ -0,0 +1,20 @@
+import * as Knex from 'knex';
+
+const tableName = 'objectives';
+
+export async function up(knex: Knex): Promise<any> {
+  return knex.schema.createTable(tableName, function(table) {
+    table.string('objective_id').primary();
+    table.string('status');
+    table.string('type');
```
andrewgordstewart, line 9: Can these be not nullable?

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
```diff
@@ -0,0 +1,20 @@
...
+    table.string('target_channel_id');
```
andrewgordstewart, line 10: Why is this not a foreign key?

andrewgordstewart (commented):
What was the argument for using a single table for different objective types?

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
geoknee: Yes #2708

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
geoknee: See #2708 . I guess it's not 100% clear to me that we want to enforce that constraint. I might want to store an objective that refers to a channel that I do not yet know about?

geoknee:
> What was the argument for using a single table for different objective types?

I took your arguments in the previous PR as an argument for fewer tables, given that the objectives have similar shapes. Fewer tables will probably lead to less duplication, and I don't think it will hurt us on the type safety side if we do things carefully enough? Don't feel super strongly though

andrewgordstewart:
> > What was the argument for using a single table for different objective types?
> 
> I took your arguments in the previous PR as an argument for fewer tables, given that the objectives have similar shapes. Fewer tables will probably lead to less duplication, and I don't think it will hurt us on the type safety side if we do things carefully enough? Don't feel super strongly though

In the previous PR, the two tables had identical schema (except for a mistaken column).

In this one, it is now possible to have an objective that has:
- type equal to `'OpenAndLedgerFund'` (or whatever it is to signify that meaning)
- a non-null `guarantor_id`

When feasible, it's desirable for your database schema to remove these kinds of possibilities. Nevertheless, in this case, it doesn't seem that harmful, and I don't think we have the time (or even the knowledge?) to define the "one true schema".

DB schema is usually worth additional scrutiny because it can be excessively difficult to change it in a production DB. It's also possible to write migrations that either [can or cannot run in a living DB](https://github.com/statechannels/statechannels/issues/2570), depending on the _content_ of the DB.

andrewgordstewart:
https://github.com/statechannels/statechannels/pull/2708/files#r508634858

This, for example, is suggestive that the current schema of the `objectives` is "not correct".

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
andrewgordstewart: Foreign key constraints are about _referential integrity_. If this objective refers to a channel, the constraint enforces that the channel exists.

If we want to enforce that, this should be a foreign key.

geoknee: Correct. I'm saying that I'm not sure we want to enforce that.

Anyways, we _will_ enforce that for the time being (current tests pass). We can revisit if needs be.

PR #2707, opened 2020-10-20
test: add 100 internal transfers case

geoknee:


geoknee:
I believe this was not merged due to the tests becoming prohibitively slow.

PR #2708, opened 2020-10-20
Tweaks to objective schema

geoknee:


packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
```diff
@@ -5,13 +5,18 @@ const tableName = 'objectives';
 export async function up(knex: Knex): Promise<any> {
   return knex.schema.createTable(tableName, function(table) {
     table.string('objective_id').primary();
-    table.string('status');
-    table.string('type');
+    table.string('status').notNullable();
+    table.string('type').notNullable();
     table.string('target_channel_id');
     table.string('funding_strategy');
```
andrewgordstewart, line 11: Should this be not nullable?

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
```diff
@@ -5,13 +5,18 @@ const tableName = 'objectives';
 export async function up(knex: Knex): Promise<any> {
   return knex.schema.createTable(tableName, function(table) {
     table.string('objective_id').primary();
-    table.string('status');
-    table.string('type');
+    table.string('status').notNullable();
+    table.string('type').notNullable();
     table.string('target_channel_id');
```
andrewgordstewart, line 10: My gut says that the table should have a unique constraint on `('type', `target_channel_id`)`. But I think that's taken care of by constructing the `objective_id` as a deterministic function of `('type', `target_channel_id`)`?

andrewgordstewart (approved):

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
geoknee: no

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
andrewgordstewart: If that's the case, you can consider using this strategy:
```
    table.primary(['target_channel_id', 'type']);
```
which is used here: https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/db/migrations/20200814010138_app_bytecode.ts#L11

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
geoknee: You want this to be null in the case of inserting an objective that does not have this property. 

Is your point that all objectives _should_ have this property? Currently, they do not all have this property.

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
geoknee: Yes I think I agree. I think rolling our own ids might help in coordinating objectives across participants' wallets.

tomclose: I don't think that objectives will necessarily all have a `target_channel_id`. An example would be the objective proposed here: https://www.notion.so/Bulk-ledger-fund-objective-bb05565337174fa88498fc98fb9e79e7

geoknee: Current approach to ids: 
https://github.com/statechannels/statechannels/blob/2783e044c9786cd8ac6e05ad3a3f33996cc72f1b/packages/wallet-core/src/types.ts#L153-L165

I think the bytecode case that you linked to @andrewgordstewart is different in that different wallets do not need to agree on the id: it is not used outside of the db context.

PR #2709, opened 2020-10-20
[server-wallet] Ensure Nonce gets bumped on reception of new signed states

snario:
The `Nonce` is currently not bumped on the receiving end and this can lead to issues when a party who has received many states decides to create a channel with an existing counterparty.

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -278,8 +278,13 @@ export class Store {
 
   async pushMessage(message: WirePayload): Promise<Bytes32[]> {
     return this.knex.transaction(async tx => {
+      // Sorted to ensure channel nonces arrive in ascending order
+      const sortedSignedStates = (message.signedStates || []).sort(
+        (a, b) => a.channelNonce - b.channelNonce
+      );
+
       const stateChannelIds = message.signedStates?.map(ss => ss.channelId) || [];
-      for (const ss of message.signedStates || []) {
+      for (const ss of sortedSignedStates || []) {
         await this.addSignedState(ss.channelId, undefined, ss, tx);
       }
 
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -341,9 +341,19 @@ export class Store {
         )
     );
 
+    maybeChannel =
+      maybeChannel || (await timer('get channel', async () => getChannel(channelId, tx)));
+
+    if (!maybeChannel) {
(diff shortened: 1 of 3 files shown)


packages/server-wallet/src/wallet/store.ts
```diff
@@ -341,9 +346,19 @@ export class Store {
...
 
+    maybeChannel =
+      maybeChannel || (await timer('get channel', async () => getChannel(channelId, tx)));
+
+    if (!maybeChannel) {
+      // TODO: Discuss where to put the code that bumps a channel nonce (is here ok?)
+      await Nonce.fromJson({
+        value: wireSignedState.channelNonce,
+        addresses: wireSignedState.participants.map(p => p.signingAddress),
+      }).use(tx);
```
andrewgordstewart, line 357: I think whether or not this succeeds depends on the order in which new states are put in the store, something that I'm not sure I thought about when writing `Nonce.use`. It might need some thought/work.

tomclose (changes requested):
Could you add a test to ensure that you can still push channels in any order. Something like:
```ts
const { outbox: outbox1 } = await a.createChannel({..})
const { outbox: outbox2 } = await a.createChannel({..})

await b.pushPayload(outbox2) // or whatever it needs to be
await b.pushPayload(outbox1)
```
and ensure it passes?

snario:
I realized that we need to keep track of more than once nonce. We need to keep track of all ranges of nonces we can consider as having been used. Ideally, this range is the minimal amount of information like "all less than 33", but it might need to be "all less than 23, 25 to 28, and 30 to 33" because no states exist for nonces 23, 24, 26, 29, 31, 32, 33 due to some concurrency / message ordering scenario. I think `Nonce` needs to be updated to include this ability.

tomclose:
One strategy that gets us most of the way there: just don't raise an error if you're receiving a new channel with a nonce that's lower than the maximum. That's fine - if there's a conflicting channel in the db, it will just be added to this channel, but we'll see it's an invalid transition.

The only problem with this approach is if the conflicting channel isn't in the db anymore. So we've cleared out channel with nonce 1, and then our opponent sends us a new channel with nonce 1. We could just not handle this case for the time being. In the longer term it seems like we could actually just maintain the largest nonce that we've deleted, and not accept channels under that. That would solve this state ordering problem, as you'd only delete old completed channels, which both participants agree are completed.

snario:
I have implemented the above suggestion in #2715 — a640d226c

tomclose (approved):

PR #2711, opened 2020-10-20
[server-wallet] Add test helper method

snario:


andrewgordstewart (approved):

packages/server-wallet/src/wallet/__test__/integration/join-channel.test.ts
```diff
@@ -40,25 +41,22 @@ describe('directly funded app', () => {
...
+      turnNum: 1,
     });
 
+    expect(getChannelResultFor(c2.channelId, channelResults)).toMatchObject({
+      channelId: c2.channelId,
+      turnNum: 1,
+    });
+
+    expect(getSignedStateFor(c1.channelId, outbox)).toMatchObject({...state1, turnNum: 1});
+    expect(getSignedStateFor(c2.channelId, outbox)).toMatchObject({...state2, turnNum: 1});
```
tomclose, line 58: I'm not sure whether we should be testing what states go across the wire. It seems like it's a bit of an implementation detail.

PR #2714, opened 2020-10-21
fix: enter run loop once for joinChannels API call

snario:


andrewgordstewart:
What is this fixing?

snario:
- With ledger channels a call to `joinChannel` might trigger a ledger channel update
- That ledger channel update should be generated at the end of the run loop for an API calls
- With the current implementation, it produces a single ledger channel update funding the first channel in `joinChannels`
- With this implementation, it produces a single ledger update funding all channels it joins

tomclose (approved):
I'm keen to refactor this shortly, but this solves the problem for now.

Potential refactor: it seems like this is a bit nasty because the wallet api method `joinChannels`, calls the wallet api method `joinChannel`. It seems that what it really should be doing is calling the wallet _operation_ `joinChannel`, which is the thing responsible for approving the objective and signing the next state. It seems like this separation would generally make things nicer: we would have operations that all look similar, and wallet api methods that trigger an operation, run the run loop, and return stuff.

snario:
@tomclose ^ https://github.com/statechannels/statechannels/pull/2717

PR #2718, opened 2020-10-21
refactor: simplify close channel logic

snario:


tomclose:
Could you give some context around this change and why it is required?

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -11,28 +11,17 @@ const stageGuard = (guardStage: Stage) => (s: State | undefined): s is State =>
...
-        turnNum: ps.app.latest.turnNum,
-      })));
+  !!ps.app.supported &&
+  !isFinal(ps.app.latestSignedByMe) &&
+  signState({
+    channelId: ps.app.channelId,
+    ...ps.app.supported,
+    turnNum: ps.app.supported.turnNum + (isFinal(ps.app.supported) ? 0 : 1),
+    isFinal: true,
+  });
```
snario, line 21: This allows both participants to sign the same states if they have decided to close a channel (one or both has created / sent a CloseChannel objective). It's written more concisely.

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -11,28 +11,17 @@ const stageGuard = (guardStage: Stage) => (s: State | undefined): s is State =>
...
 
 const completeCloseChannel = (ps: ProtocolState): CompleteObjective | false =>
-  isFinal(ps.app.supported) &&
+  (ps.app.support || []).every(isFinal) &&
```
snario, line 24: This is actually a bug fix I think — you should not consider a channel closed unless every state in the support is final.

andrewgordstewart: I remember making a similar change a couple of months ago. Is there now a test that detects this regression?

tomclose (approved):

PR #2719, opened 2020-10-21
refactor: add support for requesting too much funds from ledger channel

snario:


tomclose:
Could you explain what the overall behaviour that this PR seeks to accomplish? Is it aiming to completely solve the over-allocated ledger problem (coordinating a resolution between the two parties) or is it just making sure that the ledgers won't go negative?

snario:
It rejects sets the status of `OpenChannel`s to `rejected` for both parties for channels which do not have enough funding available in the ledger for them.

PR #2720, opened 2020-10-21
Replace individual columns with a single jsonb column in objectives db table, and introduce objectives_channels linker table

geoknee:
TODO
---
- [ ] ~reinstate `participants` property (should clean up the code quite a bit)~
- [ ] ~throw error on unsupported objectvies?~
- [x] debug and fix failing tests: currently
```
 Objective.relationMappings.objectivesChannels: join: either `from` or `to` must point to the owner model table and the other one to the related table. It might be that specified table 'channels' is not correct
```
- [x] debug and fix e2e tests (same error ^^)

### DIFF
--- a/packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
+++ b/packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
@@ -8,7 +8,7 @@ export async function up(knex: Knex): Promise<any> {
     table.string('objective_id').primary();
     table.string('status').notNullable();
     table.string('type').notNullable();
-    table.jsonb('data');
+    table.jsonb('data').notNullable();
   });
 
   await knex.schema.createTable(associativeTableName, function(table) {
@@ -17,7 +17,7 @@ export async function up(knex: Knex): Promise<any> {
       .string('channel_id')
       .references('channels.channel_id')
       .notNullable();
-    table.primary(['objective_id,channel_id']);
+    table.primary(['objective_id', 'channel_id']);
   });
 }
 
--- a/packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
+++ b/packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
@@ -7,16 +7,7 @@ export async function up(knex: Knex): Promise<any> {
     table.string('objective_id').primary();
     table.string('status').notNullable();
     table.string('type').notNullable();
-    table.string('target_channel_id');
-    table.string('funding_strategy');
-    table.string('ledger_id');
-    table.string('joint_channel_id');
-    table.string('guarantor_id');
-
-    table.foreign('target_channel_id').references('channels.channel_id');
-    table.foreign('ledger_id').references('channels.channel_id');
-    table.foreign('joint_channel_id').references('channels.channel_id');
-    table.foreign('guarantor_id').references('channels.channel_id');
+    table.jsonb('data');
   });
 }
 
--- a/packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
+++ b/packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
@@ -1,16 +1,27 @@
 import * as Knex from 'knex';
 
 const tableName = 'objectives';
+const associativeTableName = 'objectives_channels';
 
 export async function up(knex: Knex): Promise<any> {
-  return knex.schema.createTable(tableName, function(table) {
+  await knex.schema.createTable(tableName, function(table) {
     table.string('objective_id').primary();
     table.string('status').notNullable();
     table.string('type').notNullable();
     table.jsonb('data');
   });
+
+  await knex.schema.createTable(associativeTableName, function(table) {
+    table.string('objective_id').references(tableName + '.objective_id').notNullable;
+    table
+      .string('channel_id')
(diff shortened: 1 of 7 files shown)


packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
```diff
@@ -1,25 +1,27 @@
...
+  await knex.schema.createTable(tableName, function(table) {
     table.string('objective_id').primary();
     table.string('status').notNullable();
     table.string('type').notNullable();
-    table.string('target_channel_id');
-    table.string('funding_strategy');
-    table.string('ledger_id');
-    table.string('joint_channel_id');
-    table.string('guarantor_id');
+    table.jsonb('data').notNullable();
```
geoknee, line 11: Hopefully this presents a lower risk to future changes in our objective types.

packages/server-wallet/src/models/objective.ts
```diff
@@ -1,121 +1,90 @@
...
+      data: objectiveToBeStored.data,
+    });
 
-    return objectives.map(extract);
+    // Associate the objective with any channel that it references
+    // By inserting an ObjectiveChannel row for each channel
+    // Requires objective and channels to exist
+    extractReferencedChannels(objectiveToBeStored).map(
+      async value => await ObjectiveChannel.query(tx).insert({objectiveId: id, channelId: value})
+    );
```
geoknee, line 86: These lines capture how we would track relations between objectives and channels.

packages/server-wallet/src/db/migrations/20201014174039_objectives.ts
andrewgordstewart: I would prefer a data model with fewer foot guns, but I think this is an entirely reasonable choice given our current timeline. (I might go further and say this is our best choice for the moment.)

packages/server-wallet/src/models/__test__/objectives.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  },
+};
+beforeEach(async () => {
+  await seedAlicesSigningWallet(knex);
+});
+
+describe('Objective > insert', () => {
+  it.skip('fails to insert / associate an objective when it references a channel that does not exist', async () => {
+    // For some reason this does not catch the error :/
+    await expect(Objective.insert({...objective, status: 'pending'}, knex)).rejects.toThrow();
```
geoknee, line 26: @andrewgordstewart do you have any ideas about how to test this properly? It seems that the promise successfully resolves but that an error is thrown afterwards. I'd like to assert that this is the "desired" behaviour (I think?) but I can't seem to catch the error, even when[ wrapping in an arrow function. ](https://jestjs.io/docs/en/expect#tothrowerror)

andrewgordstewart: `Objective.insert` is a bit magical. I think it is a different promise than `new Promise()` is.
```suggestion
  it.skip('fails to insert / associate an objective when it references a channel that does not exist', async () => {
    // For some reason this does not catch the error :/
    await expect(
      new Promise(async resolve => { await Objective.insert({...objective, status: 'pending'}, knex); resolve()})
      ).rejects.toThrow();
```

So, something like this might work?

andrewgordstewart: Screen Shot 2020-10-21 at 2 05 22 PM

`Nonce.query().insert` returns a `QueryBuilder`, which you _can_ `await`, but it's not a `Promise`.

andrewgordstewart: Ok, so `QueryBuilder` is a class that extends `Promise`:
```
export class QueryBuilder<M extends Model, R = M[]> extends Promise<R> { //snip }
```

Perhaps this is incompatible with the implementation `expect(somePromise).rejects`?

geoknee: good thinking; I think I found the issue, though:  I'm awaiting my own `.insert` fn, not the usual one, and it doesn't reject properly when it should.

packages/server-wallet/src/models/objective.ts
```diff
@@ -134,4 +103,18 @@ export class Objective extends Model {
...
+    tx: TransactionOrKnex
+  ): Promise<ObjectiveStoredInDB[]> {
+    const objectiveIds = (
+      await ObjectiveChannel.query(tx)
+        .column('objectiveId')
+        .select()
+        .whereIn('channelId', targetChannelIds)
+    ).map(oc => oc.objectiveId);
+
+    return (await Objective.query(tx).findByIds(objectiveIds)).map(extract);
```
geoknee, line 122: I suspect there may be a smarter / more efficient way of doing this (with a single query, say).

andrewgordstewart: It's easy when you can write SQL :P

```
SELECT objectives.*
FROM objectives
INNER JOIN objective_channels ON objectives.id = objective_channels.objective_id
WHERE channel_id in ($1)
```

andrewgordstewart: (you can use `withGraphFetched`, if your models have the correct relations, and you'll get results. But it's a bit complex.)

tomclose (approved):

PR #2721, opened 2020-10-21
Validate ForceMove transition rules in typescript

lalexgap:
Performs the validation done by the `validTransition`in the ForceMove contract in typescript. This allows us to validate the basic transition rules without using  the `EVM`.

Also we no longer try to validate `Prefund/postfundSetup`  state transitions in the `EVM`.

### DIFF
--- a/packages/server-wallet/src/evm-validator.ts
+++ b/packages/server-wallet/src/evm-validator.ts
@@ -44,7 +44,18 @@ export const validateTransitionWithEVM = async (
     Uint8Array.from(Buffer.from(bytecode.substr(2), 'hex')),
     Uint8Array.from(Buffer.from(data ? data.toString().substr(2) : '0x00', 'hex'))
   );
-
   // We need to ensure the result is the correct length otherwise we might be interpreting a failed assertion
-  return result.length === 32 && (utils.defaultAbiCoder.decode(['bool'], result)[0] as boolean);
+  const transitionPassed =
+    result.length === 32 && (utils.defaultAbiCoder.decode(['bool'], result)[0] as boolean);
+
+  if (!transitionPassed) {
+    // TODO: Figure out the proper encoding.
+    // Right now the revert reason is readable but slightly garbled
+    logger.error(`Call to ValidTransition failed in the EVM`, {
+      result: new TextDecoder().decode(result),
+    });
+    return false;
+  }
+
+  return true;
 };
--- a/packages/server-wallet/src/evm-validator.ts
+++ b/packages/server-wallet/src/evm-validator.ts
@@ -49,13 +49,17 @@ export const validateTransitionWithEVM = async (
     result.length === 32 && (utils.defaultAbiCoder.decode(['bool'], result)[0] as boolean);
 
   if (!transitionPassed) {
-    // TODO: Figure out the proper encoding.
-    // Right now the revert reason is readable but slightly garbled
-    logger.error(`Call to ValidTransition failed in the EVM`, {
-      result: new TextDecoder().decode(result),
+    logger.error(`Call to ValidTransition failed in the EVM ${parseRevertReason(result)}`, {
+      result: parseRevertReason(result),
     });
     return false;
   }
 
   return true;
 };
+
+function parseRevertReason(result: Uint8Array) {
+  // TODO: Figure out the proper encoding.
+  // Right now the revert reason is readable but slightly garbled
+  return new TextDecoder().decode(result.filter(r => r !== 0));
+}
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -253,6 +253,8 @@ export class Store {
     )?.protocolState;
   }
 
+  // Port of the following solidity code
+  // https://github.com/statechannels/statechannels/blob/a3d21827e340c0cc086f1abad7685345885bf245/packages/nitro-protocol/contracts/ForceMove.sol#L492-L534
   async validateTransition(
     fromState: SignedState,
     toState: SignedState,
@@ -275,13 +277,15 @@ export class Store {
       toState.signatures.some(s => s.signer === toMover);
 
     if (!basicValidation) {
-      logger.error('Basic ForceMove transition validation failed.', {fromState, toState});
+      const VALIDATION_ERROR = 'Basic ForceMove transition validation failed.';
+      logger.error(VALIDATION_ERROR, {fromState, toState, error: Error(VALIDATION_ERROR)});
       return false;
     }
 
     // Final state specific validation
     if (toState.isFinal && fromState.outcome !== toState.outcome) {
-      logger.error('Outcome changed on a final state', {fromState, toState});
+      const VALIDATION_ERROR = 'Outcome changed on a final state.';
+      logger.error(VALIDATION_ERROR, {fromState, toState, error: Error(VALIDATION_ERROR)});
       return false;
     }
 
@@ -293,10 +297,10 @@ export class Store {
       toState.appData === fromState.appData;
 
     if (isInFundingStage && !fundingStageValidation) {
-      logger.error(
-        'Invalid setup state transition. The outcome, appData have changed or the state is final',
-        {fromState, toState}
-      );
+      const VALIDATION_ERROR =
+        'Invalid setup state transition. The outcome, appData have changed or the state is final';
+      logger.error(VALIDATION_ERROR, {fromState, toState, error: Error(VALIDATION_ERROR)});
+
       return false;
     }
 
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -182,7 +182,7 @@ export class Store {
 
     if (
       supported &&
-      (await timer('validating transition', async () =>
+      !(await timer('validating transition', async () =>
         this.validateTransition(supported, signedState, tx)
       ))
     ) {
@@ -508,9 +508,9 @@ export class Store {
       const {supported} = channel;
 
       if (
-        await timer('validating transition', async () =>
+        !(await timer('validating transition', async () =>
           this.validateTransition(supported, deserializeState(wireSignedState), tx)
-        )
+        ))
       ) {
         throw new StoreError('Invalid state transition', {
           from: channel.supported,
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -266,17 +266,18 @@ export class Store {
     const toMoverIndex = toState.turnNum % toState.participants.length;
     const toMover = toState.participants[toMoverIndex].signingAddress;
 
-    const turnNumCheck = toState.turnNum === fromState.turnNum + 1;
+    const turnNumCheck = _.isEqual(toState.turnNum, fromState.turnNum + 1);
     if (!turnNumCheck) {
       const VALIDATION_ERROR = `Turn number check failed.`;
       logger.error(VALIDATION_ERROR, {fromState, toState, error: Error(VALIDATION_ERROR)});
     }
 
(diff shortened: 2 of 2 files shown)


packages/server-wallet/src/evm-validator.ts
```diff
@@ -44,7 +44,18 @@ export const validateTransitionWithEVM = async (
...
     Uint8Array.from(Buffer.from(data ? data.toString().substr(2) : '0x00', 'hex'))
   );
-
   // We need to ensure the result is the correct length otherwise we might be interpreting a failed assertion
-  return result.length === 32 && (utils.defaultAbiCoder.decode(['bool'], result)[0] as boolean);
+  const transitionPassed =
+    result.length === 32 && (utils.defaultAbiCoder.decode(['bool'], result)[0] as boolean);
+
+  if (!transitionPassed) {
+    // TODO: Figure out the proper encoding.
```
lalexgap, line 52: This might be an easy fix but my knowledge of text encoding is pretty weak.

andrewgordstewart: Do you have a sample output from this?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -260,6 +246,69 @@ export class Store {
     )?.protocolState;
   }
 
+  async validateTransition(
```
andrewgordstewart, line 249:
```suggestion
  // Port of the following solidity code
  // https://github.com/statechannels/statechannels/blob/a3d21827e340c0cc086f1abad7685345885bf245/packages/nitro-protocol/contracts/ForceMove.sol#L492-L534
  async validateTransition(
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -260,6 +246,69 @@ export class Store {
...
+      toState.turnNum === fromState.turnNum + 1 &&
+      toState.chainId === fromState.chainId &&
+      toState.participants === fromState.participants &&
+      toState.appDefinition === fromState.appDefinition &&
+      toState.challengeDuration === fromState.challengeDuration &&
+      fromState.signatures.some(s => s.signer === fromMover) &&
+      toState.signatures.some(s => s.signer === toMover);
+
+    if (!basicValidation) {
+      logger.error('Basic ForceMove transition validation failed.', {fromState, toState});
```
andrewgordstewart, line 271: 💡 It's nice to have stack traces in a case like this. One way of achieving that would be 
```suggestion
      const VALIDATION_ERROR = 'Basic ForceMove transition validation failed.'
      logger.error(VALIDATION_ERROR, {fromState, toState, error: Error(VALIDATION_ERROR)});
```

(This suggestion is might not be `prettier` enough.)

packages/server-wallet/src/evm-validator.ts
lalexgap: input: 
```
[8,195,121,160,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,32,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,40,67,111,117,110,116,105,110,103,65,112,112,58,32,67,111,117,110,116,101,114,32,109,117,115,116,32,98,101,32,105,110,99,114,101,109,101,110,116,101,100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```

output:
```
\b�y�\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000 \u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000(CountingApp: Counter must be incremented\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000
```

andrewgordstewart: My naive attempt:
```
> b = Buffer.from(<INPUT>);
> encoder.decode(['string'], '0x' + b.toString('hex'))
Uncaught:
Error: overflow (fault="overflow", operation="toNumber", value="3963877391197344453575983046348115674221700746820753546331534351508065746944", code=NUMERIC_FAULT, version=bignumber/5.0.7)
    at Logger.makeError (/Users/andrewstewart/Code/magmo/monorepo/node_modules/@ethersproject/logger/lib/index.js:179:21)
    at Logger.throwError (/Users/andrewstewart/Code/magmo/monorepo/node_modules/@ethersproject/logger/lib/index.js:188:20)
    at throwFault (/Users/andrewstewart/Code/magmo/monorepo/node_modules/@ethersproject/bignumber/lib/bignumber.js:278:19)
    at BigNumber.toNumber (/Users/andrewstewart/Code/magmo/monorepo/node_modules/@ethersproject/bignumber/lib/bignumber.js:146:13)
    at /Users/andrewstewart/Code/magmo/monorepo/node_modules/@ethersproject/abi/lib/coders/array.js:89:60
    at Array.forEach (<anonymous>)
    at Object.unpack (/Users/andrewstewart/Code/magmo/monorepo/node_modules/@ethersproject/abi/lib/coders/array.js:85:12)
    at TupleCoder.decode (/Users/andrewstewart/Code/magmo/monorepo/node_modules/@ethersproject/abi/lib/coders/tuple.js:39:49)
    at AbiCoder.decode (/Users/andrewstewart/Code/magmo/monorepo/node_modules/@ethersproject/abi/lib/abi-coder.js:93:22) {
  reason: 'overflow',
  code: 'NUMERIC_FAULT',
  fault: 'overflow',
  operation: 'toNumber',
  value: '3963877391197344453575983046348115674221700746820753546331534351508065746944'
}
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -260,6 +246,73 @@ export class Store {
...
+    // This is the basic validation from the force move contract
+    // that applies to all state transitions
+    const basicValidation =
+      toState.turnNum === fromState.turnNum + 1 &&
+      toState.chainId === fromState.chainId &&
+      toState.participants === fromState.participants &&
+      toState.appDefinition === fromState.appDefinition &&
+      toState.challengeDuration === fromState.challengeDuration &&
+      fromState.signatures.some(s => s.signer === fromMover) &&
+      toState.signatures.some(s => s.signer === toMover);
```
andrewgordstewart, line 270: It's non-trivial to verify that this is a port of the solidity code. In fact, I don't think it is -- for example, this isn't exactly the signature check performed in the smart contracts.

andrewgordstewart (approved):
This looks ok.

It's not an exact port of the code, so it's a little hard to verify that it's "correct". But, it's readable, and it seems correct to me!

packages/server-wallet/src/evm-validator.ts
andrewgordstewart: Wait, when you said it's readable but garbled, it took me a long time to realize there was actual text in this line :P
> \b�y�\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000 \u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000(CountingApp: Counter must be incremented\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000

andrewgordstewart: Another naive attempt:
```
for (let i = 0; i < input.length; i++) {
  try {
    console.log(decoder.decode(['string'], input.slice(idx)))
  } catch (e) {
    
  }
}
```

Returns `undefined`, and logs nothing to the console :(

packages/server-wallet/src/wallet/store.ts
lalexgap: I modeled this based on the pseudocode @geoknee  has in the documentation. I chose that because it seemed clearer  than trying to mimic the contract which has multiple checks spread across multiple contracts. That does introduce the risk of a discrepancy but the rules seem simple enough 🤞

packages/server-wallet/src/evm-validator.ts
```diff
@@ -44,7 +44,18 @@ export const validateTransitionWithEVM = async (
...
+    // Right now the revert reason is readable but slightly garbled
+    logger.error(`Call to ValidTransition failed in the EVM`, {
+      result: new TextDecoder().decode(result),
```
andrewgordstewart, line 55:
```suggestion
      result: new TextDecoder().decode(result).replace(/\0/g, ''),
```

Based on
```
> decoded = new TextDecoder().decode(Buffer.from(input)).replace(
'\b�y�\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000 \u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000(CountingApp: Counter must be incremented\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000'
> decoded.replace(/\0/g, '')
```

andrewgordstewart: (It looks like the `result` buffer contains a lot of null characters, perhaps because the first few bytes do not represent valid UTF-8 or something like that?)

lalexgap: `\b�y� (CountingApp: Counter must be incremented` is actually what I initially saw when logging out with `console.log`. It appears that  using `pino` causes `\u0000` to be logged out.

andrewgordstewart: My guess is that `console.log` prints nothing over and over when it sees the `null` character, but `pino` prints the escaped `\u0000` version of the `null` character.

andrewgordstewart: If you ask `pino` to `prettyPrint`, for example, it does not print `\u0000`:
```
> pino = require('pino')({prettyPrint: true});
Pino {
  // snip
}
> pino.info(decoded)
[1603313859603] INFO  (45642 on MacBook-Pro-10.local):�y� (CountingApp: Counter must be incremented
undefined
```

lalexgap: I've done something similar in 83cc99e to filter out the null bytes

tomclose (approved):

PR #2726, opened 2020-10-22
feat: wallet emits channel ready events

kerzhner:
Also contains changes from https://github.com/statechannels/statechannels/pull/2725.

Consider the following scenario. A channel manager **CM** is trying to open one channel with a receipt manager **RM**. How does this work with fake funded channels? CM creates prefund 1 state and sends the state to RM. CM creates a promise **P** that resolves when:
1. A response is received from RM.
1. AND CM has no more messages to send to RM.

In practice, the promise P resolves when CM receives postfund2 from RM. This poses a problem for directly funded channels. Consider the following scenario.
1. CM sends prefund1 to RM.
1. CM receives prefund2 from RM.
1. CM triggers the chain deposit transaction.
1. CM has no more states to send to RM.
1. Promise P resolves.

There are several strategies for CM to block on a channel progressing to a ready state (aka last prefund has been signed). In this PR we chose to implement a wallet event that notifies any wallet consumer when a channel is ready.

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -645,7 +645,7 @@ export class Wallet extends EventEmitter<WalletEvent>
   };
 
   // ChainEventSubscriberInterface implementation
-  onHoldingUpdated(arg: HoldingUpdatedArg): void {
+  holdingUpdated(arg: HoldingUpdatedArg): void {
     const channelUpdate: WalletEventName = 'channelUpdate';
     this.updateChannelFundingForAssetHolder(arg).then(singleChannelOutput =>
       this.emit(channelUpdate, singleChannelOutput)
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -72,7 +72,9 @@ export type MultipleChannelOutput = {
   objectivesToApprove?: Omit<ObjectiveStoredInDB, 'status'>[];
 };
 type Message = SingleChannelOutput | MultipleChannelOutput;
-type WalletEvent = {singleChannelOutput: SingleChannelOutput};
+
+export type WalletEventName = 'channelUpdate' | 'channelReady';
+type WalletEvent = {[key in WalletEventName]: SingleChannelOutput};
 
 const isSingleChannelMessage = (message: Message): message is SingleChannelOutput =>
   'channelResult' in message;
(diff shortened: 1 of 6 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -634,7 +646,10 @@ export class Wallet extends EventEmitter<WalletEvent>
 
   // ChainEventSubscriberInterface implementation
   onHoldingUpdated(arg: HoldingUpdatedArg): void {
```
tomclose, line 648: I realise that this change isn't part of this PR, but I'm wondering if this is the best name for this method. As far as I'm aware `onXXX` methods are normally exposed on the event emitter:
```ts
emitter.on('some-event', (event) => doAction(event))
// or
emitter.onSomeEvent((event) => doAction(event))
```
The `onXXX` method exists on the emitter, and is called by the subscriber, who passes in a callback that is used to bind an action to that event.

In this case it's the other way around. The emitter is calling the `onHoldingUpdated` method on the subscriber with the event data. Of course this works, it just seems to run contrary to the normal pattern around events and `onXXX` methods, which could be confusing. For example, it took me a while to figure out what was going on here, as the naming pushed me in a different direction.

Maybe something like `holdingUpdated` or `holdingChanged` would work?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -617,8 +619,18 @@ export class Wallet extends EventEmitter<WalletEvent>
...
+          markObjectiveAsDone();
+          // todo: is there a better check for "has the final postfund state been signed"?
+          if (app.supported?.turnNum === app.participants.length * 2 - 1) {
+            const toEmit: SingleChannelOutput = {
+              channelResult: ChannelState.toChannelResult(app),
+              outbox: [],
+            };
+            const channelReady: WalletEventName = 'channelReady';
+            this.emit(channelReady, toEmit);
+          }
```
tomclose, line 632: In the description of the PR you characterise the problem as a promise resolving before the channel is open, and therefore the need to separately notify the app when the channel is ready. It seems to me that the problem goes beyond this.

I'd characterise the problem as "we need to tell the CM that it needs to act when the next step of the protocol is triggered by a chain-service event". When the deposit event that fully funds the channel arrives, we need to (a) trigger a post-fund-setup to be sent to the opponent, and (b) when all messages have been exchanged, notify the app that funding is complete. It seems that to accomplish (a), you need to have a notification that also includes messages to send. The interaction would be like this:

1. ChainService sends funding event to wallet
2. Wallet updates funding on channels and runs `takeActions`
3. Take actions realises that the next step is to send the post-fund-0
4. Wallet sends an event to the CM, containing both a channel-result and the post-fund-0 to send to the opponent
    * ☝️  **this is the bit that we're currently missing (I think?)**
5. CM sends to RM, creating a promise that resolves when all messaging is done
6. When the promise resolves, then the channel will be open, and the CM will be given the channel result

It seems that if you solve (a), then you don't need the notification here - you need notifications to be emitted exactly when activity is triggered by a chain service event. If that's true, it seems worth removing this event emission, as one of the confusions in the xstate wallet is when the double notifications that the app would receive about some events - one through a notification, and the other from the return value.

packages/server-wallet/src/wallet/index.ts
kerzhner: Good point! Let's change the name.

kerzhner: Addressed in 5bdbb0c3630656341274fa0c7323a24d25721a84

kerzhner:
After a discussion with @tomclose, closing this PR. The renames are now https://github.com/statechannels/statechannels/pull/2736

PR #2727, opened 2020-10-23
[nitro-protocol] post-audit changes

geoknee:
Mostly straightforward things, apart from:
- changing name of `forceMove` method to `challenge` (recommended to avoid confusion). Note that I marked this as a **breaking change**
- [ ] SafeMath-ifying `ForceMove.sol` -- still to be done, but not straightforward because we use a lot of `uint48` and `uint8` which I don't think work straightforwardly with `SafeMath`. Workaround might involve a bunch of casting (yuk), extending the SafeMath library, or relying on our own `require()` statements. I did a few of those already.

### Rejected suggestions
---
- `_recoverSigner` should not be `pure`. It depends on a constant defined in the scope of the contract (not the function), which is why I think it may have been flagged. But I think it is still pure.
- arithmetic functions flagged inside of `SafeMath` (these are obviously necessary)
- unused function params in virtual functions (will be overriden, so OK to have empty function body)

### Punted suggestions that require more thought
--- 
- looping over unbounded data structures

### DIFF
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -36,7 +36,7 @@ contract AssetHolder is IAssetHolder {
         uint256 finalPayoutAmount;
         uint256 firstNewAllocationItemAmount;
 
-        for (uint256 i = 0; i < allocation.length; i++) {
+        for (uint256 i = 0; i < allocation.length; i = i.add(1)) {
             if (balance == 0) {
                 // if funds are completely depleted, keep the allocationItem and do not pay out
             } else {
@@ -44,15 +44,15 @@ contract AssetHolder is IAssetHolder {
                 if (balance < _amount) {
                     // if funds still exist but are insufficient for this allocationItem, payout what's available and keep the allocationItem (but reduce the amount allocated)
                     // this block is never executed more than once
-                    numPayouts++;
+                    numPayouts = numPayouts.add(1);
                     overlap = true;
                     finalPayoutAmount = balance;
-                    firstNewAllocationItemAmount = _amount - balance;
+                    firstNewAllocationItemAmount = _amount.sub(balance);
                     balance = 0;
                 } else {
                     // if ample funds still exist, pay them out and discard the allocationItem
-                    numPayouts++;
-                    numNewAllocationItems--;
+                    numPayouts = numPayouts.add(1);
+                    numNewAllocationItems = numNewAllocationItems.sub(1);
                     balance = balance.sub(_amount);
                 }
             }
@@ -66,8 +66,8 @@ contract AssetHolder is IAssetHolder {
             Outcome.AllocationItem[] memory newAllocation = new Outcome.AllocationItem[](
                 numNewAllocationItems
             );
-            for (uint256 k = 0; k < numNewAllocationItems; k++) {
-                newAllocation[k] = allocation[allocation.length - numNewAllocationItems + k];
+            for (uint256 k = 0; k < numNewAllocationItems; k = k.add(1)) {
+                newAllocation[k] = allocation[allocation.length.sub(numNewAllocationItems).add(k)];
                 if (overlap && k == 0) {
                     newAllocation[k].amount = firstNewAllocationItemAmount;
                 }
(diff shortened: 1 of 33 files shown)


snario (approved):

packages/nitro-protocol/contracts/ETHAssetHolder.sol
```diff
@@ -62,6 +67,7 @@ contract ETHAssetHolder is AssetHolder {
      * @param amount Quantity of wei to be transferred.
      */
     function _transferAsset(address payable destination, uint256 amount) internal override {
-        destination.transfer(amount);
+        (bool success, ) = destination.call{value: amount}('');
```
GNSPS, line 70: This is actually problematic. Because the `.transfer()` method was preventing reentrancy by only giving the callee the 2300 gas stipend. We should either keep this as was or add a reentrancy guard somehow.

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -367,7 +376,7 @@ contract ForceMove is IForceMove {
             uint256 offset = (nParticipants + largestTurnNum - i) % nParticipants;
             // offset is the difference between the index of participant[i] and the index of the participant who owns the largesTurnNum state
             // the additional nParticipants in the dividend ensures offset always positive
-            if (whoSignedWhat[i] + offset < nStates - 1) {
+            if (whoSignedWhat[i] + offset + 1 < nStates) {
```
GNSPS, line 379: This was the nastiest of nasty! 😄 😉  👀

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -35,23 +36,23 @@ contract AssetHolder is IAssetHolder {
         uint256 finalPayoutAmount;
         uint256 firstNewAllocationItemAmount;
 
-        for (uint256 i = 0; i < allocation.length; i++) {
+        for (uint256 i = 0; i < allocation.length; i = i.add(1)) {
```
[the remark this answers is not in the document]
andrewgordstewart: So the reasoning is that we know that allocation.length cannot exceed the maximum integer, because allocation was passed in as calldata, and an array that long in calldata would cost too much gas?

packages/nitro-protocol/contracts/ForceMove.sol
andrewgordstewart: 😬

packages/nitro-protocol/contracts/ETHAssetHolder.sol
geoknee: I was following this advice https://diligence.consensys.net/blog/2019/09/stop-using-soliditys-transfer-now/. 

It seems best to not rely on the gas stipend for safety. We currently use a checks-effects-interaction pattern to protect against reentrancy attacks in the `AssetHolder` base contract. If we think an explicit reentrancy guard is a good idea (I don't think it is necessary), it would probably make most sense to put it in the base contract, too.

packages/nitro-protocol/contracts/ETHAssetHolder.sol
```diff
@@ -49,10 +50,14 @@ contract ETHAssetHolder is AssetHolder {
...
         // require successful deposit before updating holdings (protect against reentrancy)
-        // refund whatever wasn't deposited.
-        msg.sender.transfer(amount.sub(amountDeposited));
+
         holdings[destination] = holdings[destination].add(amountDeposited);
         emit Deposited(destination, amountDeposited, holdings[destination]);
+
+        // refund whatever wasn't deposited.
+        uint256 refund = amount.sub(amountDeposited);
+        (bool success, ) = msg.sender.call{value: refund}('');
```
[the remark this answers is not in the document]
geoknee: #2827

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -195,10 +196,10 @@ contract AssetHolder is IAssetHolder {
         uint256 newAllocationLength = allocation.length;
 
         // first increase payouts according to guarantee
-        for (uint256 i = 0; i < guarantee.destinations.length; i++) {
+        for (uint256 i = 0; i < guarantee.destinations.length; i = i.add(1)) {
             // for each destination in the guarantee
             bytes32 _destination = guarantee.destinations[i];
-            for (uint256 j = 0; j < allocation.length; j++) {
+            for (uint256 j = 0; j < allocation.length; j = j.add(1)) {
```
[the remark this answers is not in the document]
geoknee: I take your point about these. 

 #2828

PR #2728, opened 2020-10-23
fix: log error on invalid state transition of received state instead of throwing

snario:
You may want to store a state that is not a valid transition from a previously supported one (e.g., a new ledger update) but not consider it supported. In that case, you still want to check if a valid transition is there to determine if it is in fact supported but you don't want to discard the new state entirely.

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -520,7 +520,7 @@ export class Store {
           this.validateTransition(supported, incomingState, tx)
         ))
       ) {
-        throw new StoreError('Invalid state transition', {
+        logger.error('Invalid state transition — state is stored but not supported', {
           from: channel.supported,
           to: wireSignedState,
         });


tomclose (approved):

andrewgordstewart (commented):
How is the wallet supposed to know how to act on this information?

tomclose:
Further context: in ledger funding, anytime you propose a new state you sign a state that isn't a valid transition from the old state. It therefore doesn't seem that performing the `validTransition` check at the time of signing is the right thing to do. This PR disables the check while we rethink the strategy, as the current approach is blocking ledger channels.

In general, it seems like whether `s1 -> s2` is a validTransition is just one factor that the wallet needs to consider when signing transitions. It's also a piece of information that the wallet needs to store, as it needs to know it when figuring out if a state is supported or not. (Complicating this is the fact that, when/if we allow multiple states per turnNum, it could be a valid transition from _multiple_ states, and this would be important to know because it might be supported along one path and not another. We could probably leave this for now though.)

tomclose:
Possible solution for the short term:

1. If a channel has a non-empty app-definition then we only store states that are validTransitions.
     * This covers `updateChannel` and `pushMessage`
     * It means we don't have to store the valid transition information, as we know that if `s1` and `s2` are in the store, and the turn num increments, then `s1 -> s2` is a valid transition (as we don't allow duplicate states with the same turn number)
2. If a channel has an empty app-definition, we don't require states to be validTransitions once the channel is running.

PR #2731, opened 2020-10-23
[server-wallet] Support ledger defunding

snario:
When `closeChannel` is called on a channel that is ledger funded, it will kick off an objective for the wallet to close that channel and now with this PR, that objective does not complete unless the funds are re-allocated to its parent ledger channel. Works concurrently.

### DIFF
--- a/packages/server-wallet/src/protocols/process-ledger-queue.ts
+++ b/packages/server-wallet/src/protocols/process-ledger-queue.ts
@@ -8,6 +8,7 @@ import {
   SignedStateWithHash,
   SimpleAllocation,
   areAllocationItemsEqual,
+  BN,
 } from '@statechannels/wallet-core';
 import {Transaction} from 'knex';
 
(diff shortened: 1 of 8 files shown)


packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -26,9 +27,33 @@ import {
...
+            ? _.update(allocationItems, idx, to => ({
+                destination,
+                amount: BN.add(amount, to.amount),
+              }))
+            : [...allocationItems, {destination, amount}];
+        }, allocationItems)
+        .filter(item => item.destination !== channel.channelId),
+    original.allocationItems
+  ),
+});
```
snario, line 55: This is kinda complicated. Curious what folks would prefer I do with this to make it easier to grok? Basically it is a map-reduce de-allocating funds into a single ledger outcome.

PR #2732, opened 2020-10-23
refactor: move invalid transition check from store to wallet API

snario:


tomclose:
Alternative approach: https://github.com/statechannels/statechannels/pull/2735

PR #2733, opened 2020-10-23
Add assetHolderAddress column to channels table and store leger-requests in new db table

geoknee:
- includes `crashAndRestart` #2730 
- Tests failing currently (inherited failures from base branch).
- ~closely follows the data model sketched on the base branch~
- changes data model from base branch a bit.


packages/server-wallet/src/db/migrations/20201014174041_ledgers.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function up(knex: Knex): Promise<any> {
+  await knex.schema.createTable(tableName, function(table) {
+    table
+      .string('ledger_channel_id')
+      .primary()
+      .references('channel_id')
+      .inTable('channels');
+    table.string('asset_holder_address').notNullable();
+  });
```
tomclose, line 13: I think it might be better to add this info to the `channels` table. That would make it easy to answer the question "is this channel a ledger channel", which we will need e.g. [here](https://github.com/statechannels/statechannels/pull/2735#discussion_r510957955).

geoknee: OK 👍

andrewgordstewart: Just a side note, but putting such a flag on the `channels` table "smells bad" from a data normality point of view. It feels to me like things ought to be organized differently:
- there is a `ledgers` table that has a foreign key relation to the `channels` table
- there is an `application_channels` table that has a foreign key relation to the `channels` table
- `channels` table is precisely the common set of data that must be stored for any nitro state channel.

geoknee: In that case, to support a synchronous `isLedger()` function, I guess we would have to wrap the `forId(channelId)` around a query that touches multiple tables, so the in-memory objects have the flag?

I'm not sure I see a big benefit from that?

andrewgordstewart: I think it would require the wallet to think about its data differently. I don't have a well-formed answer for that mode of thinking.

snario:
Replaced by #2750

PR #2734, opened 2020-10-23
Close channels with sequential turn number orders

snario:
Opened just to address some discussion in Slack — not necessarily proposing we do this.


tomclose:
Could you add a description of the change this makes and the reason behind making it?

andrewgordstewart (changes requested):
I'd like some justification for this, considering that this has the potential to cost a very large amount of real money.

Before merging, can you include the gas cost of concluding a channel on-chain before and after this change?

PR #2735, opened 2020-10-23
fix: disable validTransition check for ledger channels

tomclose:
Following the approach outlined [here](https://github.com/statechannels/statechannels/pull/2728#issuecomment-715311687):

* it's too restrictive to check all state transitions for valid transitions, because new ledger states will always fail this check
* in the longer term we probably want to think more carefully about where we do this check
* but for now, we just turn it off for ledger channels

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -189,6 +189,7 @@ export class Store {
       supported &&
       // If the state is the same as the support state then its not a transition just adding signatures
       !statesEqual(supported, state) &&
+      !(await this.isLedger(channelId)) &&
       !(await timer('validating transition', async () =>
         this.validateTransition(supported, signedState, tx)
       ))
@@ -515,7 +516,8 @@ export class Store {
     if (
       channel.supported &&
       // If the state is the same as the support state then its not a transition just adding signatures
-      !statesEqual(channel.supported, incomingState)
+      !statesEqual(channel.supported, incomingState) &&
+      !(await this.isLedger(channelId))
     ) {
       const {supported} = channel;
 
@@ -524,7 +526,7 @@ export class Store {
           this.validateTransition(supported, incomingState, tx)
         ))
       ) {
-        logger.error('Invalid state transition — state is stored but not supported', {
+        throw new StoreError('Invalid state transition', {
           from: channel.supported,
           to: wireSignedState,
         });


packages/server-wallet/src/wallet/store.ts
```diff
@@ -189,6 +189,7 @@ export class Store {
       supported &&
       // If the state is the same as the support state then its not a transition just adding signatures
       !statesEqual(supported, state) &&
+      !(await this.isLedger(channelId)) &&
```
andrewgordstewart, line 192: This is a new, async call on the critical path. How will it affect performance?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -524,7 +526,7 @@ export class Store {
           this.validateTransition(supported, incomingState, tx)
         ))
       ) {
-        logger.error('Invalid state transition — state is stored but not supported', {
+        throw new StoreError('Invalid state transition', {
```
andrewgordstewart, line 529: :+1:

packages/server-wallet/src/wallet/store.ts
tomclose: Good point. Right now it won't, but once @geoknee's work to put ledgers in the db is complete it might. Will investigate further.

andrewgordstewart: I think it could be replaced with a performant, synchronous check on the channel's `appDefinition`, like you've suggested somewhere else
- if the user wants to use a null app, then the wallet can infer that it should not validate transitions in the running phase, since there are none
- therefore, `validateTransition` could just return `true` when the `appDefinition` is zero

tomclose: The check should definitely be synchronous and based on the channel's state somehow. @geoknee is going to add the `isLedger` property to the channels table, so we have an alternative.

I think it probably makes sense to check `isLedger` for now, rather than inferring from a zero `appDefinition`, as this is more explicit with what we expect to support right now. It seems like there might be more thought required on how to handle users using the null app. For example, should any address where there isn't bytecode available be treated as a null app? What should happen if the wallet just doesn't have the bytecode? Do we support null-app-style updates in apps that do have a valid transition e.g. if the user passes a special flag? If so, should users pass this flag in if they're doing null-app updates?

andrewgordstewart: But the question is simply about whether we should validate transitions, and that is a question about what rules the app uses, and not what the purpose of the channel is?

If we decide to use the consensus app for ledger channels, then `isLedger` would presumably return `true`, and you would not validate a transition in a channel where you ought to?

geoknee: > therefore, validateTransition could just return true when the appDefinition is zero

it should return `false` then , as the on-chain code would. We can still "skip the check" if we want to for performance reasons, but we should definitely return false. 

The code that decides whether to _store_ the state should not require a valid transition in general. In the short term, we could require it for non-null apps / application channels as a simplification that wouldn't hurt us too much. In the longer term we should support progression-by-consensus updates in non-null apps, too, and therefore have a smarter scheme for deciding whether to store a state (@andrewgordstewart mentioned storing them in a validTransition tree).

The code that decides whether a state is _supported_ should require validTransition checks _as necessary_ (depending on how states are signed in the support proof) in the same way that the on chain code does it.

andrewgordstewart (approved):
It'd be better if this were a performant, synchronous check already, but seems ok as is, if it becomes performant quickly.

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: ^ quite clarifying

>  (@andrewgordstewart mentioned storing them in a validTransition tree).

_technically_, it would be a forest 🌴 🎄 🌳 🌲

PR #2737, opened 2020-10-23
Fix EVM Transition check

lalexgap:
Previously we were passing in the `fromState` twice into `validateTransitionWithEVM` causing the call to always fail.

This PR fixes this and updates the `update-channel` integration test to use EVM validation so issues like this will be caught in the future.


packages/server-wallet/src/wallet/__test__/integration/update-channel.test.ts
```diff
@@ -61,4 +81,23 @@ describe('error cases', () => {
       Error('channel not found')
     );
   });
+
+  it('throws when it is an invalid app transition', async () => {
```
andrewgordstewart, line 85: I'd prefer for this test to live in the test of the pure function -- it is significantly faster to run (so our test suite doesn't grow in duration)

andrewgordstewart: For reference:
```
monorepo/packages/server-wallet on  master [$⇣] is 📦 v1.0.0 via ⬢ v12.16.3 on ☁️  (us-west-1) on ☁️ andrew.stewart@mesh.xyz took 14s 
❯ yarn test src/wallet/__test__/integration/update-channel.test.ts
yarn run v1.22.4
$ yarn jest -c ./jest/jest.config.js --runInBand src/wallet/__test__/integration/update-channel.test.ts
$ /Users/andrewstewart/Code/magmo/monorepo/node_modules/.bin/jest -c ./jest/jest.config.js --runInBand src/wallet/__test__/integration/update-channel.test.ts
(node:30758) V8: /Users/andrewstewart/Code/magmo/monorepo/node_modules/rustbn.js/lib/index.asm.js:2 Linking failure in asm.js: Unexpected stdlib member
 PASS  src/wallet/__test__/integration/update-channel.test.ts (6.686 s)
  ✓ updates a channel (717 ms)
  error cases
    ✓ throws when it is not my turn (570 ms)
    ✓ throws when the channel isn't found (176 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        6.732 s, estimated 7 s
Ran all test suites matching /src\/wallet\/__test__\/integration\/update-channel.test.ts/i.
✨  Done in 8.05s.

monorepo/packages/server-wallet on  master [$⇣] is 📦 v1.0.0 via ⬢ v12.16.3 on ☁️  (us-west-1) on ☁️ andrew.stewart@mesh.xyz took 8s 
❯ yarn test src/handlers/__test__/update-channel.test.ts          
yarn run v1.22.4
$ yarn jest -c ./jest/jest.config.js --runInBand src/handlers/__test__/update-channel.test.ts
$ /Users/andrewstewart/Code/magmo/monorepo/node_modules/.bin/jest -c ./jest/jest.config.js --runInBand src/handlers/__test__/update-channel.test.ts
(node:30835) V8: /Users/andrewstewart/Code/magmo/monorepo/node_modules/rustbn.js/lib/index.asm.js:2 Linking failure in asm.js: Unexpected stdlib member
 PASS  src/handlers/__test__/update-channel.test.ts (5.284 s)
  ✓ validUpdate (3 ms)
  ✓ error cases [Error: must have latest state] (1 ms)
  ✓ error cases [Error: channel must be in running state]
  ✓ error cases [Error: it is not my turn]

Test Suites: 1 passed, 1 total
Tests:       4 passed, 4 total
Snapshots:   0 total
Time:        5.33 s
Ran all test suites matching /src\/handlers\/__test__\/update-channel.test.ts/i.
✨  Done in 6.39s.
```

Maybe it doesn't matter, because both tests take 5s to compile a large project. But the `handler` tests run in <10ms.

lalexgap: > I'd prefer for this test to live in the test of the pure function -- it is significantly faster to run (so our test suite doesn't grow in duration)

The `UpdateChannel` handler does not validate the transition so we can't add the test there. I think this integration test is the only test that calls `updateChannel` on the `wallet`.

We could probably create a new test file for tests directly against the `store` but I feel like it would still be a good idea to have this enabled in the "integration" test.

andrewgordstewart: Right. I guess the `validateTransition` has to fetch the bytecode (once) from the DB.

We could refactor it to accept the bytecode -- at that point it can be a pure function. The wallet could pass the bytecode to `updateChannel`

andrewgordstewart (approved):

packages/server-wallet/src/wallet/__test__/integration/update-channel.test.ts
lalexgap: > Right. I guess the `validateTransition` has to fetch the bytecode (once) from the DB.
> 
> We could refactor it to accept the bytecode -- at that point it can be a pure function. The wallet could pass the bytecode to `updateChannel`

Just a caveat that the original bug wasn't in `validateTransitionWithEVM` but with `validateTransition` in the `store`.

andrewgordstewart: Yes, but I'm suggesting that `updateChannel` should do the transition validation (and as much business logic as possible?)

IE. get business logic as far away from DB code as possible.

PR #2741, opened 2020-10-24
fix: completeOpenChannel uses turn number

kerzhner:
`completeOpenChannel` should only generate the `completeObjective` after the last `postfund` state has been signed, NOT any postfund state.

After this is fixed and new server wallet package is published, the direct funding e2e test can be commented back in in the-graph repo: https://github.com/statechannels/the-graph/commit/6184cf377c4ae8069a34a5ffd8778f74c0275a81

### DIFF
--- a/packages/server-wallet/src/protocols/open-channel.ts
+++ b/packages/server-wallet/src/protocols/open-channel.ts
@@ -20,7 +20,7 @@ const isPrefundSetup = stageGuard('PrefundSetup');
 
 // These are currently unused, but will be used
 // const isRunning = stageGuard('Running');
-const isPostfundSetup = stageGuard('PostfundSetup');
+// const isPostfundSetup = stageGuard('PostfundSetup');
 const isRunning = stageGuard('Running');
 // const isMissing = (s: State | undefined): s is undefined => stage(s) === 'Missing';
 
@@ -122,7 +122,7 @@ const signPostFundSetup = (ps: ProtocolState): ProtocolResult | false =>
   });
 
 const completeOpenChannel = (ps: ProtocolState): CompleteObjective | false =>
-  (isRunning(ps.app.supported) || isPostfundSetup(ps.app.supported)) &&
+  (isRunning(ps.app.supported) || ps.app.supported === ps.app.participants.length * 2 - 1) &&
   completeObjective({channelId: ps.app.channelId});
 
 export const protocol: Protocol<ProtocolState> = (ps: ProtocolState): ProtocolResult =>
--- a/packages/server-wallet/src/protocols/open-channel.ts
+++ b/packages/server-wallet/src/protocols/open-channel.ts
@@ -122,7 +122,7 @@ const signPostFundSetup = (ps: ProtocolState): ProtocolResult | false =>
   });
 
 const completeOpenChannel = (ps: ProtocolState): CompleteObjective | false =>
-  (isRunning(ps.app.supported) || ps.app.supported === ps.app.participants.length * 2 - 1) &&
+  (isRunning(ps.app.supported) || ps.app.supported.turnNum === ps.app.participants.length * 2 - 1) &&
   completeObjective({channelId: ps.app.channelId});
 
 export const protocol: Protocol<ProtocolState> = (ps: ProtocolState): ProtocolResult =>
--- a/packages/server-wallet/src/protocols/open-channel.ts
+++ b/packages/server-wallet/src/protocols/open-channel.ts
@@ -122,7 +122,8 @@ const signPostFundSetup = (ps: ProtocolState): ProtocolResult | false =>
   });
 
 const completeOpenChannel = (ps: ProtocolState): CompleteObjective | false =>
-  (isRunning(ps.app.supported) || ps.app.supported.turnNum === ps.app.participants.length * 2 - 1) &&
+  (ps.app.supported?.turnNum === ps.app.participants.length * 2 - 1 ||
+    isRunning(ps.app.supported)) &&
   completeObjective({channelId: ps.app.channelId});
 
 export const protocol: Protocol<ProtocolState> = (ps: ProtocolState): ProtocolResult =>
--- a/packages/server-wallet/src/protocols/__test__/open-channel.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/open-channel.test.ts
@@ -18,8 +18,9 @@ const outcome2 = simpleEthAllocation([
 ]);
 const participants = [alice(), bob()];
 const prefundState = {outcome, turnNum: 0, participants};
-const prefundState2 = {outcome: outcome2, turnNum: 0, participants};
-const postFundState = {outcome, turnNum: 2, participants};
+const prefundState1 = {outcome: outcome2, turnNum: 0, participants};
+const postfundState2 = {outcome, turnNum: 2, participants};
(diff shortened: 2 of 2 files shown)


packages/server-wallet/src/protocols/open-channel.ts
```diff
@@ -122,7 +122,7 @@ const signPostFundSetup = (ps: ProtocolState): ProtocolResult | false =>
   });
 
 const completeOpenChannel = (ps: ProtocolState): CompleteObjective | false =>
-  (isRunning(ps.app.supported) || isPostfundSetup(ps.app.supported)) &&
+  (isRunning(ps.app.supported) || ps.app.supported === ps.app.participants.length * 2 - 1) &&
```
andrewgordstewart, line 125: I'd like to note this as an argument for changing the nitro protocol logic.

Suppose we change the adjudicator to have the following rules:
A transition from state `s_n` to state `s_m` is valid if and only if:
R1. `n > 0` and `m == n+1` (these are turn numbers)
R2. `s_n` and `s_m` are not final
R3. `AppRules.validTransition(s_n, s_m)` is true
R4. `constantPart(s_n) == constantPart(s_m)`

In other words, turns 0 & 1 are specially reserved for pre- and post-FS states. This offers the same safety* as the current set of rules.

* I am 99% sure of this.

# Benefits
B1. This is a simpler set of rules than the current implementation. It is easier for us to audit, and easier to explain to other folks _conditional on being able to explain signatures_.
B2. This is more gas efficient in both the happy path (we force conclusion proofs to be single states) and the sad path (we force challenges before running states

# Claimed neutrals
CN3. Wallets can get the same benefits with the current implementation, since they can opt to sign states `0` and then states `3`.

# Drawbacks
D4. We would have to change the contracts.
D5. It seemed to be the consensus preference to take turns for pre- and post-FS.  We could not opt into turn-taking if we wanted to.

I am happy for others to add to this. I am biased.

# Claimed?
I say that (CN3) is a _claimed_ neutral because, as is evidenced here, the current nitro rules allow for more foot guns. The following are three foot guns that are avoided by the rules R1-R4.

## Footgun 1
Suppose I want to know "Is this channel running? Is it safe for me to try to make an update?"
### Current answer
A. BOTH `channel.supportedState.turnNumber >= channel.participants.length * 2 - 1`
B. AND "I have not seen any `final` states"

The error in `master` is that (A) is replaced with `isPostFundSetup(channel.supported) || isRunning(channel.supported)`, which is succinct, but wrong.

### Simpler answer
A. BOTH `channel.supportedState.turnNumber >= 1`
B. AND "I have not seen any `final` states"

In this case, (A) is now equivalent to `isPostFundSetup(channel.supported) || isRunning(channel.supported)`.
So, the developer who wrote this line's mental model is in line with the correct answer (in spite of it being more complicated.)

It is a good property for mental models to be correct. (It is a better property for them to be correct _and_ simple!)

## Footgun 2
Suppose I want to quickly write a wallet.
Then I will assume that each channel has two participants, and use the (often magic) constant `3` instead of the correct, dynamic number `channel.participants.length * 2 - 1`.

You might not have noticed my accidental use of this magic constant in (CN3). I didn't when I wrote it.

## Footgun 3
Suppose I want to quickly write a wallet.
Then I might assume that the outcome of the channel has one outcome per participant, _in the same order as the participants_
This would lead me to make assumptions about when a participant should sign a post fund state based on its _index in the channel_

# Drawbacks?
D4: We are going to change the contracts anyway, post-audit.
  - This retort has another drawback, that the audit would now be a bit stale?
D5. I assert that the consensus was wrong, and that turn-taking while funding nudges our mental models to be slightly wrong, manifesting in bugs like this one.

andrewgordstewart (approved):
In addition to the comment left on the single line change, I'll add these two observations:

## Observation 1
This was a simple logic bug that was difficult to detect where it manifested itself, in the e2e tests between managers, because we were so far up the stack.
It should have been caught by comprehensive unit tests in the server-wallet package.

Why was it not caught by a comprehensive unit test? 
These [two lines](https://github.com/statechannels/statechannels/commit/37ed94c85ce984fdf583eef92e1250625c591565#diff-628e269381530e87d15f7308ff9f9bc669d27adaec648c675002768ec3f1f7cbR60-R61) indicate how the mental model of developers do not line up with correctness.

It is easy to see how such a subtle detail can be missed: the two lines of test, and [three lines of code](https://github.com/statechannels/statechannels/commit/37ed94c85ce984fdf583eef92e1250625c591565#diff-ecb671e7b0194f31631f7c3aa478c9a997dd7e8ae25822f845be16bfce7fe26dR114-R116), are a part of an incredibly complicated commit:

Screen Shot 2020-10-24 at 12 23 39 PM

That commit is part of an even more complex pull request:
Screen Shot 2020-10-24 at 12 25 16 PM


It is difficult to say with any level of certainty what should be done about that: I could say that the work should have been better planned, but the given feature was implemented under significant time pressure. Instead, I am simply aiming to highlight the events that lead to this bug.

## Observation 2
Why was it so difficult to pinpoint where in the stack the error was?
Answer: The logs when debugging this in the `payment`/`receipt`-manager e2e tests are insanely noisy and unreadable.

packages/server-wallet/src/protocols/open-channel.ts
snario, line 125:
```suggestion
  (isRunning(ps.app.supported) || ps.app.supported.turnNum === ps.app.participants.length * 2 - 1) &&
```

This is a typo right?

snario: Seems reasonable to me.

andrewgordstewart: This should have been a type error. The type of `ps.app.supported` appears to be `undefined` on this line though ...

Screen Shot 2020-10-24 at 12 35 29 PM

snario: Yeah there is a general code cleanup that can be done to these protocols to assert certain properties inside the smaller functions the `protocol` function is composed of that would ensure it is defined.

andrewgordstewart: This is really counterintuitive:
Screen Shot 2020-10-24 at 12 42 54 PM

snario: I think it has something to do with non-structural / nominal typing and maybe the lack of negated types

andrewgordstewart: typescript is structurally typed, and not nominally typed?

I didn't expect `barGuard(foobar: Foo | Bar): foobar is Bar` to change the type at all if `barGuard` returned `false`.

andrewgordstewart: This is perhaps an argument in favour of the `Option(State)` type instead of `State | undefined`?

andrewgordstewart: so the [example](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) in the docs is pretty similar to the `isSomething` example.

It seems, then, that the mistake is that `isRunning` is a _type guard_ that asserts that it is a `State` when it returns `true`. The compiler then narrows it down to the remaining allowed types, when it returns false, which is `undefined`.

Instead, `isRunning` should narrow it to a `RunningState` type, leaving open the possibility that it is a `NotRunningState | undefined`.

PR #2715, opened 2020-10-21
[server-wallet] 🧾 Ledger Channels

snario:
Replaces commit history of #2699 to be far more readable and cherry picks commits of other work I've split into separate pull requests.

PR #2743, opened 2020-10-25
📝💬🔏 Add two phase commit for ledger channel updates

snario:
Changes the implementation in #2715, which has each part sign a state of its proposed update immediately, this PR first ensures the two parties exchange a sequence of "commits", sharing what they propose the next ledger update be before actually signing it (or the intersection of it and the others commit). This implementation uses "requests", but it might be argued that it should be an objective instead.

### Pros
- Maintains that a Store need only store a single state per turn number

### Cons
- This adds networking overhead

### DIFF
--- a/packages/server-wallet/src/protocols/process-ledger-queue.ts
+++ b/packages/server-wallet/src/protocols/process-ledger-queue.ts
@@ -5,7 +5,6 @@ import {
   checkThat,
   isSimpleAllocation,
   Outcome,
-  SignedStateWithHash,
   SimpleAllocation,
   areAllocationItemsEqual,
   BN,
@@ -20,12 +19,15 @@ import {Protocol, ProtocolResult, ChannelState} from './state';
 import {
   MarkLedgerFundingRequestsAsComplete,
   noAction,
+  ProposeLedgerState,
   ProtocolAction,
   SignLedgerState,
 } from './actions';
 
 export type ProtocolState = {
   fundingChannel: ChannelState;
+  counterpartyLedgerCommit?: SimpleAllocation;
+  myProposedLedgerCommit?: SimpleAllocation;
   channelsRequestingFunds: ChannelState[];
   channelsReturningFunds: ChannelState[];
 };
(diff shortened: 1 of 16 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -490,32 +491,44 @@ export class Wallet extends EventEmitter<WalletEvent>
...
     );
 
-    const {channelResults, outbox} = await this.takeActions(channelIds);
+    let outbox: Outgoing[] = [];
+
+    if (wirePayload.requests && wirePayload.requests.length > 0)
+      // Modifies outbox, may append new messages
+      await Promise.all(wirePayload.requests.map(deserializeRequest).map(handleRequest(outbox)));
+
+    const runLoopResult = await this.takeActions(channelIds);
```
snario, line 521: Had to reverse the ordering of (1) take actions (2) handle requests. This might be the reason that using an objective is preferred, since objectives already have the ordering appropriate.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -811,6 +841,11 @@ function validateInvariants(stateVars: SignedStateVarsWithHash[], myAddress: str
 
   const turnNums = _.map(stateVars, s => s.turnNum);
 
+  const duplicateTurnNums = turnNums.some((t, i) => turnNums.indexOf(t) != i);
+  if (duplicateTurnNums) {
+    throw new StoreError(StoreError.reasons.duplicateTurnNums);
+  }
```
snario, line 847: Reverts 2c0aae271dfc9d03ccf8ef25b4458111f812b3de

packages/wire-format/src/generated-schema.json
```diff
@@ -58,21 +58,46 @@
...
+              "type": "string"
+            },
+            "outcome": {
+              "$ref": "#/definitions/Allocation"
+            },
+            "type": {
+              "const": "ProposeLedger",
+              "type": "string"
+            }
+          },
```
snario, line 92: The new message

snario:
TODO (probably on Monday) — update the push message test

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -102,114 +113,135 @@ const xorOutcome = (a: SimpleAllocation, b: SimpleAllocation): SimpleAllocation
   allocationItems: _.xorWith(a.allocationItems, b.allocationItems, areAllocationItemsEqual),
 });
 
-const computeNewOutcome = ({
+const mergeProposedLedgerUpdates = (
+  mine: SimpleAllocation,
+  theirs: SimpleAllocation,
```
andrewgordstewart, line 118: This is probably totally fine as is, but this could be easily extended to an `n`-party null channel.

I'm not sure if it's worth the abstraction, since the only `n`-party null channel we've come across is the joint channel when virtual funding, and for that, there is no chance of disagreement in the next state update.

It would be worth it if eg. we were to implement some token exchange in a directly funded, 3-party ledger channel.

PR #2745, opened 2020-10-26
feat: add closeChannels API to server wallet

snario:


packages/server-wallet/src/wallet/index.ts
```diff
@@ -421,6 +421,40 @@ export class Wallet extends EventEmitter<WalletEvent>
...
+        data: {targetChannelId: channelId, fundingStrategy: channel.fundingStrategy},
+      };
+
+      const objectiveToStore: ObjectiveStoredInDB = {
+        ...objective,
+        status: 'approved',
+        objectiveId: objectiveId(objective),
+      };
+
+      await ObjectiveModel.insert(objectiveToStore, this.knex);
```
tomclose, line 444: Should this use `this.store.addObjective`?

snario: Yes

packages/server-wallet/src/wallet/index.ts
```diff
@@ -391,6 +391,44 @@ export class Wallet extends EventEmitter<WalletEvent>
     return this.store.lockApp(channelId, criticalCode, handleMissingChannel);
   }
 
+  async closeChannels(channelIds: Bytes32[]): Promise<MultipleChannelOutput> {
```
kerzhner, line 394: Should this leverage the `closeChannel` api defined below?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -391,6 +391,44 @@ export class Wallet extends EventEmitter<WalletEvent>
...
+            CloseChannel.CloseChannelError.reasons.channelMissing,
+            {channelId}
+          );
+        }
+      );
+    }
+
+    const {channelResults, outbox} = await this.takeActions(channelIds);
+
+    (outbox[0].params.data as Payload).objectives = channelIds.map(channelId => ({
```
kerzhner, line 423: Is it possible the type assertion?

snario: The `Store` creates an outbound objective right now and the type of this is `unknown` which is strategically set (cc @tomclose) for outbound objects. One solution might be to put this line of code inside the store... but that would create an even bigger mess and contribute to the code smell of the store handling communications. I think this type assertion is OK for now.

packages/server-wallet/src/wallet/__test__/integration/close-channel.test.ts
```diff
@@ -57,3 +58,28 @@ it("reject when it's not my turn", async () => {
...
+      channelNonce: i,
+      vars: [stateWithHashSignedBy([alice(), bob()])(runningState)],
+    });
+    await Channel.query(w.knex).insert(c);
+    channelIds.push(c.channelId);
+  }
+
+  await expect(w.closeChannels(channelIds)).resolves.toMatchObject({
+    channelResults: channelIds.map(_ => ({
+      status: 'closing',
```
geoknee, line 80: Can you put the `channelId` in each object of this array?

packages/server-wallet/src/wallet/index.ts
geoknee: Seems to me that the wallet _should_ be using the `Payload` type. It should be `unknown` only from the point of view of the app / messaging layer (outside the wallet's interface).

geoknee (approved):

packages/server-wallet/src/wallet/index.ts
snario: There should probably be a cleanup of the wallet code and `takeActions` to account for this, maybe a new issue

PR #2747, opened 2020-10-26
refactor: encapsulate ObjectiveModel inside Store

snario:


tomclose (approved):

packages/server-wallet/src/wallet/index.ts
```diff
@@ -574,10 +562,7 @@ export class Wallet extends EventEmitter<WalletEvent>
               });
               return;
             case 'CompleteObjective':
-              if (objective.type === 'OpenChannel')
-                await ObjectiveModel.succeed(objective.objectiveId, tx);
-              if (objective.type === 'CloseChannel')
-                await ObjectiveModel.succeed(objective.objectiveId, tx);
+              await this.store.markObjectiveAsSucceeded(objective);
```
andrewgordstewart, line 565: This marks the objective as complete outside of a transaction, which can lead to a deadlock.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -404,6 +404,18 @@ export class Store {
...
+  }
+
+  async approveObjective(objective: ObjectiveStoredInDB, tx?: Transaction): Promise<void> {
+    await ObjectiveModel.approve(objective.objectiveId, tx || this.knex);
+  }
+
+  async markObjectiveAsSucceeded(objective: ObjectiveStoredInDB, tx?: Transaction): Promise<void> {
+    await ObjectiveModel.succeed(objective.objectiveId, tx || this.knex);
+  }
+
```
andrewgordstewart, line 418: To protect against deadlocks (such as the one this PR introduced) these methods could _require_ a transaction as a second argument, when appropriate.

packages/server-wallet/src/wallet/index.ts
tomclose: Could you explain how the deadlock happens? Is it just the case where there aren't enough threads in the transaction pool, or is there something else that can happen?

andrewgordstewart: You are
1. starting a transaction (acquires a connection)
2. executing a query (acquires a connection)
3. committing the transaction

if at (2), there are no connections available, you are blocked

If the entire connection pool is at (1), every transaction is blocked and none of them proceed.

(Even if this event never happens, the current code is an inefficient use of connections, which has a serious impact on performance.)

PR #2753, opened 2020-10-27
[server-wallet] Add withdrawal on conclusion

kerzhner:
For a directly funded channel, the server wallet requests the chain service to withdraw funds when:
- There is a conclusion proof.
- We are the first participant.


andrewgordstewart:
>     * We are the first participant.

Is this to avoid two people racing to conclude the channel? (If so, is that a problem?)

kerzhner:
> > ```
> > * We are the first participant.
> > ```
> 
> Is this to avoid two people racing to conclude the channel? (If so, is that a problem?)

Only one conclude + transfer all transaction needs to be submitted. This is a heuristic to determine which participant submits the transaction.

We can remove the heuristic and allow participants to race to transaction submission. In that case, if a participant submits the transaction and it fails, should the participant retry the transaction? Should the chain service have logic to determine whether one of the conclude + transfer all transactions has already succeeded?

andrewgordstewart:
> > > ```
> > > * We are the first participant.
> > > ```
> > 
> > 
> > Is this to avoid two people racing to conclude the channel? (If so, is that a problem?)
> 
> Only one conclude + transfer all transaction needs to be submitted. This is a heuristic to determine which participant submits the transaction.
> 
> We can remove the heuristic and allow participants to race to transaction submission. In that case, if a participant submits the transaction and it fails, should the participant retry the transaction? Should the chain service have logic to determine whether one of the conclude + transfer all transactions has already succeeded?

Yes, presumably the revert reason would indicate whether or not the peer has to retry.

I think the heuristic is fine for the moment. I don't see it as the correct long-term decision.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -1,52 +1,64 @@
...
+const rpcEndpoint = defaultTestConfig.rpcEndpoint;
+let provider: providers.JsonRpcProvider;
 const b = new Wallet({...defaultTestConfig, postgresDBName: 'TEST_B'});
 const a = new Wallet({...defaultTestConfig, postgresDBName: 'TEST_A'});
 
+const aAddress = '0x50Bcf60D1d63B7DD3DAF6331a688749dCBD65d96';
+const bAddress = '0x632d0b05c78A83cEd439D3bd6C710c4814D3a6db';
+
+const aFunding = BigNumber.from(1).toHexString();
+const bFunding = BigNumber.from(0).toHexString();
```
andrewgordstewart, line 21:
```suggestion
const aFunding = BN.from(1);
const bFunding = BN.from(0);
```

?

andrewgordstewart (approved):

packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
```diff
@@ -22,7 +22,7 @@ export async function up(knex: Knex): Promise<any> {
     ALTER TABLE ${channels}
     ADD CONSTRAINT valid_chain_service_requests
     CHECK (
-      ${chainServiceRequests} <@ ARRAY['fund']
+      ${chainServiceRequests} <@ ARRAY['fund', 'withdraw']
```
andrewgordstewart, line 25: Wait, you can't change this migration!

andrewgordstewart (changes requested):

packages/server-wallet/src/db/migrations/20200915140300_chain_service.ts
kerzhner: Nice catch! 7d9406d10

packages/server-wallet/src/db/migrations/20201027090000_chain-service-constraint.ts
```diff
@@ -0,0 +1,28 @@
...
+    ALTER TABLE ${channels}
+    ADD CONSTRAINT valid_chain_service_requests
+    CHECK (
+      ${chainServiceRequests} <@ ARRAY['fund', 'withdraw']
+    )
+  `);
+}
+
+export async function down(knex: Knex): Promise<any> {
+  await dropConstraint(knex);
```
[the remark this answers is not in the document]
kerzhner: That's a valid point, though in practice I find down migrations not very useful. eb3757a43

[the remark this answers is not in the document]
andrewgordstewart: Touche.

andrewgordstewart (approved):

PR #2756, opened 2020-10-27
fix: fix bug in allocateToTarget preventing funding with amount 0x00

snario:
If I had an outcome {A: 2, B: 0} and I wanted to fund (C1) {A: 1, B: 0} and (C2) {A: 1, B 0} then I would make two calls to `allocateToTarget`. The first would have produced {A: 0, C1: 1} and the second would have thrown an error (since B was not in this result and the destination was missing. With this change, the second call will produce {C1: 1, C2: 1} and each channel would be funded as needed.

### DIFF
--- a/packages/wallet-core/src/utils/outcome.ts
+++ b/packages/wallet-core/src/utils/outcome.ts
@@ -73,17 +73,19 @@ export function allocateToTarget(
   let total = Zero;
   let currentItems = currentOutcome.allocationItems;
 
-  deductions.forEach(targetItem => {
-    const ledgerItem = currentItems.find(i => i.destination === targetItem.destination);
-    if (!ledgerItem) {
-      throw new Error(Errors.DestinationMissing);
-    }
+  deductions
+    .filter(i => BN.gt(i.amount, 0))
+    .forEach(targetItem => {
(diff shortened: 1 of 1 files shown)


andrewgordstewart:
> If I had an outcome {A: 2, B: 0} and I wanted to fund (C1) {A: 1, B: 0} and (C2) {A: 1, B 0} then I would make two calls to `allocateToTarget`. The first would have produced {A: 0, C1: 1} and the second would have thrown an error (since B was not in this result and the destination was missing. With this change, the second call will produce {C1: 1, C2: 1} and each channel would be funded as needed.

Can you create this test case?

snario:
> > If I had an outcome {A: 2, B: 0} and I wanted to fund (C1) {A: 1, B: 0} and (C2) {A: 1, B 0} then I would make two calls to `allocateToTarget`. The first would have produced {A: 0, C1: 1} and the second would have thrown an error (since B was not in this result and the destination was missing. With this change, the second call will produce {C1: 1, C2: 1} and each channel would be funded as needed.
> 
> Can you create this test case?

ff3b3937e

andrewgordstewart (approved):

PR #2759, opened 2020-10-27
A few refactors to the Objective model

tomclose:
This PR:

* Renames `Objective -> ObjectiveModel` and `ObjectiveChannel -> ObjectiveChannelModel` to make the different types more explicit.
* Changes `extract` from being a standalone function that took an `ObjectiveModel` to being an instance method, `toObjective` on the `ObjectiveModel`.
* Moves and renames the `ObjectiveStoredInDB` model to `DBObjective` and adds a brief description.

These changes should hopefully make it less confusing to work with objectives 🤞 .

### DIFF
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -1,8 +1,6 @@
 import {objectiveId, Objective} from '@statechannels/wallet-core';
 import {Model, TransactionOrKnex} from 'objection';
 
-import {ObjectiveStoredInDB} from '../wallet/store';
-
 function extractReferencedChannels(objective: Objective): string[] {
   switch (objective.type) {
     case 'OpenChannel':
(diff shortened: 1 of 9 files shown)


packages/server-wallet/src/models/objective.ts
```diff
@@ -27,8 +17,16 @@ function extractReferencedChannels(objective: ObjectiveType): string[] {
...
 
-export class ObjectiveChannel extends Model {
-  readonly objectiveId!: ObjectiveStoredInDB['objectiveId'];
+/*
+  A DBObjective is a wire objective with a status and an objectiveId
+*/
+export type DBObjective = Objective & {
+  objectiveId: string;
+  status: 'pending' | 'approved' | 'rejected' | 'failed' | 'succeeded';
+};
```
geoknee, line 26: I thought your aim was to just have `ObjectiveModel` (formerly `Objective) be this type?

Also: moving the type seems good. Renaming it seems unecessary?

geoknee (approved):

packages/server-wallet/src/models/objective.ts
tomclose: That was the goal. The problem with that is that if you have just the `ObjectiveModel`, you don't get any type narrowing, because you don't have a union type. So it seems like we need another thing.

I'm open to a different name. I thought that `ObjectiveStoredInDB` was a bit confusing, as that suggested that it should be what the `ObjectiveModel` actually is. That said I'm not sure that `DBObjective` is much better. Maybe it should actually be `Objective` and the wallet-core objective should become `WireObjective`?

PR #2761, opened 2020-10-27
📝💬🔏 Unsigned consensus protocol before signing ledger channel updates

snario:
Changes the implementation in #2715, which has each part sign a state of its proposed update immediately, this PR first ensures the two parties exchange a sequence of "proposals", sharing what they propose the next ledger update be before actually signing it (or the intersection of it and the others proposal). This implementation uses "requests" in the wallet for the wire formatting, but it might be argued that it should be an objective instead.

### Details

https://github.com/statechannels/statechannels/blob/74e8dd13a7d7a9f8ad913458890d6d4c69b298f1/packages/server-wallet/src/protocols/process-ledger-queue.ts#L1-L39

### Pros
- Maintains that a Store need only store a single state per turn number

### Cons
- This adds networking overhead

### TODO

- [x] Ensure `SyncChannel` can also sync these proposals.
- [x] Handle outstanding ledger requests in O₁ / O₂ (Separate PR ?)
- [x] Unit tests


andrewgordstewart:
I'm closing this and putting it in the "ice box" column to revisit later.

andrewgordstewart:
Re-opening to see the potential merge conflicts.

snario:
I have been keeping rebasing this to keep it up to date every few days by the way

andrewgordstewart:
Meta-comment: I might be reluctant to continue to use "two phase commit" to describe this algorithm.

2PC is a [well-known algorithm ](https://en.wikipedia.org/wiki/Two-phase_commit_protocol)involving a _coordinator_ who
1. sends a message proposing a transaction and waits for confirmations from all nodes saying "I can commit this transaction"
2. when confirmation is received from all nodes, sends a message committing a transaction

I might have called this a 2-phase commit algorithm simply because it similarly has "two phases"
1. propose a ledger update, and wait for proposals from all other peers
2. when proposal is received from all other peers, compute the next ledger update (using the "intersection" of all the proposals) and sign the next ledger update

So it's different from 2PC in two ways:
1. there is no coordinator. each peer in a ledger channel can spontaneously propose an update
2. the _value_ proposed in the first phase is not always equal to the _value_ committed (signed) in the second phase.

Instead, the proposal was meant to be loosely modeled after the types of messages that are exchanged in 2PC.

andrewgordstewart:
The PR description does not describe how the protocol should when the two proposed ledger updates don't match.

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -5,7 +5,6 @@ import {
   checkThat,
```
andrewgordstewart, line 31: I don't see any unit tests for this file. Because of that, it's difficult to understand what's going on, since this file contains a large amount of complex code.

andrewgordstewart (commented):
checking in some comments before the next force-push

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -1,11 +1,36 @@
...
+ * Algorithm:
+ *
+ * IF there are any pending ledger updates (to fund OR defund a channel)
+ *
+ *   i. Check if I already proposed a new ledger update O₁. Otherwise, iteratively allocate funds
+ *      from the supported state's (i.e., S) outcome to each request, call this O₁.
+ *
+ *  ii. Check if there is an existing proposal for a ledger update, O₂.
+ *
+ * iii. Compute O₁ ⋂ O₂ (NOTE: O₂ might be NULL i.e., not received yet)
```
andrewgordstewart, line 16: If O2 is NULL, how should I interpret O1 \cap O2?

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -1,11 +1,36 @@
...
+ *
+ *      a. If I haven't sent my proposal yet, propose O₁ ⋂ O₂.
+ *
+ *      b. If I have sent my proposal and received theirs sign O₁ ⋂ O₂ at
+ *         the turn number 2 higher than the most recently supported one.
```
andrewgordstewart, line 21: Why increase the turn number by 2? Why not 1?

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -1,11 +1,36 @@
...
+ *
+ *      c. If O₂ is NULL (not received a proposal), wait for the counterparty.
```
andrewgordstewart, line 23: This allows for a deadlock:
1. my message with my proposal got dropped
2. their message with their proposal got dropped

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -1,11 +1,36 @@
...
+ * Follows an algorithm to guarantee that an agreed upon ledger update will be signed by
+ * 2 participants in at most 2 round trips. Either can propose a state at any time and either
+ * can counterpropose a state at any time. However, if they follow the algorithm below they
+ * should always be able to compute the expected outcome by the second round trip.
+ *
+ * Algorithm:
+ *
+ * IF there are any pending ledger updates (to fund OR defund a channel)
+ *
+ *   i. Check if I already proposed a new ledger update O₁. Otherwise, iteratively allocate funds
```
andrewgordstewart, line 11: How should the requests be ordered?

andrewgordstewart (commented):
Add some questions.

packages/server-wallet/src/protocols/process-ledger-queue.ts
snario: They're sorted by channel nonce (proxy for creation time) at the moment

packages/server-wallet/src/protocols/process-ledger-queue.ts
snario: Actually this should read `n` — but in fact 1 also works. Can change.

packages/server-wallet/src/protocols/process-ledger-queue.ts
snario: O1

packages/server-wallet/src/models/channel.ts
```diff
@@ -81,6 +82,9 @@ export class Channel extends Model implements RequiredColumns {
   readonly assetHolderAddress!: string; // only Ledger channels have this
   readonly fundingLedgerChannelId!: Bytes32; // only App channels funded by Ledger have this
 
+  readonly myUnsignedCommitment: Outcome | undefined; // only Ledger channels have this
+  readonly theirUnsignedCommitment: Outcome | undefined; // only Ledger channels have this
+
```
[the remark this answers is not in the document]
snario: https://github.com/statechannels/statechannels/issues/3059

packages/server-wallet/src/protocols/process-ledger-queue.ts
snario: Started a related discussion here: https://github.com/statechannels/statechannels/issues/3057

One strategy could be to automatically trigger a channel sync if this stage of the protocol is reached. I'm not opposed to doing that.

andrewgordstewart: > Started a related discussion here: #3057

👍 

I think we need some tighter organization of this project. Creating issues is good, but we ought to have a "checklist" of things that need to get done to ship a robust ledger funding protocol.

> One strategy could be to automatically trigger a channel sync if this stage of the protocol is reached. I'm not opposed to doing that.

I think this is the most natural solution here. It is a bit complex, because you want to ensure you don't enter an infinite loop in the case where channel syncing fails to change state on either end.

lalexgap: > I think we need some tighter organization of this project. Creating issues is good, but we ought to have a "checklist" of things that need to get done to ship a robust ledger funding protocol.
> 

I've created some labels for challenging-v0 to tag issues so maybe the same thing could work here?

andrewgordstewart: I like that suggestion, and I'd be keen to give that process a try for one or two projects.

snario:
Reminder for myself: I think ledger proposals need a nonce

snario:
In 20443ec I added the concept of "proposal nonces". These are locally managed monotonically increasing integers that any particular wallet uses on a per-ledger channel basis. 

### Motivation
It is possible to receive a proposal over the network which is not the latest proposal your counterparty has proposed. It may be referring to some set of channels being funded which were computed before one wallet had full information about its outstanding ledger requests.

### Solution
Add a nonce on each proposal so that if a wallet receives a stale proposal, it ignores it.

### Properties
- Each ledger channel has "my proposal nonce" and "their proposal nonce"
- Starts at 0 for each
- Whenever my wallet creates a new proposal, it increments my proposal nonce by 1
- Whenever my wallet receives a ledger proposal, it **only inserts it into the DB if their nonce is greater than the last one I received from them**

snario:
Added unit tests
```
 PASS  src/protocols/__test__/process-ledger-queue.test.ts (28.124 s)
  marking ledger requests as complete
    ✓ takes no action if there are no ledger requests (2 ms)
    ✓ detects completed funding requests from the outcome of supported state (638 ms)
    ✓ detects completed defunding requests from the outcome of supported state (149 ms)
  exchanging ledger proposals
    as initial proposer
      ✓ takes no action if proposal is identical to existing supported outcome (452 ms)
      ✓ takes no action if proposal exists but counterparties does not (464 ms)
      ✓ proposes new outcome funding 1 channel (462 ms)
      ✓ proposes new outcome funding many channels (2428 ms)
      ✓ proposes new outcome exhausting 100% of funds (505 ms)
      ✓ proposes new outcome requiring defund before having sufficient funds (1451 ms)
      ✓ makes no proposal but identifies channel when ledger has insufficient funds (485 ms)
      ✓ proposes outcome funding some channels, identifying insufficient funds for others (2336 ms)
    as responding proposer
      ✓ will propose identical proposal to counterparty with same requests (483 ms)
      ✓ will propose intersected proposal to counterparty with superset of requests (950 ms)
      ✓ will not propose new state if intersection is identical to supported outcome (945 ms)
  exchanging signed ledger state updates
    as initial signer
      ✓ does not sign ledger update if one has already been signed by me (748 ms)
      ✓ generates signed ledger update when proposals were identical (456 ms)
      ✓ generates signed ledger update when proposals were not identical but overlapped (1443 ms)
      ✓ dismisses proposals when the intersected result is identical to supported (938 ms)
    as responding signer
      ✓ throws an error if counterparty signed update does not follow protocol (790 ms)
```

andrewgordstewart:
> In [20443ec](https://github.com/statechannels/statechannels/commit/20443ec457642da2837974de047bf34885e9d075) I added the concept of "proposal nonces". These are locally managed monotonically increasing integers that any particular wallet uses on a per-ledger channel basis.
> ### Motivation
> 
> It is possible to receive a proposal over the network which is not the latest proposal your counterparty has proposed. It may be referring to some set of channels being funded which were computed before one wallet had full information about its outstanding ledger requests.
> ### Solution
> 
> Add a nonce on each proposal so that if a wallet receives a stale proposal, it ignores it.
> ### Properties
> 
>     * Each ledger channel has "my proposal nonce" and "their proposal nonce"
> 
>     * Starts at 0 for each
> 
>     * Whenever my wallet creates a new proposal, it increments my proposal nonce by 1
> 
>     * Whenever my wallet receives a ledger proposal, it **only inserts it into the DB if their nonce is greater than the last one I received from them**

Is there a reason the proposal can't be "the next state" (which already has a nonce -- `turnNum`)?

packages/server-wallet/src/db/migrations/20201201095501_ledger_proposals.ts
```diff
@@ -0,0 +1,27 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<any> {
+  await knex.schema.table('channels', function(table) {
```
andrewgordstewart, line 4: request: instead of placing two-party-oriented data on `channels`, there ought to be a new `ledger_proposals` table with this rough schema:
```
ledger_id: foreign key to channels
proposal: jsonb
nonce: integer
signing_address: text
```

this allows the DB to store data for `n`-party ledger channels (such as the joint channel in a virtual channel construction!)

packages/server-wallet/src/protocols/__test__/process-ledger-queue.test.ts
```diff
@@ -0,0 +1,380 @@
...
+    it('takes no action if proposal is identical to existing supported outcome', () => {
+      const requestChannel = prefundChannelWithAllocations(); // Empty outcome
+      const protocolArgs = {
+        fundingChannel: defaultLedgerChannel,
+        channelsRequestingFunds: [requestChannel.protocolState],
+      };
+      expect(protocol(processLedgerQueueProtocolState(protocolArgs))).toBeUndefined();
+    });
+
+    it('takes no action if proposal exists but counterparties does not', () => {
```
andrewgordstewart, line 125: Should it request a proposal from the peer in this case?

packages/wire-format/src/types.ts
```diff
@@ -142,8 +142,9 @@ export const isVirtuallyFund = guard<VirtuallyFund>('VirtuallyFund');
 export const isFundGuarantor = guard<FundGuarantor>('FundGuarantor');
 
 // channel requests
-type GetChannel = {type: 'GetChannel'; channelId: Bytes32};
-export type ChannelRequest = GetChannel;
+type GetChannel = {type: 'GetChannel'; channelId: string};
+type ProposeLedger = {type: 'ProposeLedger'; nonce: number; channelId: string; outcome: Outcome};
```
andrewgordstewart, line 146: Shouldn't this be `ProposeLedgerUpdate`? You're not proposing a _ledger_, you're proposing the next outcome in an existing ledger.

packages/server-wallet/src/protocols/__test__/process-ledger-queue.test.ts
```diff
@@ -0,0 +1,380 @@
...
+  theirLedgerProposal: undefined,
+  theirLedgerProposalNonce: 0,
+
+  myLedgerProposal: undefined,
+  myLedgerProposalNonce: 0,
+
+  ...args,
+});
+
+describe('marking ledger requests as complete', () => {
```
andrewgordstewart, line 80: I didn't read these test cases carefully, but I skimmed them, and they generally make sense to me.

I have the feeling that there are edge cases that can lead to deadlocks. (But we're aware of deadlocks that are fixed by this change.) I'd like to explore some strategies for detecting potential issues with a long-running randomized test.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -200,6 +203,13 @@ export class SingleThreadedWallet extends EventEmitter<EventEmitterType>
 
     response.queueChannelRequest(channelId, myIndex, participants);
     response.queueChannelState(channelState);
+
+    if (await this.store.isLedger(channelId)) {
```
andrewgordstewart, line 207: I don't think I have the contextual awareness to actually review this component anymore 😞

andrewgordstewart (changes requested):
This is a cursory review requesting one change (to the DB schema) before approving the merge into master.

It's difficult to properly review a PR of this size and complexity. In the future, this type of change would probably benefit from a couple of rounds of "planning discussions" so that we can agree on a rough plan. For example, we could have avoided the proposed modification of the `channels` table, and gone straight for a `ledger_proposal` table. Or, we could have landed on this `LedgerProposal` type:
```
type LedgerProposal = {nonce: number; defund: AllocationItem[]; fund: AllocationItem[]};
```

which is amenable to a simpler algorithm description and implementation.

Let's plan to make these simplifications in the future.

packages/server-wallet/src/models/ledger-proposal.ts
```diff
@@ -0,0 +1,74 @@
...
+    await LedgerProposal.query(txOrKnex)
+      .where({channelId})
+      .patch({proposal: null});
+  }
+
+  static async storeProposal(
+    {channelId, signingAddress, proposal, nonce}: RequiredColumns,
+    txOrKnex: TransactionOrKnex
+  ): Promise<void> {
+    // TODO: Do this in a single SQL query (could not figure out how in time)
```
snario, line 51: Could use help if anyone knows how. Tried 2 strategies:

Strategy 1 — raw SQL
```ts
await (tx || this.knex).raw(`
  INSERT INTO ${LedgerProposal.tableName}
    (channel_id, signing_address, proposal, nonce)
      values ("${channelId}", "${signingAddress}", ${JSON.stringify(proposal)}, ${nonce}
  ON CONFLICT (channel_id, signing_address)       //   ^-- unsure how to do jsonb insertion
  DO
    UPDATE
    WHERE proposal IS NOT NULL
    AND nonce > ${nonce}
    SET proposal = "${JSON.stringify(proposal)}", nonce = ${nonce};`);
```

Strategy 2 — use .onConflict() API
```ts

await (tx || this.knex)(LedgerProposal.tableName)
  .insert({channelId, signingAddress, proposal, nonce})
  .onConflict(['channelId', 'signingAddress'])
  .merge({proposal, nonce})
  .whereNull('proposal')           //   <-- I don't think this applies in the
  .andWhere('nonce', '<', nonce);  //          right place for the query
```

packages/server-wallet/src/protocols/__test__/process-ledger-queue.test.ts
snario: The _wallet_ sends the proposal in this case, but the _protocol_ does not. The reason why is that there would need to be some state for the _protocol_ to be able to know "I have decided to request a proposal from my peer", so that the run loop can terminate.

packages/server-wallet/src/db/migrations/20201201095501_ledger_proposals.ts
```diff
@@ -0,0 +1,25 @@
...
+      .defaultTo(0)
+      .unsigned();
+
+    table.index('channel_id');
+
+    table.unique(['channel_id', 'signing_address']);
+
+    table.foreign('channel_id').references('channels.channel_id');
+  });
+}
```
snario, line 21: New schema ☝️

packages/server-wallet/src/db/migrations/20201201095501_ledger_proposals.ts
```diff
@@ -1,27 +1,25 @@
...
-  await knex.schema.table('channels', function(table) {
-    table.jsonb('my_ledger_proposal');
-    table.jsonb('their_ledger_proposal');
-    table
-      .integer('my_ledger_proposal_nonce')
-      .notNullable()
-      .defaultTo(0)
-      .unsigned();
+  await knex.schema.createTable('ledger_proposals', table => {
+    table.string('channel_id').notNullable();
```
andrewgordstewart, line 5: Should this be the table's primary key?

snario: I don't think so because there will be `numParticipants` number of rows

andrewgordstewart: Right. I guess `['channel_id', 'signing_address']` might be a natural private key.

packages/server-wallet/src/models/ledger-proposal.ts
```diff
@@ -0,0 +1,74 @@
+import {JSONSchema, Model, TransactionOrKnex} from 'objection';
+import {Address, SimpleAllocation} from '@statechannels/wallet-core';
+
+import {Bytes32} from '../type-aliases';
+
+interface RequiredColumns {
+  readonly channelId: Bytes32;
+  readonly proposal: SimpleAllocation | null;
```
andrewgordstewart, line 8: Why should this column ever take the value `null`?

snario: When we want to explicitly remove the proposal, which is an indication of the protocol having "dealt with" it

packages/server-wallet/src/models/ledger-proposal.ts
andrewgordstewart: Are you saying that
1. fails because you are not inserting a jsonb column correctly
2. fails because you think the `whereNull` is not in the correct place
?

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -55,17 +55,15 @@ import {
...
   channelsRequestingFunds: ChannelState[];
   channelsReturningFunds: ChannelState[];
 };
 
 type ProtocolStateWithCommits = ProtocolState & {
-  theirLedgerProposal: SimpleAllocation;
-  myLedgerProposal: SimpleAllocation;
+  theirLedgerProposal: {proposal?: SimpleAllocation; nonce: number};
+  myLedgerProposal: {proposal?: SimpleAllocation; nonce: number};
 };
```
andrewgordstewart: The types here are confusing.
- Sometimes `proposal` must be present, but it may have the `null` value
- Other times, `proposal` may be `undefined`

It smells bad that the presence of a `null` value contains different information than the absence of the key.

packages/server-wallet/src/models/ledger-proposal.ts
andrewgordstewart: Why not have a `dealth_with: boolean` column? (by a more meaningful name)

packages/server-wallet/src/models/ledger-proposal.ts
snario: (1) I wasn't sure how to write the query to insert an `object` and coerce it into a `jsonb` literal inside the `INSERT` and `SET` command, so I tried using the Knex API but

(2) It didn't seem like the ORM created the right SQL query to do what I want (which is what the raw SQL query was doing), so two concurrency related tests failed

—

I simply stopped trying and wrote two queries with the nonce check done in JS. Have not revisited this since last night

packages/server-wallet/src/protocols/process-ledger-queue.ts
snario: You're right, this is a typo — this type was meant to have definitely not null or undefined types

packages/server-wallet/src/models/ledger-proposal.ts
snario: We can certainly do this, I guess I had preferred deleting data we no longer needed

andrewgordstewart: We can still delete rows where `dismissed: true`, if we wish. It would be sensible to do so when the next state becomes supported, for instance.

packages/server-wallet/src/models/ledger-proposal.ts
andrewgordstewart: The issue goes away with the "soft-delete" strategy of having a `dismissed: boolean` column, since then you simply insert a new row for new proposals.

andrewgordstewart (approved):

PR #2762, opened 2020-10-27
[server-wallet] iteration on conclude and withdraw

kerzhner:
The two major changes are:
1. The server wallet runloop does not wait for chain service calls. See https://github.com/statechannels/statechannels/commit/e68ea1596b5da9d122c3cef4c3a65d7df172a464 for reasoning.
1. all participants submit conclude and withdraw transactions. Addresses https://github.com/statechannels/statechannels/pull/2753#issuecomment-717302221

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -565,6 +565,7 @@ export class Wallet extends EventEmitter<WalletEvent>
             case 'Withdraw':
               await this.store.addChainServiceRequest(action.channelId, 'withdraw', tx);
               // Note, this returns a promise. We are NOT waiting on the promise
+              // app.supported is defined (if the wallet is functioning correctly), but the compiler is not aware of that
               this.chainService.concludeAndWithdraw(app.supported ? [app.supported] : []);
               return;
             default:
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -551,8 +551,7 @@ export class Wallet extends EventEmitter<WalletEvent>
             }
             case 'FundChannel':
               await this.store.addChainServiceRequest(action.channelId, 'fund', tx);
-              // Note, this returns a promise. We are NOT waiting on the promise
-              this.chainService.fundChannel({
+              await this.chainService.fundChannel({
                 ...action,
                 expectedHeld: BN.from(action.expectedHeld),
                 amount: BN.from(action.amount),
@@ -564,9 +563,9 @@ export class Wallet extends EventEmitter<WalletEvent>
               return;
             case 'Withdraw':
               await this.store.addChainServiceRequest(action.channelId, 'withdraw', tx);
-              // Note, this returns a promise. We are NOT waiting on the promise
               // app.supported is defined (if the wallet is functioning correctly), but the compiler is not aware of that
-              this.chainService.concludeAndWithdraw([app.supported!]);
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              await this.chainService.concludeAndWithdraw([app.supported!]);
               return;
             default:
               throw 'Unimplemented';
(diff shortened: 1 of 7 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -563,7 +564,8 @@ export class Wallet extends EventEmitter<WalletEvent>
               return;
             case 'Withdraw':
               await this.store.addChainServiceRequest(action.channelId, 'withdraw', tx);
-              await this.chainService.concludeAndWithdraw(app.supported ? [app.supported] : []);
+              // Note, this returns a promise. We are NOT waiting on the promise
+              this.chainService.concludeAndWithdraw(app.supported ? [app.supported] : []);
```
andrewgordstewart, line 568: Is the ternary operator here just to appease the compiler? IE. is `app.supported` known to be defined at this point?
```suggestion
              this.chainService.concludeAndWithdraw([app.supported!]);
```

kerzhner: If the wallet is functioning correctly, `app.supported` should be defined. Clarified here: 0bbd70baf.

Another option is to throw an error if app.supported is undefined. The non-null assertion you suggest also looks good, but, in general, I try to avoid those.

andrewgordstewart:
> The implication is that a failing concludeAndWithdraw will prevent the CloseObjective from completing. We probably don't want that.

I don't follow this argument.
1. If my wallet failed to _submit_ a `concludeAndWithdraw` transaction, I certainly wouldn't expect the objective to complete.
2. If the `concludeAndWithdraw` transaction was submitted but then _failed_, there is an acceptable reason for the objective to complete: the channel was already finalized.
  a. If this is the case, then it's ok for the objective to be completed.
  b. If it's a different error, I might not expect the DB transaction to abort, but I _also_ do not expect the objective to be completed.

Isn't it the case that you want to
1. try to submit the transaction. If it fails, handle that error
2. When the transaction is mined ... it doesn't matter, does it? Shouldn't the chain watcher push a `channelFinalized` event into the wallet?

packages/server-wallet/src/wallet/index.ts
andrewgordstewart: But in this case, if `app.supported` happens to be `undefined`, the code is misbehaving, silently, which seems really bad.

At least the assertion would result in a runtime error. I agree with the rule that we should avoid them (or maybe not even have them).

kerzhner:
On second thought, I remember that the wallet must await chain service calls that return a [transaction response](https://docs.ethers.io/v5/single-page/#/v5/api/providers/types/-%23-providers-TransactionResponse). Otherwise, we are opening. Otherwise, we are likely to get nonce errors: https://github.com/ethers-io/ethers.js/issues/972#event-3856427649.

Chain service can implement a transaction submission queue to get around this restriction.

packages/server-wallet/src/wallet/index.ts
kerzhner: That reasoning makes sense to me.

andrewgordstewart:
> I think an open question is: when does the `CloseChannel` objective complete for a directly funded channel? The options are:
> 
> 1. When the final state is signed by all participants (this is currently implemented in this PR).
> 2. When `concludeAndWithdraw` transaction succeeds.
> 3. When `concludeAndWithdraw` transaction is successfully submitted (this is implemented before the PR).
> 
> I would vote **against** option 3. With option 3, the `CloseChannel` objective can succeed regardless of whether the transaction ends up getting mined. Which is also the case for option 1. But option 1 allows the run loop to not wait for network requests.
> 
> I am open to implementing option 2 if we think that is the right strategy.

I think it should be
4. When a `AssetTransferred` event is pushed into the wallet, indicating that assets are transferred out of the channel.

kerzhner:
> I think it should be
> 4. When a `AssetTransferred` event is pushed into the wallet, indicating that assets are transferred out of the channel.

Good call, I agree with you on that. That work probably doesn't belong in this PR. At this point, this PR implements conclude transaction submission by every participant.

andrewgordstewart (approved):

kerzhner:
> On second thought, I remember that the wallet must await chain service calls that return a [transaction response](https://docs.ethers.io/v5/single-page/#/v5/api/providers/types/-%23-providers-TransactionResponse). Otherwise, we are likely to get nonce errors: [ethers-io/ethers.js#972 (comment)](https://github.com/ethers-io/ethers.js/issues/972#event-3856427649).
> 
> Chain service can implement a transaction submission queue to get around this restriction.

I retract this statement. Transaction queue has already been implemented in the chain service, so it safe not to await chain service api calls.

PR #2768, opened 2020-10-29
refactor: rewrite retrieveFundsFromClosedChannels to be clearer to read

snario:


packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -30,27 +31,42 @@ export type ProtocolState = {
...
+  if (!channel.supported) throw new Error('state is unsupported');
+
+  const {allocationItems: channelAllocations} = checkThat(
+    channel.supported.outcome,
+    isSimpleAllocation
+  );
+
+  const [removed, remaining] = _.partition(allocationItems, ['destination', channel.channelId]);
+
+  if (removed.length !== 1) throw 'Expected to find exactly one item';
```
andrewgordstewart, line 47:
```suggestion
  if (removed.length !== 1) throw new Error('Expected to find exactly one item');
```

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -30,27 +31,42 @@ export type ProtocolState = {
...
+
+  if (removed[0].amount !== channelAllocations.map(x => x.amount).reduce(BN.add, BN.from(0)))
+    throw 'Bookkeeping error';
```
andrewgordstewart, line 50: Can we use a more appropriate error? I did not intend for this message to be used verbatim, it was meant as a placeholder.

snario: We don't really have a known error type for this, so I might just keep it for now. I suppose it would fit inside some category of "protocol errors"?

andrewgordstewart: I just mean `'Bookkeeping error'` is not very informative, but maybe it's fine. It should at least be `Error('Bookkeeping error')` though.

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -30,27 +31,42 @@ export type ProtocolState = {
...
+    channel.supported.outcome,
+    isSimpleAllocation
+  );
+
+  const [removed, remaining] = _.partition(allocationItems, ['destination', channel.channelId]);
+
+  if (removed.length !== 1) throw new Error('Expected to find exactly one item');
+
+  if (removed[0].amount !== channelAllocations.map(x => x.amount).reduce(BN.add, BN.from(0)))
+    throw 'Bookkeeping error';
```
snario, line 50:
```suggestion
    throw new Error('Expected outcome allocations to add up to the allocation in ledger');
```

andrewgordstewart (approved):

PR #2770, opened 2020-10-29
refactor: do not bind class methods

geoknee:
I think this is mostly a question of style, but it seems that the current code is unnecessarily eye-catching and unexpected to the majority.

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -131,28 +131,6 @@ export class Wallet extends EventEmitter<WalletEvent>
       this.walletConfig.chainNetworkID
     );
 
-    // Bind methods to class instance
-    this.getParticipant = this.getParticipant.bind(this);
-    this.updateChannelFunding = this.updateChannelFunding.bind(this);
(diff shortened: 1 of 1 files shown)


tomclose (approved):

andrewgordstewart:
For the record, the intention of this question was not to argue for or against binding instance methods. It was simply to understand _why we need to bind methods_, ie. to see an explanation like https://github.com/statechannels/statechannels/commit/7a9689fd13ad12dca97c8777208859fa7407db94#r43705018

A comment such as
```
    /*
    Binding methods to the class instance makes it slightly easier to use class methods in callbacks.
    If we don't bind methods to the class instance, than this results in a runtime error, due to the semantics
    of the `this` keyword in javascript:
    `channelResults.map(this.registerChannelWithChainService)`

    By calling
    `this.registerChannelWithChainService = this.this.registerChannelWithChainService.bind(this)`
    in the method constructor, `channelResults.map(this.registerChannelWithChainService)` works as a developer
    might expect.

    For further context, see eg. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind
    */
```

living in the code would make it clear to a developer why methods are bound as such.

PR #2776, opened 2020-10-29
chore: add new database indexes for channels table

snario:
[Direct quote]

> I propose we add an index in the server-wallet’s PostgreSQL channels table — in addition to channelId — on participants.
> Then, we can do this.wallet.getChannels(participants). If we also index on assetHolderAddress then we can make this.wallet.getLedgerChannel(assetHolderAddress, participants) and O(1) call. Which is an alias for getLedgerChannelForAllocation (since participants is [gateway,allocation.id]).

### DIFF
--- a/packages/server-wallet/src/db/migrations/20201027090001_channels_indexes.ts
+++ b/packages/server-wallet/src/db/migrations/20201027090001_channels_indexes.ts
@@ -1,15 +1,15 @@
 import * as Knex from 'knex';
 
+const COLUMNS = ['assetHolderAddress', 'participants'];
+
 export async function up(knex: Knex): Promise<any> {
   await knex.schema.alterTable('channels', function(t) {
-    t.index('assetHolderAddress');
-    t.index('participants');
+    t.index(COLUMNS);
   });
 }
 
 export async function down(knex: Knex): Promise<any> {
   await knex.schema.alterTable('channels', function(t) {
-    t.dropIndex('assetHolderAddress');
-    t.dropIndex('participants');
+    t.dropIndex(COLUMNS);
   });
 }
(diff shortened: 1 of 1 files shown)


packages/server-wallet/src/db/migrations/20201027090001_channels_indexes.ts
```diff
@@ -0,0 +1,15 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<any> {
+  await knex.schema.alterTable('channels', function(t) {
+    t.index('assetHolderAddress');
+    t.index('participants');
```
andrewgordstewart, line 6: This should probably use a [multi-column index](https://www.postgresql.org/docs/12/indexes-multicolumn.html), considering the desired application:
> I propose we add an index in the server-wallet’s PostgreSQL channels table — in addition to channelId — on participants.
> Then, we can do this.wallet.getChannels(participants). If we also index on assetHolderAddress then we can make this.wallet.getLedgerChannel(assetHolderAddress, participants) and O(1) call. Which is an alias for getLedgerChannelForAllocation (since participants is [gateway,allocation.id]).

Let me check that this is the correct syntax
```suggestion
    t.index(['assetHolderAddress', 'participants']);
```

andrewgordstewart: If you take this suggestion, then there should be a file-level constant `const COLUMNS = ['assetHolderAddress', 'participants'];, and then this can be
```suggestion
    t.index(COLUMNS);
```

so that you can `t.dropIndex(COLUMNS)` below.

andrewgordstewart: Note that an index on `['firstCol', 'secondCol']` is still comparably effective to an index just on `firstCol` for queries that filter `WHERE firstCol = $1`. But it's more efficient for queries that filter `WHERE firstCol = $1 AND secondCol = $2`.

snario: e3910893f

andrewgordstewart (approved):

PR #2777, opened 2020-10-29
feat: enhanced getLedgerChannels API with assetHolder and participants query args

snario:


packages/server-wallet/src/models/channel.ts
```diff
@@ -144,10 +144,14 @@ export class Channel extends Model implements RequiredColumns {
...
 
-  static allLedgerChannels(txOrKnex: TransactionOrKnex): Promise<Channel[]> {
+  static getLedgerChannels(
+    assetHolderAddress: string,
+    participants: Participant[],
+    txOrKnex: TransactionOrKnex
+  ): Promise<Channel[]> {
     return Channel.query(txOrKnex)
       .select()
-      .whereNotNull('assetHolderAddress');
```
andrewgordstewart, line 150: The before looks incorrect here, but maybe I am not familiar with the wallet schema?

andrewgordstewart (approved):

packages/server-wallet/src/models/channel.ts
snario: `assetHolderAddress` is being set to `NULL` for channels which are not ledger channels (cc @geoknee)

PR #2781, opened 2020-10-30
refactor: introduce ObjectiveManager class

tomclose:


packages/server-wallet/src/wallet/store.ts
```diff
@@ -431,6 +431,11 @@ export class Store {
     await AppBytecode.upsertBytecode(chainNetworkId, appDefinition, bytecode, this.knex);
   }
 
+  async getObjective(objectiveId: string, tx?: TransactionOrKnex): Promise<DBObjective> {
+    tx = tx || this.knex; // todo: make tx required
```
geoknee, line 435: Is this TODO now?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -117,6 +114,7 @@ export class Wallet extends EventEmitter<WalletEvent>
   knex: Knex;
   store: Store;
   chainService: ChainServiceInterface;
+  objectiveManager: ObjectiveManager;
```
geoknee, line 117: Perhaps let's rename `manager` to `workerManager` just above.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -169,6 +167,13 @@ export class Wallet extends EventEmitter<WalletEvent>
       );
       this.chainService = new MockChainService();
     }
+
+    this.objectiveManager = ObjectiveManager.create({
+      store: this.store,
+      chainService: this.chainService,
+      logger,
+      timingMetrics: !!walletConfig?.timingMetrics,
```
geoknee, line 175:
```suggestion
      timingMetrics: this.walletConfig.timingMetrics,
```

packages/server-wallet/src/wallet/store.ts
tomclose: It was really meant for a later date, when we make the store more consistent.

geoknee (commented):
I think this is looking good so far. There are quite a few TODOs to clear up.

Also the PR title isn't quite accurate. The `Wallet` still owns a `crankToCompletion` method, but that now calls into a new `ObjectiveManager.crank()`. 

FWIW I think `crankToCompletion` could be renamed to `crankToIdle` or similar. This would help disassociate "doing several cranks on each approved objective until we are finished for the moment with all of them" (accurate) from "keep cranking until all the objectives succeed" (incorrect).

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+    let channelToLock;
+
+    if (objective.type === 'OpenChannel' || objective.type === 'CloseChannel')
+      channelToLock = objective.data.targetChannelId;
+    else throw new Error('crankToCompletion: unsupported objective');
+
+    let attemptAnotherProtocolStep = true;
+
+    while (attemptAnotherProtocolStep) {
+      await this.store.lockApp(channelToLock, async tx => {
```
snario, line 64: I feel like the `lockApp` method might be getting outdated as the paradigm moves forward. In reality, we want to lock all rows related to the `protocolState` that the objective uses, I think. Maybe we can replace `lockApp` with just a simple call to `knex.transaction` or an alias to that e.g.,`store.withinTransaction`. We aren't even using the second argument in the callback anymore (e.g., after `tx` there is a 2nd argument, `channel`, that is omitted on this line 64).

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+            }
+            case 'RequestLedgerDefunding': {
+              const ledgerChannelId = await determineWhichLedgerToUse(protocolState.app, tx);
+              await LedgerRequest.requestLedgerDefunding(
+                protocolState.app.channelId,
+                ledgerChannelId,
+                tx
+              );
+              return;
+            }
```
snario, line 134: If we're in a refactoring mood, I feel like these could be combined into one

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+        }
+        return ledgerChannelId;
+      }
+      throw new Error('ledger funded app has no fundingLedgerChannelId');
+    }
+    throw new Error('cannot fund app that is not ledger funded');
+  } else {
+    throw new Error('cannot fund unsupported app');
+  }
+};
```
snario, line 197: This entire function basically amount to `return channel.fundingLedgerChannelId`. Feel free to destroy it — it's mostly an artifact of history pre- #2760

packages/server-wallet/src/wallet/index.ts
```diff
@@ -666,121 +672,25 @@ export class Wallet extends EventEmitter<WalletEvent>
       .filter(x => x !== undefined)
       .filter(o => o?.status === 'approved');
 
+    // todo(tom): why isn't this just a for loop?
```
snario, line 675: ¯\_(ツ)_/¯

packages/server-wallet/src/wallet/index.ts
```diff
@@ -654,7 +659,8 @@ export class Wallet extends EventEmitter<WalletEvent>
     return requiresAnotherCrankUponCompletion;
   }
 
-  private async crankToCompletion(
+  // todo(tom): change function to return a value instead of mutating input args
```
snario, line 662: 👍

packages/server-wallet/src/wallet/index.ts
```diff
@@ -666,121 +672,25 @@ export class Wallet extends EventEmitter<WalletEvent>
...
+      // add channel result
+      channelResults.push(...newChannelResults);
+      outbox.push(...newOutbox);
 
-        const nextAction = recordFunctionMetrics(
-          executeProtocol(),
-          this.walletConfig.timingMetrics
-        );
+      // todo(tom): this how the code behaved previously. Is it actually what we want?
+      error = newError;
```
snario, line 690: Do we have any tests for error cases within the run loop?

packages/server-wallet/src/objectives/objective-manager.ts
tomclose: Totally agree with this. Was going to tackle this in another PR.

packages/server-wallet/src/objectives/objective-manager.ts
tomclose: I was intending to move this down into an objective model shortly.

packages/server-wallet/src/objectives/objective-manager.ts
andrewgordstewart: Lock app does a really important thing: it synchronizes access to the application.

There needs to be a carefully thought out set of database queries that make updates. There used to be one, which updated a single row and acquired a lock on that row to synchronize that behaviour.

If we are updating multiple rows in multiple relations within a tx, we need to ensure it cannot create a deadlock. If it only needs to update one row that is known in advance, we can avoid locking other rows, but then we have to consider the tx isolation level if there is concurrent updates to other rows read by the transaction, and then we need to expect random Postgres errors

packages/server-wallet/src/objectives/objective-manager.ts
andrewgordstewart: I would advise against putting business logic in the model, since it couples data access to business logic, making it hard to eg. Use a prepared statement rather than an objection model to make an update.

packages/server-wallet/src/objectives/objective-manager.ts
tomclose: I might be wrong, but I'm not sure we actually get that guarantee currently. It looks like we've already moved beyond a world where our updates acquire a single row, do some calculations, and then update that row. The existing `crankToCompletion` function in the wallet appears to lock the channel, not use the value returned, [refetch the channel](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/wallet/index.ts#L696-L700) within the transaction, and in cases [update the objective](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/wallet/index.ts#L746), or [update the ledger tables](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/wallet/index.ts#L746-L750). 

Something I'm aiming for in this set of refactors is to make it clearer what operations are actually happening within each transaction, so we can better reason about whether this is all ok. The plan is to have "atomic" crank operations, that each run in their own transaction, and which make it really easy to see which operations run within that transaction. We can then make a call on whether the existing locking approach is sufficient, or whether we need to go with an approach such as the one outlined in [this doc](https://www.notion.so/Server-wallet-updated-architecture-ad0372e611ac485aa16dc2c4df65a14f) (when we have the bandwidth for a proper discussion).

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+    const channelResults: ChannelResult[] = [];
+    let maybeError: any = undefined;
+
+    const objective = await this.store.getObjective(objectiveId);
+
+    let channelToLock;
+
+    if (objective.type === 'OpenChannel' || objective.type === 'CloseChannel')
+      channelToLock = objective.data.targetChannelId;
+    else throw new Error('crankToCompletion: unsupported objective');
```
geoknee, line 59:
```suggestion
    else throw new Error('ObjectiveManager.crank(): unsupported objective');
```

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+  chainService: ChainServiceInterface;
+  timingMetrics: boolean;
+}
+
+// todo: currently duplicated in wallet/index.ts
+type ExecutionResult = {
+  outbox: Outgoing[];
+  channelResults: ChannelResult[];
+  error?: any;
+};
```
geoknee, line 29: Why not export it here, and import it in `/wallet/index.ts`

geoknee (approved):

packages/server-wallet/src/objectives/objective-manager.ts
geoknee: @tomclose would be nice to do away with this

PR #2784, opened 2020-10-30
Add walletVersion as a required string on wire-format.message.payload

geoknee:
_In the xstate wallet:_ populate this string using the githash (which we are already using and is injected during the webpack build). 

_In the server wallet:_ write the package version number into the compiled javascript before publishing. Use an empty string otherwise (e.g. with ts-node). 

Also throw an error if we got a message from a wallet at a different `walletVersion`.

### DIFF
--- a/packages/server-wallet/package.json
+++ b/packages/server-wallet/package.json
@@ -84,7 +84,7 @@
     "lint:check": "eslint \"*/**/*.{js,ts}\" --cache",
     "lint:write": "eslint \"*/**/*.{js,ts}\" --fix",
     "precommit": "lint-staged --quiet",
-    "prepare": "rm -rf lib; tsc -b .",
+    "prepare": "rm -rf lib; tsc -b . && node ./scripts/inject-wallet-version.js",
     "prestart": "yarn prepare && npm run db:rollback && npm run db:migrate && npm run db:seed",
     "prettier:check": "npx prettier --check 'src/**/*.{ts,tsx}'",
     "prettier:write": "npx prettier --write 'src/**/*.{ts,tsx}'",
--- a/packages/server-wallet/scripts/inject-wallet-version.js
+++ b/packages/server-wallet/scripts/inject-wallet-version.js
@@ -1,0 +1,10 @@
+var fs = require('fs');
+
+fs.writeFileSync(
+  'lib/src/version.js',
+  `"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+exports.WALLET_VERSION = void 0;
+exports.WALLET_VERSION = '${process.env.npm_package_name}@${process.env.npm_package_version}';
+//# sourceMappingURL=version.js.map`
+);
--- a/packages/server-wallet/src/errors/wallet-error.ts
+++ b/packages/server-wallet/src/errors/wallet-error.ts
@@ -41,7 +41,7 @@ export class PushMessageError extends WalletError {
   readonly type = WalletError.errors.PushMessageError;
 
   static readonly reasons = {
-    objectiveNotFound: 'Error during pushMessage',
+    genericPushMessageError: 'Error during pushMessage',
   } as const;
 
   constructor(
--- a/packages/server-wallet/src/errors/wallet-error.ts
+++ b/packages/server-wallet/src/errors/wallet-error.ts
@@ -9,6 +9,7 @@ export abstract class WalletError extends Error {
     NonceError: 'NonceError',
     StoreError: 'StoreError',
     OnchainError: 'OnchainError',
+    PushMessageError: 'PushMessageError',
   } as const;
 
   abstract readonly type: Values<typeof WalletError.errors>;
@@ -34,4 +35,19 @@ export function isWalletError(error: any): error is WalletError {
   if (!error?.type) return false;
   if (!(typeof error.type === 'string' || error.type instanceof String)) return false;
   return hasOwnProperty(WalletError.errors, error.type);
+}
+
+export class PushMessageError extends WalletError {
+  readonly type = WalletError.errors.PushMessageError;
+
+  static readonly reasons = {
+    objectiveNotFound: 'Error during pushMessage',
+  } as const;
+
+  constructor(
+    reason: Values<typeof PushMessageError.reasons>,
+    public readonly data: {thisWalletVersion: string; payloadWalletVersion: string; cause: Error}
+  ) {
+    super(reason);
+  }
 }
--- a/packages/server-wallet/src/logger.ts
+++ b/packages/server-wallet/src/logger.ts
@@ -9,15 +9,13 @@ export function createLogger(config: ServerWalletConfig): pino.Logger {
     config.logDestination && config.logDestination.toLocaleLowerCase() !== 'console'
       ? pino.destination(config.logDestination)
       : undefined;
-  return destination
-    ? pino({level: config.logLevel}, destination).child({
-        dbName: config.postgresDBName,
-        walletVersion: WALLET_VERSION,
-      })
-    : pino({level: config.logLevel}).child({
-        dbName: config.postgresDBName,
-        walletVersion: WALLET_VERSION,
-      });
+  return (destination
+    ? pino({level: config.logLevel}, destination)
+    : pino({level: config.logLevel})
+  ).child({
+    dbName: config.postgresDBName,
+    walletVersion: WALLET_VERSION,
+  });
 }
 
 export const logger = createLogger(defaultConfig);
--- a/packages/server-wallet/src/logger.ts
+++ b/packages/server-wallet/src/logger.ts
@@ -1,6 +1,7 @@
 import pino from 'pino';
 
 import {defaultConfig, ServerWalletConfig} from './config';
+import {WALLET_VERSION} from './version';
 
 export function createLogger(config: ServerWalletConfig): pino.Logger {
   // eslint-disable-next-line no-process-env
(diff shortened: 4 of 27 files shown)


geoknee:
I `yarn link`ed locally to the graph repo, ran the e2e tests and found what I wanted in the logs 

Screenshot 2020-10-30 at 15 58 29

geoknee:
Note that the specific short term goal this PR aims to achieve is to _detect and notify_ when your wallet is talking to a different wallet (i.e. one at a different version). 

It also makes it relatively simple to extend / relax this so that "different wallet" becomes "incompatible wallet" , for example using semantic versioning to decide whether there is a problem or not.

packages/server-wallet/package.json
```diff
@@ -84,7 +84,7 @@
     "lint:check": "eslint \"*/**/*.{js,ts}\" --cache",
     "lint:write": "eslint \"*/**/*.{js,ts}\" --fix",
     "precommit": "lint-staged --quiet",
-    "prepare": "rm -rf lib; tsc -b .",
+    "prepare": "rm -rf lib; tsc -b . && node ./scripts/inject-wallet-version.js",
```
geoknee, line 87: Here's where we inject the wallet version. It'll only get into the build js files, so won't appear in our ts-node scripts and tests.

packages/server-wallet/scripts/inject-wallet-version.js
```diff
@@ -0,0 +1,10 @@
+var fs = require('fs');
+
+fs.writeFileSync(
+  'lib/src/version.js',
+  `"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+exports.WALLET_VERSION = void 0;
+exports.WALLET_VERSION = '${process.env.npm_package_name}@${process.env.npm_package_version}';
```
geoknee, line 8: Reads env vars that exist during lifecycle scripts.

packages/server-wallet/src/version.ts
```diff
@@ -0,0 +1 @@
+export const WALLET_VERSION = ''; // Will be overwritten during npm's 'prepare' lifecycle script
```
geoknee, line 1: Isolated to a single file to reduce risk of corrupting other bits of code.

packages/wire-format/src/types.ts
```diff
@@ -146,6 +146,7 @@ type GetChannel = {type: 'GetChannel'; channelId: Bytes32};
 export type ChannelRequest = GetChannel;
 
 export interface Payload {
+  walletVersion: string; // e.g. @statechannels/server-wallet@1.4.0
```
geoknee, line 149: New required property. This is a breaking change (I did not mark it as such in my commit, to avoid bumping  major version number).

packages/xstate-wallet/src/config.ts
```diff
@@ -18,6 +18,8 @@ function getBool(val: string | undefined): boolean {
 
 export const GIT_VERSION = process.env.GIT_VERSION;
 
+export const WALLET_VERSION = '@statechannels/xstate-wallet@' + GIT_VERSION;
```
geoknee, line 21: xstate-wallet uses the GIT_VERSION instead of the package version

packages/server-wallet/scripts/inject-wallet-version.js
tomclose: What's the advantage of putting the `npm_package_name` here? Won't that always be the same?

packages/server-wallet/src/wallet/index.ts
```diff
@@ -501,16 +506,22 @@ export class Wallet extends EventEmitter<WalletEvent>
 
     const wirePayload = validatePayload(rawPayload);
 
+    if (wirePayload.walletVersion !== WALLET_VERSION)
+      throw Error(
+        `Mismatch in wallet versions: this wallet is a ${WALLET_VERSION} but received a message from a ${wirePayload.walletVersion}`
+      );
```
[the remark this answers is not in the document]
tomclose: This is probably too restrictive for the moment e.g. if we want to make a performance improvement for the wallet deployed in the gateway, then this would break all the indexers. Maybe a warning - but even that could become pretty spammy, as we'd be logging a line for every single state both on the gateway and the indexer.

It seems like the time we really need to know this information is if something goes wrong. Ideally we'd somehow add this information to any error thrown or logged during the execution of a `pushMessage` - something like `{ ourWalletVersion: ..., theirWalletVersion: ...}`. We might be able to do that with a `try / catch` block around the `pushMessage`.

tomclose (approved):
A few comments/suggestions that are worth looking at, but fundamentally looks good!

packages/server-wallet/scripts/inject-wallet-version.js
geoknee: It's for the benefit of any other wallet that this one may wish to communicate with: for example our very own `xstate-wallet` or some third party nitro wallet.

geoknee: We could hard-code the package name into this file, but this has the benefit of being robust to changing the package name.

geoknee:
In response to feedback:
- no new errors thrown anymore
- catch pushMessage errors and insert context about 2x wallet versions (this + incoming message)
- wallets _own their child logger_ and **every log message** has log their DB_NAME and walletVersion

packages/server-wallet/src/wallet/index.ts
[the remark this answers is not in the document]
geoknee: d170b0a, 140a77c

packages/server-wallet/scripts/inject-wallet-version.js
snario: If I am publishing a canary build would this pick up that version e.g., `next.ab423+72.ab423`

packages/server-wallet/src/errors/wallet-error.ts
```diff
@@ -35,3 +36,18 @@ export function isWalletError(error: any): error is WalletError {
   if (!(typeof error.type === 'string' || error.type instanceof String)) return false;
   return hasOwnProperty(WalletError.errors, error.type);
 }
+
+export class PushMessageError extends WalletError {
+  readonly type = WalletError.errors.PushMessageError;
+
+  static readonly reasons = {
+    objectiveNotFound: 'Error during pushMessage',
```
snario, line 44:
```suggestion
    objectiveNotFound: 'PushMessageError: Objective not found.',
```

packages/server-wallet/scripts/inject-wallet-version.js
geoknee: I believe that it would

packages/server-wallet/src/logger.ts
```diff
@@ -1,14 +1,23 @@
...
-  return destination ? pino({level: config.logLevel}, destination) : pino({level: config.logLevel});
+  return destination
+    ? pino({level: config.logLevel}, destination).child({
+        dbName: config.postgresDBName,
+        walletVersion: WALLET_VERSION,
+      })
+    : pino({level: config.logLevel}).child({
+        dbName: config.postgresDBName,
+        walletVersion: WALLET_VERSION,
+      });
```
snario, line 20:
```suggestion
  return pino({level: config.logLevel}, destination).child({
        dbName: config.postgresDBName,
        walletVersion: WALLET_VERSION,
      });
```
Shouldn't this work?

snario:
I guess at a high level I'm confused as to whether or not there should be two concepts; one for the _wallet software version_ and another for the _wire format version_ that perhaps multiple wallets (e.g., browser and server) at multiple versions (e.g., August 2020 and November 2020) might both be compatible with

packages/server-wallet/src/logger.ts
geoknee: Yeah I think it should

geoknee: hmm maybe not

geoknee: 7049bea

packages/server-wallet/src/errors/wallet-error.ts
geoknee: b0ce125

packages/server-wallet/src/wallet/index.ts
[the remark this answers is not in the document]
andrewgordstewart: You can log the first time you see it, so it's not spammy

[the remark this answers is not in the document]
geoknee: What I settled on is adding versioning information both to i) existing logs and ii) existing errors. I hope that qualifies it as "not spammy"?

geoknee:
> I guess at a high level I'm confused as to whether or not there should be two concepts; one for the _wallet software version_ and another for the _wire format version_ that perhaps multiple wallets (e.g., browser and server) at multiple versions (e.g., August 2020 and November 2020) might both be compatible with

Perhaps: at least this PR gets us some of the way there. The wire-format version (if we chose to use that at some point) would be inserted and checked at a slightly different level, I think. As it is we are using exact versioning, so it's easy to infer a version of the wire format that each of our wallet versions is compatible with.

packages/server-wallet/src/version.ts
andrewgordstewart: Did you consider using `read-pkg` to read the version (asynchronously)?

Since we now create wallets through a [static `create` function](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/wallet/index.ts#L16), we can make it asynchronous and inject the wallet version there.

PR #2785, opened 2020-10-30
fix: remove forced turn taking requirment on post-fund setup state

snario:


packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -127,26 +127,32 @@ it('Create a directly funded channel between two wallets ', async () => {
...
 
   const postFundA = await postFundAPromise;
-  await channelFundedBPromise;
-  expect(getChannelResultFor(channelId, [postFundA.channelResult])).toMatchObject({
+  const postFundB = await postFundBPromise;
+
+  expect(postFundA.channelResult).toMatchObject({
     status: 'running',
     turnNum: 2,
   });
```
kerzhner, line 135: This removes the channelId check that exists with `getChannelResultFor(channelId, [postFundA.channelResult])`
```suggestion
  expect(postFundA.channelResult).toMatchObject({
    channelId,
    status: 'running',
    turnNum: 2,
  });
```

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -127,26 +127,32 @@ it('Create a directly funded channel between two wallets ', async () => {
...
 
+  await b.pushMessage(getPayloadFor(participantB.participantId, postFundA.outbox));
+
   const resultA1 = await a.pushMessage(getPayloadFor(participantA.participantId, postFundB.outbox));
   expect(getChannelResultFor(channelId, resultA1.channelResults)).toMatchObject({
     status: 'running',
     turnNum: 3,
   });
 
+  const resultB1 = await a.pushMessage(getPayloadFor(participantB.participantId, postFundA.outbox));
```
kerzhner, line 150: This looks wrong, `a` should not push `postFundA.outbox`

packages/server-wallet/src/__test__/crash-tolerance/crash-tolerance.test.ts
```diff
@@ -125,12 +125,14 @@ it('Create a fake-funded channel between two wallets, of which one crashes midwa
   const resultB2 = await b.updateFundingForChannels([depositByB]);
 
   expect(getChannelResultFor(channelId, resultA2.channelResults)).toMatchObject({
-    status: 'running', // Still opening because turnNum 3 is not supported yet, but is signed by A
+    // Running because turnNum 2 is supported and channel result says 'running' for state > 1
```
kerzhner, line 128: This doesn't make sense to me. Mind explaining?

packages/server-wallet/src/protocols/open-channel.ts
```diff
@@ -191,10 +174,9 @@ const signPreFundSetup = (ps: ProtocolState): ProtocolResult | false =>
   });
 
 const signPostFundSetup = (ps: ProtocolState): ProtocolResult | false =>
-  myTurnToPostfund(ps) &&
-  (isFunded(ps) || isFakeFunded(ps.app)) &&
   ps.app.latestSignedByMe &&
-  ps.app.supported &&
+  ps.app.latestSignedByMe.turnNum < ps.app.participants.length &&
```
kerzhner, line 178: Should we check that there exists a supported prefund state before signing a postfund state?

kerzhner (changes requested):

packages/server-wallet/src/__test__/crash-tolerance/crash-tolerance.test.ts
snario: Turn number 2 is a valid chain from 0, 1, 2 which A has now and the wallet shows `status = "running"` for turn number 2

andrewgordstewart: I think that's a bug though?

snario: Yeah, but can be fixed in a separate PR.

snario: https://github.com/statechannels/statechannels/pull/2794

andrewgordstewart: Presumably this will fail when both this pr and #2794 are merged?

snario: Rebased.

snario:
> Are there going to be issues with Alice receiving Bob's postfund state turn 3 before Alice has signed post fund turn 2?

I think the current tests confirm that this should not be an issue. What needs to be ensured is that the protocol will sign a state regardless of whether or not it has the others' signed post-fund state. I added a test for that: 38b4315b7.

PR #2788, opened 2020-10-30
Fix: Do not evm validate final states

lalexgap:
Following the force move contracts we should not be calling `validTransition` when transitioning to a `isFinal` state. 

This fixes the validation logic to respect that.

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -330,8 +330,8 @@ export class Store {
     }
 
     // Validates app specific rules by running the app rules contract in the EVM
-    // We only want to run the validation for states not in the funding stage per the force move contract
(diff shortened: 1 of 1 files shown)


andrewgordstewart (approved):
Ok with this as is.

If it's not a burden, can you add a regression unit test for this? An effective one might require a refactor to, for example, extract code that validates state transitions into a method not attached to the store.

A reasonable alternative is to create an issue for this request.

lalexgap:
> Ok with this as is.
> 
> If it's not a burden, can you add a regression unit test for this? An effective one might require a refactor to, for example, extract code that validates state transitions into a method not attached to the store.
> 
> A reasonable alternative is to create an issue for this request.

I'd like to get this in as soon as possible so I've created a separate issue here: https://github.com/statechannels/statechannels/issues/2789#issue-733455024

PR #2792, opened 2020-10-31
Close channel objective only completes when funds have been transferred

kerzhner:
This work only impacts directly funded channels.

How it works before this PR: when processing the close channel objective, the wallet submits a `concludeAndWithdraw` transaction. If the transaction is successfully submitted or fails for an accepted reason (regardless of what happens during mining), the wallet allows the close channel objective to succeed.

How it works after this PR: the wallet only allows the close channel objective to succeed if the wallet has been notified of `AssetTransferred` events that show that all of the assets for the participant have been transferred out of the channel.

Follow up work includes the wallet emitting an event when the close channel objective succeeds. At the moment, the tests do not verify that the close channel objective succeeds for directly funded channels.

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -217,10 +217,10 @@ export class Channel extends Model implements RequiredColumns {
       fundingLedgerChannelId,
     } = this;
     const funding = (assetHolder: Address): ChannelStateFunding => {
+      const noFunding = {amount: Zero, transferredOut: []};
+      if (!this.funding) return noFunding;
       const result = this.funding.find(f => f.assetHolder === assetHolder);
-      return result
-        ? {amount: result.amount, transferredOut: result.transferredOut}
-        : {amount: Zero, transferredOut: []};
+      return result ? {amount: result.amount, transferredOut: result.transferredOut} : noFunding;
     };
     return {
       myIndex: myIndex as 0 | 1,
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -14,7 +14,12 @@ import {JSONSchema, Model, Pojo, QueryContext, ModelOptions, TransactionOrKnex} 
 import {ChannelResult, FundingStrategy} from '@statechannels/client-api-schema';
 
 import {Address, Bytes32, Uint48} from '../type-aliases';
-import {ChannelState, toChannelResult, ChainServiceRequests} from '../protocols/state';
+import {
+  ChannelState,
+  toChannelResult,
+  ChainServiceRequests,
+  ChannelStateFunding,
+} from '../protocols/state';
 import {WalletError, Values} from '../errors/wallet-error';
 
 import {SigningWallet} from './signing-wallet';
@@ -211,9 +216,11 @@ export class Channel extends Model implements RequiredColumns {
       fundingStrategy,
       fundingLedgerChannelId,
     } = this;
-    const funding = (assetHolder: Address): string => {
+    const funding = (assetHolder: Address): ChannelStateFunding => {
       const result = this.funding.find(f => f.assetHolder === assetHolder);
-      return result ? result.amount : Zero;
+      return result
+        ? {amount: result.amount, transferredOut: result.transferredOut}
+        : {amount: Zero, transferredOut: []};
     };
     return {
       myIndex: myIndex as 0 | 1,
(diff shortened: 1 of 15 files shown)


packages/server-wallet/src/models/funding.ts
```diff
@@ -47,13 +61,40 @@ export class Funding extends Model implements RequiredColumns {
...
+  ): Promise<Funding> {
+    return knex.transaction(async tx => {
+      const existing = await Funding.query(tx)
+        .where({channelId, assetHolder})
+        .first();
+
+      if (!existing) {
+        const errorMessage = `Expected for funding row to exists with channelId ${channelId}, assetHolder ${assetHolder}`;
+        logger.error(errorMessage, {channelId, assetHolder, toAddress, amount});
+        throw new Error(errorMessage);
```
geoknee, line 89: Alternative way to implement this: https://vincit.github.io/objection.js/api/query-builder/other-methods.html#throwifnotfound

geoknee (approved):

kerzhner:
Update: follow up work in the PR description has been completed in f653bce..27a82dc

packages/server-wallet/src/models/channel.ts
```diff
@@ -210,9 +215,11 @@ export class Channel extends Model implements RequiredColumns {
       fundingStrategy,
       fundingLedgerChannelId,
     } = this;
-    const funding = (assetHolder: Address): string => {
+    const funding = (assetHolder: Address): ChannelStateFunding => {
       const result = this.funding.find(f => f.assetHolder === assetHolder);
-      return result ? result.amount : Zero;
+      return result
+        ? {amount: result.amount, transferredOut: result.transferredOut}
+        : {amount: Zero, transferredOut: []};
```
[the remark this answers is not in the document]
kerzhner: Apologies, not sure I fully understand. Would you mind elaborating?

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -33,7 +33,9 @@ function everyoneSignedFinalState(ps: ProtocolState): boolean {
 }
 
 function successfulWithdraw({app}: ProtocolState): boolean {
+  if (app.fundingStrategy !== 'Direct') return true;
```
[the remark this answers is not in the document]
kerzhner: Agreed. I think I need to understand better how funding is stored for ledger funded channel to see how we can abstract away this logic for a channel regardless of funding.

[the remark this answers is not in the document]
kerzhner: 8cfad599e

packages/server-wallet/src/models/funding.ts
kerzhner: Excellent point: 07b806a33

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -46,7 +46,8 @@ function successfulWithdraw({app}: ProtocolState): boolean {
     .reduce((soFar, currentAi) => BN.add(soFar, currentAi.amount), BN.from(0));
   const amountTransferredToMe = app
     .funding(assetHolderAddress)
-    .transferredOut.filter(tf => tf.toAddress === myDestination)
+    // todo: figure out why destination in emitted events are lower case
+    .transferredOut.filter(tf => tf.toAddress === myDestination.toLocaleLowerCase())
```
geoknee, line 52: nit: I think you probably want `.toLowerCase()` (not sure the `locale` is doing anything for us, here.

packages/server-wallet/src/wallet/index.ts
```diff
@@ -719,14 +746,21 @@ export class Wallet extends EventEmitter<WalletEvent>
 
   // ChainEventSubscriberInterface implementation
   holdingUpdated(arg: HoldingUpdatedArg): void {
-    const channelUpdate: WalletEventName = 'channelUpdate';
+    const channelUpdate: ChannelUpdateEventName = 'channelUpdate';
```
geoknee, line 749: nit: I think we have used "ChannelUpdat**ed**" in the past and this is what you'd find in the `client-api-schema`.

geoknee (approved):
👍

packages/server-wallet/src/models/channel.ts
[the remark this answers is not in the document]
geoknee: ah I may have made the comment on an intermediate commit / something that got force-pushed over. It was inconsequential anyway.

packages/server-wallet/src/protocols/close-channel.ts
kerzhner: Good catch, I think `toLocaleLowerCase` was the first autocomplete option :) d8510e23e

packages/server-wallet/src/wallet/index.ts
kerzhner: 👍 a4b8dd1c5

PR #2793, opened 2020-10-31
[server-wallet] Simplify ledger funding algorithm

snario:
Based on a suggestion by @geoknee to always sign the intersection

### DIFF
--- a/packages/server-wallet/src/protocols/process-ledger-queue.ts
+++ b/packages/server-wallet/src/protocols/process-ledger-queue.ts
@@ -184,7 +184,7 @@ const computeNewOutcome = ({
   let insufficientFundsFor: Bytes32[] = [];
 
   // Check if I already signed a new ledger update, if so continue with that
-  if (latestSignedByMe.turnNum > supported.turnNum) myExpectedOutcome = myLatestOutcome;
+  if (latestSignedByMe.turnNum === supported.turnNum + n) myExpectedOutcome = myLatestOutcome;
   // Otherwise, iteratively allocate funds from the supported state's (i.e., S) outcome
   else
     ({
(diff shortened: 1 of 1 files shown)


packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -134,16 +134,13 @@ const xorOutcome = (a: SimpleAllocation, b: SimpleAllocation): SimpleAllocation
  *
  * IF there are any pending ledger updates (to fund OR defund a channel)
  *
- *   i. Check if I already signed a new ledger update O₁. If I did, then use that. [1]
+ *   i. Check if I already signed a new ledger update. If I did, then call this O₁. [1]
```
andrewgordstewart, line 137:
```suggestion
 *   i. Check if I already signed a new ledger update. If I did, then call its outcome O₁. [1]
```

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -134,16 +134,13 @@ const xorOutcome = (a: SimpleAllocation, b: SimpleAllocation): SimpleAllocation
...
  *      Otherwise, iteratively allocate funds from the supported state's (i.e., S) outcome
  *      to each request, call this O₁.
  *
- *  ii. Check if there is a conflicting ledger update, O₂.
- *
- *      a. If O₁ ⋂ O₂ ≠ O₁ sign an outcome for O₁ ⋂ O₂ with turn number 2
- *         higher than the one corresponding to O₂.
- *
- *      b. If O₁ ⋂ O₂ = O₁ = O then sign O with turn number 2 corresponding to O₂.
+ *  ii. If there is a conflicting ledger update, call that O₂. Sign O₁ ⋂ O₂ with the higher turn
```
andrewgordstewart, line 141:
```suggestion
 *  ii. If there is a conflicting ledger update, call its outcome O₂. Sign a state with outcome O₁ ⋂ O₂ and the higher turn number
```

I really like the mathematical notation in the code, but we **also need to define** what O₁ ⋂ O means.

Also, what is "the higher turn number"?

andrewgordstewart (approved):
I need to think a bit more about the algorithm described, and whether this is a correct implementation of it. But since:
1. I didn't understand it before
2. I still don't understand it
3. The tests pass
4. The code is now simpler

I don't see the point in blocking based on (2), considering (1), (3) and (4)!

PR #2797, opened 2020-10-31
Purge some TODOs

snario:


packages/server-wallet/src/wallet/store.ts
```diff
@@ -650,11 +650,6 @@ export class Store {
         tx
       );
 
-      /** todo:
-       * What happens if Bob is adding his signature to prefund0 from Alice?
```
andrewgordstewart, line 654: Why is this stale? It seems like it might have manifested in a confusing situation that lead to a DB constraint violation in staging.

Screen Shot 2020-11-01 at 12 36 45 PM

packages/server-wallet/src/wallet/store.ts
```diff
@@ -491,15 +490,15 @@ export class Store {
...
+        status: 'approved',
         type: objective.type,
         participants: [],
         data: {
           targetChannelId,
           fundingStrategy,
         },
       };
-      // TODO: (Stored Objectives) Does it make sense to do the INSERT here?
+
```
andrewgordstewart, line 501: Why does it make sense to do the INSERT here?

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -95,7 +95,6 @@ it('Create a directly funded channel between two wallets ', async () => {
   // PreFund0
   const preFundA = await a.createChannel(channelParams);
 
-  // TODO compute the channelId for a better test
```
andrewgordstewart, line 98: Is the test now a better test?

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -160,26 +156,3 @@ const createOutboxFor = (
       method: 'MessageQueued' as const,
       params: serializeMessage(data, recipient, participants[myIndex].participantId, channelId),
     }));
-
-const determineWhichLedgerToUse = async (
```
andrewgordstewart, line 164: This looks like a great thing to remove.

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -103,19 +101,17 @@ export class ObjectiveManager {
               await this.chainService.concludeAndWithdraw([protocolState.app.supported!]);
               return;
             case 'RequestLedgerFunding': {
-              const ledgerChannelId = await determineWhichLedgerToUse(protocolState.app, tx);
               await LedgerRequest.requestLedgerFunding(
                 protocolState.app.channelId,
-                ledgerChannelId,
+                protocolState.app.fundingLedgerChannelId as string,
```
andrewgordstewart, line 106: Why does this need type assertion? The type system should be narrowing the type of `fundingLedgerChannelId` for us.

packages/server-wallet/src/protocols/process-ledger-queue.ts
```diff
@@ -315,19 +311,25 @@ export const getProcessLedgerQueueProtocolState = async (
   const ledgerRequests = await store.getPendingLedgerRequests(ledgerChannelId, tx);
   return {
     fundingChannel: await store.getChannel(ledgerChannelId, tx),
-    channelsRequestingFunds: await Promise.all(
-      compose(
-        map(({channelToBeFunded}: LedgerRequestType) => store.getChannel(channelToBeFunded, tx)),
-        filter(['status', 'pending']),
-        filter(['type', 'fund'])
-      )(ledgerRequests)
+    channelsRequestingFunds: _.sortBy(
```
andrewgordstewart, line 314: Suggestion: Define a helper function `sortByNonce = channelStates => _.sortBy(channelStates, a => a.latest.channelNonce)`, and make this
`await Promise.all(compose(//snip)).then(sortByNonce)`

packages/server-wallet/src/wallet/store.ts
```diff
@@ -268,10 +268,13 @@ export class Store {
     const toMoverIndex = toState.turnNum % toState.participants.length;
     const toMover = toState.participants[toMoverIndex].signingAddress;
 
-    const turnNumCheck = _.isEqual(toState.turnNum, fromState.turnNum + 1);
-    if (!turnNumCheck) {
-      const VALIDATION_ERROR = `Turn number check failed.`;
-      logger.error({fromState, toState, error: Error(VALIDATION_ERROR)}, VALIDATION_ERROR);
+    if (fromState.appDefinition !== constants.AddressZero) {
```
andrewgordstewart, line 271: This predicate should not be inlined. There should be an `isLedgerChannel(signedState)` function that is called here.

packages/server-wallet/src/db-admin/knexfile.ts
```diff
@@ -3,7 +3,7 @@ import * as path from 'path';
 import {Config} from 'knex';
 
 // Populate env vars as knexfile is used directly in yarn scripts
-// FIXME We should not need to depend on devtools at this step.
+// TODO: We should not need to depend on devtools at this step.
```
andrewgordstewart, line 6: New issue.

andrewgordstewart (changes requested):
Some of these are definitely positive, but I am uncertain about all of these changes.

There should be some context around why some of these TODOs are removed -- otherwise, I have no way of knowing whether this should be approved or not.

packages/server-wallet/src/wallet/store.ts
snario: This comment was originally added before we used PostgreSQL at all to help @geoknee write the PostgreSQL implementation later and so I wrote it because I wasn't sure if it would be a good place to make a call to an object that needed to be within the context of a database transaction. After we added the database integration, it seemed like the answer was that we _could_ do the insert at this place in the code.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
snario: I disagree with this necessarily being something we should do for a better test — we have other tests to ensure that channel IDs are being computed correctly inside the wallet, and an integration test does not seem the right place to check that implementation detail.

packages/server-wallet/src/objectives/objective-manager.ts
snario: It could be `undefined` because not _all_ `ChannelState` objects have this value and we do not have any type narrowing to detect it (yet).

packages/server-wallet/src/wallet/store.ts
snario: I actually specifically did not do this because I think this is a valid condition regardless of the type of channel — if there is no app definition then I don't think it matters to do turn number verification.

packages/server-wallet/src/protocols/process-ledger-queue.ts
snario: 11508ec0f

packages/server-wallet/src/wallet/store.ts
snario: This method is not called by Bob — it is only ever called inside `wallet.createChannel` or `wallet.createLedgerChannel`

packages/server-wallet/src/db-admin/knexfile.ts
snario: https://github.com/statechannels/statechannels/issues/2801

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: Oh, fair. I guess it should be a `isNullApp` function but that seems ok.

andrewgordstewart (approved):

PR #2800, opened 2020-11-01
[server-wallet] Clean store.ts

snario:
Ran through some (not all) functions in the `Store` implementation and tried to simplify them using generally reasonable coding standards.

### DIFF
--- a/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
@@ -67,7 +67,7 @@ it('stores states contained in the message, in a single channel model', async ()
   expect(channelsAfter[0].vars).toHaveLength(2);
 
   // The Channel model adds the state hash before persisting
-  expect(signedStates.map(addHash)).toMatchObject(channelsAfter[0].vars);
+  expect(channelsAfter[0].vars).toMatchObject(signedStates);
 });
 
 it('ignores duplicate states', async () => {
@@ -92,7 +92,7 @@ it('ignores duplicate states', async () => {
   expect(channelsAfter[0].vars).toHaveLength(2);
 
   // The Channel model adds the state hash before persisting
-  expect(signedStates.map(addHash)).toMatchObject(channelsAfter[0].vars);
+  expect(channelsAfter[0].vars).toMatchObject(signedStates);
 });
 
 it('adds signatures to existing states', async () => {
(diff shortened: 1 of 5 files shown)


packages/server-wallet/src/wallet/store.ts
```diff
@@ -455,101 +444,79 @@ export class Store {
...
+   * @param channelId - the channel ID the state refers to
+   * @param state - the new signed state to be added
+   * @param tx - an existing database transaction context
+   *
+   * This function serves as both a data validator (ensure the state is valid) and
+   * an efficient operator on the database.
+   */
   async addSignedState(
     channelId: string,
-    maybeChannel: Channel | undefined,
```
geoknee, line 460: By ripping out this parameter, you are making this function go to the DB _for sure_. Previously, I could skip that if  I already had the channel?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -455,101 +444,79 @@ export class Store {
...
-    wireSignedState: WireSignedState,
+    wireState: WireSignedState,
```
geoknee, line 459: why?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -632,103 +599,75 @@ class StoreError extends WalletError {
...
-    {
-      ...constants,
-      channelId,
-      vars: [],
-      chainServiceRequests: [],
-      signingAddress,
-      fundingStrategy,
-    },
-    ...CHANNEL_COLUMNS
+  await Nonce.ensureLatest(constants.channelNonce, addresses, txOrKnex);
```
geoknee, line 620: This is _new_ behaviour, if I am not mistaken, and not just a refactor?

packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
```diff
@@ -67,7 +67,7 @@ it('stores states contained in the message, in a single channel model', async ()
   expect(channelsAfter[0].vars).toHaveLength(2);
 
   // The Channel model adds the state hash before persisting
-  expect(signedStates.map(addHash)).toMatchObject(channelsAfter[0].vars);
+  expect(channelsAfter[0].vars).toMatchObject(signedStates);
```
geoknee, line 70: Is the addHash not necessary?

geoknee (approved):

packages/server-wallet/src/wallet/store.ts
snario: It was never used in any instance of the codebase, so I removed it for that reason

packages/server-wallet/src/wallet/store.ts
snario: Just shortening the code — not attached to this

packages/server-wallet/src/wallet/store.ts
snario: Same behaviour as before! It used to be a bit more hidden

packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
snario: No, it was being used to mask an internal implementation detail

PR #2805, opened 2020-11-01
fix: remove subtely introduced additional JS ecrecover call (via deserialization fn)

snario:
Calling `deserializeState` triggered an additional call to `getStateSignerAddress` which does `ethers.utils.keccak256` and `ethers.utils.verifyMessage` in JS

1. https://github.com/statechannels/statechannels/blob/6aa64ccdb346722a40e1412df907d08f20ef7f60/packages/server-wallet/src/wallet/store.ts#L576-L576

2. https://github.com/statechannels/statechannels/blob/6aa64ccdb346722a40e1412df907d08f20ef7f60/packages/wallet-core/src/serde/wire-format/deserialize.ts#L84-L84

3. https://github.com/statechannels/statechannels/blob/6aa64ccdb346722a40e1412df907d08f20ef7f60/packages/wallet-core/src/state-utils.ts#L95-L98

4. https://github.com/statechannels/statechannels/blob/6aa64ccdb346722a40e1412df907d08f20ef7f60/packages/nitro-protocol/src/signatures.ts#L12-L14

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -23,7 +23,6 @@ import {
   convertToParticipant,
   SignedState,
   objectiveId,
-  deserializeState,
   isSimpleAllocation,
   checkThat,
 } from '@statechannels/wallet-core';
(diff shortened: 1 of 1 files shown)


andrewgordstewart (approved):
How did you catch this?

snario:
Was trying to refactor the code by removing unnecessary calls to things and then noticed this while inspecting the call to `deserializeState` which seemed unnecessary.

PR #2807, opened 2020-11-02
refactor: de-couple outgoing messages from Store

snario:
The `Store` was preparing outbound messages including objectives and signed states when it did not need to. This refactors that code into the wallet API handlers and has the added benefit of increasing re-usability by other API handler methods (i.e., `createChannel`, `createChannels`, and `createLedgerChannel` all re-use the same logic which re-uses `this.store.createChannel`).

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -8,6 +8,7 @@ import {
   GetStateParams,
   Address,
   ChannelId,
+  FundingStrategy,
 } from '@statechannels/client-api-schema';
 import {
   deserializeAllocations,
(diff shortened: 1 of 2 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -248,85 +248,116 @@ export class Wallet extends EventEmitter<WalletEvent>
...
+    fundingStrategy: FundingStrategy,
+    fundingLedgerChannelId?: Bytes32,
+    role: 'app' | 'ledger' = 'app'
+  ): Promise<SingleChannelOutput> {
+    const channelNonce = await this.store.nextNonce(participants.map(p => p.signingAddress));
+
+    const constants = {
+      appDefinition,
+      chainId: this.walletConfig.chainNetworkID,
+      challengeDuration: 9001,
```
andrewgordstewart, line 330: How long is this going to go on? 😆

andrewgordstewart (approved):

PR #2809, opened 2020-11-02
fix: fix broken query on ledger channels in DB

snario:
I observed this error:

```
[1604329033593] ERROR (7338): [] PaymentServer - DataError: select "channels".* from "channels" where "asset_holder_address" = $1 and "participants" = $2 - invalid input syntax for type json
    at wrapError (/node_modules/db-errors/lib/dbErrors.js:19:14)
    at handleExecuteError (/node_modules/objection/lib/queryBuilder/QueryBuilder.js:1506:32)
    at QueryBuilder.execute (/node_modules/objection/lib/queryBuilder/QueryBuilder.js:687:20) {
  nativeError: error: invalid input syntax for type json
      at Parser.parseErrorMessage (/node_modules/pg-protocol/src/parser.ts:357:11)
      at Parser.handlePacket (/node_modules/pg-protocol/src/parser.ts:186:21)
      at Parser.parse (/node_modules/pg-protocol/src/parser.ts:101:30)
      at Socket.<anonymous> (/node_modules/pg-protocol/src/index.ts:7:48)
      at Socket.emit (events.js:310:20)
      at Socket.EventEmitter.emit (domain.js:482:12)
      at addChunk (_stream_readable.js:286:12)
      at readableAddChunk (_stream_readable.js:268:9)
      at Socket.Readable.push (_stream_readable.js:209:10)
      at TCP.onStreamRead (internal/stream_base_commons.js:186:23) {
```

And found [this solution](https://github.com/brianc/node-postgres/issues/442#issuecomment-24496481).

This applies that solution and adds a test to find this in the future.

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -152,7 +152,7 @@ export class Channel extends Model implements RequiredColumns {
   ): Promise<Channel[]> {
     return Channel.query(txOrKnex)
       .select()
-      .where({assetHolderAddress, participants});
+      .where({assetHolderAddress, participants: JSON.stringify(participants)});
   }
 
   static allChannelsWithPendingLedgerRequests(txOrKnex: TransactionOrKnex): Promise<Channel[]> {
--- a/packages/server-wallet/src/wallet/__test__/integration/create-ledger-channel.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/create-ledger-channel.test.ts
@@ -1,0 +1,80 @@
+import {constants} from 'ethers';
+
+import {Channel} from '../../../models/channel';
+import {Wallet} from '../..';
+import {createChannelArgs} from '../fixtures/create-channel';
+import {seedAlicesSigningWallet} from '../../../db/seeds/1_signing_wallet_seeds';
+import {defaultTestConfig} from '../../../config';
+import {alice, bob} from '../fixtures/participants';
+import {DBAdmin} from '../../../db-admin/db-admin';
+
+let w: Wallet;
+beforeEach(async () => {
+  w = new Wallet(defaultTestConfig);
+  await new DBAdmin(w.knex).truncateDB();
+});
+
+afterEach(async () => {
+  await w.destroy();
+});
+
+describe('happy path', () => {
+  beforeEach(async () => seedAlicesSigningWallet(w.knex));
+
+  it('creates a ledger channel', async () => {
+    expect(await Channel.query(w.knex).resultSize()).toEqual(0);
+
+    const {participants, allocations} = createChannelArgs();
+
+    const createPromise = w.createLedgerChannel({participants, allocations});
+
+    await expect(createPromise).resolves.toMatchObject({
+      outbox: [
+        {
+          params: {
+            recipient: 'bob',
+            sender: 'alice',
+            data: {
+              signedStates: [{turnNum: 0}],
+              objectives: [
+                {
+                  participants: [alice(), bob()],
+                  data: {
+                    fundingStrategy: 'Direct',
+                  },
+                  type: 'OpenChannel',
+                },
+              ],
+            },
+          },
+        },
+      ],
+      channelResult: {channelId: expect.any(String), turnNum: 0},
+    });
+
+    const {channelId} = (await createPromise).channelResult;
+
+    expect(await Channel.query(w.knex).resultSize()).toEqual(1);
+
+    const updated = await Channel.forId(channelId, w.knex);
+    const expectedState = {
+      turnNum: 0,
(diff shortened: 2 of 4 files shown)


packages/server-wallet/src/wallet/__test__/integration/create-ledger-channel.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      outbox: [
+        {
+          params: {
+            recipient: 'bob',
+            sender: 'alice',
+            data: {
+              signedStates: [{turnNum: 0}],
+              objectives: [
+                {
+                  participants: [alice(), bob()],
```
geoknee, line 41: Are these wallet core participants or api participants?

packages/server-wallet/src/models/channel.ts
```diff
@@ -152,7 +152,7 @@ export class Channel extends Model implements RequiredColumns {
   ): Promise<Channel[]> {
     return Channel.query(txOrKnex)
       .select()
-      .where({assetHolderAddress, participants});
+      .where({assetHolderAddress, participants: JSON.stringify(participants)});
```
geoknee, line 155: How does this approach compare to [this approach](https://github.com/statechannels/statechannels/pull/2792/commits/4ec1b65185bc8748fffa29e3de1f0317878f74a2)?

geoknee (approved):

packages/server-wallet/src/wallet/__test__/integration/create-ledger-channel.test.ts
snario: `wallet-core` though in this context it doesn't matter; the only different is that `{ _isDestination: void }` is added on the typescript type for `wallet-core`

packages/server-wallet/src/models/channel.ts
snario: I don't think that approach works and that issue is hidden since only `[]` is passed in on that PR

packages/server-wallet/src/wallet/__test__/integration/create-ledger-channel.test.ts
geoknee: I thought there was an actual runtime difference: I think we get zero-padding to bytes32 in wallet core but not in the api type? There's a chance that could fall foul of our runtime type validation?

snario: Hm, well, inside the wallet we only every use the `wallet-core` variant and it's what is placed in the DB. The only place the API variant is used is in the wallet API handlers which immediately convert them into the wallet-core variant, so I don't think this an issue right now and at the least it's not an issue in this PR

geoknee: My point is that we are fixing types on the request side of the API (e.g. #2808 ) and we should also fix them on the response side. AFAICT we are spitting out wallet-core types in the return value of an API call (assuming this test passes and is sensitive to that).

packages/server-wallet/src/models/channel.ts
geoknee: @kerzhner

kerzhner: I am not 100% sure what "this approach doesn't work" means in this case :/ My observations are:
1. I don't know what the implications of `JSON.stringify` are. Is some/all validation skipped for the participants column?
1. For the `transferredOut` column, it is **not** the case that `[]` is the only value used: https://github.com/statechannels/statechannels/blob/asset-transfer/packages/server-wallet/src/models/funding.ts#L93

snario: I mean that when I add
```
participants: {
  type: 'array',
  items: {type: 'object'},
}
```
in the same way as `transferredOut` above and run the test I wrote, I get the above `DataError`

andrewgordstewart: It may be the case that "you're doing it wrong", and objection + knex makes it difficult to know what you are doing wrong, and what is the correct thing to do.

snario: If so, I was unable to figure out how to do it "right"

andrewgordstewart: So leaving be the issue of whether or not this change is required, and coming back to
> I don't know what the implications of JSON.stringify are. Is some/all validation skipped for the participants column?

Do you know what the implications are? If there are no implications, then we might not need to understand the "right" way to do it.

snario: I thought about this for a few minutes and created [this](https://github.com/statechannels/statechannels/pull/2812) to continue the discussion

PR #2812, opened 2020-11-02
fix: use knex whereComposite instead of where for composite index

snario:
For future consideration. I'm not totally sure if this is right or not. A good analysis might review the raw SQL that gets generated between the two approaches.

The [docs](https://vincit.github.io/objection.js/api/query-builder/find-methods.html#wherecomposite) seem to indicate this is at least in theory meant for composite indexes which this is one.

Also, I'm not sure we can get around stringifying the participants. It could be that a composite index in PostgreSQL between a `jsonb` and a `string` must be a `string`...?

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -152,7 +152,10 @@ export class Channel extends Model implements RequiredColumns {
   ): Promise<Channel[]> {
     return Channel.query(txOrKnex)
       .select()
-      .where({assetHolderAddress, participants: JSON.stringify(participants)});
+      .whereComposite(
+        ['assetHolderAddress', 'participants'],
+        [assetHolderAddress, JSON.stringify(participants)]
+      );
   }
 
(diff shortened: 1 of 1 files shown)


andrewgordstewart:
@snario what is the status on this PR?

lalexgap:
I think we should close this until we have time to compare performance of `whereComposite` vs `where`. @snario sound good to you?

snario:
I think it's pretty clear that this is the correct way to query composite indexes using `knex`

geoknee:
Can we get a hint of any performance changes by looking at the jobs that have run on circle (stress test?)

andrewgordstewart:
> Can we get a hint of any performance changes by looking at the jobs that have run on circle (stress test?)

No, because there's essentially no data in The DB

> I think it's pretty clear that this is the correct way to query composite indexes using `knex`

I remember understanding from the documentation what whereComposite does, so I disagree.

lalexgap:
This seems pretty stale and I don't think `where` vs `whereComposite` is terribly important so closing for now.

snario:
I'm still pretty confident that it's the correct way to do a `WHERE` on a composite index

PR #2813, opened 2020-11-02
Ignore older states when validating

lalexgap:
Ignore old states that may be received from syncing when validating. Also moves the shouldValidate logic into it's own function for clarity.

I'll be adding tests but this can be done in a separate PR to prevent blocking @snario .

### DIFF
--- a/packages/server-wallet/src/utilities/validate-transition.ts
+++ b/packages/server-wallet/src/utilities/validate-transition.ts
@@ -1,7 +1,6 @@
 import _ from 'lodash';
 import {SignedState, StateWithHash, toNitroState} from '@statechannels/wallet-core';
 import {constants} from 'ethers';
-import {Transaction} from 'knex';
 
 import {logger} from '../logger';
 import {validateTransitionWithEVM} from '../evm-validator';
@@ -10,20 +9,15 @@ import {Channel} from '../models/channel';
 
 export async function shouldValidateTransition(
   incomingState: StateWithHash,
-  channel: Channel,
-  tx: Transaction
+  channel: Channel
 ): Promise<boolean> {
-  const {channelId} = channel;
+  const {supported, isLedger} = channel;
 
   // If we already have the state we should of already validated it
   const alreadyHaveState = _.some(channel.sortedStates, ['stateHash', incomingState.stateHash]);
 
-  const {supported} = channel;
-
   // Ignore older states that may be added via syncing
   const isOldState = incomingState.turnNum < (supported?.turnNum || 0);
-
-  const isLedger = await Channel.isLedger(channelId, tx);
 
   return !!supported && !isOldState && !alreadyHaveState && !isLedger;
 }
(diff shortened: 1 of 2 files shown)


packages/server-wallet/src/utilities/validate-transition.ts
```diff
@@ -1,11 +1,32 @@
...
+
+  // If we already have the state we should of already validated it
+  const alreadyHaveState = _.some(channel.sortedStates, ['stateHash', incomingState.stateHash]);
+
+  const {supported} = channel;
+
+  // Ignore older states that may be added via syncing
+  const isOldState = incomingState.turnNum < (supported?.turnNum || 0);
+
+  const isLedger = await Channel.isLedger(channelId, tx);
```
snario, line 26: I'd rather not couple `Channel` in a utility file if we can avoid it. We could parameterize this perhaps? You can check if it's a ledger channel by seeing if `channel.assetHolderAddress` is `undefined` or not

andrewgordstewart: Wait, why should we need to make a DB query to answer this question? (Should `Channel.isLedger` even exist as a method?)

lalexgap: Good point, fixed in 09704fa0094a4e7b486a82a956c3933952dc523c

lalexgap: > Wait, why should we need to make a DB query to answer this question? (Should `Channel.isLedger` even exist as a method?)

As far as I can tell it's only used by `Store.isLedger` which I don't think is used anywhere? So it probably should be removed.

packages/server-wallet/src/utilities/validate-transition.ts
```diff
@@ -1,11 +1,22 @@
...
 import {constants} from 'ethers';
 
 import {logger} from '../logger';
 import {validateTransitionWithEVM} from '../evm-validator';
 import {Bytes} from '../type-aliases';
+import {Channel} from '../models/channel';
 
+export function shouldValidateTransition(incomingState: StateWithHash, channel: Channel): boolean {
+  const {supported, isLedger} = channel;
+  // If we already have the state we should of already validated it
```
tomclose, line 12: "should of" -> "should have" 🤓

packages/server-wallet/src/utilities/validate-transition.ts
```diff
@@ -1,11 +1,22 @@
...
+  const alreadyHaveState = _.some(channel.sortedStates, ['stateHash', incomingState.stateHash]);
+
+  // Ignore older states that may be added via syncing
+  const isOldState = incomingState.turnNum < (supported?.turnNum || 0);
```
tomclose, line 16: It seems like we should go beyond not validating, and actually not store states that are older than our latest supported state. Is that accomplished elsewhere?

snario: We do not need to store states older than our latest supported state in some situations, e.g., a ledger update

snario (approved):

packages/server-wallet/src/utilities/validate-transition.ts
andrewgordstewart: What about `should'f`?

PR #2814, opened 2020-11-02
[nitro-protocol] Add a transfer method

geoknee:
Adding a fine-grained transfer method helps get around the case where there is a "bad" destination in an `assetOutcome` (say, that `reverts` when we try to transfer an asset to it using `transferAll`) -- currently this would effectively lock all of the funds in the channel. The `transfer` method added in this PR would be an escape hatch, allowing the specific destination to be specified in the method call and selectively paid out.

TODO 
---
- [x] `transfer` method
- [x] `transfer` test suite
- [ ] also add a `claim` operation
- [ ] `claim` test suite

### DIFF
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -17,6 +17,110 @@ contract AssetHolder is IAssetHolder {
     mapping(bytes32 => bytes32) public assetOutcomeHashes;
 
     /**
+     * @notice Transfers as many funds escrowed against `channelId` as can be afforded for a specific destination. Assumes no repeated entries. Performs no checks.
+     * @dev Transfers as many funds escrowed against `channelId` as can be afforded for a specific destination. Assumes no repeated entries. Performs no checks.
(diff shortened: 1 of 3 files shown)


packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -16,6 +16,113 @@ contract AssetHolder is IAssetHolder {
...
+            }
+            if (balance < _amount) {
+                balance = 0;
+            } else {
+                balance = balance.sub(_amount);
+            }
+        }
+
+        // effects
+        holdings[fromChannelId] = holdings[fromChannelId] -= affordsForDestination;
```
tomclose, line 64: Why the repetition?

tomclose (approved):
Looks good to me.

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: typo!

geoknee, line 64:
```suggestion
        holdings[fromChannelId] -= affordsForDestination;
```

PR #2816, opened 2020-11-02
Add validate transition tests

### LINKED ISSUE #2789: Add testing for Validation Transition method in the store
See https://github.com/statechannels/statechannels/pull/2788#pullrequestreview-520965345

> Ok with this as is.
> 
> If it's not a burden, can you add a regression unit test for this? An effective one might require a refactor to, for example, extract code that validates state transitions into a method not attached to the store.
> 
> A reasonable alternative is to create an issue for this request.

lalexgap:
Adds some tests of the validate transition function.

Fixes #2789 .


geoknee (approved):
I pegged on a commit that neatens up the test format a bit 😛

PR #2818, opened 2020-11-03
refactor: only process ledgers related to arguments to takeActions

snario:
Applications can run into confusing situations where a message pertaining to some channels A and B might return an outbound message related to some ledger L funding other channels C and D. In general this isn't too harmful, but if an application is relying on several concurrent streams of messages related to channels it might expect that sequences of exchanges relate to the same set of channels and don't involve channels related to other ledgers.

### DIFF
--- a/packages/server-wallet/src/wallet/index.ts
+++ b/packages/server-wallet/src/wallet/index.ts
@@ -593,23 +593,30 @@ export class Wallet extends EventEmitter<WalletEvent>
     let needToCrank = true;
     while (needToCrank) {
       await this.crankUntilIdle(channels, accumulator);
-      needToCrank = await this.processLedgerQueue(accumulator);
+      needToCrank = await this.processLedgerQueue(channels, accumulator);
     }
 
     return accumulator;
   };
 
-  private async processLedgerQueue({
-    outbox,
-    channelResults,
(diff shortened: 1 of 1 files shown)


packages/server-wallet/src/wallet/index.ts
```diff
@@ -593,23 +593,25 @@ export class Wallet extends EventEmitter<WalletEvent>
...
 
-    const ledgersToProcess = _.uniqBy(
-      await this.store.getAllPendingLedgerRequests(),
-      'ledgerChannelId'
-    );
+    // Fetch ledger channels related to the channels argument where related means, either:
+    // - The ledger channel is in the channels array
+    // - The ledger channel is funding one of the channels in the channels array
+    const ledgersToProcess = _.chain(await this.store.getAllPendingLedgerRequests())
+      .filter(r => _.some(channels, cid => cid === r.channelToBeFunded || r.ledgerChannelId))
```
tomclose, line 612: Does this do what you expect? I think you might need:
```ts
cid => cid === r.channelToBeFunded || cid === r.ledgerChannelId
```

tomclose (approved):
Happy for this to be merged without further review once the comment has been addressed.

PR #2824, opened 2020-11-03
ci: remove race condition in prepare step

tomclose:
Previously we had the following problem

1. Running `yarn prepare` at the root, would trigger a `yarn prepare` in the `nitro-protocol` package
2. This triggered the following set of commands: `run-s patch-package contract:compile build:typescript docgen`
3. Unfortunately the `patch-package` command looked like this: `cd ../.. && yarn patch-package`
4. ... which meant that all the other commands were run at the root level too, which meant that they ran through lerna in all packages
5. In particular, `build:typescript` would run a second time in the `client-api-schema`, potentially when it was in the middle of running the api-extractor, which didn't like it when the build folder was suddenly deleted

This should actually fix the problems with the canary-publish step, unlike the previous attempts: https://github.com/statechannels/statechannels/pull/2767, https://github.com/statechannels/statechannels/pull/2821


geoknee:
I'm not 100% sure if we actually need to run patch-package as a part of nitro-protocol's `prepare` step. Probably we don't?

geoknee:
Can we rip it out altogether? If the tests pass, we should be good.

snario:
> I'm not 100% sure if we actually need to run patch-package as a part of nitro-protocol's `prepare` step. Probably we don't?

I think it only needs to be in the monorepo root

geoknee (approved):

PR #2827, opened 2020-11-04
fix: do not require refunds or payouts to succeed

geoknee:


packages/nitro-protocol/contracts/ETHAssetHolder.sol
```diff
@@ -67,7 +67,7 @@ contract ETHAssetHolder is AssetHolder {
      * @param amount Quantity of wei to be transferred.
      */
     function _transferAsset(address payable destination, uint256 amount) internal override {
-        (bool success, ) = destination.call{value: amount}('');
-        require(success, 'Could not transfer asset');
+        destination.send(amount); // forwards a stipend of only 2300 gas, will not revert on failure
+        // We do not require success here, to block a griefing vector
```
geoknee, line 71: There's an issue with this approach in that an `AssetTransferred` event will be emitted regardless. So we _should_ inspect the `.send()` return value to properly conditionalise the event emission.

geoknee: Fixed in 7e276cc but requires breaking change (expect some tests to fail).

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -116,8 +116,7 @@ contract AssetHolder is IAssetHolder {
         // storage updated BEFORE asset transferred (prevent reentrancy)
 
         if (_isExternalDestination(destination)) {
-            _transferAsset(_bytes32ToAddress(destination), affordsForDestination);
-            emit AssetTransferred(fromChannelId, destination, affordsForDestination);
+            _transferAssetAndEmitIfSuccessful(fromChannelId, _bytes32ToAddress(destination), affordsForDestination);
         } else {
             holdings[destination] += affordsForDestination;
         }
```
geoknee, line 122: This raises another issue: we do not currently emit any event for **internal** transfers. (This PR bakes that issue in even more by changing the existing event to be incompatible with external transfers). 

@kerzhner and I agreed that in the short term we should emit `AssetTransferred` also for internal transfers.

Medium term, it might make more sense to move away from `Deposit` and `AssetTransferred` events to `HoldingsUpdated` and `AssetOutcomesUpdated`. The second containing the preimage of the `assetOutcomeHash` that is stored on chain. This is more useful to our wallet's chain-watching service, and would fulfil a principle of "emit events according to known changes in _this_ contract" rather than trying to emit events that relate to changes in other (possibly bad) contracts. (And we bypass the problem of deciding whether to emit the event if the external transfer failed). 

`AssetOutcomeUpdated(channelId, [])` would indicate that this channel had been finalised on chain and all assets in this holder had been paid out (the storage slot would still be deleted). That (along with similar events from any other relevant asset holders) could be the trigger for a wallet to consider a directly funded channel closed and withdrawn from (and therefore to wrap up some objectives and clear space in its DB).

We would emit `AssetOutcomeUpdated(channelId, residualAllocations)` in edge cases: e.g. the channel was underfunded for some reason, or we called the new `transfer` (not `transferAll`) method. Then, a wallet simply needs to inspect `residualAllocations` to decide whether it still cares about the channel. 

This proposal is also adopting a pattern of "emit the _new state_ and not the _delta_", which seems better to me.

geoknee:
#2844

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: The short term fix: #2844

PR #2832, opened 2020-11-04
style: remove eslint rule disabling and add type information to utils

snario:
This removes four TODOs from the codebase. Some eslint rules were disabled which allowed a lack of return values on functions and other basic typing related information.

We should periodically do this kind of work to keep the codebase clean


tomclose:
Could you add a description to the PR explaining what this change accomplishes?

tomclose (approved):

PR #2833, opened 2020-11-04
refactor: snake case all migrations

kerzhner:
In general, we shouldn't be modifying or renaming existing migrations. However, hopefully, no production/graph testnet deployments have run this migration yet.

So it would be nice to rename the migration instead of having to live with the incorrect case.


kerzhner:
Kebab case migration was introduced in https://github.com/statechannels/statechannels/commit/d07c5ab608221aef04b123148b6f08dca8a23473.

andrewgordstewart (approved):

PR #2835, opened 2020-11-05
Separate out multi-threaded wallet

tomclose:
This PR moves the multi-threaded logic out of the main wallet and into a separate class. The aim is both to simplify the logic in the main wallet by removing implementation details related to multiple threads, and to also make it clearer which parts of the wallet are affected by multi-threaded operation.

All the logic for multi-threadedness is now contained in the following class:
```ts
export class MultiThreadedWallet extends SingleThreadedWallet {
  private workerManager: WorkerManager;

  public static create(walletConfig?: ServerWalletConfig): MultiThreadedWallet {
    return new this(walletConfig);
  }

  protected constructor(walletConfig?: ServerWalletConfig) {
    super(walletConfig);
    this.workerManager = new WorkerManager(this.walletConfig);
  }

  async updateChannel(args: UpdateChannelParams): Promise<SingleChannelOutput> {
    return this.workerManager.updateChannel(args);
  }

  async pushMessage(rawPayload: unknown): Promise<MultipleChannelOutput> {
    return this.workerManager.pushMessage(rawPayload);
  }

  public async destroy(): Promise<void> {
    await this.workerManager.destroy();
    await super.destroy();
  }

  public async warmUpThreads(): Promise<void> {
    return this.workerManager.warmUpThreads();
  }
}
```

In order to accomplish this, I had to change the way wallets are constructed:
```ts
// old
const wallet = new Wallet(config);
// new
const wallet = Wallet.create(config);
```
The new `Wallet.create` method will return a `MultiThreadedWallet` or a `SingleThreadedWallet`, depending on the config. I have changed the `Wallet` constructor to be private, so that calling `new Wallet..` will give a type error. 

While the change to the construction api will require consumers to update their code, I think it's worth it for the greater flexibility we get with the approach. For example, we could in future change the `create` method to be `async` if we needed to.


snario:
I'm not sure if git will do this automatically or not but it looks like the move from `index.ts` to `wallet.ts` did a bulk line removal and then bulk line insertion which might make git blame a bit messy. Is it possible to write the history as a rename?

packages/server-wallet/src/wallet/multi-threaded-wallet/worker.ts
```diff
@@ -13,12 +13,13 @@ import {isStateChannelWorkerData} from './worker-data';
 const walletConfig: ServerWalletConfig = {
   ...(workerData as ServerWalletConfig),
```
geoknee, line 14: Can we get away without this type assertion?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -0,0 +1,819 @@
...
+    return this._pushMessage(wirePayload).catch(e => {
+      throw new PushMessageError('Error during pushMessage', {
+        thisWalletVersion: WALLET_VERSION,
+        payloadWalletVersion: wirePayload.walletVersion,
+        cause: e,
+      });
+    });
+  }
+
+  async _pushMessage(wirePayload: WirePayload): Promise<MultipleChannelOutput> {
```
geoknee, line 557: I feel like many of these methods should be `private. But we can tackle that in separate a PR.

geoknee (approved):

packages/server-wallet/src/wallet/multi-threaded-wallet/worker.ts
tomclose: I think this isn't easy to do, as `workerData` is the data that's passed to the thread, so typescript can't check it, as it has no control over how the thread is called. The alternative would be to write a validation function that checks the format of the `workerData`, throwing an error if it's the wrong format and type asserting otherwise. I think that would only accomplish failing sooner though, and probably isn't worth the effort at this stage.

PR #2837, opened 2020-11-05
test: anticipate race condition

geoknee:
This will hopefully prevent test flickers https://app.circleci.com/pipelines/github/statechannels/statechannels/10193/workflows/fea71c0a-1f5e-40e4-a48f-cadb3be53423/jobs/50136

### DIFF
--- a/packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
+++ b/packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
@@ -476,10 +476,8 @@ describe('Funding multiple channels concurrently (in bulk)', () => {
       allocations: [{allocationItems}],
     } = ledger;
 
-    expect(ledger).toMatchObject({
-      turnNum: 5,
-      status: 'running',
-    });
+    expect(ledger.turnNum === 5 || ledger.turnNum === 7).toBe(true);
+    // there is a race condition which means the turNum should be either 5 or 7
 
     expect(allocationItems).toHaveLength(N * 2);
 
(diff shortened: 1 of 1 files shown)


snario (approved):

packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
```diff
@@ -476,10 +476,8 @@ describe('Funding multiple channels concurrently (in bulk)', () => {
       allocations: [{allocationItems}],
     } = ledger;
 
-    expect(ledger).toMatchObject({
-      turnNum: 5,
-      status: 'running',
-    });
+    expect(ledger.turnNum === 5 || ledger.turnNum === 7).toBe(true);
+    // there is a race condition which means the turNum should be either 5 or 7
```
tomclose, line 480: Would it be better to remove the race condition? Something like:
```ts
const results = [await createMessageAndJoinBatch(), wait createMessageAndJoinBatch()]
```

tomclose: I'm also a bit confused about what race condition would cause different turn numbers here.

snario: I think the intent of the test is to prove that the wallet can handle multiple requests happening concurrently

geoknee: > Would it be better to remove the race condition? Something like:

I think that depends on whether we think  concurrent usage of the API (as currently expressed in this test) is a desirable / useful thing to test or support. 

About the race condition: @snario may understand this best?

geoknee: Seems that we should allow the race condition but relax the assertion on the ledger channel turn number. 

As I understand it, the wallet may or may not batch multiple ledger updates into one state update, depending on when it processes request. a54ccdc

PR #2838, opened 2020-11-05
Make sender and objective.participants optional in wire-format

tomclose:
We currently have a couple of properties on the wire-format that we're not using, and probably won't use, but make it harder to write wallet code because the wallet needs them available when constructing messages. This PR removes those properties.

The properties are:
* `participants` on the objective
   * objectives always reference one or more channels, which can provide the participants
   * we originally had participants on objectives because there was a thought that objectives could exist before the channels. This is no-longer the case in any of the current objectives, and I can't see it happening in the future.
* `sender` on the message
    * we don't use this in the wallet, and shouldn't need to, as the wallet is deliberately agnostic to how messages arrived
    * as messaging is handled by the app, it's more appropriate for the app to handle this. For example, the ongoing http request essentially does this in the indexer.

The plan for deploying is:
1. make the properties optional, but still provide them (this PR), so that the wire-format is backwards compatible
2. wait for wallets to upgrade
3. simultaneously with 2, we can start providing fake values for these properties (they aren't used for anything, so this should be ok)
4. once wallets have upgraded, we can remove the properties from the api altogether

### DIFF
--- a/packages/client-api-schema/client-api-schema.api.md
+++ b/packages/client-api-schema/client-api-schema.api.md
@@ -468,7 +468,7 @@ export interface JsonRpcResponse<ResponseType = object> {
 export interface Message {
     data: unknown;
     recipient: string;
-    sender: string;
+    sender?: string;
 }
 
 // Warning: (ae-incompatible-release-tags) The symbol "MessageQueuedNotification" is marked as @public, but its signature references "JsonRpcNotification" which is marked as @beta
@@ -543,7 +543,9 @@ export type StateChannelsNotification = ChannelProposedNotification | ChannelUpd
 //
 // @public (undocumented)
 export type StateChannelsNotificationType = {
-    [T in StateChannelsNotification['method']]: [FilterByMethod<StateChannelsNotification, T>['params']];
+    [T in StateChannelsNotification['method']]: [
+        FilterByMethod<StateChannelsNotification, T>['params']
+    ];
 };
 
 // @public (undocumented)
--- a/packages/client-api-schema/client-api-schema.api.md
+++ b/packages/client-api-schema/client-api-schema.api.md
@@ -543,9 +543,7 @@ export type StateChannelsNotification = ChannelProposedNotification | ChannelUpd
 //
 // @public (undocumented)
 export type StateChannelsNotificationType = {
-    [T in StateChannelsNotification['method']]: [
-        FilterByMethod<StateChannelsNotification, T>['params']
-    ];
+    [T in StateChannelsNotification['method']]: [FilterByMethod<StateChannelsNotification, T>['params']];
 };
 
 // @public (undocumented)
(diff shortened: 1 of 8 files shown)


geoknee:
To prevent DoS attacks it might be desirable to filter out messages from blacklisted peers -- which could be an argument for keeping `sender`. OTOH I suppose it is also easy for a malicious wallet to spoof the  `sender` field, and as you say the app can handle blacklisting via IP address. So I think I agree that we don't want it.

andrewgordstewart:
> To prevent DoS attacks it might be desirable to filter out messages from blacklisted peers -- which could be an argument for keeping `sender`. OTOH I suppose it is also easy for a malicious wallet to spoof the `sender` field, and as you say the app can handle blacklisting via IP address. So I think I agree that we don't want it.

To do this, the recipient would have to fill out the sender field anyway.

package.json
```diff
@@ -60,6 +60,9 @@
   "workspaces": {
     "packages": [
       "packages/*"
+    ],
+    "nohoist": [
+      "**/html-minifier-terser"
```
geoknee, line 65: Can you leave a link to a relevant docusaurus GH issue on this PR?

packages/docs-website/package.json
```diff
@@ -43,7 +43,7 @@
     "rename-version": "docusaurus-rename-version",
     "start": "yarn trigger-api-generation && yarn api-documenter && docusaurus start",
     "swizzle": "docusaurus swizzle",
-    "trigger-api-generation": "cd ../.. && yarn lerna run generate-api cd packages/docs-website",
+    "trigger-api-generation": "yarn lerna run generate-api",
```
geoknee, line 46: 👍

.circleci/config.yml
```diff
@@ -57,19 +52,14 @@ save_built_artifacts: &save_built_artifacts
     paths:
       # Various TypeScript emissions
       - packages/channel-client/lib
-      - packages/iframe-channel-provider/lib
       - packages/client-api-schema/lib
       - packages/devtools/lib
       - packages/jest-gas-reporter/lib
       - packages/nitro-protocol/lib
-      - packages/simple-hub/lib
```
geoknee, line 65: 😄

geoknee (approved):

geoknee:
This PR does cause some broken links on the website https://5faa7b0b938db7ad157338ab--nitro-protocol.netlify.app/app-devs/make-api-calls


Could you make a new GH issue to revisit and fix those, and/or reinstate the typescript API docs for legacy packages (we can do this by installing them from npm).

package.json
tomclose: https://github.com/facebook/docusaurus/issues/3515#issuecomment-707145367

tomclose:
Issue for broken links: https://github.com/statechannels/statechannels/issues/2860

PR #2844, opened 2020-11-06
[nitro-protocol] Do not require external calls to succeed

geoknee:
- Limit gas forwarded when making external calls (using `.send`)
- emit AssetTransferred unconditionally, even for internal updates to `holdings`.

### DIFF
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -61,6 +61,8 @@ contract AssetHolder is IAssetHolder {
             }
         }
 
+        require(affordsForDestination > 0, '_transfer fromChannel allocates 0 to destination');
+        
         // effects
         holdings[fromChannelId] -= affordsForDestination;
 
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -430,13 +430,13 @@ contract AssetHolder is IAssetHolder {
                 } else {
                     holdings[allocation[j].destination] += payouts[j];
                 }
-            }
             // Event emitted regardless of success of external calls
             emit AssetTransferred(
                         guarantorChannelId,
                         allocation[j].destination,
                         payouts[j]
                     );
+            }
         }
     }
 
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -18,8 +18,8 @@ contract AssetHolder is IAssetHolder {
     mapping(bytes32 => bytes32) public assetOutcomeHashes;
 
     /**
-     * @notice Transfers as many funds escrowed against `channelId` as can be afforded for a specific destination. Assumes no repeated entries. Performs no checks.
-     * @dev Transfers as many funds escrowed against `channelId` as can be afforded for a specific destination. Assumes no repeated entries. Performs no checks.
+     * @notice Transfers as many funds escrowed against `channelId` as can be afforded for a specific destination. Assumes no repeated entries. Does not check allocationBytes against on chain storage.
+     * @dev Transfers as many funds escrowed against `channelId` as can be afforded for a specific destination. Assumes no repeated entries. Does not check allocationBytes against on chain storage.
      * @param fromChannelId Unique identifier for state channel to transfer funds *from*.
      * @param allocationBytes The abi.encode of AssetOutcome.Allocation
      * @param destination External destination or channel to transfer funds *to*.
@@ -130,8 +130,8 @@ contract AssetHolder is IAssetHolder {
     }
 
     /**
-     * @notice Transfers the funds escrowed against `channelId` to the beneficiaries of that channel. No checks performed.
-     * @dev Transfers the funds escrowed against `channelId` and transfers them to the beneficiaries of that channel. No checks performed.
+     * @notice Transfers the funds escrowed against `channelId` to the beneficiaries of that channel. Does not check allocationBytes against on chain storage.
+     * @dev Transfers the funds escrowed against `channelId` and transfers them to the beneficiaries of that channel. Does not check allocationBytes against on chain storage.
      * @param channelId Unique identifier for a state channel.
      * @param allocationBytes The abi.encode of AssetOutcome.Allocation
      */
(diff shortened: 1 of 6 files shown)


packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -61,6 +61,8 @@ contract AssetHolder is IAssetHolder {
             }
         }
 
+        require(affordsForDestination > 0, '_transfer fromChannel allocates 0 to destination');
```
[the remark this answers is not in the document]
kerzhner: I think this `require` means that this function cannot be used for `transferAll`. For example, let's say we finalize a channel with Alice: 0Eth, Bob 5Eth and then call transferAll.

I am assuming that is ok?

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -113,14 +115,17 @@ contract AssetHolder is IAssetHolder {
             );
         } 
 
-        // storage updated BEFORE asset transferred (prevent reentrancy)
 
+        // Event emitted regardless of success of external calls
+        emit AssetTransferred(fromChannelId, destination, affordsForDestination);
+
+        // storage updated BEFORE asset transferred (prevent reentrancy)
```
kerzhner, line 122: It seems like this comment doesn't apply any longer? Do we understand why it's important to emit `AssetTransferred` after updating storage?

packages/nitro-protocol/contracts/ETHAssetHolder.sol
```diff
@@ -58,8 +58,8 @@ contract ETHAssetHolder is AssetHolder {
 
         // refund whatever wasn't deposited.
         uint256 refund = amount.sub(amountDeposited);
-        (bool success, ) = msg.sender.call{value: refund}('');
-        require(success, 'Could not refund excess funds');
+        msg.sender.send(refund); // forwards a stipend of only 2300 gas, will not revert on failure
```
kerzhner, line 61: I am far from expert on this matter, but I wanted to highlight [this](https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now/)

kerzhner (approved):

packages/nitro-protocol/contracts/ETHAssetHolder.sol
```diff
@@ -69,7 +69,7 @@ contract ETHAssetHolder is AssetHolder {
      * @param amount Quantity of wei to be transferred.
      */
     function _transferAsset(address payable destination, uint256 amount) internal override {
-        (bool success, ) = destination.call{value: amount}('');
-        require(success, 'Could not transfer asset');
+        destination.send(amount); // forwards a stipend of only 2300 gas, will not revert on failure
```
kerzhner, line 72: Does this mean that if the assets are transferred to a contracts with an expensive fallback functions, the gas limit of 2300 will result in the assets being stuck in the asset holder forever?

packages/nitro-protocol/contracts/AssetHolder.sol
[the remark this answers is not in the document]
geoknee: Yes,  "Performs no checks" is too vague. Have updated in 11f587cfd530ab1664ef138ce9af3f2a38e4d598.

You're right that we couldn't easily reuse `_transfer` inside `transferAll`. For now, the two methods have entirely separate implementations, mainly to be more gas efficient.

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: I think the comment does apply, as it doesn't mention events. Regardless I do think it is slightly strange to have the event emitted before the assets are actually transferred, so I will move it to afterwards. Also, in the case of reentrancy, we could potentially get more events emitted than we should.

The more important thing is to do any external contract calls only after updating storage.

packages/nitro-protocol/contracts/ETHAssetHolder.sol
geoknee: 👍 This is a marginal call that we have discussed with the auditors. On balance the current approach of forwarding all gas was deemed a griefing vector that was worth blocking, despite the downside of "having a hard dependency on gas costs".

geoknee: For reference:

[ _5.3 Gas siphoning attack possible in “transfer all” and “claim all”-related methods (Medium)_](https://consensys.net/diligence/audits/private/ljryv8j1b18epn/)

Note that there is no analogous solution offered in this PR for the `ERC20AssetHolder`, since it is difficult to know how much gas to forward (seeing as it depends on the token contract implementation).

packages/nitro-protocol/contracts/ETHAssetHolder.sol
geoknee: Yes.

Perhaps a better solution would be to rename this function` _transferAssetWithGasStipendAndIgnoreFailures`, and have that used by `_transferAll` and `_claimAll` (which are in some sense "collective" methods called by one participant on behalf of all participants, and therefore the ones that need to be careful about gas siphoning or other griefing problems such as transfer failure), and have a regular `_transferAsset` function which forwards all gas and requires success, and have _that_ used by `_transfer` (which is not a "collective" method).

geoknee: Of course this isn't a really satisfactory solution, since funds could still get locked forever if `transferAll` is called first.

Another idea would be to maintain the requirement for the external call to succeed, and tackle[ _5.2 Recipient can block transferAsset/payouts for all participants (Major)_](https://consensys.net/diligence/audits/private/ljryv8j1b18epn/) in a different way -- namely, by the recently-introduced `transfer` function. In the case of a `_transferAsset` reverting, the participants would need to fall back to each calling `transfer` separately (and spend more gas in the process).

kerzhner: Thank for the explanations @geoknee. I think it is good to highlight these issues, but, to be clear, I am not pushing for any immediate action :)

geoknee:
After a chat with @kerzhner, we decided that this PR is probably going a bit too far and unnecessarily so. 

Agreed approach:

- lean on `transfer` (and remember to implement a `claim` method) as a way of extracting funds for a single destination, which won't be impacted by any _other_ destination being "broken" or "malicious"
- maintain `transferAll` as a happy-case, may-revert or may-leak-gas method. (especially given that it is already somewhat altruistic)
- *always* require external calls to succeed, and *always* forward all available gas: to avoid funds being locked in the asset holder.

- keep the changes from this PR that cause `AssetTransferred` to be emitted for internal and external transfers.

PR #2847, opened 2020-11-06
[nitro-protocol] Run chain tests concurrently

geoknee:
This cuts down nitro test run time by about a half (10 mins to about 5) on circle. 

It also closes the gap between local development (where tests will run concurrently, and *fail* some of the time because of that) and circle (where they do not). 

The key fix was not asserting on the balance of one of our test accounts (e.g. one that deploys contracts). 

Other fixes that I put in to avoid interference between tests: 

- generate a psuedorandom nonce for each channel used in each test, using the hash of the test name as a seed.

### DIFF
--- a/packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
+++ b/packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
@@ -169,10 +169,7 @@ describe('concludePushOutcomeAndTransferAll', () => {
       if ('ERC20' in heldBefore) {
         await (
           await Token.transfer(ERC20AssetHolder.address, BigNumber.from(heldBefore.ERC20.c))
-        ).wait();
-        expect(await Token.balanceOf(ERC20AssetHolder.address)).toStrictEqual(
-          BigNumber.from(heldBefore.ERC20.c)
-        );
+        ).wait(); // if the tx is mined, we know we the transfer succeeded
       }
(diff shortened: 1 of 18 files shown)


packages/nitro-protocol/test/test-helpers.ts
```diff
@@ -385,3 +385,7 @@ export async function writeGasConsumption(
     if (err) throw err;
   });
 }
+
+export function getRandomNonce(seed: string): number {
+  return Number.parseInt(ethers.utils.id(seed).slice(2, 11), 16);
+}
```
geoknee, line 391: Pseudorandom nonce generator.

packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
```diff
@@ -168,10 +169,7 @@ describe('concludePushOutcomeAndTransferAll', () => {
       if ('ERC20' in heldBefore) {
         await (
           await Token.transfer(ERC20AssetHolder.address, BigNumber.from(heldBefore.ERC20.c))
-        ).wait();
-        expect(await Token.balanceOf(ERC20AssetHolder.address)).toStrictEqual(
-          BigNumber.from(heldBefore.ERC20.c)
```
geoknee, line 173: This is highly prone to failing if another test is running concurrently.

snario:
This is god's work! 👍

tomclose (approved):

Issue #2815, opened 2020-11-02
Destinations in chain events are lower cased


kerzhner:
Destinations stored by the server wallet are checksummed. Destinations that arrive in chain events (such as `AssetTransferred`) are lower case. It would be nice to
- Understand what component lower cases destinations.
- Have consistent casing of values of all variables that store a destination in the server wallet.

PR #2850, opened 2020-11-07
Intersection types for byte strings

kerzhner:
In our world, byte strings are hex strings. For hex strings, case of the string carries no information. However, when casing of hex strings is inconsistent, comparison mistakes happen. This PR:
- Creates new intersection types for `Address` and `PrivateKey` that enforce lowercase byte strings.
- Eliminates the `Address` type alias that equates to a string.
- Adds database constraints that enforce lower casing.
- Pins `xstate-wallet` to an older version of the `wallet-core` to avoid converting `xstate-wallet` to the new intersection types.

The benefits of using intersection types across the wallet are:
- ensure consistent case for byte strings.
- add length constraints where required (ie addresses, private keys, channel ids, etc)
- ensure that the above are validated at the boundaries of the wallet.

There is quite a bit of follow up work that can be done, but this PR is already large:
- Convert [all type-aliases](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/type-aliases.ts#L1) to intersection types.
- Add more database constraints for every column that stores an intersection type (including json columns).

Fixes https://github.com/statechannels/statechannels/issues/2815


packages/nitro-protocol/src/signatures.ts
```diff
@@ -15,7 +15,7 @@ export function getStateSignerAddress(signedState: SignedState): string {
   const {channel} = signedState.state;
   const {participants} = channel;
 
-  if (participants.indexOf(recoveredAddress) < 0) {
+  if (participants.map(p => p.toLowerCase()).indexOf(recoveredAddress.toLowerCase()) < 0) {
```
tomclose, line 18: Are these conversions still required? It seems like we should know that the addresses are all lower case by this point?

packages/server-wallet/src/db/migrations/20201104115400_lower_case_constraints.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  await dropConstraint(knex, app_bytecode, 'app_definition_is_address');
+  await dropConstraint(knex, app_bytecode, 'app_bytecode_is_bytes');
+
+  await dropConstraint(knex, 'funding', 'asset_holder_is_address');
+}
+
+export async function up(knex: Knex): Promise<any> {
+  await dropOldByteConstraints(knex);
+  await addByteContraints(knex);
```
tomclose, line 46: Should this include an operation to lowercase any existing columns? Otherwise it looks like the migration could fail here.

tomclose (approved):
A few questions to address, but happy for this to be merged without further review after.

packages/nitro-protocol/src/signatures.ts
kerzhner: Good question. In this specific case, `recoveredAddress` will be a checksum as that address is [derived using ethers ](https://github.com/statechannels/statechannels/blob/3ac43c32923f09ab9a1c21efa4faa7243cc2b957/packages/nitro-protocol/src/signatures.ts#L14). The question, then, is should we assume that the participants array contains lower case addresses?

This code is part of the `nitro-protocol` package. We currently do **not** enforce/assume lowercase addresses or bytes in nitro protocol. We only do so in `server-wallet` and `wallet-core` packages.  Why don't we not assume lower case addresses in `nitro-protocol`? My thinking was that I did not want to create a breaking change in `nitro-protocol`. I was primarily thinking about not breaking`xstate-wallet`.

However, `xstate-wallet` depends on `wallet-core`. So the breaking changes in `wallet-core` required pinning `xstate-wallet` to a version of `wallet-core` before this PR. So perhaps we should just introduce the breaking change into `nitro-protocol`?

packages/server-wallet/src/db/migrations/20201104115400_lower_case_constraints.ts
kerzhner: That's a good point. In fact, the migration will fail if there is existing data as, before this PR, addresses were checksummed.

packages/nitro-protocol/src/signatures.ts
kerzhner: After giving this more thought: if we are going to enforce lower cases in `nitro-protocol`, should `Address` and `Destination` intersection types be defined in the `nitro-protocol` package?

kerzhner: During the discussion in standup, the conclusion was to enforce address and bytes cases in nitro protocol. In the interest of reducing the changeset of this PR, leaving this work for future PRs: https://github.com/statechannels/statechannels/issues/2859

packages/server-wallet/src/db/migrations/20201104115400_lower_case_constraints.ts
kerzhner: Now that addresses are checksummed, I don't think this comment is applicable.

kerzhner:
8b1b768 through 0b30ed9 convert this PR to use checksummed addresses instead of lowercase addresses. We made this decision due to:
1. Existing databases store checksummed addresses. It is nice to avoid having to convert existing data.
1. Ethers return values always checksum addresses.

Destinations are treated [a bit strangely at the moment](https://github.com/statechannels/statechannels/blob/lower-case/packages/wallet-core/src/utils/outcome.ts#L102):
1. If a destination is a channel id, the destination is expected to be lower case (just like all other byte values except addresses).
1. If a destination is zeros prefixed to an ethereum address, the ethereum address is checksummed.

Do we want to adhere to this heuristic? @snario @tomclose @andrewgordstewart

PR #2852, opened 2020-11-09
[server-wallet] Export a few more symbols

geoknee:


packages/server-wallet/src/index.ts
```diff
@@ -1,4 +1,10 @@
 export {Payload as Message} from '@statechannels/wallet-core';
 
-export {Wallet, SingleChannelOutput} from './wallet';
+export {
+  Wallet,
+  SingleChannelOutput,
+  SingleThreadedWallet,
+  MultiThreadedWallet,
+  ServerWalletConfig,
+} from './wallet';
```
geoknee, line 9: These are useful types for the consuming application to import.

tomclose: I'm not sure that we want to make `SingleThreadedWallet` or `MultiThreadedWallet` part of the public interface - I'd consider them more an implementation detail of the `Wallet` class.

geoknee: ok 8660ede. The motivation was partly due to the api-documenter tool not generating anything very useful at the moment, since we use an object and a class with the same name `Wallet`.

geoknee:
Note also the generated api report is out of sync currently, this PR fixes that issue.

snario (approved):

PR #2853, opened 2020-11-09
Nuclear option 💣

tomclose:
This PR removes a load of packages from the monorepo. It was prompted by a commit that Mike did in #2850, where he pinned the xstate-wallet to old versions of wallet-core and wire-format. The logic goes:

1. If we're not planning on using xstate-wallet in the future, and don't have anyone using it currently, we shouldn't spend time and effort keeping the xstate-wallet in sync with current changes
2. So let's pin xstate-wallet to old versions of packages in the monorepo.
3. But if we're going to do that, we might as well just remove the xstate-wallet, as we (a) still have it in the history (and could separate it out into a different repo, if needed), and (b) that's less confusing than having a monorepo where some packages are pinned.
4. And if we're removing xstate-wallet, we might as well remove all the packages that are only used by it.

Interested to hear what people think.


snario:
I think it's important that the code be easily publicly reference-able. Maybe you could move it into a separate repo e.g., `legacy-monorepo`? https://web3torrent.statechannels.org/ is still live and functional and still has people playing with it, so we should ensure that stays up too, as well is easily changeable for when we want to update it.

tomclose:
Here's a repo with the packages extracted: https://github.com/statechannels/legacy-packages. I've filtered the history down to just those packages, so the repo is much smaller.

andrewgordstewart:
This sounds pretty reasonable. I'd be on board if we think that both
A. In the future, when we consolidate the browser- and server- wallet, we do not anticipate refactoring the `xstate-wallet` to fit the bill.

I think (A) is true. And if (A) is false, then we can revert a single commit (1805187712d917f023c99736da19e744f31fa66a) and get right back where we were. The only consequence I can think of is that the `git blame` will be annoying for any changes pre 1805187712d917f023c99736da19e744f31fa66a

snario:
Are we sure that the existing sites hosted on Netlify will stay running after this is merged?

tomclose:
@snario which existing sites do you mean? The docs?

andrewgordstewart:
> I think it's important that the code be easily publicly reference-able. Maybe you could move it into a separate repo e.g., `legacy-monorepo`? https://web3torrent.statechannels.org/ is still live and functional and still has people playing with it, so we should ensure that stays up too, as well is easily changeable for when we want to update it. (same with https://rps.statechannels.org)

The code is still easily referencable -- it's in this repo's git history!

Public references to `master` is not good on a website anyway, since `master` can change underneath your feet.

snario:
https://xstate-wallet.statechannels.org for example

snario:
> > I think it's important that the code be easily publicly reference-able. Maybe you could move it into a separate repo e.g., `legacy-monorepo`? https://web3torrent.statechannels.org/ is still live and functional and still has people playing with it, so we should ensure that stays up too, as well is easily changeable for when we want to update it. (same with https://rps.statechannels.org)
> 
> The code is still easily referencable -- it's in this repo's git history!
> 
> Public references to `master` is not good on a website anyway, since `master` can change underneath your feet.

Sure, but most people interested in seeing code will navigate to https://github.com/statechannels and it will not be obvious they would have to go into a specific branch to see code we have running online.

andrewgordstewart:
> > > I think it's important that the code be easily publicly reference-able. Maybe you could move it into a separate repo e.g., `legacy-monorepo`? https://web3torrent.statechannels.org/ is still live and functional and still has people playing with it, so we should ensure that stays up too, as well is easily changeable for when we want to update it. (same with https://rps.statechannels.org)
> > 
> > 
> > The code is still easily referencable -- it's in this repo's git history!
> > Public references to `master` is not good on a website anyway, since `master` can change underneath your feet.
> 
> Sure, but most people interested in seeing code will navigate to https://github.com/statechannels and it will not be obvious they would have to go into a specific branch to see code we have running online.

Most people who navigate to https://github.com/statechannels would have no idea where to look to see what code is running online. We would have to provide links anyway, and those links should not point at `master`!

PR #2857, opened 2020-11-09
refactor: remove 'stage' concept and document 'status'

snario:
This always caused confusion and was error prone, especially since PostFundSetup as a 'stage' was not determinable based on a single state. Also, we rarely (only two instances) used the `stage` helper

### DIFF
--- a/packages/server-wallet/src/protocols/close-channel.ts
+++ b/packages/server-wallet/src/protocols/close-channel.ts
@@ -1,10 +1,10 @@
-import {BN, checkThat, isSimpleAllocation, State} from '@statechannels/wallet-core';
+import {BN, checkThat, isSimpleAllocation} from '@statechannels/wallet-core';
 import {Transaction} from 'knex';
 
 import {Store} from '../wallet/store';
 import {Bytes32} from '../type-aliases';
 
-import {Protocol, ProtocolResult, ChannelState, stage, Stage} from './state';
+import {Protocol, ProtocolResult, ChannelState} from './state';
 import {
   completeObjective,
   Withdraw,
@@ -23,13 +23,14 @@ export type ProtocolState = {
   ledgerChannelId?: Bytes32;
 };
 
-const stageGuard = (guardStage: Stage) => (s: State | undefined): s is State =>
-  !!s && stage(s) === guardStage;
-
-const isFinal = stageGuard('Final');
-
 function everyoneSignedFinalState(ps: ProtocolState): boolean {
-  return (ps.app.support || []).every(isFinal) && isFinal(ps.app.latestSignedByMe);
+  return (
+    !!ps.app.latestSignedByMe &&
+    ps.app.latestSignedByMe.isFinal &&
+    !!ps.app.supported &&
+    ps.app.supported.isFinal &&
+    (ps.app.support || []).every(s => s.isFinal)
+  );
 }
 
 // TODO: where is the corresponding logic for ledger channels?
@@ -66,9 +67,10 @@ const isMyTurn = (ps: ProtocolState): boolean =>
 //
 const signFinalState = (ps: ProtocolState): ProtocolResult | false =>
   !!ps.app.supported && // there is a supported state
-  !isFinal(ps.app.latestSignedByMe) && // I haven't yet signed a final state
+  !!ps.app.latestSignedByMe &&
+  !ps.app.latestSignedByMe.isFinal && // I haven't yet signed a final state
   // if there's an existing final state double-sign it
-  ((isFinal(ps.app.supported) &&
+  ((ps.app.supported.isFinal &&
     signState({
       channelId: ps.app.channelId,
       ...ps.app.supported,
--- a/packages/server-wallet/src/protocols/open-channel.ts
+++ b/packages/server-wallet/src/protocols/open-channel.ts
@@ -1,11 +1,11 @@
-import {BN, isSimpleAllocation, checkThat, State} from '@statechannels/wallet-core';
+import {BN, isSimpleAllocation, checkThat, SignedState} from '@statechannels/wallet-core';
 import _ from 'lodash';
 import {Transaction} from 'knex';
 
 import {Store} from '../wallet/store';
 import {Bytes32} from '../type-aliases';
 
-import {Protocol, ProtocolResult, ChannelState, stage, Stage} from './state';
+import {Protocol, ProtocolResult, ChannelState} from './state';
 import {
   signState,
   noAction,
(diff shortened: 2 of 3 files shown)


packages/server-wallet/src/protocols/open-channel.ts
```diff
@@ -33,16 +33,13 @@ export type LedgerFundedAppState = AppState & {
...
+  !!channelState.supported && channelState.supported.turnNum < channelState.participants.length;
 
-const isPrefundSetup = stageGuard('PrefundSetup');
-
-// These are currently unused, but will be used
-// const isRunning = stageGuard('Running');
-// const isPostfundSetup = stageGuard('PostfundSetup');
-const isRunning = stageGuard('Running');
-// const isMissing = (s: State | undefined): s is undefined => stage(s) === 'Missing';
+const isPostFundSetupSupported = (channelState: ChannelState): boolean =>
```
andrewgordstewart, line 39: This name is confusing. In a 2-party channel, state `2` can be supported, and state `2` is a post-fund setup. But if state `3` is not supported, then `isPostFundSetupSupported` would return `false`.

snario: How about `isEveryPostFundSetupStateSigned` or `isEverySupportPostFund` or `isSupportChainPostFunded`

andrewgordstewart: `isRunning` is exactly what the implementation is, which is what I would suggest.

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -23,13 +23,14 @@ export type ProtocolState = {
...
-const stageGuard = (guardStage: Stage) => (s: State | undefined): s is State =>
-  !!s && stage(s) === guardStage;
-
-const isFinal = stageGuard('Final');
-
 function everyoneSignedFinalState(ps: ProtocolState): boolean {
-  return (ps.app.support || []).every(isFinal) && isFinal(ps.app.latestSignedByMe);
+  return (
+    !!ps.app.latestSignedByMe &&
+    ps.app.latestSignedByMe.isFinal &&
```
andrewgordstewart, line 29: I don't see why this is needed.
```suggestion
```
- If `ps.app.support.every(isFinal)`, then I have signed a final state
- My wallet should not be signing any other states once it signs a final state.

packages/server-wallet/src/protocols/state.ts
```diff
@@ -50,47 +49,54 @@ type SignedByMe = {latestSignedByMe: SignedStateWithHash};
...
-    : state.isFinal
-    ? 'Final'
-    : state.turnNum <= state.participants.length - 1
-    ? 'PrefundSetup'
-    : state.turnNum <= state.participants.length * 2 - 1
-    ? 'PostfundSetup'
-    : 'Running';
+export const status = (channelState: ChannelState): ChannelStatus => {
+  const {supported, latest, latestSignedByMe, support} = channelState;
+  const {participants} = supported ?? latest;
```
andrewgordstewart, line 60: If `supported` is defined, so is `latest`.
```suggestion
  const {participants} = latest;
```

packages/server-wallet/src/protocols/state.ts
```diff
@@ -50,47 +49,54 @@ type SignedByMe = {latestSignedByMe: SignedStateWithHash};
...
+  if (latest.isFinal) {
+    if (support?.every(s => s.isFinal)) {
+      return 'closed'; // the entire support chain isFinal
+    } else {
+      return 'closing'; // at least one isFinal state proposed
```
andrewgordstewart, line 65: This might be incorrect. It could be that something in the `support` is `isFinal`, but `latest` is not `isFinal`. Is there another bit of code that guarantees this cannot happen?

packages/server-wallet/src/protocols/state.ts
```diff
@@ -50,47 +49,54 @@ type SignedByMe = {latestSignedByMe: SignedStateWithHash};
...
+    }
+  } else {
+    if (latest.turnNum >= participants.length * 2) {
```
andrewgordstewart, line 68: In this branch, we know that `latest.isFinal` is true. That does not necessarily imply that there are no final states.

packages/server-wallet/src/protocols/state.ts
```diff
@@ -50,47 +49,54 @@ type SignedByMe = {latestSignedByMe: SignedStateWithHash};
...
-  !state
-    ? 'Missing'
-    : state.isFinal
-    ? 'Final'
-    : state.turnNum <= state.participants.length - 1
-    ? 'PrefundSetup'
-    : state.turnNum <= state.participants.length * 2 - 1
-    ? 'PostfundSetup'
-    : 'Running';
+export const status = (channelState: ChannelState): ChannelStatus => {
```
andrewgordstewart, line 58: I'm not sure I agree with the change. It is harder for me to understand the implementation of this function, and it appears like it could be incorrect?

packages/server-wallet/src/protocols/close-channel.ts
snario: I think it's just because the unit tests provide `undefined` for `support` that I did this

packages/server-wallet/src/protocols/state.ts
snario: Is that actually possible? [It shouldn't be possible to transition from an `isFinal` state to a state that is not `isFinal`](https://github.com/statechannels/statechannels/blob/357a488076e4129f5029093412c0a9062d358d4d/packages/nitro-protocol/contracts/ForceMove.sol#L517-L520), so any support with an `isFinal` state must have `latest.isFinal`?

packages/server-wallet/src/protocols/state.ts
snario: I think it does (ref: https://github.com/statechannels/statechannels/pull/2857#discussion_r523023180)

packages/server-wallet/src/protocols/state.ts
snario: I think the previous functions were harder to understand and this function more explicitly shows the current logic branch that the previous code was hiding the complexity of — the fact of the matter is that these statuses are somewhat subjective and still not decided upon, which the code also references to the github issue for

packages/server-wallet/src/protocols/state.ts
andrewgordstewart: Don't we store unsupported states? What stops Bob from sending Alice a later state with `isFinal: false`?

packages/server-wallet/src/protocols/state.ts
andrewgordstewart: Maybe I am confused because I don't know what branch I am in. I think this function is difficult to understand.

andrewgordstewart: Why can't it be the case that we have a supported `isFinal: true` state, but Bob sent a `isFinal: false` state with a higher turn number?

packages/server-wallet/src/protocols/close-channel.ts
andrewgordstewart: The following line should return `false` in that case?

lalexgap:
This seems pretty stale so I'm going to close it. Please re-open if still relevant!

PR #2861, opened 2020-11-10
build: remove now-unused code and configurations 

snario:
Post-work for https://github.com/statechannels/statechannels/pull/2853


.env.integration-tests
```diff
@@ -11,13 +11,3 @@ LOG_DESTINATION = 'console'
 SHOW_VERBOSE_GANACHE_OUTPUT = 'true'
```
kerzhner, line 11: Are these env files used anywhere at this point? I suspect not...

kerzhner (approved):

PR #2862, opened 2020-11-10
refactor: migrate chain-service into its own package

snario:
This was initially the intention, but we placed it inside `server-wallet` for short-term convenience.

```
.
├── deployment
│   └── deploy.ts
├── jest
│   ├── chain-setup.ts
│   ├── jest.config.js
│   └── test-teardown.ts
├── src
│   ├── __chain-test__
│   │   ├── fixtures
│   │   │   ├── participants.ts
│   │   │   ├── signing-wallets.ts
│   │   │   └── utils.ts
│   │   ├── chain-service.test.ts
│   │   └── stress.test.ts
│   ├── chain-service.ts
│   ├── config.ts
│   ├── index.ts
│   └── types.ts
├── README.md
├── package.json
└── tsconfig.json

5 directories, 16 files
```

See [README.md](https://github.com/statechannels/statechannels/blob/liam/chain-service/packages/chain-service/README.md)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,7 @@ save_dep: &save_dep
     paths:
       - node_modules
       - packages/client-api-schema/node_modules
+      - packages/chain-service/node_modules
       - packages/devtools/node_modules
       - packages/docs-website/node_modules
       - packages/jest-gas-reporter/node_modules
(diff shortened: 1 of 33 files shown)


kerzhner:
My preference is to delay factoring out the chain service into its own package.
1. Each new package contains configuration overhead.
1. At the moment, the chain service is used as a library by only the server wallet. A chain service in a separate directory perhaps seems sufficient for separation of concerns at the moment.

That being said, there are conversations around NOT passing an ethereum private key to the server wallet. With that strategy, the application would instantiate the chain service and pass the instance to the server wallet. If we end up going down that route, we should factor out the chain service into a separate package.

andrewgordstewart:
This is out of date and controversial.

PR #2864, opened 2020-11-11
[breaking change (new config option)] consolidate chain service configuration options 

snario:
~The `ChainService` should not be instantiated inside the `server-wallet` as it is fundamentally an external component that is designed to be provided _to_ the `server-wallet`. Similarly, the implementation details of the wallet should not need to know about mocks or test fixtures, that should be left to the tests. So, this PR moves the instantiation of a `ChainService` into the constructor of a `server-wallet`. Applications are expected to provide it in the constructor.~

~It seems reasonable to force the `ChainService` to be in the constructor as without it the `server-wallet` is not able to do things like make deposits, withdrawals, or monitor the chain. These are fundamental requirements of a state channels wallet.~

Simplified this PR to simply consolidate wallet config options into one place as they relate to the chain service after some discussions

### DIFF
--- a/packages/server-wallet/deployment/deploy.ts
+++ b/packages/server-wallet/deployment/deploy.ts
@@ -1,9 +1,7 @@
 import {Address} from '@statechannels/client-api-schema';
 import {ContractArtifacts} from '@statechannels/nitro-protocol';
-import {GanacheDeployer} from '@statechannels/devtools';
+import {ETHERLIME_ACCOUNTS, GanacheDeployer} from '@statechannels/devtools';
 import {Wallet} from 'ethers';
-
-import {defaultTestConfig} from '../src/config';
 
 // NOTE: deploying contracts like this allows the onchain service package to
 // be easily extracted
@@ -15,8 +13,10 @@ export type TestNetworkContext = {
   NITRO_ADJUDICATOR_ADDRESS: Address;
 };
 
+
 export async function deploy(): Promise<TestNetworkContext> {
-  const {ethereumPrivateKey} = defaultTestConfig();
+  const ethereumPrivateKey = ETHERLIME_ACCOUNTS[0].privateKey;
+
   // TODO: best way to configure this?
   const deployer = new GanacheDeployer(8545, ethereumPrivateKey);
   const {
--- a/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
+++ b/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
@@ -13,28 +13,40 @@ import {getChannelResultFor, getPayloadFor} from '../__test__/test-helpers';
 
 // eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion
 const ethAssetHolderAddress = makeAddress(process.env.ETH_ASSET_HOLDER_ADDRESS!);
+// eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion
+if (!process.env.RPC_ENDPOINT) throw new Error('RPC_ENDPOINT must be defined');
+// eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion
+const rpcEndpoint = process.env.RPC_ENDPOINT;
+
 const config = {
   ...defaultTestConfig(),
   networkConfiguration: {
     ...defaultTestConfig().networkConfiguration,
     // eslint-disable-next-line no-process-env
-    rpcEndpoint: process.env.RPC_ENDPOINT,
-    // eslint-disable-next-line no-process-env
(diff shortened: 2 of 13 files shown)


packages/server-wallet/deployment/deploy.ts
```diff
@@ -13,9 +13,12 @@ export type TestNetworkContext = {
   NITRO_ADJUDICATOR_ADDRESS: Address;
 };
 
+// TODO: Where is this from?
+const ETHEREUM_PRIVATE_KEY = '0x7ab741b57e8d94dd7e1a29055646bafde7010f38a900f55bbd7647880faa6ee8';
```
kerzhner, line 17: https://github.com/statechannels/statechannels/blob/master/packages/devtools/src/constants.ts#L4

packages/server-wallet/jest/chain-setup.ts
```diff
@@ -6,21 +6,13 @@ import {deploy} from '../deployment/deploy';
...
   const accounts = ETHERLIME_ACCOUNTS.map(account => ({
     ...account,
     amount: utils.parseEther('100').toString(),
   }));
 
-  if (!process.env.GANACHE_PORT) {
-    throw new Error('process.env.GANACHE_PORT must be defined');
-  }
-  const ganacheServer = new GanacheServer(parseInt(process.env.GANACHE_PORT), 1337, accounts);
+  const ganacheServer = new GanacheServer(8545, 1337, accounts);
```
kerzhner, line 15: Have you tried changing these magic constants to ensure that this is the only place that relies these? Specifically, other places our code might rely on `process.env.GANACHE_PORT` and might default to port 8545. If that were the case, then changing 8545 to 8546 would result in failing tests.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -5,22 +5,23 @@ import {BigNumber, constants, ethers, providers} from 'ethers';
...
 import {defaultTestConfig} from '../config';
 import {ObjectiveSucceededValue, SingleChannelOutput, Wallet} from '../wallet';
 import {getChannelResultFor, getPayloadFor} from '../__test__/test-helpers';
 
-if (!defaultTestConfig.rpcEndpoint) throw new Error('rpc endpoint must be defined');
-const rpcEndpoint = defaultTestConfig.rpcEndpoint;
 let provider: providers.JsonRpcProvider;
-const b = Wallet.create({
+
+const GANACHE_RPC_ENDPOINT = 'http://localhost:8545';
```
kerzhner, line 15: This seems like a bad proliferation of a magic constant.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -26,9 +25,9 @@ const erc20AssetHolderAddress = makeAddress(process.env.ERC20_ASSET_HOLDER_ADDRE
 const erc20Address = makeAddress(process.env.ERC20_ADDRESS!);
 /* eslint-enable no-process-env, @typescript-eslint/no-non-null-assertion */
 
-if (!defaultTestConfig.rpcEndpoint) throw new Error('rpc endpoint must be defined');
-const rpcEndpoint = defaultTestConfig.rpcEndpoint;
-const provider: providers.JsonRpcProvider = new providers.JsonRpcProvider(rpcEndpoint);
+const GANACHE_RPC_ENDPOINT = 'http://localhost:8545';
```
kerzhner, line 28: This needs to match the host and port of test setup phase. We should use either use environment variables or some other methodology.

kerzhner (approved):
A chain service defined externally seems like the right pattern to adopt. It would be nice to clean up the magic constants before this work is merged in.

andrewgordstewart (changes requested):
I would like to temporarily block merging of this PR, not because I disagree with it (I agree with the change in theory)

The reason to block it is, this is actually a breaking change, not a refactor. Considering that, the managers will now need to be updated, so if
1. this gets into `master`
2. we want to fix some other issues, and release a new `server-wallet`
then we will be forced to also release this.

I'd like anyone (@snario ?) who has the time (I do not) to spend a few minutes thinking about if this poses an issue. If not, dismiss this review.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
snario: Right now on `master` the jest tests in the `server-wallet` that use a chain all hard code the usage of `http://localhost:8545` inside `chain-setup.ts`. I think this change just makes it much more obvious. I think we may want to open a separate issue if we want to refactor the tests to easily have configurable chain endpoints.

PR #2865, opened 2020-11-11
Add response builder

tomclose:
The `ResponseBuilder` is an attempt to clean up the output-generating logic that is currently spread throughout the wallet.

Each call to the wallet's api will start by creating a response object, that is then passed through to the objectives etc. The response object gives a simple interface for adding objectives/states/channelResults etc. to the output, and takes care of any message merging required:
```ts
// initialize a response

// add a signed state to be sent (will be added to outbox.signedStates)
response.stateSigned(state, myIndex, channelId);

// notify user of channel update (will be added to channelResults)
response.channelUpdated(channel)

// queue an objective to be sent to opponent (will be added to outbox.objectives)
response.objectiveCreated(objective)

// queue an objective to be approved by the app (will be added to objectivesToApprove)
response.objectiveReceived(objective)

// at the end you can return a single channel result ..
return response.singleChannelResult() // errors if channelUpdated has been called more than once

// ..or a multiple channel result
return response.multipleChannelResult()
```

This PR introduces the `WalletResponse` and `ResponseBuilder` interfaces. It converts the `syncChannels` api call to use the new interface. Future work will convert other api calls, if we're happy with the approach.

### DIFF
--- a/packages/server-wallet/package.json
+++ b/packages/server-wallet/package.json
@@ -81,6 +81,8 @@
     "db:rollback": "npx knex migrate:rollback --cwd src/db-admin",
     "db:seed": "npx knex seed:run --cwd src/db-admin",
     "db:setup:dev": "yarn db:drop && yarn db:create && yarn db:migrate",
+    "db:reset": "yarn db:drop; yarn db:create && yarn db:migrate",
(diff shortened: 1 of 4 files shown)


tomclose:
Merged as part of #2867

PR #2867, opened 2020-11-11
Wallet fully converted to use ResponseBuilder

tomclose:
Completes the work started in https://github.com/statechannels/statechannels/pull/2865.

### DIFF
--- a/packages/server-wallet/package.json
+++ b/packages/server-wallet/package.json
@@ -81,6 +81,8 @@
     "db:rollback": "npx knex migrate:rollback --cwd src/db-admin",
     "db:seed": "npx knex seed:run --cwd src/db-admin",
     "db:setup:dev": "yarn db:drop && yarn db:create && yarn db:migrate",
+    "db:reset": "yarn db:drop; yarn db:create && yarn db:migrate",
+    "db:reset:test": "(NODE_ENV=test yarn db:reset)",
     "generate-api": "yarn api-extractor run --local",
     "lint:check": "eslint \"*/**/*.{js,ts}\" --cache",
     "lint:write": "eslint \"*/**/*.{js,ts}\" --fix",
--- a/packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
@@ -3,7 +3,6 @@ import {Wallet} from '../..';
 import {createChannelArgs} from '../fixtures/create-channel';
 import {seedAlicesSigningWallet} from '../../../db/seeds/1_signing_wallet_seeds';
 import {defaultTestConfig} from '../../../config';
-import {alice, bob} from '../fixtures/participants';
 import {DBAdmin} from '../../../db-admin/db-admin';
 
 let w: Wallet;
@@ -38,7 +37,7 @@ describe('happy path', () => {
               signedStates: [{turnNum: 0, appData}],
               objectives: [
                 {
-                  participants: [alice(), bob()],
+                  participants: [], // todo: remove, currently deprecating participants
                   data: {
                     fundingStrategy: 'Direct',
                   },
--- a/packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
@@ -37,7 +37,7 @@ describe('happy path', () => {
               signedStates: [{turnNum: 0, appData}],
               objectives: [
                 {
-                  participants: [], // todo: remove, currently deprecating participants
+                  participants: [], // TODO: remove, currently deprecating participants
                   data: {
                     fundingStrategy: 'Direct',
                   },
--- a/packages/server-wallet/src/wallet/response-builder.ts
+++ b/packages/server-wallet/src/wallet/response-builder.ts
@@ -24,7 +24,7 @@ export class WalletResponse {
   /**
    * Queues channel for notification to user
    */
-  channelUpdated(channel: Channel): void {
+  queueChannel(channel: Channel): void {
     this._channelResults[channel.channelId] = channel.channelResult;
   }
 
@@ -33,14 +33,14 @@ export class WalletResponse {
    *
    * Plan to deprecate.
    */
-  channelUpdatedResult(channelResult: ChannelResult): void {
+  queueChannelResult(channelResult: ChannelResult): void {
     this._channelResults[channelResult.channelId] = channelResult;
   }
 
   /**
    * Queues state for sending to opponent
    */
-  stateSigned(state: SignedState, myIndex: number, channelId?: string): void {
+  queueState(state: SignedState, myIndex: number, channelId?: string): void {
     const myParticipantId = state.participants[myIndex].participantId;
     state.participants.forEach((p, i) => {
       if (i !== myIndex) {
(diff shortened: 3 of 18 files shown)


packages/server-wallet/package.json
```diff
@@ -81,6 +81,8 @@
     "db:rollback": "npx knex migrate:rollback --cwd src/db-admin",
     "db:seed": "npx knex seed:run --cwd src/db-admin",
     "db:setup:dev": "yarn db:drop && yarn db:create && yarn db:migrate",
+    "db:reset": "yarn db:drop; yarn db:create && yarn db:migrate",
+    "db:reset:test": "(NODE_ENV=test yarn db:reset)",
```
snario, line 85: Is there any reason why we do different things based on `NODE_ENV` being `test` or `development`?

andrewgordstewart: It's just a convenience script to set the shell variable

snario: Yeah but I'm just wondering why we need to even care about this at all — is there a fundamental reason why we need two environments

packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
```diff
@@ -38,7 +37,7 @@ describe('happy path', () => {
               signedStates: [{turnNum: 0, appData}],
               objectives: [
                 {
-                  participants: [alice(), bob()],
+                  participants: [], // todo: remove, currently deprecating participants
```
snario, line 40: Can we use all caps `TODO:`? I use [this extension](https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight) to highlight these things

packages/server-wallet/package.json
andrewgordstewart: There are different databases. The database used depends on the node env

snario: Yeah I know but I am wondering why we need to use different databases based on the node env

snario (commented):
Is there any particular reason why you took the approach of mutating the response as opposed to the usual builder pattern of creating a new object? i.e.,

```
response.addSomething();
```
vs
```
response = response.addSomething();
```

tomclose:
I took the first approach to make it easier to handle concurrency. If you return the `reponse` object you then have to have methods for merging responses if you ever do something like:
```ts
this.channelIds.map((id) => this._closeChannel(id, response))
```
It's possible that it's misleading to call it a builder and that it should instead just be `WalletResponse`.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -198,26 +195,28 @@ export class SingleThreadedWallet extends EventEmitter<EventEmitterType>
...
-        walletVersion: WALLET_VERSION,
-        signedStates: states,
-        requests: [{type: 'GetChannel', channelId}],
-      }),
-      channelResult: ChannelState.toChannelResult(channelState),
-    };
+    states.forEach(s => responseBuilder.stateSigned(s, myIndex, channelId));
+
+    responseBuilder.requestGetChannel(channelId, myIndex, participants);
+    responseBuilder.channelUpdatedResult(ChannelState.toChannelResult(channelState));
```
geoknee, line 219: have you forgotten to `return responseBuilder.singleChannelOutput()` ?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -236,37 +235,29 @@ export class SingleThreadedWallet extends EventEmitter<EventEmitterType>
...
-    amount,
-  }: HoldingUpdatedArg): Promise<SingleChannelOutput> {
+  private async _updateChannelFunding(
+    {channelId, token, amount}: UpdateChannelFundingParams,
+    response: ResponseBuilder
+  ): Promise<void> {
+    const assetHolderAddress = getAssetHolderAddress(token || Zero);
     await this.store.updateFunding(channelId, BN.from(amount), assetHolderAddress);
-    const {channelResults, outbox} = await this.takeActions([channelId]);
-    return {outbox, channelResult: channelResults[0]};
```
geoknee, line 269: Here again you changed the return type of the function: why?

packages/server-wallet/src/wallet/wallet.ts
tomclose: I think this is ok. `_syncChannel` modifies the `responseBuilder` that it was passed, and the caller will call either `response.singleChannelOutput()` or `response.multipleChannelOutput()` depending on whether it was `syncChannel` or `syncChannels`.

geoknee: nm, I see now that we are mutating the  `responseBuilder` that was passed in.

packages/server-wallet/src/wallet/response-builder.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Queues objectives for sending to opponent
+   */
+  objectiveCreated: (objective: DBObjective, myIndex: number, participants: Participant[]) => void;
+
+  /**
+   * Queues objectives for approval by the user
+   */
+  objectiveReceived: (objective: DBObjective) => void;
```
geoknee, line 38: I'm wondering if there are more intuitive names for these methods. For example `queueChannel`, `queueSignedState` etc.  You could even use function overloading and have a single `queue` method that you can pass various types into.

geoknee (approved):

packages/server-wallet/src/wallet/response-builder.ts
tomclose: The original intent was that the style would be as declarative as possible: the wallet would tell the response that something has happened, and the response would then decide what to do. That was the thinking behind having all the methods named as "X happened". I'm open to a change though. It sounds like the full set would become:

* `queueChannel`
* `queueChannelResult`
* `queueState`
* `queueCreatedObjective`
* `queueReceivedObjective`
* `queueCompletedObjective`
* `queueChannelRequest`

Do these sound reasonable?

geoknee: Yes that sounds reasonable. I don't feel super strongly about it, as the original intent you mentioned also seems reasonable -- and would probably make it a bit easier if we want to change the response format in the future. 

There is, however, a bit of a disconnect between that and the docstrings you put in the interface -- would be good to keep those in tighter sync.

tomclose: Made the change. I think I prefer it. d2583cf

geoknee: +1 for docstrings, by the way. I think we should be doing more of that.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -732,56 +670,47 @@ export class SingleThreadedWallet extends EventEmitter<EventEmitterType>
...
-    await this.takeActions([arg.channelId]);
+    await this.takeActions([arg.channelId], response);
+
+    response.channelUpdatedEvents().forEach(event => this.emit('channelUpdated', event.value));
   }
 
-  private registerChannelWithChainService(cr: ChannelResult): void {
-    const assetHolderAddresses = cr.allocations.map(a => getAssetHolderAddress(a.token));
-    this.chainService.registerChannel(cr.channelId, assetHolderAddresses, this);
+  private async registerChannelWithChainService(channelId: string): Promise<void> {
```
tomclose, line 708: This method does a query, where it didn't before.

PR #2868, opened 2020-11-11
Mark _transferAsset as virtual

lalexgap:
The current implementation of `GRTAssetHolder` expects to override `_transferAsset` so this marks it as `virtual` allowing for the `GRTAssetHolder` .

I don't think there's any security implications here but would be good to get a second opinion on that.


andrewgordstewart (approved):
I think it's fine, here's my rational: If someone wants to extend the ERC20AssetHolder, they need to be careful. Changing the implementation of `_transferAsset` is obviously "dangerous".

But adding `virtual` here doesn't pose any risk -- it just lets someone extend `ERC20AssetHolder`.

PR #2871, opened 2020-11-12
Close channel not on turn

tomclose:
Previously, calling `closeChannel` twice, or when not your turn, would cause a "not your turn" or "insert into objectives duplicate key error", as seen in https://github.com/statechannels/the-graph/issues/317.

This PR makes calling `closeChannel` idempotent, and enables `closeChannel` to be called when it is not your turn. When it isn't your turn, calling `closeChannel` will:

1. Add an approved close-channel objective to the store, which will cause the channel will be closed at the next available point.
2. Send this objective to your opponent, to prompt them to close the channel.


packages/server-wallet/src/__test__/create-and-close-channel/direct-funding.test.ts
```diff
@@ -152,25 +152,27 @@ it('Create a directly funded channel between two wallets ', async () => {
     status: 'running',
     turnNum: 3,
   });
-});
 
-it('Rejects b closing with `not your turn`', async () => {
```
tomclose, line 157: Combining these three tests into one, as they only work if run sequentially in order anyway.

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -53,15 +53,35 @@ function successfulWithdraw({app}: ProtocolState): boolean {
   return BN.eq(amountOwedToMe, amountTransferredToMe);
 }
 
+const isMyTurn = (ps: ProtocolState): boolean =>
+  !!ps.app.supported && (ps.app.supported.turnNum + 1) % 2 === ps.app.myIndex;
```
andrewgordstewart, line 57: Can we do something like this?
```suggestion
  !!ps.app.supported && (ps.app.supported.turnNum + ps.app.participants.length) % 2 === ps.app.myIndex;
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -398,7 +398,11 @@ export class Store {
         tx
       );
 
-    await ObjectiveModel.insert(objectiveToBeStored, tx);
+    try {
+      await ObjectiveModel.insert(objectiveToBeStored, tx);
+    } catch (e) {
+      if (e.name !== 'UniqueViolationError') throw e;
```
andrewgordstewart, line 404: You could also upsert, ie. `ON CONFLICT whatever_the_name_of_the_constraint_is DO NOTHING`, but it's difficult to do that with `knex`.

The current implementation is potentially dangerous: we might later have a _different_ uniqueness constraint that gets violated, and that error _should_ result in throwing `e`. IE. the test `e.name !== 'UniqueViolationError'` can have false negatives.

andrewgordstewart (approved):
LGTM

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -53,15 +53,35 @@ function successfulWithdraw({app}: ProtocolState): boolean {
...
+      isFinal: true,
+    })) ||
+    // otherwise, if it's my turn, sign a final state
+    (isMyTurn(ps) &&
+      signState({
+        channelId: ps.app.channelId,
+        ...ps.app.supported,
+        turnNum: ps.app.supported.turnNum + 1,
+        isFinal: true,
+      })));
```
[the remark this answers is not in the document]
tomclose: I'd like to make a change like this as part of a larger refactor of the protocols. With the change in this PR, I thought I should stick to the existing style as much as possible.

Issue #2872, opened 2020-11-12
Does the directly-funded-channel test work with Rinkeby contracts?


kerzhner:

andrewgordstewart:
@kerzhner This is done, right?

kerzhner:
> @kerzhner This is done, right?

This is the outcome of the testing work.

andrewgordstewart:
Ah, I was too quick reading the title, and thought in my head it said
" Test the chain service with Rinkeby contracts"

> > @kerzhner This is done, right?
> 
> This is the outcome of the testing work.

Does "The testing work" refer to something you're doing right now, or does it refer to testing on `sc-staging`?

kerzhner:
Updated the title of the issue to be clearer. One follow up: https://github.com/statechannels/the-graph/issues/334

PR #2874, opened 2020-11-13
fix: do not rely on revert reason in chain service

kerzhner:
With Ganache, when transaction execution reverts, the revert reason is included in the thrown exeption. This is not the case for testnets like rinkeby (https://github.com/ethers-io/ethers.js/issues/446).

The alternative to querying the nitro adjudicator for its state is to use the [`eth-revert-reason` npm module](https://www.npmjs.com/package/eth-revert-reason). However, that module is not reliable as we see the following error: `You cannot use a blocknumber that has not yet happened.`

`eth-revert-reason` can be used more reliably if a delay is added before `ether-revert-reason` is invoked.

FIxes https://github.com/statechannels/statechannels/issues/2872

### DIFF
--- a/packages/server-wallet/.eslintrc.js
+++ b/packages/server-wallet/.eslintrc.js
@@ -43,7 +43,10 @@ module.exports = {
         '@typescript-eslint/camelcase': 'off',
         '@typescript-eslint/no-explicit-any': 'off',
         '@typescript-eslint/no-use-before-define': 'off',
-        '@typescript-eslint/no-unused-vars': [1, {argsIgnorePattern: '^_'}],
+        '@typescript-eslint/no-unused-vars': [
+          1,
+          {argsIgnorePattern: '^_', varsIgnorePattern: '^_'},
+        ],
       },
     },
   ],
--- a/packages/server-wallet/src/chain-service/chain-service.ts
+++ b/packages/server-wallet/src/chain-service/chain-service.ts
@@ -213,15 +213,11 @@ export class ChainService implements ChainServiceInterface {
         })
       );
 
-      // Check if the channel has been finalized in the past or will finalize in the next 10 minutes
-      const timeNowSeconds = Date.now() / 1_000;
-      if (
-        finalizesAt &&
-        (timeNowSeconds - Number(finalizesAt) >= 0 ||
-          Number(finalizesAt) - timeNowSeconds <= 60 * 10)
-      ) {
-        return;
-      }
+      const latestBlockTimestamp = (
+        await this.provider.getBlock(await this.provider.getBlockNumber())
+      ).timestamp;
+      // Check if the channel has been finalized in the past
+      if (latestBlockTimestamp >= Number(finalizesAt)) return;
 
       throw reason;
     };
--- a/packages/server-wallet/src/chain-service/chain-service.ts
+++ b/packages/server-wallet/src/chain-service/chain-service.ts
@@ -2,6 +2,7 @@ import {
   ContractArtifacts,
   createERC20DepositTransaction,
   createETHDepositTransaction,
+  getChannelId,
   Transactions,
 } from '@statechannels/nitro-protocol';
 import {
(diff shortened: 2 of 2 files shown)


packages/server-wallet/.eslintrc.js
```diff
@@ -43,7 +43,10 @@ module.exports = {
         '@typescript-eslint/camelcase': 'off',
         '@typescript-eslint/no-explicit-any': 'off',
         '@typescript-eslint/no-use-before-define': 'off',
-        '@typescript-eslint/no-unused-vars': [1, {argsIgnorePattern: '^_'}],
+        '@typescript-eslint/no-unused-vars': [
+          1,
+          {argsIgnorePattern: '^_', varsIgnorePattern: '^_'},
```
andrewgordstewart, line 48: ohhhhhhhhhhhhhhhhhhhhhhhhhhhh

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -190,8 +197,33 @@ export class ChainService implements ChainServiceInterface {
...
+          participants: firstState.participants.map(p => p.signingAddress),
+        })
+      );
+
+      // Check if the channel has been finalized in the past or will finalize in the next 10 minutes
+      const timeNowSeconds = Date.now() / 1_000;
+      if (
+        finalizesAt &&
+        (timeNowSeconds - Number(finalizesAt) >= 0 ||
+          Number(finalizesAt) - timeNowSeconds <= 60 * 10)
```
andrewgordstewart, line 221: If the channel finalizes in the next 10 minutes, that indicates that there is a _challenge underway_. Do we really want to return in that case? Is that an expected behaviour in the happy path?

Is this instead due to the fact that `timeNowSeconds` is an approximation of `latestBlock.timestamp`?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -190,8 +197,33 @@ export class ChainService implements ChainServiceInterface {
...
+        _fingerprint,
+      ] = await this.nitroAdjudicator.getChannelStorage(
+        getChannelId({
+          ...firstState,
+          participants: firstState.participants.map(p => p.signingAddress),
+        })
+      );
+
+      // Check if the channel has been finalized in the past or will finalize in the next 10 minutes
+      const timeNowSeconds = Date.now() / 1_000;
```
andrewgordstewart, line 217: This should be `block.timestamp` shouldn't it? Who's to say whether `Date.now()` is aligned with the important timestamp?

andrewgordstewart (changes requested):
There's some question about timestamps

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: This case represents the following:
1. We have submitted a conclude and transfer all transaction and that transaction has failed
1. We have queried the chain and found out that the channel storage has a `finalizesAt` timestamp.

We need a heuristic to account for skew between timestamps onchain and timestamps offchain. In this PR the heuristic is to allow for skew of 10 minutes. Perhaps this is too large of an allowance?

If we want to consider challenges in this logic, we probably have to compare the fingerprint to the latest state in the wallet. This is not trivial as the chain service does not store states.

Thoughts?

kerzhner: Actually, I see you comment below about not using a local timestamp. So I think we might be able to remove this skew allowance.

andrewgordstewart: There may still be some issues due to chain reorgs. Something like
1. I submit tx1.
2. It is reverted in block 10 because a transaction tx2 succeeded in block 9.
3. The chain is forked at block 9 and no longer includes tx1 or tx2.
4. I get the channel state from block 11 in this fork.

In this specific case, we would want to observe that the channel was not concluded and retry, so I think the overall logic in this pr is still correct. But I'm not sure I have a full grasp on the possible edge cases.

andrewgordstewart: I guess, to clarify, I think the approach in this pr is sufficient for now, and probably a good approach, but it should just use the block timestamp rather than the system's timestamp.

kerzhner: 9d1de309a.

Note that this PR does not attempt to retry transactions. The chain service throws an error if a transaction fails for any reason besides acceptable failure reasons. Transaction retries are something that would be nice to have down the line.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Yep, this is a great point 9d1de30

andrewgordstewart (approved):

PR #2882, opened 2020-11-14
test: fix flickering test

andrewgordstewart:
fixes these types of test failures:
https://app.circleci.com/pipelines/github/statechannels/statechannels/10374/workflows/4b6029dc-d67b-41eb-bfb8-f1987dc0cf8d/jobs/51093


kerzhner (approved):
Do we understand the underlying cause for why the error occurs some of the time?

andrewgordstewart:
> Do we understand the underlying cause for why the error occurs some of the time?

No, but I don't think it matters.

PR #2883, opened 2020-11-14
chore: try upgrading solidity to 0.7.4?

andrewgordstewart:


geoknee:
blockers:
- we need a new image on circle with upgraded solidity (easy, and done)
- `ethereum-waffle/compile@latest` only installs `solc@0.6`. It throws errors if I try and force it to use `0.7.x` via yarn resolutions. We get the errors on _importing_ `solc`: so setting `USE_NATIVE_SOLC=true` is not a successful workaround (and would require all devs to install native solc anyway).

see #2895 for an alternative approach, which leans on the incoming `hardhat`package.

andrewgordstewart:
Should we close in favour of https://github.com/statechannels/statechannels/pull/2895 then?

geoknee:
yeah

PR #2889, opened 2020-11-14
chore: add sync and async timer for performance metrics

snario:
Addresses a TODO

### DIFF
--- a/packages/server-wallet/src/utilities/validate-transition.ts
+++ b/packages/server-wallet/src/utilities/validate-transition.ts
@@ -25,12 +25,12 @@ export function shouldValidateTransition(incomingState: StateWithHash, channel: 
 }
 // Port of the following solidity code
 // https://github.com/statechannels/statechannels/blob/a3d21827e340c0cc086f1abad7685345885bf245/packages/nitro-protocol/contracts/ForceMove.sol#L492-L534
-export async function validateTransition(
+export function validateTransition(
   fromState: SignedState,
   toState: SignedState,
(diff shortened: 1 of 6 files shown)


packages/server-wallet/src/utilities/validate-transition.ts
```diff
@@ -25,12 +25,12 @@ export function shouldValidateTransition(incomingState: StateWithHash, channel:
 }
 // Port of the following solidity code
 // https://github.com/statechannels/statechannels/blob/a3d21827e340c0cc086f1abad7685345885bf245/packages/nitro-protocol/contracts/ForceMove.sol#L492-L534
-export async function validateTransition(
```
geoknee, line 28: Why was this ever async?

geoknee (approved):

packages/server-wallet/src/utilities/validate-transition.ts
snario: The original implementation of `validateTransitionWithEVM` hit the DB https://github.com/statechannels/statechannels/commit/c683bceda1ae4a2a7198b69fc75045a88b50e42a#diff-897a8f5bbf39f2c61f4fcfa823b514bfb6208c27196d096352ea5e92a4a2a19bR14

snario: I removed it but didn't mark it as synchronous https://github.com/statechannels/statechannels/commit/6385b690ab906b043fcf89f88c30e39fa689fc44

PR #2891, opened 2020-11-15
build: remove netlify CLI from node_modules and related scripts

snario:
We now only have a single package which uses Netlify, `docs-website`, and since our usage of Netlify builds is dramatically lower right now (since we aren't changing code for `xstate-wallet`, `web3torrent`, etc on each commit) we can probably ask Netlify to build this stuff for us on their servers instead of on CircleCI.

Also, the `netlify-cli` package is _huge_, by far our largest dependency.


snario:
Will want to update Netlify config

image

image

https://app.netlify.com/sites/nitro-protocol/settings/deploys

geoknee (approved):
I think this makes sense, especially given #2901 
What we want in the short term is for the docs website to stay frozen until we can devote some time to get it automatically updating / editable again.

PR #2878, opened 2020-11-14
build: add hardhat for testnet deploys

kerzhner:
[hardhat](https://www.npmjs.com/package/hardhat-deploy) provides a nice set of tools for managing deployments across multiple networks. The downside of this work is that there are now 2 compilation and deploy strategies.
1. For Ganache: we have a custom strategy that uses etherlime.
2. For all other networks, we have hardhat.

In the future, it would be nice to migrate all compilation and deployment to hardhat: https://github.com/statechannels/statechannels/issues/2877

Fixes https://github.com/statechannels/statechannels/issues/2875

PR #2893, opened 2020-11-16
[nitro-protocol] Switch contract compilation to hardhat

geoknee:
~⚠️ base is not master, see #2878~

TODO
- [ ] unbreak docs-website (hardhat is not emitting the `ast` into the artifacts) :-( -- __for now, just freeze docs website__
- [x] figure out why unit tests are failing on ci while they pass locally for me


geoknee:
Because hardhat handles native solc compilation and fallback to solcjs, if we go this way we can deprecate the USE_NATIVE_SOLC environment variable, and also stop worrying about pushing new docker images with solc installed at a specific version. 

We may be able to save a bit of time on circle if we cache the binary that hardhat downloads to do the compilation.

packages/docs-website/package.json
```diff
@@ -34,12 +34,12 @@
...
   "scripts": {
     "api-documenter": "node scripts/api-documenter.js",
-    "build:netlify": "yarn trigger-api-generation && yarn api-documenter && yarn docusaurus build",
+    "build:netlify:disabled-for-now": "yarn trigger-api-generation && yarn api-documenter && yarn docusaurus build",
     "deploy": "docusaurus deploy",
     "docusaurus": "docusaurus",
     "examples": "docusaurus-examples",
     "publish-gh-pages": "docusaurus-publish",
-    "release:netlify": "netlify deploy --site $NITRO_PROTOCOL_NETLIFY_ID --dir=build",
+    "release:netlify:disabled-for-now": "netlify deploy --site $NITRO_PROTOCOL_NETLIFY_ID --dir=build",
```
geoknee, line 42: We should make a new issue to restore the docs website to its former glory.

.circleci/config.yml
```diff
@@ -6,13 +6,12 @@ orbs:
 
 defaults: &defaults
   environment:
-    USE_NATIVE_SOLC: true
     PSQL_ARGS: -p 5432 -h localhost
     NODE_ENV: test
   working_directory: /home/circleci/project
   resource_class: medium
   docker:
-    - image: geoknee/statechannels:12.18-solc0.6.12 # Fast contract compilation with solc installed
```
kerzhner, line 15: Would you mind creating an issue for using a lighter-weight docker image without `solc`?

.circleci/config.yml
```diff
@@ -58,12 +57,12 @@ save_built_artifacts: &save_built_artifacts
       - packages/wire-format/lib
       - packages/server-wallet/lib
       # Smart contract compiles
-      - packages/nitro-protocol/build
+      - packages/nitro-protocol/artifacts
 
-restore_built_artifacts: &restore_built_artifacts
+restore_built_artifacts: &restore_built_artifacts # only used by release-hub-production
```
kerzhner, line 62: Are we still using `release-hub-production`?

packages/nitro-protocol/hardhat.config.ts
```diff
@@ -5,36 +5,49 @@
...
   paths: {
     sources: 'contracts',
     deploy: 'hardhat-deploy',
     deployments: 'hardhat-deployments',
   },
 };
+
+const infuraToken = process.env.INFURA_TOKEN;
+const rinkebyDeployerPK = process.env.RINKEBY_DEPLOYER_PK;
+if (infuraToken && rinkebyDeployerPK) {
```
kerzhner, line 39: What happens if `yarn hardhat deploy --network rinkeby` is run without these env vars defined? An error, or a silent failure?

packages/nitro-protocol/package.json
```diff
@@ -70,13 +69,13 @@
   "scripts": {
     "build:typescript": "rm -rf lib; yarn tsc -b .",
     "build:webpack": "webpack",
-    "contract:compile": "yarn node ./bin/compile.js",
+    "contract:compile": "yarn hardhat compile --force",
```
kerzhner, line 72: do you think that the `force` flag is necessary?

kerzhner (approved):
🎩

packages/nitro-protocol/package.json
geoknee: My instinct is that it is worth waiting a few seconds to get a bit more certainty that the compilation artifacts are up to date, especially as these are form part of the npm package. 

If developing with the contracts, one has the option to just run `yarn hardhat compile` and get a bit of a faster flow going. (On my machine it takes 2s to decide "nothing to compile" and 7s to compile everything.)

There's also the option to watch for file changes, and then recompile and/or redeploy as necessary https://hardhat.org/plugins/hardhat-watcher.html

packages/nitro-protocol/hardhat.config.ts
geoknee: Good point. In 39116a1 I have restored the previous behaviour by making the current config file extend off of a leaner base config.

PR #2895, opened 2020-11-16
[nitro-protocol] Upgrade solidity to 0.7.4

### LINKED ISSUE #2870: Bump solidity version to 0.7

geoknee:
~⚠️ base is not master, see #2893~

Closes #2870

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ defaults: &defaults
   working_directory: /home/circleci/project
   resource_class: medium
   docker:
-    - image: geoknee/statechannels:12.18-solc0.6.12 # solc installed (unecessary since hardhat will re-download solc)
+    - image: geoknee/statechannels:12.18-solc0.7.4 # solc installed (unecessary since hardhat will re-download solc)
     - image: circleci/postgres:12-alpine-ram # server-wallet uses a postgresql DB. alpine-ram is a minimal installation that runs in memory (ie. faster tests!)
       environment:
         POSTGRES_USER: root
--- a/packages/devtools/package.json
+++ b/packages/devtools/package.json
@@ -20,7 +20,7 @@
     "lockfile": "1.0.4",
     "pino": "6.2.0",
     "pino-pretty": "4.0.0",
-    "solc": "0.6.12",
+    "solc": "0.7.4",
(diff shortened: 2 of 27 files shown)


packages/devtools/package.json
```diff
@@ -20,7 +20,7 @@
     "lockfile": "1.0.4",
     "pino": "6.2.0",
     "pino-pretty": "4.0.0",
-    "solc": "0.6.12",
+    "solc": "0.7.4",
```
geoknee, line 23: I suspect we no longer need to install `solc` (anywhere?)

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ defaults: &defaults
   working_directory: /home/circleci/project
   resource_class: medium
   docker:
-    - image: geoknee/statechannels:12.18-solc0.6.12 # solc installed (unecessary since hardhat will re-download solc)
+    - image: geoknee/statechannels:12.18-solc0.7.4 # solc installed (unecessary since hardhat will re-download solc)
```
geoknee, line 14: This is harmless but probably unnecessary given our use of hardhat for compilation.

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -84,7 +84,7 @@ contract ForceMove is IForceMove {
         emit ChallengeRegistered(
             channelId,
             largestTurnNum,
-            uint48(now) + fixedPart.challengeDuration, // This could overflow, so don't join a channel with a huge challengeDuration
+            uint48(block.timestamp) + fixedPart.challengeDuration, // This could overflow, so don't join a channel with a huge challengeDuration
```
snario, line 87: What's the max `challengeDuration`? We could add a check to the wallet

snario (approved):

packages/nitro-protocol/contracts/ForceMove.sol
geoknee: I think (2**48 − 1 - now) but I would need to think a bit more to be sure. But I think this is probably the safe lower bound on the maximum.

packages/devtools/package.json
geoknee: #2902

PR #2898, opened 2020-11-17
server-wallet unit tests work with Rinkeby

kerzhner:
This PR is a set of changes that helped run the `server-wallet` unit tests against Rinkeby. I am reluctant to update the readme with instructions on how to run the unit tests against Rinkeby as:
- My sense is that running unit tests against Rinkeby won't happen frequently.
- The tests do not work out of the box necessarily. At the very least, jest timeouts must be increased as Rinkeby transactions are slower than Ganache. Also, tests that rely on fresh chain state (for example, a channel with no funds on chain) might fail.


packages/nitro-protocol/hardhat-deployments/rinkeby/solcInputs/421853c27fa211830d511ef4302cc5f0.json
```diff
@@ -0,0 +1,50 @@
...
+        "": [
+          "ast"
+        ]
+      }
+    },
+    "metadata": {
+      "useLiteralContent": true
+    }
+  }
+}
```
snario, line 50: Is this file meant to be ignored?

.env.rinkeby
```diff
@@ -0,0 +1,10 @@
+SERVER_DB_NAME=server_wallet_test
+SERVER_DB_USER=postgres
+
+CHAIN_NETWORK_ID=4
+RPC_ENDPOINT=https://rinkeby.infura.io/v3/${INFURA_API_KEY}
+
+NITRO_ADJUDICATOR_ADDRESS=0x1d25Bbf952781e0c75d248904B5422FeE9FcA558
+ETH_ASSET_HOLDER_ADDRESS=0x0d9F0670170Fab72F0f879D020f9E77590aEE642
+ERC20_ASSET_HOLDER_ADDRESS=0x59be89a0DE5cA11c74463A696128CcD80ab0182A
+ERC20_ADDRESS=0x78573871B098da14Dc33f8c155b2ddf382eABfe3
```
geoknee, line 10: Do we have to remember to keep these addresses in sync going forward? Is there a way where we can export them from `nitro-protocol/...../deployments` ?

kerzhner: Yes, we would have to remember to update these. Perhaps it is not worth investing in exporting these from `nitro-protocol` given that:
1. At the moment, this env file is used for just unit tests.
1. We probably won't run unit tests against Rinkeby that often.

However, if either of the above does not hold, we may want to reconsider.

packages/nitro-protocol/hardhat-deployments/rinkeby/solcInputs/421853c27fa211830d511ef4302cc5f0.json
kerzhner: I think we can probably ignore the `solcInputs` directory. f6dfcd5be

geoknee (approved):

PR #2904, opened 2020-11-17
refactor: remove all 'token' arguments and replace with 'assetHolderAddress'

snario:
The wallet does not need to be responsible for doing any lookups of token addresses to specific asset holder addresses.

The app can be in charge of choosing the asset holder address.

If the app is not in charge, then some other component must be responsible for populating the token -> asset holder address mapping as certainly this cannot be a centralized list (i.e., in the server wallet source code). So, there would need to be some kind of `setTokenToAssetHolder(token: string, assetHolder: Address)` function that the app would call.... which defeats the point I think.


packages/client-api-schema/client-api-schema.api.md
```diff
@@ -270,7 +270,7 @@ export type ErrorCodes = {
     CreateChannel: {
         SigningAddressNotFound: 1000;
         InvalidAppDefinition: 1001;
-        UnsupportedToken: 1002;
+        UnsupportedAssetHolderAddress: 1002;
```
snario, line 273: Realistically we can just remove this error type

packages/wire-format/src/types.ts
```diff
@@ -36,7 +36,7 @@ export interface AllocationItem {
 }
 
 export interface Allocation {
-  assetHolderAddress: Address; // The token's contract address.
```
geoknee, line 39: 👍

geoknee (approved):
Another reason to make this change: The current pattern presupposes at most one asset holder per token.

PR #2907, opened 2020-11-17
build: do not use npm-run-all

geoknee:
I'm hoping this will iron out issues we have had with our GH actions, where lerna will run inside of itself

```
lerna info Executing command in 7 packages: "yarn run prepare"

...

@statechannels/nitro-protocol: lerna info Executing command in 3 packages: "yarn run build:typescript"
```


lalexgap (approved):

geoknee:
https://github.com/lerna/lerna/issues/2145

PR #2913, opened 2020-11-18
test: increase number of tokens minted in Token.sol contract

snario:
In an end to end stress test with The Graph we require several billion tokens and absent the asset holder I'm using this test artifact, but I'm limited to 10,000 tokens.


andrewgordstewart (approved):
Why not make the amount minted a constructor variable?

geoknee (approved):

PR #2914, opened 2020-11-18
[nitro-protocol] Check chainId on chain

geoknee:


andrewgordstewart (approved):
We should document the security offered by `chainId`/`channelId` somewhere.

geoknee:
> We should document the security offered by `chainId`/`channelId` somewhere.

Here's one place we can do that #2915 .

Issue #2901, opened 2020-11-17
Restore the docs website to its former glory.


geoknee:
We should make a new issue to restore the docs website to its former glory.

_Originally posted by @geoknee in https://github.com/statechannels/statechannels/pull/2893#discussion_r524503146_

geoknee:
via #3029

PR #2915, opened 2020-11-18
docs: add precautions section

geoknee:
I'd like to invite feedback on the wording here. 

For chainId I have assumed #2914 will be merged.


snario:
Might be cleaner / easier to consume in a table?

snario:
Also in general feels like it implies there should be some sort of "how to build a wallet" section

snario (approved):

geoknee:
Reminder: the docs website is currently frozen #2901 so this change (and any others) won't go live until we fix that.

Issue #2873, opened 2020-11-12
Chain service should provide some minimal protection against chain reorgs


kerzhner:
From the audit team:
There are two schelling points within production services (e.g., exchanges and the like) on 6 and 25 blocks. I think given the nature of state channels, that 6 blocks are OK

kerzhner:
Reference the following section in the audit: _6.6 server-wallet - The wallet does not account for chain reorgs by waiting for confirmations to events received_

PR #2922, opened 2020-11-19
Wait for 6 blocks before emitting a chain event

kerzhner:
Fixes https://github.com/statechannels/statechannels/issues/2873.

### DIFF
--- a/packages/devtools/src/ganache/server.ts
+++ b/packages/devtools/src/ganache/server.ts
@@ -130,8 +130,7 @@ export class GanacheServer {
     accounts: Account[] = ETHERLIME_ACCOUNTS,
     public readonly timeout: number = 10_000,
     gasLimit = 1_000_000_000,
-    gasPrice = 1,
-    blockTime: number | undefined = undefined
+    gasPrice = 1
   ) {
     logger.info(`Starting ganache on port ${this.port} with network ID ${this.chainId}`);
     this.fundedPrivateKey = accounts[0].privateKey;
@@ -152,8 +151,7 @@ export class GanacheServer {
             ? extractLogsFromVerboseGanacheOutput(this.buffer, x)
             : logger.info(x);
         }
-      },
-      blockTime
+      }
     };
 
     this.server = ganache.server({...serverOptions, _chainId: this.chainId} as any);
--- a/packages/devtools/src/ganache/server.ts
+++ b/packages/devtools/src/ganache/server.ts
@@ -129,7 +129,7 @@ export class GanacheServer {
     public readonly chainId: number = 9001,
     accounts: Account[] = ETHERLIME_ACCOUNTS,
     public readonly timeout: number = 10_000,
-    gasLimit = 10_00_000_000,
+    gasLimit = 1_000_000_000,
     gasPrice = 1,
     blockTime: number | undefined = undefined
   ) {
--- a/packages/devtools/src/ganache/server.ts
+++ b/packages/devtools/src/ganache/server.ts
@@ -129,8 +129,9 @@ export class GanacheServer {
     public readonly chainId: number = 9001,
     accounts: Account[] = ETHERLIME_ACCOUNTS,
     public readonly timeout: number = 10_000,
-    gasLimit = 1000000000,
-    gasPrice = 1
+    gasLimit = 10_00_000_000,
+    gasPrice = 1,
+    blockTime: number | undefined = undefined
   ) {
     logger.info(`Starting ganache on port ${this.port} with network ID ${this.chainId}`);
     this.fundedPrivateKey = accounts[0].privateKey;
@@ -151,7 +152,8 @@ export class GanacheServer {
             ? extractLogsFromVerboseGanacheOutput(this.buffer, x)
             : logger.info(x);
         }
-      }
+      },
+      blockTime
     };
 
     this.server = ganache.server({...serverOptions, _chainId: this.chainId} as any);
--- a/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
+++ b/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
@@ -191,4 +191,4 @@ it('Create a directly funded channel between two wallets ', async () => {
 
   expect(BN.sub(aBalanceFinal, aBalanceInit)).toEqual(aFunding);
   expect(BN.sub(bBalanceFinal, bBalanceInit)).toEqual(bFunding);
-}, 20_000);
+}, 50_000);
--- a/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
+++ b/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
@@ -198,7 +198,7 @@ it('Create a directly funded channel between two wallets ', async () => {
   expect(BN.sub(aBalanceFinal, aBalanceInit)).toEqual(aFunding);
   expect(BN.sub(bBalanceFinal, bBalanceInit)).toEqual(bFunding);
 
-  // todo: remove this
+  // TODO: remove this
   // The reason for the wait:
   // - B CloseChannel objective succeeds BEFORE AssetTransferred event arrives
   // - B has no funds in the channel, so B does not wait for an AssetTransferred event to complete the CloseObjective
--- a/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
+++ b/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
@@ -1,13 +1,18 @@
 import {CreateChannelParams, Participant, Allocation} from '@statechannels/client-api-schema';
 import {ETHERLIME_ACCOUNTS} from '@statechannels/devtools';
+import {ContractArtifacts} from '@statechannels/nitro-protocol';
 import {BN, makeAddress, makeDestination} from '@statechannels/wallet-core';
-import {BigNumber, constants, ethers, providers} from 'ethers';
+import {BigNumber, constants, Contract, ethers, providers} from 'ethers';
+import _ from 'lodash';
 import {fromEvent} from 'rxjs';
 import {take} from 'rxjs/operators';
 
 import {defaultTestConfig} from '../config';
 import {ObjectiveSucceededValue, SingleChannelOutput, Wallet} from '../wallet';
 import {getChannelResultFor, getPayloadFor} from '../__test__/test-helpers';
+
+// eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion
+const ethAssetHolderAddress = makeAddress(process.env.ETH_ASSET_HOLDER_ADDRESS!);
 
 if (!defaultTestConfig.rpcEndpoint) throw new Error('rpc endpoint must be defined');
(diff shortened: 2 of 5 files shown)


kerzhner:
Tests are passing, but seeing these logs:
```
@statechannels/server-wallet:     console.error
@statechannels/server-wallet:       Unhandled error
@statechannels/server-wallet:       at process.uncaught (../../node_modules/jest-jasmine2/build/jasmine/Env.js:248:21)
@statechannels/server-wallet:       at processEmit (../../node_modules/signal-exit/index.js:161:32)
@statechannels/server-wallet:       at process.emit (../../node_modules/ganache-core/node_modules/source-map-support/source-map-support.js:465:21)
@statechannels/server-wallet:     console.error
@statechannels/server-wallet:       Error: Unable to acquire a connection
@statechannels/server-wallet:           at Client_PG.acquireConnection (/home/circleci/project/node_modules/knex/lib/client.js:339:13)
@statechannels/server-wallet:           at Transaction.acquireConnection (/home/circleci/project/node_modules/knex/lib/transaction.js:213:46)
@statechannels/server-wallet:           at Transaction._evaluateContainer (/home/circleci/project/node_modules/knex/lib/transaction.js:158:17)
@statechannels/server-wallet:           at /home/circleci/project/node_modules/knex/lib/transaction.js:60:12
@statechannels/server-wallet:       at process.uncaught (../../node_modules/jest-jasmine2/build/jasmine/Env.js:249:21)
@statechannels/server-wallet:       at processEmit (../../node_modules/signal-exit/index.js:161:32)
@statechannels/server-wallet:       at process.emit (../../node_modules/ganache-core/node_modules/source-map-support/source-map-support.js:465:21)
```

kerzhner:
Error logs are dealt with in bca18aae2

packages/devtools/src/ganache/server.ts
```diff
@@ -129,8 +129,9 @@ export class GanacheServer {
     public readonly chainId: number = 9001,
     accounts: Account[] = ETHERLIME_ACCOUNTS,
     public readonly timeout: number = 10_000,
-    gasLimit = 1000000000,
-    gasPrice = 1
+    gasLimit = 10_00_000_000,
```
snario, line 132:
```suggestion
    gasLimit = 1_000_000_000,
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -291,25 +294,48 @@ export class ChainService implements ChainServiceInterface {
     subscriptions.map(s => s.unsubscribe());
   }
 
+  private async waitForConfirmations(event: Event | null): Promise<void> {
+    const numConfirmations = 6;
```
snario, line 298: This should probably be a configurable value in the constructor of the `ChainService`. Tests with ganache might want `1`, Rinkeby maybe `3`, Mainnet maybe `6`.

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -291,25 +294,48 @@ export class ChainService implements ChainServiceInterface {
     subscriptions.map(s => s.unsubscribe());
   }
 
+  private async waitForConfirmations(event: Event | null): Promise<void> {
```
snario, line 297: Seems like a kind of strange API to me. When is `event` going to be `null`?

snario: We probably also can just pass in a transaction hash object instead of the entire ethers `Event`

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -183,12 +183,27 @@ it('Create a directly funded channel between two wallets ', async () => {
...
+  });
 
   const aBalanceFinal = await getBalance(aAddress);
   const bBalanceFinal = await getBalance(bAddress);
 
   expect(BN.sub(aBalanceFinal, aBalanceInit)).toEqual(aFunding);
   expect(BN.sub(bBalanceFinal, bBalanceInit)).toEqual(bFunding);
-}, 20_000);
+
+  // todo: remove this
```
snario, line 201: Can we make this all caps `TODO:`? I use [this extension](https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight) which highlights them

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: We can certainly make this configurable!

I would actually argue that we should use the same value for Ganache and Rinkeby as Mainnet:
- Using a longer confirmation time might have race conditions. It would be bad to wait to discover these on Mainnet.
- Seems like it is in general a good practice to configure test environments as close as possible to production environments.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
kerzhner, line 201:
```suggestion
  // TODO: remove this
```

packages/server-wallet/src/chain-service/chain-service.ts
snario: AFAIK
- `ganache` does not mine blocks by default
- if it does mine blocks, it mines them one at a time at an interval set by some config
- there are no uncle blocks

Does our test environment have the autominer turned on?

geoknee: From the `ganache-core` readme: 

> "blockTime": number - Specify blockTime in seconds for automatic mining. If you don't specify this flag, ganache will instantly mine a new block for every transaction. Using the blockTime option is discouraged unless you have tests which require a specific mining interval.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Good call on limiting the parameter to be a transaction.

When a wallet registers a channel, the chain service reads the funding amount from the chain and calls `holdingUpdated` with the value. This read is **not** associated with a transaction. However, we still want to protect against chain reorgs for this read.

The goal with emitting a funding value after channel registration is to allow wallets that start after some downtime to have a correct funding value in the database.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: https://github.com/statechannels/statechannels/blob/4ab70e25e0653d41603148e1dd3b3a579d4b4799/packages/server-wallet/jest/chain-setup.ts#L37

andrewgordstewart: You can also instruct ganache to mine a block.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: On second thought, I think it might be cleaner to stick with passing an ethers event into this function. If this functionality needs to be reused, we can do more refactoring/rethinking of function signatures.

Also, I cleaned up the types a bit de5085582

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: I like that idea! a25bdf8

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -183,12 +205,27 @@ it('Create a directly funded channel between two wallets ', async () => {
...
+  });
 
   const aBalanceFinal = await getBalance(aAddress);
   const bBalanceFinal = await getBalance(bAddress);
 
   expect(BN.sub(aBalanceFinal, aBalanceInit)).toEqual(aFunding);
   expect(BN.sub(bBalanceFinal, bBalanceInit)).toEqual(bFunding);
-}, 20_000);
+
+  // TODO: remove this
```
andrewgordstewart, line 223: Let's file an issue for this?

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -183,12 +205,27 @@ it('Create a directly funded channel between two wallets ', async () => {
     turnNum: 4,
   });
 
-  expect(await closeCompletedA).toMatchObject({channelId, objectiveType: 'CloseChannel'});
-  expect(await closeCompletedB).toMatchObject({channelId, objectiveType: 'CloseChannel'});
+  expect(await closeCompletedA).toMatchObject({
```
andrewgordstewart, line 208: I _think_ this logic is only testing the happy path. There doesn't appear to be coverage in this test file that says:
- This promise does _not_ resolve to a `'CloseChannel'` event after 3 block confirmations

I think you could do something like the following
```
async function rejectIfResolves(promiseThatShouldNotResolve: Promise<any>, duration = 1_000) {
  return new Promise((resolve, reject) => {
    promiseThatShouldNotResolve.then(reject);
    setTimeout(resolve, duration);
  });
}

ganache.mineBlocks(2);
await rejectIfResolves(closeCompletedA);
ganache.mineBlocks(3);
await expect(closeCompletedA).resolves.toMatchObject(snip);
```
to assert that, though this is a little hacky.

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -291,25 +288,45 @@ export class ChainService implements ChainServiceInterface {
...
+      type: Deposited,
+      channelId,
+      assetHolderAddress: makeAddress(contract.address),
+      amount: BN.from(holding),
+    };
+  }
+
+  private async waitForConfirmations(event: Event | undefined): Promise<void> {
+    if (event) {
+      await (await event.getTransaction()).wait(blockConfirmations + 1);
```
andrewgordstewart, line 304: Can you explain this logic in a comment to highlight what's going on?

It seems like the core change in this PR, and while it is possible to infer what's going on, it might not be obvious. 

Something like
```suggestion
      // `tx.wait(n)` resolves after n blocks are mined that include the given transaction `tx`
      // See https://docs.ethers.io/v5/api/providers/types/#providers-TransactionResponse
      await (await event.getTransaction()).wait(blockConfirmations + 1);
```

andrewgordstewart (approved):
🔒

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
kerzhner: Should be covered by https://github.com/statechannels/statechannels/issues/2928. Updated the description of that issue.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
kerzhner: You are right that the non-happy path is not tested. This test would be quite a bit easier to write once https://github.com/statechannels/statechannels/issues/2928 is resolved. In that case, we can simply:
1. Read channel status. Channel status should be `closed`.
1. Mine 3 blocks. Channel status should still be `closed`.
1. Mine 2 more blocks. Channel status should be `closedAndDefunded`.

Thoughts?

andrewgordstewart: Added some TODOs to #2928

PR #2924, opened 2020-11-19
feat: catch external destination collisions

geoknee:
when computing channelId
add a test for helper isExternalDestination

Note that this is extremely unlikely to happen in the first place.

### DIFF
--- a/packages/nitro-protocol/src/contract/channel.ts
+++ b/packages/nitro-protocol/src/contract/channel.ts
@@ -8,12 +8,19 @@ export interface Channel {
   chainId: Uint256;
 }
 
+export function isExternalDestination(bytes32: Bytes32): boolean {
+  return /^0x(0{24})([a-fA-F0-9]{40})$/.test(bytes32);
+}
+
 export function getChannelId(channel: Channel): Bytes32 {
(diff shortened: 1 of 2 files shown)


geoknee:
Additional thoughts: 

- we could have a `ChannelId` type that excludes external destinations, and validate against this in our schema validators
-  the check that this PR introduces will be triggered in our server wallet, via the `calculateChannelId` function in `wallet-core`:

https://github.com/statechannels/statechannels/blob/be0c73c76458fa9674411ab39fb3b43ca6064ec5/packages/server-wallet/src/models/channel.ts#L182-L184

So we are protected against a malign actor proposing a channel with an external id -- those states won't get into the DB.

packages/nitro-protocol/src/contract/channel.ts
```diff
@@ -8,12 +8,19 @@ export interface Channel {
   chainId: Uint256;
 }
 
+export function isExternalDestination(bytes32: Bytes32): boolean {
```
snario, line 11: I think you can do something like `: bytes32 is Destination {` to return a now-typed `Destination` object

snario (approved):

PR #2925, opened 2020-11-19
fix: require addresses returned from ecrecover to be nonzero

geoknee:


packages/docs-website/docs/protocol-tutorial/execute-state-transitions.md
```diff
@@ -106,7 +106,7 @@ const getVariablePart = getVariablePart(state);
 As a state channel participant, it is advised to check the `FixedPart` of any channel before participating in it. A good state channels wallet will perform these checks for you:
 
 - `chainId` -- This needs to match the id of the chain where assets are to be locked. The root of the funding tree for this channel.
-- `participants` -- This should have length at least 2, and include a public key (account) that you control.
+- `participants` -- This should have length at least 2, and include a public key (account) that you control. Each entry should be a nonzero ethereum address.
```
geoknee, line 109: If it is zero, what should happen?

a) [current] an invalid signature may be used to change adjudicator storage for this channel on chain (since `ecrecover` will default to 0 in these cases)
b) [proposed] it would be impossible to finalize the channel

I'm not sure that b is better than a? If 

In either case it's a very important check to make in the wallet.

geoknee: Looks like our native signing functions are going to catch invalid addresses
https://github.com/statechannels/native-utils/blob/308b66758e89b334a62f0c25ab3b7834cc7caf62/packages/native-utils/tests/recover.test.ts#L92-101
which means our wallet would never allow a channel with a zero participant to get to the funding point.

snario: I think it would still be worth adding some human readable error at the API level of the wallet

snario: Probably these sorts of checks should be provided in `wallet-core` as pure functions that can easily understood and provided in any language

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -395,6 +395,7 @@ contract ForceMove is IForceMove {
     function _recoverSigner(bytes32 _d, Signature memory sig) internal pure returns (address) {
         bytes32 prefixedHash = keccak256(abi.encodePacked(prefix, _d));
         address a = ecrecover(prefixedHash, sig.v, sig.r, sig.s);
+        require(a != address(0), 'Invalid signature');
```
snario, line 398: Error message could specify `— null address is not a valid signer`?

snario (approved):

packages/nitro-protocol/contracts/ForceMove.sol
geoknee: I took inspiration from here https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ef32c04795e0288c0f71b08352b1877ddb509832/contracts/cryptography/ECDSA.sol#L60

I

PR #2931, opened 2020-11-20
refactor: tidy up and dry out AssetHolder.sol a bit

geoknee:
Preparatory work for adding a `claim` method.

PR #2932, opened 2020-11-20
feat(nitro-protocol): add claim method to AssetHolder

geoknee:
Includes #2931

See also #2814 .

Gas costs (see accompanying test suite)

```
1. straight-through guarantee, 3 destinations:
40559 gas
3. swap guarantee,             3 destinations:
40547 gas
4. straight-through guarantee, 2 destinations:
37586 gas
7. swap guarantee, overfunded, 2 destinations:
53094 gas
8. underspecified guarantee, overfunded      :
52393 gas
```

### DIFF
--- a/packages/docs-website/docs/protocol-tutorial/outcomes.md
+++ b/packages/docs-website/docs/protocol-tutorial/outcomes.md
@@ -67,6 +67,10 @@ A channel that has a guarantee outcome is said to be a guarantor channel.
 
 Now while we can _transfer_ assets out of a channel, the terminology is instead to _claim_ on a guarantor channel. Assets will be paid to beneficiaries of the _target_ channel, but in an order of precedence defined by the guarantor. More on this later.
 
+:::warning
+For simplicity, the on-chain code assumes that neither kind of allocation has any repeated entries.
+:::
+
 Constructing the right kind of object in typescript is straightforward:
 
 ```typescript
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -367,7 +367,7 @@ contract AssetHolder is IAssetHolder {
 
         for (j = 0; j < guarantee.destinations.length; j++) {
             // for each _destination in the guarantee
-            bytes32 _destination = guarantee.destinations[i];
+            bytes32 _destination = guarantee.destinations[j];
             // if destination matches the one that is passed in
             // loop over allocations in the target channel and decrease balance until we hit the specified destination
             if (_destination == destination) {
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -411,8 +411,9 @@ contract AssetHolder is IAssetHolder {
             );
             for (uint256 k = 0; k < allocation.length; k++) {
                 newAllocation[k] = allocation[k];
-                if (allocation[k].destination == destination) {
+                if (k == i) {
                     newAllocation[k].amount = residualAllocationAmount;
+                    break;
                 }
             }
             assetOutcomeHashes[guarantee.targetChannelId] = keccak256(
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -53,6 +53,28 @@ contract AssetHolder is IAssetHolder {
     }
 
     /**
+     * @notice Transfers as many funds escrowed against `guarantorChannelId` as can be afforded for a specific destination in the beneficiaries of the __target__ of that channel. Checks against the storage in this contract.
+     * @dev Transfers as many funds escrowed against `guarantorChannelId` as can be afforded for a specific destination in the beneficiaries of the __target__ of that channel. Checks against the storage in this contract.
+     * @param guarantorChannelId Unique identifier for a guarantor state channel.
+     * @param guaranteeBytes The abi.encode of Outcome.Guarantee
+     * @param allocationBytes The abi.encode of AssetOutcome.Allocation for the __target__
+     * @param destination External destination or channel to transfer funds *to*.
+     */
+    function claim(
+        bytes32 guarantorChannelId,
+        bytes memory guaranteeBytes,
+        bytes memory allocationBytes,
+        bytes32 destination
+    ) public {
+        // checks
+        _requireCorrectGuaranteeHash(guarantorChannelId, guaranteeBytes);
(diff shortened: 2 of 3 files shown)


packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -17,6 +17,128 @@ contract AssetHolder is IAssetHolder {
...
+     * @param guarantorChannelId Unique identifier for a guarantor state channel.
+     * @param guaranteeBytes The abi.encode of Outcome.Guarantee
+     * @param allocationBytes The abi.encode of AssetOutcome.Allocation for the __target__
+     * @param destination External destination or channel to transfer funds *to*.
+     */
+    function claim(
+        bytes32 guarantorChannelId,
+        bytes memory guaranteeBytes,
+        bytes memory allocationBytes,
+        bytes32 destination
```
geoknee, line 64: Idea from @andrewgordstewart: Instead of passing in the `destination`, better to pass in the `index` of the destination in the allocation. Same goes (actually more urgently so) for `transfer`. 

We could even pass `indices: number[]` to target multiple destinations -- and have `[]` imply "all".

Because we pass in the abi-encoded allocation, this implies we know which destination is at which index (so no need to recheck this on chain).

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -312,15 +331,153 @@ contract AssetHolder is IAssetHolder {
...
+        uint256 _amount;
+        uint256 j; // indexes guarantee 
+        uint256 i; // indexes target allocations
+
+        for (j = 0; j < guarantee.destinations.length; j++) {
+            // for each _destination in the guarantee
+            bytes32 _destination = guarantee.destinations[i];
+            // if destination matches the one that is passed in
+            // loop over allocations in the target channel and decrease balance until we hit the specified destination
+            if (_destination == destination) {
```
tomclose, line 370: I'm confused whether this does the right thing? It seems like it should be:

* For each _destination in guarantee:
   * Iterate over the allocation to find _amt it sends to that destination
   * If _destination != destination
        * Decrease balance by _amt
        * Next _destination
    * Else
        * affords = min(balance, _amt)
        * break

whereas it seems like once you find the destination in the guarantee, you then walk along the allocation as normal decreasing the balance?

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -312,15 +331,153 @@ contract AssetHolder is IAssetHolder {
...
+        uint256 balance = holdings[guarantorChannelId];
+        uint256 affordsForDestination;
+        uint256 residualAllocationAmount;
+        uint256 _amount;
+        uint256 j; // indexes guarantee 
+        uint256 i; // indexes target allocations
+
+        for (j = 0; j < guarantee.destinations.length; j++) {
+            // for each _destination in the guarantee
+            bytes32 _destination = guarantee.destinations[i];
```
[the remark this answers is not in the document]
geoknee: yes good catch

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: Yeah you are correct. I even put a bad test case in that passed with this behaviour. 

Hopefully fixed in a6cac4b

packages/docs-website/docs/protocol-tutorial/outcomes.md
```diff
@@ -67,6 +67,10 @@ A channel that has a guarantee outcome is said to be a guarantor channel.
 
 Now while we can _transfer_ assets out of a channel, the terminology is instead to _claim_ on a guarantor channel. Assets will be paid to beneficiaries of the _target_ channel, but in an order of precedence defined by the guarantor. More on this later.
 
+:::warning
+For simplicity, the on-chain code assumes that neither kind of allocation has any repeated entries.
+:::
```
geoknee, line 72: 👀  we should put explicit checks in our wallet or else augment the on chain code to fully treat repeated destinations.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -312,15 +330,138 @@ contract AssetHolder is IAssetHolder {
...
+                revert('_claim | guarantorChannel affords 0 for destination');
+            }
+            // for each destination in the guarantee,
+            // find the first corresponding allocationItem in the target allocation
+            bytes32 guaranteeDestination = guarantee.destinations[j];
+            for (i = 0; i < allocation.length; i++) {
+                if (allocation[i].destination == guaranteeDestination) {
+                    // decrease balance
+                    uint256 _amount = allocation[i].amount;
+                        if (balance < _amount) {
```
tomclose: The alignment is off here.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -312,14 +333,138 @@ contract AssetHolder is IAssetHolder {
...
+                abi.encode(
+                    Outcome.AssetOutcome(
+                        uint8(Outcome.AssetOutcomeType.Allocation),
+                        abi.encode(newAllocation)
+                    )
+                )
+            );
+        }
+
+        if (residualAllocationAmount == 0) {
```
tomclose, line 428: Could this be an `else`?

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -312,14 +333,138 @@ contract AssetHolder is IAssetHolder {
...
+
+        // construct new outcome for target
+        if (residualAllocationAmount > 0) {
+            // new allocation identical save for a single entry
+            Outcome.AllocationItem[] memory newAllocation = new Outcome.AllocationItem[](
+                allocation.length
+            );
+            for (uint256 k = 0; k < allocation.length; k++) {
+                newAllocation[k] = allocation[k];
+                if (allocation[k].destination == destination) {
```
tomclose, line 414: Could you use `k==i` here instead?

tomclose (approved):

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: It could be. Personally, I prefer doing it this way as it seems a bit easier to read and reduces indentation.

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: Yes, and that is actually better since here we can potentially go around again and find a second (repeated) entry if it exists (whereas we are not doing that above).

geoknee: 20bfad5

PR #2933, opened 2020-11-20
perf: make many public methods external

geoknee:
In theory this saves gas, although I am not seeing that in our test suite.

### DIFF
--- a/packages/nitro-protocol/contracts/ForceMove.sol
+++ b/packages/nitro-protocol/contracts/ForceMove.sol
@@ -22,7 +22,7 @@ contract ForceMove is IForceMove {
      * @return fingerprint Unique identifier for the channel's current state, up to hash collisions.
      */
     function getChannelStorage(bytes32 channelId)
-        public
+        external
         view
         returns (
             uint48 turnNumRecord,
@@ -52,7 +52,7 @@ contract ForceMove is IForceMove {
         Signature[] memory sigs,
         uint8[] memory whoSignedWhat,
         Signature memory challengerSig
-    ) public override {
+    ) external override {
         bytes32 channelId = _getChannelId(fixedPart);
 
         if (_mode(channelId) == ChannelMode.Open) {
@@ -121,7 +121,7 @@ contract ForceMove is IForceMove {
         // variablePartAB[0] = challengeVariablePart
         // variablePartAB[1] = responseVariablePart
         Signature memory sig
-    ) public override {
+    ) external override {
         bytes32 channelId = _getChannelId(fixedPart);
         (uint48 turnNumRecord, uint48 finalizesAt, ) = _getChannelStorage(channelId);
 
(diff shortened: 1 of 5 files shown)


andrewgordstewart (approved):

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -52,7 +52,7 @@ contract ForceMove is IForceMove {
         Signature[] memory sigs,
         uint8[] memory whoSignedWhat,
         Signature memory challengerSig
-    ) public override {
+    ) external override {
```
geoknee, line 55: Point of confusion. This contract implements the `IForceMove` interface, where `external` method parameters have their parameters marked as `calldata` instead of `memory`. 

If I make that change in *this* contract, I get 
```shell
➜  nitro-protocol git:(externalize) ✗ yarn contract:compile --show-stack-traces
yarn run v1.22.5
$ yarn hardhat compile --force --show-stack-traces
$ /Users/georgeknee/statechannels/statechannels/node_modules/.bin/hardhat compile --force --show-stack-traces
Compiling 25 files with 0.7.4
UnimplementedFeatureError: 

Error HH600: Compilation failed

HardhatError: HH600: Compilation failed
    at SimpleTaskDefinition.action (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/builtin-tasks/compile.ts:741:15)
    at Environment._runTaskDefinition (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at Environment.run (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/internal/core/runtime-environment.ts:130:14)
    at SimpleTaskDefinition.action (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/builtin-tasks/compile.ts:954:7)
    at Environment._runTaskDefinition (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at Environment.run (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/internal/core/runtime-environment.ts:130:14)
    at SimpleTaskDefinition.action (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/builtin-tasks/compile.ts:384:45)
    at Environment._runTaskDefinition (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at Environment.run (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/internal/core/runtime-environment.ts:130:14)
    at SimpleTaskDefinition.action (/Users/georgeknee/statechannels/statechannels/node_modules/hardhat/src/builtin-tasks/compile.ts:1172:63)
error Command failed with exit code 1.
```

geoknee: Digging into this a bit more, I was starting to see errors like this

```
contracts/ForceMove.sol:479:17: TypeError: Invalid type for argument in function call. Invalid implicit conversion from struct IForceMove.FixedPart memory to struct IForceMove.FixedPart calldata requested.
                fixedPart,
```
and now things make sense. This PR didn't actually change any parameters from `memory` to `calldata` (exceptions being base contracts which were overridden anyway), hence no gas savings. Doing so causes problems when internal functions expect `memory` and not `calldata`. 

Making such changes can be attempted in separate PRs.

PR #2935, opened 2020-11-20
chore: upgrade prettier-plugin-solidity

geoknee:
Confusingly, running `yarn lint:write` after this change gives me a bunch of formatting changes in several packages, both to `.sol` and `.ts` files 😖


geoknee:
I dug into this a bit and concluded that for whatever reason, the changes in this PR are good because they seem to implement our configuration (whereas before this was not happening). 

For example: I believe the defaults in prettier (that we do not override) are to have spaces around bracketed imports and to have trailing commas in multiline object literals. These make up the vast majority of linting changes on this PR. 

To change these we could use
`--no-bracket-spacing`
`--trailing-comma`
options.

snario:
Usually for these sorts of PRs I like to explicitly tell the linting config files to keep the status quo and then separate out styling changes into separate PRs. For example, you could set the bracket spacing and trailing comma rules in `eslintrc.js` to be the same as what we have right now like you mentioned. Then in a different PR change those if you want — but more importantly in the immediate term we can get the solidity linting in

andrewgordstewart:
I think we _definitely_ want to limit the changeset in this PR to be as small as possible.

There are 8 other open PRs in this repo that will be effected by this. If we want to make a config change that adds or removes commas in hundreds of files, we need to have a single commit `f00` such that, when we rebase or merge, we can programmatically fix merge conflicts:
1. run `git checkout --ours *` (or `--yours`, I always have to look it up)
2. run `yarn lint:write`

geoknee:
#2948

PR #2937, opened 2020-11-20
wip: test challenging with a  large outcome

geoknee:
I have hacked this together rather quickly, but wanted to share the methodology and some preliminary results. 

```
It copes with a large outcome with 10 allocationItems :
133788 gas

It copes with a large outcome with 100 allocationItems :
271493 gas
```

I ran out of memory when trying to do 1000 allocations 😆 .

But these two results suggest adding 90 allocationItems incurred an additional 140K gas:  so 1.5K gas per allocation. This is very back of the envelope kind of stuff, but... this might mean we could raise a challenge in a ledger channel with an outcome allocating to around 5000 payment channels.

### DIFF
--- a/packages/nitro-protocol/test/contracts/ForceMove/challenge-large-outcome.ts
+++ b/packages/nitro-protocol/test/contracts/ForceMove/challenge-large-outcome.ts
@@ -1,0 +1,149 @@
+import {configureEnvVariables, expectRevert, GanacheServer} from '@statechannels/devtools';
+import {Contract, Wallet, ethers, Signature} from 'ethers';
+
+const {HashZero} = ethers.constants;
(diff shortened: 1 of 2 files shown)


geoknee:
By moving out of `jest` and to `ts-node`, and by upping the memory limit to 8gb(!), I got the 1000 outcomes case to run successfully (locally). 

`1000 allocations:  1710675 gas`

I hit out-of-memory issues somewhere above 1000 allocation items.

geoknee:
The out of memory issues *seem* to coincide with the size of the transaction going north of 128KB, which is the point at which clients such as `geth` will throw an "oversized data" error. 

Closing in favour of #2972, which adopts the approach of first finding a safe number of allocations given this 128KB limit, and *then* checking that the method call doesn't consume excessive amounts of gas.

PR #2944, opened 2020-11-23
test: reliably generate incorrect signatures

### LINKED ISSUE #2941: Flickering test
https://app.circleci.com/pipelines/github/statechannels/statechannels/10584/workflows/d4814a01-de95-4df5-92d7-009e2dfb9b3e/jobs/52037

geoknee:
Fixes #2941

### DIFF
--- a/packages/nitro-protocol/src/signatures.ts
+++ b/packages/nitro-protocol/src/signatures.ts
@@ -71,7 +71,7 @@ function hashMessage(hashedData: string): string {
   return utils.hashMessage(utils.arrayify(hashedData));
 }
 
-function signData(hashedData: string, privateKey: string): Signature {
+export function signData(hashedData: string, privateKey: string): Signature {
   const signingKey = new utils.SigningKey(privateKey);
 
   return utils.splitSignature(signingKey.signDigest(hashMessage(hashedData)));
--- a/packages/nitro-protocol/test/contracts/ForceMove/challenge.test.ts
+++ b/packages/nitro-protocol/test/contracts/ForceMove/challenge.test.ts
@@ -24,6 +24,7 @@ import {
   getPlaceHolderContractAddress,
   getRandomNonce,
   getTestProvider,
+  nonParticipant,
   ongoingChallengeHash,
   setupContracts,
   writeGasConsumption,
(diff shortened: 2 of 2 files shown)


packages/nitro-protocol/test/contracts/ForceMove/challenge.test.ts
```diff
@@ -180,7 +181,7 @@ describe('challenge', () => {
 
       switch (challengeSignatureType) {
         case 'incorrect':
-          challengeSignature = {...correctChallengeSignature, v: correctChallengeSignature.v + 1};
```
geoknee, line 183: This (very hacky) approach was sometimes generating a valid but incorrect signature (desired), and sometimes generating an entirely invalid signature (undesired)

packages/nitro-protocol/test/contracts/ForceMove/challenge.test.ts
```diff
@@ -283,3 +284,23 @@ describe('forceMove with transaction generator', () => {
...
+  while (participants.find(participant => participant == nonParticipant)) {
+    const wallet = Wallet.createRandom();
+    nonParticipant = wallet.address;
+    pK = wallet.privateKey;
+  }
+  const challengeState = signedStates[signedStates.length - 1].state;
+  const challengeHash = hashChallengeMessage(challengeState);
+
+  return signData(challengeHash, pK);
+}
```
geoknee, line 306: This may be over-engineered, but it should be pretty reliable.

andrewgordstewart:
Couldn't we used a constant, valid signature? The chance that it's correct is essentially zero

geoknee:
> Couldn't we used a constant, valid signature? The chance that it's correct is essentially zero

2fb9c8f

packages/nitro-protocol/src/signatures.ts
```diff
@@ -71,7 +71,7 @@ function hashMessage(hashedData: string): string {
   return utils.hashMessage(utils.arrayify(hashedData));
 }
 
-function signData(hashedData: string, privateKey: string): Signature {
+export function signData(hashedData: string, privateKey: string): Signature {
   const signingKey = new utils.SigningKey(privateKey);
 
   return utils.splitSignature(signingKey.signDigest(hashMessage(hashedData)));
```
snario, line 77: Why is the argument called `hashedData` when we pass it into `hashMessage`? I forget if we're intentionally double hashing, or if the naming is just confusing

geoknee: We are double hashing.

snario (approved):

PR #2750, opened 2020-10-26
liam/ledgers db

snario:

PR #2945, opened 2020-11-23
perf: optimize queries for ledger requests on run loop

snario:
When thinking about performance a bit I looked at this code and thought I could optimize the query made to the DB a bit instead of doing filtering in JS.

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -469,6 +469,44 @@ export class Store {
     return LedgerRequest.getRequest(channelId, type, tx || this.knex);
   }
 
+  async getChannelIdsPendingLedgerFundingFrom(
+    ledgerChannelIds: Bytes32[],
+    tx?: Transaction
+  ): Promise<Bytes32[]> {
+    const query = await LedgerRequest.query(tx || this.knex)
+      .whereIn('ledgerChannelId', ledgerChannelIds)
+      .where({status: 'pending'})
+      .distinct('channelToBeFunded')
+      .select('channelToBeFunded');
+
+    return _.map(query, 'channelToBeFunded');
+  }
+
+  async getLedgerChannelIdsFundingChannels(
+    channelToBeFunded: Bytes32[],
+    tx?: Transaction
+  ): Promise<Bytes32[]> {
+    const query = await LedgerRequest.query(tx || this.knex)
+      .whereIn('channelToBeFunded', channelToBeFunded)
+      .where({status: 'pending'})
+      .distinct('ledgerChannelId')
+      .select('ledgerChannelId');
+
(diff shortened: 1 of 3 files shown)


packages/server-wallet/src/wallet/store.ts
```diff
@@ -469,6 +469,44 @@ export class Store {
     return LedgerRequest.getRequest(channelId, type, tx || this.knex);
   }
 
+  async getChannelIdsPendingLedgerFundingFrom(
+    ledgerChannelIds: Bytes32[],
+    tx?: Transaction
+  ): Promise<Bytes32[]> {
+    const query = await LedgerRequest.query(tx || this.knex)
+      .whereIn('ledgerChannelId', ledgerChannelIds)
```
andrewgordstewart, line 477: There's no index on `ledger_channel_id`.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -469,6 +469,44 @@ export class Store {
...
+    return _.map(query, 'ledgerChannelId');
+  }
+
+  async filterChannelIdsByIsLedger(
+    maybeLedgerChannelIds: Bytes32[],
+    tx?: Transaction
+  ): Promise<Bytes32[]> {
+    const query = await Channel.query(tx || this.knex)
+      .whereIn('channelId', maybeLedgerChannelIds)
+      .whereNotNull('assetHolderAddress')
```
andrewgordstewart, line 504: Why is `assetHolderAddress !== null` equivalent to being a ledger channel?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -469,6 +469,44 @@ export class Store {
...
   }
 
+  async getChannelIdsPendingLedgerFundingFrom(
+    ledgerChannelIds: Bytes32[],
+    tx?: Transaction
+  ): Promise<Bytes32[]> {
+    const query = await LedgerRequest.query(tx || this.knex)
+      .whereIn('ledgerChannelId', ledgerChannelIds)
+      .where({status: 'pending'})
+      .distinct('channelToBeFunded')
```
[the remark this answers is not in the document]
snario: There is no garbage collection so this table could have both requests to `fund` and `defund`, which is why we do distinct

packages/server-wallet/src/wallet/store.ts
snario: Instead of creating a new `ledgers` table #2750 added this column to the `channels` table and this column is unique for ledger channels

packages/server-wallet/src/wallet/store.ts
snario: ba64b3f2e

andrewgordstewart (approved):

PR #2545, opened 2020-09-09
Validate the db config

andrewgordstewart:
Provides a useful error in this sort of scenario.
```
packages/e2e-testing/src on  master [!] on ☁️  us-west-1 
❯ SERVER_DB_NAME=payment ts-node payment-server.ts start
[@statechannels/devtools] NODE_ENV is undefined — setting to "development"

/Users/andrewstewart/Code/magmo/monorepo/packages/server-wallet/src/config.ts:30
  else throw new Error(`Expected process.env.${key} to be a string`);
             ^
Error: Expected process.env.SERVER_DB_USER to be a string
```

PR #2949, opened 2020-11-23
Refactor Server Wallet Config

lalexgap:
Refactor the server wallet configuration to hopefully make it easier to use and understand.

The config has now been separated into two types: `RequiredServerWalletConfig` and `OptionalServerWalletConfig`. 

`RequiredServerWalletConfig` are the configuration options that **must** be specified by the user of the wallet.

`OptionalServerWalletConfig` are various options that can be optional provided. If these options are not provided then the `defaultConfig` is used (which contains sane defaults).

The `RequiredServerWalletConfig` must contain either a connection string of a database name.

Configuration options are now grouped to clean up clutter and keep things organised.

The `defaults.ts` file contains default configuration values. There is a `defaultConfig` used to provide `OptionalServerWalletConfig` and a `defaultTestConfig` that can be used for testing (that disables evm validation and worker threading)

The `utils.ts` contains some functions for working with the configuration. `overwriteConfigWithDatabaseConnection` allows a config file to be updated with a connection string which is useful in tests with multiple databases.

TODO: 
- [x] passing tests?
- [ ] validation
- [x] documentation
- [x] code organisation
- [x] update `the-graph` repo to use new config (or just accept a wallet?)


packages/server-wallet/src/config/defaults.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Network config
+ */
+
+export const defaultNetworkConfiguration: NetworkConfiguration = {
+  chainNetworkID: '0x00',
+  erc20Address: undefined,
+  ethAssetHolderAddress: undefined,
+  erc20AssetHolderAddress: undefined,
```
snario, line 26: I don't think `erc20AssetHolderAddress` nor `erc20Address` should be here, and `ethAssetHolderAddress` is always `0x0000000000000000000000000000000000000000`

packages/server-wallet/src/config/defaults.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Server Wallet config
+ */
+
+/**
+ * These are the default values that will be used by the server wallet
+ * if not overidden in the provided config
+ */
+export const defaultConfig: OptionalServerWalletConfig = {
+  databaseConfiguration: defaultDatabaseConfiguration,
```
snario, line 48: `databaseConfiguration` probably better named `postgresqlConfiguration` as it is quite specific

packages/server-wallet/src/config/defaults.ts
```diff
@@ -0,0 +1,76 @@
...
+  networkConfiguration: defaultNetworkConfiguration,
+  loggingConfiguration: defaultLoggingConfiguration,
+  metricConfiguration: defaultMetricConfiguration,
+  // TODO: List addresses these keys correspond to
+  stateChannelPrivateKey: '0x1b427b7ab88e2e10674b5aa92bb63c0ca26aa0b5a858e1d17295db6ad91c049b',
```
snario, line 53: https://github.com/statechannels/statechannels/pull/2952

packages/server-wallet/src/config/defaults.ts
```diff
@@ -0,0 +1,76 @@
...
+  ethereumPrivateKey: '0x7ab741b57e8d94dd7e1a29055646bafde7010f38a900f55bbd7647880faa6ee8',
+  skipEvmValidation: false,
+  workerThreadAmount: 10,
```
snario, line 56: Probably warrants its own section in the config object like "`concurrencyConfiguration`" or something

packages/server-wallet/src/config/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export function extractDBConfigFromServerWalletConfig(
+  serverWalletConfig: ServerWalletConfig
+): Config {
+  const connectionConfig = getDatabaseConnectionConfig(serverWalletConfig);
+
+  return {
+    client: 'postgres',
+    // TODO: Might make sense to use `database` instead of `dbName` so its consitent with knex
```
snario, line 66: 👍

packages/server-wallet/src/db-admin/knexfile.ts
```diff
@@ -16,35 +23,32 @@ WARNING: @statechannels/devtools not detected.
...
+    migrations: {
+      directory: path.join(BASE_PATH, 'migrations'),
+      loadExtensions: extensions,
+    },
+    seeds: {
+      directory: path.join(BASE_PATH, 'seeds'),
+      loadExtensions: extensions,
+    },
+    /**
+     * SEE: db-admin-connection.ts
```
snario, line 41: This file does not exist anymore

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
```diff
@@ -6,37 +6,38 @@ import {UpdateChannelParams} from '@statechannels/client-api-schema';
...
+          this.logger.trace('Starting worker');
 
           const worker = new Worker(path.resolve(__dirname, './loader.js'), {
             workerData: walletConfig,
           });
 
-          worker.stdout.on('data', data => logger.info(data.toString()));
-          worker.stderr.on('data', data => logger.error(data.toString()));
+          worker.stdout.on('data', data => this.logger.info(data.toString()));
+          worker.stderr.on('data', data => this.logger.error(data.toString()));
```
snario, line 36: If `data` is a stringified object e.g., `"{type:'debug',msg:'blah',channelId:'0xabc'}"` would `this.logger.info` interpret it as an object or a string?

packages/server-wallet/src/config/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+      ...defaultTestConfig.databaseConfiguration,
+      connection: {host, port, dbName: databaseName || DEFAULT_DB_NAME},
+    },
+  };
+}
+type DatabaseConnectionConfigObject = Exclude<DatabaseConnectionConfiguration, string>;
+export function getDatabaseConnectionConfig(
+  config: ServerWalletConfig
+): DatabaseConnectionConfigObject {
+  if (typeof config.databaseConfiguration.connection === 'string') {
```
snario, line 88: When is this true in our projects?

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
andrewgordstewart: This explains some noisy and unreadable logs I've seen. What is even the purpose of this?

lalexgap: I think the purpose of this is to capture logging output from the `worker` when `LOG_DESTINATION=console`. However it looks like the parent `stdout` gets passed in by default so this probably just messes up the logs.

packages/server-wallet/src/config/utils.ts
lalexgap: This allows the wallet to take in a connection string. It seems a nice property to either accept a connection string or a config object (like `knex`). Although perhaps it's not worth the overhead.

packages/server-wallet/src/config/defaults.ts
lalexgap: Why don't you think they should be there?

packages/server-wallet/src/config/utils.ts
andrewgordstewart: https://github.com/statechannels/the-graph/blob/master/.circleci/config.yml#L10

Here's how it could be in `server-wallet`.

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
andrewgordstewart: I think the problem would be solved if the worker had the same logging config as the main thread.
- if it's `console`, then there's no need to print the same log line twice, once in an unreadable format
- if the main thread's logger logs to a filename, then so should the worker thread's right?

packages/server-wallet/src/config/defaults.ts
lalexgap: Sweet, I've incorporated that into this.

packages/server-wallet/src/config/defaults.ts
snario: There is no such thing as an "ERC20 asset" — it is an asset class. The assets are things like DAI, GRT, etc.

andrewgordstewart: > `ethAssetHolderAddress` is always `0x0000000000000000000000000000000000000000`

Why would it always be the zero address?

packages/server-wallet/src/config/defaults.ts
```diff
@@ -0,0 +1,75 @@
...
+  networkConfiguration: defaultNetworkConfiguration,
+  loggingConfiguration: defaultLoggingConfiguration,
+  metricConfiguration: defaultMetricConfiguration,
+  // TODO: List addresses this corresponds to
+  ethereumPrivateKey: '0x7ab741b57e8d94dd7e1a29055646bafde7010f38a900f55bbd7647880faa6ee8',
+  skipEvmValidation: false,
+  workerThreadAmount: 10,
+};
+
+export const DEFAULT_DB_NAME = 'server_wallet_test';
```
andrewgordstewart, line 58: This should not be the default.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -25,9 +25,9 @@ const ethAssetHolderAddress = makeAddress(process.env.ETH_ASSET_HOLDER_ADDRESS!)
 const erc20AssetHolderAddress = makeAddress(process.env.ERC20_ASSET_HOLDER_ADDRESS!);
 const erc20Address = makeAddress(process.env.ERC20_ADDRESS!);
 /* eslint-enable no-process-env, @typescript-eslint/no-non-null-assertion */
```
andrewgordstewart, line 27: This comment can be removed. (Applies to many files changed here.)

packages/server-wallet/src/config/defaults.ts
```diff
@@ -0,0 +1,75 @@
...
+export const DEFAULT_DB_USER = 'postgres';
```
andrewgordstewart, line 59: I don't think this should be the default either. It seems like the default configuration in _our setup_ (the postgres app on macOS) is leaking into the source code of the wallet.

packages/server-wallet/src/config/defaults.ts
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable no-process-env */
+import {
+  LoggingConfiguration,
+  NetworkConfiguration,
+  OptionalDatabaseConfiguration,
+  OptionalServerWalletConfig,
+  ServerWalletConfig,
+} from './types';
+
+export const defaultDatabaseConfiguration: OptionalDatabaseConfiguration & {
```
andrewgordstewart, line 10: It seems like these defaults are actually the parameters that we expect in our _test environment_. I don't think it's appropriate for such values to act as _default configs_.

Some of them, such as the `host` and `port` of the postgres DB might be reasonable defaults:
- postgres _always_ uses port `5432` by default
- I'm not actually sure about `host`, because I am not sure it is the "default" to run the DB on the same server as the process running the wallet

packages/server-wallet/src/config/defaults.ts
```diff
@@ -0,0 +1,75 @@
...
+ * These are the default values that will be used by the server wallet
+ * if not overidden in the provided config
+ */
+export const defaultConfig: OptionalServerWalletConfig = {
+  databaseConfiguration: defaultDatabaseConfiguration,
+  networkConfiguration: defaultNetworkConfiguration,
+  loggingConfiguration: defaultLoggingConfiguration,
+  metricConfiguration: defaultMetricConfiguration,
+  // TODO: List addresses this corresponds to
+  ethereumPrivateKey: '0x7ab741b57e8d94dd7e1a29055646bafde7010f38a900f55bbd7647880faa6ee8',
```
andrewgordstewart, line 53: I don't think this belongs in the defaults. (I also read somewhere that it's not used anymore?)

packages/server-wallet/src/wallet/store.ts
```diff
@@ -46,7 +46,10 @@ import {ObjectiveModel, DBObjective} from '../models/objective';
 import {AppBytecode} from '../models/app-bytecode';
 import {LedgerRequest, LedgerRequestType} from '../models/ledger-request';
 import {shouldValidateTransition, validateTransition} from '../utilities/validate-transition';
-import {logger as defaultLogger} from '../logger';
+import {defaultTestConfig} from '../config';
+import {createLogger} from '../logger';
+
+const defaultLogger = createLogger(defaultTestConfig);
```
andrewgordstewart, line 52: This smells pretty bad. Why is the default logger using the `defaultTestConfig`?

packages/server-wallet/src/db-admin/knexfile.ts
```diff
@@ -16,35 +23,31 @@ WARNING: @statechannels/devtools not detected.
...
+     * overwrite the two config options set by knexSnakeCaseMappers
+     */
+    postProcessResponse: undefined,
+    wrapIdentifier: undefined,
+  };
 }
 
-export const {client, connection, debug, migrations, seeds, pool} = knexConfig;
+export const {client, connection, debug, migrations, seeds, pool} = createKnexConfig(
+  overwriteConfigWithEnvVars(defaultTestConfig)
```
andrewgordstewart, line 52: On first glance, this seems wrong. This is the `db-admin` file. Is it _only_ going to be used in a test scenario?

packages/server-wallet/src/config/types.ts
```diff
@@ -0,0 +1,92 @@
...
+ * The minimum required configuration to use the server wallet.
+ */
+export type RequiredServerWalletConfig = {
+  databaseConfiguration: RequiredDatabaseConfiguration;
+};
+
+/**
+ * Additional configuration options for the server wallet that are not required.
+ */
+export interface OptionalServerWalletConfig {
```
andrewgordstewart, line 61: I think some of these config variables should be required, such as `networkConfiguration`.

andrewgordstewart (changes requested):
The change looks great, but it also looks like it exposes problems with the current configuration.

There appear to be two different settings where we want the wallet to be configured.
1. In some environment controlled by us, where we for some reason want to declare the configuration by environment variables
2. In some environment not controlled by us (users, such as a Graph Gateway or Indexer), where we want to be told _very precisely_ what configuration to use.

## World 1, env variables
In (1), I'd prefer to not use environment variables anymore. The `SC_ENV` environment variable that we previous used, to point to a file defining the entire configuration, could instead be used here. Since the configuration as of this PR is a JSON object, the SC_ENV file can also be a JSON object. Future work could validate this JSON object against a schema, if we deem something like that effort worthwhile. (I think there are [lighter-weight validation strategies](https://github.com/statechannels/statechannels/pull/2545/files#diff-797a4d689c0809d196c06ddda00e8c32b4397c2748c4068571fa5f50324c7865R63) that will do just fine.)

If we were to do this, then lines like
```
const config = overwriteConfigWithEnvVars(defaultTestConfig)
```
can be replaced with
```
// path/to/test.config.ts
export const config = validateConfig(require(process.env.SC_ENV))

// some/component.test.ts
import {config} from 'path/to/test.config.ts'
```

## World 2, "production"-y environment
In (2), I think we want to make sure that the only defaults we use affect the _behaviour_ of the wallet. In other words, the configuration of external resources (contract addresses, DB configuration, network configuration) should not have default values. This seems important to protect against users accidentally using `config.networkConfiguration.chainId == '0x00'`.


## Requested changes
- I don't think we need to worry about world 1 right now.
- I do think we ought to address issues around world 2 before merging this PR.

packages/server-wallet/src/config/defaults.ts
lalexgap: There was a different private key that was removed already.

packages/server-wallet/src/wallet/store.ts
lalexgap: Yeah this should probably be cleaned up so that the `Store` **requires** a logger  (or only tries to log when provided one).

andrewgordstewart: https://github.com/statechannels/statechannels/issues/2959

packages/server-wallet/src/db-admin/knexfile.ts
lalexgap: This is slightly hacky. `overwriteConfigWithEnvVars` expects a config with a `dbName` defined so we use the `defaultTestConfig`.

I think the correct behavior here is to read in the config from env and throw if it doesn't define a dbname or connection string

lalexgap:
> ## Requested changes
> * I don't think we need to worry about world 1 right now.
> * I do think we ought to address issues around world 2 before merging this PR.

I agree about World 1 and I think that might be something I can look at in the separate PR.

For point 2 I'll go through and try to move more appropriate fields into the `Required` types.

packages/server-wallet/src/config/defaults.ts
```diff
@@ -46,24 +41,29 @@ export const defaultMetricsConfiguration = {timingMetrics: false};
  */
 export const defaultConfig: OptionalServerWalletConfig = {
   databaseConfiguration: defaultDatabaseConfiguration,
-  networkConfiguration: defaultNetworkConfiguration,
   loggingConfiguration: defaultLoggingConfiguration,
   metricsConfiguration: defaultMetricsConfiguration,
-  // TODO: List addresses this corresponds to
-  ethereumPrivateKey: '0x7ab741b57e8d94dd7e1a29055646bafde7010f38a900f55bbd7647880faa6ee8',
   skipEvmValidation: false,
-  workerThreadAmount: 10,
```
andrewgordstewart: IMO this could reasonably be a default, but the value should probably be `0`?

packages/server-wallet/deployment/deploy.ts
```diff
@@ -1,9 +1,10 @@
 import {Address} from '@statechannels/client-api-schema';
 import {ContractArtifacts} from '@statechannels/nitro-protocol';
 import {GanacheDeployer} from '@statechannels/devtools';
-import {defaultConfig} from '../src/config';
 import {Wallet} from 'ethers';
 
+import {defaultTestConfig} from '../src/config';
```
andrewgordstewart, line 6: Is `deploy.ts` a test-specific deploy file? If yes, should it be renamed `deploy.test.ts`?

andrewgordstewart (approved):

packages/server-wallet/src/config/defaults.ts
lalexgap: Done

Issue #2928, opened 2020-11-19
Get rid of wallet objectiveSucceeded events?


kerzhner:
Some potential reasons to remove `objectiveSucceeded` events:
1. The events are emitted during runloop execution as opposed to after the runloop has finished. This is different from the behavior of `channelUpdated` event.
1. The events are only used in `directly-funded-channel.test.ts`. This test can be modified to use the `objectiveSucceeded` event if the channel status contains information on whether a channel has been defunded.

This is relevant to https://github.com/statechannels/statechannels/blob/01638f13f532a86853a83596f727e222a49bf7bc/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts#L223 

## TODO: 
- [x] https://github.com/statechannels/statechannels/pull/2922#discussion_r527987072

Issue #2947, opened 2020-11-23
Audit repeated destinations in outcomes


andrewgordstewart:
In some places in both the contracts and the wallet code, it is _assumed_ that there are no repeated outcome items.

1. When is this assumption valid, and why?
2. What are the consequences of making this assumption? 
3. When might we want to invalidate that assumption?

andrewgordstewart:
> 3. When might we want to invalidate that assumption?

Top-ups in ledger channels are more "performant" (smaller number of state updates) if we allow repeat outcome items. The algorithm looks something like this:
1. Alice wants to top-up. Alice proposes an update with an unfunded `{destionation: alicesAddress, amount: topUpAmount}` outcome item at the end of the outcome
2. Once an outcome is supported with that outcome item at the end, Alice deposits
3. Once Alice's deposit is mined and "finalized", Alice proposes an update to merge the `amount: topUpAmount` with the `amount: previousAmount` outcome items

PR #2950, opened 2020-11-23
Add fundingStatus, remove ObjectiveSucceeded

kerzhner:
Fixes https://github.com/statechannels/statechannels/issues/2928.

This PR:
- Concentrates funding status logic for directly funded channels in `Channel` model instead of distributing the logic between open channel and close channel protocols. This lays the groundwork for switching the wallet to handle allocation types other than `SimpleAllocation`. It also removes potential inconsistencies with funding calculations. This is probably the least controversial part of this PR.
- Remove `ObjectiveSucceeded` wallet event.  Reference https://github.com/statechannels/statechannels/issues/2928 for reason.
- Adds `fundingStatus` to `ChannelResult` as a replacement to the `ObjectiveSucceeded` event.

One behavior I noticed is that `funding` property of `ChannelState` is sometimes correct and sometimes incorrect in the wallet. This is due to some channel reads joining the `channels` table with `funding` table while others not doing so. As a result of this, `ChannelState` to `ChannelResult` conversion (which populates the `fundingStatus` property) depends on how `ChannelState` was fetched. The options going forward are:
1. Fix this inconsistency.
1. Chose not to fix this inconsistency due to performance implications of joining `funding` to `channels`. In this case, 
 a consistent `fundingStatus` property is difficult to implement.


packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -136,26 +133,30 @@ it('Create a directly funded channel between two wallets ', async () => {
...
     status: 'proposed',
     turnNum: 0,
+    fundingStatus: 'Uncategorized',
   });
 
   const prefundB = await b.joinChannel({channelId});
   expect(getChannelResultFor(channelId, [prefundB.channelResult])).toMatchObject({
     status: 'opening',
     turnNum: 1,
+    fundingStatus: 'Uncategorized',
```
snario, line 151: Why isn't this `ReadyToFund`? It's not its turn?

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -179,39 +182,57 @@ it('Create a directly funded channel between two wallets ', async () => {
...
+    fundingStatus: 'Funded',
   });
 
   const resultB1 = await b.pushMessage(getPayloadFor(participantB.participantId, postFundA.outbox));
   expect(getChannelResultFor(channelId, resultB1.channelResults)).toMatchObject({
     status: 'running',
     turnNum: 3,
+    // TODO: the fundingStatus is incorrect as the funding table is not joined with channels table
+    //  when processing the pushMessage above
+    // fundingStatus: 'Funded',
```
snario, line 194: I doubt there is much of a performance hit — the table id indexed on `channelId`.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
kerzhner: `ReadyToFund` === all prefund states singed + all deposits before me have been made. I should probably document what each `FundingStatus` means.

snario:
What would the funding status be for a channel with `fundingStrategy` as `"Ledger"` that has a supported prefund setup state?

andrewgordstewart:
The options going forward are:
> 
>     1. Fix this inconsistency.
> 
>     2. Chose not to fix this inconsistency due to performance implications of joining `funding` to `channels`. In this case,
>        a consistent `fundingStatus` property is difficult to implement.

Another option is the following, right?
3. Choose to duplicate the funding status on the `channels` table, which gets updated whenever modifying the `funding` table in a relevant way.

packages/client-api-schema/src/data-types.ts
```diff
@@ -114,6 +114,8 @@ export type ChannelStatus =
   | 'closing' // You cannot use the channel anymore, but your funds are still locked up.
   | 'closed'; // Your funds have been released from the channel.
 
+export type FundingStatus = 'Uncategorized' | 'ReadyToFund' | 'Funded' | 'Defunded';
```
andrewgordstewart, line 117: What does `'Uncategorized'` mean?

packages/server-wallet/src/models/channel.ts
```diff
@@ -173,6 +177,67 @@ export class Channel extends Model implements RequiredColumns {
     });
   }
 
+  public static directFundingStatus(
```
andrewgordstewart, line 180: What's the purpose of this as a static function on a model? It doesn't seem to have anything to do with the `Channel` model. Could it be a standalone function?

packages/server-wallet/src/models/channel.ts
```diff
@@ -173,6 +177,67 @@ export class Channel extends Model implements RequiredColumns {
...
+
+    const outcome = supported?.outcome;
+    if (!supported || !outcome) {
+      return 'Uncategorized';
+    }
+
+    const {allocationItems, assetHolderAddress} = checkThat(outcome, isSimpleAllocation);
+
+    // Collapse all allocation items with my destination into one
+    const myAllocation = allocationItems
```
andrewgordstewart, line 198: This would probably be easier to read if you defined
```
const myDestination = myParticipant.destination;
const myAllocation = allocationItems.filter(**snip**).map(ai => ai.amount).reduce(BN.add, BN.from(0));
```

packages/server-wallet/src/models/channel.ts
```diff
@@ -173,6 +177,67 @@ export class Channel extends Model implements RequiredColumns {
...
+          ...soFar,
+          amount: BN.add(soFar.amount, currentAi.amount),
+        }),
+        {
+          destination: myParticipant.destination,
+          amount: BN.from(0),
+        }
+      );
+
+    const funding = fundingFn(assetHolderAddress);
```
andrewgordstewart, line 211: I am no longer sure the `fundingFn` does what it was meant to do. I don't even remember why it was a function in the first place 😆

packages/server-wallet/src/models/channel.ts
```diff
@@ -173,6 +177,67 @@ export class Channel extends Model implements RequiredColumns {
...
+    }
+
+    const fullFunding = allocationItems.map(a => a.amount).reduce(BN.add, BN.from(0));
+    if (BN.eq(amountTransferredToMe, 0) && BN.gte(funding.amount, fullFunding)) {
+      return 'Funded';
+    }
+
+    const allocationsBeforeMe = _.takeWhile(
+      allocationItems,
+      a => a.destination !== myAllocation.destination
```
andrewgordstewart, line 231: Consider this: https://github.com/statechannels/statechannels/issues/2947#issuecomment-732364811

If we do, or plan to, allow repeated destinations, do you know how to corret this?

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
```diff
@@ -51,10 +53,22 @@ test.each`
...
+      latestSignedByMe,
+      funding,
+      myIndex,
+      directFundingStatus: Channel.directFundingStatus(
+        supported,
+        funding,
+        participants[myIndex],
+        'Direct'
+      ),
+    },
```
andrewgordstewart, line 71: nit: If you defined
```
const directFundingStatus = Channel.directFundingStatus(
        supported,
        funding,
        participants[myIndex],
        'Direct'
      )
```
then this probably fits on one line.

```suggestion
    app: { supported, latest, latestSignedByMe, funding, myIndex, directFundingStatus },
```

packages/server-wallet/src/models/channel.ts
andrewgordstewart, line 180: I think this deserves a unit test with decent coverage.

packages/server-wallet/src/models/channel.ts
```diff
@@ -173,6 +177,67 @@ export class Channel extends Model implements RequiredColumns {
...
   }
 
+  public static directFundingStatus(
+    supported: SignedStateVarsWithHash | undefined,
+    fundingFn: (address: Address) => ChannelStateFunding,
+    myParticipant: Participant,
+    fundingStrategy: FundingStrategy
+  ): FundingStatus {
+    if (fundingStrategy !== 'Direct') {
+      return 'Uncategorized';
```
andrewgordstewart, line 187: Isn't it the case that the direct funding _status_ is independent of the funding _strategy_?

It is unlikely, but possible, that
1. Alice & Bob have a directly funded ledger L, funding a channel C
2. Alice goes on vacation
3. Bob challenges in L, and transfers funds to C

Now, C is directly funded, though its funding strategy was `'Ledger'`.

andrewgordstewart (changes requested):
The change overall looks good, but there's a complicated function that deserves a good unit test.

I'm also curious about why the direct funding status should be `'Uncategorized'` when the strategy is not `'Direct'`.

packages/server-wallet/src/models/channel.ts
kerzhner: Should the `fundingStrategy` get updated to `Direct` in the scenario you describe?

packages/server-wallet/src/models/channel.ts
kerzhner: I think this is a good point, but should probably be tackled in a separate PR. Now that all funding logic is in one place, we can implement and test allowing repeated destinations.

packages/server-wallet/src/models/channel.ts
kerzhner: I agree with this point 2f47878de

packages/client-api-schema/src/data-types.ts
kerzhner: It is a catch all label. So anything that cannot be categorized as `ReadyToFund`, `Funded` or `Defunded`

packages/server-wallet/src/models/channel.ts
kerzhner: Good call 7fd576927

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
kerzhner: Unfortunately, it does not fit in one line 1a338b5c1

kerzhner:
> I'm also curious about why the direct funding status should be `'Uncategorized'` when the strategy is not `'Direct'`.

In the future, it would be nice to have a `ChannelState.fundingStatus` property (instead of `directFundingStatus`). For now, my strategy was to create the `directFundingStatus` that is only applicable for directly funded channels and to ignore all other funding types.

kerzhner:
> What would the funding status be for a channel with `fundingStrategy` as `"Ledger"` that has a supported prefund setup state?

@snario In this PR, if the channel funding strategy is not `Direct`, the `fundingStatus` in `ChannelResult` is always `Uncategorized`

packages/server-wallet/src/protocols/state.ts
```diff
@@ -92,9 +109,71 @@ export const toChannelResult = (channelState: ChannelState): ChannelResult => {
...
+  }
+
+  const outcome = supported?.outcome;
+  if (!supported || !outcome) {
+    return 'Uncategorized';
+  }
+
+  const {allocationItems, assetHolderAddress} = checkThat(outcome, isSimpleAllocation);
+
+  // Collapse all allocation items with my destination into one
```
tomclose, line 141: I think the contracts assume that there is only one allocation item per destination, and therefore ignore items after the first (@geoknee could you confirm?). The wallet should therefore do the same.

tomclose (commented):
This looks like a step in the right direction to me, and I think it's worth merging.

That said, I think we probably need a different approach here in the medium-term. In our current approach, to run a protocol we (a) fetch all the information we could possibly need about a channel, (b) do some pure calculations, (c) write some information back into the store.

The main issue with this appears to be (a) - that we need to gather all the information we could possibly need at any stage of the protocol, which in this case means gathering information about the funding status, regardless on whether we're in the funding stage. To make this efficient we therefore need to store this information on the channels table, which makes it hard to update the funding of multiple channels atomically (without risking deadlocks), which we need to do to prevent [funding flicker](https://www.notion.so/Server-wallet-updated-architecture-ad0372e611ac485aa16dc2c4df65a14f#55c858a3107f496594b1c51e32a70590).

I think the best solution here is to largely separate the funding from the channels, and request this data separately in the protocols, as described [here](https://www.notion.so/Server-wallet-updated-architecture-ad0372e611ac485aa16dc2c4df65a14f#450f5ed57b1c4c3f854429e15a4bce85). It would probably be good to revisit that document sometime soon, and think through it as a team - otherwise we just end up independently working around the problems it solves (cc @andrewgordstewart ).

packages/server-wallet/src/protocols/state.ts
geoknee: Yes that is the case for `claimAll` and `transfer` (and `claim` #2932). It happens that `transferAll` does not rely on such an assumption. 

It seems like a big assumption to make so I think we should a) be sure that we really need or want to make it and b) if we do, put it very prominently in the docs, advise `ForceMoveApps` to check it*, check it in the wallet when we do ledger funding etc.

*Similar to:
https://github.com/statechannels/statechannels/blob/76ad4a64a9f4bdc02f05afe121ea9cb5d3aaeca7/packages%2Fnitro-protocol%2Fcontracts%2Fexamples%2FSingleAssetPayments.sol#L61-L64

packages/server-wallet/src/models/channel.ts
kerzhner: Unit tests added https://github.com/statechannels/statechannels/pull/2950/files#diff-7d310f70f103b57a4ab31608a210eda379528970c7804f919e58389b440584d6

packages/server-wallet/src/models/channel.ts
kerzhner: `funding` should probably be a dictionary. https://github.com/statechannels/statechannels/issues/2969

packages/server-wallet/src/protocols/__test__/state.test.ts
```diff
@@ -0,0 +1,100 @@
...
+});
+const fullyTransferred = () => ({
+  amount: five,
+  transferredOut: [{toAddress: alice().destination, amount: five}],
+});
+
+describe('Only Alice funds channel', () =>
+  test.each`
+    supported        | fundingFn           | myParticipant      | fundingStrategy | result
+    ${preFundState0} | ${noFunding}        | ${participants[0]} | ${'Direct'}     | ${'ReadyToFund'}
```
andrewgordstewart, line 32: 😻

andrewgordstewart (approved):

packages/server-wallet/src/protocols/__test__/state.test.ts
andrewgordstewart: I'd like someone to try the alternative `test.each` api, where you call `test.each(testCases: TestCase[])`

It looks less pretty when it's static, but, it:
1. is type-safe. The type of `noFunding` is known within the test case callback.
2. is much easier to understand in git diffs

andrewgordstewart: Screen Shot 2020-11-20 at 2 47 17 PM

PR #2955, opened 2020-11-24
perf: remove unecessary abi.encode

geoknee:
- [x] will require updates to `wasm-utils` to get the tests passing.

### DIFF
--- a/packages/nitro-protocol/contracts/CountingApp.sol
+++ b/packages/nitro-protocol/contracts/CountingApp.sol
@@ -39,6 +39,7 @@ contract CountingApp is IForceMoveApp {
             appData(b.appData).counter == appData(a.appData).counter + 1,
             'CountingApp: Counter must be incremented'
         );
+        // Note this is gas inefficient, and inferior to _bytesEqual in use elsewhere
         require(
             keccak256(b.outcome) == keccak256(a.outcome),
             'CountingApp: Outcome must not change'
--- a/packages/nitro-protocol/contracts/ForceMove.sol
+++ b/packages/nitro-protocol/contracts/ForceMove.sol
@@ -107,7 +107,7 @@ contract ForceMove is IForceMove {
                 uint48(block.timestamp) + fixedPart.challengeDuration,
                 supportedStateHash,
                 challenger,
-                keccak256(abi.encode(variableParts[variableParts.length - 1].outcome))
+                keccak256(variableParts[variableParts.length - 1].outcome)
             )
         );
     }
@@ -135,8 +135,8 @@ contract ForceMove is IForceMove {
         bytes32 channelId = _getChannelId(fixedPart);
         (uint48 turnNumRecord, uint48 finalizesAt, ) = _getChannelStorage(channelId);
 
-        bytes32 challengeOutcomeHash = _hashOutcome(variablePartAB[0].outcome);
-        bytes32 responseOutcomeHash = _hashOutcome(variablePartAB[1].outcome);
+        bytes32 challengeOutcomeHash = keccak256(variablePartAB[0].outcome);
+        bytes32 responseOutcomeHash = keccak256(variablePartAB[1].outcome);
         bytes32 challengeStateHash = _hashState(
             turnNumRecord,
             isFinalAB[0],
@@ -268,7 +268,7 @@ contract ForceMove is IForceMove {
      * @param largestTurnNum The largest turn number of the submitted states; will overwrite the stored value of `turnNumRecord`.
      * @param fixedPart Data describing properties of the state channel that do not change with state updates.
      * @param appPartHash The keccak256 of the abi.encode of `(challengeDuration, appDefinition, appData)`. Applies to all states in the finalization proof.
-     * @param outcomeHash The keccak256 of the abi.encode of the `outcome`. Applies to all stats in the finalization proof.
+     * @param outcomeHash The keccak256 of the `outcome`. Applies to all stats in the finalization proof.
      * @param numStates The number of states in the finalization proof.
      * @param whoSignedWhat An array denoting which participant has signed which state: `participant[i]` signed the state with index `whoSignedWhat[i]`.
      * @param sigs An array of signatures that support the state with the `largestTurnNum`.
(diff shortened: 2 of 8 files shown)


packages/nitro-protocol/contracts/CountingApp.sol
```diff
@@ -39,6 +39,7 @@ contract CountingApp is ForceMoveApp {
             appData(b.appData).counter == appData(a.appData).counter + 1,
             'CountingApp: Counter must be incremented'
         );
+        // Note this is gas inefficient, and inferior to _bytesEqual in use elsewhere
         require(
             keccak256(b.outcome) == keccak256(a.outcome),
```
geoknee, line 44: This is not the best way to compare bytes, but this code is only used in testing so I just left a comment.

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -814,16 +814,6 @@ contract ForceMove is IForceMove {
...
-    /**
-     * @notice Computes the hash of a given outcome.
-     * @dev Computes the hash of a given outcome.
-     * @param outcome An outcome
-     * @return The outcomeHash
-     */
-    function _hashOutcome(bytes memory outcome) internal pure returns (bytes32) {
-        return keccak256(abi.encode(outcome));
-    }
-
```
geoknee, line 826: We were advised to reuse this function instead of sometimes directly doing `keccak256(abi.encode(...))`. However: I spotted that we don't actually need to `abi.encode` here at all. Removing that makes `_hashOutcome` an alias for `keccak256` and therefore redundant.

packages/nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -125,7 +125,7 @@ abstract contract IForceMove {
         uint8 numStates,
         uint8[] memory whoSignedWhat,
         Signature[] memory sigs
-    ) public virtual;
+    ) public virtual returns (bytes32);
```
geoknee, line 128: This saves a `require`, some `abi.encode` and a `keccak256` on our happy path. It takes a little bit more gas on the slightly-less happy path.

packages/wallet-core/src/serde/wire-format/example.ts
```diff
@@ -45,10 +45,9 @@ export const wireStateFormat: WireState = {
   ],
   turnNum: 1,
   signatures: [
-    '0xef7e226a43c52d4b8f7b14f13acdf9e75d871ea5c51235fbc4d538acf84c61c4727431f0cc83d0f566e222a21d35ae4d8d2a0dd4428cba7bf95bf7b3f11ad0c61c'
+    '0x59069c99dd52a7766e439ef9344e5cee7a51625ec366095cbd7bad66075a54e924ab48b08a683deff7e3642d8017178e55990b2073d169eec64a7fef60648a6a1b'
```
geoknee, line 48: I had to regenerate this signature manually, which was a pain.

packages/nitro-protocol/src/contract/outcome.ts
```diff
@@ -140,7 +140,7 @@ export type Outcome = AssetOutcome[];
 
 export function hashOutcome(outcome: Outcome): Bytes32 {
   const encodedOutcome = encodeOutcome(outcome);
-  return utils.keccak256(utils.defaultAbiCoder.encode(['bytes'], [encodedOutcome]));
```
geoknee, line 143: This is the change we would need to replicate in `native-utils`.

andrewgordstewart: I'll take a look today.

geoknee: https://github.com/statechannels/native-utils/pull/9

packages/nitro-protocol/contracts/interfaces/IForceMove.sol
geoknee: This got cherry-picked off this branch and made it's way into `master` #2980 .

geoknee:
test passing 🎉  and some decent gas savings e.g. for `challenge`:

master
```
It accepts for an open channel, and updates storage correctly, when the slot is empty, 1 state submitted:
105301 gas

It accepts for an open channel, and updates storage correctly, when the slot is empty, 3 states submitted:
147794 gas

It accepts for an open channel, and updates storage correctly, when the slot is not empty, 3 states submitted:
90313 gas

It accepts for an open channel, and updates storage correctly, when the slot is not empty, 3 states submitted:
90301 gas

It accepts for an open channel, and updates storage correctly, when the slot is not empty, 1 state submitted:
132782 gas

It accepts when a challenge is present, and updates storage correctly, when the turnNumRecord increases, 1 state:
91443 gas

It accepts when a challenge is present, and updates storage correctly, when the turnNumRecord increases, 3 states:
133912 gas

```

here
```
It accepts for an open channel, and updates storage correctly, when the slot is empty, 1 state submitted:
102947 gas

It accepts for an open channel, and updates storage correctly, when the slot is empty, 3 states submitted:
142953 gas

It accepts for an open channel, and updates storage correctly, when the slot is not empty, 3 states submitted:
87935 gas

It accepts for an open channel, and updates storage correctly, when the slot is not empty, 3 states submitted:
87935 gas

It accepts for an open channel, and updates storage correctly, when the slot is not empty, 1 state submitted:
127965 gas

It accepts when a challenge is present, and updates storage correctly, when the turnNumRecord increases, 1 state:
89053 gas

It accepts when a challenge is present, and updates storage correctly, when the turnNumRecord increases, 3 states:
129083 gas

```

packages/wallet-core/src/serde/wire-format/example.ts
andrewgordstewart: It's a nice touch to leave [instructions in a comment](https://github.com/statechannels/the-graph/blob/master/packages/statechannels-contracts/test/attestationApp.test.ts#L67-L74) on how to generate the signature. Thank @lalexgap for that!

andrewgordstewart (approved):

PR #2962, opened 2020-11-25
Small refactors to ObjectiveManager and Protocols

tomclose:
This is a first step towards simplifying the protocol logic. It makes a couple of (hopefully non-controversial) refactors to the ObjectiveManager and to the Protocols.

### DIFF
--- a/packages/server-wallet/src/objectives/objective-manager.ts
+++ b/packages/server-wallet/src/objectives/objective-manager.ts
@@ -1,5 +1,6 @@
 import {Logger} from 'pino';
 import {BN} from '@statechannels/wallet-core';
+import {Transaction} from 'objection';
 
 import {Bytes32} from '../type-aliases';
 import {
@@ -16,10 +17,11 @@ import * as ChannelState from '../protocols/state';
 import {Store} from '../wallet/store';
 import {LedgerRequest} from '../models/ledger-request';
 import {ChainServiceInterface} from '../chain-service';
-import {ProtocolAction} from '../protocols/actions';
+import {FundChannel, ProtocolAction, SignState, Withdraw} from '../protocols/actions';
 import {recordFunctionMetrics} from '../metrics';
 import {WalletResponse} from '../wallet/response-builder';
 import {Channel} from '../models/channel';
+import {DBObjective} from '../models/objective';
 
 import {ObjectiveManagerParams} from './types';
 import {CloseChannelObjective} from './close-channel';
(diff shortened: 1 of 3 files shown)


packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -54,87 +66,58 @@ export class ObjectiveManager {
...
-          protocolState = await OpenChannelProtocol.getOpenChannelProtocolState(
+          protocolState = await getOpenChannelProtocolState(
             this.store,
             objective.data.targetChannelId,
             tx
           );
-          executeProtocol = () =>
-            OpenChannelProtocol.protocol(protocolState as OpenChannelProtocol.ProtocolState);
+          nextAction = recordFunctionMetrics(
+            openChannelProtocol(protocolState as OpenChannelProtocolState),
```
geoknee, line 79: if `protocolState` is the return of `getOpenChannelProtocolState`, we shouldn't need to cast it on L79 (or anywhere).

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -54,87 +66,58 @@ export class ObjectiveManager {
...
-          protocolState = await CloseChannelProtocol.getCloseChannelProtocolState(
+          protocolState = await getCloseChannelProtocolState(
             this.store,
             objective.data.targetChannelId,
             tx
           );
-          executeProtocol = () =>
-            CloseChannelProtocol.protocol(protocolState as CloseChannelProtocol.ProtocolState);
+          nextAction = recordFunctionMetrics(
+            closeChannelProtocol(protocolState as OpenChannelProtocolState),
```
geoknee, line 89: Same comment here about casting, although I am surprised that this isn't an error?

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -1,21 +1,33 @@
...
+import {FundChannel, SignState, Withdraw} from '../protocols/actions';
 import {recordFunctionMetrics} from '../metrics';
 import {WalletResponse} from '../wallet/response-builder';
 import {Channel} from '../models/channel';
+import {DBObjective} from '../models/objective';
 
 import {ObjectiveManagerParams} from './types';
 import {CloseChannelObjective} from './close-channel';
 
+type SupportedProtocolState = OpenChannelProtocolState | CloseChannelProtocolState;
```
geoknee, line 29: Is there a way to avoid the naming collision of 'supported state'? Maybe just use `ProtocolState` as the union type , and put a comment that these are the only ones that are implemented so far.

packages/server-wallet/src/objectives/objective-manager.ts
```diff
@@ -54,87 +66,58 @@ export class ObjectiveManager {
...
         } else if (objective.type === 'CloseChannel') {
-          protocolState = await CloseChannelProtocol.getCloseChannelProtocolState(
+          protocolState = await getCloseChannelProtocolState(
             this.store,
             objective.data.targetChannelId,
             tx
           );
-          executeProtocol = () =>
-            CloseChannelProtocol.protocol(protocolState as CloseChannelProtocol.ProtocolState);
+          nextAction = recordFunctionMetrics(
```
geoknee, line 88: Having `recordFunctionMetrics` sprinkled everywhere seems pretty bad for readability. This PR increases the number of these wrappers. Is there a strategy we can employ to improve this? Hide them inside the imported functions implementations?

geoknee (approved):

packages/server-wallet/src/objectives/objective-manager.ts
tomclose: Yeah, I looked at this. Unfortunately line 70 explicitly declares `protocolState` to be the union of the two possible return types, so we need to cast. It needs to do this because it's used later in the function. This issue will go away as I continue to refactor though.

packages/server-wallet/src/objectives/objective-manager.ts
tomclose: I'm intending to remove this type completely in future refactors.

packages/server-wallet/src/objectives/objective-manager.ts
tomclose: It's not clear to me whether we even still need these. In this case it's wrapping a pure function, which as far as I can tell doesn't do hashing or signing, so I'm not sure what the point it. Will ask the others when they're up.

snario (commented):
👍

PR #2966, opened 2020-11-25
perf: keep FixedPart in calldata

geoknee:
I tried doing this for other parameters, but ran into `Stack too deep` errors. 🌰


geoknee:
Closing as I am actually seeing worse gas efficiency with this change, and a failing test to 🥾

PR #2968, opened 2020-11-25
[server-wallet] Prevent closed ledger channels from being able to fund channels

snario:
Presently a ledger channel can be used to fund a channel after the ledger channel has already been closed. This adds a check to prevent `createChannel` from allowing it to be specified and a check in the protocol to ensure the ledger being used to fund a channel is actually running before proceeding with ledger funding.

### DIFF
--- a/packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
+++ b/packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
@@ -342,6 +342,35 @@ describe('Funding a single channel with 50% of ledger funds', () => {
   });
 });
 
+describe('Closing a ledger channel and preventing it from being used again', () => {
+  let ledgerChannelId: Bytes32;
+
+  afterAll(async () => {
+    await a.dbAdmin().truncateDB(['channels', 'ledger_requests']);
+    await b.dbAdmin().truncateDB(['channels', 'ledger_requests']);
+  });
+
+  it('can close the ledger channel', async () => {
+    ledgerChannelId = await createLedgerChannel(10, 10);
+    const {outbox} = await a.closeChannel({channelId: ledgerChannelId});
+    const {outbox: close} = await b.pushMessage(getPayloadFor(participantB.participantId, outbox));
+    await exchangeMessagesBetweenAandB([close], []);
+    const {channelResults} = await a.getChannels();
+    await expect(b.getChannels()).resolves.toEqual({channelResults, outbox: []});
+    const ledger = getChannelResultFor(ledgerChannelId, channelResults);
+    expect(ledger).toMatchObject({
+      turnNum: 4,
+      status: 'closed',
+    });
+  });
+
+  it('fails to fund a new channel with this ledger', async () => {
+    const params = testCreateChannelParams(10, 10, ledgerChannelId);
+    await expect(a.createChannel(params)).rejects.toThrow(/already been closed/);
+    await expect(a.createChannels(params, 1)).rejects.toThrow(/already been closed/);
+  });
+});
+
 describe('Funding multiple channels syncronously (in bulk)', () => {
   const N = 4;
   let ledgerChannelId: Bytes32;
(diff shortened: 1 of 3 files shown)


packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
```diff
@@ -342,6 +342,35 @@ describe('Funding a single channel with 50% of ledger funds', () => {
...
+    await expect(b.getChannels()).resolves.toEqual({channelResults, outbox: []});
+    const ledger = getChannelResultFor(ledgerChannelId, channelResults);
+    expect(ledger).toMatchObject({
+      turnNum: 4,
+      status: 'closed',
+    });
+  });
+
+  it('fails to fund a new channel with this ledger', async () => {
+    const params = testCreateChannelParams(10, 10, ledgerChannelId);
```
andrewgordstewart, line 368: This test seems to require that the previous test case on line 353 runs and succeeds

packages/server-wallet/src/wallet/store.ts
```diff
@@ -697,6 +699,13 @@ async function createChannel(
   fundingLedgerChannelId: Bytes32 | undefined,
   txOrKnex: TransactionOrKnex
 ): Promise<Channel> {
+  if (fundingLedgerChannelId) {
+    const ledger = await Channel.forId(fundingLedgerChannelId, txOrKnex);
+    if (!ledger) throw new StoreError(StoreError.reasons.invalidFundingLedgerChannelId);
+    if (ledger.supported && ledger.supported.isFinal)
```
andrewgordstewart, line 705: It is also possible for the ledger channel to be _closing_, but `ledger.supported.isFinal` to be false.

andrewgordstewart (changes requested):
I think there's two issues with this change, but otherwise, good catch!

packages/server-wallet/src/wallet/store.ts
```diff
@@ -697,6 +699,13 @@ async function createChannel(
   fundingLedgerChannelId: Bytes32 | undefined,
   txOrKnex: TransactionOrKnex
 ): Promise<Channel> {
+  if (fundingLedgerChannelId) {
+    const ledger = await Channel.forId(fundingLedgerChannelId, txOrKnex);
+    if (!ledger) throw new StoreError(StoreError.reasons.invalidFundingLedgerChannelId);
+    if (ledger.supported && _.some(ledger.support, 'isFinal'))
```
geoknee, line 705: Not sure I understand this. Is the channel `closing` as soon as I have received an `isFinal` state? Does that state need to be in `ledger.support`? Or should we check against `ledger.vars` ?

geoknee: Should we have a `.status()` getter on the `Channel` model, so that we don't have to keep revisiting channel statuses and recomputing them ad hoc?

geoknee: Or we use a `ledger.isRunning` method?

geoknee (approved):
Approving as I think this is already an important improvement. Would like to see the `isRunning` logic cleared up, though.

packages/server-wallet/src/wallet/store.ts
snario: This is not in the protocol code, just in the `createChannel` method on the `Store`. In theory it's possible you might be closing the ledger channel (proposed an `isFinal` state) and then later decide that you don't want to close it, but this logic ignores that. The logic is "throw an error if trying to create a channel funded by a ledger channel with any `isFinal` state in the support). The user could always call `createChannel` again later. For now, this is enough to prevent an issue and is at worst _too strict_ versus _too liberal_

geoknee: I think it might be too liberal, still. 

If I have proposed an `isFinal` state that is currently unsupported, my counterparty can unilaterally `conclude` the channel with that state at any point in the future. Deciding later to not close it is not an option, I think -- so the state should instantly and irrecoverably become `!running`.

 So I think it is unsafe to fund any new application channels with a ledger channel once I have signed an `isFInal` state.

PR #2972, opened 2020-11-25
test: add off-chain tests against max outcome size, and on-chain test for challenging with max outcome size

geoknee:
Happy: by checking the tx size off chain, ganache was happy to accept a `challenge` transaction under 128KB without gorging on memory. 

Sad: tried rolling out the same thing for `pushOutcome`. Can get the same number of allocationItems in under 128KB but still seeing an out-of-memory error. 

- exports new constants
- tidies up test helpers
- adds comments around block gas limits


andrewgordstewart:
Neat!

tomclose (approved):

PR #2973, opened 2020-11-26
Config clean up

lalexgap:
Some minor changes to clean up the config.

- Removes some unused assetholder address config
- Exported config types from `ServerWallet` package
- Accept `chainId` as a number.

### DIFF
--- a/packages/server-wallet/src/config/types.ts
+++ b/packages/server-wallet/src/config/types.ts
@@ -84,9 +84,6 @@ export type IncomingServerWalletConfig = RequiredServerWalletConfig &
  * Various network configuration options
  */
 export type NetworkConfiguration = {
-  erc20Address: string;
-  ethAssetHolderAddress: string;
-  erc20AssetHolderAddress: string;
   rpcEndpoint?: string;
   chainNetworkID: string;
 };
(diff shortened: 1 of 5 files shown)


andrewgordstewart (approved):

andrewgordstewart:
I think exporting the types is also a "fix" btw.

packages/server-wallet/src/config/types.ts
```diff
@@ -84,9 +84,6 @@ export type IncomingServerWalletConfig = RequiredServerWalletConfig &
  * Various network configuration options
  */
 export type NetworkConfiguration = {
-  erc20Address: string;
-  ethAssetHolderAddress: string;
-  erc20AssetHolderAddress: string;
```
snario, line 89: Keep in mind these values are still being read from the environment

https://github.com/statechannels/statechannels/blob/a67effab94da3301f5dbe21e32e7f8ba8b9afd48/packages/server-wallet/src/chain-service/chain-service.ts#L28-L33

```typescript
const ethAssetHolderAddress = makeAddress(
  process.env.ETH_ASSET_HOLDER_ADDRESS || ethers.constants.AddressZero
);
const nitroAdjudicatorAddress = makeAddress(
  process.env.NITRO_ADJUDICATOR_ADDRESS! || ethers.constants.AddressZero
);
```

PR #2981, opened 2020-11-27
fix: improve semantic meaning of input parameters via runtime validation

geoknee:
We are not able to use the type system to constrain the length of dynamically sized arrays; yet we _do_ use the type system  to constrain `numStates` and `whoSignedWhat` entries to be `uint8`s. 

Adding a runtime validation check makes it clear that we do not support more than 255 participants in a channel, and also (therefore) that the chain ought never to be supplied with more than 255 states (meaning `variableParts.length` and `whoSignedWhat[i] \forall i` are also always less than or equal to 255).

### DIFF
--- a/packages/nitro-protocol/contracts/ForceMove.sol
+++ b/packages/nitro-protocol/contracts/ForceMove.sol
@@ -54,7 +54,7 @@ contract ForceMove is IForceMove {
         Signature memory challengerSig
     ) external override {
         // input type validation
-        _requireValidInput(
+        requireValidInput(
             fixedPart.participants.length,
             variableParts.length,
             sigs.length,
@@ -205,7 +205,7 @@ contract ForceMove is IForceMove {
         uint8[] memory whoSignedWhat
     ) external override {
         // input type validation
-        _requireValidInput(
+        requireValidInput(
             fixedPart.participants.length,
             variableParts.length,
             sigs.length,
@@ -286,7 +286,7 @@ contract ForceMove is IForceMove {
         _requireChannelNotFinalized(channelId);
 
         // input type validation
-        _requireValidInput(
+        requireValidInput(
             fixedPart.participants.length,
             numStates,
             sigs.length,
@@ -913,12 +913,12 @@ contract ForceMove is IForceMove {
      * @param numSigs Number of signatures submitted
      * @param numWhoSignedWhats whoSignedWhat.length
      */
-    function _requireValidInput(
+    function requireValidInput(
         uint256 numParticipants,
         uint256 numStates,
         uint256 numSigs,
         uint256 numWhoSignedWhats
-    ) internal pure {
+    ) public pure returns (bool) {
         require(
             (numParticipants >= numStates) && (numStates > 0),
             'ForceMove | You must submit at least one but no more than numParticipants states'
@@ -928,5 +928,6 @@ contract ForceMove is IForceMove {
             'ForceMove | You must provide exactly one signature per participant, and assert who signed what for all participants'
         );
         require(numParticipants < type(uint8).max, 'ForceMove | Too many participants!');
+        return true;
     }
 }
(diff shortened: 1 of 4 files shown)


packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -53,13 +53,16 @@ contract ForceMove is IForceMove {
         uint8[] memory whoSignedWhat,
         Signature memory challengerSig
     ) external override {
-        bytes32 channelId = _getChannelId(fixedPart);
-
-        require(
-            (fixedPart.participants.length >= variableParts.length) && (variableParts.length > 0),
-            'challenge | You must submit at least one but no more than numParticipants states'
+        // input type validation
+        _requireValidInput(
```
andrewgordstewart, line 57: Can we consider an optional app definition of a valid state?

geoknee: couldn't that be checked inside `ForceMoveApp.ValidTransition`?

andrewgordstewart: er yes, and it _should_ be (if we make it a thing)

I think it's related to this change, but perhaps not this line?

geoknee: My point is, can't we leave that decision up to the `ForceMoveApp` developer? Thinking of it that way, it already *is* optional. I'm not sure if you are suggesting a change to *our* contracts?

andrewgordstewart: Unfortunately not, in the case that a fully-supported state is provided.

geoknee: We could recommend that a `ForceMoveApp` expose a public, pure `validState` function (that can be used internally by `validTransition` and also externally). Then, clients can optionally call or emulate `validState` before countersigning a state, particularly a state which is not a `validTransition` from any known state.

I think we aren't far off from having such functions in our examples already: 
https://github.com/statechannels/statechannels/blob/172b4529352d4af2dc9fcf8c1d7438b7de3c83f1/packages/nitro-protocol/contracts/CountingApp.sol#L16-L23

Is there are compelling reason for this check to be performed on chain? It seem pretty "edge case", to me. `checkpoint` and `conclude` are about getting a `turnNum` or `outcome` on chain. `challenge` is maybe the problematic one, since it is currently possible to get a bad state on chain which causes `validTransition` to do unexpected things during `respond`*. Is that the kind of thing you are hoping to prevent?

*of course only by all parties agreeing, which the off-chain check I am proposing would help to prevent.

packages/nitro-protocol/test/contracts/ForceMove/requireValidInput.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  { numParticipants: 256, numStates: 256, numSigs: 256, numWhoSignedWhats: 256 },
+  { numParticipants: 1, numStates: 0, numSigs: 1, numWhoSignedWhats: 1 },
+  { numParticipants: 2, numStates: 3, numSigs: 3, numWhoSignedWhats: 3 },
+  { numParticipants: 2, numStates: 3, numSigs: 2, numWhoSignedWhats: 2 },
+];
+
+describe('requireValidInput', () => {
+  it.each(goodTable)('Valid input: %o', checkGoodCaseAgainstContract);
+  it.each(badTable)('Invalid input: %o', checkBadCaseAgainstContract);
+});
```
geoknee, line 66: @andrewgordstewart this if for you

packages/nitro-protocol/test/contracts/ForceMove/requireValidInput.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    ForceMove.requireValidInput(
+      testCase.numParticipants,
+      testCase.numStates,
+      testCase.numSigs,
+      testCase.numWhoSignedWhats
+    )
+  ).rejects.toThrow(/ForceMove |/);
+}
+
+// prettier-ignore
```
andrewgordstewart, line 46: This is awesome. Good tip on `prettier-ignore`...

andrewgordstewart (approved):

PR #2984, opened 2020-11-30
emit finalizesAt in Conclude event

### LINKED ISSUE #2983: emit block.timestamp in `Conclude` event
Most often we use the `concludePushOutcomeAndTransferAll` path. But we must support a separate `conclude` (or `challenge` and timeout) and `pushOutcome`.

The `ChallengeRegistered` event is rich with information to allow responding or pushing the outcome. The `Concluded` event is rather sparse, only featuring the `channelId`
https://docs.statechannels.org/contract-api/natspec/ForceMove#events

When working through this path manually recently, I resorted to doing things like this:

```ts
const finalizesAt = (await provider.getBlock(concludeReceipt.blockNumber)).timestamp
```

We could make the experience nicer by emitting the arguments for `pushOutcome` https://docs.statechannels.org/contract-api/natspec/NitroAdjudicator#pushoutcome in the `Concluded` event. 

Note, however, that many of these arguments take a default value when the channel is concluded, so do not need to be emitted. 

https://github.com/statechannels/statechannels/blob/30f7226d6eac88eb28d1e5a4da4ce60842c70533/packages%2Fnitro-protocol%2Fcontracts%2FForceMove.sol#L311-L315

geoknee:
Closes #2983 

- enhances `createPushOutcomeTransaction` to deal with concluded channels
- adds a test case to cover a concluded channel.

### DIFF
--- a/packages/nitro-protocol/contracts/ForceMove.sol
+++ b/packages/nitro-protocol/contracts/ForceMove.sol
@@ -224,7 +224,7 @@ contract ForceMove is IForceMove {
      * @param largestTurnNum The largest turn number of the submitted states; will overwrite the stored value of `turnNumRecord`.
      * @param fixedPart Data describing properties of the state channel that do not change with state updates.
      * @param appPartHash The keccak256 of the abi.encode of `(challengeDuration, appDefinition, appData)`. Applies to all states in the finalization proof.
-     * @param outcomeHash The keccak256 of the abi.encode of the `outcome`. Applies to all stats in the finalization proof.
+     * @param outcomeHash The keccak256 of the abi.encode of the `outcome`. Applies to all states in the finalization proof.
      * @param numStates The number of states in the finalization proof.
      * @param whoSignedWhat An array denoting which participant has signed which state: `participant[i]` signed the state with index `whoSignedWhat[i]`.
      * @param sigs An array of signatures that support the state with the `largestTurnNum`.
@@ -312,7 +312,7 @@ contract ForceMove is IForceMove {
         channelStorageHashes[channelId] = _hashChannelData(
             ChannelData(0, uint48(block.timestamp), bytes32(0), address(0), outcomeHash)
         );
-        emit Concluded(channelId);
+        emit Concluded(channelId, uint48(block.timestamp));
     }
 
     // Internal methods:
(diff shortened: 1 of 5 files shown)


packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -312,7 +312,7 @@ contract ForceMove is IForceMove {
         channelStorageHashes[channelId] = _hashChannelData(
             ChannelData(0, uint48(block.timestamp), bytes32(0), address(0), outcomeHash)
         );
-        emit Concluded(channelId);
+        emit Concluded(channelId, uint48(block.timestamp));
```
geoknee, line 315: The only other non-null variable we could emit is `outcomeHash`. But that isn't useful unless you have the pre-image.

andrewgordstewart: Can you output the preimage? How much gas does that cost?

andrewgordstewart: How hard a "no" is that? Is it reasonable to change the function signature to allow this?

andrewgordstewart: (the point is to wonder: should our short term strategy be to make it as easy as possible to manage finalized outcomes?)

andrewgordstewart (approved):

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -312,7 +312,7 @@ contract ForceMove is IForceMove {
         channelStorageHashes[channelId] = _hashChannelData(
             ChannelData(0, uint48(block.timestamp), bytes32(0), address(0), outcomeHash)
```
geoknee, line 313: An alternative would be to set `finalizesAt` to 0 during a conclude. Then it's a null value like most of the others and would not need to be emitted.

geoknee: ... this would be a bad idea, though, since we use`0`as a magic value for `finalizesAt` indicating that a channel is open

https://github.com/statechannels/statechannels/blob/30f7226d6eac88eb28d1e5a4da4ce60842c70533/packages%2Fnitro-protocol%2Fcontracts%2FForceMove.sol#L735-L746

PR #2991, opened 2020-11-30
docs: rip out contract-api section

geoknee:
- upgrade api-extractor
- disbable solidoc in nitro-protocol
- delete sidebar and navbar entries

this is a fallback option in case we cannot get solidity -> docs pipeline working again. 

Not to be merged yet, since it could end up overwriting the existing docs website.


geoknee:
in favour of #3005

PR #2995, opened 2020-12-01
Open channel protocol test

tomclose:
This PR changes the style of test for the open channel protocol. It also fixes a pretty serious issue that was exposed by the change: that once it was a participant's turn to deposit, the protocol would deposit repeatedly, stopping only when the channel was funded (e.g. in an `{a: 5, b: 5}` channel, A would deposit 5 multiple times, possibly funding b's stake ... :/ ).

The open channel protocol test has changed from being a test of the pure protocol function, to encompassing more of the system. It (a) loads state into the store, (b) runs the protocol, and then (c) asserts on side-effects. This style will make it easier to refactor the protocols to make the queries more targeted and efficient.

### DIFF
--- a/packages/server-wallet/src/protocols/__test__/open-channel.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/open-channel.test.ts
@@ -31,50 +31,57 @@ afterEach(async () => await store.dbAdmin().truncateDB());
 describe(`pre-fund-setup phase`, () => {
   describe(`as participant 0`, () => {
     it(`takes no action if I haven't received my opponent's state and I've already signed`, async () => {
-      await testCase({participant: 0, statesHeld: [0], statesToSign: []});
+      const obj = await setup({participant: 0, statesHeld: [0]});
+      await crankAndAssert(obj, {statesToSign: []});
     });
     it(`signs the pre-fund-setup if I have my opponent's state (and deposits)`, async () => {
-      await testCase({participant: 0, statesHeld: [1], statesToSign: [0], fundsToDeposit: 5});
+      const obj = await setup({participant: 0, statesHeld: [1], totalFunds: 0});
+      await crankAndAssert(obj, {statesToSign: [0], fundsToDeposit: 5});
     });
   });
   describe(`as participant 1`, () => {
     it(`takes no action if I haven't received my opponent's state and I've already signed`, async () => {
-      await testCase({participant: 1, statesHeld: [1], statesToSign: []});
+      const obj = await setup({participant: 1, statesHeld: [1], totalFunds: 0});
+      await crankAndAssert(obj, {statesToSign: []});
     });
     it(`signs the pre-fund-setup if I have my opponent's state`, async () => {
-      await testCase({participant: 1, statesHeld: [0], statesToSign: [1]});
+      const obj = await setup({participant: 1, statesHeld: [0], totalFunds: 0});
+      await crankAndAssert(obj, {statesToSign: [1]});
     });
   });
 });
 
 describe(`funding phase`, () => {
-  const statesHeld = [0, 1];
-
   describe(`as participant 0`, () => {
     it(`submits the transaction if it is my turn`, async () => {
-      const o = await testCase({participant: 0, statesHeld, totalFunds: 0, fundsToDeposit: 5});
+      const obj = await setup({participant: 0, statesHeld: [0, 1], totalFunds: 0});
+      await crankAndAssert(obj, {fundsToDeposit: 5});
       // and then doesn't submit it a second time
-      await crankAndAssert(o, {fundsToDeposit: 0});
+      await crankAndAssert(obj, {fundsToDeposit: 0});
     });
     it(`submits a top-up if there's a partial deposit`, async () => {
-      const o = await testCase({participant: 0, statesHeld, totalFunds: 2, fundsToDeposit: 3});
+      const obj = await setup({participant: 0, statesHeld: [0, 1], totalFunds: 2});
+      await crankAndAssert(obj, {fundsToDeposit: 3});
       // and then doesn't submit it a second time
-      await crankAndAssert(o, {fundsToDeposit: 0});
+      await crankAndAssert(obj, {fundsToDeposit: 0});
     });
     it(`does nothing if I've already deposited`, async () => {
-      await testCase({participant: 0, statesHeld, totalFunds: 5, fundsToDeposit: 0});
+      const obj = await setup({participant: 0, statesHeld: [0, 1], totalFunds: 5});
+      await crankAndAssert(obj, {fundsToDeposit: 0});
     });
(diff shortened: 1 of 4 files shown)


packages/server-wallet/src/protocols/__test__/open-channel.test.ts
```diff
@@ -1,95 +1,210 @@
...
-      directFundingStatus,
-    },
+describe(`funding phase`, () => {
+  const statesHeld = [0, 1];
+
+  describe(`as participant 0`, () => {
+    it(`submits the transaction if it is my turn`, async () => {
+      const o = await testCase({participant: 0, statesHeld, totalFunds: 0, fundsToDeposit: 5});
+      // and then doesn't submit it a second time
+      await testCrankOutcome(o, {fundsToDeposit: 0});
```
geoknee, line 60: Looks like `fundsToDeposit` defaults to `0`, do you need to specify it again here?

It's not immediately obvious that awaiting `testCrankOutcome` with a certain input is equivalent to asserting that a transaction is not submitted. a second time.

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
```diff
@@ -1,95 +1,210 @@
...
+    });
+    it(`submits a top-up if there's a partial deposit`, async () => {
+      const o = await testCase({participant: 1, statesHeld, totalFunds: 7, fundsToDeposit: 3});
+      // and then doesn't submit it a second time
+      await testCrankOutcome(o, {fundsToDeposit: 0});
+    });
+    it.skip(`does nothing if I've already deposited`, async () => {
+      // todo: this fails, as it sends the post-fund-setup (as it probably should)
```
geoknee, line 83: I'm confused about whether this test should exist, be fixed, be skipped, or what?

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
```diff
@@ -1,95 +1,210 @@
...
+    });
+    it(`signs the post-fund-setup (and completes objective), when my opponent has signed`, async () => {
+      await testCase({
+        participant: 0,
+        statesHeld: [0, 1, 3],
+        totalFunds: 10,
+        statesToSign: [2],
+        completesObj: true,
+      });
+      // todo: and marks the objective as complete
```
geoknee, line 105: you have many `todo`s here. I think @snario prefers them to be `TODO`s. 

When should these be todone?

packages/server-wallet/src/protocols/open-channel.ts
```diff
@@ -234,18 +234,24 @@ const requestFundChannelIfMyTurn = ({app}: ProtocolState): FundChannel | false =
   );
 
   const currentFunding = app.funding(assetHolderAddress).amount;
+  const allocationsBeforeMe = _.takeWhile(allocationItems, a => a.destination !== myDestination);
```
geoknee, line 237: nice use of `takeWhile.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -664,6 +669,10 @@ export class Store {
   async nextNonce(signingAddresses: Address[]): Promise<number> {
     return await Nonce.next(this.knex, signingAddresses);
   }
+
+  dbAdmin(): DBAdmin {
+    return new DBAdmin(this.knex);
+  }
```
geoknee, line 675: This is replicated in the `Wallet` class, I believe? Do we want to keep both?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -162,6 +163,10 @@ export class Store {
     });
   }
 
+  async transaction<T>(callback: (tx: Transaction) => T): Promise<T> {
+    return this.knex.transaction(async tx => callback(tx));
+  }
```
geoknee, line 168: Is there a chance this is unrelated to this PR, and sneaking in?

packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
```diff
@@ -0,0 +1,135 @@
...
+    this.startBals = [aBal, bBal];
+  }
+
+  /**
+   * Gives the nth state in the history, signed by the correct participant
+   *
+   * @param n - the index of the state in the history
+   *
+   * Note - in cases where participants double-sign the same states, n might _not_
+   * be the turnNum
```
geoknee, line 58: Not sure I understand this Note?

geoknee (approved):
There's a lot of repetition in your test suite, which makes it pretty ripe for some `it.each()`, in my opinion. But these things are a matter of taste and there can be good reason for more verbosity. 

Anyway: great that it caught a critical bug!

>  (e.g. in an {a: b, b: 5} channel, A would deposit 5 multiple times, possibly funding b's stake ... :/ ).

Is there a typo in this example that you give?

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
tomclose: This one has actually been done ✂️

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
tomclose: Me too. I'll remove it.

packages/server-wallet/src/wallet/store.ts
tomclose: This is used in the test setup. We'll also need it in other places going forward though.

packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
tomclose: What I mean is if we're following the "A signs turnNum 0, B signs turnNum 0" school of the prefundsetup, then `state(0)` would be turnNum 0 with A's signature, and `state(1)` would be turnNum 0 with B's signature (i.e. what B sends back to A).

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
tomclose: I wanted to make it more obvious what I was checking for here. You're right though that it still isn't very obvious. Maybe `crankAndAssert` would be a better name?

packages/server-wallet/src/wallet/store.ts
tomclose: I think we need it here. Thought it wouldn't be too bad to keep it in both places for the time being.

packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
geoknee: ok it makes sense to me now

PR #2996, opened 2020-12-01
Logging and configuration improvements to ChainService

snario:
Done while trying to get the chain service working in production

### DIFF
--- a/packages/server-wallet/src/chain-service/chain-service.ts
+++ b/packages/server-wallet/src/chain-service/chain-service.ts
@@ -60,8 +60,8 @@ export type ChainServiceConfig = {
   provider: string;
   pk: string;
   pollingInterval?: number;
-  logger: Logger;
   allowanceMode: AllowanceMode;
+  logger?: Logger;
 };
 
 export class ChainService implements ChainServiceInterface {
@@ -77,7 +77,9 @@ export class ChainService implements ChainServiceInterface {
   private transactionQueue = new PQueue({concurrency: 1});
 
   constructor({provider, pk, pollingInterval, logger, allowanceMode}: ChainServiceConfig) {
-    this.logger = logger ?? createLogger(defaultTestConfig());
+    this.logger = logger
+      ? logger.child({module: 'ChainService'})
+      : createLogger(defaultTestConfig());
     this.provider = new providers.JsonRpcProvider(provider);
     this.allowanceMode = allowanceMode;
     if (provider.includes('0.0.0.0') || provider.includes('localhost')) {
@@ -138,16 +140,20 @@ export class ChainService implements ChainServiceInterface {
   ): Promise<providers.TransactionResponse> {
     return this.transactionQueue.add(async () => {
       try {
+        this.logger.debug({...transactionRequest}, 'Submitting transaction to the blockchain');
         return await this.ethWallet.sendTransaction(transactionRequest);
-      } catch (e) {
+      } catch (err) {
         // https://github.com/ethers-io/ethers.js/issues/972
         this.ethWallet.incrementTransactionCount(-1);
-        throw e;
+        this.logger.error({err}, 'Transaction failed');
+        throw err;
       }
     });
   }
 
   async fundChannel(arg: FundChannelArg): Promise<providers.TransactionResponse> {
+    this.logger.info({...arg}, 'Attempting to fund channel');
+
     const assetHolderAddress = arg.assetHolderAddress;
     const isEthFunding = isEthAssetHolder(assetHolderAddress);
 
@@ -163,7 +169,21 @@ export class ChainService implements ChainServiceInterface {
       to: assetHolderAddress,
       value: isEthFunding ? arg.amount : undefined,
     };
-    return this.sendTransaction(depositRequest);
+
+    const tx = await this.sendTransaction(depositRequest);
+
+    if (!tx) throw new Error('fundChannel failed to submit deposit transaction');
+
+    this.logger.info(
+      {
+        channelId: arg.channelId,
+        assetHolderAddress,
+        tx: tx.hash,
+      },
+      'Finished funding channel'
+    );
+
+    return tx;
   }
 
   async concludeAndWithdraw(
@@ -328,7 +348,20 @@ export class ChainService implements ChainServiceInterface {
           to: tokenContract.address,
         };
 
-        await this.sendTransaction(increaseAllowanceRequest);
+        const tx = await this.sendTransaction(increaseAllowanceRequest);
+
+        if (!tx) throw new Error('fundChannel failed to increase allowance');
+
+        this.logger.info(
+          {
+            assetHolderAddress,
+            tokenAddress,
+            amount,
+            tx: tx.hash,
+          },
+          'Increased asset holder token allowance'
+        );
+
         break;
       }
       case 'MaxUint': {
@@ -347,7 +380,20 @@ export class ChainService implements ChainServiceInterface {
             to: tokenContract.address,
           };
 
-          await this.sendTransaction(approveAllowanceRequest);
+          const tx = await this.sendTransaction(approveAllowanceRequest);
+
+          if (!tx) throw new Error('fundChannel failed to call approve');
+
+          this.logger.info(
+            {
+              assetHolderAddress,
+              tokenAddress,
+              amount,
+              tx: tx.hash,
+            },
+            'Approved maximum amount of asset holder spending'
+          );
+
           break;
         }
       }
--- a/packages/server-wallet/src/chain-service/chain-service.ts
+++ b/packages/server-wallet/src/chain-service/chain-service.ts
@@ -407,16 +407,6 @@ export class ChainService implements ChainServiceInterface {
           const tx = await this.sendTransaction(approveAllowanceRequest);
 
           this.logger.info(
-            {
-              assetHolderAddress,
-              tokenAddress,
-              amount,
-              tx: tx.hash,
-            },
-            'Submitted transaction to approve maximum amount of asset holder spending'
-          );
-
-          this.logger.info(
             {tx: tx.hash},
             'Transaction to approve maximum amount of asset holder spending successfully submitted'
           );
--- a/packages/server-wallet/src/chain-service/chain-service.ts
+++ b/packages/server-wallet/src/chain-service/chain-service.ts
@@ -53,6 +53,12 @@ function isEthAssetHolder(address: Address): boolean {
 
 const blockConfirmations = 5;
 
+export type ChainServiceConfig = {
+  provider: string;
+  pk: string;
+  pollingInterval?: number;
+};
+
 export class ChainService implements ChainServiceInterface {
   private readonly ethWallet: NonceManager;
   private provider: providers.JsonRpcProvider;
--- a/packages/server-wallet/src/chain-service/chain-service.ts
+++ b/packages/server-wallet/src/chain-service/chain-service.ts
@@ -186,7 +186,7 @@ export class ChainService implements ChainServiceInterface {
   async concludeAndWithdraw(
     finalizationProof: SignedState[]
   ): Promise<providers.TransactionResponse | void> {
-    if (finalizationProof.length === 0) {
+    if (!finalizationProof.length) {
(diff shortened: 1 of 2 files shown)


packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -62,9 +64,19 @@ export class ChainService implements ChainServiceInterface {
...
-  constructor({provider, pk, allowanceMode, pollingInterval}: ChainServiceArgs) {
+  constructor({
+    provider,
+    pk,
+    pollingInterval,
+    logger,
+    blockConfirmations,
+    allowanceMode,
+  }: ChainServiceArgs) {
+    this.blockConfirmations = blockConfirmations || 5;
```
kerzhner, line 78: 0 is a valid blockConfirmations value.
```suggestion
    this.blockConfirmations = blockConfirmations ?? 5;
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -120,19 +132,25 @@ export class ChainService implements ChainServiceInterface {
...
     return this.transactionQueue.add(async () => {
       try {
+        this.logger.debug({...transactionRequest}, 'Submitting transaction to the blockchain');
         return await this.ethWallet.sendTransaction(transactionRequest);
-      } catch (e) {
+      } catch (err) {
         // https://github.com/ethers-io/ethers.js/issues/972
         this.ethWallet.incrementTransactionCount(-1);
-        throw e;
+        this.logger.error({err}, 'Transaction failed');
```
kerzhner, line 145:
```suggestion
        this.logger.error({err}, 'Transaction submission failed');
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -120,19 +132,25 @@ export class ChainService implements ChainServiceInterface {
     return obs;
   }
 
-  private async sendTransaction(transactionRequest: providers.TransactionRequest) {
+  private async sendTransaction(
+    transactionRequest: providers.TransactionRequest
+  ): Promise<providers.TransactionResponse | undefined> {
```
kerzhner, line 137: Why should this function return a Promise<undefined>?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -120,19 +132,25 @@ export class ChainService implements ChainServiceInterface {
...
       try {
+        this.logger.debug({...transactionRequest}, 'Submitting transaction to the blockchain');
         return await this.ethWallet.sendTransaction(transactionRequest);
-      } catch (e) {
+      } catch (err) {
         // https://github.com/ethers-io/ethers.js/issues/972
         this.ethWallet.incrementTransactionCount(-1);
-        throw e;
+        this.logger.error({err}, 'Transaction failed');
+        return;
```
kerzhner, line 146: What is the reason for this `return`?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -148,7 +166,21 @@ export class ChainService implements ChainServiceInterface {
       to: assetHolderAddress,
       value: isEthFunding ? arg.amount : undefined,
     };
-    return this.sendTransaction(depositRequest);
+
+    const tx = await this.sendTransaction(depositRequest);
+
+    if (!tx) throw new Error('fundChannel failed to submit deposit transaction');
```
kerzhner, line 172: It seems wrong to use `undefined` return values to propagate errors. `sendTransaction` throws an error. Is it possible to catch the error and throw a new error here?

Also, by discarding the original error, we are likely losing valuable debugging information.

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -313,7 +345,22 @@ export class ChainService implements ChainServiceInterface {
           to: tokenContract.address,
         };
 
-        await this.sendTransaction(increaseAllowanceRequest);
+        const tx = await this.sendTransaction(increaseAllowanceRequest);
+
+        if (!tx) throw new Error('fundChannel failed to increase allowance');
```
kerzhner, line 350: Same comment as above about mixing return values and errors.

kerzhner (changes requested):
The unit tests are failing for this PR.

packages/server-wallet/src/chain-service/chain-service.ts
snario: It already does; it was just missing the return type (which is odd that the linter didn't notice?)

packages/server-wallet/src/chain-service/chain-service.ts
snario: Yeah, I agree, changing

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -332,7 +419,25 @@ export class ChainService implements ChainServiceInterface {
...
+            {
+              assetHolderAddress,
+              tokenAddress,
+              amount,
+              tx: tx.hash,
+            },
+            'Submitted transaction to approve maximum amount of asset holder spending'
+          );
+
+          await tx.wait();
```
kerzhner, line 434: Why wait for transaction mining here? The wallet **can** await requests to the chain service. So the expectation is for chain service requests to return in a reasonable amount of time. At the moment, the heuristic is that it is reasonable to wait for transaction submission, but not transaction mining.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Can this condition ever get triggered?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -148,12 +168,38 @@ export class ChainService implements ChainServiceInterface {
...
+      'Finished funding channel'
+    );
+
+    return tx;
   }
 
   async concludeAndWithdraw(
     finalizationProof: SignedState[]
   ): Promise<providers.TransactionResponse | void> {
+    if (finalizationProof.length === 0) {
```
kerzhner, line 191:
```suggestion
    if (!finalizationProof.length) {
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -148,12 +168,38 @@ export class ChainService implements ChainServiceInterface {
...
+      this.logger.warn('ChainService: concludeAndWithdraw was called with an empty array?');
+      return;
```
kerzhner, line 193: We might want to throw an error?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -163,24 +209,30 @@ export class ChainService implements ChainServiceInterface {
 
     const captureExpectedErrors = async (reason: any) => {
       if (reason.error?.message.includes('Channel finalized')) {
+        this.logger.warn(
+          {channelId, determinedBy: 'Revert reason'},
```
kerzhner, line 213: We might want to include `reason.error.message` in the log?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -163,24 +209,30 @@ export class ChainService implements ChainServiceInterface {
...
-      const latestBlockTimestamp = (
-        await this.provider.getBlock(await this.provider.getBlockNumber())
-      ).timestamp;
       // Check if the channel has been finalized in the past
-      if (latestBlockTimestamp >= Number(finalizesAt)) return;
+      if (latestBlockTimestamp >= Number(finalizesAt)) {
+        this.logger.warn(
+          {channelId, determinedBy: 'Javascript check'},
+          'Transaction to conclude channel failed: channel is already finalized'
+        );
```
kerzhner, line 230: A return is missing. I am surprised unit tests didn't fail...

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -313,7 +382,25 @@ export class ChainService implements ChainServiceInterface {
...
+          {
+            assetHolderAddress,
+            tokenAddress,
+            amount,
+            tx: tx.hash,
+          },
+          'Submitted transaction to increase asset holder token allowance'
+        );
+
+        await tx.wait();
```
kerzhner, line 397: This is a significant change to the chain service API. See my comment on the other transaction mining wait.

kerzhner (changes requested):

packages/server-wallet/src/chain-service/chain-service.ts
snario: I think this is just a bad merge conflict resolution; it used to call `.wait()` before the recent refactor

packages/server-wallet/src/chain-service/chain-service.ts
snario: The reason is that the channel is already finalized which is in the message; did you mean verbatim?

packages/server-wallet/src/chain-service/chain-service.ts
snario: It's kind of a harmless situation, so I'm not sure if we want to

packages/server-wallet/src/chain-service/chain-service.ts
snario: I think it is always covered by the revert reason. TBH I don't know why do this check twice

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Is there ever a valid scenario for calling `concludeAndWithdraw` with no finalization proof?

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: I guess the only thing we could find out by including `reason.error.message` is whether the error is coming from transaction submission (in which case we will see a "failed to estimate gas" error) or transaction mining.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Now that I think about it, the revert reason is not present on Rinkeby: https://github.com/ethers-io/ethers.js/issues/446. So I think we need the return.

Unit tests are not failing because we run the tests against Ganache.

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -315,7 +381,13 @@ export class ChainService implements ChainServiceInterface {
           to: tokenContract.address,
         };
 
-        await this.sendTransaction(increaseAllowanceRequest);
+        const tx = await this.sendTransaction(increaseAllowanceRequest);
+
+        this.logger.info(
+          {tx: tx.hash},
+          'Transaction to increase asset holder token allowance successfully mined'
```
kerzhner, line 388:
```suggestion
          'Transaction to increase asset holder token allowance successfully submitted'
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -334,7 +406,23 @@ export class ChainService implements ChainServiceInterface {
...
+            {
+              assetHolderAddress,
+              tokenAddress,
+              amount,
+              tx: tx.hash,
+            },
+            'Submitted transaction to approve maximum amount of asset holder spending'
+          );
+
+          this.logger.info(
```
kerzhner, line 421: I think this log should be removed?

kerzhner (approved):

packages/server-wallet/src/chain-service/chain-service.ts
snario: a8bd7ea02

packages/server-wallet/src/chain-service/chain-service.ts
snario: In general I think the `ChainService` should hide the gas estimation wrapped errors from the consumer — the actual error is the interesting bit that the consumer wants to know about, it just happens to be that the RPC call to estimate gas consumption provides it to the `ChainService` to read

PR #2997, opened 2020-12-01
pushUpdate 

tomclose:
Adds a `pushUpdate` method to the wallet api, for pushing a message containing a single state that updates a running app channel. Should address some of the performance issues seen in https://github.com/statechannels/project-management/issues/4, by allowing objective cranking to be skipped in this case.

Still todo:
- [x] Write tests for the handler to test success and the various error conditions
- [x] Update the e2e/stress tests to use the new method where appropriate

### DIFF
--- a/packages/server-wallet/src/handlers/single-app-updater.ts
+++ b/packages/server-wallet/src/handlers/single-app-updater.ts
@@ -1,0 +1,51 @@
+import {WirePayload} from '../type-aliases';
+import {WalletResponse} from '../wallet/wallet-response';
+import {Store} from '../wallet/store';
+
+/**
+ * For making a single update to a running application channel
+ * */
+export class SingleAppUpdater {
+  private store: Store;
+
+  public static create(store: Store): SingleAppUpdater {
+    return new SingleAppUpdater(store);
+  }
+
+  private constructor(store: Store) {
+    this.store = store;
+  }
+
+  /**
+   * For pushing a message containing a single update to a running application channel
+   */
+  async update(wirePayload: WirePayload, response: WalletResponse): Promise<void> {
+    // check that we have exactly one signedState and nothing else
+    if (wirePayload.signedStates?.length !== 1)
+      throw new Error(`The payload sent to pushUpdate must contain exactly 1 signedState.`);
+    if (wirePayload.objectives)
+      throw new Error(`The payload sent to pushUpdate must not contain objectives.`);
+    if (wirePayload.requests)
+      throw new Error(`The payload sent to pushUpdate must not contain requests.`);
+
+    const wireState = wirePayload.signedStates[0];
+    const channelId = wireState.channelId;
+
+    await this.store.transaction(async tx => {
+      const channel = await this.store.addSignedState(channelId, wireState, tx);
+
+      // check channel is an application channel. If not, transaction will rollback when promise rejects.
+      if (!channel.isAppChannel)
+        throw new Error(`The update sent to pushUpdate must be for an application channel`);
+
+      // check that the channel is running. If not, transaction will rollback when promise rejects
+      // note that this actually checks that the channel is still running _after_ the update has
+      // been applied, as this is the case where no cranking is required
+      if (!channel.isRunning)
+        throw new Error(`The update sent to pushUpdate must be for a running channel`);
+
+      // and add to response
+      response.queueChannel(channel);
+    });
+  }
+}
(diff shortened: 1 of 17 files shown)


packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -523,6 +524,20 @@ export class SingleThreadedWallet extends EventEmitter<EventEmitterType>
...
+  /**
+   * For pushing a message containing a single update to a running application channel
+   */
+  async pushUpdate(rawPayload: unknown): Promise<SingleChannelOutput> {
+    const wirePayload = validatePayload(rawPayload);
+
+    const response = WalletResponse.initialize();
+
+    const singleAppUpdater = SingleAppUpdater.create(this.store);
+    await singleAppUpdater.update(wirePayload, response);
```
geoknee, line 536:
```suggestion
    await SingleAppUpdater.create(this.store).update(wirePayload, response);
```

snario:
Suppose `pushUpdate` _should_ cause a crank to be done, for example if the update is set to `isFinal: true`. What is the method for realizing this at some point later to respond accordingly (i.e., signing a new state with `isFinal`)

geoknee (approved):

packages/server-wallet/src/handlers/single-app-updater.ts
```diff
@@ -0,0 +1,51 @@
...
+
+      // check channel is an application channel. If not, transaction will rollback when promise rejects.
+      if (!channel.isAppChannel)
+        throw new Error(`The update sent to pushUpdate must be for an application channel`);
+
+      // check that the channel is running. If not, transaction will rollback when promise rejects
+      // note that this actually checks that the channel is still running _after_ the update has
+      // been applied, as this is the case where no cranking is required
+      if (!channel.isRunning)
+        throw new Error(`The update sent to pushUpdate must be for a running channel`);
```
geoknee, line 45: Thinking about @snario's scenario:  this places a requirement on the app to understand what _not_ to send through this method. I think we should alter the docstring on the API to reflect this more fully, i.e. explicitly mention to not send `isFinal` states?

packages/server-wallet/src/models/channel.ts
```diff
@@ -343,6 +343,21 @@ export class Channel extends Model implements RequiredColumns {
...
+    return !this.isLedger;
+  }
+
+  public get isRunning(): boolean {
+    // running if:
+    //  1. the supported state implies a post-fund-setup
+    //  2. no isFinal states exist
+
+    const havePostFund = !!this.supported && this.supported.turnNum >= 2 * this.nParticipants() - 1;
+    const noFinalStates = _.every(this.sortedStates, s => !s.isFinal);
```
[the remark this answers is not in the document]
andrewgordstewart: There shouldn't be more than `numParticipants` states in normal behaviour, since the wallet removes states that are older than those in the support of the latest supported state.

packages/server-wallet/src/handlers/single-app-updater.ts
snario: This was my opposition to this method when speaking with Tom last week. I'd prefer if we only had `pushMessage` and defaulted `doAnotherCrank` to be `true` unless this `isRunning` check is `true`

packages/server-wallet/src/handlers/single-app-updater.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * For making a single update to a running application channel
+ * */
+export class SingleAppUpdater {
+  private store: Store;
+
+  public static create(store: Store): SingleAppUpdater {
+    return new SingleAppUpdater(store);
+  }
```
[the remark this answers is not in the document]
tomclose: It's something that I'm doing with most classes now, as it gives you more flexibility if you want to refactor later and keep the same api (e.g. [this sort of thing](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/wallet/index.ts#L10-L18)). 

Also it means that you don't have to write `new`, which spoils the flow of the code :).

packages/server-wallet/src/handlers/single-app-updater.ts
tomclose: I agree with this. I think it's worth merging this (to get the short-term benefit, as it seems to fit well with the-graph's use-case) but scheduling work to make the logic part of pushMessage later.

Issue #2929, opened 2020-11-19
Ensure that channels don't get "stuck" if the chain service fails for some reason


snario:
Yesterday I had been using the `MockChainService` and `Direct` channels and saw several channels make request to be funded to the `MockChainService`. I realized later that I was using the `MockChainService`. So, I restarted the wallet and told it to use a real chain service. However, now, since the request to fund the channel was logged and stored in `Channel.chainServiceRequests` in the DB, it did _not_ try to ask for funding again to the (now real) chain service.

I think, in general, this is an issue. If the chain service fails for some reason and the wallet thinks that it successfully requested funding, then we are now in a spot where the wallet thinks that the channel will be funded and the chain service is doing nothing, and doesn't know it has to do anything.

What do we do about this?

snario:
I think this relates to something I wrote earlier https://github.com/statechannels/statechannels/issues/2534. It seems like I initially expected `fundChannel` to be idempotent and that the wallet would not store information about whether or not it already called `fundChannel`.

kerzhner:
> I think this relates to something I wrote earlier #2534. It seems like I initially expected `fundChannel` to be idempotent and that the wallet would not store information about whether or not it already called `fundChannel`.

Idempotency of `fundChannel` is probably a good property! Unfortunately, given the run loop architecture of the wallet, the wallet needs to store something to indicate that `fundChannel` is being attempted. Otherwise, the run loop will spin, continuously requesting the chain service to fund a channel over and over again.

An alternative is for wallet to await channel funding. Channel funding takes a long time (6 blocks), so that might not be a good idea.

andrewgordstewart:
There could (should?) be some way for the channel manager to ask the wallet to re-try failed attempts.

kerzhner:
Alternatively, the wallet can clear the chain request status after some time. The wallet can re-request `fundChannel` if holdings are not updated after a certain amount of time from the initial chain request.

However, this moves the wallet further away from being a reducer on top of a store. The wallet would now schedule tasks.

Perhaps the right place for this logic is either at the application level or some other cron-type system that calls the wallet.

kerzhner:
Proposed strategy: https://www.notion.so/Chain-service-resilience-83c40060b2dc401e8e2efa6852b9c503

snario:
[...]

PR #3001, opened 2020-12-02
Add timeouts to chain service requests

kerzhner:
The high level purpose of this PR is to add timeouts to requests from the wallet to the chain service. This should help the following scenario:
1. The wallet requests the chain service to fund or conclude the channel.
1. Transaction mining fails for some reason.
1. The wallet is not able to make any progress.

With this PR, each request to the chain service is retried once after 10 minutes. It is the job of the application to trigger a protocol rerun that retries the chain service request.

Future work includes:
- The chain service might want to notify the wallet if any transaction associated with a request from the wallet fails.
- We might want to clean up chain service requests from the database after the request has succeeded.

Also, this PR does NOT include a seamless migration from the old chain service requests **column** to the new chain service requests **table**. This will impact wallets that are mid channel funding or concluding. However, I don't believe we have wallets in production that are directly funding channels.

Fixes https://github.com/statechannels/statechannels/issues/2929.


packages/server-wallet/src/protocols/__test__/open-channel.test.ts
```diff
@@ -58,7 +59,12 @@ describe(`funding phase`, () => {
       await crankAndAssert(obj, {fundsToDeposit: 5});
       // and then doesn't submit it a second time
       await crankAndAssert(obj, {fundsToDeposit: 0});
-    });
+      // but then let's "wait" for 10 minutes for the fund request to the chain service to get stale
+      await ChainServiceRequest.query(knex)
+        .findOne({channelId: testChan.channelId})
+        .patch({timestamp: new Date(Date.now() - requestTimeout - 1)});
+      await crankAndAssert(obj, {fundsToDeposit: 5});
```
tomclose, line 66: Might be worth adding a test to ensure that it doesn't retry after 10mins if the transaction went through?

tomclose (approved):
This looks good to me. 🌋 

I wonder whether we're almost at the point where it's worth moving some of this logic (and storage) into the chain service. As a wallet, it would be nice to not to have to worry about this stuff (whether 10mins is the right retry interval, and whether someone's going to trigger a crank at the right time etc.). As a wallet, I would probably like to send a transaction to the chain service, and have it do whatever's necessary to get the transaction in (retries, increasing gas, backing off when the chain is busy/whatever) or tell me if it can't get the transaction in.

tomclose:
Also worth asking whether you've thought about any potential performance impacts here?

kerzhner:
> Also worth asking whether you've thought about any potential performance impacts here?

I have a bit, but not rigorously. Do you know if there is a good way to measure performance impact?

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
kerzhner: You are actually pointing out something that potentially could be an issue. The chain service request retry does not depend on the success/failure of the chain service request. At the moment, every chain service request is associated with a chain event: the funding request is associated with the deposit event; the close + withdraw request is associated with the asset transferred event.

So the retry will happen in the following case:
1. 10 minutes have elapsed.
1. The protocol is rerun.
1. The chain service request has **succeeded**.
1. The wallet has not received a chain event.

But maybe that's ok for now?

snario:
I'm curious to hear what your thoughts are on this feedback:

> The proposal is definitely an improvement over what we have now 👍 My main feedback would be that we could consider higher fidelity communication between the chain service and the wallet than "time elapsed since last call". For example, the chain service could identify and relay information like (a) transaction reverted, out of gas (b) transaction reverted: "thrown error from contract", (c) insufficient ETH to pay for transaction, or (d) unable to reach internet, API error, or (e) transaction pending. There could be some API on the chains service like `getStatusOnRequest(fingerprint)` which returns some information about the request.

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
tomclose: I think it's probably ok for the receipt of the chain event to be the signal that the chain service request succeeded, which is essentially the situation you're describing above.

The thing it might be worth confirming is that if a chain event _has_ been received, then no retry occurs. I expect that to work, but it would be nice to be sure.

kerzhner:
> This looks good to me. 🌋
> 
> I wonder whether we're almost at the point where it's worth moving some of this logic (and storage) into the chain service. As a wallet, it would be nice to not to have to worry about this stuff.

@tomclose, I think this is an excellent point to bring up. Here is how I have been thinking about this. The chain service has a high level API such as fund, conclude, challenge. Each API call can result in many transactions. For example, the fund call can lead to 2 transactions: a transaction to increase token allowance for the asset holder and a transaction to deposit tokens. Down the line, with channels holding multiple different assets, the chain service fund API call will trigger many deposits.

Due to the runloop wallet design, the wallet needs to store whether a chain service api has been called. The wallet does not and, probably, should not need to worry about how to ensure transactions are submitted. The chain service requests table records high level information about communication between the wallet and the chain service. The chain service requests table also allows protocols to remain somewhat synchronous and avoid the need to wait for chain service responses.

Your comment highlights that:
1. The chain service (not the wallet) should attempt transaction retries in certain cases. This is not built yet.
1. There should be a way to communicate chain service API failures that might happen long after the API response. For example, the chain service fund API returns successfully after transactions have been submitted. But how should we communicate transaction mining failures?

If you have any thoughts about this, would love to hear them! @snario, this is related to your comment above.

packages/server-wallet/src/protocols/__test__/open-channel.test.ts
kerzhner: Thanks for the feedback! If all prefund states are singed (a prerequisite for a fund chain service request), then a chain service requests will get submitted if:
1. There is not a valid existing chain service request. This is currently tested here https://github.com/statechannels/statechannels/blob/chain-service-resilience/packages/server-wallet/src/protocols/__test__/open-channel.test.ts#L57.
2. The channel needs the participant's funds. This is already tested with existing unit tests. One of which is here: https://github.com/statechannels/statechannels/blob/chain-service-resilience/packages/server-wallet/src/protocols/__test__/open-channel.test.ts#L81 

We should add a test similar to 1 for the close protocol.

kerzhner: Tests added in ed415705d.

PR #3002, opened 2020-12-02
fix: do not do turn number check for funding or closing stage channels

snario:
A legit "transition" is from turn 1 to 3 as there are no app rules checked in this phase to warrant enforcing turn taking.


tomclose (approved):
LGTM.

Separate to the review, I'm not 100% sure why we have that check at all. It seems like either failing the turnNumCheck is a problem, and the wallet should proceed, or it isn't a problem, and the wallet shouldn't log an error. Does anyone know why we log an error and continue?

Issue #3000, opened 2020-12-02
Concluding 20k channels blocks the main nodejs thread for hours


andrewgordstewart:
> [Jannis] `mergeProp` concatenates old objectives and one new objective and then does `uniqueWith` using a deep comparison to ensure the result of the concatenation is still unique values. It should be enough to just check if the new objective isn’t already included.

There isn't necessarily an argument for concluding 20k channels at once. However, there is still a `O(n^2)` (?) algorithm running in the main thread of the gateway manager when closing (and opening?) channels.

tomclose:
https://github.com/statechannels/the-graph/pull/442#issuecomment-737063311

PR #3004, opened 2020-12-02
WalletResponse improvements

tomclose:
Makes some improvements to `WalletResponse`, including addressing the issue in https://github.com/statechannels/the-graph/issues/441.

* moves the `mergeOutcomes` helper into `WalletResponse`, removing a utils file 🎉 
* rewrites `mergeOutcomes` to be clearer and more efficient
* refactors wallet so that it is only used by `WalletResponse`
* moves wallet type definitions out of `wallet.ts` and into `types.ts`

fixes https://github.com/statechannels/statechannels/issues/3000


packages/server-wallet/src/utilities/messaging.ts
```diff
@@ -19,43 +19,34 @@ export function mergeOutgoing(outgoing: Notice[]): Notice[] {
...
-    if (!mergedOutgoing[recipient]) {
-      mergedOutgoing[recipient] = {walletVersion: WALLET_VERSION};
+  const messages = outgoing.map(o => o.params as WireMessage);
+
+  return _.map(
+    _.groupBy(messages, o => o.recipient),
+    (rcptMsgs, recipient) => {
+      const states = uniqueAndSorted(rcptMsgs.flatMap(n => n.data.signedStates || []));
+      const requests = uniqueAndSorted(rcptMsgs.flatMap(n => n.data.requests || []));
+      const objectives = uniqueAndSorted(rcptMsgs.flatMap(n => n.data.objectives || []));
```
andrewgordstewart: This plus L38-40 makes me think that `n.data.signedStates` should _always_ be an array (maybe of length 0)?

packages/server-wallet/src/utilities/messaging.ts
```diff
@@ -19,43 +19,34 @@ export function mergeOutgoing(outgoing: Notice[]): Notice[] {
...
-      ['desc', 'asc']
-    );
-  } else if (a) {
-    return a;
-  } else if (b) {
-    return b;
-  } else {
-    return undefined;
-  }
+function uniqueAndSorted<T>(array: T[]): T[] {
```
andrewgordstewart: Considering that this commit is about removing `O(N^2)` behaviour, can you leave a comment about the complexity of this function?

andrewgordstewart (approved):
I only skimmed 0f43c05eb7b57117bee504eed1e38780aa03893f, as it was a sensible looking refactor. I think the point about removing the `O(N^2)` hit should be clarified in the codebase.

PR #3005, opened 2020-12-02
[docs-website] Replace solidoc with solidity-docgen

geoknee:
We have our own fork of `solidoc` (pretty much unmaintained package) to generate the "Contract API" section of the docs-website. It relies on contract compilation artifacts having entries such as `ast`, which I can't convince `hardhat` to output. 

This branch is exploring using `solidity-docgen` instead, which is maintained by openzeppelin. It has the downsides of 

- producing somewhat more vanilla output, but this is partly due to having not yet fined tuned the handlebars templates
- recompiling contracts from source using `solcjs` (somewhat slow and inelegant)

I imagine these downsides may go away as new versions are released. There may also be the possibility of piggybacking on hardhat's compiler https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/scripts/prepare-docs-solc.js.

TODO
- [ ] fixup some tests
- [ ] improve the handlebars template


geoknee:
In favour of #3029

PR #3006, opened 2020-12-02
Add bytecode checking to ChainService constructor

snario:
Note this is a stacked PR.

This checks the on-chain code before instantiating a `ChainService`; helpful when building an application which requires the `ChainService` to function in avoiding finding out about a misconfigured `ChainService` until mid-way through operation

### DIFF
--- a/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
+++ b/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
@@ -8,34 +8,45 @@ import {fromEvent} from 'rxjs';
 import {take} from 'rxjs/operators';
 
 import {defaultTestConfig, overwriteConfigWithDatabaseConnection} from '../config';
+import {ChainService} from '../chain-service';
 import {Wallet, SingleChannelOutput} from '../wallet';
 import {getChannelResultFor, getPayloadFor} from '../__test__/test-helpers';
 
 // eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion
 const ethAssetHolderAddress = makeAddress(process.env.ETH_ASSET_HOLDER_ADDRESS!);
+// eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion
+if (!process.env.RPC_ENDPOINT) throw new Error('RPC_ENDPOINT must be defined');
+// eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion
+const rpcEndpoint = process.env.RPC_ENDPOINT;
+
 const config = {
   ...defaultTestConfig(),
   networkConfiguration: {
     ...defaultTestConfig().networkConfiguration,
     // eslint-disable-next-line no-process-env
-    rpcEndpoint: process.env.RPC_ENDPOINT,
-    // eslint-disable-next-line no-process-env
     chainNetworkID: parseInt(process.env.CHAIN_NETWORK_ID || '0'),
   },
 };
-if (!config.networkConfiguration.rpcEndpoint) throw new Error('rpc endpoint must be defined');
-const {rpcEndpoint} = config.networkConfiguration;
+
 let provider: providers.JsonRpcProvider;
-const b = Wallet.create({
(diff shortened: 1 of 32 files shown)


snario:
I just realized `ContractArtifacts` only refers to some locally deployed artifacts, so `deployedBytecode` is incorrect. I propose some kind of API like this on `@statechannels/nitro-protocol`:

```ts
import { chainArtifacts } from '@statechannels/nitro-protocol';
const { deployedBytecode, abi, address } = chainArtifacts['rinkeby' /* or 4 */].ETHAssetHolder;
```

This way also we can fetch the contract addresses too instead of having to thread them into the config files. cc @geoknee

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -37,46 +37,55 @@ type DepositedEvent = {type: 'Deposited'; ethersEvent?: Event} & HoldingUpdatedA
...
+    allowanceMode,
+    nitroAdjudicatorAddress,
+    ethAssetHolderAddress,
+  }: ChainServiceArgs): Promise<ChainService> {
+    const jsonrpc = new providers.JsonRpcProvider(provider);
+
+    if (pollingInterval) jsonrpc.pollingInterval = pollingInterval;
+
+    if (provider.includes('0.0.0.0') || provider.includes('localhost'))
+      jsonrpc.pollingInterval = pollingInterval ?? 50;
```
kerzhner, line 62: This might be more clear. This condition is satisfied when the provider is `0.0.0.0` or localhost AND polling interval has not been set in the line above.
```suggestion
    if ((provider.includes('0.0.0.0') || provider.includes('localhost')) && !pollingInterval )
      jsonrpc.pollingInterval = 50;
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -353,3 +358,27 @@ export class ChainService implements ChainServiceInterface {
...
+  }: {contractName: string; abi: ethers.ContractInterface; deployedBytecode: string},
+  provider: providers.JsonRpcProvider
+): Promise<Contract> {
+  const bytecode = await provider.getCode(address);
+
+  // TODO: Throw errors once we can be certain the bytecode being referenced is correct;
+  //       right now it may be incorrect since our ContractArtifacts refer to local deployments
+  //       and not deployed bytecode on various other chains e.g., rinkeby, mainnet, etc
+
+  if (BN.eq(bytecode, 0))
```
kerzhner, line 377: If I understand correctly, bytecode is not a number.

kerzhner (approved):

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -72,6 +60,31 @@ function mineOnEvent(contract: Contract) {
 
 beforeAll(async () => {
   provider = new providers.JsonRpcProvider(rpcEndpoint);
+
+  a = Wallet.create(
+    await ChainService.create({
```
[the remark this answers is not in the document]
snario: 1. The original plan was for the `ChainService` to be an external component to the `server-wallet`, and writing it inside the `server-wallet` was only ever a short-term convenience.
2. Arguably it's easier to write the tests because it's easier to understand when you're using a mock and when you're expecting to use and test this external component which has special characteristics (i.e., connects to the internet).
3. I've already rewritten all of the tests and propagated all of the changes upstream for convenience, so I've made a best attempt to make the transition as smooth as possible.

packages/server-wallet/src/chain-service/chain-service.ts
snario: It's a `Bytes` which is a `BigNumberish`

snario: Might be easier to do a direct string comparison instead like `== "0x00"`.

kerzhner: Perhaps it's pedantic, but it would be nice to use `BigNumber` and `BigNumberish` for variables/constants that represent numbers.

snario:
I'm going to re-open this PR with a slightly different approach

PR #3011, opened 2020-12-03
[nitro-protocol] accept a list of indices in transfer method

### LINKED ISSUE #2936: Pass index to transfer (instead of destination)
Idea from @andrewgordstewart: Instead of passing in the `destination`, better to pass in the `index` of the destination in the allocation. Same goes (actually more urgently so) for `transfer`. 

We could even pass `indices: number[]` to target multiple destinations -- and have `[]` imply "all".

Because we pass in the abi-encoded allocation, this implies we know which destination is at which index (so no need to recheck this on chain).

_Originally posted by @geoknee in https://github.com/statechannels/statechannels/pull/2932#r527880096_

geoknee:
... instead of a single `destination`. 

An empty list implies "all".

I think this is a much cleaner implementation and DRYs out the contract a lot. 

TODO
- [x] factor out function that computes new outcome to a pure function, so we can `PureEVM` it. cc @kerzhner and the chain service.
- [x] delete the hash if all outcomeItems allocate 0 (save a *lot* of gas if we can, spend a bit every time to check)
- [ ] something similar for `claimAll` (out of scope for this PR)
- [ ] rethink and overhaul event emission (out of scope for this PR)

Closes #2936


geoknee:
I went ahead with factoring `_computeNewOutcome` into a pure function. It consumes more gas (circa 500 gas in our test cases) due to the need for copying the entire outcome (rather than mutating it in place), which means iterating over all of the allocation items. 

Master:

>  1. funded          -> 1 EOA:
> 20655 gas
> 
>  2. overfunded      -> 1 EOA:
> 26309 gas
> 
>  3. underfunded     -> 1 EOA:
> 27395 gas
> 
>  4. funded      -> 1 channel:
> 32407 gas
> 
>  5. overfunded  -> 1 channel:
> 47407 gas
> 
>  6. underfunded -> 1 channel:
> 48493 gas
> 
>  7. -> 2 EOA            full:
> 43419 gas
> 
>  8. -> 2 EOA            full:
> 28407 gas
> 
>  9. -> 2 EOA         partial:
> 44427 gas
> 
> 11. -> 2 chan           full:
> 49661 gas
> 
> 12. -> 2 chan        partial:
> 65681 gas
>
> 12. -> 2 chan        partial:
> 65681 gas


5315d6e:

>  1. funded          -> 1 EOA:
> 21003 gas
> 
>  2. overfunded      -> 1 EOA:
> 27001 gas
> 
>  3. underfunded     -> 1 EOA:
> 28054 gas
> 
>  4. funded      -> 1 channel:
> 32998 gas
> 
>  5. overfunded  -> 1 channel:
> 47994 gas
> 
>  6. underfunded -> 1 channel:
> 49035 gas
> 
>  7. -> 2 EOA         1 index:
> 45114 gas
> 
>  8. -> 2 EOA         1 index:
> 30062 gas
> 
>  9. -> 2 EOA         partial:
> 45126 gas
> 
> 10. -> 2 chan             no:
> 34809 gas
> 
> 11. -> 2 chan           full:
> 51187 gas
> 
> 12. -> 2 chan        partial:
> 66263 gas
> 
> 13. -> 2 indices:
> 59762 gas
> 
> 

The *original* motivation for doing this may or may not stand the test of time*, but it has the useful side effect of easily tracking when it is safe to delete the allocation's hash entirely. If we optimise for the happy path of paying out the entirety of a (small) outcome, this is the right way forward IMO.

 FWIW I think readability was improved, too.

*@tomclose suggests an alternative route to tracking changes in the client without emitting the entire allocation in an event. Simply emit the computed payouts^ (e.g. `uint[] indices` and `uint[] payouts` here), from which it should be trivial to compute the new allocation in client code. 

^deductions might be a better name, since they can lead to assets moving only _internally_ to this contract.

andrewgordstewart:
Emitting the computed payouts would cost ~500 gas _per allocation item_, correct? 256 gas per word, 2 words per allocation item.

geoknee:
We already emit an `AssetTransferred` event for each internal and external payout: 

https://github.com/statechannels/statechannels/blob/3250047a6104780e43ca8c0cc71ea28b089ff0dc/packages%2Fnitro-protocol%2Fcontracts%2Finterfaces%2FIAssetHolder.sol#L53

375 + ( 2 * 375 ) + ( 3 * 256 ) = 1893 gas per deduction.

[source](https://www.reddit.com/r/ethdev/comments/78ugki/do_indexed_parameters_in_events_cost_more_gas/) 

We could make some decent savings by having an `AssetsTransferred` event, (only one LOG opcode and one `channelId` indexed parameter). We could tackle that in a separate PR.

packages/nitro-protocol/test/contracts/AssetHolder/transfer.test.ts
```diff
@@ -122,6 +122,7 @@ describe('transfer', () => {
           }
         });
         const {events: eventsFromTx, gasUsed} = await (await tx).wait();
+        // NOTE: _transferAsset is a NOOP in TESTAssetHolder, so gas costs will be much lower than for a real Asset Holder
```
andrewgordstewart, line 125: Is it possible to "spend" some expected amount of gas spent transferring ERC20 tokens? eg. transferring a vanilla token to an address currently holding no tokens.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -139,203 +139,120 @@ contract AssetHolder is IAssetHolder {
...
+
+        // loop over allocations and decrease surplus
+        for (uint256 i = 0; i < allocation.length; i++) {
+            // copy destination part
+            newAllocation[i].destination = allocation[i].destination;
+            // compute new amount part
+            uint256 affordsForDestination = (allocation[i].amount > surplus)
+                ? surplus
+                : allocation[i].amount; // min(amount,surplus)
+            if ((indices.length == 0) || ((k < indices.length) && (indices[k] == i))) {
```
andrewgordstewart, line 171: required: The correctness here seems to rely on `indices` being sorted.

Can you confirm that? If so, I think we should assert that `indices` is sorted, when present.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -139,203 +139,120 @@ contract AssetHolder is IAssetHolder {
...
-            }
-            for (uint256 k = i + 1; k < allocation.length; k++) {
-                splicedAllocation[k - 1] = allocation[k];
-            }
-            if (splicedAllocation.length == 0) {
-                deleteHash = true;
-            }
-            encodedAllocation = abi.encode(splicedAllocation);
+        if (newHoldings == 0) {
+            delete holdings[fromChannelId];
```
andrewgordstewart, line 221: Considering that mapping values default to `0`, does setting `holdings[fromChannelId] = 0` have a different effect than `delete holdings[fromChannelId]`?

andrewgordstewart (changes requested):
There's one required review item. Otherwise, looks good!

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -139,203 +139,120 @@ contract AssetHolder is IAssetHolder {
...
+        internal
+        pure
+        returns (
+            Outcome.AllocationItem[] memory newAllocation,
+            bool safeToDelete,
+            uint256[] memory payouts,
+            uint256 totalPayouts
+        )
+    {
+        payouts = new uint256[](indices.length > 0 ? indices.length : allocation.length); // [] means "all"; values default to 0
```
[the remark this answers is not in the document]
geoknee: It's just the value the array elements are initialised to. If you don't explicitly set an entry of this array, and later access it, you will get 0.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -139,203 +139,120 @@ contract AssetHolder is IAssetHolder {
...
+        totalPayouts = 0;
+        newAllocation = new Outcome.AllocationItem[](allocation.length);
+        safeToDelete = true; // switched to false if there is an item with amount > 0
+        uint256 surplus = initialHoldings; // virtual funds available during calculation
+        uint256 k = 0; // indexes indices
```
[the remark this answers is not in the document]
geoknee: yeah indexes is a verb, indices is plural noun.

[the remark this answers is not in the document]
andrewgordstewart: touche!

ok, so now I managed to read this as "`k` indexes indices" 😄 It still took a second to unpack this as "`k` indexes the `indices` array".

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: I don't think it has a different effect. So yeah we don't need this branching logic here.

andrewgordstewart: I suppose it's important to know for sure, since we wait to provide the gas refund is possible.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -26,17 +26,17 @@ contract AssetHolder is IAssetHolder {
...
      * @param allocationBytes The abi.encode of AssetOutcome.Allocation
-     * @param destination External destination or channel to transfer funds *to*.
+     * @param indices Array with each entry denoting the index of a destination to transfer funds to.
      */
     function transfer(
         bytes32 fromChannelId,
         bytes calldata allocationBytes,
-        bytes32 destination
-    ) external {
+        uint256[] memory indices
```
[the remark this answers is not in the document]
geoknee: You should use `calldata` if you can because it is a special storage location with lower gas costs associated with it. But in practice I have often found it necessary to use `memory` for function parameters so that those functions can be used internally and externally.

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: The [docs ](https://docs.soliditylang.org/en/v0.7.5/types.html?highlight=delete#delete)are pretty clear 
> delete a assigns the initial value for the type to a. I.e. for integers it is equivalent to a = 0, 

and

> It is important to note that delete a really behaves like an assignment to a, i.e. it stores a new object in a.

packages/nitro-protocol/contracts/AssetHolder.sol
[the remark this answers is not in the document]
geoknee: "indexes" is also a plural noun. So maybe I should put '`k` indexes indexes' 😃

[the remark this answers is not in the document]
andrewgordstewart: My suggestion:
```suggestion
        uint256 k = 0; // indexes the `indices` array
```

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: d867e74

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: The behaviour of the function certainly depends on the ordering of `indices` in this implementation. There is a test case that shows this:
https://github.com/statechannels/statechannels/blob/3391f1fd482e511e16f38b9a79da71faadc14a1c/packages/nitro-protocol/test/contracts/AssetHolder/transfer.test.ts#L59-L61

I think we have two options: 

a) your suggestion: revert if unsorted `indices` is provided. This will cost some gas to do the check,
b) warn in the documentation about providing unsorted `indices`: 

> if at some point `indices[k+1] <= indices[k]`, then all indices >= k + 1 will be effectively ignored. 

This will not cost any extra gas.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -139,203 +139,120 @@ contract AssetHolder is IAssetHolder {
...
+        uint256 k = 0; // indexes indices
+
+        // loop over allocations and decrease surplus
+        for (uint256 i = 0; i < allocation.length; i++) {
+            // copy destination part
+            newAllocation[i].destination = allocation[i].destination;
+            // compute new amount part
+            uint256 affordsForDestination = (allocation[i].amount > surplus)
+                ? surplus
+                : allocation[i].amount; // min(amount,surplus)
```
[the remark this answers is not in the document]
geoknee: 1b919e2

packages/nitro-protocol/contracts/AssetHolder.sol
geoknee: I implemented a) in 20153dd and 4e243df so we can make a judgement call based on the extra gas spent and either keep or revert/omit those commits.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -45,11 +45,11 @@ contract AssetHolder is IAssetHolder {
      * @param channelId Unique identifier for a state channel.
      * @param allocationBytes The abi.encode of AssetOutcome.Allocation
      */
-    function transferAll(bytes32 channelId, bytes calldata allocationBytes) external override {
+    function transferAll(bytes32 channelId, bytes calldata allocationBytes) external {
```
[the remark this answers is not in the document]
andrewgordstewart: https://github.com/statechannels/statechannels/issues/3047

packages/nitro-protocol/contracts/AssetHolder.sol
andrewgordstewart: Before: from [here](https://app.circleci.com/pipelines/github/statechannels/statechannels/10879/workflows/f9b7c266-56bc-41a4-93da-6717b3258b5e/jobs/53260/artifacts)
```
13. -> 2 indices:
59975 gas

14. -> 3 indices:
48208 gas
```

After https://github.com/statechannels/statechannels/commit/4e243df18c5ef0b4976647a03731a7a9390a177c, from [here](https://app.circleci.com/pipelines/github/statechannels/statechannels/10936/workflows/3f3bdd67-0bc3-4305-9ee2-2acbe5716190/jobs/53492/artifacts)
```
13. -> 2 indices:
59762 gas

14. -> 3 indices:
47869 gas

15. -> reverse order (see 13):
66606 gas
```

It seems as if the gas consumption _went down_. I assume from this that there is some variance, and that the additional require adds less gas than the variance between test runs?

Since the sorting order of the indices appears vital from a protocol security point of view (right? see below) I think we need to accept whatever the gas cost is, which is fortunately small.

### Percieved problem

If the `indices` array that is passed is not sorted, does it allow one to extract funds that one doesn't deserve? IE. is the order of the outcome items still respected? I could not immediately tell. I had the sense that the order is still preserved, but that the end effect was unintended (only a subset was withdrawn corresponding to a particular subset of the indices that are sorted).

geoknee: Firstly about the gas usage. There should not be any variance, in my understanding, although I do see some small (10 or 15 gas say) variance on circle. I don't currently know what to attribute this to; it might be linked somehow to running chain tests concurrently. 

The first commit you picked is ahead of an unrelated change which I would expect to _also_ reduce gas. 

I think the gas usage went up (as it has to), but it's not easy to glean that from my commit history because the tests didn't always run. I reran this test file on its own locally and got this (increases shown inline)

```
without:
  1. funded          -> 1 EOA:
20977 gas 
 2. overfunded      -> 1 EOA:
26966 gas
 3. underfunded     -> 1 EOA:
28055 gas
 4. funded      -> 1 channel:
32947 gas
 5. overfunded  -> 1 channel:
47947 gas
 6. underfunded -> 1 channel:
49036 gas
 7. -> 2 EOA         1 index:
45066 gas
 8. -> 2 EOA         1 index:
30050 gas
 9. -> 2 EOA         partial:
45130 gas
10. -> 2 chan             no:
34801 gas
11. -> 2 chan           full:
51175 gas
12. -> 2 chan        partial:
66255 gas
13. -> 2 indices:
59750 gas
14. -> 3 indices:
47804 gas
with:
 1. funded          -> 1 EOA:
21018 gas (+41)
 2. overfunded      -> 1 EOA:
27024 gas (+58)
 3. underfunded     -> 1 EOA:
28125 gas (+70)
 4. funded      -> 1 channel:
33017 gas (+70)
 5. overfunded  -> 1 channel:
48017 gas (+70)
 6. underfunded -> 1 channel:
49106 gas (+70)
 7. -> 2 EOA         1 index:
45136 gas (+70)
 8. -> 2 EOA         1 index:
30120 gas (+70)
 9. -> 2 EOA         partial:
45200 gas (+70)
10. -> 2 chan             no:
34871 gas (+70)
11. -> 2 chan           full:
51245 gas (+70)
12. -> 2 chan        partial:
66325 gas (+70)
13. -> 2 indices:
59999 gas (+249)
14. -> 3 indices:
48232 gas (+428)
```

Either way I think the gas increase is quite small and is tolerable.

geoknee: About your perceived problem: I stand by what I put as the [suggested documentation](https://github.com/statechannels/statechannels/pull/3011#discussion_r538174606).

The algorithm walks along the allocation and "virtually" pays out to each destination, in order. If it hits the first supplied index, it will "actually" pay it out. It will then continue walking, without resetting to the start of the allocation, and look for the second index. Once it reaches the end of the allocation, the algorithm terminates. If the second supplied index is not strictly greater than the first, the algorithm will never find it and no additional "actual" payouts will happen (and similarly no additional changes to the allocation will be made). If the second index _is_ greater than the first, the algorithm will find it, but we can make the same argument for the third index (and so on). 

So my claim is that current implementation will payout `indices[i]` for `i=0,1,...,m` where `m` is the largest integer such that `indices[m] > indices[m-1] > ... indices[0]`.

andrewgordstewart: I'm comfortable with the suggested documentation!

andrewgordstewart: > About your perceived problem: I stand by what I put as the [suggested documentation](https://github.com/statechannels/statechannels/pull/3011#discussion_r538174606).
> 
> The algorithm walks along the allocation and "virtually" pays out to each destination, in order. If it hits the first supplied index, it will "actually" pay it out. It will then continue walking, without resetting to the start of the allocation, and look for the second index. Once it reaches the end of the allocation, the algorithm terminates. If the second supplied index is not strictly greater than the first, the algorithm will never find it and no additional "actual" payouts will happen (and similarly no additional changes to the allocation will be made). If the second index _is_ greater than the first, the algorithm will find it, but we can make the same argument for the third index (and so on).
> 
> So my claim is that current implementation will payout `indices[i]` for `i=0,1,...,m` where `m` is the largest integer such that `indices[m] > indices[m-1] > ... indices[0]`.

Perhaps the algorithm deserves some high-level description like this?

packages/nitro-protocol/test/contracts/AssetHolder/transfer.test.ts
geoknee: That's possible, but I think this test is not the place to do that. We have other tests that target e.g. the `ERC20AssetHolder`, that do actually cover payouts to token holders and newbies.

geoknee:
Rebased over at #3049

PR #3018, opened 2020-12-04
build: update .env.rinkeby

geoknee:


snario:
We probably don't need this anymore since we have the hardhat artifacts file checked in, but might as well update it now anyway

snario (approved):

geoknee:
https://github.com/statechannels/statechannels/pull/2898#discussion_r525408335

PR #3019, opened 2020-12-04
[nitro-protocol] Run solhint (instead of prettier)

geoknee:
This brings our solidity linting/formatting setup closer to what we enjoy for typescript.

The linter has a plugin that will flag issues from prettier, so we don't need to run that separately. 

Quirks: 
- there doesn't seem to be a preconfigured way to disable the extra / conflicting style rules that solhint brings (which we want to do, since we will rely on prettier for those). So I disabled them manually
- automatically fixing issues does not seem to work. So I kept our `yarn prettier: write` command for that. 

I also failed to get my local VSCode to lint / format on save at all for .sol files, I am not sure how to fix that. 

- I disabled a few extra rules to prevent a large changeset. These should be undisabled and fixed in a future PR (then we will actually get the benefits of linting!)


packages/nitro-protocol/.solhint.json
```diff
@@ -0,0 +1,24 @@
...
+    "func-param-name-mixedcase": "off",	
+    "modifier-name-mixedcase": "off", 
+    "private-vars-leading-underscore": "off", 
+    "use-forbidden-name": "off", 	
+    "var-name-mixedcase": "off", 
+    "func-order": "off",
+    "imports-on-top": "off",
+    "ordering": "off",
+    "visibility-modifier-order": "off",
+    "prettier/prettier": "error",
```
geoknee, line 19: Manually disabling style rules and then enabling all of the prettier rules.

packages/nitro-protocol/.solhint.json
```diff
@@ -0,0 +1,24 @@
...
+    "compiler-version": ["error","0.7.4"],
+    "func-visibility": ["warn",{"ignoreConstructors":true}],
```
geoknee, line 21: Rules that I added that we could keep.

packages/nitro-protocol/.solhint.json
```diff
@@ -0,0 +1,24 @@
...
+    "reason-string": "off"
```
geoknee, line 22: Rules that I added that we should probably not keep.

tomclose (approved):

Issue #3010, opened 2020-12-03
nitro-protocol should export contract addresses and deployed bytecode of on-chain contracts


snario:
I just realized `ContractArtifacts` only refers to some locally deployed artifacts, so `deployedBytecode` is incorrect when run against a non-local chain. I propose some kind of API like this on `@statechannels/nitro-protocol`:

```ts
import { chainArtifacts } from '@statechannels/nitro-protocol';
const { deployedBytecode, abi, address } = chainArtifacts['rinkeby' /* or 4 */].ETHAssetHolder;
```

This way also we can fetch the contract addresses too instead of having to thread them into the config files. cc @geoknee

_Originally posted by @snario in https://github.com/statechannels/statechannels/pull/3006#issuecomment-737984347_

PR #3025, opened 2020-12-04
Export Nitro Protocol deployed addresses in nitro-protocol

snario:
With this we will get

```ts
const addresses = require('@statechannels/nitro-protocol/addresses.json')
```
to be available, addressing https://github.com/statechannels/statechannels/issues/3010


packages/nitro-protocol/hardhat-deploy/02-token.ts
```diff
@@ -2,8 +2,16 @@ import {HardhatRuntimeEnvironment} from 'hardhat/types';
...
 const func: DeployFunction = async function(hre: HardhatRuntimeEnvironment) {
-  const {deployments, getNamedAccounts} = hre;
-  const {deploy} = deployments;
+  const {
+    deployments: {deploy},
+    getChainId,
+    getNamedAccounts,
+  } = hre;
+
+  if ((await getChainId()) === '4') {
```
andrewgordstewart, line 11: Why is Rinkeby isolated here? Shouldn't we blacklist any non-internal testnet?

packages/nitro-protocol/addresses.json
```diff
@@ -0,0 +1,1139 @@
+{
+  "4": {
+    "rinkeby": {
+      "name": "rinkeby",
+      "chainId": "4",
+      "contracts": {
+        "NitroAdjudicator": {
+          "address": "0x94Fca2E2F57B25b2D9eA5d4FCEd6c5d8d9ad3172",
+          "abi": [
```
andrewgordstewart, line 9: This file is called `addresses.json`, but it contains more than addresses. Is that intentional?

snario: The most important information are the addresses, the ABI is just conveniently next to it... I think since the most important information are the addresses, `addresses.json` is reasonable

packages/nitro-protocol/hardhat-deploy/02-token.ts
snario: Probably, yeah

packages/nitro-protocol/addresses.json
andrewgordstewart: This has the effect of growing the size of the git repo by a reasonable amount on each deploy. Considering the repo is already very large, if the intent is to simply list contract addresses, I think it should contain just the addresses.

snario: 947c0b729

packages/nitro-protocol/hardhat-deploy/02-token.ts
```diff
@@ -2,8 +2,18 @@ import {HardhatRuntimeEnvironment} from 'hardhat/types';
...
-  const {deploy} = deployments;
+  const {
+    deployments: {deploy},
+    getChainId,
+    getNamedAccounts,
+  } = hre;
+
+  const chainId = await getChainId();
+
+  if (chainId !== '31337') {
```
kerzhner, line 13: Might want to comment that this is the default hardhat chainId.

kerzhner:
This PR removes deploy artifacts from the repository. So we lose information like:
- What PK deployed each contract.
- What was the transaction of the deploy.

Are we sure we don't want to retain this information? If you find deploy artifacts too verbose, perhaps we can trim some of the keys.

One of the nice things about hardhat + CREATE2 is that if a contract has not changed since the last deploy and a redeploy is triggered, the contract is NOT redeployed. Have you checked that this behavior holds with deploy artifacts removed?

packages/nitro-protocol/scripts/postdeploy
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cat addresses.json | jq 'walk(if (type == "object" and .abi) then del(.abi) else . end )' > addresses.json.tmp
```
kerzhner, line 3: What happens if jq is not installed locally?

snario:
> One of the nice things about hardhat + CREATE2 is that if a contract has not changed since the last deploy and a redeploy is triggered, the contract is NOT redeployed. Have you checked that this behavior holds with deploy artifacts removed?

I did try this and I also did find that that contract was not re-deployed

> - What PK deployed each contract.
>- What was the transaction of the deploy.
>
>Are we sure we don't want to retain this information? 

This information is available on etherscan, so we're not really losing it, just not storing it inside the repo explicitly; so I think it's fine

kerzhner (approved):

packages/nitro-protocol/scripts/postdeploy
geoknee: I hit this problem and decided to just rewrite what we need in javascript: 
https://github.com/statechannels/statechannels/pull/3072: https://github.com/statechannels/statechannels/pull/3072/commits/444c6130ee405a6bc1b647168e91c79e87a46223

geoknee: @snario is `addresses.json` included in the npm package?

PR #3029, opened 2020-12-04
[nitro-protocol, docs-website] Resurrect docs#Contract API via patching hardhat

geoknee:
- patches hardhat to incluce `ast` field in contract artifacts
- updates our [solidoc fork](https://github.com/statechannels/solidoc) to ignore some of the `.json` files that hardhat outputs
- configures netlfiy gui to deploy from master and generate deploy previews for PRs (may turn the previews off, but for now they are on to check this works)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,2 +1,2 @@
 [build]
-  ignore = "git diff --quiet HEAD^ HEAD packages/docs-website/"
+  ignore = "git diff --quiet HEAD^ HEAD packages/docs-website/ packages/nitro-protocol"
--- a/patches/hardhat+2.0.3.patch
+++ b/patches/hardhat+2.0.3.patch
@@ -1,0 +1,33 @@
+diff --git a/node_modules/hardhat/builtin-tasks/compile.js b/node_modules/hardhat/builtin-tasks/compile.js
+index ab562c3..bdf669f 100644
+--- a/node_modules/hardhat/builtin-tasks/compile.js
++++ b/node_modules/hardhat/builtin-tasks/compile.js
+@@ -448,12 +448,14 @@ config_env_1.subtask(task_names_1.TASK_COMPILE_SOLIDITY_EMIT_ARTIFACTS)
+             continue;
+         }
+         const artifactsEmitted = [];
++
+         for (const [contractName, contractOutput] of Object.entries((_b = (_a = output.contracts) === null || _a === void 0 ? void 0 : _a[file.sourceName]) !== null && _b !== void 0 ? _b : {})) {
+             log(`Emitting artifact for contract '${contractName}'`);
++            const richerOutput = {...contractOutput,ast: output.sources[file.sourceName].ast, sourcePath: file.absolutePath}
+             const artifact = await run(task_names_1.TASK_COMPILE_SOLIDITY_GET_ARTIFACT_FROM_COMPILATION_OUTPUT, {
+                 sourceName: file.sourceName,
+                 contractName,
+-                contractOutput,
++                contractOutput: richerOutput,
+             });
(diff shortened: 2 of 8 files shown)


geoknee:
🎉  https://5fcab3e265ab4f000739d540--nitro-protocol.netlify.app/contract-api/natspec/iforcemoveapp/

geoknee:
Observation: here I used `patch-package`, although it may be possible to achieve the same thing through an [overriding ](https://hardhat.org/guides/create-task.html#overriding-tasks)`hardhat`  task.

Observation 2: We have now moved away from netlify cli #2891 and back to the GUI. This means we need to spend a lot more time doing the build on netlify, which is pretty inefficient. There's a lot of scope for optimization.

geoknee:
A new page which will go live when this is merged
https://deploy-preview-3029--nitro-protocol.netlify.app/protocol-tutorial/execute-state-transitions#precautions

andrewgordstewart:
> Observation 2: We have now moved away from netlify cli #2891 and back to the GUI. This means we need to spend a lot more time doing the build on netlify, which is pretty inefficient. There's a lot of scope for optimization.

What's the link to the netlify gui?

geoknee:
https://app.netlify.com/sites/nitro-protocol/overview

andrewgordstewart:
> we need to spend a lot more time doing the build on netlify, which is pretty inefficient. 

Can you clarify what this means? Do you mean we spend more CPU time, or we need to click through the GUI to do something?

geoknee:
More explicitly, we have to build and bootstrap the entire monorepo before running the website generator tool. Previously, we would have already done all of that on circle. We could also be a bit smarter about deciding whether to skip a build because we could use `lerna` and not just git.

andrewgordstewart:
Got it, so it's a computer, not a human, spending more time to generate the website.

patches/hardhat+2.0.3.patch
```diff
@@ -0,0 +1,33 @@
+diff --git a/node_modules/hardhat/builtin-tasks/compile.js b/node_modules/hardhat/builtin-tasks/compile.js
```
andrewgordstewart, line 1: It'd be nice to include a rough explaination in the git commit message here.

netlify.toml
```diff
@@ -1,2 +1,2 @@
 [build]
-  ignore = "git diff --quiet HEAD^ HEAD packages/docs-website/"
\ No newline at end of file
+  ignore = "git diff --quiet HEAD^ HEAD packages/docs-website/ packages/nitro-protocol"
```
andrewgordstewart, line 2: How does this command work? What's its purpose?

andrewgordstewart (approved):

netlify.toml
geoknee: It wires up netlify's ["ignore a build"](https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds) configuration to the exit code of a git command. It's purpose is to skip the build *unless* the docs-website has changed OR nitro-protocol has changed. It should prevent a lot of unnecessary deploy previews being built.

andrewgordstewart: > It wires up netlify's ["ignore a build"](https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds) configuration to the exit code of a git command. It's purpose is to skip the build _unless_ the docs-website has changed OR nitro-protocol has changed. It should prevent a lot of unnecessary deploy previews being built.

Does `.toml` support comments? This could be useful information to live above the `ignore = ` line.

patches/hardhat+2.0.3.patch
geoknee: I added a bit more context in a squashed commit message for the PR.

PR #3037, opened 2020-12-07
[server-wallet] Do not close a channel until no channel is funded by it

snario:
Adds a check before closing a channel to ensure it does not fund any other channel before being closed.

### DIFF
--- a/packages/server-wallet/src/protocols/close-channel.ts
+++ b/packages/server-wallet/src/protocols/close-channel.ts
@@ -1,7 +1,7 @@
 import {checkThat, isSimpleAllocation, State, unreachable} from '@statechannels/wallet-core';
 import {Transaction} from 'knex';
 import {Logger} from 'pino';
-import {isExternalDestination} from '@statechannels/nitro-protocol/src/contract/channel';
+import {isExternalDestination} from '@statechannels/nitro-protocol';
 
 import {Store} from '../wallet/store';
 import {Bytes32} from '../type-aliases';
--- a/packages/server-wallet/src/protocols/close-channel.ts
+++ b/packages/server-wallet/src/protocols/close-channel.ts
@@ -224,13 +224,14 @@ const defundIntoLedger = (ps: ProtocolState): RequestLedgerDefunding | false =>
  * for funding.
  */
 const ensureAllAllocationItemsAreExternalDestinations = (ps: ProtocolState): boolean =>
-  !isDirectlyFunded(ps.app) ||
-  (isDirectlyFunded(ps.app) &&
+  (!isDirectlyFunded(ps.app) && !isFakeFunded(ps.app)) ||
+  ((isDirectlyFunded(ps.app) || isFakeFunded(ps.app)) &&
     !!ps.app.supported &&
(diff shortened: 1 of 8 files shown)


packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -216,6 +217,22 @@ const defundIntoLedger = (ps: ProtocolState): RequestLedgerDefunding | false =>
...
   });
 
+/**
+ * If the channel is directly funded, then ensure none of its allocation items are other
+ * channels being funded by this channel (e.g., if it is a ledger channel). This should
+ * cause the protocol to "pause" / "freeze" until no channel depends on this channel
+ * for funding.
+ */
+const ensureAllAllocationItemsAreExternalDestinations = (ps: ProtocolState): boolean =>
+  (!isDirectlyFunded(ps.app) && !isFakeFunded(ps.app)) ||
```
geoknee, line 227:
```suggestion

```
I believe this line is equivalent to `if (true) then` since direct and fake funding are mutually exclusive ?

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -216,6 +217,22 @@ const defundIntoLedger = (ps: ProtocolState): RequestLedgerDefunding | false =>
...
+  ((isDirectlyFunded(ps.app) || isFakeFunded(ps.app)) &&
```
geoknee, line 228: The name of the function suggests no condition on funding. The docstring suggests the condition should be `isDirectlyFunded` . The  the actual condition seems to be either directly funded or fake funded".

I'd like us to consider here what the long term rule should be for the wallet, and why. Perhaps the rule could be "do not close any channel that allocates to other channels, regardless of funding". (In that case, you should keep the name of the function but remove the conditional logic highlighted).

packages/server-wallet/src/protocols/close-channel.ts
snario: It's only true if it's Ledger funded I think, otherwise false

packages/server-wallet/src/protocols/close-channel.ts
snario: Hm yeah, "do not close any channel that allocates to other channels, regardless of funding" might be better

packages/server-wallet/src/protocols/close-channel.ts
geoknee: Yeah ok, but I think it is still unnecessary given the next line?

Also, why not use your `isLedgerFunded` helper here?

snario: The original intent was for this to only apply to channels which were directly funded. However we still have this legacy "fake" funded channel type, so I added that, and we only have three types in the codebase now, so I suppose it's equivalent to checking if it's not ledger funded but then it kind of defeats the point.

I think the suggestion of just ensuring no internal destination exists inside the outcome for any channel before closing is reasonable.

geoknee (approved):

geoknee:
Can you fix up the PR description before merging?

PR #3038, opened 2020-12-08
Chain service supports basic challenges

kerzhner:
For background, reference https://www.notion.so/Direct-funding-f0fe906bb6fd4a298602b140c3bb0899.

## Features added
- `challenge` ChainServiceAPI.
- `pushOutcomeAndWithdraw` ChainServiceAPI.
- `channelFinalized` ChainEventSubscriber API is called when a channel is finalized.

## Refactor
While adding the above features, it became clear that creating a new Observable subscriptions on every `registerChannel` call is clunky, hard to understand, and, likely, memory-expensive. This PR contains a refactor with:
- One observable per contract. One subscription to that observable.
- `registerChannel` modifies a mapping of channelId to listeners. This should scale to thousands of registered channels.

Follow up work
- [ ] https://github.com/orgs/statechannels/projects/1#card-50847289

### DIFF
--- a/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
+++ b/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
@@ -8,22 +8,28 @@ import {getFixedPart, hashAppPart, hashState, State} from '../state';
 // https://github.com/ethers-io/ethers.js/issues/602#issuecomment-574671078
 const NitroAdjudicatorContractInterface = new utils.Interface(NitroAdjudicatorArtifact.abi);
 
+export type PushOutcomeTransactionArg = {
+  turnNumRecord: number;
+  finalizesAt: number;
+  state: State;
+  outcome: Outcome;
+  channelWasConcluded?: boolean;
+  challengerAddress?: string;
+};
+
 export const createPushOutcomeTransactionFactory = (transferAll: boolean) => (
-  turnNumRecord: number,
-  finalizesAt: number,
-  state: State,
-  outcome: Outcome,
-  channelWasConcluded = false
+  arg: PushOutcomeTransactionArg
 ): providers.TransactionRequest => {
+  const defaults = {channelWasConcluded: false, challengerAddress: constants.AddressZero};
+  const {turnNumRecord, finalizesAt, state, outcome, channelWasConcluded, challengerAddress} = {
+    ...defaults,
+    ...arg,
+  };
   if (channelWasConcluded && turnNumRecord !== 0)
     throw Error('If the channel was concluded, you should use 0 for turnNumRecord');
   const channelId = getChannelId(state.channel);
   const stateHash = channelWasConcluded ? constants.HashZero : hashState(state);
   const {participants} = state.channel;
-  const challengerAddress = channelWasConcluded
-    ? constants.AddressZero
-    : // todo: this is not correct
-      participants[state.turnNum % participants.length];
   const encodedOutcome = encodeOutcome(outcome);
 
   const funcName = transferAll ? 'pushOutcomeAndTransferAll' : 'pushOutcome';
--- a/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
+++ b/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
@@ -23,7 +23,8 @@ export function createPushOutcomeTransaction(
   const {participants} = state.channel;
   const challengerAddress = channelWasConcluded
     ? constants.AddressZero
-    : participants[state.turnNum % participants.length];
+    : // todo: this is not correct
+      participants[state.turnNum % participants.length];
   const encodedOutcome = encodeOutcome(outcome);
 
   const funcName = transferAll ? 'pushOutcomeAndTransferAll' : 'pushOutcome';
--- a/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
+++ b/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
@@ -29,7 +29,6 @@ export const createPushOutcomeTransactionFactory = (transferAll: boolean) => (
     throw Error('If the channel was concluded, you should use 0 for turnNumRecord');
   const channelId = getChannelId(state.channel);
   const stateHash = channelWasConcluded ? constants.HashZero : hashState(state);
-  const {participants} = state.channel;
   const encodedOutcome = encodeOutcome(outcome);
 
   const funcName = transferAll ? 'pushOutcomeAndTransferAll' : 'pushOutcome';
--- a/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
+++ b/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
@@ -13,7 +13,8 @@ export function createPushOutcomeTransaction(
   finalizesAt: number,
   state: State,
   outcome: Outcome,
-  channelWasConcluded = false
+  channelWasConcluded = false,
+  transferAll = false
 ): providers.TransactionRequest {
   if (channelWasConcluded && turnNumRecord !== 0)
     throw Error('If the channel was concluded, you should use 0 for turnNumRecord');
@@ -25,7 +26,8 @@ export function createPushOutcomeTransaction(
     : participants[state.turnNum % participants.length];
   const encodedOutcome = encodeOutcome(outcome);
 
-  const data = NitroAdjudicatorContractInterface.encodeFunctionData('pushOutcome', [
+  const funcName = transferAll ? 'pushOutcomeAndTransferAll' : 'pushOutcome';
+  const data = NitroAdjudicatorContractInterface.encodeFunctionData(funcName, [
     channelId,
     turnNumRecord,
     finalizesAt,
--- a/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
+++ b/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
@@ -8,14 +8,13 @@ import {getFixedPart, hashAppPart, hashState, State} from '../state';
 // https://github.com/ethers-io/ethers.js/issues/602#issuecomment-574671078
 const NitroAdjudicatorContractInterface = new utils.Interface(NitroAdjudicatorArtifact.abi);
 
-export function createPushOutcomeTransaction(
+export const createPushOutcomeTransactionFactory = (transferAll: boolean) => (
   turnNumRecord: number,
   finalizesAt: number,
   state: State,
   outcome: Outcome,
-  channelWasConcluded = false,
-  transferAll = false
-): providers.TransactionRequest {
+  channelWasConcluded = false
+): providers.TransactionRequest => {
   if (channelWasConcluded && turnNumRecord !== 0)
     throw Error('If the channel was concluded, you should use 0 for turnNumRecord');
   const channelId = getChannelId(state.channel);
@@ -38,7 +37,7 @@ export function createPushOutcomeTransaction(
   ]);
 
   return {data};
-}
+};
 
 export function concludePushOutcomeAndTransferAllArgs(
   states: State[],
--- a/packages/server-wallet/.vscode/launch.json
+++ b/packages/server-wallet/.vscode/launch.json
@@ -38,8 +38,8 @@
       "autoAttachChildProcesses": false,
       "internalConsoleOptions": "neverOpen",
       "env": {
-        "NODE_ENV": "test"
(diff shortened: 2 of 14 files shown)


packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
```diff
@@ -13,7 +13,8 @@ export function createPushOutcomeTransaction(
   finalizesAt: number,
   state: State,
   outcome: Outcome,
-  channelWasConcluded = false
+  channelWasConcluded = false,
+  transferAll = false
```
snario, line 17: It seems like this is passing on some additional logic into the `nitro-protocol` package that we may not want to put in there. Should it be the `nitro-protocol` packages job to decide what transaction to construct or the `ChainService`'s?

Some points worth considering:
- The method is named `createPushOutcomeTransaction`, but sometimes returns a `pushOutcomeAndTransferAll` transaction.
- The `ChainService` has all of the information and context it needs to decide which transaction to construct; it could decide to call `createPushOutcomeTransaction`, `createPushOutcomeAndTransferAllTransaction`, or `createConcludeAndPushOutcomeAndTransferAllTransaction`, etc based on configured options / logic inside the `ChainService`, and the logic inside `nitro-protocol` could be simplified.

Ultimately seems like a question about separation of concerns that could go either way. Seems to me like it's "cleaner" to have the `nitro-protocol` expose TS helpers to simplify making known external API calls on the `NitroAdjudicator` contract as opposed to abstracting those methods, but that's just my two cents

packages/server-wallet/.vscode/launch.json
```diff
@@ -39,6 +39,7 @@
       "internalConsoleOptions": "neverOpen",
       "env": {
         "NODE_ENV": "test"
+        //"SHOW_VERBOSE_GANACHE_OUTPUT": "true"
```
snario, line 42:
```suggestion
        "SHOW_VERBOSE_GANACHE_OUTPUT": "false"
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -245,6 +257,47 @@ export class ChainService implements ChainServiceInterface {
     return transactionResponse;
   }
 
+  async challenge(
+    challengeStates: SignedState[],
+    privateKey: PrivateKey
```
snario, line 252: Probably not something we want or need to do now, but perhaps worth considering and making an issue for — it's not great from a security perspective to be passing around this private key in-memory too often. If we split out the `ChainService` into a separate process, or even on a different host, then this will lead to the key being passed over the network. Maybe the challenge message could be something that is easily sign-able within the `Store` for example, maybe?

packages/server-wallet/.vscode/launch.json
kerzhner, line 42: @snario, good call, but need a comma on the line above
```suggestion
        "NODE_ENV": "test",
        "SHOW_VERBOSE_GANACHE_OUTPUT": "false"
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -245,6 +257,47 @@ export class ChainService implements ChainServiceInterface {
...
+  ): Promise<providers.TransactionResponse> {
+    if (!challengeStates.length) {
+      throw new Error('Must challenge with at least one state');
+    }
+
```
snario, line 257: It might be useful to add an `INFO` log here with the `challengeStates` being used

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Thanks for flagging this https://github.com/statechannels/statechannels/issues/3044

packages/server-wallet/src/chain-service/types.ts
```diff
@@ -45,6 +50,7 @@ interface ChainModifierInterface {
   concludeAndWithdraw(
     finalizationProof: SignedState[]
   ): Promise<providers.TransactionResponse | void>;
+  pushOutcomeAndWithdraw(state: State): Promise<providers.TransactionResponse>;
```
snario, line 60: It's worth discussing whether or not there should just be a single API on the `ChainService` for both of these cases, something equivalent to the wallet saying "here is everything I have, please get the money out of this channel", and then letting the `ChainService` decide what to do when it gets called. The benefit would be that the `Wallet` would not need to be as concerned about what is on-chain and can rely on the `ChainService` to figure out any of those details.

packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
kerzhner: I think it's a good call to clean up the Nitro protocol API.

packages/server-wallet/src/chain-service/chain-service.ts
andrewgordstewart: Why INFO and not ERROR?

packages/server-wallet/src/wallet/__test__/fixtures/participants.ts
```diff
@@ -1,24 +1,24 @@
-import {Participant, makeDestination} from '@statechannels/wallet-core';
+import {Participant, makeDestination, makeAddress} from '@statechannels/wallet-core';
 
 import * as wallets from './signing-wallets';
 import {fixture} from './utils';
 
 const _alice: Participant = {
-  signingAddress: wallets.alice().address,
+  signingAddress: makeAddress(wallets.alice().address),
```
snario, line 7: Curious why this is necessary since `SigningWallet.address` is of type `Address`?

kerzhner: Good call, I did this when I was hitting a bizarre stress test error. Fixed in 6c1b1a05d

packages/server-wallet/src/chain-service/chain-service.ts
snario: I think the line I chose to comment on just happened to include 265 within the 4 lines GitHub shows. I mean to say there should be an INFO log in the empty space on line 267. It would probably also be useful to log on line 265.

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -255,110 +298,140 @@ export class ChainService implements ChainServiceInterface {
...
+    if (nowMs < finalizesAtMs) {
+      setTimeout(() => this.whenFinalized(channelId, finalizesAtS, 0), nowMs - finalizesAtMs);
+      return;
     }
-    subscriptions.map(s => s.unsubscribe());
+    const [, storageFinalizesAtS] = await this.nitroAdjudicator.getChannelStorage(channelId);
+    const blockTimeS = (await this.provider.getBlock(await this.provider.getBlockNumber()))
+      .timestamp;
+    if (blockTimeS >= storageFinalizesAtS) {
+      // Should we wait for 6 blocks before emitting the finalized event?
```
snario, line 334: It seems like this is a general purpose configuration option on the `ChainService` that should be respected whenever a chain re-org occurring is important to detect before making any irreversible decisions. So, IMO, we should use `this.blockConfirmations` in situations like this one.

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -255,110 +298,140 @@ export class ChainService implements ChainServiceInterface {
...
+      // Will the wallet sign new states or deposit into the channel based on this event?
+      // The answer is likely no.
+      // So we probably want to emit this event as soon as possible.
```
snario, line 337: I think it's true that the _wallet_ might not, but it may emit an event that states that the channel's status is now `closed` officially, which the _app_ might take some action on.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: In general, I agree with you. But what is an example of an irreversible decision that would be affected by a chain reorg, in this case?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -255,110 +298,140 @@ export class ChainService implements ChainServiceInterface {
...
+
+  private async whenFinalized(channelId: Bytes32, finalizesAtS: number, tryNum: number) {
+    if (tryNum > 4) return;
+
+    const nowMs = Date.now();
+    const finalizesAtMs = finalizesAtS * 1_000;
+    if (nowMs < finalizesAtMs) {
+      setTimeout(() => this.whenFinalized(channelId, finalizesAtS, 0), nowMs - finalizesAtMs);
+      return;
     }
```
snario, line 329: It seems to me like we can leverage the fact that the `ethers` library exposes `provider.on('block', block => ...)` which emits each time a new block is produced. Instead of relying on `Date.now()` we could rely on `block.timestamp`

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Maybe we should have 2 chain service notifications: channel finalized and channel finalized confirmed (that is fired after 6 blocks)?

It seems to me that it is beneficial for the wallet to get a notification as soon as possible that a channel has been finalized.

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: I think that's an interesting point. But what if a challenge is 2 weeks? Do you still want to fire a callback on every block?

kerzhner: That approach is probably ok as long as there is one global callback instead of one callback per channel.

snario: Considering that the `ChainService` is already inspecting every block (e.g., for `Deposited` events), if it's just doing a single additional comparison operation, I don't think it's adding any meaningful overhead in practice. Arguably the code simplification benefits might outweigh the extra check

kerzhner: Good point!

packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
kerzhner: da452420e

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Addressed in 0d194d4b9.

packages/server-wallet/src/chain-service/types.ts
kerzhner: That's a good point! https://github.com/statechannels/statechannels/issues/3048

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Addressed in 974b971fa

snario (approved):

kerzhner:
> nit: I don't love having both `_onBlockMined` and `onBlockMined` which are both `private`

@lalexgap agreed! fbeec9031

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -255,110 +299,140 @@ export class ChainService implements ChainServiceInterface {
...
-    const subscriptions = this.channelToSubscription.get(channelId);
-    if (subscriptions?.length !== 1) {
-      throw new Error(
-        'Unregister channel implementation only works when there is one subscriber per channel'
-      );
+    this.channelToSubscribers.set(channelId, []);
+  }
+
+  private async whenFinalized(channelId: Bytes32, finalizesAtS: number, tryNum: number) {
+    if (tryNum > 4) return;
```
lalexgap: Does this ever get used? It seem like we're not incrementing the `tryNum`?

kerzhner: That code has been removed in https://github.com/statechannels/statechannels/pull/3038/commits/0d194d4b92ca7a3fbc4a52245b43f5ab19101009

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -152,7 +159,7 @@ export class ChainService implements ChainServiceInterface {
   }
 
   async fundChannel(arg: FundChannelArg): Promise<providers.TransactionResponse> {
-    this.logger.info({...arg}, 'Attempting to fund channel');
+    this.logger.info({...arg}, 'ChainService.fundChannel');
```
lalexgap, line 162: nit: I don't think we need to include `ChainService` in the string since the `logger` has the module set to `ChainService` already: `logger.child({module: 'ChainService'})`.

lalexgap (approved):
Looks good to me!

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: Good call ce7244d5e

Issue #3035, opened 2020-12-07
[audit] 6.1, 6.3, 6.4 Fix validTransition


andrewgordstewart:
[Explaination required] See https://github.com/statechannels/statechannels/issues/2738 for some context.

- [x] add a check for channelNonce inside validateTransition #3039 
- [x] incorporate turnNum check inside `_requireExtraImplicitChecks(fromState, toState)` ;#3041
- [x] introduce `requireValidProtocolTransition` typescript emulator for `ForceMove.sol` function (with cross-language test) #3041 
- [ ] clearly separate:
      - transition validation (protocol, app); 
      - inference of a support proof of a state in the `store`; 
      - garbage collection / early discard of unwanted states during `pushMessage`.

geoknee:
> 6.1 server-wallet - validateTransition fails to return on turn number check failed Major
6.3 sever-wallet - validateTransition is missing a check for channelNonce Major
6.4 server-wallet - validateTransition checks differ from ForceMove._requireValidTransition Major
I think we'll take the recommendation to call out to the NitroAdjudicator contract for the `validTransition` call.
This would be a fairly simple refactor, but it may have throughput & latency consequences.
If we do it, it will tick all of these three

geoknee:
Some observations about calling out to the NitroAdjudicator contract:
- it hooks an external API (e.g. infura) into our critical path. Even if it is usually fast and reliable, it may not always be and we can't control that. If there are other possibilities, such as running the EVM code in our client we should explore those first. We already do this but in a way that is easy to become (and actually is) out of sync.
- the on-chain `validTransition` function works in tandem with other on-chain functions to ensure a state is supported. Some of the checks are done implicitly as a part of the signature validation. So we either need to a) rewrite some of the implicit checks in our client code b) refactor the on chain code to do some more (redundant) explicit checks at a higher gas cost or c) refactor the client code to perform the signature validation and transition validation together in the same implicit manner. Option a) might be workable especially if we protect ourselves with a comprehensive test suite.

geoknee:
Pinching off the remaining issue into #3137

PR #3041, opened 2020-12-08
refactor: un-nest requireValidTransition

geoknee:
Preparatory work for tackling #3035 

- It is now a very thin wrapper around two pure functions, avoiding the need for cross-contract calls 
- The branching logic in the wrapper should be simple to emulate in other languages and _is_ emulated in the `server-wallet` - The pure functions can be emulated separately. My strategy here is to use PureEVM for the app check, and write a typescript function that manually emulates the protocol check. The justification for this being that the protocol check is unlikely to change frequently and also easy to make an exact match for. 
- the typescript function in question has been placed in `nitro-protocol` and has an accompanying test suite. 
- this is imported into `server-wallet`, where it replaces the existing function. 
- side effects of this: I had to move the special logic around setup states into the existing `shouldValidateTransition` function. 

This is good because no such special case exists on chain and I want to keep the transition validation as close to the solidity code as possible. However, this also means that *no* checks at all are performed in the setup phase (so the app definition could change, for example). I think this may be an acceptable regression for now.


packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -547,25 +547,25 @@ contract ForceMove is IForceMove {
         return stateHashes;
     }
 
+    enum IsValidTransition {True, NeedToCheckApp}
```
lalexgap, line 550: nit: `TransitionIsValid` (or something) might be a clearer name for the `True` enum?

lalexgap (approved):
This looks great to me and I'd be happy to get it merged in!!

Like you mentioned in standup we should probably think about when we validate a state vs when we store a state, but that can probably take place in a separate scope of work (after we spec it out in a notion doc)?

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -547,25 +547,25 @@ contract ForceMove is IForceMove {
...
     /**
     * @notice Check that the submitted pair of states form a valid transition
     * @dev Check that the submitted pair of states form a valid transition
     * @param nParticipants Number of participants in the channel.
     transition
     * @param isFinalAB Pair of booleans denoting whether the first and second state (resp.) are final.
     * @param ab Variable parts of each of the pair of states
-    * @param turnNumB turnNum of the later state of the pair.
-    * @param appDefinition Address of deployed contract containing application-specific validTransition function.
+    * @param turnNumB turnNum of the later state of the pai
```
geoknee, line 559:
```suggestion
    * @param turnNumB turnNum of the later state of the pair
```

packages/nitro-protocol/src/valid-transition.ts
```diff
@@ -0,0 +1,71 @@
...
+export enum Status {
+  True,
+  NeedToCheckApp,
+}
+
+/**
+ *  Port of ForceMove#_requireValidProtocolTransition solidity code
+ * either:
+ * - returns Status.True (no further checks required)
+ * -returns  Status.NeedToCheckApp (clears protocol checks, overall validity requires App.validTransition)
```
geoknee, line 14:
```suggestion
 * - returns  Status.NeedToCheckApp (clears protocol checks, overall validity requires App.validTransition)
```

PR #3045, opened 2020-12-08
Close channel protocol tests

tomclose:
This PR updates the close channel protocol tests, so they use the same approach as the open channel protocol tests (see https://github.com/statechannels/statechannels/pull/2995). Having the tests in this format will allow us to refactor the close channel protocols. 

It also makes a few improvements to the `TestChannel` class:

1. You can now set the `fundingStrategy` on a test channel
2. test channels can now insert themselves into a store
3. when creating a test channel, you can set `finalFrom: 10`, which will e.g. cause `state(11)` to return a double-signed 10

### DIFF
--- a/packages/server-wallet/src/handlers/__test__/single-app-updater.test.ts
+++ b/packages/server-wallet/src/handlers/__test__/single-app-updater.test.ts
@@ -8,7 +8,8 @@ import {TestChannel} from '../../wallet/__test__/fixtures/test-channel';
 import {TestLedgerChannel} from '../../wallet/__test__/fixtures/test-ledger-channel';
 import {SingleAppUpdater} from '../single-app-updater';
 
-const testChan = TestChannel.create({channelNonce: 1});
+const FINAL = 8;
+const testChan = TestChannel.create({channelNonce: 1, startClosingAt: FINAL});
 const testChan2 = TestChannel.create({channelNonce: 2});
 
 let store: Store;
@@ -23,6 +24,7 @@ beforeEach(async () => {
     '0'
   );
   await store.dbAdmin().truncateDB();
+  await store.dbAdmin().migrateDB();
   singleAppUpdater = SingleAppUpdater.create(store);
   response = WalletResponse.initialize();
 });
@@ -70,21 +72,24 @@ describe('app channel not running', () => {
     );
   });
   it(`errors if the channel has a final state`, async () => {
-    // put state5 and state6final in the store
-    await setup({states: [testChan.wirePayload(5), testChan.wirePayload(6, [5, 5], 'final')]});
+    // put state7 and state8final in the store
+    await setup({
+      states: [testChan.wirePayload(FINAL - 1), testChan.wirePayload(FINAL, [5, 5])],
+    });
 
-    // send state7final
+    // send state9final
     await expect(
-      singleAppUpdater.update(testChan.wirePayload(7, [5, 5], 'final'), response)
+      singleAppUpdater.update(testChan.wirePayload(FINAL + 1, [5, 5]), response)
     ).rejects.toThrow('The update sent to pushUpdate must be for a running channel');
   });
-  it(`errors if given a final state`, async () => {
-    // put state5 and state6final in the store
-    await setup({states: [testChan.wirePayload(5), testChan.wirePayload(6, [5, 5], 'final')]});
+  it(`errors if given a first final state`, async () => {
+    await setup({
+      states: [testChan.wirePayload(FINAL - 2), testChan.wirePayload(FINAL - 1, [5, 5])],
+    });
 
     // send state7final
     await expect(
-      singleAppUpdater.update(testChan.wirePayload(7, [5, 5], 'final'), response)
+      singleAppUpdater.update(testChan.wirePayload(FINAL, [5, 5]), response)
     ).rejects.toThrow('The update sent to pushUpdate must be for a running channel');
   });
 });
--- a/packages/server-wallet/src/handlers/__test__/single-app-updater.test.ts
+++ b/packages/server-wallet/src/handlers/__test__/single-app-updater.test.ts
@@ -9,7 +9,7 @@ import {TestLedgerChannel} from '../../wallet/__test__/fixtures/test-ledger-chan
 import {SingleAppUpdater} from '../single-app-updater';
 
 const FINAL = 8;
-const testChan = TestChannel.create({channelNonce: 1, startClosingAt: FINAL});
+const testChan = TestChannel.create({channelNonce: 1, finalFrom: FINAL});
 const testChan2 = TestChannel.create({channelNonce: 2});
 
 let store: Store;
--- a/packages/server-wallet/src/protocols/__test__/close-channel.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/close-channel.test.ts
@@ -1,70 +1,151 @@
-import {simpleEthAllocation, BN, State} from '@statechannels/wallet-core';
-
-import {protocol} from '../close-channel';
-import {alice, bob} from '../../wallet/__test__/fixtures/participants';
-import {SignState} from '../actions';
+import {testKnex as knex} from '../../../jest/knex-setup-teardown';
+import {MockChainService} from '../../chain-service';
+import {defaultTestConfig} from '../../config';
+import {createLogger} from '../../logger';
 import {ChainServiceRequest, requestTimeout} from '../../models/chain-service-request';
-import {channel} from '../../models/__test__/fixtures/channel';
+import {DBCloseChannelObjective} from '../../models/objective';
+import {Store} from '../../wallet/store';
+import {WalletResponse} from '../../wallet/wallet-response';
+import {TestChannel} from '../../wallet/__test__/fixtures/test-channel';
+import {ChannelCloser} from '../close-channel';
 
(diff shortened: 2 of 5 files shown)


packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
```diff
@@ -65,31 +73,31 @@ export class TestChannel {
...
    */
-  public state(n: number, bals?: [number, number], final = ''): State {
+  public state(n: number, bals?: [number, number]): State {
     return {
       ...this.channelConstants,
       appData: '0x',
-      isFinal: final === 'final',
-      turnNum: n,
+      isFinal: !!this.finalFrom && n >= this.finalFrom,
+      turnNum: Math.min(n, this.finalFrom || n),
```
geoknee, line 81: I think an explanatory comment is called for?

```suggestion
      turnNum: Math.min(n, this.finalFrom || n), // test channels adopt a countersigning strategy for final states, so the turn number doesn't progress after finalFrom.
```

packages/server-wallet/src/handlers/__test__/single-app-updater.test.ts
```diff
@@ -8,7 +8,8 @@ import {TestChannel} from '../../wallet/__test__/fixtures/test-channel';
 import {TestLedgerChannel} from '../../wallet/__test__/fixtures/test-ledger-channel';
 import {SingleAppUpdater} from '../single-app-updater';
 
-const testChan = TestChannel.create({channelNonce: 1});
+const FINAL = 8;
+const testChan = TestChannel.create({channelNonce: 1, finalFrom: FINAL});
```
geoknee, line 12: I think 
```suggestion
const finalFrom = 8;
const testChan = TestChannel.create({channelNonce: 1, finalFrom});
```
(and so on below) is clearer.

packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
```diff
@@ -39,6 +44,8 @@ export class TestChannel {
   public signingWalletB: SigningWallet = bobWallet();
   public startBals: [number, number];
   public channelNonce: number;
+  public finalFrom?: number;
```
geoknee, line 47: Can you put a docstring above `TestChannelArgs` to help explain `finalFrom`? Something like 

```ts
/**
 * Arguments for constructing a TestChannel
 *
 * @param finalFrom - a turnnumber where the channel bceoms finalized. All subsequent states willl be final and have this turn number
 */
interface TestChannelArgs {
  aBal?: number;
  bBal?: number;
  channelNonce?: number;
  finalFrom?: number;
  fundingStrategy?: FundingStrategy;
}

```

packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
```diff
@@ -172,6 +191,46 @@ export class TestChannel {
   static get emptyPayload(): Payload {
     return {walletVersion: WALLET_VERSION};
   }
+
+  public get startBal(): number {
+    return this.startBals.reduce((s, n) => s + n);
+  }
+
+  public async insertInto(store: Store, args: InsertionParams): Promise<DBOpenChannelObjective> {
```
geoknee, line 199: Again a docstring would be nice, explaining all the things that happen when this method is called. 
```suggestion
/**
* Calls addSigningKey, pushMessage, updateFunding, ensureObjective and approveObjective on the supplied store.
*/
  public async insertInto(store: Store, args: InsertionParams): Promise<DBOpenChannelObjective> {
```

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
```diff
@@ -1,70 +1,151 @@
...
-import {channel} from '../../models/__test__/fixtures/channel';
+import {DBCloseChannelObjective} from '../../models/objective';
+import {Store} from '../../wallet/store';
+import {WalletResponse} from '../../wallet/wallet-response';
+import {TestChannel} from '../../wallet/__test__/fixtures/test-channel';
+import {ChannelCloser} from '../close-channel';
 
-import {applicationProtocolState} from './fixtures/protocol-state';
+const FINAL = 10; // this will be A's state to sign
+const testChan = TestChannel.create({aBal: 5, bBal: 5, finalFrom: FINAL});
```
geoknee, line 13:
```suggestion
const finalFrom = 10; // this will be A's state to sign
const testChan = TestChannel.create({aBal: 5, bBal: 5, finalFrom});
```
and so on

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
```diff
@@ -1,70 +1,151 @@
...
+    defaultTestConfig().metricsConfiguration.timingMetrics,
+    defaultTestConfig().skipEvmValidation,
+    '0'
+  );
+  await store.dbAdmin().truncateDB();
+});
 
-const runningState = {outcome, turnNum: 7, participants};
-const closingState2 = {outcome, turnNum: 8, isFinal: true, participants};
+afterEach(async () => await store.dbAdmin().truncateDB());
```
geoknee, line 27: Do you need to truncate before _and_ after?

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
```diff
@@ -1,70 +1,151 @@
...
 
-test('when I have signed a final state, unfunded', () => {
-  const ps = applicationProtocolState({
-    app: {
-      supported: closingState,
-      latest: closingState,
-      latestSignedByMe: closingState,
-      directFundingStatus: 'Uncategorized',
-    },
+// TODO: when should it complete the objective?
```
geoknee, line 82: When should this be todone?

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
```diff
@@ -1,70 +1,151 @@
...
-  ${closingState}  | ${closingState}  | ${closingState}  | ${undefined}         | ${withdrawAction}           | ${'when the channel is closed'}                                       | ${'submits withdraw transaction'}
-  ${closingState}  | ${closingState}  | ${closingState}  | ${[validCSR]}        | ${undefined}                | ${'when the channel is closed + valid chain service request'}         | ${'no action'}
-  ${closingState}  | ${closingState}  | ${closingState}  | ${[staleCSR]}        | ${withdrawAction}           | ${'when the channel is closed + stale chain service request'}         | ${'sumbits a withdraw transaction'}
-`('$result $cond', ({supported, latest, latestSignedByMe, chainServiceRequests, action}) => {
-  const ps = applicationProtocolState({
-    app: {supported, latest, latestSignedByMe, chainServiceRequests},
+      // then let's "wait" for 10 minutes for the fund request to the chain service to get stale
+      await ChainServiceRequest.query(knex)
+        .findOne({channelId: testChan.channelId})
+        .patch({timestamp: new Date(Date.now() - requestTimeout - 10)});
```
geoknee, line 58: Is this 10 minutes, or 10 milliseconds?

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
```diff
@@ -1,70 +1,151 @@
...
 });
 
-const staleCSR = ChainServiceRequest.fromJson({
-  channelId: 0,
-  request: 'withdraw',
-  timestamp: new Date(Date.now() - requestTimeout - 1),
-});
+describe(`defunding phase (when the channel is closed)`, () => {
+  describe(`direct funding`, () => {
+    it('submits the withdrawal transaction', async () => {
```
geoknee, line 46: The description should reflect the assertions better. 
```suggestion
    it('submits the withdrawal transaction, and submits once again when it becomes stale, but not a third time', async () => {
```
^ is this what you are actually testing? why do we have that behaviour?

geoknee (approved):

packages/server-wallet/src/handlers/__test__/single-app-updater.test.ts
tomclose: I'm not sure about this.

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
tomclose: Not sure. I'll change the `afterEach` to an `afterAll`, so it does it before each test and once after all the tests.

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
tomclose: I'm not sure why this is the behaviour. All I've done here is replicate the test the @kerzhner added for open channel test.

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
tomclose: ✂️

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
kerzhner: Good call on a testcase description that does not match the logic. At the moment, the wallet attempts each chain service request (for each channel) twice: https://github.com/statechannels/statechannels/pull/3001.

Happy to reconsider this heuristic!

PR #3050, opened 2020-12-09
Check finalization status when registering a channel

### LINKED ISSUE #3043: Chain service should check for pending challenges and finalized channels when a channel is registered

lalexgap:
Fixes #3043 

This updates the `ChainService` to check whether a channel is finalized/finalizing when the channel is registered with chain service. 

Currently, it's enough to just call `getChannelStorage` and check the `finalizationAt` value on there. 

When challenging becomes more fully-featured we'll want to also look at looking for `Challenge` events for the channel. Since we're not handling responding to a challenge right now I considered that out of scope.

### DIFF
--- a/packages/nitro-protocol/src/index.ts
+++ b/packages/nitro-protocol/src/index.ts
@@ -4,6 +4,8 @@ import TokenArtifact from '../artifacts/contracts/Token.sol/Token.json';
 import AssetHolderArtifact from '../artifacts/contracts/AssetHolder.sol/AssetHolder.json';
 import Erc20AssetHolderArtifact from '../artifacts/contracts/ERC20AssetHolder.sol/ERC20AssetHolder.json';
 import EthAssetHolderArtifact from '../artifacts/contracts/ETHAssetHolder.sol/ETHAssetHolder.json';
+import TestNitroAdjudicatorArtifact from '../artifacts/contracts/test/TESTNitroAdjudicator.sol/TESTNitroAdjudicator.json';
+import TestAssetHolderArtifact from '../artifacts/contracts/test/TESTAssetHolder.sol/TESTAssetHolder.json';
 
 export const ContractArtifacts = {
   NitroAdjudicatorArtifact,
@@ -12,6 +14,16 @@ export const ContractArtifacts = {
   EthAssetHolderArtifact,
   TokenArtifact, // TODO do we want to export this?
   AssetHolderArtifact, // TODO do we want to export this?
+};
+
+/**
+ * Various test contract artifacts used for testing
+ * They expose helper functions to allow for easier testing
+ * They should NEVER be used in a production environment.
+ */
+export const TestContractArtifacts = {
+  TestNitroAdjudicatorArtifact,
+  TestAssetHolderArtifact,
 };
 
 export {
--- a/packages/nitro-protocol/src/index.ts
+++ b/packages/nitro-protocol/src/index.ts
@@ -17,7 +17,7 @@ export const ContractArtifacts = {
 };
 
 /**
- * Various test contract artifacts used for testing
+ * Various test contract artifacts used for testing.
  * They expose helper functions to allow for easier testing
  * They should NEVER be used in a production environment.
  */
--- a/packages/nitro-protocol/src/index.ts
+++ b/packages/nitro-protocol/src/index.ts
@@ -18,7 +18,7 @@ export const ContractArtifacts = {
 
 /**
  * Various test contract artifacts used for testing.
- * They expose helper functions to allow for easier testing
+ * They expose helper functions to allow for easier testing.
  * They should NEVER be used in a production environment.
  */
 export const TestContractArtifacts = {
--- a/packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
+++ b/packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
@@ -213,8 +213,8 @@ describe('fundChannel', () => {
 describe('registerChannel', () => {
   it('dispatches a channel finalized event if the channel has been finalized BEFORE registering', async () => {
     const channelId = randomChannelId();
-    const CHALLENGE_EXPIRE_TIME = 6000;
-    const FUTURE_TIME = 7000;
+    const CHALLENGE_EXPIRE_TIME = 2_000_000_000;
+    const FUTURE_TIME = 2_000_500_000;
 
     const tx = await testAdjudicator.functions.setChannelStorageHash(
       channelId,
@@ -240,11 +240,13 @@ describe('registerChannel', () => {
     expect(channelFinalizedHandler).toHaveBeenCalledWith({channelId});
   });
 
-  it('registers a channel in the finalizing channel list and fires an event when that channel is fainlized', async () => {
+  it('registers a channel in the finalizing channel list and fires an event when that channel is finalized', async () => {
     const channelId = randomChannelId();
-    const CURRENT_TIME = 5000;
-    const CHALLENGE_EXPIRE_TIME = 6000;
-    const FUTURE_TIME = 7000;
+    // We use large values so we don't have to worry about ganache
+    // mining a block with the current timestamp setting off the expiry
+    const CURRENT_TIME = 2_000_000_000;
+    const CHALLENGE_EXPIRE_TIME = 2_000_400_000;
+    const FUTURE_TIME = 2_000_800_000;
 
     const tx = await testAdjudicator.setChannelStorageHash(
       channelId,
--- a/packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
+++ b/packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
@@ -1,3 +1,5 @@
+import {setTimeout} from 'timers';
+
 import {ETHERLIME_ACCOUNTS} from '@statechannels/devtools';
 import {
   channelDataToChannelStorageHash,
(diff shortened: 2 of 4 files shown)


packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -198,6 +210,29 @@ describe('fundChannel', () => {
 });
 
 describe('registerChannel', () => {
+  it('dipatches a channel finalized event if the channel has been finalized BEFORE registering', async () => {
```
lalexgap, line 213: I only tested the `Finalized` case as the `In Progress` case can be fragile to test due to timing issues.  I'd really hate to introduce another blinking test.

andrewgordstewart: I think there ought to be an issue tracking that. While I recognize the complexity, it would be good to have a solid test suite providing 100% coverage (in the sense of 100% of the chain's state)

Barring that, we ought to have a really good understanding of _why_ it will work in the `In Progress` case.

lalexgap: I've added a test in db07285 by creating a `TestChainService` that can expose the `finalizingChannels` property.

This allows us to ensure that an "In progress" channel gets added to that list.

kerzhner: There is a way to specify the timestamp with `evm_mine` rpc (`await provider.send('evm_mine', [timestampInSeconds]);`. We can:
1. Set channel storage to finalize the channel far in the future.
1. Register the channel.
1. Mine a block with a timestamp that is further in the future.
1. Check that the `channelFinalized` callback has fired.

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -73,7 +73,7 @@ export class ChainService implements ChainServiceInterface {
   private readonly blockConfirmations: number;
   private transactionQueue = new PQueue({concurrency: 1});
 
-  private finalizingChannels: {finalizesAtS: number; channelId: Bytes32}[] = [];
+  protected finalizingChannels: {finalizesAtS: number; channelId: Bytes32}[] = [];
```
kerzhner, line 76: See my comment around `evm_mine` on how we can avoid upgrading `finalizingChannels` to `protected`

packages/nitro-protocol/src/index.ts
```diff
@@ -14,6 +16,16 @@ export const ContractArtifacts = {
...
 };
 
+/**
+ * Various test contract artifacts used for testing
+ * They expose helper functions to allow for easier testing
+ * They should NEVER be used in a production environment.
+ */
+export const TestContractArtifacts = {
+  TestNitroAdjudicatorArtifact,
+  TestAssetHolderArtifact,
```
kerzhner, line 26: Is this used anywhere?

kerzhner (changes requested):

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -53,14 +62,21 @@ async function mineBlockPeriodically(blocks: number) {
...
+  new providers.JsonRpcProvider(rpcEndpoint).getSigner(1)
+);
+beforeAll(async () => {
   // Try to use a different private key for every chain service instantiation to avoid nonce errors
   // Using the first account here as that is the one that:
   // - Deploys the token contract.
   // - And therefore has tokens allocated to it.
   /* eslint-disable no-process-env */
-  chainService = new ChainService({
+  chainService = new TestChainService({
```
kerzhner, line 79: Might not be needed. See `evm_mine` comment.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -198,6 +214,58 @@ describe('fundChannel', () => {
 });
 
 describe('registerChannel', () => {
+  it('dipatches a channel finalized event if the channel has been finalized BEFORE registering', async () => {
```
kerzhner, line 217:
```suggestion
  it('dispatches a channel finalized event if the channel has been finalized BEFORE registering', async () => {
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -356,6 +360,39 @@ export class ChainService implements ChainServiceInterface {
...
+
+  private async getFinalizedStatus(
+    channelId: string
+  ): Promise<
+    | {status: 'Finalized'; finalizedAtS: number}
+    | {status: 'In Progress'; finalizedAtS: number}
+    | {status: 'Not Finalized'}
+  > {
+    const latestBlock = await this.provider.getBlock(this.provider.getBlockNumber());
+    const [, finalizesAt] = await this.nitroAdjudicator.getChannelStorage(channelId);
```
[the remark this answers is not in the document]
lalexgap: I think we should probably split handling chain reorg into a separate piece of work? I think there's a lot of complexity there that will apply to a bunch of places in the chain service?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -356,6 +360,39 @@ export class ChainService implements ChainServiceInterface {
     this.finalizingChannels.sort((a, b) => a.finalizesAtS - b.finalizesAtS);
   }
 
+  private async registerFinalizationStatus(channelId: string): Promise<void> {
+    const result = await this.getFinalizedStatus(channelId);
+    if (result.status === 'In Progress') {
+      this.addFinalizingChannel(channelId, result.finalizedAtS);
+    } else if (result.status === 'Finalized') {
```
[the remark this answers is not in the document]
lalexgap: I was thinking about that approach. I was thinking that the `In progress` event sets us up nicely for challenging but perhaps I should just add this later.

packages/server-wallet/src/chain-service/chain-service.ts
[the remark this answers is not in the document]
kerzhner: Agreed!

packages/nitro-protocol/src/index.ts
lalexgap: `TestNitroAdjudicatorArtifact` is used in the tests but may not be needed if we can just set `evm_mine` timestamp

packages/server-wallet/src/chain-service/chain-service.ts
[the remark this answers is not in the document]
lalexgap: I was thinking about doing this. I've implemented that in a42519e

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
lalexgap: Done in 2f6b4fd

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -356,6 +360,39 @@ export class ChainService implements ChainServiceInterface {
     this.finalizingChannels.sort((a, b) => a.finalizesAtS - b.finalizesAtS);
   }
 
+  private async registerFinalizationStatus(channelId: string): Promise<void> {
+    const result = await this.getFinalizedStatus(channelId);
+    if (result.status === 'In Progress') {
+      this.addFinalizingChannel(channelId, result.finalizedAtS);
```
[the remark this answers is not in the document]
lalexgap: Short term check in c2445a4

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -198,6 +212,69 @@ describe('fundChannel', () => {
...
+        assetTransferred: _.noop,
+        channelFinalized: arg => {
+          channelFinalizedHandler(arg);
+          resolve();
+        },
+      })
+    );
+
+    await mineBlock(CURRENT_TIME);
+    expect(channelFinalizedHandler).not.toHaveBeenCalled();
```
kerzhner, line 272: I _think_ this check is not sufficient since it is possible that `channelFinalizedHandler` has not fired **yet**. Consider the following sequence of events:
1. current time block is mined.
1. The logic in this line is executed.
1. The `channelFinalizedHandler`fires.
1. future time block is mined.

That being said, I am not sure how to properly check that `channelFinalizedHander` did not fire after the first block is mined. We could insert a delay, but that slows down the testsuite.

kerzhner (approved):
Looks good! It would be great to address https://github.com/statechannels/statechannels/pull/3050/files#r541990529 before merging.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -53,14 +62,21 @@ async function mineBlockPeriodically(blocks: number) {
...
 jest.setTimeout(20_000);
-
-beforeAll(() => {
+// The test nitro adjudicator allows us to set channel storage
+const testAdjudicator = new Contract(
+  nitroAdjudicatorAddress,
+  TestContractArtifacts.TestNitroAdjudicatorArtifact.abi,
+  // We use a separate signer address to avoid nonce issues
+  // eslint-disable-next-line no-process-env
+  new providers.JsonRpcProvider(rpcEndpoint).getSigner(1)
```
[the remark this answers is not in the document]
kerzhner: Just to clarify: a new provider is created on this line. I think we need to call `removeAllListeners` on this provider in the `afterAll` callback.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
lalexgap: Yeah I agree. I couldn't think of a better way of doing it without introducing a delay. I think it's probably worthwhile adding a very short delay so the test is more likely to catch any issues.

lalexgap:
> Looks good! It would be great to address https://github.com/statechannels/statechannels/pull/3050/files#r541990529 before merging.

Good point, done in 75e0714

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
lalexgap: I've added 1 sec delay in d101ee4.

PR #3055, opened 2020-12-10
perf: emit a single event for AssetsTransferred

### LINKED ISSUE #3046: Move to emitting a single `AssetsTransferred` (plural) event
We already emit an `AssetTransferred` event for each internal and external payout: 

https://github.com/statechannels/statechannels/blob/3250047a6104780e43ca8c0cc71ea28b089ff0dc/packages%2Fnitro-protocol%2Fcontracts%2Finterfaces%2FIAssetHolder.sol#L53

375 + ( 2 * 375 ) + ( 3 * 256 ) = 1893 gas per deduction.

[source](https://www.reddit.com/r/ethdev/comments/78ugki/do_indexed_parameters_in_events_cost_more_gas/) 

We could make some decent savings by having an `AssetsTransferred` event, (only one LOG opcode and one `channelId` indexed parameter). We could tackle that in a separate PR.

_Originally posted by @geoknee in https://github.com/statechannels/statechannels/issues/3011#issuecomment-738665671_

geoknee:
Want to see about the gas savings before fixing the tests. 

Closes #3046

### DIFF
--- a/packages/nitro-protocol/contracts/interfaces/IAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/interfaces/IAssetHolder.sol
@@ -47,7 +47,7 @@ interface IAssetHolder {
     /**
      * @dev Indicates that `amounts` assets have been transferred (internally or externally) to the destinationss pecified the CALLDATA. allocation[indices[k]] was transferred amounts[k]
      * @param channelId The channelId of the funds being withdrawn.
-     * @param amounts Number of assets transferred (e.g. wei or tokens) to the destinations specified in the calldata
+     * @param initialHoldings initial holdings of the channel before transfer. Allows the client to compute the payouts from function calldata
      */
-    event AssetsTransferred(bytes32 indexed channelId, uint256[] amounts);
+    event AssetsTransferred(bytes32 indexed channelId, uint256 initialHoldings);
 }
--- a/packages/nitro-protocol/contracts/interfaces/IAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/interfaces/IAssetHolder.sol
@@ -45,10 +45,9 @@ interface IAssetHolder {
     );
 
     /**
-     * @dev Indicates that `amount` assets have been transferred (internally or externally) to the destination denoted by `destination`.
+     * @dev Indicates that `amounts` assets have been transferred (internally or externally) to the destinationss pecified the CALLDATA. allocation[indices[k]] was transferred amounts[k]
      * @param channelId The channelId of the funds being withdrawn.
-     * @param destination An internal destination (channelId) of external destination (padded ethereum address)
-     * @param amount Number of assets transferred (wei or tokens).
+     * @param amounts Number of assets transferred (e.g. wei or tokens) to the destinations specified in the calldata
      */
-    event AssetTransferred(bytes32 indexed channelId, bytes32 indexed destination, uint256 amount);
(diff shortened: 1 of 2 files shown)


packages/nitro-protocol/contracts/interfaces/IAssetHolder.sol
```diff
@@ -45,10 +45,9 @@ interface IAssetHolder {
...
     /**
-     * @dev Indicates that `amount` assets have been transferred (internally or externally) to the destination denoted by `destination`.
+     * @dev Indicates that `amounts` assets have been transferred (internally or externally) to the destinationss pecified the CALLDATA. allocation[indices[k]] was transferred amounts[k]
      * @param channelId The channelId of the funds being withdrawn.
-     * @param destination An internal destination (channelId) of external destination (padded ethereum address)
-     * @param amount Number of assets transferred (wei or tokens).
+     * @param amounts Number of assets transferred (e.g. wei or tokens) to the destinations specified in the calldata
      */
-    event AssetTransferred(bytes32 indexed channelId, bytes32 indexed destination, uint256 amount);
+    event AssetsTransferred(bytes32 indexed channelId, uint256[] amounts);
```
andrewgordstewart, line 52: Do you have an algorithm in mind for re-constructing the destinations in the client?
I suppose the client can get:

    the outcome pre-transfer (calldata)
    the indices transferred (calldata)
    the amounts transferred (calldata)

Now, if this is the data that your algorithm uses, you might as well emit the sum amount transferred here, no? That would save 256 gas per payout.

geoknee: > the outcome pre-transfer (calldata) < YES
the indices transferred (calldata) < YES
the amounts transferred (calldata) < I don't think so?

Yeah the algorithm is:

```
oldAllocation = abi.decode(calldata.allocationBytes);
newAllocation = oldAllocation.clone();
calldata.indices.forEach(i=> newAllocation[i].amount -= event.payouts[i]);
return newAllocation
```

I don't think this is the absolute minimum, though. We could replace `uint256[] payouts` with `uint256 initialHoldings`, and then emulate `_computeNewAllocation` in the client code.

andrewgordstewart: > > the amounts transferred (calldata) < I don't think so?

(mistake, meant to be (this event))

andrewgordstewart: > [andrew accidentally edited this comment]

I don't remember exactly what you wrote, so I'm going to stop editing it now. but we're on the same page: you can emit either:
1. the total amount transferred
2. the initial holdings
and you can re-construct the individual payouts, and only pay the gas to emit one uint256 instead of an array

(2) is what you suggested, and that probably makes sense?

geoknee: Yeah that's it

andrewgordstewart: I took the liberty to implement (2) so we can compare the gas (though I suppose we could predict the gas fee reduction).

geoknee:
see #3063

geoknee:
gas savings seem good (a few hundred for most test cases). 
Note, however, that in some cases we would not emit any events at all, and now we always emit an event. This leads to an increase in gas for those cases:

> 10. -> 2 chan             no:
34859 gas (master) ---> 36345 gas  (7d6e979)

PR #3062, opened 2020-12-11
Require Challenge Duration when creating a channel

### LINKED ISSUE #3060: The wallet should accept a challengeDuration
The wallet should accept a challenge duration when creating new channels instead of using a hardcoded value of 9001

lalexgap:
Fixes #3060.

This should not be merged until https://github.com/statechannels/the-graph/issues/489 is ready to go.

### DIFF
--- a/packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
+++ b/packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
@@ -9,6 +9,7 @@ import {
   getChannelResultFor,
   getPayloadFor,
   interParticipantChannelResultsAreEqual,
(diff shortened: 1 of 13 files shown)


packages/server-wallet/src/__test__/create-and-close-channel/ledger-funding.test.ts
```diff
@@ -1,5 +1,12 @@
 import {CreateChannelParams} from '@statechannels/client-api-schema';
-import {BN, makeAddress, makeDestination, Participant} from '@statechannels/wallet-core';
+import {
```
andrewgordstewart, line 2: this is not a `fix` :P

andrewgordstewart (approved):

PR #3063, opened 2020-12-11
replace n * AssetTransferred events with a single AllocationUpdated event

geoknee:
This emits a close-to-minimal amount of information when `transfer` or `claim` is called: the information is sufficient for clients to construct a new `AssetOutcome` for a given channel in a given `AssetHolder`. They need this up-to-date data in order to successfully call `transfer` in future.

Example: my counterparty concluded, pushed, and transferred their own money out of a channel. Now I want to transfer my money out, but my view of the outcome on chain is out of date and my transaction will fail if I don't keep it fresh. 

- [x] contracts updated with new event type
- [x]  off-chain version of `_computeNewAllocation`. We could go down the pureEVM route, but for now I just reimplemented it in typescript and
- [x] wrote a test that uses random data to check against the solidity version. 
- [x] existing tests updated to catch and assert on new event 
- [x] AssetTransferred deprecated in nitro-protocol and the documentation
- [x] Chain Service updated to watch for the new event and call into `computeNewAllocation` to infer the data it wants to understand.

### DIFF
--- a/packages/docs-website/docs/docs/app-devs/make-api-calls.md
+++ b/packages/docs-website/docs/docs/app-devs/make-api-calls.md
@@ -100,7 +100,7 @@ WalletA-->>ClientA: ChannelUpdated('closed')
 WalletA->>Chain: concludePushOutcomeAndTransferAll()
 deactivate WalletA
 deactivate WalletB
-Chain-->>WalletA: AssetTransferred x2
-Chain-->>WalletB: AssetTransferred x2
+Chain-->>WalletA: AllocationUpdated
+Chain-->>WalletB: AllocationUpdated
 end
 " />
--- a/packages/docs-website/docs/docs/protocol-tutorial/release-assets.md
+++ b/packages/docs-website/docs/docs/protocol-tutorial/release-assets.md
@@ -88,18 +88,13 @@ const assetOutcome: AllocationAssetOutcome = {
 const tx3 = ETHAssetHolder.transferAll(channelId, encodeAllocation(assetOutcome.allocationItems));
 
 /* 
-  Check that an AssetTransferred event was emitted.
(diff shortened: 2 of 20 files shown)


geoknee:
In the chain service we should be able to do something like this 

```ts
  AssetHolder.on('AllocationUpdated', async function(channelId, initialHoldings, eventObject) {

    const calldata = abi.decode((await provider.getTransaction(eventObject.transactionHash)).data);

    const newAllocation = computeNewAllocation(
      initialHoldings,
      calldata.allocation,
      calldata.indices
    ); // store this for any future `transfer` transactions we wish to submit

    const payouts = newAllocation.payouts; // trigger downstream events such as "channel defunded"
  });
```

andrewgordstewart:
> In the chain service we should be able to do something like this
> 
> ```ts
>   AssetHolder.on('AllocationUpdated', async function(channelId, initialHoldings, eventObject) {
>  // the callback seems to be passed all event args in order, followed by an object with a bunch of useful data such as the tx hash
> 
>     const calldata = abi.decode((await provider.getTransaction(eventObject.transactionHash)).data);
> 
>     const newAllocation = computeNewAllocation(
>       initialHoldings,
>       calldata.allocation,
>       calldata.indices
>     ); // store this for any future `transfer` transactions we wish to submit
> 
>     const payouts = newAllocation.payouts; // trigger downstream events such as "channel defunded"
>   });
> ```

I think we'd want to provide the wallet with a relatively rich, computed event. Something like:
```
{
  type: 'AssetOutcomeUpdated';
  channelId: string;
  payouts: ...,
  newOutcome: ...,
  oldOutcome: ... // ?
}
```

geoknee:
In favour of #3074

PR #3070, opened 2020-12-14
Add finalisation status to channel

lalexgap:
This adds the finalization status to the `ChannelState` that gets passed into protocols.

This sets us up to easily check the finalization status in the close channel protocol for [this](https://github.com/statechannels/the-graph/issues/490)


packages/server-wallet/src/db/migrations/20201211180815_finalization_status.ts
```diff
@@ -0,0 +1,17 @@
+import * as Knex from 'knex';
+
+const tableName = 'finalization_status';
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable(tableName, function(table) {
+    table.string('channel_id').notNullable();
+    table.integer('finalizes_at').notNullable();
+    table.integer('block_number').notNullable();
```
andrewgordstewart, line 8: These properties seem to be common among all channels. What's the argument for storing them in a separate table?

Skimming the rest of this PR, the logic would be simpler if these data were stored on the `channels` table.

Two reasons I can think of for creating a new table might be
1. It might be "neater" to store all blockchain-related data for a channel in a single table
2. It is possible to update the record corresponding to a channel `C` in the `finalization_status` table, while some other operation is updating the row corresponding to `C` in `channels`.

I'm not sure if either of these are a strong argument though?

lalexgap: I think 1. was the main impetus behind this. It seemed like a good idea to keep the blockchain data separate similar to the `funding` table.

andrewgordstewart: should the table be called something as specific as `finalization_status` then?

Considering that it's easy to rename a table, I don't see a real issue with it just the same.

lalexgap: I think this will probably get broadened to include all challenging info eventually, so it might make sense to rename it now.

packages/server-wallet/src/models/finalization-status.ts
```diff
@@ -0,0 +1,70 @@
...
+      .where({channelId})
+      .first();
+
+    return FinalizationStatus.convertResult(result);
+  }
+
+  static async updateFinalizationStatus(
+    knex: Knex,
+    channelId: string,
+    finalizesAt = 0,
```
andrewgordstewart, line 35: Should these have defaults?

I feel like setting `finalizesAt` to `0` should be a deliberate action, not the default action.

packages/server-wallet/src/db/migrations/20201211180815_finalization_status.ts
lalexgap: I've renamed this to `challenging_status` for now.

lalexgap:
I'm going to close this and just include these changes with the changes for #3078

lalexgap:
> I'm going to close this and just include these changes with the changes for #3078

Actually, on second thought, this should probably be its own chunk of work.

PR #3072, opened 2020-12-15
redeploy to rinkeby and verify on etherscan

geoknee:
A bit of a dry run for imminent deployment to mainnet.

Fixed some things as I went along: 

- I don't have `jq` installed on my machine, so I converted our `postdeploy` script ~to use `node-jq`~ to pure JS to prevent the script failing (it just nuked the `addresses.json` file altogther)
- tweaked one of our test contracts so that it has distinct bytecode
- added `@nomiclabs/hardhat-etherscan` plugin


geoknee:
Prefer to merge https://github.com/statechannels/statechannels/pull/3073 hence do not merge label.

packages/nitro-protocol/contracts/test/TestEthAssetHolder.sol
```diff
@@ -7,5 +7,7 @@ import '../ETHAssetHolder.sol';
  * @dev This contract is a clone of the ETHAssetHolder contract. It will be initialized to point to the TestNitroAdjudicator.
  */
 contract TestEthAssetHolder is ETHAssetHolder {
+    uint256 public dummy; // this is simply to make the contract have distinct bytecode from the ETHAssetHolder (otherwise bytecode verification can fail)
```
andrewgordstewart, line 10: nit: We could _really_ make sure the bytecode is unique?

```suggestion
    string public contractName = 'TestEthAssetHolder'; // ensures the contract has different bytecode from the ETHAssetHolder (otherwise bytecode verification can fail)
```

packages/nitro-protocol/scripts/postdeploy.js
```diff
@@ -0,0 +1,14 @@
...
+const addresses = require(jsonPath);
+
+function deepDelete(object, keyToDelete) {
+  Object.keys(object).forEach(key => {
+    if (key === keyToDelete) delete object[key];
+    else if (typeof object[key] === 'object') deepDelete(object[key], keyToDelete);
+  });
+}
+const keyToDelete = 'abi';
+deepDelete(addresses, keyToDelete);
```
andrewgordstewart, line 13: Conceptually, it makes more sense to me to _keep the keys you want_, rather than _delete they keys you don't want_.

Is there a reason for the latter here?

geoknee: Reason 1: this is just transcribing existing code, so I considered it safer to adopt the existing approach. 
Reason 2: it's a deeply nested object, so there are fare more keys to keep than to delete in this case. Although I afree with your instinct in general, I think in this case it would be more complicated to implement it that way. It's marginal.

PR #3080, opened 2020-12-16
feat: claim method accepts indices array

geoknee:
- exposes a pure public function
`_computeNewAllocationWithGuarantee` which can be emulated off chain

note two changes to behaviour
- amout: 0 allocation items are left in place
- overfunded guarantees have the excess funds locked up, instead of paid out as is currently done

overall this brings the API in line with the transfer method.

TODO
- [x] write the TS mirror of `_computeNewAllocationWithGuarantee`
- [x] write test for ^
- [x] have `computeNewAssetOutcome` call into this function where appropriate


geoknee:
There is an argument that says 
> For the foreseeable future we only expect guarantees to target channels with 2 outcomes. So we could forego a general `claim(indices)` method and only support `claimAll`. 

Although that would simplify the on and off chain code, it would suffer from the problem that one party has to pay out all of the allocation items, which might involve spending a lot of gas in someone else's contract.

packages/nitro-protocol/test/contracts/AssetHolder/computeNewAllocation.test.ts
```diff
@@ -56,7 +48,7 @@ describe('AsserHolder._computeNewAllocation', () => {
       indices
     )) as ReturnType<typeof computeNewAllocation>;
     expect(result).toBeDefined();
-
+    console.log(result);
```
tomclose, line 51: ✂️

tomclose (approved):
lgtm

Issue #3081, opened 2020-12-16
reduce costs of deployments to mainnet


geoknee:
Ideas: 

- Only deploy `NitroAdjudicator`
- reduce gasPrice
- choose deployment time wisely
- reduce size of contracts --.e.g. shorter reason strings

PR #3082, opened 2020-12-16
build: estimate gas cost of deployment

geoknee:
- store in folder uploaded as circle artifact
- deploy script converted to typescript

preparatory work for #3081


packages/nitro-protocol/deployment/deploy.ts
```diff
@@ -0,0 +1,80 @@
...
+const deploy = async () => {
+  const deployer = new GanacheDeployer(Number(process.env.GANACHE_PORT));
+
+  const nitroAdjudicatorDeploymentGas = await deployer.etherlimeDeployer.estimateGas(
+    nitroAdjudicatorArtifact as any
+  );
+  writeGasConsumption('NitroAdjudicator.gas.md', 'deployment', nitroAdjudicatorDeploymentGas);
+  console.log(
+    `\nDeploying NitroAdjudicator... (cost estimated to be ${nitroAdjudicatorDeploymentGas})\n`
+  );
```
geoknee, line 26: The diff isn't easy to read because of the extension change. These are the new lines.

andrewgordstewart: A good practice here is to separate the two changes into separate commits: one changes the extension but no logic, the second changes the logic.

geoknee: As you wish #3083

PR #3084, opened 2020-12-16
perf: use shorter reason strings

### LINKED ISSUE #3081: reduce costs of deployments to mainnet

geoknee:
The test will fail for now but easily fixed.

Reduces estimated gas costs of deploying NitroAdjudicator: `2566237` becomes `2420522` => `145715` drop.

Closes #3081

### DIFF
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -569,7 +569,7 @@ contract AssetHolder is IAssetHolder {
                         )
                     )
                 ),
-            'h(allocation) ≠ assetOutcomeHash'
+            'h(allocation)!=assetOutcomeHash'
         );
     }
 
@@ -587,7 +587,7 @@ contract AssetHolder is IAssetHolder {
                         )
                     )
                 ),
-            'h(guarantee) ≠ assetOutcomeHash'
+            'h(guarantee)!=assetOutcomeHash'
         );
     }
 
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -273,7 +273,7 @@ contract AssetHolder is IAssetHolder {
 
         for (uint256 j = 0; j < guarantee.destinations.length; j++) {
             if (balance == 0) {
-                revert('guarantor affords 0 for dest');
+                revert('guarantor affords zero for dest');
             }
             // for each destination in the guarantee,
             // find the first corresponding allocationItem in the target allocation
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -273,7 +273,7 @@ contract AssetHolder is IAssetHolder {
 
         for (uint256 j = 0; j < guarantee.destinations.length; j++) {
             if (balance == 0) {
-                revert('_claim : guarantorChannel affords 0 for destination');
+                revert('guarantor affords 0 for dest');
             }
             // for each destination in the guarantee,
             // find the first corresponding allocationItem in the target allocation
@@ -308,7 +308,7 @@ contract AssetHolder is IAssetHolder {
             }
         }
 
-        require(affordsForDestination > 0, '_claim | guarantor affords 0 for destination');
+        require(affordsForDestination > 0, 'guarantor affords 0 for dest');
 
         // effects
         holdings[guarantorChannelId] -= affordsForDestination;
@@ -569,7 +569,7 @@ contract AssetHolder is IAssetHolder {
                         )
                     )
                 ),
-            'AssetHolder | submitted allocationBytes data does not match stored assetOutcomeHash'
+            'h(allocation) ≠ assetOutcomeHash'
         );
     }
 
@@ -587,7 +587,7 @@ contract AssetHolder is IAssetHolder {
                         )
                     )
                 ),
-            'AssetHolder | submitted guaranteeBytes data does not match stored assetOutcomeHash'
+            'h(guarantee) ≠ assetOutcomeHash'
         );
     }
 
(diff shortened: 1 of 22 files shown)


andrewgordstewart:
I think we should have a maintained list of error codes. Developer experience can be maintained by storing meaningful error reasons in typescript.

geoknee:
If it is only the developer experience that benefits from the reason strings, we could run a [preprocessor](https://hardhat.org/plugins/hardhat-preprocessor.html) to strip them out of production deploys altogether?

geoknee:
From sync with @andrewgordstewart today:

- reason strings may (or may not) be important for wallets to make decisions, too. example: I try to conclude not knowing someone else already did. If I get a revert reason 'channel already concluded' this could help me out. (There are other ways of finding this out, but they may be more complex)
- we can use the code -> verbose reason mapping suggested above to inject verbose information into our jest tests.

geoknee:
Also we can adopt a phased approach to this

0. fixup and merge this PR for immediate gas savings with little to no loss in information
1. change to error codes because it enables more verbosity and more robust programmatic interpretation of reverts
2. strip reason strings out of production deploys if we deem that a good idea (save even more gas but lose chance to programmatically interpret reverts in client code). And hope this doesn't make bytecode verification more difficult.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -308,7 +308,7 @@ contract AssetHolder is IAssetHolder {
             }
         }
 
-        require(affordsForDestination > 0, '_claim | guarantor affords 0 for destination');
+        require(affordsForDestination > 0, 'guarantor affords 0 for dest');
```
andrewgordstewart, line 311: This is now the same string as above, worsening the debug experience.

packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -569,24 +563,12 @@ contract ForceMove is IForceMove {
...
             if (turnNumB < 2 * nParticipants) {
-                require(
-                    _bytesEqual(ab[1].outcome, ab[0].outcome),
-                    'InvalidTransitionError: Cannot change the default outcome during setup phase'
-                );
-                require(
-                    _bytesEqual(ab[1].appData, ab[0].appData),
-                    'InvalidTransitionError: Cannot change the appData during setup phase'
-                );
+                require(_bytesEqual(ab[1].outcome, ab[0].outcome), 'Outcome change forbidden');
```
andrewgordstewart, line 570: Duplicate reason string

packages/nitro-protocol/contracts/examples/SingleAssetPayments.sol
```diff
@@ -42,11 +42,11 @@ contract SingleAssetPayments is IForceMoveApp {
         );
         require(
             assetOutcomeA.assetOutcomeType == uint8(Outcome.AssetOutcomeType.Allocation),
-            'SingleAssetPayments: outcomeA: AssetOutcomeType must be Allocation'
+            'AssetOutcome must be Allocation'
         );
         require(
             assetOutcomeB.assetOutcomeType == uint8(Outcome.AssetOutcomeType.Allocation),
-            'SingleAssetPayments: outcomeB: AssetOutcomeType must be Allocation'
+            'AssetOutcome must be Allocation'
```
andrewgordstewart, line 49: Duplicate reason string
```suggestion
            'A.OutcomeB must be Allocation'
```

packages/nitro-protocol/contracts/examples/SingleAssetPayments.sol
```diff
@@ -58,14 +58,8 @@ contract SingleAssetPayments is IForceMoveApp {
...
-        require(
-            allocationA.length == nParticipants,
-            'SingleAssetPayments: outcomeA: Allocation length must equal number of participants'
-        );
-        require(
-            allocationB.length == nParticipants,
-            'SingleAssetPayments: outcomeB: Allocation length must equal number of participants'
-        );
+        require(allocationA.length == nParticipants, '|Allocation|!=|participants|');
+        require(allocationB.length == nParticipants, '|Allocation|!=|participants|');
```
andrewgordstewart, line 62:
```suggestion
        require(allocationB.length == nParticipants, '|AllocationB|!=|participants|');
```

packages/nitro-protocol/src/valid-transition.ts
```diff
@@ -20,24 +20,12 @@ export function requireValidProtocolTransition(fromState: State, toState: State)
 
   // directly ported checks:
   if (toState.isFinal) {
-    _require(
-      isEqual(toState.outcome, fromState.outcome),
-      'InvalidTransitionError: Cannot move to a final state with a different default outcome'
-    );
+    _require(isEqual(toState.outcome, fromState.outcome), 'Outcome change forbidden');
```
andrewgordstewart, line 23: Duplicate reason?

packages/nitro-protocol/src/valid-transition.ts
```diff
@@ -20,24 +20,12 @@ export function requireValidProtocolTransition(fromState: State, toState: State)
...
     if (toState.turnNum < 2 * toState.channel.participants.length) {
-      _require(
-        isEqual(toState.outcome, fromState.outcome),
-        'InvalidTransitionError: Cannot change the default outcome during setup phase'
-      );
-      _require(
-        toState.appData == fromState.appData,
-        'InvalidTransitionError: Cannot change the appData during setup phase'
-      );
+      _require(isEqual(toState.outcome, fromState.outcome), 'Outcome change forbidden');
```
andrewgordstewart, line 27: Duplicate reason

andrewgordstewart (approved):
There appear to be some duplicate reason strings, which worsens the debug experience.

I don't think this is a reason to block this PR, but I'd slightly prefer that they be unique.

PR #3087, opened 2020-12-17
refactor: swap JS based hashState to use wasm utility

### LINKED ISSUE #3066: Server Wallet still uses JS based state hashing
https://github.com/statechannels/statechannels/blob/a5a65d8c3225109e4ca0807170992a370ca6c759/packages/server-wallet/src/models/channel.ts#L216-L216

snario:
Fixes #3066

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -6,14 +6,15 @@ import {
   SignedStateVarsWithHash,
   SignedStateWithHash,
   calculateChannelId,
-  hashState,
   outcomesEqual,
   Zero,
   Address,
+  toNitroState,
 } from '@statechannels/wallet-core';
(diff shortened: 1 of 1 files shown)


geoknee (approved):
Is there something we can do to prevent missing these in future? A custom lint rule? Or justs top exporting `hashState` from wallet core?

snario:
I think the exported `hashState` should by default use the wasm utility if it's available in the environment (i.e., some kind of `__wasm_available__` boolean perhaps)

andrewgordstewart:
>     * The built wasm utility is not available on the machine for some reason (no rust?)

wasm utilities are already compiled -- there should be no requirement to install `rust` to use them

>     * We're in the browser, in which case it becomes an asynchronous API (or unavailable on older browsers)

why would it become asynchronous?

snario:
Loading WebAssembly modules in the browser requires using dynamic importing: https://developer.mozilla.org/en-US/docs/WebAssembly/Loading_and_running

PR #3088, opened 2020-12-17
refactor: store.getChannel -> store.getChannelState

tomclose:
* We have an object called a `Channel` and an object called a `ChannelState`.
* Currently `store.getChannel` returns a `ChannelState`. 
* This PR renames it to `getChannelState`
* In another PR, I'm going to recreate `getChannel` so that it actually returns a `Channel`

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -252,13 +252,24 @@ export class Store {
     // This is somewhat faster than Channel.forId for simply fetching a channel:
     // - the signingWallet isn't needed to construct the protocol state
     // - withGraphJoined is slightly faster in this case
-    return (
-      await Channel.query(tx ?? this.knex)
-        .where({'channels.channel_id': channelId})
-        .withGraphJoined('funding')
-        .withGraphJoined('chainServiceRequests')
-        .first()
-    )?.protocolState;
+    const channel = await this.getChannel(channelId, tx);
+
+    // TODO: throwing here is a quick fix until we can update all consumers of getChannelState
+    // to handle ChannelState | undefined
+    if (!channel) throw new Error(`Channel not found with id ${channelId}`);
+
+    return channel.protocolState;
+  }
+
+  async getChannel(channelId: Bytes32, tx?: Transaction): Promise<Channel | undefined> {
+    // This is somewhat faster than Channel.forId for simply fetching a channel:
+    // - the signingWallet isn't needed to construct the protocol state
+    // - withGraphJoined is slightly faster in this case
+    return Channel.query(tx ?? this.knex)
+      .where({'channels.channel_id': channelId})
+      .withGraphJoined('funding')
+      .withGraphJoined('chainServiceRequests')
+      .first();
   }
 
   async getStates(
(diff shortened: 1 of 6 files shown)


andrewgordstewart (commented):
Why should the store return a channel model? The channel model should be an implementation detail of the Postgres store

tomclose:
We currently have `ChannelStates`, `ProtocolStates` and `Channels` in the wallet codebase, which are all very similar but not interchangeable, and which make it generally painful to work with the code. I think we prematurely abstracted the `ChannelState` concept, before we knew how the protocols would be written, or what they would require.

As I'm working on the protocols, I'm trying to consolidate around one object, and have chosen the `Channel` for this purpose - it's already a class and therefore easy to add convenience methods, and we know it will definitely exist going forwards. Once the protocols are written, it will be easy to refactor to have an abstract `Channel` interface (decoupled from postgres), and a `ChannelModel` implementation.

andrewgordstewart:
The protocol state is not meant to be interchangeable with a channel state. For instance, the ledger funding protocol needs a lot more information, and that is reflected on its protocol state.

The virtual funding protocol would need to know about multiple channels simultaneously, and that would be reflected in its protocol state.

The channel state is meant to be the abstract channel interface you mention. So your proposal, on the surface, seems to be a backward step?

tomclose:
Part of the motivation for what I'm doing is to move away from the protocol state. Instead of fetching everything all the information we could possibly need to run a protocol upfront, I'm allowing protocols to request more information as they go along. This means, for example, that we don't need to query the LedgerRequests table at all, if we can see that the channel is already funded.

I agree that the `ChannelState` is _meant_ to be the abstract interface, but I believe it was abstracted too early, before we knew how the protocols would be written, or what they would require. So in a sense this is a backwards step - I'm choosing to use a concrete interface (the `Channel`) for the time being, so that we can more easily figure out what the abstract interface should be.

As a concrete example, by using the `Channel`, it's quick and easy to do [add methods](https://github.com/statechannels/statechannels/pull/3089/commits/b3aa196898362df2a7cda5cdeba9f621977a58a5) like `postfundSigned`, which allows protocols to be written [more naturally](https://github.com/statechannels/statechannels/blob/de1fdbe6e2db95d6465dc64ae2007985482d2d1d/packages/server-wallet/src/protocols/open-channel.ts#L36-L65) and is a natural place for the consolidation of helper methods:
```ts
// if we haven't signed the pre-fund, sign it
if (!channel.prefundSigned) {
  const signedState = await this.store.signState(channel, channel.latest, tx);
  response.queueState(signedState, channel.myIndex, channel.channelId);
}

// if we don't have a supported preFundSetup, we're done for now
if (!channel.prefundSupported) {
  return;
}
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -251,13 +251,24 @@ export class Store {
     // This is somewhat faster than Channel.forId for simply fetching a channel:
     // - the signingWallet isn't needed to construct the protocol state
     // - withGraphJoined is slightly faster in this case
```
andrewgordstewart, line 253: This comment is copied below, and thus seems to have become misplaced.

andrewgordstewart (approved):

PR #3095, opened 2020-12-18
refactor: change Wallet from object to abstract class

### LINKED ISSUE #3094: Usage of wallet's inherited EventEmitter .on does not compile
This error is shown

```
src/channel-manager.ts:436:19 - error TS2349: This expression is not callable.
  Each member of the union type '(<T extends "channelUpdated">(event: T, fn: (...args: ArgumentMap<EventEmitterType>[Extract<T, "channelUpdated">]) => void, context?: any) => SingleThreadedWallet) | (<T extends "channelUpdated">(event: T, fn: (...args: ArgumentMap<...>[Extract<...>]) => void, context?: any) => MultiThreadedWallet)' has signatures, but none of those signatures are compatible with each other.
436       this.wallet.on('channelUpdated', (channelResult) => {
                      ~~

src/channel-manager.ts:436:41 - error TS7006: Parameter 'channelResult' implicitly has an 'any' type.
436       this.wallet.on('channelUpdated', (channelResult) => {
                                            ~~~~~~~~~~~~~

Found 2 errors.
```

snario:
Fixes #3094

### DIFF
--- a/packages/server-wallet/server-wallet.api.md
+++ b/packages/server-wallet/server-wallet.api.md
@@ -231,20 +231,14 @@ export function validateServerWalletConfig(config: Record<string, any>): {
 };
 
 // Warning: (ae-forgotten-export) The symbol "SingleThreadedWallet" needs to be exported by the entry point index.d.ts
-// Warning: (ae-forgotten-export) The symbol "MultiThreadedWallet" needs to be exported by the entry point index.d.ts
 //
 // @public (undocumented)
-export type Wallet = SingleThreadedWallet | MultiThreadedWallet;
-
-// @public (undocumented)
-export const Wallet: {
-    create(walletConfig: IncomingServerWalletConfig): Wallet;
-};
-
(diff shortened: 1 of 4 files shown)


andrewgordstewart (approved):
It would be nice to give an explanation of why this fixes #3094 .

snario:
Previously `Wallet` was not a class, it was an object with a key `create` that was a function that returned wither `SingleThreadedWallet` or a `MultiThreadedWallet`. Given this type structure and the way `EventEmitter` wrote `.on( ... )`'s type definition, it led to the `.on( ... )` having two possible function definitions, one which returned `SingleThreadedWallet` and another returning a `MultiThreadedWallet`:

```ts
EventEmitter<EventEmitterType, any>.on<"channelUpdated">( ... ): SingleThreadedWallet
EventEmitter<EventEmitterType, any>.on<"channelUpdated">( ... ): MultiThreadedWallet
```

This is not valid according to TypeScript. The solution was to create an abstract `Wallet` class so change the signature to:


```ts
EventEmitter<EventEmitterType, any>.on<"channelUpdated">( ... ): Wallet
```

Which TypeScript likes better.

Issue #3097, opened 2021-01-04
fix broken links on docs website


geoknee:
From homepage https://docs.statechannels.org. 

- Broken:
https://docs.statechannels.org/app-devs/intro.md
https://docs.statechannels.org/protocol-docs/quick-start.md

- And address this huge page

```
 [BABEL] Note: The code generator has deoptimised the styling of /Users/georgeknee/statechannels/statechannels/packages/docs-website/docs/typescript-api/nitro-protocol.contractartifacts.md as it exceeds the max of 500KB.
```

PR #3099, opened 2021-01-04
[nitro-protocol] export lighter artifacts

geoknee:
1. Pick minimal fields from contract artifacts that are exported (they recently ballooned in size, and are even printed in full on the docs website! #3097)  

2. Move `TokenArtifact` underneath `TestArtifacts` object. ⚠️ this is a breaking change. It's not totally necessary, so this PR could be regenerated without that change.


packages/nitro-protocol/src/index.ts
```diff
@@ -1,29 +1,46 @@
...
+  'contractName',
+  'abi',
+  'bytecode',
+  'deployedByteCode',
+  'linkReferences',
+  'deployedLinkReferences',
+];
+
+interface MinimalArtifact {
+  contractName: typeof FULLNitroAdjudicatorArtifact.contractName;
```
andrewgordstewart, line 21: What is `typeof FULLNitroAdjudicatorArtifact`? I don't understand how it could be a meaningful type, if `FULLNitroAdjudicatorArtifact` is an imported `json` file.

packages/nitro-protocol/src/index.ts
```diff
@@ -1,29 +1,46 @@
...
 export const ContractArtifacts = {
-  NitroAdjudicatorArtifact,
-  TrivialAppArtifact, // TODO do we want to export this?
-  Erc20AssetHolderArtifact,
-  EthAssetHolderArtifact,
-  TokenArtifact, // TODO do we want to export this?
-  AssetHolderArtifact, // TODO do we want to export this?
+  NitroAdjudicatorArtifact: pick(FULLNitroAdjudicatorArtifact, fields) as MinimalArtifact,
+  Erc20AssetHolderArtifact: pick(FULLErc20AssetHolderArtifact, fields) as MinimalArtifact,
+  EthAssetHolderArtifact: pick(FULLEthAssetHolderArtifact, fields) as MinimalArtifact,
```
andrewgordstewart, line 32: nit: With a helper
```
const minimize = artifact => pick(artifact, fields) as MinimalArtifact
```

this could be a bit neater

```suggestion
  NitroAdjudicatorArtifact: minimize(FULLNitroAdjudicatorArtifact),
  Erc20AssetHolderArtifact: minimize(FULLErc20AssetHolderArtifact),
  EthAssetHolderArtifact: minimize(FULLEthAssetHolderArtifact),
```

andrewgordstewart (approved):

packages/nitro-protocol/src/index.ts
geoknee: There is some inference done by the typescript compiler, such that importing from a Json file results in a rich and typed object. I was surprised by this, too.

geoknee: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#new---resolvejsonmodule

packages/nitro-protocol/src/index.ts
```diff
@@ -94,6 +111,8 @@ export {
 
 import * as Signatures from './signatures';
 import * as Transactions from './transactions';
+import {computePublicKey} from 'ethers/lib/utils';
+import {Contract, ContractInterface} from 'ethers';
```
geoknee, line 115: These don't belong.

PR #3104, opened 2021-01-05
refactor: make fundingStatus optional in the ChannelResult

tomclose:
Currently the fundingStatus value is sometimes inaccurate. For example, it's possible to get 'Uncategorized' or 'ReadyToFund' when the channel is funded, but the funding hasn't been fetched from the db.

Fixing this would require always querying the db for funding, which might not be desirable for performance reasons. Instead, consumers will now receive `fundingStatus: undefined` if it hasn't be fetched.

### DIFF
--- a/packages/client-api-schema/src/data-types.ts
+++ b/packages/client-api-schema/src/data-types.ts
@@ -141,7 +141,7 @@ export interface ChannelResult {
   status: ChannelStatus;
   turnNum: Uint48;
   challengeExpirationTime?: number;
-  fundingStatus: FundingStatus;
+  fundingStatus?: FundingStatus;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


tomclose:
It doesn't look like this will cause any issues with the graph, as the only way that code uses the `fundingStatus` is to check that it `!== 'Defunded'` - a property that won't change: https://github.com/statechannels/the-graph/search?q=fundingStatus

snario:
This may cause an issue with this PR https://github.com/statechannels/the-graph/pull/480/files#diff-d5feac8493a10695a08de9574f4f0a695585f457b849e4ebe0cc05ffe3ea21b8R514

tomclose:
I think it should be fine on the line you highlighted, as if `fundingStatus == 'Funded'` currently, that won't change.

lalexgap (approved):
I don't love this behaviour for long term but for the immediate future it seems ok? Perhaps we should add a TODO and GH issue to implement better behaviour.

PR #3109, opened 2021-01-06
refactor: remove succeeded and failed objective statuses; do garbage collection

snario:
Presently, objectives can pile up and there is no mechanism for removing them. This PR removes them after they have succeeded.

### DIFF
--- a/packages/server-wallet/src/protocols/__test__/challenge-submitter.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/challenge-submitter.test.ts
@@ -146,7 +146,9 @@ const crankAndAssert = async (
   }
 
   if (completesObj) {
-    await expect(store.getObjective(objective.objectiveId)).rejects.toThrow();
+    await expect(store.getObjective(objective.objectiveId)).rejects.toThrow(
+      'Objective does not exist in database'
+    );
   } else {
     const reloadedObjective = await store.getObjective(objective.objectiveId);
     expect(reloadedObjective.status).toEqual('pending');
--- a/packages/server-wallet/src/protocols/__test__/challenge-submitter.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/challenge-submitter.test.ts
@@ -145,11 +145,10 @@ const crankAndAssert = async (
     expect(spy).not.toHaveBeenCalled();
   }
 
-  const reloadedObjective = await store.getObjective(objective.objectiveId);
-
   if (completesObj) {
-    expect(reloadedObjective.status).toEqual('succeeded');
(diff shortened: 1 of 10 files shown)


packages/server-wallet/src/protocols/__test__/challenge-submitter.test.ts
```diff
@@ -147,11 +147,10 @@ const crankAndAssert = async (
     expect(spy).not.toHaveBeenCalled();
   }
 
-  const reloadedObjective = await store.getObjective(objective.objectiveId);
-
   if (completesObj) {
-    expect(reloadedObjective.status).toEqual('succeeded');
+    await expect(store.getObjective(objective.objectiveId)).rejects.toThrow();
```
andrewgordstewart, line 151: Should it expect a specific error message?

packages/server-wallet/src/protocols/__test__/close-channel.test.ts
```diff
@@ -141,10 +141,10 @@ const crankAndAssert = async (
     expect(spy).not.toHaveBeenCalled();
   }
 
-  const reloadedObjective = await store.getObjective(objective.objectiveId);
   if (completesObj) {
-    expect(reloadedObjective.status).toEqual('succeeded');
+    await expect(store.getObjective(objective.objectiveId)).rejects.toThrow();
```
andrewgordstewart, line 145: Ditto

andrewgordstewart (approved):

snario:
This _may_ have surfaced an issue that the recent refactoring work on how protocols are written was designed to fix. In particular, we use `FOR UPDATE` on a channel row, but not on objectives or other rows, and so perhaps there is a race condition where concurrent API calls may both attempt to operate on the same objective at the same time.

snario:
The above comment inspired #3131

andrewgordstewart:
Would you consider deferring this, and alternatively, remove objectives at the point when we remove the channel?

Leaving the objective in the DB aids in debugging. Removing it also removes the possibility of querying the status of an objective that has failed, making robustness on startup a more difficult problem.

snario:
That seems reasonable, but then also it might be worth thinking about when we remove a channel as we don't do that at all right now.

andrewgordstewart:
Yes. I would rather spend our effort thinking about that, and leave objectives with a `failed` or `succeeded` states for both of the reasons listed above.

snario:
Closing as we'd like to keep objectives in the DB until a channel is deleted.

snario:
Closing as we'd like to keep objectives in the DB until a channel is deleted.

PR #3111, opened 2021-01-06
Add challenge status to channel

lalexgap:
Stores the challenge status for a channel in the database. This allows for protocols to get information about the challenge status.

(This is the same PR as #3070 but since I force pushed I couldn't seem to reopen it)


packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -344,27 +344,30 @@ export class ChainService implements ChainServiceInterface {
...
-      // So we probably want to emit this event as soon as possible.
-      this.channelToSubscribers.get(channelId)?.map(subscriber =>
-        subscriber.channelFinalized({
-          channelId,
-        })
-      );
-      // Chain storage has a new finalizesAt timestamp
-    } else if (finalizesAt) {
-      this.addFinalizingChannel({channelId, finalizesAtS: finalizesAt});
+    if (finalizingChannel.finalizesAtS <= block.timestamp) {
```
geoknee, line 351: I'm having trouble understanding the reason for this change. This function is not terribly easy to understand in the first place ☹️ .

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -794,8 +794,12 @@ export class SingleThreadedWallet extends EventEmitter<EventEmitterType>
     response.channelUpdatedEvents().forEach(event => this.emit('channelUpdated', event.value));
   }
 
-  async channelFinalized(_arg: ChannelFinalizedArg): Promise<void> {
-    return;
+  async channelFinalized(arg: ChannelFinalizedArg): Promise<void> {
```
geoknee, line 797: Out of scope for this PR, but I still find it strange that the wallet implements a ` ChainEventSubscriberInterface`. That approach causes functions such as this one to be exposed in the external API of the wallet. In my opinion this can cause confusion about which other bits of code should call such functions.

packages/server-wallet/src/protocols/state.ts
```diff
@@ -80,6 +82,7 @@ export const toChannelResult = (channelState: ChannelState): ChannelResult => {
     latestSignedByMe,
     support,
     directFundingStatus,
+    challengeStatus: challengeStatus,
```
geoknee, line 85:
```suggestion
    challengeStatus,
```

packages/server-wallet/src/models/challenge-status.ts
```diff
@@ -0,0 +1,73 @@
+import Knex from 'knex';
+import {Model} from 'objection';
+
+import {Bytes32, Uint48} from '../type-aliases';
+
+export type ChallengeStatusResult =
+  | {status: 'Challenge Finalized'; finalizedAt: number; finalizedBlockNumber: number}
+  | {status: 'Challenge Active'}
```
geoknee, line 8: Might we want to include a `finalizesAt` field in the `Challenge Active` type?

packages/server-wallet/src/db/migrations/20201211180815_challenge_status.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const tableName = 'challenge_status';
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable(tableName, function(table) {
+    table.string('channel_id').notNullable();
+    table.integer('finalizes_at').notNullable();
+    table.integer('block_number').notNullable();
+
+    table.primary(['channel_id']);
+    table.foreign('channel_id').references('channels.channel_id');
```
geoknee, line 11: Did you consider the choice of new columns on the existing table versus an entirely new table? Any thoughts on the tradeoffs there?

geoknee (approved):
Looks good to me. Should there be a new test (or enhancement to an existing test) for this behaviour? I left a few other comments.

packages/server-wallet/src/db/migrations/20201211180815_challenge_status.ts
geoknee: (Just spotted existing discussion under #3070)

packages/server-wallet/src/chain-service/chain-service.ts
lalexgap: This might be a rebase accident actually, I'll take a look!

packages/server-wallet/src/models/challenge-status.ts
lalexgap: That probably makes sense. I omitted it for now since we don't care about the `Challenge Active` status.

packages/server-wallet/src/wallet/wallet.ts
geoknee: I was hacking on this over at #3116.

packages/server-wallet/src/chain-service/chain-service.ts
lalexgap: This has been cleaned up in f6a1027. It should only have been a change to the `args` object but I messed up when rebasing.

lalexgap:
> Looks good to me. Should there be a new test (or enhancement to an existing test) for this behaviour? I left a few other comments.

I just found an issue when writing a test so I think the answer is yes 😄

PR #3116, opened 2021-01-07
refactor: chainServiceSubscriber part of ServerWallet is private

geoknee:
Let's keep an eye on the public interface of `server-wallet`, and try to keep it tidy and well documented.


packages/server-wallet/src/wallet/index.ts
```diff
@@ -4,6 +4,8 @@ export {SingleChannelOutput, MultipleChannelOutput} from './types';
 import {MultiThreadedWallet} from './multi-threaded-wallet';
 import {SingleThreadedWallet} from './wallet';
 
+export {SingleThreadedWallet};
```
geoknee, line 7: https://api-extractor.com/pages/messages/ae-forgotten-export/

This means:
- we get a page for this class in our docs website. 
- we get a summary for this class in the api-report

packages/server-wallet/server-wallet.api.md
```diff
@@ -223,15 +223,101 @@ export type SingleChannelOutput = {
     channelResult: ChannelResult;
 };
 
+// Warning: (ae-forgotten-export) The symbol "EventEmitterType" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "WalletInterface" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export class SingleThreadedWallet extends EventEmitter<EventEmitterType> implements WalletInterface {
```
geoknee, line 230: I intended for this file to show that the methods I made private disappear from this report. As it is, the file was not displaying this class at all until now, so we don't see the intended diff (but a much bigger one). 

The docs website is derived from this file, and we also enforce that it is kept up to date on github via a check on circle. (It is generated during `yarn prepare`, so will trigger `git diff --exit-code` unless it is checked in.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -809,7 +819,7 @@ export class SingleThreadedWallet extends EventEmitter<EventEmitterType>
     const assetHolderAddresses = channelResult.allocations.map(a =>
       makeAddress(a.assetHolderAddress)
     );
-    this.chainService.registerChannel(channelId, assetHolderAddresses, this);
+    this.chainService.registerChannel(channelId, assetHolderAddresses, this.chainEventSubscriber);
```
geoknee, line 822: This change is the main thrust of this PR.

packages/server-wallet/src/wallet/__test__/integration/update-funding.test.ts
```diff
@@ -186,7 +186,8 @@ it('emits new channel result when the funding event is provided via holdingUpdat
     )
   );
 
-  await w.holdingUpdated({
+  await (w as any).holdingUpdated({
+    // typecasting enables us to acess private method without a TS error
```
geoknee, line 190: This doesn't really work, hence the failing test on this branch. 

Perhaps it is a hint that we do actually want these functions to be public?

geoknee:
I'll close this for now, since I think we probably prioritise ease of testing (and other workarounds) over a tidy API (for now).

PR #3127, opened 2021-01-12
feat: store challenge state

### LINKED ISSUE #3126: Store challenge state in database or reconstruct it from call data
We need the state later on when defunding the channel.

lalexgap:
Fixes #3126 

This adds the `challenge-state` column to `ChallengeStatus` allowing us to store the state that was used to challenge.

We need to store the `challenge-state` so once the challenge expires we can call `PushOutcomeAndWithdraw` with that state.

It gets set when the `ChallengeRegistered` event is fired from the `chain-service`

### DIFF
--- a/packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
+++ b/packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
@@ -42,6 +42,12 @@ const chainId = process.env.CHAIN_NETWORK_ID || '9002';
 /* eslint-enable no-process-env, @typescript-eslint/no-non-null-assertion */
 const provider: providers.JsonRpcProvider = new providers.JsonRpcProvider(rpcEndpoint);
 
+const defaultNoopListeners = {
+  holdingUpdated: _.noop,
+  assetOutcomeUpdated: _.noop,
+  channelFinalized: _.noop,
+  challengeRegistered: _.noop,
+};
 let chainService: ChainService;
 let channelNonce = 0;
 async function mineBlock(timestamp?: number) {
@@ -223,13 +229,11 @@ describe('registerChannel', () => {
     const channelFinalizedHandler = jest.fn();
     const channelFinalizedPromise = new Promise<void>(resolve =>
       chainService.registerChannel(channelId, [ethAssetHolderAddress], {
-        holdingUpdated: _.noop,
-        assetOutcomeUpdated: _.noop,
+        ...defaultNoopListeners,
         channelFinalized: arg => {
           channelFinalizedHandler(arg);
           resolve();
         },
-        challengeRegistered: _.noop,
       })
     );
 
@@ -277,10 +281,8 @@ describe('registerChannel', () => {
     };
 
     chainService.registerChannel(channelId, [ethAssetHolderAddress], {
+      ...defaultNoopListeners,
       holdingUpdated,
-      assetOutcomeUpdated: _.noop,
-      channelFinalized: _.noop,
-      challengeRegistered: _.noop,
     });
     await p;
   });
@@ -291,6 +293,7 @@ describe('registerChannel', () => {
 
     await new Promise(resolve =>
       chainService.registerChannel(channelId, [ethAssetHolderAddress], {
+        ...defaultNoopListeners,
         holdingUpdated: arg => {
           expect(arg).toMatchObject({
             channelId,
@@ -299,9 +302,6 @@ describe('registerChannel', () => {
           });
           resolve(true);
         },
-        assetOutcomeUpdated: _.noop,
-        channelFinalized: _.noop,
-        challengeRegistered: _.noop,
       })
     );
   });
@@ -333,10 +333,8 @@ describe('registerChannel', () => {
       if (!objectsToMatch.length) resolve(true);
     };
     chainService.registerChannel(channelId, [ethAssetHolderAddress, erc20AssetHolderAddress], {
+      ...defaultNoopListeners,
       holdingUpdated,
-      assetOutcomeUpdated: _.noop,
-      channelFinalized: _.noop,
-      challengeRegistered: _.noop,
     });
     fundChannelAndMineBlocks(0, 5, channelId, ethAssetHolderAddress);
     fundChannelAndMineBlocks(0, 5, channelId, erc20AssetHolderAddress);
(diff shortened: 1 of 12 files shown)


andrewgordstewart:
> We need to store the `challenge-state` so once the challenge expires we can call `PushOutcomeAndWithdraw` with that state.

Is it too difficult for the chain service to inform the wallet of the finalized outcome?

lalexgap:
> > We need to store the `challenge-state` so once the challenge expires we can call `PushOutcomeAndWithdraw` with that state.
> 
> Is it too difficult for the chain service to inform the wallet of the finalized outcome?

Perhaps I'm misunderstanding but I think we need more than just the `outcome` ?
```
function pushOutcomeAndTransferAll(
        bytes32 channelId,
        uint48 turnNumRecord,
        uint48 finalizesAt,
        bytes32 stateHash,
        address challengerAddress,
        bytes memory outcomeBytes
    )
```

I think longer term we should update the `challenge-state` in the DB when a ChallengeRegistered event is fired. That event contains all the information to create the state.

packages/server-wallet/src/protocols/challenge-submitter.ts
```diff
@@ -65,6 +66,8 @@ export class ChallengeSubmitter {
 
       await this.chainService.challenge([signedState], channel.signingWallet.privateKey);
 
+      // TODO: Store the actual finalizesAt instead of 0
```
andrewgordstewart, line 69: What is the implication of storing `0` instead of the real number?

andrewgordstewart:
> > > We need to store the `challenge-state` so once the challenge expires we can call `PushOutcomeAndWithdraw` with that state.
> > 
> > 
> > Is it too difficult for the chain service to inform the wallet of the finalized outcome?
> 
> Perhaps I'm misunderstanding but I think we need more than just the `outcome` ?

So the chain service should equally be able to provide the other data that's required, right?

> I think longer term we should update the `challenge-state` in the DB when a ChallengeRegistered event is fired. That event contains all the information to create the state.

Is there a reason not to do that now?

packages/server-wallet/src/protocols/challenge-submitter.ts
lalexgap: > So the chain service should equally be able to provide the other data that's required, right?
Yes, the `chain-service` could search the events on-chain and find the state info from a `ChallengeRegistered` event before calling `pushOutcomeAndTransferAll`

Although having to scan through the events seems kinda clunky? Uthink listening for `ChallengeRegistered` and storing the challenge state in the database is more straightforward. 

> Is there a reason not to do that now?
Just avoiding extra effort since it's not needed for `challenging-v0`. Although it would be pretty straightforward to do that.

lalexgap:
> So the chain service should equally be able to provide the other data that's required, right?
Yes, the `chain-service` could search the events on-chain and find the state info from a `ChallengeRegistered` event before calling `pushOutcomeAndTransferAll`

Although having to scan through the events seems kinda clunky? Uthink listening for `ChallengeRegistered` and storing the challenge state in the database is more straightforward. 

> Is there a reason not to do that now?
Just avoiding extra effort since it's not needed for `challenging-v0`. Although it would be pretty straightforward to do that.

packages/server-wallet/src/protocols/challenge-submitter.ts
lalexgap: Hmm, this might be dangerous.

andrewgordstewart:
Ah, I think I understand my confusion now.

The chain-service does not know the finalized outcome at the time that it triggers `ChannelFinalized`. Given that, it doesn't make sense for `channelFinalized` to include the outcome.

In the use case you are going for, there would have been a previous `ChallengeRegistered` event emitted by the chain-service. That is the event that would contain the outcome (and state).

I do think it's worth doing that now, since the code in this diff seems to be temporary code -- the `challengeStatus` should probably _not_ be updated before the `ChallengeRegistered` event has occurred -- what if the transaction fails, or there is a re-org?

lalexgap:
> Ah, I think I understand my confusion now.
> 
> The chain-service does not know the finalized outcome at the time that it triggers `ChannelFinalized`. Given that, it doesn't make sense for `channelFinalized` to include the outcome.
> 
> In the use case you are going for, there would have been a previous `ChallengeRegistered` event emitted by the chain-service. That is the event that would contain the outcome (and state).
> 
> I do think it's worth doing that now, since the code in this diff seems to be temporary code -- the `challengeStatus` should probably _not_ be updated before the `ChallengeRegistered` event has occurred -- what if the transaction fails, or there is a re-org?


I've updated the PR to now update the `ChallengeStatus` table on the `ChallengeRegistered` event.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -509,6 +510,69 @@ describe('challenge', () => {
...
+      channelNonce: state1.channelNonce,
+      chainId: state1.chainId,
+      participants: [alice(), bob()].map(p => p.signingAddress),
+    });
+    const challengeRegistered: Promise<ChallengeRegisteredArg> = new Promise(resolve =>
+      chainService.registerChannel(channelId, [ethAssetHolderAddress], {
+        ...defaultNoopListeners,
+        challengeRegistered: arg => resolve(arg),
+      })
+    );
```
andrewgordstewart, line 556: nit: 
```suggestion
    const challengeRegistered: Promise<ChallengeRegisteredArg> = new Promise(challengeRegistered =>
      chainService.registerChannel(channelId, [ethAssetHolderAddress], {
        ...defaultNoopListeners,
        challengeRegistered
      })
    );
```

packages/server-wallet/src/models/challenge-status.ts
```diff
@@ -1,12 +1,18 @@
+import {State} from '@statechannels/wallet-core';
 import Knex from 'knex';
 import _ from 'lodash';
 import {Model} from 'objection';
 
 import {Bytes32, Uint48} from '../type-aliases';
 
 export type ChallengeStatusResult =
```
andrewgordstewart, line 8: I thought that this table/model was also going to be used to when a `Conclude` transaction succeeded. Is there a reason why it isn't called something like `AdjudicatorStatusResult`?

andrewgordstewart (approved):

packages/server-wallet/src/models/challenge-status.ts
lalexgap: I think something like `AdjudicatorStatusResult` makes sense. Originally I wasn't considering `Conclude`.

PR #3128, opened 2021-01-12
refactor: dedupe transition validation fn

geoknee:
- Removes the `validChain` method on the `Channel` model, replacing it with the existing `validateTransition` function in use elsewhere. 

- Changes both i) the counting app definition address and ii) the default app definition (in text fixtures) so that it is not treated as a null app. This required quite a few changes to ensure consistency across multiple tests

- Adds an `isNullApp` getter to the `Channel` model, alongside the `isLedger` getter. This was because i) we tend to use a zero `appDefinition` (in our test suite at least) to signify a ledger channel; ii) this property is not necessarily 1-1 with ledger channels and iii) this condition is actually the important thing when validating transitions. 

Towards #3035


packages/server-wallet/src/models/channel.ts
```diff
@@ -418,6 +421,12 @@ export class Channel extends Model implements RequiredColumns {
   }
 
   private get _support(): Array<SignedStateWithHash> {
+    // TODO: activate these fields for proper application checks (may be resource hungry)
```
tomclose, line 424: Should this todo be done as part of this work?

packages/server-wallet/src/models/channel.ts
```diff
@@ -426,7 +435,11 @@ export class Channel extends Model implements RequiredColumns {
     for (const signedState of this.sortedStates) {
       // If there is not a valid transition we know there cannot be a valid support
       // so we clear out what we have and start at the current signed state
-      if (previousState && !this.validChain(signedState, previousState)) {
+
+      if (
+        previousState &&
+        !validateTransition(signedState, previousState, logger, byteCode, skipAppTransition)
```
tomclose, line 441: Would you expect this change to affect performance?

tomclose (approved):

packages/server-wallet/src/models/channel.ts
geoknee: I think not, as that would alter the behaviour of the code (not my intention, here). It does highlight the fact that we do not currently check the application-level valid transition function, though. We should probably have an issue to help track our journey getting to that point.

packages/server-wallet/src/models/channel.ts
geoknee: I would not, no. The new function behaves very similarly to the old one when either `skipAppTransition=true` or `skipAppTransition=false && byteCode=undefined`.

packages/server-wallet/src/models/channel.ts
geoknee: For extra context, this comment merely replaces

```
  // We will eventually want to perform a proper validTransition check	
  // but we will have to be careful where we do that to prevent eating up a ton of cpu
```

packages/server-wallet/src/models/channel.ts
geoknee: In other words, the app transition check will fail early for null apps, without hitting pure evm.

PR #3129, opened 2021-01-12
Refactor the process ledger queue protocol

geoknee:
The idea behind these change is to bring this protocol in line with other recent refactors. i.e. to have a `LedgerManager` that is `cranked` (in a similar fashion to how the `ObjectiveManager` is cranked). 

The initial goal will be to maintain current functionality, but to increase clarity wherever possible. This may involve deprecating `protocolState` and `actions`, as we have been doing elsewhere. 

A possibly-useful side effect will be increased familiarity and understanding of the underlying protocol that this code implements. That should aid in judging it against alternatives.


snario:
Glad that we can better encapsulate the "pessimistic syncing" code

snario:
One thing I dislike is that we're coupling interactions with the `store` into logic related to the protocol. For example, if the protocol detects that the merged outcome is equal to the current outcome, there is a call to `store.removeLedgerProposals`. This makes it harder to write unit tests for the protocol by requiring a spy on the store, or a mock store, or only supporting integration tests.

geoknee:
Side comment: this draft PR is pretty old and I am probably going to close it in favour of a fresh one, although the aim is going to be the same. 

I agree that by peforming this refactor we are coupling the store and the protocol more tightly, and this makes tests harder to write, and slower. But the hope is that the other advantages outweigh this.

I suppose @tomclose can speak to those advantages better than me. I think part of the rationale in other similar refactors is that the "protocol" can make calls to the store more judiciously / parsimoniously. 

I'll take some time to think it over some more though, because I think you raise a good point.

geoknee:
I prepared a PR that reintroduces the pure-function test suite. I think that could live alongside the current implementation and the new tests https://github.com/statechannels/statechannels/pull/3192

If we follow through on the refactor, we can just delete it again.

packages/server-wallet/src/protocols/ledger-manager.ts
```diff
@@ -55,108 +46,217 @@ export class LedgerManager {
...
-    }
+        response.queueChannelState(protocolState.fundingChannel);
+        return;
+      }
+
+      const crankAgain = await this.markIncludedRequestsAsComplete(protocolState, tx);
+      requiresAnotherCrankUponCompletion = requiresAnotherCrankUponCompletion || crankAgain;
+
+      // refresh the protocol state to be safe
+      // TODO: remove if possible
```
snario, line 71: I think we ought to ensure this is linked to an issue

packages/server-wallet/src/protocols/ledger-manager.ts
```diff
@@ -55,108 +46,217 @@ export class LedgerManager {
...
+      if (finishedExchangingProposals(protocolState)) {
+        const crankAgain2 = await this.exchangeSignedLedgerStates(protocolState, response, tx);
+        requiresAnotherCrankUponCompletion = requiresAnotherCrankUponCompletion || crankAgain2;
+      }
+
+      // refresh the protocol state to be safe
+      // TODO: remove if possible
+      protocolState = await getProcessLedgerQueueProtocolState(this.store, ledgerChannelId, tx);
+
+      // exchange proposals - when do I need to do this?
```
snario, line 83: Is this still an open question?

packages/server-wallet/src/protocols/ledger-manager.ts
```diff
@@ -55,108 +46,217 @@ export class LedgerManager {
...
+
+  private async signLedgerUpdate(
+    fundingChannel: ChannelStateWithSupported,
+    stateToSign: StateVariables,
+    response: WalletResponse,
+    tx: Transaction
+  ) {
+    const {myIndex, channelId} = fundingChannel;
+    const channel = await Channel.forId(channelId, tx);
+    const signedState = await this.store.signState(channel, stateToSign, tx);
```
snario, line 109: Why are we fetching the channel from the DB again if we have a `fundingChannel` available?

packages/server-wallet/src/protocols/ledger-manager.ts
tomclose: No. Will remove.

packages/server-wallet/src/protocols/ledger-manager.ts
```diff
@@ -55,108 +46,217 @@ export class LedgerManager {
...
+    /**
+     * After we have completed some funding requests (i.e., a new ledger state
+     * has been signed), we can confidently clear now-stale proposals from the DB.
+     */
+    await this.store.removeLedgerProposals(ledgerChannelId, tx);
+
+    await this.store.markLedgerRequests(fundedChannels, 'fund', 'succeeded', tx);
+    await this.store.markLedgerRequests(defundedChannels, 'defund', 'succeeded', tx);
+
+    return true; // requiesAnotherCrankUponCompletion
```
snario, line 166: typo

packages/server-wallet/src/protocols/ledger-manager.ts
```diff
@@ -55,108 +46,217 @@ export class LedgerManager {
...
+      // sign ledger update
+      const stateToSign = {
+        turnNum: nextTurnNum,
+        outcome,
+        appData: NULL_APP_DATA,
+        isFinal: false,
+      };
+
+      await this.signLedgerUpdate(protocolState.fundingChannel, stateToSign, response, tx);
+      return false;
```
snario, line 217: Why doesn't this require another crank?

packages/server-wallet/src/protocols/ledger-manager.ts
tomclose: This was just a direct translation of what was there before:
```ts
  case 'SignLedgerUpdate': {	
      const {myIndex, channelId} = protocolState.fundingChannel;	
      const channel = await Channel.forId(channelId, tx);	
      const signedState = await this.store.signState(channel, action.stateToSign, tx);	
      response.queueState(signedState, myIndex, channelId);	
      return;	
    }
```

snario (approved):

tomclose:
Todo:
- [ ] Link against the graph repo and run tests

geoknee:
> Todo:
> 
> * [ ]  Link against the graph repo and run tests [? turns out the two codebases aren't compatible at the moment :( ]

I linked this branch against [this one in the other repo](https://github.com/statechannels/graph-payments/pull/10). Unit tests passed. I wasn't able to run e2e tests yet, I just see errors on my local machine. To be continued.

geoknee:
I was in fact able to link and get tests to pass in the other repo https://github.com/statechannels/graph-payments/pull/15. 

However tests are now failing in this repo since I solved conflicts with `master` 😞

geoknee:
The test issue could be a flicker https://github.com/statechannels/statechannels/issues/3208

PR #3131, opened 2021-01-12
refactor: wrap crank in a transaction

snario:
Wraps `crank` in a transaction as a means of ensuring a crank is done inside of a transaction. Enables the ability to provide `tx` object to `getObjectives` which enables usage of `.forUpdate` to avoid cranking the same objective in two threads or processes.

This pull request was inspired after noticing that #3109 appeared to [demonstrate a flickering test](https://app.circleci.com/pipelines/github/statechannels/statechannels/11244/workflows/8dd29206-1558-4bc9-92ec-3b1733ee98af/jobs/54895) that indicated an objective was being operated on despite having been deleted. The [server wallet architecture doc](https://www.notion.so/elfour/Server-wallet-updated-architecture-ad0372e611ac485aa16dc2c4df65a14f) implies that an objective ought to only ever be handled by a single process as there is always only one logical "crank" to be done with it. We have a partial implementation of that now as the _channel_ is locked within the execution of the crank, but the objective remains freely usable. It seems safe to wrap the crank as a whole inside a transaction to mitigate this risk.

### DIFF
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -177,6 +177,7 @@ export class ObjectiveModel extends Model {
     const objectiveIds = (
       await ObjectiveChannelModel.query(tx)
         .column('objectiveId')
+        .forUpdate()
         .select()
         .whereIn('channelId', targetChannelIds)
     ).map(oc => oc.objectiveId);
--- a/packages/server-wallet/src/objectives/objective-manager.ts
+++ b/packages/server-wallet/src/objectives/objective-manager.ts
@@ -1,6 +1,5 @@
 import {Logger} from 'pino';
 import {unreachable} from '@statechannels/wallet-core';
-import {Transaction} from 'objection';
 
 import {Bytes32} from '../type-aliases';
 import {ChannelOpener} from '../protocols/channel-opener';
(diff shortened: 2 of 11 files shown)


andrewgordstewart:
Can you explain why `crankUntilIdle` should be an atomic action? Is it obvious?

Also, is it obvious that two actors should not be able to concurrently operate on the same objective? (It seems reasonable to me, but I do not know exactly how the wallet works.)

tomclose:
This looks like it introduces the possibility of deadlocks, as we have one transaction over the course of which multiple channels and objectives are locked for update. I don't believe any of those locks are released until the transaction completes. If we end up processing two messages at once, or a message and an app/chain update, we could be in trouble.

snario:
@tomclose yeah, I had intended for the transaction to wrap just a single crank, fixed in be12e51

packages/server-wallet/src/models/objective.ts
```diff
@@ -177,6 +177,7 @@ export class ObjectiveModel extends Model {
     const objectiveIds = (
       await ObjectiveChannelModel.query(tx)
         .column('objectiveId')
+        .forUpdate()
```
geoknee, line 180: Two observations about this: 

1. You are only putting the lock on the join table `objectives_channels`
2. `forChannelIds` is not actually used in the objective manager.

geoknee:
Closing in favour of #3321, because

- this PR is pretty old and has many conflicts
- the bulk of the changes have actually been implemented on `master` in the mean time. 
- the remaining changes looked very small, but a bit wrong, to me. So I just redid them.

geoknee:
(@snario hope this makes sense to you, else feel free to reopen).

Issue #3122, opened 2021-01-08
EnsureObjective always returns a status of pending


lalexgap:
Currently if an objective already exists we will return a status of `pending` for it even if the existing objective has a different status.

_Originally posted by @lalexgap in https://github.com/statechannels/statechannels/pull/3121#discussion_r554099702_

lalexgap:
I think the solution is to create a `insertOrFetch` method that `ensureObjective` can use

geoknee:
via #3259

PR #3133, opened 2021-01-12
Defund Channel Protocol

### LINKED ISSUE #3113: Create `DefundChannel` protocol
See discussion [here](https://www.notion.so/Jan-11-Close-channel-protocol-4a167972a7264e1ebbb594c4624c152b#04c8a45842c645d9914802b956f3c3b7)

Create a new protocol (`DefundChannel` ?) that initially handles this part of the current `CloseChannel` protocol.

> kicking off blockchain transactions, **when needed (?)**

Initially, we assume that **the channel is directly funded**.

It should ensure that the funds deposited against the channel are transferred to the wallet owner's destinations, by triggering a `PushOutcomeAndWithdraw` via the chain service

### Out of scope

- Emitting/handling ChannelDefunded event.

lalexgap:
Fixes #3113 

Introduces a `DefundChannel` protocol that is responsible for defunding a channel by calling either `pushOutcomeAndTransferAll` or `concludePushOutcomeAndTransferAll` on chain.

A `DefundChannel` objective now gets added when a channel is finalized.

The `DefundChannel` protocol is pretty simple and does the following:
- Checks if the channel is finalized on chain (by consulting the `ChallengeStatus` table) 
- If so it calls `pushOutcomeAndTransferAll`
- If not check if the channel has a conclusion proof
- If so it calls `concludePushOutcomeAndTransferAll`
- If not do nothing

### DIFF
--- a/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
@@ -30,18 +30,40 @@ beforeEach(async () => {
   await seedAlicesSigningWallet(knex);
 });
 
+test('it should submit a conclude transaction even if there is an active challenge if there is a conclusion proof', async () => {
+  const chainService = new MockChainService();
+  const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
+
+  const c = channel({
+    channelNonce: 1,
+    vars: [stateWithHashSignedBy([alice(), bob()])({isFinal: true})],
+  });
+
+  await Channel.query(knex).withGraphFetched('signingWallet').insert(c);
+
+  const challengeState = {
+    ...c.channelConstants,
+    ...stateVars(),
+  };
+  await ChallengeStatus.insertChallengeStatus(knex, c.channelId, 100, challengeState);
+  const obj = createPendingObjective(c.channelId);
+  await knex.transaction(tx => ObjectiveModel.insert(obj, tx));
+  const response = WalletResponse.initialize();
+  const concludeSpy = jest.spyOn(chainService, 'concludeAndWithdraw');
+
+  await channelDefunder.crank(obj, response);
+
+  const reloadedObjective = await store.getObjective(obj.objectiveId);
+  expect(reloadedObjective.status).toEqual('succeeded');
+  expect(concludeSpy).toHaveBeenCalled();
+});
+
 test('when the channel does not exist it should fail the objective', async () => {
   const chainService = new MockChainService();
   const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
   const c = channel();
 
-  const obj: DBDefundChannelObjective = {
-    type: 'DefundChannel',
-    status: 'pending',
-    objectiveId: ['DefundChannel', c.channelId].join('-'),
-    data: {targetChannelId: c.channelId},
-  };
-
+  const obj = createPendingObjective(c.channelId);
   await knex.transaction(tx => ObjectiveModel.insert(obj, tx));
   const response = WalletResponse.initialize();
 
@@ -50,16 +72,13 @@ test('when the channel does not exist it should fail the objective', async () =>
   const reloadedObjective = await store.getObjective(obj.objectiveId);
   expect(reloadedObjective.status).toEqual('failed');
 });
-
 test('when using non-direct funding it fails', async () => {
   const chainService = new MockChainService();
   const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
 
   const c = channel({fundingStrategy: 'Fake'});
 
-  await Channel.query(knex)
-    .withGraphFetched('signingWallet')
-    .insert(c);
+  await Channel.query(knex).withGraphFetched('signingWallet').insert(c);
 
   const obj = createPendingObjective(c.channelId);
   await knex.transaction(tx => store.ensureObjective(obj, tx));
@@ -71,7 +90,7 @@ test('when using non-direct funding it fails', async () => {
   expect(reloadedObjective.status).toEqual('failed');
 });
 
-test('when there is an active challenge it should do nothing', async () => {
+test('when there is an active challenge and no conclusion proof it should do nothing', async () => {
   const chainService = new MockChainService();
   const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
 
@@ -80,9 +99,7 @@ test('when there is an active challenge it should do nothing', async () => {
     ...c.channelConstants,
     ...stateVars(),
   };
-  await Channel.query(knex)
-    .withGraphFetched('signingWallet')
-    .insert(c);
+  await Channel.query(knex).withGraphFetched('signingWallet').insert(c);
   await ChallengeStatus.insertChallengeStatus(knex, c.channelId, 100, challengeState);
 
   const obj = createPendingObjective(c.channelId);
@@ -103,9 +120,7 @@ test('when the channel is not finalized it defunds the channel by calling pushOu
     ...c.channelConstants,
     ...stateVars(),
   };
-  await Channel.query(knex)
-    .withGraphFetched('signingWallet')
-    .insert(c);
+  await Channel.query(knex).withGraphFetched('signingWallet').insert(c);
   await ChallengeStatus.insertChallengeStatus(knex, c.channelId, 100, challengeState);
   await ChallengeStatus.setFinalized(knex, c.channelId, 200);
   const obj = createPendingObjective(c.channelId);
@@ -129,9 +144,7 @@ test('when the channel has not been concluded on chain it should call withdrawAn
     vars: [stateWithHashSignedBy([alice(), bob()])({isFinal: true})],
   });
 
-  await Channel.query(knex)
-    .withGraphFetched('signingWallet')
-    .insert(c);
+  await Channel.query(knex).withGraphFetched('signingWallet').insert(c);
 
   await ChallengeStatus.setFinalized(knex, c.channelId, 200);
   const obj = createPendingObjective(c.channelId);
--- a/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
@@ -72,6 +72,7 @@ test('when the channel does not exist it should fail the objective', async () =>
   const reloadedObjective = await store.getObjective(obj.objectiveId);
   expect(reloadedObjective.status).toEqual('failed');
 });
+
 test('when using non-direct funding it fails', async () => {
   const chainService = new MockChainService();
   const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
--- a/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
@@ -1,0 +1,127 @@
+import {State} from '@statechannels/wallet-core';
+
+import {defaultTestConfig} from '../..';
+import {createLogger} from '../../logger';
+import {DBDefundChannelObjective} from '../../models/objective';
+import {Store} from '../../wallet/store';
+import {testKnex as knex} from '../../../jest/knex-setup-teardown';
+import {seedAlicesSigningWallet} from '../../db/seeds/1_signing_wallet_seeds';
+import {channel} from '../../models/__test__/fixtures/channel';
+import {Channel} from '../../models/channel';
+import {MockChainService} from '../../chain-service';
+import {WalletResponse} from '../../wallet/wallet-response';
+import {ChannelDefunder} from '../defund-channel';
+import {ChallengeStatus} from '../../models/challenge-status';
+import {stateVars} from '../../wallet/__test__/fixtures/state-vars';
+import {stateWithHashSignedBy} from '../../wallet/__test__/fixtures/states';
+import {alice, bob} from '../../wallet/__test__/fixtures/signing-wallets';
+
+const logger = createLogger(defaultTestConfig());
+const timingMetrics = false;
+
+let store: Store;
+beforeEach(async () => {
+  store = new Store(
+    knex,
+    defaultTestConfig().metricsConfiguration.timingMetrics,
+    defaultTestConfig().skipEvmValidation,
+    '0'
+  );
+
+  await store.dbAdmin().truncateDB();
+  await seedAlicesSigningWallet(knex);
+});
+
+describe('when a channel is finalized on chain', () => {
+  it('defunds the channel by calling pushOutcomeAndWithdraw', async () => {
+    const c = channel();
+
+    await Channel.query(knex)
+      .withGraphFetched('signingWallet')
+      .insert(c);
+    await ChallengeStatus.insertChallengeStatus(knex, c.channelId, 100, {
+      ...c.channelConstants,
+      ...stateVars(),
+    });
+    await ChallengeStatus.setFinalized(knex, c.channelId, 200);
+    const obj: DBDefundChannelObjective = {
+      type: 'DefundChannel',
+      status: 'pending',
+      objectiveId: ['DefundChannel', c.channelId].join('-'),
+      data: {targetChannelId: c.channelId},
+    };
+
+    await knex.transaction(tx => store.ensureObjective(obj, tx));
+    await await crankAndAssert(obj, {calls: 'pushOutcomeAndWithdraw', completesObj: true});
+  });
+  it('if the channel has not been concluded on chain it should call withdrawAndPushOutcome', async () => {
+    const c = channel({
+      channelNonce: 1,
+      vars: [stateWithHashSignedBy([alice(), bob()])({isFinal: true})],
+    });
+
+    await Channel.query(knex)
+      .withGraphFetched('signingWallet')
+      .insert(c);
+
+    await ChallengeStatus.setFinalized(knex, c.channelId, 200);
+    const obj: DBDefundChannelObjective = {
+      type: 'DefundChannel',
+      status: 'pending',
+      objectiveId: ['DefundChannel', c.channelId].join('-'),
+      data: {targetChannelId: c.channelId},
+    };
+
+    await knex.transaction(tx => store.ensureObjective(obj, tx));
+    await await crankAndAssert(obj, {completesObj: true, calls: 'concludeAndWithdraw'});
+  });
+});
+afterEach(async () => {
+  await store.dbAdmin().truncateDB();
+});
+
+interface AssertionParams {
+  challengeState?: State;
+  calls: 'pushOutcomeAndWithdraw' | 'concludeAndWithdraw' | 'none';
+  completesObj?: boolean;
+}
+
+const crankAndAssert = async (
+  objective: DBDefundChannelObjective,
+  args: AssertionParams
+): Promise<void> => {
+  const completesObj = args.completesObj || false;
+
+  const chainService = new MockChainService();
+  const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
+  const response = WalletResponse.initialize();
+  const pushSpy = jest.spyOn(chainService, 'pushOutcomeAndWithdraw');
+  const withdrawSpy = jest.spyOn(chainService, 'concludeAndWithdraw');
+  await channelDefunder.crank(objective, response);
+
+  if (args.calls === 'none') {
+    expect(pushSpy).not.toHaveBeenCalled();
+    expect(withdrawSpy).not.toHaveBeenCalled();
+  } else if (args.calls === 'pushOutcomeAndWithdraw') {
+    if (args.challengeState) {
+      expect(pushSpy).toHaveBeenCalledWith(
+        expect.objectContaining(args.challengeState),
+        expect.anything()
+      );
+    } else {
+      expect(pushSpy).toHaveBeenCalled();
+    }
+    expect(withdrawSpy).not.toHaveBeenCalled();
+  } else {
+    expect(withdrawSpy).toHaveBeenCalled();
+    expect(pushSpy).not.toHaveBeenCalled();
+  }
+
+  const reloadedObjective = await store.getObjective(objective.objectiveId);
+
+  if (completesObj) {
+    expect(reloadedObjective.status).toEqual('succeeded');
+  } else {
+    expect(reloadedObjective.status).toEqual('pending');
+  }
+};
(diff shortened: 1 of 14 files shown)


andrewgordstewart:
🔥

packages/server-wallet/src/protocols/defund-channel.ts
```diff
@@ -0,0 +1,71 @@
...
+        // TODO: https://github.com/statechannels/statechannels/issues/3124
+        this.logger.error(`Only direct funding is currently supported.`);
+        await this.store.markObjectiveStatus(objective, 'failed', tx);
+        return;
+      }
+
+      const result = await ChallengeStatus.getChallengeStatus(tx, channelId);
+
+      if (result.status === 'Challenge Active') {
+        this.logger.warn('There is a challenge active so cannot defund the channel');
```
tomclose, line 50: Is this true? If there's a conclusion proof I think we should still be able to defund it, right?

lalexgap: Hmm it looks like you're right. I'll fix this up.

lalexgap: Updated in 7237e79

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  store = new Store(
+    knex,
+    defaultTestConfig().metricsConfiguration.timingMetrics,
+    defaultTestConfig().skipEvmValidation,
+    '0'
+  );
+
+  await store.dbAdmin().truncateDB();
+  await seedAlicesSigningWallet(knex);
+});
```
andrewgordstewart, line 31: I imagine that this is a common pattern in the server-wallet codebase now. Is there a test utility to dry up this test setup code?

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+test('it should submit a conclude transaction even if there is an active challenge if there is a conclusion proof', async () => {
+  const chainService = new MockChainService();
```
andrewgordstewart, line 34: I find it really difficult to read this type of test.
- the majority of the test case is spent setting up the DB
- there are six test cases, and there appears to be a lot of repeated code
- it is difficult to look at the code, and draw a connection to the test's description

Is there no common pattern like
```
function setUpStore(testData: TestData) {
  ///
}
```

so that you can call
```
setUpStore({
  channels: [channelWithConclusionProof],
  challengeStatuses: [{channelId: c.channelId, finalizesAt: 100, challengeState}],
  ...
})
```

As something that you can do in this PR, can you at least clearly label the test setup from testing the behaviour?
```
// SETUP
const channelWithConclusionProof = ...
Store.insertChannel(channelWithConclusionProof)

// TEST
await channelDefunder.crank(obj, response);
expect(obj.status).toEqual('succeeded')
```

packages/server-wallet/src/protocols/defund-channel.ts
```diff
@@ -0,0 +1,66 @@
...
+      // applies to both co-operatively concluding or challenging
+      // see https://github.com/statechannels/statechannels/issues/3132
+      if (result.status === 'Challenge Finalized') {
+        await ChainServiceRequest.insertOrUpdate(channelId, 'pushOutcome', tx);
+        this.chainService.pushOutcomeAndWithdraw(result.challengeState, channel.myAddress);
+        await this.store.markObjectiveStatus(objective, 'succeeded', tx);
+
+        return;
+      } else if (channel.hasConclusionProof) {
+        await ChainServiceRequest.insertOrUpdate(channelId, 'withdraw', tx);
```
andrewgordstewart, line 59: Why is this model used here? Shouldn't it use a `Store` method?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -391,14 +400,37 @@ export class Store {
...
+    objective: DefundChannel,
+    tx: Transaction
+  ): Promise<DBObjective> {
+    const {data} = objective;
+    const {targetChannelId} = data;
+    // fetch the channel to make sure the channel exists
+    const channel = await this.getChannelState(targetChannelId, tx);
+    if (!channel) {
+      throw new StoreError(StoreError.reasons.channelMissing, {channelId: targetChannelId});
+    }
```
andrewgordstewart, line 422: Shouldn't there be a foreign key constraint that guarantees this?

packages/server-wallet/src/wallet/store.ts
```diff
@@ -391,14 +400,37 @@ export class Store {
...
+
+    try {
+      await ObjectiveModel.insert({...objective, status: 'pending'}, tx);
+    } catch (e) {
+      // If the objective exists our job is done
+      if (e.name !== 'UniqueViolationError') throw e;
+    }
```
andrewgordstewart, line 429: I think the `ObjectiveModel` should handle this logic? Right now, the internal model implementation has leaked into the store.

Postgres allows `ON CONFLICT`, so the `ObjectiveModel` can do, for instance,
```
INSERT foo
ON CONFLICT (name of uniqueness constraint) DO NOTHING
```

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -744,7 +744,19 @@ export class SingleThreadedWallet
 
   async channelFinalized(arg: ChannelFinalizedArg): Promise<void> {
     const response = WalletResponse.initialize();
+
     await this.store.setFinalizedChallengeStatus(arg.channelId, arg.blockNumber);
+    await this.knex.transaction(async tx => {
+      const {objectiveId} = await this.store.ensureObjective(
+        {
+          type: 'DefundChannel',
+          participants: [],
```
andrewgordstewart, line 753: Why is `participants` a property of this objective if it's empty?

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  };
+  await ChallengeStatus.insertChallengeStatus(knex, c.channelId, 100, challengeState);
+  const obj = createPendingObjective(c.channelId);
+  await knex.transaction(tx => ObjectiveModel.insert(obj, tx));
+  const response = WalletResponse.initialize();
+  const concludeSpy = jest.spyOn(chainService, 'concludeAndWithdraw');
+
+  await channelDefunder.crank(obj, response);
+
+  const reloadedObjective = await store.getObjective(obj.objectiveId);
```
andrewgordstewart, line 56: Why doesn't `crank` return the new objective's value?

andrewgordstewart (commented):
Not approving this yet, since I did not carefully read and understand the test cases.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  const obj = createPendingObjective(c.channelId);
+  await knex.transaction(tx => ObjectiveModel.insert(obj, tx));
+  const response = WalletResponse.initialize();
+
+  await channelDefunder.crank(obj, response);
+
+  const reloadedObjective = await store.getObjective(obj.objectiveId);
+  expect(reloadedObjective.status).toEqual('failed');
+});
+test('when using non-direct funding it fails', async () => {
```
andrewgordstewart, line 75:
```suggestion
});

test('when using non-direct funding it fails', async () => {
```

Whitespace makes tests easier to read.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  expect(concludeSpy).toHaveBeenCalled();
+});
+
+test('when the channel does not exist it should fail the objective', async () => {
+  const chainService = new MockChainService();
+  const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
+  const c = channel();
+
+  const obj = createPendingObjective(c.channelId);
+  await knex.transaction(tx => ObjectiveModel.insert(obj, tx));
```
andrewgordstewart, line 67: Why do you need to start a transaction here? Can you abuse javascript and just do something like
```suggestion
  await ObjectiveModel.insert(obj, undefined as any);
```

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  await knex.transaction(tx => store.ensureObjective(obj, tx));
+  const response = WalletResponse.initialize();
+
+  await channelDefunder.crank(obj, response);
+
+  const reloadedObjective = await store.getObjective(obj.objectiveId);
+  expect(reloadedObjective.status).toEqual('pending');
+});
+
+test('when the channel is not finalized it defunds the channel by calling pushOutcomeAndWithdraw', async () => {
```
andrewgordstewart, line 115:
```suggestion
test('when the channel is now finalized it defunds the channel by calling pushOutcomeAndWithdraw', async () => {
```

?

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  const chainService = new MockChainService();
+  const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
+  const c = channel();
+  const challengeState = {
+    ...c.channelConstants,
+    ...stateVars(),
+  };
+  await Channel.query(knex).withGraphFetched('signingWallet').insert(c);
+  await ChallengeStatus.insertChallengeStatus(knex, c.channelId, 100, challengeState);
+  await ChallengeStatus.setFinalized(knex, c.channelId, 200);
```
andrewgordstewart, line 125: Why does it take two DB queries to set it as finalized?

(What if the chain service came online, and found a finalized channel?)

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  const pushSpy = jest.spyOn(chainService, 'pushOutcomeAndWithdraw');
+
+  await channelDefunder.crank(obj, response);
+
+  expect(pushSpy).toHaveBeenCalledWith(challengeState, c.myAddress);
+  const reloadedObjective = await store.getObjective(obj.objectiveId);
+  expect(reloadedObjective.status).toEqual('succeeded');
+});
+
+test('when the channel has not been concluded on chain it should call withdrawAndPushOutcome', async () => {
```
andrewgordstewart, line 139: This test description does not seem to match the content of the test.

packages/server-wallet/src/protocols/defund-channel.ts
```diff
@@ -0,0 +1,66 @@
...
+    objective: DBDefundChannelObjective,
+    _response: WalletResponse
+  ): Promise<void> {
+    const {targetChannelId: channelId} = objective.data;
+    await this.store.transaction(async tx => {
+      const channel = await this.store.getAndLockChannel(channelId, tx);
+
+      if (!channel) {
+        this.logger.error(`No channel found for channel id ${channelId}`);
+        await this.store.markObjectiveStatus(objective, 'failed', tx);
```
andrewgordstewart, line 36: Why don't objectives have failure reasons? It seems important for a wallet user to know why an objective failed.

packages/server-wallet/src/wallet/store.ts
lalexgap: The `data` of an `objective` does not necessarily have a `channelId`. It's also just `jsonb` column so it seems like this would be hard to encode as a foreign key.

packages/server-wallet/src/wallet/wallet.ts
lalexgap: Since `wallet-core` defines `participants` as a property on the `objective` and `ensureObjective` takes a `Objective` from `wallet-core`.

Sidenote: I'd personally prefer having the `server-wallet` stop using `wallet-core` for the `objective` structure.

```
type _Objective<Name, Data> = {
  participants?: Participant[];
  type: Name;
  data: Data;
};
```

packages/server-wallet/src/wallet/store.ts
lalexgap: I agree with this. Mea culpa: I just copied this from the existing `ensure` methods.

packages/server-wallet/src/protocols/defund-channel.ts
lalexgap: I thought a bit about this but decided it was outside the scope of this PR. [It seems like we're removing the `failed` and` succeeded` status anways?](https://github.com/statechannels/statechannels/pull/3109/files)

I think this flags a good point though about how we want to handle "failed" objectives longer term. Do we just throw an exception or log an entry? Or do we want to update the objective database?

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
lalexgap: Wouldn't that fail horribly? We use the `tx` that's passed in?

packages/server-wallet/src/protocols/defund-channel.ts
lalexgap: Is the intention of this that our protocols only rely on a `store` and not on the specific `Postgres` implementation?

If so that makes sense to me. However, our protocols make references to models in a bunch of places besides this one, so we should probably change those as well.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
lalexgap: The `ObjectiveManager` assumes a signature of`Promise<void>`. I think it would be fine to change all `crank` methods to return the objective but I think the behaviour should be consistent across all of them.

Also, I think there is value in fetching the objective from the DB to avoid issues like #3122 where we return something different than what we store.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
lalexgap: I think there is a benefit to explicitly doing the setup. From this, you know that before each test
- creating a new `store`
- truncating the database
- seeding only alice's signing wallet

Perhaps I'm being paranoid but I'm worried about some utility obscuring some of this.  Explicitly saying who's wallet you'll be using and whether you're wiping the database seems to be pretty relevant info that various by test.

I'd be happy with something like this?:
```
const store = TestUtil.createStore();
await TestUtil.truncate();
await seedAlicesSigningWallet(knex);
```

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
lalexgap: I agree. It would be nice if this was a `lint` rule and was automatically formatted. I think I assumed that linting would take care of it

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
lalexgap: I think these tests can be cleaned up! I think my preference is for descriptions with `comments`?  I feel like some attempts to DRY up test code makes it harder to understand but perhaps that a preference?

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
lalexgap: Originally this was designed **only** with challenges in mind so we could always rely on the `ChallengeRegistered` event occurring first. Oops!

We should probably fix this in #3134

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
andrewgordstewart: I'm not sure if the transaction _javascript_ methods are used internally. My understanding is that the main point of the transaction that gets passed around is to ensure consistency.

I remember now that `ObjectiveModel` does need to be passed a DB connection. So maybe this will do?
```suggestion
  await ObjectiveModel.insert(obj, knex as any);
```

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
andrewgordstewart: 💡 `crank` can be a generic function on the `ObjectiveManager`, accepting an objective of type `T`, and returning `Promise<T>`

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: Happy to leave this as a future refactoring work, if the pattern exists elsewhere.

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: I am unfortunately unfamiliar with the current schema. My gut thought, when I wrote this, that there ought to be built in protection against there being no channel for an objective, if it's important for the objective to be attached to a channel.

packages/server-wallet/src/protocols/defund-channel.ts
andrewgordstewart: If it's an existing pattern, I think it's fine. But I am doubtful we will be able to maintain that pattern.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
andrewgordstewart: Assuming that it's commonly used (which I think it is), I think the fact that there's already `seedAlicesSigningWallet` is an indication that it's a common enough pattern to encapsulate all that into something like
```
store = TestUtil.blankStore(knex, {seededWalet: 'Alice'})
```

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
andrewgordstewart: I see your point. One particular issue here is, it is difficult to connect the test description to the test data, which is worsened because the test data is interspersed with general test setup & data insertion.

If you don't want to DRY up the test code, maybe one option is something like this?
```
test('it should submit a conclude transaction even if there is an active challenge if there is a conclusion proof', async () => {
  // Insert channel with conclusion proof
  const c = channel({
    channelNonce: 1,
    vars: [stateWithHashSignedBy([alice(), bob()])({isFinal: true})],
  });
  await Channel.query(knex).withGraphFetched('signingWallet').insert(c);

  // Insert active challenge
  const challengeState = {
    ...c.channelConstants,
    ...stateVars(),
  };
  await ChallengeStatus.insertChallengeStatus(knex, c.channelId, 100, challengeState);

  // Insert pending objective
  const obj = createPendingObjective(c.channelId);
  await knex.transaction(tx => ObjectiveModel.insert(obj, tx));

  // Test setup
  const chainService = new MockChainService();
  const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
  const concludeSpy = jest.spyOn(chainService, 'concludeAndWithdraw');

  // Test
  await channelDefunder.crank(obj, response);

  const reloadedObjective = await store.getObjective(obj.objectiveId);
  expect(reloadedObjective.status).toEqual('succeeded');
  expect(concludeSpy).toHaveBeenCalled();
});
```

Though I'd note here that you could already DRY up the code by overwriting a `chainService` and `channelDefunder` in the `beforeEach`, since they are common to all the tests.

packages/server-wallet/src/protocols/defund-channel.ts
andrewgordstewart: Can someone to explain how "failed" objectives are meant to be handled?

andrewgordstewart (approved):
BTW, I'm happy with whatever previous comments you address.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
andrewgordstewart: I think your suggestion is also fine.

When I also see this [over here](https://github.com/statechannels/statechannels/pull/3146/files#diff-039e86d4e9d5638d7541ebf8d0a2922f71c752b20b6cb6c15165b454ab664899R93-R105), it seems like a code smell. What happens when you add a new required argument to the store function? You'd then need to update every test file that creates a store.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
lalexgap: I've cleaned up the tests in ec574d4. I think grouping them with descriptions and creating some simple helper functions has gone a long way to make them easier to understand

```
 PASS  src/protocols/__test__/defund-channel.test.ts (16.891 s)
  ✓ should fail the objective when the channel does not exist (555 ms)
  ✓ should fail when using non-direct funding (509 ms)
  when there is no challenge or finalized channel
    ✓ should do nothing (589 ms)
  when there is an active challenge
    ✓ should submit a conclude transaction if there is a conclusion proof (1020 ms)
    ✓ should do nothing if there is no conclusion proof (510 ms)
  when the channel is finalized on chain
    ✓ should call pushOutcome if the on-chain state is not final (494 ms)
    ✓ should do nothing if the on-chain state is final (526 ms)
```

andrewgordstewart: 👍

packages/server-wallet/src/protocols/defund-channel.ts
lalexgap: > Can someone to explain how "failed" objectives are meant to be handled?

I don't think there's much you can do besides retry them? They might be useful for debugging but I feel like we could just log that data out.

Considering this is the only place we use them and t[hey're being removed soon](https://github.com/statechannels/statechannels/pull/3109/files), I'll just remove them.

andrewgordstewart: I don't really know that they should be removed.

Currently, if the channel manager calls createChannel and it fails because there's insufficient funding in the ledger channel, there's no way for the channel manager to find that out. That appears to be a hole in the wallet API?

packages/server-wallet/src/wallet/store.ts
lalexgap: I think this probably makes sense but it would probably require changing the current objective schema a bit and I don't think it's worth the pain right now?

andrewgordstewart: Yep!

packages/server-wallet/src/protocols/defund-channel.ts
andrewgordstewart: (in this case, it would seem that the channel manager might react by depositing in the ledger channel before trying again.)

I'm not suggesting you make any change in this pr

lalexgap: Ah good point!

Perhaps a discussion or spike is warranted here

PR #3134, opened 2021-01-13
refactor: rename challenge status to adjudicator status

### LINKED ISSUE #3132: Refactor Challenge Status to AdjudicatorStatusResult
I thought that this table/model was also going to be used to when a `Conclude` transaction succeeded. Is there a reason why it isn't called something like `AdjudicatorStatusResult`?

_Originally posted by @andrewgordstewart in https://github.com/statechannels/statechannels/pull/3127#discussion_r556107554_

lalexgap:
Fixes #3132 

This renames `ChallengeStatus` to the more accurate name of `AdjudicatorStatusResult` as the table stores on chain finalizing status info and is not limited to challenges.

The `challengeState` column has also been changed to `states` and changed to a collection of `SignedState`s as we may be storing a conclusion proof.


packages/server-wallet/src/models/adjudicator-status.ts
```diff
@@ -0,0 +1,93 @@
...
+import {Bytes32, Uint48} from '../type-aliases';
+
+export type AdjudicatorStatus =
+  | {
+      status: 'Channel Finalized';
+      finalizedAt: number;
+      states: SignedState[];
+      finalizedBlockNumber: number;
+    }
+  | {status: 'Challenge Active'; finalizesAt: number; states: SignedState[]}
```
geoknee, line 15: I'm not sure I understand why we need to store many signed states for these two channel modes. 

In these modes, we store states to enable either 
i) [`pushOutcome`](https://docs.statechannels.org/docs/protocol-tutorial/release-assets#using-pushoutcome) (and friends) or
ii) [`respond`](https://docs.statechannels.org/docs/protocol-tutorial/clear-a-challenge#call-respond)

i) only requires a single `stateHash`, and ii) only requires a single `VariablePart`. So no signatures needed, either. 

Of course there are other methods that require more: such as `checkpoint`. But these are available for the `Open` (or 'Nothing') channel mode, too. In order to `checkpoint` or `challenge`,  I would imagine that the requisite states could be pulled from the existing `channels` table ---because  it has been setup to maintain the support proof for the latest state.

packages/server-wallet/src/protocols/challenge-submitter.ts
```diff
@@ -38,10 +38,10 @@ export class ChallengeSubmitter {
...
       }
 
-      if (
-        channel.challengeStatus &&
-        channel.challengeStatus.toResult().status !== 'No Challenge Detected'
-      ) {
+      const status = channel.adjudicatorStatus
+        ? channel.adjudicatorStatus.toResult().status
+        : 'Nothing';
+      if (status !== 'Nothing') {
```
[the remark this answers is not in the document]
lalexgap: I think you're right. I think this is the correct check to do since if there is a `finalized` channel (from either a challenge or conclude) we don't want to submit a challenge. So I think the warning needs to be updated.

packages/server-wallet/src/models/adjudicator-status.ts
lalexgap: I was slightly on the fence about this. I ended up switching the single `state` to `states` to keep it in line with [event](https://github.com/statechannels/statechannels/blob/f2ef253a6bc3b9f7fa85e5ccf2bfaa73358135ad/packages/nitro-protocol/src/contract/challenge.ts#L22) NitroProtocol gives us.

My thinking was that this `AdjudicatorStatus` represents the on-chain status for a channel and I'd rather include more information than less even if it's not actively being used at the moment.

However I don't feel too strongly about it so I'd be happy to switch it back to a single `state`.

packages/server-wallet/src/db/migrations/20210112170805_rename_challenge_status.ts
```diff
@@ -0,0 +1,21 @@
+import * as Knex from 'knex';
+
+const old_table_name = 'challenge_status';
+const new_table_name = 'adjudicator_status';
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`ALTER TABLE ${old_table_name} RENAME TO ${new_table_name}`);
```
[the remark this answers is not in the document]
andrewgordstewart: It's never run "in the wild", and I'm ok with this.

packages/client-api-schema/src/data-types.ts
```diff
@@ -140,7 +140,7 @@ export interface ChannelResult {
   channelId: ChannelId;
   status: ChannelStatus;
   turnNum: Uint48;
-  challengeStatus: 'Challenge Finalized' | 'Challenge Active' | 'No Challenge Detected';
+  adjudicatorStatus: 'Channel Finalized' | 'Challenge Active' | 'Nothing';
```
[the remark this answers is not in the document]
lalexgap: In 70c8bb9 I've updated the PR to use channel mode based on the docs.

packages/server-wallet/src/models/channel.ts
```diff
@@ -264,7 +261,7 @@ export class Channel extends Model implements RequiredColumns {
       fundingStrategy,
       fundingLedgerChannelId,
       directFundingStatus: dfs,
-      challengeStatus,
+      adjudicatorStatus: this.adjudicatorStatus?.toResult().status || 'Nothing',
```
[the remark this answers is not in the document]
lalexgap: Resolved in a093a98

packages/server-wallet/src/protocols/challenge-submitter.ts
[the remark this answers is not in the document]
lalexgap: The warning has been updated in d33e8bd

packages/server-wallet/src/db/migrations/20210112170805_rename_challenge_status.ts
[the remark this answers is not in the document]
lalexgap: In 70c8bb9 I collapsed the three migrations around `ajudicatorStatus/challengeStatus` into one migration for simplicity.

packages/server-wallet/src/protocols/challenge-submitter.ts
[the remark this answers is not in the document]
geoknee: Cool. I guess there are scenarios where you might want to [submit another challenge](https://docs.statechannels.org/docs/protocol-tutorial/clear-a-challenge#call-challenge-again) when there is one ongoing -- but that is probably out of scope for this PR. Just wanted to mention, for completeness.

packages/server-wallet/src/models/adjudicator-status.ts
geoknee: Good point about the `ChallengeRegistered` event. I'm struggling to remember / understand why it doesn't just emit a single state though. The current test shows us just checking the final state in that array:

https://github.com/statechannels/statechannels/blob/f2ef253a6bc3b9f7fa85e5ccf2bfaa73358135ad/packages/nitro-protocol/test/contracts/ForceMove/challenge.test.ts#L237-L254

@snario do you have some insight about this, given https://github.com/statechannels/statechannels/pull/176 ?

packages/server-wallet/src/db/migrations/20201211180815_ajudicator_status.ts
```diff
@@ -1,11 +1,13 @@
...
 
-const tableName = 'challenge_status';
+const tableName = 'adjudicator_status';
 export async function up(knex: Knex): Promise<void> {
   await knex.schema.createTable(tableName, function (table) {
     table.string('channel_id').notNullable();
     table.integer('finalizes_at').notNullable();
     table.integer('block_number').notNullable();
+    table.integer('block_timestamp').notNullable();
+    table.specificType('states', 'jsonb[]');
```
kerzhner, line 10: Does this need to be an array of `jsonb`? I _think_ `jsonb` allows for arrays.
```suggestion
    table.specificType('states', 'jsonb');
```

packages/server-wallet/src/models/adjudicator-status.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import {Bytes32, Uint48} from '../type-aliases';
+
+export type AdjudicatorStatus =
+  | {
+      channelMode: 'Finalized';
+      states: SignedState[];
+    }
+  | {channelMode: 'Challenge'; states: SignedState[]}
+  | {channelMode: 'Open'};
```
kerzhner, line 15: Could also use
```suggestion
  | {
      channelMode: 'Finalized' | 'Challenge';
      states: SignedState[];
    }
    | {channelMode: 'Open'};
```

packages/server-wallet/src/models/adjudicator-status.ts
```diff
@@ -0,0 +1,88 @@
...
+  static async insertAdjudicatorStatus(
+    knex: Knex,
+    channelId: string,
+    finalizesAt: number,
+    states: SignedState[]
+  ): Promise<AdjudicatorStatus> {
+    const result = await AdjudicatorStatusModel.query(knex).insert({
+      channelId,
+      finalizesAt,
+      blockNumber: 0,
```
kerzhner, line 68: Curious, is there a reason we are choosing to make these columns non-nullable and defaulting to values of `0`? An alternative is to allow nulls.

packages/server-wallet/src/models/adjudicator-status.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+export class AdjudicatorStatusModel extends Model implements RequiredColumns {
+  readonly channelId!: Bytes32;
+  readonly finalizesAt!: Uint48;
+  readonly blockNumber!: Uint48;
+  readonly blockTimestamp!: Uint48;
+  readonly states!: SignedState[];
+  static tableName = 'adjudicator_status';
+  static get idColumn(): string[] {
+    return ['channelId'];
```
kerzhner, line 31: Does this need to be an array?
```suggestion
    return 'channelId';
```

kerzhner (approved):

packages/server-wallet/src/db/migrations/20201211180815_ajudicator_status.ts
lalexgap: I ran into [this issue](https://jawj.github.io/zapatos/#casting-parameters-to-json) where postgres gets confused by the array and won't automatically `JSON.stringify/parse` for us. By using `jsonb[]` we can let postgres handle the stringifying and parsing.

packages/server-wallet/src/models/adjudicator-status.ts
lalexgap: I think I did this for simplicity so that the `model` always returns a `number` when looking at the `blockNumber`\`blockTimestamp`  so we didn't have to add `model.BlockNumber ?? 0` in places.

PR #3135, opened 2021-01-13
style: upgrade prettier et al

geoknee:
We already using Typescript 4, so this catches up Prettier https://prettier.io/blog/2020/08/24/2.1.0.html#typescript , as well as other packages such as eslint and various packages. 

In particular, we would like to use short-circuiting assignments like this:

```ts
crankAgain ||= someFunction();
```

which prettier will currently complain about.

The changeset is not as yuge as it may have been, because I changed our rules to ignore the new defaults for trailing commas, bracket spacing, and parentheses for arrow functions. Still, there are some changes I can't easily prevent.


packages/client-api-schema/src/validation.ts
```diff
@@ -18,7 +18,9 @@ const apiSchema = require('./generated-schema.json'); // because https://github.
 ajv.addSchema(apiSchema, 'api.json');
 
 export const validateRequest = ajv.compile({$ref: 'api.json#/definitions/StateChannelsRequest'});
-export const validateResponse = ajv.compile({$ref: 'api.json#/definitions/StateChannelsResponse'});
+export const validateResponse = ajv.compile({
+  $ref: 'api.json#/definitions/StateChannelsResponse'
+});
```
tomclose, line 23: I'm confused by this change. Has the line length changed?

packages/server-wallet/src/config/validation.ts
```diff
@@ -74,34 +60,18 @@ const chainServiceConfigurationSchema = joi.object({
...
-    .negative()
-    .optional(),
-  allowanceMode: joi
-    .string()
-    .valid('MaxUint', 'PerDeposit')
-    .optional(),
+  provider: joi.string().uri().optional(),
+  pollingInterval: joi.number().positive().optional(),
+  blockConfirmations: joi.number().not().negative().optional(),
+  allowanceMode: joi.string().valid('MaxUint', 'PerDeposit').optional(),
```
tomclose, line 66: 🎉

tomclose (approved):

packages/client-api-schema/src/validation.ts
geoknee: The `printWidth` option has not changed, and remains at 100. This is [not a hard and fast line length limit](https://prettier.io/docs/en/options.html#print-width) in prettier, though. 

There's also this https://prettier.io/blog/2020/03/21/2.0.0.html#improved-method-chain-breaking-heuristic-6685httpsgithubcomprettierprettierpull6685-by-mmkalhttpsgithubcommmkal

which might be causing it. Overall I am not sure.

geoknee: This line is 100 chars long. So it should in fact be broken.  Not sure why it wasn't before.

PR #3138, opened 2021-01-13
Refactor logic around computing 'status' of channel result

snario:
Simplifies the logic around `status` without changing any of the behaviour.

### DIFF
--- a/packages/server-wallet/src/protocols/state.ts
+++ b/packages/server-wallet/src/protocols/state.ts
@@ -69,12 +69,12 @@ export const status = (channelState: ChannelState): ChannelStatus => {
     return 'opening';
   }
 
-  if (supported.turnNum < 2 * participants.length - 1) return 'opening';
-
   if (supported.isFinal) {
     if (_.every(support, 'isFinal')) return 'closed';
     return 'closing';
   }
+
+  if (supported.turnNum < 2 * participants.length - 1) return 'opening';
 
   return 'running';
 };
--- a/packages/server-wallet/src/protocols/state.ts
+++ b/packages/server-wallet/src/protocols/state.ts
@@ -3,7 +3,6 @@ import {
   serializeAllocation,
   checkThat,
   isAllocation,
-  State,
   Participant,
   Address,
   Destination,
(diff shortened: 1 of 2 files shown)


tomclose:
I don't understand why, for example, turnNum 2 in a 2-person channel should be considered `'running'`. Historically the channel has been considered running once the post-fund setup is complete. That means that the app can make updates to the channel if, and only if, it is running. What's the motivation behind changing this?

snario:
This PR might negate the need for this: https://github.com/statechannels/the-graph/pull/518

andrewgordstewart:
> This PR might negate the need for this: [statechannels/the-graph#518](https://github.com/statechannels/the-graph/pull/518)

Which "this" refers to this PR? 😛 

Should I interpret this as

> This PR might negate the need for [#518](https://github.com/statechannels/the-graph/pull/518)

andrewgordstewart (dismissed):
> The cache presently gets updated whenever a channel result in the output of an API call to the wallet has channel result has `status` equal to `"running"`. If a message is ever dropped and a channel has one half of a signed post-fund setup, there is currently no mechanism in the `ChannelManager` to identify this case.

You've described here why you think the [definition of "running" ](https://github.com/statechannels/statechannels/pull/3138/files#diff-6c80021eab5aa797ef4ed3a4a994440dad2accc19dea55dfbeecf6eaefb5bc87R77)should be updated in terms of the current behaviour of an external component, when it sees a "running" channel.

Why is it a *safe* definition of `running`? What are the security & usability implications?

Why is this a preferred solution over changing the behaviour in the `ChannelManager`?

It seems to me like the change is *not a good idea*. For instance,
- If we have an `[alice, bob]` channel
- they both sign state 0
- alice signs state 3

then as of this changeset, Alice would be told that the channel is "running". But if Alice tried to _use_ the channel (call `updateChannel` with an app update) the wallet throw an error.

Are you sure that this behaviour is a good idea?

snario:
No, I think a better approach would be to implement `fundingStatus` for ledger funded status, which the `ChannelManager` could then use to identify a channel which is funded but not yet finished the post-fund setup.

snario:
I've updated this PR to _only_ be the code that simplifies the logic so I can separate out the discussion about how we deal with the discussion about what "running" means.

packages/server-wallet/src/protocols/state.ts
```diff
@@ -55,62 +54,42 @@ type SignedByMe = {latestSignedByMe: SignedStateWithHash};
...
-        return (supported?.turnNum || 0) < participants.length * 2 - 1 ? 'opening' : 'running';
-      case 'Running':
-        return 'running';
-      case 'Final':
-        return support?.find(s => !s.isFinal) ? 'closing' : 'closed';
-    }
-  })();
-
-  const allocations = serializeAllocation(checkThat(outcome, isAllocation));
+  return 'running'; // unambiguously running e.g., 4, 5, 6, ...
```
andrewgordstewart, line 78: Alice might have signed state `0` only. If Bob sends Alice state 4, this line would indicate to her app that the channel is now "running", right?

snario: Yeah — let me write another variant that is simpler and handles this

snario:
Updated to something even simpler
```ts
export const status = (channelState: ChannelState): ChannelStatus => {
  const {supported, latest, latestSignedByMe, support} = channelState;
  const {participants} = supported ?? latest;

  if (!supported) {
    if (!latestSignedByMe) return 'proposed';
    return 'opening';
  }

  if (supported.turnNum < 2 * participants.length - 1) return 'opening';

  if (supported.isFinal) {
    if (_.every(support, 'isFinal')) return 'closed';
    return 'closing';
  }

  return 'running';
};
```

andrewgordstewart:
> Updated to something even simpler
> 
> ```ts
> export const status = (channelState: ChannelState): ChannelStatus => {
>   const {supported, latest, latestSignedByMe, support} = channelState;
>   const {participants} = supported ?? latest;
> 
>   if (!supported) {
>     if (!latestSignedByMe) return 'proposed';
>     return 'opening';
>   }
> 
>   if (supported.turnNum < 2 * participants.length - 1) return 'opening';
> 
>   if (supported.isFinal) {
>     if (_.every(support, 'isFinal')) return 'closed';
>     return 'closing';
>   }
> 
>   return 'running';
> };
> ```

What if Alice decided to close the channel mid-way through funding, which results in a conclusion proof at turn `2`?

snario:
> What if Alice decided to close the channel mid-way through funding, which results in a conclusion proof at turn 2?

Good point, updated 8a4ea66

packages/server-wallet/src/protocols/state.ts
```diff
@@ -55,62 +54,43 @@ type SignedByMe = {latestSignedByMe: SignedStateWithHash};
...
-    : state.isFinal
-    ? 'Final'
-    : state.turnNum <= state.participants.length - 1
-    ? 'PrefundSetup'
-    : state.turnNum <= state.participants.length * 2 - 1
-    ? 'PostfundSetup'
-    : 'Running';
+export const status = (channelState: ChannelState): ChannelStatus => {
+  const {supported, latest, latestSignedByMe, support} = channelState;
+  const {participants} = supported ?? latest;
```
andrewgordstewart, line 65: nit: this should always be the same thing, right?

```suggestion
  const {participants} = latest;
```

andrewgordstewart (approved):

PR #3144, opened 2021-01-14
refactor: use instanceof test for UniqueViolationError

geoknee:
When writing tests that use multiple `TestChannel.insertInto(store,...)` calls, I was seeing a `UniqueViolationError` despite us having code that ostensibly catches and swallows these errors. It seems that the test for catching that error is a little too specific on how the error should look. 

The solution here: 

- Increases consistency (uniform check used everywhere, rather than sometimes inspecting the name and sometimes inspecting the name and extra properties)
- Uses imported type rather than rolling our own (less cruft, higher chance to stay in sync with the library that defines the error)


lalexgap (approved):
👍

PR #3145, opened 2021-01-14
test: prefer nullish coallescing operator ?? to ||

geoknee:
I discovered this bug while trying to write a test for ledger funding, where I spin up several application channels with various nonces. Unexpected results ensued. 

As in my case, a test author may often wish to pass `channelNonce: 0` or `bBal: 0`, to the `TestChannel` constructor, but the current implementation will silently and surprisingly override those values if they do. The intention was likely more to have
defaults set when the passed values are undefined (rather than 0). Hence this change https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator.

Side remark: I'm not sure why `channelNonce` defaults to 5. I think 0 would be less magic?

### DIFF
--- a/packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
+++ b/packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
@@ -65,8 +65,8 @@ export class TestChannel {
 
   protected constructor(args: TestChannelArgs) {
     this.fundingStrategy = args.fundingStrategy || 'Direct';
-    this.startBals = [args.aBal || 5, args.bBal || 5];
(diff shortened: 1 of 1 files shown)


lalexgap (approved):

lalexgap:
What do you think about using a `eslint` rule? https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md

geoknee:
Yeah I would be in favour of that. We may get some false warnings, but better to silence those when they arise than to not have any warnings at all.

PR #3146, opened 2021-01-14
test: rewrite ledger-manager.test.ts (part 0)

geoknee:
The aim here is to preserve the current test suite, but reach farther in to the store/database in the setup and assertions of each test. This has the (beneficial?) effect of integrating more code into the tests, but also bypassing some of the implementation details of the `LedgerManager`. 

Specifically, when the work on this branch is done we shall no longer have any dependence on the `protocol()` function in the tests, nor on any `ProtocolAction`s. This means that those concepts can safely be refactored away in separate PR. 

Performing the refactor in this order means that we will maintain full test coverage throughout the refactor. This is in line with similar refactors executed/planned in the `open-channel` / `close-channel` protocols. 

Because I am moving case by case, I am hoping to get a review + merge having only converted some of the cases over. The rest can be done in another PR.

### DIFF
--- a/packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
@@ -4,6 +4,7 @@ import {
   BN,
   makeDestination,
   Participant,
+  serializeRequest,
   SimpleAllocation,
   simpleEthAllocation as coreSimpleEthAllocation,
 } from '@statechannels/wallet-core';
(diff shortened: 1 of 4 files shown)


packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
```diff
@@ -133,99 +194,193 @@ describe('exchanging ledger proposals', () => {
...
+        // for each one, insert the channel and a ledger funding request
+        await appChannel.insertInto(store, {states: [0, 1]});
+        await ledgerChannel.insertFundingRequest(store, appChannel.channelId);
+      }
+
+      // crank the ledger manager
+      const response = new WalletResponse();
+      await ledgerManager.crank(ledgerChannel.channelId, response);
+
+      // assert that there is a proposal some of those channels in the outbox
```
geoknee, line 366:
```suggestion
      // assert that there is a proposal for funding some of those channels in the outbox
```

snario (approved):
Seems like a much more reasonable approach!

packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
```diff
@@ -75,36 +86,86 @@ const processLedgerQueueProtocolState = (args: Partial<ProtocolState> = {}) => (
   ...args,
 });
 
+let store: Store;
+let ledgerManager: LedgerManager;
+
+beforeEach(async () => {
+  store = new Store(
```
andrewgordstewart, line 93: Relevant conversation over here: https://github.com/statechannels/statechannels/pull/3133#discussion_r557801985

PR #3148, opened 2021-01-14
Call register channel on wallet start up

lalexgap:
Fixes https://github.com/statechannels/project-management/issues/11

Changes the `wallet` to call `registerChannel` for existing channels on startup. Only channels that are not finalized on chain are called with `registerChannel`.

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -293,7 +293,7 @@ export class Store {
   /**
    * Returns all channels that are not finalized on chain
    */
-  async getActiveChannels(): Promise<ChannelState[]> {
+  async getNonFinalizedChannels(): Promise<ChannelState[]> {
     return (await Channel.query(this.knex))
       .filter(
         c => !c.challengeStatus || c.challengeStatus.toResult().status !== 'Challenge Finalized'
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -291,7 +291,7 @@ export class Store {
   }
 
   /**
-   * Returns all channels that are not finalized chain
+   * Returns all channels that are not finalized on chain
    */
   async getActiveChannels(): Promise<ChannelState[]> {
     return (await Channel.query(this.knex))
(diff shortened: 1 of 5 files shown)


packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -100,7 +100,27 @@ export class SingleThreadedWallet
   readonly walletConfig: ServerWalletConfig;
 
   public static create(walletConfig: IncomingServerWalletConfig): SingleThreadedWallet {
-    return new SingleThreadedWallet(walletConfig);
+    const wallet = new SingleThreadedWallet(walletConfig);
+    // This is an async method so it could continue executing after this method returns
```
lalexgap, line 104: It might make sense to change `create` to an async method so we avoid this.

However, I was reluctant to change the external wallet API as we'd need to update lots of code in this repo and in `the-graph`

packages/server-wallet/src/wallet/store.ts
```diff
@@ -290,6 +290,17 @@ export class Store {
     return (await Channel.query(this.knex)).map(channel => channel.protocolState);
   }
 
+  /**
+   * Returns all channels that are not finalized chain
```
geoknee, line 294:
```suggestion
   * Returns all channels that are not finalized on chain
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -290,6 +290,17 @@ export class Store {
...
+   */
+  async getActiveChannels(): Promise<ChannelState[]> {
```
geoknee, line 296: I would prefer we keep the terminology as small as possible, so suggest `getOpenChannels` (see https://github.com/statechannels/statechannels/pull/3134#discussion_r559580530)

geoknee (approved):

packages/server-wallet/src/wallet/store.ts
lalexgap: Unfortunately `open` may be overloaded in our `wallet` as we use open as a channel status as well. I've just updated this to `getNonFinalizedChannels` which is a bit clunky but clearer

PR #3152, opened 2021-01-15
test: default to a random channel nonce if none is provided.

geoknee:
This issue is scoped purely to _test_ code, and concerns the behaviour of the `TestChannel` class in use in many of our suites.

There are obviously several approaches here

a) make `channelNonce` random (this proposal). Nonce collisions are made very unlikely automatically.
b) make `channelNonce` a required parameter, at least things are then very explicit in our tests. The test author has to take care to avoid nonce collisions (this is laborious but very easy typically).
c) implement a method to ensure `channelNonce` increments each time a new instance is created (similar to how our store does it):

```ts
class TestChannel {
    static overallNonce = 0;
    public nonce: number;
    constructor(providedNonce?:number ) {
        this.nonce = providedNonce ?? TestChannel.overallNonce++;
    }
}
```
(although in this case, it might be better to disallow `providedNonce` altogether). This might be overengineering. The main thing is to have it not default to 5 (current behaviour).

Note: our on-chain code stipulates that `channelNonce` is a `uint48`. Our DB schema in the server wallet, the `value` column of the `nonces` table is an `integer` (32 bit). We might consider using `bigint` instead (64 bit), but this may affect performance. Here, I constrain test channel nonces to be small enough to not cause problems.

### DIFF
--- a/packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
+++ b/packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
@@ -72,7 +72,8 @@ export class TestChannel {
       ['a', args.aBal ?? 5],
       ['b', args.bBal ?? 5],
     ];
-    this.channelNonce = args.channelNonce ?? Math.ceil(Math.random() * 1e12);
+    const MAX_INTEGER_POSTGRES = 2147483647;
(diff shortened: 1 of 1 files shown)


lalexgap (approved):
I think I have a slight preference towards option c) but I agree that that might be overengineering.

As long as the chances of a collision are pretty low I'm happy with this.

PR #3153, opened 2021-01-15
test: enhance TestLedgerChannel

geoknee:
This brings the `TestLedgerChannel` closer to how real ledger channels work, so it is pretty important IMO. 

- allow a non-mover to sign a state, but default to everyone signing every state.
(passing in an optional list of indices that defaults to everyone)
- force the appDefinition to be null

### DIFF
--- a/packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
+++ b/packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
@@ -41,7 +41,7 @@ interface TestChannelArgs {
   fundingStrategy?: FundingStrategy;
 }
 
-type Bals = [string, number][] | [number, number];
+export type Bals = [string, number][] | [number, number];
 
 /** A two-party channel between Alice and Bob, with state history. For testing purposes. */
 export class TestChannel {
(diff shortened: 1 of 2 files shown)


geoknee:
The issue this has thrown up: if ledger channels are null apps (they should be), then we can't rely on turn taking in order to get a supported state. The only way is countersigning. The `TestLedgerChannel` needs to be *further* enhanced (or have more overrides) to make it easy to prepare the channel in a desired supported state during testing.

packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
```diff
@@ -98,7 +98,7 @@ export class TestChannel {
 
     return {
       ...this.channelConstants,
-      appData: '0x',
+      appData: '0x00',
```
snario, line 101: I'm confused by this in other places also — what's the right thing to be using here and what makes it right or wrong?

snario (approved):

packages/server-wallet/src/wallet/__test__/fixtures/test-channel.ts
andrewgordstewart: Is the right thing to avoid magic strings, and have either a `ZERO_APP_DATA` or a `BYTES_ZERO` constant?

geoknee: I'll try and clear this up in another PR.

PR #3154, opened 2021-01-15
fix: support fake funding in directFundingStatus

snario:
If a ledger channel is funding a payment channel, when the payment channel checks to see if its parent is funded it will not work if the parent is fake funded.

https://github.com/statechannels/statechannels/blob/f2b0d6b919ecb4eeaf2a86f5ef41c32e7ad97d43/packages/server-wallet/src/protocols/ledger-funder.ts#L66-L68

This makes this function above return `true`

### DIFF
--- a/packages/server-wallet/src/protocols/state.ts
+++ b/packages/server-wallet/src/protocols/state.ts
@@ -109,7 +109,6 @@ export function directFundingStatus(
   if (fundingStrategy === 'Fake') {
     if (!supported) return 'Uncategorized';
     if (supported.turnNum < 3) return 'ReadyToFund';
-    if (supported.isFinal) return 'Defunded';
     return 'Funded';
   }
 
--- a/packages/server-wallet/src/protocols/state.ts
+++ b/packages/server-wallet/src/protocols/state.ts
@@ -106,6 +106,13 @@ export function directFundingStatus(
   myParticipant: Participant,
   fundingStrategy: FundingStrategy
 ): FundingStatus | undefined {
+  if (fundingStrategy === 'Fake') {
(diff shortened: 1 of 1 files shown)


andrewgordstewart:
Does the payment channel need to check if the parent is funded? Can that be implicit ( the responsibility of another protocol?)

snario:
I think this is arguable.

1. When I originally wrote the code there was a TODO which asked the question you just posed (should we even do this?)
2. Tom's architecture doc implies that we ought to do this check
3. At some point the TODO was removed

I'm fine not doing it.

packages/server-wallet/src/protocols/state.ts
```diff
@@ -106,6 +106,13 @@ export function directFundingStatus(
   myParticipant: Participant,
   fundingStrategy: FundingStrategy
 ): FundingStatus | undefined {
+  if (fundingStrategy === 'Fake') {
+    if (!supported) return 'Uncategorized';
+    if (supported.turnNum < 3) return 'ReadyToFund';
+    if (supported.isFinal) return 'Defunded';
```
andrewgordstewart, line 112: Is this mean to be correct? It is incorrect, in the case when the latest supported state is final, but there is no conclusion proof.

snario: I'll just remove this branch for now — no code relies on the "Defunded" status of a closed fake funded channel at this moment.

andrewgordstewart: Now, a fake-funded channel with a conclusion proof is listed as `'Funded'`. Is that intended?

andrewgordstewart (approved):

PR #3155, opened 2021-01-18
perf: TestChannel nonce is incrementing, not random

geoknee:
A recent change #3152 caused the tests on master to fail with jest timeouts. My hypothesis is that the change introduced some inefficient code, namely generation of random integers, which made the difference for the timeouts.

This PR swaps out a random nonce for an incrementing nonce approach, which should be faster.

### DIFF
--- a/packages/server-wallet/src/handlers/__test__/single-app-updater.test.ts
+++ b/packages/server-wallet/src/handlers/__test__/single-app-updater.test.ts
@@ -9,8 +9,8 @@ import {TestLedgerChannel} from '../../wallet/__test__/fixtures/test-ledger-chan
 import {SingleAppUpdater} from '../single-app-updater';
 
 const FINAL = 8;
-const testChan = TestChannel.create({channelNonce: 1, finalFrom: FINAL});
-const testChan2 = TestChannel.create({channelNonce: 2});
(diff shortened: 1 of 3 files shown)


andrewgordstewart:
Why do you hypothesize that generating a random nonce is inefficient?

geoknee:
> Why do you hypothesize that generating a random nonce is inefficient?

Mainly i) the observed failures on master with the random nonce change included and lack thereof on this branch and ii) a little bit of googling about javascript being pretty slow at generating random integers.

On the other side, we aren't generating many random integers, so even if JS is "slow" I would not have expected it to have much of an effect. And I will admit we don't really have compelling evidence that the hypothesis is true :-P 

However, the incrementing nonce seems like a [better approach anyway ](https://github.com/statechannels/statechannels/pull/3152#pullrequestreview-569754750). And on the small amount of evidence we have, it seems empirically better, too (whether that has anything to do with the hypothesis, IDK).

andrewgordstewart:
```
❯ node __tests__/foo.js
randomly generate 1000000 numbers: 132.029ms
```

I agree that incrementing the nonce is better anyway. I would refrain from assuming that generating random nonces would cause timeouts by hogging CPU.

andrewgordstewart (approved):

PR #3158, opened 2021-01-18
fix: remove 'Fake' funding strategy

snario:
The `"Fake"` funding strategy was initially added to simplify development of the server wallet. It has lived out its initial purpose, and now must be terminated. Fake funding functionality is provided by using the `"Direct"` funding type and adding extra calls to the `Wallet`'s `updateChannelFunding` with fake data.


snario:
This is an alternative to #3154

andrewgordstewart:
Before merging this, don't we want to be certain we can implement the required changes to the `MockChainService` to
> [add] extra calls to the Wallet's updateChannelFunding with fake data.
?

snario:
Yeah. Working on this now.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -187,26 +187,6 @@ it('should fail the objective when the channel does not exist', async () => {
   expect(reloadedObjective.status).toEqual('failed');
 });
 
-it('should fail when using non-direct funding', async () => {
-  const chainService = new MockChainService();
-  const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
-
-  // Create a channel with fake funding strategy
-  const c = channel({fundingStrategy: 'Fake'});
```
lalexgap, line 195: I think we should keep this test but use ledger funding:
 ` const c = channel({fundingStrategy: 'Ledger'});`

lalexgap (dismissed):
LGTM

lalexgap:
> LGTM

Ha nevermind, didn't notice the `do-not-merge` label

snario:
Removing fake funding makes it difficult to co-ordinate multiple wallets across different processes since they would need to rely on a fake `ChainService` emitting events based on the behaviour of a counterparty which is not yet implemented. For now we'll keep Fake funding until a better alternative comes along.

PR #3169, opened 2021-01-20
refactor: throw err instead of rolling back tx

geoknee:
Simplifies the existing "rollback on error" behaviour by relying on this happening automatically at the ORM level.

https://vincit.github.io/objection.js/guide/transactions.html#using-a-transaction

It should also make it easier to write tests that want to assert on thrown errors.


tomclose:
Have you thought about the implications this will have for the wallet behaviour? It seems that currently an error processing one channel would cause just that change to rollback, and an error to be logged, but for the processing to continue. After this change, processing will abort after the first error. Is that likely to cause problems with the graph?

geoknee:
I had not considered that. I'm not sure about the answer, to be honest. Maybe @snario or @andrewgordstewart could pitch in?

If the behaviour you describe is in fact desired, I think pushing the try/catch up into the `wallet.ts` itself might be clearer 

https://github.com/statechannels/statechannels/blob/766e8e5d2385bb48098d7be74bc2ac05d13a6396/packages/server-wallet/src/wallet/wallet.ts#L673-L676
(i.e. around L674). Adding some comments would be nice, too. This proposal would make testing a bit nicer, I feel. 

We should probably have a documented strategy around error handling that can be referenced in conversations like this one
https://shapeshed.com/uncaught-exceptions-in-node/

snario:
Deferring to @andrewgordstewart.

andrewgordstewart:
Since this is a refactor, it should not change behaviour! Why can't we keep the former behaviour?

andrewgordstewart:
Also, apologies for triggering 1e95e75 . I clicked on a GH button saying "Update branch" expecting to see a dropdown menu, but it chose a merge commit to update the branch.

geoknee:
> Since this is a refactor, it should not change behaviour! 

One could just as well say: this changes the behaviour, so it should not be a refactor. At the time of authoring the PR, I did not realise it changed the behaviour. 

Since @tomclose raised the point that it does change the behaviour, I would like to discuss whether the current behaviour is intended and/or good idea.

> Why can't we keep the former behaviour?

If the current behaviour is intended, we should keep it but make changes so that we don't up here having this conversation again. That could be as simple as adding a comment. 

If the intention was simply to "rollback a running transaction in the case of an error being thrown", this PR implements that intention in a more canonical way which aids testing. 

So really I am just looking for someone with an opinion or sufficient context to say whether the current behaviour, [as explained above](https://github.com/statechannels/statechannels/pull/3169#issuecomment-763637894), is intended or not.

andrewgordstewart:
> So really I am just looking for someone with an opinion or sufficient context to say whether the current behaviour, [as explained above](https://github.com/statechannels/statechannels/pull/3169#issuecomment-763637894), is intended or not.

I do not know the historical context, and have no idea what the intended behaviour is. I also do not really know the context in which a ledger manager is "cranked", so I have no opinion on that matter.

It is easy to [restore the existing functionality](https://github.com/statechannels/statechannels/pull/3178/files?diff=split&w=1), so why change it?

geoknee:
I'm trying again here #3185

Issue #3177, opened 2021-01-21
Issue labels


andrewgordstewart:
Recorded for posterity

```
💣 BREAKING CHANGE (#e57147) - fix or feature that would cause existing functionality to not work as expected
✨ enhancement     (#a2eeef) - change which adds new functionality
🐛 bug             (#d73a4a) - non-breaking change which fixes an issue
🧹  chore          (#bdc2b6) - tech debt, improved developer experience, etc
🐎  performance    (#3033F4) - either gas efficiency or computational efficiency
📖  documentation  (#0075ca) - Improvements or additions to documentation
🔒 security        (#191708) - requires two sets of eyes before merging
```

geoknee:
@andrewgordstewart the way things are setup, we are getting a new mention for this issue on every new PR :-/

geoknee:
@andrewgordstewart the way things are setup, we are getting a new mention for this issue on every new PR :-/

andrewgordstewart:
Let's find a more appropriate place to store this info. I put it in an info in a slight rush...

wunnaii:
Documented in notion https://www.notion.so/Team-working-agreements-2a95c926bb5642e5a5c42e4b74a9dd24#b304e56734a74dfbb341b8b4b27b1c0c

PR #3179, opened 2021-01-21
chore: export SingleThreadedWallet class from pkg, disable removeComments in tsconfig

geoknee:
# Description

https://github.com/statechannels/project-management/issues/28 calls for documenation of the server wallet design, mostly from an internal-developer perspective. I consider the API of the `Wallet` class as an important part of that documentation, as well as for public-facing documentation. I think it will be very useful for "App developers" (currently these are chiefly other members of our core team) to understand the purpose and responsibility of each API call they may want to make. 

This change is preparatory work for actually documenting the API using [TSDoc](https://tsdoc.org/). Some of this documentation is in place already, but very patchy.  This change will:
- Spur on a more comprehensive set of TSDoc comments, which have the advantage of enhancing[ intellisence in VSCode](https://code.visualstudio.com/docs/nodejs/working-with-javascript). I believe this has the potential to greatly reduce confusion, duplication and regressions in our code-base, by disseminating knowledge and design principles throughout the codebase. As an example, see #3180 . I can churn out a lot more, but will wait until the overall approach has the green light.
- Cause [this page to appear](https://6009614b7f8fd5c2c94b0585--nitro-protocol.netlify.app/typescript-api/server-wallet.singlethreadedwallet) in our docs website. It's automatically generated and reasonably pretty, and very easy to "show off" to those who are unfamiliar with the codebase
- Quiet a warning from the [`api-extractor`](https://api-extractor.com/) tool that is hooked into the `prepare` lifecycle of this package. I believe this is a great tool to help us "build better typescript packages", and we should mostly do what it recommends us to do :-). The recommendation has a good [justification](https://api-extractor.com/pages/messages/ae-forgotten-export/)

## [Optional] Changes
1. I also inserted `yarn` keyword in `docs-website`. This isn't really necessary so I can rewrite history to drop that commit. 

# How Has This Been Tested?
This change is tested via CI. The docs website successfully got built: meaning that the server-wallet package was successfully compiled, had it's api successfully extracted and documented, and the result was successfully included in a draft build of our docs website (see link above). 

# Checklist:

## Code quality
- [x] I have written clear commit messages
- [x] I have performed a self-review of my own code
- [ ] I have scoped this change as narrowly as possible
- [x] I have separated logic changes from refactor changes (formatting, renames, etc.)
- [x] I have commented my code wherever necessary
- [ ] I have added tests that prove my fix is effective or that my feature works
## Project management
[...]

PR #3181, opened 2021-01-21
Change DBAdmin to a collection of static functions

lalexgap:
# Description
Removes the DBAdmin property from the wallet and exposes it instead as a collection of static methods.

It is quite common to follow the pattern of using `dbAdmin` off the `wallet` to create or migrate databases:
```
const wallet = Wallet.create(someConfig);
wallet.dbAdmin().createDB();
wallet.dbAdmin().migrateDB();
```
However, this is dangerous as the `wallet` could happen to query the database before we get a chance to create the database. With the changes in #3148 we are now running into that situation.

The solution implemented here is to remove the `DBAdmin` property from the `wallet` and expose it separately. This discourages the pattern of manipulating the database before instantiating the wallet.

I have also switched `DbAdmin` to a collection of static methods instead of an object. This removes any concerns about managing the lifetime of a `DBAdmin` object. 

For each type of operation (like createDatbase), there are two functions. 
1. One takes in a wallet config and uses the database connection for the creation/migration/etc
2. One takes in a knex instance uses that for the creation/migration/etc

## Shortcomings
- Currently, I'm exporting DbAdmin as a class with static members instead of a module of functions due to api-generator not working with something like `export * as DBAdmin from '../db-admin`
- For migrating and truncating methods that take in a config create and then destroy a `knex` instance. This is probably not too performant.

 # How Has This Been Tested?
These functions are used across our test suites so I am relying on those tests to catch any issues. No new tests have been introduced.

# Checklist:
- [ ] I have updated the graph repository to work with this change. 
## Code quality
## Project management

### DIFF
--- a/packages/server-wallet/server-wallet.api.md
+++ b/packages/server-wallet/server-wallet.api.md
@@ -78,13 +78,13 @@ export type DatabasePoolConfiguration = {
 
 // @public
 export class DBAdmin {
-    static createDatabase(config: ServerWalletConfig): Promise<void>;
+    static createDatabase(config: IncomingServerWalletConfig): Promise<void>;
     static createDatabaseFromKnex(knex: Knex): Promise<void>;
-    static dropDatabase(config: ServerWalletConfig): Promise<void>;
+    static dropDatabase(config: IncomingServerWalletConfig): Promise<void>;
     static dropDatabaseFromKnex(knex: Knex): Promise<void>;
-    static migrateDatabase(config: ServerWalletConfig): Promise<void>;
+    static migrateDatabase(config: IncomingServerWalletConfig): Promise<void>;
     static migrateDatabaseFromKnex(knex: Knex): Promise<void>;
-    static truncateDatabase(config: ServerWalletConfig, tables?: string[]): Promise<void>;
+    static truncateDatabase(config: IncomingServerWalletConfig, tables?: string[]): Promise<void>;
     static truncateDataBaseFromKnex(knex: Knex, tables?: string[]): Promise<void>;
 }
 
--- a/packages/server-wallet/server-wallet.api.md
+++ b/packages/server-wallet/server-wallet.api.md
@@ -65,12 +65,6 @@ export type ChainServiceConfiguration = {
 } & Partial<ChainServiceArgs>;
 
 // @public
-export function createDatabase(config: ServerWalletConfig): Promise<void>;
-
-// @public
-export function createDatabaseFromKnex(knex: Knex): Promise<void>;
-
-// @public
 export type DatabaseConfiguration = RequiredDatabaseConfiguration & OptionalDatabaseConfiguration;
 
 // @public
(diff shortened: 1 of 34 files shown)


packages/server-wallet/server-wallet.api.md
```diff
@@ -76,6 +76,26 @@ export type DatabasePoolConfiguration = {
     min?: number;
 };
 
+// @public (undocumented)
+export class DBAdmin {
```
lalexgap, line 80: I did add documentation to the methods. @geoknee is there anything special I need to do?

geoknee: Thanks! I wouldn't worry about this too much now, it should show up as documented when my open PR lands.

geoknee: This one https://github.com/statechannels/statechannels/pull/3179

geoknee:
> ## Shortcomings
> * Currently, I'm exporting DbAdmin as a class with static members instead of a module of functions due to api-generator not working with something like `export * as DBAdmin from '../db-admin`

Thanks for working around api-generator's quirks. FWIW I think the class approach is good anyway. 

> * For migrating and truncating methods that take in a config create and then destroy a `knex` instance. This is probably not too performant.

Are there any data on a possible performance hit? The circle run on efc1373 did seem to be unusually slow, but I don't know whether to attribute that to this set of changes or not.

lalexgap:
> Are there any data on a possible performance hit? The circle run on [efc1373](https://github.com/statechannels/statechannels/commit/efc1373e873b304c876d1c7a120e7349e00ea53a) did seem to be unusually slow, but I don't know whether to attribute that to this set of changes or not.

Most of are test favour the `migrateDatabaseFromKnex` which uses an existing `knex` instance to create/migrate/etc.. and shouldn't receive any performance hit.

I've also updated the `createDatabase` and `dropDatabase` to avoid creating and destroying a `knex` instance.

I'll take a look at comparing the latest circle run (with the changes above) against master.

lalexgap:
> > Are there any data on a possible performance hit? The circle run on [efc1373](https://github.com/statechannels/statechannels/commit/efc1373e873b304c876d1c7a120e7349e00ea53a) did seem to be unusually slow, but I don't know whether to attribute that to this set of changes or not.
> 
> Most of are test favour the `migrateDatabaseFromKnex` which uses an existing `knex` instance to create/migrate/etc.. and shouldn't receive any performance hit.
> 
> I've also updated the `createDatabase` and `dropDatabase` to avoid creating and destroying a `knex` instance.
> 
> I'll take a look at comparing the latest circle run (with the changes above) against master.

Comparing [master](https://app.circleci.com/pipelines/github/statechannels/statechannels/11492/workflows/f75a6abb-b889-4657-a547-a26db3503cbf) vs [this branch](https://app.circleci.com/pipelines/github/statechannels/statechannels/11497/workflows/eb3b57a4-9068-47e2-a4bc-08556c255a20) performance doesn't look too different.

geoknee:
I think I was probably getting confused about the circle data I mentioned, actually (forgot that although master runs a lot more tests than this branch, they run in parallel). Looking again carefully, you're right there doesn't seem to be any change. I think that is expected, too, given these changes.

packages/server-wallet/src/db-admin/db-admin.ts
```diff
@@ -15,53 +16,123 @@ import {LedgerRequest} from '../models/ledger-request';
...
+  ObjectiveModel.tableName,
+  ObjectiveChannelModel.tableName,
+  Funding.tableName,
+  AppBytecode.tableName,
+  LedgerRequest.tableName,
+  LedgerProposal.tableName,
+  Funding.tableName,
+  AdjudicatorStatusModel.tableName,
+  ChainServiceRequest.tableName,
+];
```
kerzhner, line 138: Something like this might be a bit cleaner
```suggestion
const defaultTables = [
  SigningWallet,
  Channel,
  Nonce,
  ObjectiveModel,
  ObjectiveChannelModel,
  Funding,
  AppBytecode,
  LedgerRequest,
  LedgerProposal,
  Funding,
  AdjudicatorStatusModel,
  ChainServiceRequest,
].map(model => model.tableName);
```

kerzhner (approved):

geoknee:
For posterity: 

>  This discourages the pattern of manipulating the database before instantiating the wallet.

should read

> This discourages the pattern of instantiating the wallet before preparing the database

PR #3182, opened 2021-01-21
chore: remove format on save vscode setting

kerzhner:
The below vscode settings take effect when vscode is opened in the `packages/server-wallet` directory. This is my preferred way of working on the server wallet package.

Recently, I started to experience the following behaviour:
1. Save a typescript file.
2. Observe the eslint plugin reformat the file on-save to adhere to the eslint rules.
3. Observe vscode editor reformat the file to adhere to some other rules. See eslint warnings. Note, I do *not* have the prettier plugin installed.

It seems to me that the way to avoid the behavior above is to:
1. Disable formatOnSave
2. Use [editor.codeActionsOnSave](https://github.com/microsoft/vscode-eslint) which are enabled for in my User settings. Perhaps this should be enabled in the Workspace settings as well?

### DIFF
--- a/packages/server-wallet/.vscode/settings.json
+++ b/packages/server-wallet/.vscode/settings.json
@@ -4,5 +4,6 @@
     "javascriptreact",
     "typescript",
     "typescriptreact"
-  ]
+  ],
+  "editor.codeActionsOnSave": {"source.fixAll.eslint": true}
 }
--- a/packages/server-wallet/.vscode/settings.json
+++ b/packages/server-wallet/.vscode/settings.json
@@ -1,4 +1,8 @@
 {
-  "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
-  "editor.formatOnSave": true
+  "eslint.validate": [
+    "javascript",
+    "javascriptreact",
+    "typescript",
+    "typescriptreact"
+  ]
 }


geoknee:
It might be good to try and get someone else to reproduce your issues @kerzhner. 

Do you have an idea of what the "other rules" are, that vscode seems to want to enforce?

Tinkering with this locally, I noticed that my vscode was set up to have my prettier extension as the default formatter. I turned that off but didn't see the problems that you describe (I think it was falling back to eslint as specified in the workspace settings. In principle they should have the same rules, and my experience seems to confirm that)

Is there a chance you have some other local setting that overrides the workspace settings?

kerzhner:
@geoknee you bring up an excellent point. What is clear is that:
- Different folks might have different User settings (as opposed to Workspace settings).
- The Workspace settings in this PR are maybe **not** sufficient to configure eslint.

That being said, [it looks like](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) "The old eslint.autoFixOnSave setting is now deprecated and can safely be removed."

geoknee:
For the avoidance of doubt, when I have stated "it works fine for me" I mean that vscode formats my `.ts` files (and my `.json` files and my `.sol` files) on save, in a way that eslint never complains about. 

If I get a purely linting-related warning (e.g, unused import), vscode does not fix that on save (and I'm not sure I would want it to, either). I tend to hover and right-click to make those fixes, or use `yarn lint:write`. 

My user settings are 
```json
{
  "git.enableSmartCommit": true,
  "workbench.colorTheme": "Default Dark+",
  "window.zoomLevel": 1,
  "eslint.format.enable": true,
  "eslint.alwaysShowStatus": true,
  "eslint.packageManager": "yarn",
  "git.confirmSync": false,
  "diffEditor.ignoreTrimWhitespace": false,
  "diffEditor.renderSideBySide": true,
  "typescript.preferences.importModuleSpecifier": "relative",
  "editor.formatOnSave": true,
  "solidity.compileUsingRemoteVersion": "v0.7.4",
  "solidity.validationDelay": 100,
  "editor.accessibilitySupport": "off",
  "editor.defaultFormatter": "esbenp.prettier-vscode"
}
```

kerzhner:
Maybe the User setting configuration that is causing havoc for me is:  
```
"[typescript]": {
    "editor.defaultFormatter": "vscode.typescript-language-features"
  }
```

kerzhner:
I resolved my specific issue by removing the typescript default formatter in my user settings. Closing this PR in favor of https://github.com/statechannels/statechannels/issues/3186. That issue should probably be addressed before new developers are onboarded.

PR #3183, opened 2021-01-21
Wallet.Create is now async

lalexgap:
# Description
Updates the `Wallet.create` method to be `async`.

In #3148 an async function  `registerExistingChannelsWithChainService` is called in the constructor. Since it is async it can fail after the constructor has finished executing, causing issue with the tests.

The solution implemented is to change the `Wallet.Create` method to be async. This allows us to perform the registration or any other initialization logic before returning the `wallet`.  

#  How Has This Been Tested?
 These methods are using throughout our tests so I am relying on existing test coverage.

# Checklist:
- [ ] Update graph repo to work with changes.
## Code quality
## Project management

### DIFF
--- a/packages/server-wallet/e2e-test/payer/client.ts
+++ b/packages/server-wallet/e2e-test/payer/client.ts
@@ -12,17 +12,26 @@ import {defaultConfig, ServerWalletConfig} from '../../src/config';
 import {ONE_DAY} from '../../src/__test__/test-helpers';
 
 export default class PayerClient {
-  private readonly wallet: ServerWallet;
-  constructor(
+  readonly config: ServerWalletConfig;
+  private constructor(
     private readonly pk: Bytes32,
     private readonly receiverHttpServerURL: string,
-    readonly config?: ServerWalletConfig
+    private readonly wallet: ServerWallet
   ) {
-    this.wallet = recordFunctionMetrics(
-      ServerWallet.create(this.config || payerConfig),
+    this.config = wallet.walletConfig;
+  }
+  public static async create(
+    pk: Bytes32,
+    receiverHttpServerURL: string,
+    config?: ServerWalletConfig
+  ): Promise<PayerClient> {
+    const wallet = await recordFunctionMetrics(
+      ServerWallet.create(config ?? payerConfig),
       payerConfig.metricsConfiguration.timingMetrics
     );
+    return new PayerClient(pk, receiverHttpServerURL, wallet);
   }
+
   public async warmup(): Promise<void> {
     await this.wallet.warmUpThreads();
   }
--- a/packages/server-wallet/e2e-test/payer/client.ts
+++ b/packages/server-wallet/e2e-test/payer/client.ts
@@ -25,8 +25,8 @@ export default class PayerClient {
     receiverHttpServerURL: string,
     config?: ServerWalletConfig
   ): Promise<PayerClient> {
-    const wallet = await recordFunctionMetrics(
-      ServerWallet.create(config ?? payerConfig),
+    const wallet = recordFunctionMetrics(
+      await ServerWallet.create(config ?? payerConfig),
       payerConfig.metricsConfiguration.timingMetrics
     );
     return new PayerClient(pk, receiverHttpServerURL, wallet);
--- a/packages/server-wallet/e2e-test/payer/start.ts
+++ b/packages/server-wallet/e2e-test/payer/start.ts
@@ -40,8 +40,8 @@ export default {
   handler: async (argv: {[key: string]: any} & Argv['argv']): Promise<void> => {
     const {database, numPayments, channels} = argv;
 
-    const payerClient = await recordFunctionMetrics(
-      PayerClient.create(
+    const payerClient = recordFunctionMetrics(
+      await PayerClient.create(
         alice().privateKey,
         `http://127.0.0.1:65535`,
         overwriteConfigWithDatabaseConnection(defaultTestConfig(), {database})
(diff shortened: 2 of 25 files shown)


packages/server-wallet/e2e-test/payer/client.ts
```diff
@@ -12,17 +12,26 @@ import {defaultConfig, ServerWalletConfig} from '../../src/config';
...
-      ServerWallet.create(this.config || payerConfig),
+    this.config = wallet.walletConfig;
+  }
+  public static async create(
+    pk: Bytes32,
+    receiverHttpServerURL: string,
+    config?: ServerWalletConfig
+  ): Promise<PayerClient> {
+    const wallet = await recordFunctionMetrics(
+      ServerWallet.create(config ?? payerConfig),
```
kerzhner, line 29: Is the `await` in the correct place here? Should we await `ServerWallet.create` instead of `recordFunctionMetrics`?

kerzhner: Also, does `recordFunctionMetrics` work with promises?

packages/server-wallet/e2e-test/payer/start.ts
```diff
@@ -40,8 +40,8 @@ export default {
   handler: async (argv: {[key: string]: any} & Argv['argv']): Promise<void> => {
     const {database, numPayments, channels} = argv;
 
-    const payerClient = recordFunctionMetrics(
-      new PayerClient(
+    const payerClient = await recordFunctionMetrics(
```
kerzhner, line 43: Same comment as above.

packages/server-wallet/e2e-test/receiver/controller.ts
```diff
@@ -9,10 +9,15 @@ import {defaultConfig} from '../../src/config';
 import {WALLET_VERSION} from '../../src/version';
 
 export default class ReceiverController {
-  private readonly wallet: Wallet = recordFunctionMetrics(
-    Wallet.create(receiverConfig),
-    defaultConfig.metricsConfiguration.timingMetrics
-  );
+  static async create(): Promise<ReceiverController> {
+    const wallet = await recordFunctionMetrics(
```
kerzhner, line 13: Same comment as above.

packages/server-wallet/src/wallet/multi-threaded-wallet/worker.ts
```diff
@@ -22,45 +22,46 @@ const walletConfig: ServerWalletConfig = {
...
 
 logger.debug(`Worker %o starting`, threadId);
-const wallet = Wallet.create(walletConfig);
-parentPort?.on('message', async (message: any) => {
-  if (isMainThread) {
-    parentPort?.postMessage(
-      left(new Error('Attempting to execute worker thread script on the main thread'))
-    );
-  }
+Wallet.create(walletConfig).then(wallet => {
```
kerzhner, line 25: Another option is to factor out the bulk of this file into an `async` function and then call the function (without awaiting).

packages/server-wallet/e2e-test/payer/client.ts
lalexgap: > Is the `await` in the correct place here? Should we await `ServerWallet.create` instead of `recordFunctionMetrics`?

I think you're right. I wasn't paying too much attention and while `await recordFunctionMetrics` would still return the `wallet` I don't think metrics were being recorded for the wallet members.

Fixed in 62871e5

packages/server-wallet/e2e-test/payer/start.ts
lalexgap: Fixed in 62871e5

packages/server-wallet/e2e-test/receiver/controller.ts
lalexgap: Fixed in 62871e5

packages/server-wallet/e2e-test/payer/client.ts
kerzhner: It looks like `recordFunctionMetrics` uses node `timerify` under the hood. Are we sure that `timerify` can accept promises? If not `recordFunctionMetrics` might not be measuring anything useful here.

lalexgap: > It looks like `recordFunctionMetrics` uses node `timerify` under the hood. Are we sure that `timerify` can accept promises? If not `recordFunctionMetrics` might not be measuring anything useful here.

I'm not sure if it handles promises but with the changes in 62871e5 we're now passing in a wallet object instead of a promise so I don't think it should matter?

packages/server-wallet/src/wallet/multi-threaded-wallet/worker.ts
lalexgap: Done in 60b5400

packages/server-wallet/e2e-test/payer/client.ts
kerzhner: Gotcha! I wonder what we are timing in that case then...

lalexgap: > Gotcha! I wonder what we are timing in that case then...

I think the idea is that [it iterates through each function on the wallet and wraps it in a `timerify`](https://github.com/statechannels/statechannels/blob/08add256bf70bffd5551882a6b019938f8670a18/packages/server-wallet/src/metrics.ts#L99). That way we should get metrics on all the `wallet` functions. Not sure if it still works or provides useful information right now though

kerzhner: Ah gotcha. That makes sense, thank you for the explanation!

andrewgordstewart: This might be an appropriate link here: https://mtlynch.io/code-review-love/#4-answer-questions-with-the-code-itself

(though I appreciate that the resulting conversation is somewhat off topic from the scope of the PR)

kerzhner (approved):

Issue #3176, opened 2021-01-20
directly-funded-channel chain test fails


kerzhner:
directly-funded-channel test is currently suspect due to https://github.com/statechannels/statechannels/issues/3174.

My approach has been to apply the following diff to get around the above:
```
diff --git a/packages/server-wallet/src/wallet/wallet.ts b/packages/server-wallet/src/wallet/wallet.ts
index 6af6b9077..8e490d4fe 100644
--- a/packages/server-wallet/src/wallet/wallet.ts
+++ b/packages/server-wallet/src/wallet/wallet.ts
@@ -103,7 +102,7 @@ export class SingleThreadedWallet
   public static create(walletConfig: IncomingServerWalletConfig): SingleThreadedWallet {
     const wallet = new SingleThreadedWallet(walletConfig);
     // This is an async method so it could continue executing after this method returns
-    wallet.registerExistingChannelsWithChainService();
+    // wallet.registerExistingChannelsWithChainService();
     return wallet;
   }
```

After the diff is applied, the test fails locally with:
```
 FAIL  src/__chain-test__/directly-funded-channel.test.ts (16.822 s)
  ✕ Create a directly funded channel between two wallets  (10597 ms)

  ● Create a directly funded channel between two wallets 

    expect(received).toMatchObject(expected)

    - Expected  - 1
    + Received  + 1

      Object {
    -   "fundingStatus": "Defunded",
    +   "fundingStatus": "Funded",
        "status": "closed",
        "turnNum": 4,
      }

      202 |   expect(getChannelResultFor(channelId, resultB1.channelResults)).toMatchObject({
      203 |     status: 'running',
    > 204 |     turnNum: 3,
```

PR #3187, opened 2021-01-22
Fix race condition in Funding table writes

kerzhner:
Use `lockApp` to avoid stale writes to the Funding table.

The underlying issue here is a read and a stale write. We use read-committed Postgres isolation level. It is possible that with two simultaneous `updateTransferredOut` calls A and B the following happens:
1. A reads row R.
1. B reads row R.
1. A updates R to f(R)
1. B updates R to f’(R). B should update R to f’(f(R)).

We have a few different solutions:
1. A more strict transaction isolation level.
1. select-for-update on a Funding table row at the start of the function.
1. select-for-update on a Channels table row at the start of the function.

It seems prudent to use the existing locking mechanism (option 3) to avoid having to think about deadlocks. If we introduce a table specific lock for the Funding table, we need to think about order of lock acquisition.

Fixes https://github.com/statechannels/statechannels/issues/3176.

## Code quality
I have NOT scoped this change as narrowly as possible as this change involves a one line commit to revert a change to a test that was not nessary.


andrewgordstewart:
It's not a big deal for this one, but you have to manually link the issue to the PR using zenhub. (I know it's a pain)
Screen Shot 2021-01-22 at 8 27 55 PM

This action causes then to merge into a single card on the zenhub board:

Screen Shot 2021-01-22 at 8 29 59 PM

andrewgordstewart (approved):

PR #3191, opened 2021-01-25
ci: inject date into canary preid

### LINKED ISSUE #3190: Anticipate errors when publishing identical canary builds on a schedule
This is very likely to happen on weekends. NPM complains that we are trying to publish over an existing version. 

https://github.com/statechannels/statechannels/runs/1756460631?check_suite_focus=true

One solution would be to alter the build version to include a date as well as a githash.

geoknee:
Prevents the github action from failing when there have been no commits made between runs (avoids tag collisions between nightly builds, even when they run on the same commit). 

Fixes #3190 

Shortcomings: long version numbers getting longer. Does nothing to address existing shortcoming (generation and release of at least 1 new version per day / hundreds per year). 

Alternatives considered: 
- reduce frequency of this github action to weekly (very likely a commit has been made; side effect of reducing number of published versions; we only learn about breaking changes relatively slowly)
- ignore the failed github actions (nothing bad really happened)
- "catch" the error and report success from the GH action.

### [Optional] How Has This Been Tested?
By manually running `echo next.$(git rev-parse --short HEAD).$(date +%Y-%m-%d)`. 

In terms of version numbers, this means that: 
`1.14.2-next.2522b345b` will become `1.14.2-next.2522b345b.2021-01-25`

## Code quality
- [x] I have added tests that prove my fix is effective or that my feature works
## Project management

### DIFF
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -32,4 +32,4 @@ jobs:
           git config --global user.email "github-actions@users.noreply.github.com"
 
       - name: publish packages to npm
-        run: yarn run lerna publish --canary --force-publish --exact --preid next.$(git rev-parse --short HEAD) --dist-tag next --yes
+        run: yarn run lerna publish --canary --force-publish --exact --preid next.$(git rev-parse --short HEAD).$(date +%Y-%m-%d) --dist-tag next --yes


andrewgordstewart:
> "catch" the error and report success from the GH action.

Why is this, or some variation, not preferred?

If there's no difference between the current HEAD of master and yesterday's HEAD of master, doing nothing seems like the straightforward thing to do.

geoknee:
I agree that is a better solution. I think it would take me longer to engineer, that's why I started with this option.

andrewgordstewart:
then I'm happy with this!

andrewgordstewart (approved):

geoknee:
Taking aim at the better solution #3221

PR #3193, opened 2021-01-25
refactor: rip out getFirstParticipant() from store

geoknee:
AFAICT this code is unreferenced, and somewhat confusing to me. It seems like a good opportunity to thin out the wallet API, but I want to invite feedback in case someone knows of a good reason to keep it.


packages/server-wallet/src/wallet/store.ts
```diff
@@ -95,15 +95,6 @@ export class Store {
     await this.getOrCreateSigningAddress(new ethers.Wallet(privateKey));
   }
 
-  async getFirstParticipant(): Promise<Participant> {
-    const signingAddress = await this.getOrCreateSigningAddress();
```
geoknee, line 99: I am confused by this, because the "first" participant won't necessarily have *my* signing address.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -95,15 +95,6 @@ export class Store {
...
-    return {
-      participantId: signingAddress,
-      signingAddress,
-      destination: makeDestination(signingAddress),
```
geoknee, line 103: I think in general these are the wrong values for these fields.

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: "first" doesn't mean first in the channel. It means first in the wallet.

Since participant management is not actually performed by the wallet, the wallet enforces that it only stores one signing key. Therefore, the "first" participant here is the only participant the wallet stores.

Moreover, as a `Participant` object, it's misleading. The participantId does not match the Id in channels that the receipt manager will join, for example: in that case, the participantId in the channel is a URL, while the Id returned by the wallet is a hex string.

andrewgordstewart (commented):
Is it used by the payment manager?

geoknee:
> Is it used by the payment manager?
No.

andrewgordstewart (approved):

PR #3194, opened 2021-01-25
chore: mergeMessages -> mergeOutputs

geoknee:
...and make into a static method. It's just a wrapper for
a static mehod for another class.

This function is in use in just one place in one place in the graph presently:

https://github.com/statechannels/the-graph/blob/ec3007419ca2747f4ae444576d22ebb7842e4b3d/packages/receipt-manager/src/receipt-manager.ts#L100

But it should be an easy fix `this.wallet.mergeMessages(...) => Wallet.mergeOutputs(...)`


andrewgordstewart (approved):

andrewgordstewart:
nit: this doesn't follow
> I have applied the [appropriate labels](https://github.com/statechannels/statechannels/issues/3177)

from the PR template. (It includes a breaking change)

PR #3198, opened 2021-01-25
fix: await concludeAndWithdraw transaction submission

kerzhner:
Chain service APIs that result in transaction submission return a promise that resolves to [`ethers` `TransactionResponse`](https://docs.ethers.io/v5/single-page/#/v5/api/providers/types/-%23-providers-TransactionResponse). The heuristic thus far has been for the chain service API consumer to await these promises to ensure that a transaction is submitted properly.

Note that this does **not** await transaction mining.

### DIFF
--- a/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
@@ -10,7 +10,7 @@ import {testKnex as knex} from '../../../jest/knex-setup-teardown';
 import {seedAlicesSigningWallet} from '../../db/seeds/1_signing_wallet_seeds';
 import {channel} from '../../models/__test__/fixtures/channel';
 import {Channel} from '../../models/channel';
-import {MockChainService} from '../../chain-service';
+import {ErorringMockChainService, MockChainService} from '../../chain-service';
 import {WalletResponse} from '../../wallet/wallet-response';
 import {ChannelDefunder} from '../defund-channel';
 import {AdjudicatorStatusModel} from '../../models/adjudicator-status';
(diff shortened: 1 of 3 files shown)


lalexgap (approved):

andrewgordstewart:
Does this PR need to address
> I have added tests that prove my fix is effective or that my feature works

from the PR template?

kerzhner:
@andrewgordstewart you bring up a good point. Thus far we have not tested how our protocols behave with unexpected Chain Service errors. I added a testcase to this PR as a start.

Also, https://github.com/statechannels/statechannels/issues/3201 has been filed to add more testcases like this.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -31,7 +31,6 @@ beforeEach(async () => {
     '0'
   );
 
-  await DBAdmin.truncateDataBaseFromKnex(knex);
```
andrewgordstewart, line 34: > So if the database is truncated before all testcases and after each testcase,
we should not need to truncate the database beforeEach testcase.

I have a slight preference truncating in `beforeEach` instead of `afterEach`, since this allows a developer to inspect the DB upon test completion.

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -213,7 +243,7 @@ it('should fail when using non-direct funding', async () => {
 });
 
 // Helpers
-async function setChallengeStatus(
+async function setAdjudicatorStatus(
```
andrewgordstewart, line 246: nit: This violates
> I have separated logic changes from refactor changes (formatting, renames, etc.)

from our PR template.

andrewgordstewart (approved):

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
kerzhner: Good call baf7002ad

PR #3209, opened 2021-01-26
build: generate api report during prepare

geoknee:
The api report checked in for the client-api-schema package is out of sync with the source code. This PR hooks report generation into the prepare lifecycle, as it is for the server-wallet. If the API of that package changes, Circle will now fail unless the report is regenerated and checked in. 

Additionally, if running the docs-website locally, the correct report _is_ generated, which dirties your working directory if the report is out of sync. This change will prevent that issue. 

Shortcomings: `prepare` will take a little longer. PRs may unexpectedly fail, it may not be clear why or how to fix. 

# [Optional] :warn: Does this require multiple approvals?
- [x] Is it a significant process change? Arguably, yes. Although it is really an aligning of several processes already in play. 

# Checklist:

## Code quality
## Project management
- [ ] I have linked to relevant issues

### DIFF
--- a/packages/client-api-schema/client-api-schema.api.md
+++ b/packages/client-api-schema/client-api-schema.api.md
@@ -110,13 +110,13 @@ export type ChannelProposedNotification = JsonRpcNotification<'ChannelProposed',
 // @public (undocumented)
 export interface ChannelResult {
     // (undocumented)
+    adjudicatorStatus: 'Challenge' | 'Open' | 'Finalized';
+    // (undocumented)
     allocations: Allocation[];
     // (undocumented)
     appData: string;
     // (undocumented)
     appDefinition: Address;
-    // (undocumented)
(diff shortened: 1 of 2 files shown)


andrewgordstewart:
What's the argument for including this in `prepare`, which as I understand, is meant to make a package consumable by other, local packages?

`prepare` is a blocker for all remaining CI jobs, so lengthening it has a compounding effect on productivity. IMO, we ought to _shorten_ `prepare` as aggressively as possible.

geoknee:
Hmm that's a fair point. I think the implicit argument in my head is that it enables us to enforce that this report is up to date before merging a PR, using existing pipeline. 

I think we can have the best of both worlds. I'll close this for now and try again with something that won't lengthen `prepare`.

geoknee:
See #3220

PR #3210, opened 2021-01-26
chore: remove no-op method from SingleThreadedWallet

geoknee:
This method is a no-op, and it unnecessarily bloats the wallet API. This change restores a more usual approach to class inheritance.

Consumers will now need to be sure they have a `MultiThreadedWallet` before calling `warmUpThreads`. This is easy and this PR makes such a change by way of example. 

## Code quality
## Project management
- [ ] I have linked to relevant issues


andrewgordstewart (approved):
Given that this is a breaking change, it ought to force some type of version change.

Given that we don't want to bump major versions (right?), it shouldn't include `breaking` as a conventional keyword, but it probably deserves a `fix` or a `feat`, right?

This can be accomplished via "squash & merge"

PR #3211, opened 2021-01-26
Improve PR template

geoknee:
We resolved to try out the new templates and iterate on them as we go along. Here's a new iterate!

- the current template looks a bit ugly when it gets turned into html
- some of the prompts are a bit ambiguous or confusing (to me at least)

## Code quality
- [ ] I have scoped this change as narrowly as possible
- [x] I have added tests that prove my fix is effective or that my feature works
## Project management

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -36,7 +36,7 @@ Please explain which reason, if any, why this requires more than one approval.
 ### Code quality
 - [ ] I have written clear commit messages
 - [ ] I have performed a self-review of my own code
-- [ ] I have scoped this change as narrowly as possible
+- [ ] This change does not have an unduly wide scope
 - [ ] I have separated logic changes from refactor changes (formatting, renames, etc.)
 - [ ] I have commented my code wherever necessary (can be 0)
 - [ ] I have added tests that prove my fix is effective or that my feature works
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -39,7 +39,7 @@ Please explain which reason, if any, why this requires more than one approval.
 - [ ] This change does not have an unduly wide scope
 - [ ] I have separated logic changes from refactor changes (formatting, renames, etc.)
 - [ ] I have commented my code wherever necessary (can be 0)
-- [ ] I have added tests that prove my fix is effective or that my feature works
+- [ ] I have added tests that prove my fix is effective or that my feature works, if necessary
 ### Project management
 - [ ] I have applied the [appropriate labels](https://www.notion.so/Team-working-agreements-2a95c926bb5642e5a5c42e4b74a9dd24#b304e56734a74dfbb341b8b4b27b1c0c)
 - [ ] I have linked to all relevant issues (can be 0)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,7 +15,7 @@ It may include
 The summary may omit some of these details if this PR fixes a single ticket that includes these details. It is the reviewer's discretion. 
 
 ## Changes [Optional] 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -13,34 +14,35 @@ It may include
...
 
-# Checklist:
+---
+## Checklist:
 
-## Code quality
+### Code quality
 - [ ] I have written clear commit messages
 - [ ] I have performed a self-review of my own code
-- [ ] I have scoped this change as narrowly as possible
```
geoknee, line 37: Taking this PR as a (slightly contrived) example: I could have put each atomic change in its own PR: but that seems like overkill. I think the author can mostly just use their judgement to manage the scope of a PR.

I also think that, rather than trying to pre-empt the desires of the reviewer, we can allow the reviewer to ask the originator to split a PR if they deem the scope to be too wide. This point goes for some of the other prompts in the template, too: a reviewer can and should ask for more context if they need it, for example.

.github/pull_request_template.md
```diff
@@ -13,34 +14,35 @@ It may include
...
-- [ ] I have commented my code wherever necessary
-- [ ] I have added tests that prove my fix is effective or that my feature works
-## Project management
+- [ ] I have commented my code wherever necessary (can be 0)
+- [ ] I have added tests that prove my fix is effective or that my feature works, if necessary
+### Project management
 - [ ] I have applied the [appropriate labels](https://www.notion.so/Team-working-agreements-2a95c926bb5642e5a5c42e4b74a9dd24#b304e56734a74dfbb341b8b4b27b1c0c)
-- [ ] I have linked to relevant issues
-- [ ] I have added dependent tickets
+- [ ] I have linked to all relevant issues (can be 0)
```
andrewgordstewart, line 45: 💡 I think we could leave a blanket statement adding "where necessary" to each item in the checklist, leaving it up to the author's discretion of when it's necessary.

.github/pull_request_template.md
```diff
@@ -13,34 +14,35 @@ It may include
...
+- [ ] I have added all dependent tickets (can be 0)
 - [ ] I have assigned myself to this PR
-- [ ] I have chosen the appropriate pipeline
+- [ ] I have associated this PR to a GitHub Project
```
andrewgordstewart, line 48: [required]: I don't understand this. We aren't using a Github Project at the moment. What's your intended meaning here? Can you explain the action with a screenshot?

"pipeline" refers to the dropdown in the top right:

Screen Shot 2021-01-26 at 9 04 06 AM

andrewgordstewart (changes requested):
I think there's one mistake, but otherwise LGTM

.github/pull_request_template.md
geoknee, line 48: I don't have the zenhub plugin for chrome installed, so I don't see that dropdown. I made an incorrect guess that a pipeline was mapped onto a GitHub project. 

I think it is true to say that the card should be added to the zenhub board? I think the zenhub board does map on to a github project? This is what I do:
Screenshot 2021-01-26 at 18 16 32

Perhaps we want to have both checkboxes in place, and an explainer that the pipelines is a zenhub concept? I think a new de would have a very hard time figuring that out.

andrewgordstewart: I am not aware of an integration between zenhub boards and github projects. To test whether there currently exists an integration, I placed this "in progress" in the "graph integration" github project. The pipeline where the card lives is still "Review/QA". So, I believe there **is no integration** at the moment.

My preferred solution here is for there to be a document (eg. notion) explaining this process. The template may link to the document. The checklist serves as a reminder to go through the documented steps.

geoknee: ok, looks like you are right about the lack of integration
```suggestion
- [ ] I have chosen the appropriate pipeline on zenhub
```

andrewgordstewart (approved):

PR #3219, opened 2021-01-27
Store and challenge with initial supported state for ledger channels

lalexgap:
# Description
Store and challenges with the initial supported state for ledger channels.

Previously the challenge method on the wallet accepted a single unsigned state. However, that is not enough for a valid challenge proof, you need a supported state.

The solution implemented here is to:
1.  store the initial supported state for a ledger channel in the wallet.
2. When `challenge` is called used the stored initial support to create the challenge transaction.

The ledger manager protocol is now responsible for handling 1.  I've updated the `MarkLedgerFundingRequestsAsComplete` action handler to store the initial support. Since this is short-term behavior I thought that was a reasonable place and I wanted to limit the scope of the changes.

I considered storing the `initialSupport` somewhere lower in the stack like on the `store.addState`  but figured since this is temporary behavior it makes sense to avoid changing "core" functionality.

The initial support is stored in the database as an additional column so we can easily get rid of it when it's no longer needed. 

### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -156,7 +156,7 @@ export class Channel extends Model implements RequiredColumns {
       .withGraphFetched('adjudicatorStatus')
       .first();
   }
-
+  // CHALLENGING_V0 temporary method
   static async setInitialSupport(
     channelId: string,
     support: SignedState[],
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -8,6 +8,7 @@ import {
   Zero,
   Address,
   toNitroState,
+  SignedState,
 } from '@statechannels/wallet-core';
 import {JSONSchema, Model, Pojo, QueryContext, ModelOptions, TransactionOrKnex} from 'objection';
 import {ChannelResult, FundingStrategy} from '@statechannels/client-api-schema';
@@ -84,6 +85,8 @@ export class Channel extends Model implements RequiredColumns {
   readonly assetHolderAddress!: string; // only Ledger channels have this
   readonly fundingLedgerChannelId!: Bytes32; // only App channels funded by Ledger have this
 
+  readonly initialSupport!: SignedState[];
+
   static get jsonSchema(): JSONSchema {
     return {
       type: 'object',
@@ -142,7 +145,7 @@ export class Channel extends Model implements RequiredColumns {
     },
   };
 
-  static jsonAttributes = ['vars', 'participants'];
+  static jsonAttributes = ['vars', 'participants', 'initialSupport'];
 
   static async forId(channelId: Bytes32, txOrKnex: TransactionOrKnex): Promise<Channel> {
     return Channel.query(txOrKnex)
@@ -152,6 +155,14 @@ export class Channel extends Model implements RequiredColumns {
       .withGraphFetched('chainServiceRequests')
       .withGraphFetched('adjudicatorStatus')
       .first();
+  }
+
+  static async setInitialSupport(
+    channelId: string,
+    support: SignedState[],
+    txOrKnex: TransactionOrKnex
+  ): Promise<void> {
+    await Channel.query(txOrKnex).findOne({channelId}).patch({initialSupport: support});
   }
 
   static async setLedger(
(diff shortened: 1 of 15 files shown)


packages/server-wallet/src/wallet/__test__/challenge.test.ts
```diff
@@ -20,12 +20,28 @@ beforeEach(async () => {
...
+    vars: [stateWithHashSignedBy([alice(), bob()])({turnNum: 1})],
+    initialSupport: [stateWithHashSignedBy([alice(), bob()])({turnNum: 1})],
+  });
+  await Channel.query(w.knex).insert(c);
 
+  const {channelId} = c;
+
+  expect(c.isLedger).toBe(false);
+
+  await expect(w.challenge(channelId)).rejects.toThrow('Only ledger channels support challenging');
```
andrewgordstewart, line 35: nit: 
```suggestion
  expect(c.isLedger).toBe(false);

  await expect(w.challenge(c.channelId)).rejects.toThrow('Only ledger channels support challenging');
```

packages/server-wallet/src/wallet/__test__/challenge.test.ts
```diff
@@ -34,12 +50,9 @@ it('submits a challenge when no challenge exists for a channel', async () => {
   const current = await AdjudicatorStatusModel.getAdjudicatorStatus(w.knex, channelId);
 
   expect(current.channelMode).toEqual('Open');
-  const challengeState = {
-    ...c.channelConstants,
-    ..._.pick(c.latest, ['turnNum', 'outcome', 'appData', 'isFinal']),
-  };
-  await w.challenge(challengeState);
-  expect(spy).toHaveBeenCalledWith([expect.objectContaining(challengeState)], alice().privateKey);
+
```
andrewgordstewart, line 53: nit: can you
```suggestion
    expect(c.isLedger).toBe(true)
```
for reader clarity?

packages/server-wallet/src/models/channel.ts
```diff
@@ -154,6 +157,14 @@ export class Channel extends Model implements RequiredColumns {
       .first();
   }
 
+  static async setInitialSupport(
```
andrewgordstewart, line 160: Can you add the `CHALLENGING_V0` tag here, so we remember to remove it along with the below code?

I think that would be sufficient, since removing this would trigger a cascade of compile errors that leads to the remaining cleanup code.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -269,21 +267,23 @@ export class SingleThreadedWallet
...
-    const channelId = calculateChannelId(challengeState);
+  async challenge(channelId: string): Promise<SingleChannelOutput> {
     const response = WalletResponse.initialize();
 
     await this.knex.transaction(async tx => {
       const channel = await this.store.getChannel(channelId, tx);
       if (!channel) {
         throw new Error(`No channel found for channel id ${channelId}`);
       }
+      if (!channel.isLedger) {
```
andrewgordstewart, line 278: Maybe this also needs a `CHALLENGING_V0` comment?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -142,7 +140,7 @@ export class SingleThreadedWallet
 
     if (!valid) {
       errors.forEach(error =>
-        this.logger.error({error}, `Validation error occured ${error.message}`)
+        this.logger.error({error}, `Validation error occurred ${error.message}`)
```
andrewgordstewart, line 143: I think `err` is the magic keyword for pino, which triggers to use [the `pino-std-serializers/err` serializer](https://github.com/pinojs/pino-std-serializers#exportserrerror)

```suggestion
        this.logger.error({err: error}, `Validation error occurred ${error.message}`)
```

andrewgordstewart (approved):
🔥

packages/server-wallet/src/wallet/wallet.ts
lalexgap: Done in 26da120

packages/server-wallet/src/models/channel.ts
lalexgap: Done in 26da120

packages/server-wallet/src/wallet/__test__/challenge.test.ts
lalexgap: Done in b5544fe

PR #3221, opened 2021-01-27
ci: succeed early if canary already published

geoknee:
Please see #3191 for more context. This is a superior solution. 

I tested this by running the GH action manually from this branch. The first time the action ran, packages were published. I then reran the action and it [succeeded early](https://github.com/statechannels/statechannels/actions/runs/517468861) after 17 seconds. Unfortunately the original run seems to have disappeared since I re-ran: but proof that it ran is [here.](https://www.npmjs.com/package/@statechannels/server-wallet/v/1.14.2-next.2ee26f4ed.441)

### DIFF
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -19,17 +19,25 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: 12.16.3
-
-      - name: yarn install 
-        # skip the prepare step here, as lerna publish will run prepare before publishing
-        run: |
-          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-canary.yml
```diff
@@ -20,6 +20,10 @@ jobs:
         with:
           node-version: 12.16.3
 
+      - name: Succeed early if a canary has already been published at this commit
+        if: $(npm view @statechannels/server-wallet dist-tags.next) =~ $(git rev-parse --short HEAD)
```
geoknee, line 24: I have singled out `server-wallet` here, but *all* packages are published when in canary mode (so no need to check the others).

andrewgordstewart (approved):
Can you explain in the the pr description, or preferably the squashed commit message, why it's appropriate to single out server-wallet?

PR #3237, opened 2021-02-01
perf: use uint8 types for numParticipants

geoknee:
An upper limit on the number of participants (255) is currently imposed in our contracts at runtime. The same limit is implied on the number of states, signatures, and `whoSignedWhats`, and any temporary variables that are used as indices in loops but will not exceed 255. In most places, we use the `uint8` type for these variables, but not all. This PR converts all remaining instances I could find. 

Advantages: Reduces confusion about what the type of `numParticipants` is. There may be small gas savings to be had as a result.
Disadvantages: small chance of introducing an [over/underflow error. ](https://docs.soliditylang.org/en/v0.5.3/security-considerations.html#two-s-complement-underflows-overflows) This can be mitigated by careful review of this PR. 

Follows on from https://github.com/statechannels/statechannels/pull/2981, see also the [docs](https://docs.statechannels.org/docs/protocol-tutorial/precautions-and-limits#precautions)

### Additional changes

Export CountingApp artifact from nitro protocol package, so that the testing code in the server wallet can import the relevant bytecode (which was previously hardcoded and now has changed). 

---
### Code quality
### Project management


tomclose (approved):
The changes here look good, but it doesn't seem feasible for me as a reviewer to identify whether this has introduced overflow problems. I vaguely remember the auditors pointing out a place where we were being saved from overflows only because we were using `uint256` instead of `uint8`. Did we fix that?

geoknee:
Update: it seems that the gas costs have actually *increased* slightly as a result of these changes. This is likely due to type casting, and the fact that variables can't be efficiently packed in memory / calldata. [link.](https://mudit.blog/solidity-gas-optimization-tips/)

geoknee:
Given the increase in gas consumption, I don't think the benefits weigh enough to merge this. Have cherry-picked the final two commits on #3248.

PR #3242, opened 2021-02-02
emit operationStarted/operationSucceeded events

geoknee:
Thoughts: 
- How do events share responsibility with api call return values?
- do we really have type safety in our event emitter?
- could/should we integrate event emission into store methods (store cannot emit events)
- is "operation" a good name? "quest"? ⚔️ 🐉  we don't have a hierarchy of objectives at the moment so we could just stick with "objective"

> # Description
> **Title** Clear, single line description of the pull request, written as though it was an order.
> 
> Then, following a single blank line, a summary of the change and which issue is fixed, including 
> - [ ] what is the problem being solved
> - [ ] why this is a good approach
> - [ ] what are the shortcomings of this approach, if any
> 
> It may include
> - [ ] background information, such as "Fixes #"
> - [ ] benchmark results
> - [ ] links to design documents
> - [ ] dependencies required for this change
> 
> The summary may omit some of these details if this PR fixes a single ticket that includes these details. It is the reviewer's discretion. 
> 
> ## Changes [Optional] 
> Multiple changes are not recommended, so this section should normally be omitted. Unfortunately, they are sometimes unavoidable. If there are multiple logical changes, list them separately.
> 
> 1. `'foo'` is replaced by `'bar'`
> 2. `'fizzbuzz'` is optimized for gas
> 
> ## How Has This Been Tested? [Optional] 
> 
> Did you need to run manual tests to verify this change? If so, please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
> 
> ## :warning: Does this require multiple approvals? [Optional]
> Please explain which reason, if any, why this requires more than one approval.
> - [ ] Is it security related?
> - [ ] Is it a significant process change?
> - [ ] Is it a significant change to architectural, design?
> 
> ---
> ## Checklist:
> 
> ### Code quality
> - [ ] I have written clear commit messages
> - [ ] I have performed a self-review of my own code
> - [ ] This change does not have an unduly wide scope
> - [ ] I have separated logic changes from refactor changes (formatting, renames, etc.)
> - [ ] I have commented my code wherever necessary (can be 0)
> - [ ] I have added tests that prove my fix is effective or that my feature works, if necessary
> ### Project management
> - [ ] I have applied the [appropriate labels](https://www.notion.so/Team-working-agreements-2a95c926bb5642e5a5c42e4b74a9dd24#b304e56734a74dfbb341b8b4b27b1c0c)
> - [ ] I have linked to all relevant issues (can be 0)
> - [ ] I have added all dependent tickets (can be 0)
> - [ ] I have assigned myself to this PR
> - [ ] I have chosen the appropriate pipeline on zenhub
>

### DIFF
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -63,7 +63,7 @@ export function isSharedObjective(
 /**
  * A DBObjective is a wire objective with a status and an objectiveId
  *
- * Limited to 'OpenChannel' and 'CloseChannel', which are the only objectives
+ * Limited to 'OpenChannel', 'CloseChannel', 'SubmitChallenge' and 'DefundChannel' which are the only objectives
  * that are currently supported by the server wallet
  */
 export type DBObjective =
--- a/packages/server-wallet/src/wallet/__test__/challenge.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/challenge.test.ts
@@ -36,6 +36,8 @@ it('throws an error when challenging with a non ledger channel', async () => {
 });
 it('submits a challenge when no challenge exists for a channel', async () => {
   const spy = jest.spyOn(w.chainService, 'challenge');
+  const callback = jest.fn();
+  w.once('operationStarted', callback);
   const c = channel({
     channelNonce: 1,
     // Set a random address so this will be a "ledger" channel
@@ -54,6 +56,7 @@ it('submits a challenge when no challenge exists for a channel', async () => {
 
   await w.challenge(channelId);
   expect(spy).toHaveBeenCalledWith(c.initialSupport, alice().privateKey);
+  expect(callback).toHaveBeenCalledWith(expect.objectContaining({type: 'SubmitChallenge'}));
 });
 
 it('stores the challenge state on the challenge created event', async () => {
--- a/packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/create-channel.test.ts
@@ -18,15 +18,16 @@ afterEach(async () => {
 describe('happy path', () => {
   beforeEach(async () => seedAlicesSigningWallet(w.knex));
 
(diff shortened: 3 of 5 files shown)


andrewgordstewart:
> Thoughts:
>     * is "operation" a good name? "quest"? ⚔️ 🐉  we don't have a hierarchy of objectives at the moment so we could just stick with "objective"

I'm happy to stick with "objective".

_The purpose of choosing "Operation" as opposed to "Objective" was to avoid overloading. The reason to avoid overloading is, in discussions between myself and @kerzhner, it seemed as if objectives could be nested or sequential. Therefore, there ought to be a 1-1 relationship between _certain_ API methods and "operations"._

Per our discussion this morning, it now feels like there's a 1-1 relationship between _certain_ API methods and "objectives". This was not known to us at the time.

In the future, the wallet may break objectives into subtasks, which would be an implementation detail. (The subtasks may end up in the wire protocol, but that's still not relevant to the wallet API.)

>     * How do events share responsibility with api call return values?

My opinion:
1. An API call which creates a new objective should, at a minimum, return an objective id. This should probably also trigger an `ObjectiveStarted` event for consistency.
2. `pushMessage` doesn't create a new objective. It's just sharing an already existing objective from another wallet. This should trigger a `ObjectiveStarted` event.
3. `ObjectiveSucceeded` event should not be a return value of a wallet call. It should only be an emitted event.

>     * do we really have type safety in our event emitter?

How strongly do we need it? Is it hard to get?

I think it's acceptable for the event emitter types to not be inferred, but just created from a common source. Eg.
- The objective types are defined in `objective-types.ts`
- the event emitter emits a type which is derived from the `Objective` type in `objective-types.ts`
- the Objective model stores a type which is derived from the `Objective` type in `objective-types.ts`

>     * could/should we integrate event emission into store methods (store cannot emit events)

I think the wallet can emit these events itself. Having the store emit the event seems to be leaking responsibility?

geoknee:
In favour of #3243

Issue #3124, opened 2021-01-08
Canonical defund channel protocol


andrewgordstewart:
See discussion [here](https://www.notion.so/Jan-11-Close-channel-protocol-4a167972a7264e1ebbb594c4624c152b#0e9cf1069c4a4317b7ed0a490d6ebd62)

Certain protocols share defunding logic. At the moment `CloseChannel` and `DefundChannel` protocols defund a channel. In the future, `ChallengeChannel` will replace `DefundChannel`. We can create a new `Defunder` protocol that unifies defunding logic for any type of channel (regardless of how a channel is funded).

Some of the functionality of the `Defunder` protocol is:
- When the channel is ledger-funded, create a LedgerDefund request
- When we have a conclusion proof and the channel is directly funded, submit a `ConcludeAndWithdraw` transaction.

Once the `Defunder` protocol exists, `CloseChannel`'s responsibility will become:
- Make sure the channel is finalizable, by exchanging a conclusion proof
- Then, trigger a `Defunder` protocol.

wunnaii:
note: Liam would like to chat/discuss this topic if needed (per sprint planning session).

PR #3244, opened 2021-02-02
Close channel refactor

kerzhner:
The refactor eliminates the action pattern previously used by protocol.

This work is a progression toward https://github.com/statechannels/statechannels/issues/3124. To tackle that issue, the first step is to isolate the defunding logic in the close channel protocol. I had trouble doing so with the current implementation. Also, we are no longer using the "action pattern" in any other protocol.

As this PR is a refactor, the PR relies on existing tests to verify correctness of changes.


kerzhner:
> Is this ready for review? I see that some project management tasks in the PR checklist were not done for this PR. For instance, it is not connected to the issue it pertains to.

Yep it is ready for review. I added this PR as a blocker to https://github.com/statechannels/statechannels/issues/3124. I don't think I am missing anything else, but maybe I am...

andrewgordstewart:
Making this _block_ #3124 is confusing to me, since it is _addressing_ #3124. So it seems to me like it should be _connected_ to #3124:

Screen Shot 2021-02-02 at 3 56 42 PM

packages/server-wallet/src/protocols/close-channel.ts
```diff
@@ -55,70 +43,100 @@ export class ChannelCloser {
...
+            attemptAnotherProtocolStep = false;
+            await this.completeObjective(objective, protocolState, tx, response);
+          } else {
+            response.queueChannelState(protocolState.app);
             attemptAnotherProtocolStep = false;
           }
-        } else {
-          response.queueChannelState(protocolState.app);
+        } catch (error) {
+          this.logger.error({error}, 'Error handling action');
```
andrewgordstewart, line 67: It's not really handling an action anymore, right?

andrewgordstewart (approved):

packages/server-wallet/src/protocols/close-channel.ts
kerzhner: fa753ddba

kerzhner:
> Making this _block_ #3124 is confusing to me, since it is _addressing_ #3124. So it seems to me like it should be _connected_ to #3124:
> 
> Screen Shot 2021-02-02 at 3 56 42 PM

This PR does not resolve https://github.com/statechannels/statechannels/issues/3124. That issue should probably get reworded since its purpose is for channel closing objective and channel defunding objective (which might actually become the channel challenge objective) to share the defunding logic.

Note that this PR refactors the close channel protocol to create an avenue to share the defunding logic between any objective that needs the defunding logic.

PR #3249, opened 2021-02-04
Clean up + simplify close channel protocol

kerzhner:
This PR:
- Removes the custom close channel `ProtocolState`. Since protocols are no longer pure functions, it is not necessary to gather all data upfront before running the protocol. `ProtocolState` introduces an unneeded level of indirection.
- Remove the loop in the protocol. Protocols are no longer written in a way where a protocol can only return a single action. Consequently, we can many actions in a row in the crank method without a loop.

This work is building toward addressing https://github.com/statechannels/statechannels/issues/3124. This PR is a refactor and relies on existing tests.


packages/server-wallet/src/protocols/channel-closer.ts
```diff
@@ -39,142 +38,137 @@ export class ChannelCloser {
...
-          } else {
-            response.queueChannelState(protocolState.app);
-            attemptAnotherProtocolStep = false;
-          }
-        } catch (error) {
-          this.logger.error({error}, 'Error taking a protocol step');
-          await tx.rollback(error);
-          attemptAnotherProtocolStep = false;
+        const turnNum = turnNumberToSign(channel);
+        // Explicitely check undefined just in case we want to sign turnNum 0
```
tomclose, line 52: typo: explicitly

packages/server-wallet/src/protocols/channel-closer.ts
```diff
@@ -39,142 +38,137 @@ export class ChannelCloser {
   public async crank(objective: DBCloseChannelObjective, response: WalletResponse): Promise<void> {
     const channelToLock = objective.data.targetChannelId;
 
-    let attemptAnotherProtocolStep = true;
+    await this.store.lockApp(channelToLock, async tx => {
+      const channel = await this.store.getChannel(channelToLock, tx);
```
tomclose, line 42: I might be wrong, but isn't `channelToLock` already this channel?

tomclose (approved):
LGTM 🎉

packages/server-wallet/src/protocols/channel-closer.ts
```diff
@@ -39,142 +38,137 @@ export class ChannelCloser {
...
-            attemptAnotherProtocolStep = false;
-          }
-        } catch (error) {
-          this.logger.error({error}, 'Error taking a protocol step');
-          await tx.rollback(error);
-          attemptAnotherProtocolStep = false;
+        const turnNum = turnNumberToSign(channel);
+        // Explicitely check undefined just in case we want to sign turnNum 0
+        if (turnNum !== undefined) {
+          await this.signState(channel, turnNum, tx, response);
```
tomclose, line 54: Should there be a check to ensure that we haven't already done this on a previous crank?

packages/server-wallet/src/protocols/channel-closer.ts
```diff
@@ -39,142 +38,137 @@ export class ChannelCloser {
...
+        if (await this.shouldCompleteObjective(channel, tx)) {
+          await this.completeObjective(objective, channel, tx, response);
+          return;
         }
-      });
+
+        response.queueChannel(channel);
+      } catch (error) {
+        this.logger.error({error}, 'Error taking a protocol step');
+        await tx.rollback(error);
```
tomclose, line 69: I'm not sure if we need this anymore. If the something throws an errors, the promise will reject and the transaction automatically rolls back, and the error is logged. It might(?) be better to allow `crank` to error and catch it elsewhere instead of swallowing here. It seems like any error thrown here will be developer errors, rather than data errors.

geoknee: We recently made a similar change for ledger manager cranking, moving the equivalent try/catch up into the wallet's `processLedgerQueue` function https://github.com/statechannels/statechannels/pull/3185. We could do the same thing for objectives, and put the try/catch in the `crankUntilIdle` function.

packages/server-wallet/src/protocols/channel-closer.ts
kerzhner: a29374f23

packages/server-wallet/src/protocols/channel-closer.ts
kerzhner: Thanks for pointing this out. The logic is changed in 39151e5a9.

My reasoning for using `store.getChannel` was that the method joins the channel row with the `ChainServiceRequests` table https://github.com/statechannels/statechannels/blob/e7ff05aef6e876d8e872137e3d2ceeb23aedfece/packages/server-wallet/src/wallet/store.ts#L250.

That's probably overkill. What was confusing me is that different channel database read methods join differring tables with the channels table

packages/server-wallet/src/protocols/channel-closer.ts
kerzhner: That's a good point. Filed https://github.com/statechannels/statechannels/issues/3252.

packages/server-wallet/src/protocols/channel-closer.ts
kerzhner: @tomclose I think you are pointing out that the implementation of the `crank` function could be clearer. I think the approach in f69f89c should help.

kerzhner:
39151e5...7ffeacf have been implemented after thinking about @tomclose's feedback. These changes are non-trivial, so it might be good to get a second pair of eyes on these. Maybe @geoknee or @andrewgordstewart?

packages/server-wallet/src/protocols/channel-closer.ts
```diff
@@ -39,226 +36,155 @@ export class ChannelCloser {
...
-    const vars: StateVariables = {...protocolState.app.supported, turnNum, isFinal: true};
-    const signedState = await this.store.signState(channel, vars, tx);
-    response.queueState(signedState, myIndex, channelId);
+    return false;
   }
 
-  private async defund(protocolState: ProtocolState, tx: Transaction): Promise<void> {
-    switch (protocolState.app.fundingStrategy) {
+  private async isChannelDefunded(c: Channel, tx: Transaction): Promise<boolean> {
+    const {protocolState: ps} = c;
```
andrewgordstewart, line 99: Is this a different `procotolState` than the protocol state that this PR claims to remove?

packages/server-wallet/src/protocols/channel-closer.ts
```diff
@@ -39,226 +36,155 @@ export class ChannelCloser {
...
-        await this.withdraw(protocolState, tx);
-        break;
-      case 'Ledger':
-        await this.requestLedgerDefunding(protocolState, tx);
-        break;
-      case 'Fake':
+        if (c.protocolState.directFundingStatus === 'Defunded') {
+          return true;
+        }
+        if (!c.chainServiceRequests.find(csr => csr.request === 'withdraw')?.isValid()) {
```
andrewgordstewart, line 106: I find `!c.chainServiceRequests.find(csr => csr.request === 'withdraw')?.isValid();` pretty hard to parse.
I think it says:
> There is no valid withdraw chain service request

What is a "valid" chain service request?

Is this clearer?
```suggestion
        const existingWithdrawRequest = c.chainServiceRequests.find(csr => csr.request === 'withdraw')?.isValid()
        if (!existingWithdrawRequest) {
```

packages/server-wallet/src/protocols/channel-closer.ts
kerzhner: This protocol state is https://github.com/statechannels/statechannels/blob/9ca60fdaf99196348aec8651a89349dd22d294fa/packages/server-wallet/src/models/channel.ts#L236

The removed protocol state is https://github.com/statechannels/statechannels/blob/9ca60fdaf99196348aec8651a89349dd22d294fa/packages/server-wallet/src/protocols/channel-closer.ts#L171

This confusion is yet another reason the second protocol state has been removed.

andrewgordstewart: Thanks for the clarification

packages/server-wallet/src/protocols/channel-closer.ts
kerzhner: Good point. This logic is going to get refactored in an upcoming PR as the defunding code will get factored out per https://github.com/statechannels/statechannels/issues/3124. So I am tempted not to address this now.

A valid chain service request is https://github.com/statechannels/statechannels/blob/9ca60fdaf99196348aec8651a89349dd22d294fa/packages/server-wallet/src/models/chain-service-request.ts#L57

We allow 3 retries for submitting a chain service request.

packages/server-wallet/src/protocols/channel-closer.ts
```diff
@@ -42,6 +42,10 @@ export class ChannelCloser {
       if (!channel) {
         throw new Error('Channel must exist');
       }
+
+      await channel.$fetchGraph('funding', {transaction: tx});
```
andrewgordstewart, line 44: Am I right to understand that we are changing the `channel-closer` implementation here because of the `directly-funded-channel` test?

Or, are you instead saying that
> Actually, it's important that we fetch the `funding` relationship here, as it's an important part of the `closeChannel` api

packages/server-wallet/src/protocols/channel-closer.ts
andrewgordstewart: `isValid` seems like a confusing name for that concept. `isStale`seems more accurate. Even better, calling its complement`failed/isFailed/hasFailed` seems the most clear.

packages/server-wallet/src/protocols/channel-closer.ts
kerzhner: This is a good question, and I do not know the answer. We are **not** changing the behavior of `channel-closer`. The channel-closer protocol state (which is now removed) used to fetch `funding` and `chainServiceRequests` for us via `store.getChannelState`.

andrewgordstewart:
> [39151e5...7ffeacf](https://github.com/statechannels/statechannels/compare/39151e5a9c0fd51207ffa63a73564a893f1729ad...7ffeacf2b1ca31604ed7a8503e9c78ae17401c18) have been implemented after thinking about @tomclose's feedback. These changes are non-trivial, so it might be good to get a second pair of eyes on these. Maybe @geoknee or @andrewgordstewart?

It felt impossible to review this subset, so I reviewed each commit in this PR. I didn't get to the point where I'd 👍 it, but there were no red flags other than https://github.com/statechannels/statechannels/pull/3249#discussion_r570624682. So, merge at your discretion!

packages/server-wallet/src/protocols/channel-closer.ts
andrewgordstewart: Ah, I understand now.
1. The refactor unintentionally changed the behaviour
2. https://github.com/statechannels/statechannels/pull/3249/commits/57a759ceaee22302b97158bf20e161aff869b210 reverted that behaviour change

kerzhner: Yep. I probably should have added that to the commit message.

andrewgordstewart: If you squash & merge, the context can be included in its commit message

Issue #3230, opened 2021-01-29
Objectives track times


andrewgordstewart:
# Why: Problem description :confused:
- A channel manager (CM) aware of "Operations" would want to know about long-running operations.
- An app using a CM may wish to monitor statistics about how long operations take

# What: Describe the solution you'd like :sparkles:  
1. Objectives should have a `created_at` and an `updated_at` column.
2. `updated_at` should be set to `now()` when the objective is updated at

## Open question
What should happen when a child objective is updated?

**Additional context**
- [Objective stalls due to an incorrect protocol design](https://www.notion.so/First-class-objectives-402f747da81b4bf38de1a3f28dd6ca31#c6deeaa3fb504875bbe023d7bd097b81)
- [Objective stalls due to a network error](https://www.notion.so/First-class-objectives-402f747da81b4bf38de1a3f28dd6ca31#b2d0ca76b807453e8f0d8017f6f8647a)

wunnaii:
Do not start this ticket if #[3225 ](https://github.com/statechannels/statechannels/issues/3225) is in progress

geoknee:
Thoughts from standup 4th Feb:

`createdAt` could be `lastCrankedAt` or `progressLastMadeAt`, depending on what we deem important to store. Then, we may need to think more carefully about how to detect those events.

geoknee:
The API types should be updated to include 
- the timestamps
- the objective id
- *not* the status (this will be clear from context, e.g. the name of the event fired when an objective completes)

PR #3250, opened 2021-02-04
refactor: make toWireObjective type safe / more robust

geoknee:
The lodash utility currently in use does not provide type safety. For example, if we add properties to `dbObj: DBObjective`, they will not be omitted unless we update this function, and this will lead to a runtime error when we try to validate the objective coming over the wire, due to pollutant properties.

Here, any required properties will be checked by the TS compiler. Optional properties are still a bit of a risk -- I will tackle that in another PR  #3251 

(Extra context: we may well add new properties to objectives, soon. See #3230). 

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -78,7 +78,12 @@ export const toWireObjective = (dbObj: DBObjective): SharedObjective => {
       'SubmitChallenge and DefundChannel objectives are not supported as wire objectives'
     );
   }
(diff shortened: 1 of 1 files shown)


andrewgordstewart:
> The lodash utility currently in use does not provide type safety.

To be pendantic, I suppose the utility does not provide _runtime safety_, ie. the type system does not adequately prevent the runtime behaviour that we want to avoid.

andrewgordstewart (approved):

PR #3253, opened 2021-02-04
WIP objective timestamps

geoknee:
work in progress


packages/server-wallet/src/models/__test__/objectives.test.ts
```diff
@@ -23,6 +23,10 @@ beforeEach(async () => {
   await seedAlicesSigningWallet(knex);
 });
 
+afterAll(async () => {
+  await DBAdmin.truncateDataBaseFromKnex(knex);
```
andrewgordstewart, line 27: Why does this pattern exist across our codebase? Don't tests truncate before they run? In that case, this leads to double truncating.

geoknee:
In favour of #3263

Issue #3241, opened 2021-02-02
`channels.vars` should not store channel constants


andrewgordstewart:
**Describe the bug**
The `vars` column should be an array of [`SignedStateVarsWithHash`](https://github.com/statechannels/statechannels/blob/a79b3a58b9e0204c67d0ee1b51a5602a1b3262e9/packages/wallet-core/src/types.ts#L89) JSON blobs. In practice, the array elements store channel constants, among other fields.

**To Reproduce**

First, remove the `afterAll` function from `ledger-manager.test.ts`:
<img width="923" alt="Screen Shot 2021-02-01 at 7 37 49 PM" src="https://user-images.githubusercontent.com/8432675/106544581-0333e680-64c5-11eb-8992-e4b05f888fc3.png">

This means you can deterministically run this test, and observe the data stored in the database after the test run.

Then, CD into the `server-wallet` package, and run

```
> yarn test ledger-manager -t 'generates signed ledger update when proposals were not identical but overlapped'
```

Now, you should have some channels where the `vars` column containing non-variables.
```
> psql server_wallet_payer -c 'select vars->0 from channels limit 1;'
```

After my test run, for instance, one `var -> 0` entry (piped through `jq`) looks like
```
❯ pbpaste | jq                                                    
{
  "appData": "0x",
  "chainId": "0x01",
  "isFinal": false,
  "outcome": {
    "type": "SimpleAllocation",
    "allocationItems": [
      {
        "amount": "0x01",
        "destination": "0x00000000000000000000000000000000000000000000000000000000000AaAa1"
      },
      {
        "amount": "0x00",
        "destination": "0x00000000000000000000000000000000000000000000000000000000000BbBB2"
      }
    ],
    "assetHolderAddress": "0x0000000000000000000000000000000000000000"
  },
  "turnNum": 0,
  "channelId": "0x3dc8e97155e1d74f9ba973780ced196d0d0974a2c387e20db58871b39641a136",
  "stateHash": "0x913ca5312745cecc8aa22d84c23d67aa6873294afb804b7e1b1e777de915bd73",
  "signatures": [
    {
      "signer": "0x11115FAf6f1BF263e81956F0Cc68aEc8426607cf",
      "signature": "0xe10890c9be256b68962d46b2771775eb2bb9a6222cd52ba1541667e2e89a1c8d7bde62a760d0a756d0b815d154ae445431b5f868e5e1d4a7d93204881a6e04551c"
    },
    {
[...]

PR #3151, opened 2021-01-15
fix: do not store constants in every vars

andrewgordstewart:
This should not be stored in the `vars` column of the DB:
```
[
  {
    "appData": "0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007e1fb3bdad8c8939b218cc471e51d81ce52bdf90d9d673a67d9e5be7091aac59000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000000",
    "chainId": "0x00",
    "isFinal": false,
    "outcome": {
      "type": "SimpleAllocation",
      "allocationItems": [
        {
          "amount": "0x02540be400",
          "destination": "0x00000000000000000000000011115FAf6f1BF263e81956F0Cc68aEc8426607cf"
        },
        {
          "amount": "0x00",
          "destination": "0x0000000000000000000000009F7ffcb016b0f7b142529bF27ef1eC5b0039C32C"
        }
      ],
      "assetHolderAddress": "0x0000000000000000000000000000000000000000"
    },
    "turnNum": 0,
    "stateHash": "0xe94ffb72d6edd40b5292cab7b89abe90668fbc971d5c56943e9a22cee7d5c65b",
    "signatures": [
      {
        "signer": "0xe252bA50dfC0Eb7705387acEd453c03963969092",
        "signature": "0x80295bb8356622ba8b8e0ae60b23fc65da36bf9954ffbf801a8d8751567c14a72408df6e72a2d35e88c319e4cbaf84174804c306c151b2ddd0a4710801e822e71c"
      },
      {
        "signer": "0x333356F625259653A2B6d7Da44162631DCbdF93F",
        "signature": "0x2adb475fc5dec52d2dc323945b3966ff84132fba96f9bdda25e69cc93a90ea1b03c064be47737ac03803f21584e933c486ecb0680ae6514a2a261385280180681b"
      }
    ],
    "channelNonce": 20,
    "participants": [
      {
        "destination": "0x00000000000000000000000011115FAf6f1BF263e81956F0Cc68aEc8426607cf",
[...]

PR #3255, opened 2021-02-05
Drop non variables when updating channels

isoosiss7:
# Description
Drop non variables in "vars" field when updating a channel

Fixes issue #3241

1. what is the problem being solved
As stated in issue https://github.com/statechannels/statechannels/pull/3151

2. why this is a good approach
Ensure non variables are dropped each time a new state is added

3. what are the shortcomings of this approach, if any
N/A

### DIFF
--- a/packages/server-wallet/src/state-utils.ts
+++ b/packages/server-wallet/src/state-utils.ts
@@ -67,7 +67,7 @@ export function addState(
   if (existing) {
     existing.signatures = _.uniqBy([...existing.signatures, ...signatures], 'signature');
   } else {
-    ret.push(signedState);
+    ret.push(dropNonVariables(signedState));
   }
 
   return ret;


andrewgordstewart:
[This](https://github.com/statechannels/statechannels/blob/25cda975d5645c087a9b08b8f60fa94e07debcc3/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts#L228) is an example of one of the failing assertions.

- `stateVar` has type `SignedState`
- `record.vars[0]` has type `SignedStateVarsWithHash`

These types are not really compatible: the former includes constants, but no `stateHash`, while the latter includes a `stateHash` but no constants. So, they shouldn't be compared in a test.

What's particularly sad, is the compiler would happily tell us that, if you try to compare them directly! 
Screen Shot 2021-02-04 at 10 34 58 PM
But, `expect(...).toMatchObject` isn't smart enough to realize that.

We could solve this test case with some tedious manipulations like this:
```
    const record = await Channel.forId(calculateChannelId(signedStates[1]), wallet.knex);
    expect(record.vars[0]).toMatchObject(dropNonVariables(addHash(signedStates[1])));
```

But, these tests shouldn't be concerned with _how_ the data is stored internally, ie. what the [`vars` column](https://github.com/statechannels/statechannels/blob/25cda975d5645c087a9b08b8f60fa94e07debcc3/packages/server-wallet/src/models/channel.ts#L70) contains.

Instead, they should probably be concerned with which `SignedStates` get stored. There is a getter function [here](https://github.com/statechannels/statechannels/blob/25cda975d5645c087a9b08b8f60fa94e07debcc3/packages/server-wallet/src/models/channel.ts#L299) which is more appropriate. So, the test could be changed to be
```
    expect(record.sortedStates[0]).toMatchObject(stateVar);
```

As an aside, `stateVar` is a confusing variable name here, because it's a `SignedState`, not a `StateVariables` or a `SigneStateVars`.

andrewgordstewart (approved):
Welcome to the project!

PR #3256, opened 2021-02-07
Do not rely on directFundingStatus internally

kerzhner:
`directFundingStatus` of `ChannelState` has likely outlived its usefulness. The original intention with `directFundingStatus` was to create a function that maps a channel + funding database state to a channel funding string.

We are seeing the channel opener and defunding protocols reimplement parts of the `directFundingStatus`. That's probably the right approach. This way we can build up smaller functions that can answer specific questions such as:
- Is this directly funded channel fully funded?
- Do I have any funds in this channel that I can claim?

This PR is working toward addressing https://github.com/statechannels/statechannels/issues/3124.

This PR is a refactor, so no additional tests have been added.

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -420,6 +420,15 @@ export class Channel extends Model implements RequiredColumns {
 
   // Does the channel have funds to pay out to all allocation items?
   public get isFullyDirectFunded(): boolean {
+    return this.isDirectFunded(true);
+  }
+
+  // Does the channel have any funds that I can claim?
+  public get isPartlyDirectlyFunded(): boolean {
+    return this.isDirectFunded();
+  }
+
+  private isDirectFunded(fully?: boolean): boolean {
     const outcome = this.supported?.outcome;
     if (!outcome) {
       throw new Error(`Channel passed to isFullyDirectFunded has no supported state`);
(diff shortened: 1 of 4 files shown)


packages/server-wallet/src/models/channel.ts
```diff
@@ -414,8 +418,71 @@ export class Channel extends Model implements RequiredColumns {
...
+
+  /**
+   * Returns a triple of balance [targetBefore, targetAfter, targetTotal], where
+   *  - targetBefore is the balance where I should start depositing
+   *  - targetAfter should be the balance where I stop depositing
+   *  - targetTotal is the total amount that should be in the channel
+   *
+   * @param channel
+   */
+  public get fundingMilestones(): [Uint256, Uint256, Uint256] {
```
lalexgap, line 458: Perhaps this is my preference but I'd prefer returning a type like `{ targetBefore: Uint256, targetAfter: Uint256, targetTotal: Uint256}`. I feel like I'd often get the correct order of the array confused (although to be fair, it will display in intellisense).

packages/server-wallet/src/models/channel.ts
```diff
@@ -414,8 +418,71 @@ export class Channel extends Model implements RequiredColumns {
...
+  public get isFullyDirectFunded(): boolean {
+    return this.isDirectFunded(true);
+  }
+
+  // Does the channel have any funds that I can claim?
+  public get isPartlyDirectFunded(): boolean {
+    return this.isDirectFunded();
+  }
+
+  private isDirectFunded(fully?: boolean): boolean {
```
lalexgap, line 431: Nit: With `this.isDirectFunded(true)` it's not immediately clear what the boolean means.

If we switched it to accept a `"Fully"|"Partially"` type we'd have `this.isDirectFunded('Fully')` which I'd argue is more straightforward.

lalexgap (approved):
🌋

packages/server-wallet/src/models/channel.ts
kerzhner: This is a good point! `fundingMilestones` was copy/pasted from another location, so I didn't think too deeply about the interface. There are 3 PRs that are based on this branch. So my preference is to file an issue and address the issue shortly after this PR is merged.

FIled https://github.com/statechannels/statechannels/issues/3268.

packages/server-wallet/src/models/channel.ts
kerzhner: Good point! Filed https://github.com/statechannels/statechannels/issues/3269

PR #3257, opened 2021-02-07
Unify defunding logic between Close and Defund protocols

kerzhner:
This PR creates a new `Defunder` class. The purpose of the class is to implement the `isChannelDefunded` function. The function is meant to operate as a cranker (and maybe should be renamed to `crank`?). This approach mimics the funding approach in channel opener.

CloseChannel and DefundChannel (and later ChallengeChannel) protocols share the `Defunder` class to make progress toward defunding a channel. The Defunder interprets the complete state of the channel (states signed and chain state) and attempts to take the next step in defunding the channel if such a step can be taken.

This PR is a refactor and has no additional tests. This PR is making progress toward https://github.com/statechannels/statechannels/issues/3124.

### DIFF
--- a/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
@@ -1,5 +1,4 @@
 import {makeAddress, State} from '@statechannels/wallet-core';
-import {ETH_ASSET_HOLDER_ADDRESS} from '@statechannels/wallet-core/src/config';
 
 import {DBAdmin} from '../../db-admin/db-admin';
 import {defaultTestConfig} from '../..';
@@ -14,14 +13,19 @@ import {ErorringMockChainService, MockChainService} from '../../chain-service';
 import {WalletResponse} from '../../wallet/wallet-response';
 import {ChannelDefunder} from '../defund-channel';
 import {AdjudicatorStatusModel} from '../../models/adjudicator-status';
-import {stateSignedBy, stateWithHashSignedBy} from '../../wallet/__test__/fixtures/states';
-import {alice, bob} from '../../wallet/__test__/fixtures/signing-wallets';
+import {stateSignedBy} from '../../wallet/__test__/fixtures/states';
+import {alice} from '../../wallet/__test__/fixtures/signing-wallets';
 import {Funding} from '../../models/funding';
+import {TestChannel} from '../../wallet/__test__/fixtures/test-channel';
 
 const logger = createLogger(defaultTestConfig());
 const timingMetrics = false;
 
 let store: Store;
+
+const FINAL = 10; // this will be A's state to sign
+const testChan = TestChannel.create({aBal: 5, bBal: 5});
+const testChan2 = TestChannel.create({aBal: 5, bBal: 5, finalFrom: FINAL});
 
 beforeEach(async () => {
   await DBAdmin.truncateDataBaseFromKnex(knex);
@@ -41,11 +45,13 @@ describe('when there is no challenge or finalized channel', () => {
     const channelDefunder = ChannelDefunder.create(store, chainService, logger, timingMetrics);
 
     // Create the channel in the database
-    const c = channel();
-    await Channel.query(knex).insert(c);
+    await testChan.insertInto(store, {
+      participant: 0,
+      states: [0, 1],
+    });
 
     // Set the objective in the database
-    const objective = await createPendingObjective(c.channelId);
+    const objective = await createPendingObjective(testChan.channelId);
     await knex.transaction(tx => store.ensureObjective(objective, tx));
 
     const pushSpy = jest.spyOn(chainService, 'pushOutcomeAndWithdraw');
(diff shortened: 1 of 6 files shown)


packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
```diff
@@ -14,15 +13,20 @@ import {ErorringMockChainService, MockChainService} from '../../chain-service';
...
 import {Funding} from '../../models/funding';
+import {TestChannel} from '../../wallet/__test__/fixtures/test-channel';
 
 const logger = createLogger(defaultTestConfig());
 const timingMetrics = false;
 
 let store: Store;
 
+const FINAL = 10; // this will be A's state to sign
+const testChan = TestChannel.create({aBal: 5, bBal: 5});
```
lalexgap, line 27: This is outside the scope of this PR but right now we have `TestChannel` and the `channel` fixture which I think do more or less the same thing. We should probably just standardize on one and remove the other.

packages/server-wallet/src/protocols/defunder.ts
```diff
@@ -0,0 +1,104 @@
...
+  public static create(
+    store: Store,
+    chainService: ChainServiceInterface,
+    logger: Logger,
+    timingMetrics = false
+  ): Defunder {
+    return new Defunder(store, chainService, logger, timingMetrics);
+  }
+
+  public async isChannelDefunded(channel: Channel, tx: Transaction): Promise<DefunderResult> {
```
lalexgap, line 31: I think I'd prefer `crank` as `isChannelDefunded` sounds more like a check than a function that is doing work.

packages/server-wallet/src/protocols/defunder.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    const adjudicatorStatus = await AdjudicatorStatusModel.getAdjudicatorStatus(
+      tx,
+      channel.channelId
+    );
+    let didSubmitTransaction = false;
+    if (adjudicatorStatus.channelMode !== 'Finalized' && channel.hasConclusionProof) {
+      await ChainServiceRequest.insertOrUpdate(channel.channelId, 'withdraw', tx);
+
+      // supported is defined (if the wallet is functioning correctly), but the compiler is not aware of that
```
lalexgap, line 69: nFor cases like this (where we always expect something to be defined if the wallet is working correctly), I've often added an `if` statement and then log a warning or error.

 I'm not sure if that's the right solution but I feel like this is a pretty common case so it might make sense for us to adopt a common pattern?

packages/server-wallet/src/protocols/defund-channel.ts
```diff
@@ -27,44 +25,37 @@ export class ChannelDefunder {
...
+      const {didSubmitTransaction} = await Defunder.create(
+        this.store,
+        this.chainService,
+        this.logger,
+        this.timingMetrics
+      ).isChannelDefunded(channel, tx);
+
+      // A better methodology is likely to create a Challenge objective that succeeds after a
+      // channel has been defunded (instead of succeeding an objective on transaction submission)
+      if (didSubmitTransaction) {
```
lalexgap, line 56: Does this mean that after we submit a `challenge` the defund objective has succeeded?

packages/server-wallet/src/protocols/__test__/defund-channel.test.ts
kerzhner: Good point! https://github.com/statechannels/statechannels/issues/3270

packages/server-wallet/src/protocols/defunder.ts
kerzhner: Agreed! 1a2bc6705

packages/server-wallet/src/protocols/defund-channel.ts
kerzhner: It does. I believe that is how the Defund protocol behaves before this refactor.

packages/server-wallet/src/protocols/defunder.ts
kerzhner: This is a good point. Addressed in b1f22f2b9

I prefer throwing an error. In this case, if `channel.supported` is undefined `chainService.concludeAndWithdraw` will not succeed.

packages/server-wallet/src/protocols/defund-channel.ts
lalexgap: Originally, the defund and challenge submission were two different objectives. When a challenge expired we'd create a defund objective that would be responsible for submitting a `pushOutcomeAndWithdraw` transaction.

If that's still the case I think we could remove `didSubmitTransaction` and change the `Defunder` protocol to just worry about defunding.

kerzhner: Gotch! The goal with this PR was to have a pure refactor without changing functionalities of protocols. I agree with you that `didSubmitTransaction` should probably be removed once we implement the new Challenge objective.

At the moment though, unless we want to change the behavior of the Defund objective/protocol, I think we need `didSubmitTransacation`. But maybe I am missing something?

lalexgap (approved):
LGTM!

Issue #3240, opened 2021-02-02
`yarn test:e2e` does not run


andrewgordstewart:
**Describe the bug**
When I run `yarn test:e2e` in the `server-wallet` package, I see
```
(node:67401) UnhandledPromiseRejectionWarning: DBError: select "channels".* from "channels" - relation "channels" does not exist
```

**To Reproduce**
```
❯ dropdb server_wallet_payer;
❯ dropdb server_wallet_receiver;
```
Then, follow the instructions in the [readme](https://github.com/statechannels/statechannels/tree/master/packages/server-wallet)
```
❯ yarn test:e2e;
```

**Expected behavior**
To run the e2e test, we need to
1. create & migrate `server_wallet_payer`
2. create & migrate `server_wallet_receiver`

Either
A. the readme should instruct a developer to do 1 & 2
B. the e2e test itself should do 1 & 2
C. there should be a `run_e2e.sh` script which does 1 ^ 2, similar to [this](https://github.com/statechannels/graph-payments/blob/main/packages/e2e-testing/scripts/run-local.sh)

**Additional context**

The instructions [here](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/e2e-test/README.md#running-locally) are out of date, and are not referenced in either the root readme or the `server-wallet` readme.

isoosiss7:
I can fix this, assigned to myself

isoosiss7:
Pls review: https://github.com/statechannels/statechannels/pull/3258

andrewgordstewart:
Fixed by https://github.com/statechannels/statechannels/pull/3258

PR #3258, opened 2021-02-08
enhancement:make sure e2e dbs are ready before testing

isoosiss7:
https://github.com/statechannels/statechannels/issues/3240

# Description
make sure server_wallet_payer and server_wallet_receiver are created and migrated before e2e

- [x] what is the problem being solved
E2E requires 2 databases: server_wallet_payer and server_wallet_receiver, these could be not set up or not up to date when e2e runs
- [x] why this is a good approach
Create if not exists, then migrate
- [x] what are the shortcomings of this approach, if any
Not sure if we should do the same for server_wallet_test, I guess we should leave it to be controlled by the developer. But this causes inconsistency.


geoknee (approved):
@isoosiss7 I have two questions: 

- should we also try and catch errors when the database already exists? As in the approach of #3266 
- if you take a look at `.circleci/config.yml` you will see that there is a step called `server-wallet setup (postgresql migrations`. With your changes, it looks like we will end up trying to create and migrate these databases twice on circle. Perhaps we could just remove that step?

packages/server-wallet/src/test_db.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# 1. recreate db if not exist
+# 2. migrate db to latest
+
+for var in "$@"
+do
+    echo setting "$var"...
```
andrewgordstewart, line 7: This is probably a bit more clear, right? 

```suggestion
    echo creating and migrating "$var"...
```

isoosiss7:
> @isoosiss7 I have two questions:
> 
> * should we also try and catch errors when the database already exists? As in the approach of #3266
> * if you take a look at `.circleci/config.yml` you will see that there is a step called `server-wallet setup (postgresql migrations`. With your changes, it looks like we will end up trying to create and migrate these databases twice on circle. Perhaps we could just remove that step?

Removed create db operations in .circleci/config.yml
Errors are ignored should the db already exist. but will still try to migrate db regardless.

PR #3259, opened 2021-02-08
refactor: ObjectiveModel.insert returns a DBObjective

geoknee:
Currently, the various methods on `ObjectiveModel` have different return types: either a promise resolving to a`DBObjective` or to an `ObjectiveModel` instance. Most code in `store.ts` uses the former type, so it is far more convenient to return that. 

The current workarounds in the various `ensureObjective` functions involve preparing a `DBObjective`, `insert`ing it, and if success, returning the thing that was prepared beforehand. This PR moves to a model where the `insert` method accepts a type that may *not* be a `DBObjective`. Now in the `ensureObjective` functions, we simply return the result of `insert` if it succeeds. 

The reason for this alternative pattern is that we shortly intend to introduce timestamps to `DBObjectives`. If these timestamps are added at the data model layer, it isn't going to be possible to prepare a `DBObjective` before inserting. 

### Shortcomings
The current behaviour of returning the existing objective on a `UniqueViolationError` is preserved, at the cost of having another query in the error handler. This behaviour can be tidied up further in a future PR. 

This is preparatory work for #3230. 

## Changes 

1. ObjectiveModel.insert returns a DBObjective
2. `ensureFooObjective` functions have narrower return types
3.  `ensureFooObjective` functions are exported

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -137,11 +137,11 @@ export class ObjectiveModel extends Model {
       status: 'pending' | 'approved' | 'rejected' | 'failed' | 'succeeded';
     },
     tx: TransactionOrKnex
-  ): Promise<ObjectiveModel> {
+  ): Promise<DBObjective> {
     const id: string = objectiveId(objectiveToBeStored);
 
     return tx.transaction(async trx => {
-      const objective = await ObjectiveModel.query(trx).insert({
+      const model = await ObjectiveModel.query(trx).insert({
         objectiveId: id,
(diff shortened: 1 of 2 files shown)


lalexgap (approved):
Awesome, I think this fixes #3122 🎉 

I wonder if we should refactor all the `ensureFoo` methods into one method? It seems like they're mostly doing the same thing.

geoknee:
> I wonder if we should refactor all the `ensureFoo` methods into one method? It seems like they're mostly doing the same thing

Yeah I agree. I have started work on that already.

PR #3260, opened 2021-02-08
docs: add notes

geoknee:
_Problem 1_: we don't have a reference or specification to work to when engineering in the server wallet. We do have quite a bit of material in Notion (largely due to @tomclose) that can help with this, but it is not clear:

a) if the proposals there have been agreed on, or not
b) to what extent the proposals have already been built out / implemented, or are more aspirational

_Problem 2_: The current wallet is likely impenetrable to new devs. 

This PR adds some notes

- on responsibilities of a state channel wallet
- on objective-driven architecture

whose aim is to start to tackle both problems. 

I think these types of problems are probably never going to be "done", but something we have to constantly chip away at. 

**Proposed criteria for merging this PR**

P1a) _the additions should only contain material that we are happy to take as "our current best attempt" at a specification_. So things that are very controversial right now should not be included. Things that seem mostly good, but might not be the final way or better than all possible alternatives, _are_ included. I think we have to strike a balance here, between:
 - getting the spec "right"
 - improving DevX now, and hopefully enjoying greater productivity
 - shipping in-demand features to production quickly
 - overcoming the inertia of the code that exists

P1b) _the material remains aspirational, and doesn't try to describe the wallet as it is (except in those cases where it is exactly as it aspires to be)_. The current code on `master`, although potentially opaque, describes itself, with the help of comments and TSDoc annotations. I think it would be a mistake to try and document exactly how the wallet is now, because it is still moving very fast (but hopefully towards the agreed on spec). Furthermore, there is not such a huge chasm between where it is and where it should be, such that a clear spec will still give pretty good understanding of how it works now. 

P2) _It should make sense to those with relatively low familiarity with our codebase._

My suggestion is, therefore, that we have a meta-discussion about these criteria. When agreed, we can then judge the PR by those criteria.


packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+
+We usually talk of “The App” as the actor that makes the API calls and binds to the events. The wallet can be a library or run in a separate process to The App (e.g. on a server, or run in an iframe).
+
+A state channel wallet _is not_:
+
+- equipped with a graphical user interface (although this might be a desired plugin)
+
+A state channel wallet has the following _runtime dependencies_:
+
+- a message delivery service (deferred to The App)
```
andrewgordstewart, line 30: I think this wording is compatible with the status quo API and the "core CM" proposed API.

```suggestion
- a message delivery service (provided by The App)
```

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+- a blockchain provider (e.g. infura)
+- a database (could be a Postgres server or indexedDB)
+
+The following diagram describes the various components in the architecture. The scope of this document is limited to the components left of the dividing line.
```
andrewgordstewart, line 34: How is this diagram generated? Without an easy way of generating it, it's not that maintainable. (Though, we shouldn't be updating this doc all that often.)

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+                                                      +------------------+
+```
+
+# The responsibilities of a state channel wallet
+
+As in the [Hippocratic oath](https://en.wikipedia.org/wiki/Hippocratic_Oath): _primum non nocere,_ (first, do no harm).
+
+1. Do not lose secrets (private keys)
+2. Do not leak secrets (private keys)
+3. Store relevant states and signatures (Don’t lose any unless they are screened off by kept ones, allow states to be queried)
```
andrewgordstewart, line 75: I don't know what "screened off" means. Is this wording more clear?

```suggestion
3. Store relevant states and signatures (Don’t lose any unless they are superceded by kept ones, allow states to be queried)
```

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+4. Do not sign states or blockchain transactions unless the App grants permission
+
+Next, do some active things to protect The App’s interests.
+
+5. With implicit permission, allow an exit on chain without losing funds
+6. Detect and store blockchain challenges (for querying)
+7. Detect and store changes in channel funding (for querying)
```
andrewgordstewart, line 82: What's the purpose of the (for querying) addition here? I feel like I'd understand 6-7 without it, but adding it makes me confused.
```suggestion
6. Detect and store blockchain challenges
7. Detect and store changes in channel funding
```

andrewgordstewart, line 82: It might even be more clear not to tell the audience that we store these things. I think this is the _intent_ of the wallet, from an "external point of view".

```suggestion
6. React to blockchain challenges
7. React to changes in channel funding
```

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+
+Next, do some active things to allow The App to express itself
+
+8. With explicit permission, sign updates in running application channels
+9. With implicit permission, fund channels on chain
+
+Next, allow for advanced funding relationships
+
+10. With implicit permission, sign “safe” updates in ledger channels (meaning no funds are lost)
```
andrewgordstewart, line 91: A wallet should also sign updates in eg. guarantor & joint channels.

```suggestion
10. With implicit permission, sign “safe” updates in "auxilliary" funding channels, such as a ledger channel

 "Safe" here means that no funds are lost
```

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+
+[More detail on responsibilities.](./responsibilities.md)
+
+# Objective-driven Wallet architecture
+
+In order to fulfil its responsibilities, a state channel wallet must be carefully architected. This section describes one approach that we are pursuing. Let's consider the most important responsibility of a state channels wallet, besides secure storage of secrets and states:
```
andrewgordstewart, line 97:
```suggestion
In order to fulfill its responsibilities, a state channel wallet must be carefully designed. This section describes our approach. Let's consider the most important responsibility of a state channels wallet, besides secure storage of secrets and states:
```

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+Here, “explicit permission” shall be implied by an API call. “Implicit permission” is governed by another mechanism that we call _Objectives._ Objectives are a device that allow for
+
+- greater abstraction of the details of nitro protocol
+- greater performance and robustness of a responsible wallet
+- easier coordination between counterparties
+- short-lived request/response cycle over the API
+
+> An objective is a proposal for a goal which The App asserts that it wants to achieve.
+
+It has a _name_ and _parameters_. The parameters include a _scope_, which is a list of channelIds.
```
andrewgordstewart, line 110:
```suggestion
It has a _name_ and _parameters_. The parameters determine a predefined set of actions that the wallet will take on The App's behalf.
```

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+
+> It is _not_ to give permission to spend money, only to sign states and transactions that are overall value-neutral (up to gas fees).
+
+## Implementation: the lifecycle of an objective
+
+### Creation
+
+Objectives are spawned:
+
+1. During request handling, i.e. during an API call. The API call resolves when the objective is committed to the wallet's store.
```
andrewgordstewart, line 122:
```suggestion
Objectives are created:

1. During an API call. The API call resolves when the objective is committed to the wallet's store.
```

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+
+Some objecties are "shared" and involve other participants. They are therefore communicated across "the wire" to other wallets. Therefore we also have objectives being spawned in a wallet:
```
andrewgordstewart, line 124:
```suggestion
Some objecties are "shared" and involve other participants. They are therefore communicated across "the wire" to other wallets. Therefore, objectives can be created by the wallet itself when handling a peer message.
```

## Unrelated comment

Note that this might raise a red flag to a security-oriented engineer: this gives me the ability to DDOS your wallet and jam-pack your store with objectives.

A security engineer might suggest that The App must somehow explicitly grant permission to "create the objective", since ["allowlisting"](https://www.zdnet.com/article/uk-ncsc-to-stop-using-whitelist-and-blacklist-due-to-racial-stereotyping/) is more secure than "blocklisting"*.

\* according to some [random blog](https://blog.finjan.com/blacklisting-vs-whitelisting-understanding-the-security-benefits-of-each/)
> In fact, NIST recommends the use of whitelisting in high-risk security environments, where the integrity of individual or connected systems is critical and takes precedence over any restrictions that users might suffer in their choice or access to software.
but you can find lots of similar opinions.

(Note that there is no reference to NIST's recommendation, but they do have a [Guide to Application Whitelisting](https://csrc.nist.rip/library/alt-SP800-167.pdf).)

packages/server-wallet/docs/responsibilities.md
```diff
@@ -0,0 +1,41 @@
+## More detail on responsibilities
+### 1. Do not lose secrets
+Store them in a database. 
+
+### 2. Do not leak secrets
+Minimise time spent in memory.
```
andrewgordstewart, line 6: This seems like the _how_, not the _what_.

packages/server-wallet/docs/responsibilities.md
```diff
@@ -0,0 +1,41 @@
...
+
+### 3. Store relevant states
+Accept states from counterparties through the `pushMessage` API method.
+In the extreme (“ideal” case), keep every state ever signed or received.
+In practice, garbage collect "old" states (should only need ~`nParticipants` states per channel), by tracking `validTransition` relationships between them.
```
andrewgordstewart, line 11:
```suggestion
Accept states from counterparties through wire messages.
In the extreme (“ideal” case), keep every state ever signed or received.
In practice, garbage collect "old" states (should only need ~`nParticipants` states per channel), by tracking `validTransition` relationships between them. 
```

It might be nice to comment on why the garbage collecting is safe.

andrewgordstewart (approved):
This seems like excellent information to maintain!

I left many suggestions that might be more clear or less tied to the specific, current implementation. I'm curious to hear your opinion on them.

packages/server-wallet/docs/index.md
geoknee: In my mind, reacting to these events is a separate responsibility of the wallet that builds on the more basic responsibility of storing those events. If a partially-functional wallet only stores the challenges but does not otherwise react to them, or if the code that reacts to challenge is somehow buggy, at least users can attempt to recover their funds manually by falling back on the more basic responsibility.

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+3.  CrankToCompletion (for any objectives whose scoped channels have been ‘touched’).
+4.  Repeat 2-3 until no more changes
+5.  ProcessLedgerQueue (see below)
+6.  Repeat 2-5 until no more changes.
+7.  Return consolidated updates and states if an API call.
+
+# Updating Funding
+
+TODO
+
```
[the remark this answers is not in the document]
geoknee: As I said in the PR description, it is my preference that this document is maintained as a record of what we want the wallet to be, rather than what it is.

packages/server-wallet/docs/index.md
geoknee: I think this is an example of where I interpret the terminology slightly differently to others. I tend to think of anything that is not an application channel as a ledger channel.

From the nitro paper: 
> A ledger channel is a channel whose sole purpose is to provide funding to
other channels. 

On the other hand, I think many other people do not interpret it this way. Also your recommendation works under both interpretations. So I went with it!

packages/server-wallet/docs/index.md
```diff
@@ -0,0 +1,183 @@
...
+- a blockchain provider (e.g. infura)
+- a database (could be a Postgres server or indexedDB)
+
+The following diagram describes the various components in the architecture. The scope of this document is limited to the components left of the dividing line.
+
+```
+                                                    |
+                     +-----------------+            |     +--------------+
+                     |                 |            |     |              |
+                     |  CHAIN-SERVICE  <------------------>  BLOCKCHAIN  |
```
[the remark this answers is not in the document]
geoknee: Sadly I don't think this will work because the whole diagram is inside a code fence. 

If only github [supported mermaid diagrams](https://github.community/t/feature-request-support-mermaid-markdown-graph-diagrams-in-md-files/1922/105), which could be a nice solution.

geoknee:
Thanks for the feedback @andrewgordstewart! I went a long with many of your suggestions. Let's see if we can keep this material up to data as we make changes to the architecture

packages/server-wallet/docs/index.md
andrewgordstewart: Is a guarantor channel a ledger channel?

packages/server-wallet/docs/index.md
geoknee: I added a comment about where the diagram was created in this commit https://github.com/statechannels/statechannels/pull/3260/commits/dc92169b273699918abfc673906a0bf3b6071051 . Unfortunately the link won't provide a super easy way to generate this diagram. But it does provide an easy way to generate diagrams of this kind.

packages/server-wallet/docs/index.md
andrewgordstewart: Could one equivalently argue that a wallet which stores states but does not sign new states (for an objective) exhibits the "basic responsibility" of being a wallet? In that case, a user can attempt to fund their app channel by signing ledger updates.

It certainly feels as though the wallet has an _obligation_ to react to a `Challenge` event in some fashion. It may react by calling `checkpoint`, if it detects that the challenge was malicious. It may also react by notifying The App:  "⚠️ !!! A challenge was detected! ⚠️".

geoknee: Yes I agree, a fully featured wallet should react to challenges. I suppose the direction I am trying to go in, with this ordered list of responsibilities, is that they are built up in a certain order based on "security priority". In that case, reacting to a challenge would be a separately-enumerated responsibility with a lower priority. A wallet should still meet that responsibility in order to be considered feature-complete; but it should not meet it without first meeting all other responsibilities with a higher priority. 

By placing emphasis on _storage_ high up the priority list, I suppose I am trying to discourage wallet builders from implementing features that do not survive a crash/restart cycle.  It's a "walk before you run" approach.

packages/server-wallet/docs/index.md
andrewgordstewart: That approach does not seem very maintainable. As stated, that's fine when we don't change the diagram.

An alternative like draw.io would allow us to host the "source" of the image in `.git`. Github would render the diagram when displaying a `.md` file. (I don't know what other `.md` renderers will do that, so the ascii diagram definitely has a leg up on portability.)

By [including an `edit-diagram.html`](https://desk.draw.io/support/solutions/articles/16000042371-embed-a-diagram-in-github-markdown) file in our repo, we can even directly edit the diagram, using the `draw.io` UI.

PR #3263, opened 2021-02-08
add objective timestamps

geoknee:
Add a `createdAt` and `progressLastMadeAt` columns to the `objectives` table. Does not add any methods to query these fields, yet (the next PR in this stack does that #3264). Towards #3230.

 **what is the problem being solved**
The app requires a means to detect inactivity on an ongoing objective, so that it can i) try and synchronize data with peers to get the objective going again or ii) launch a challenge to recover funds. This first step installs a location for the wallet to store the data required to track objective progress. 
**why this is a good approach**
The most important timestamp here is the `progressLastMadeAt`. If this timestamp is long enough ago (let's say, 10x the expected network latency), then this is a very strong indication that the objective will no longer progress and a different course of action is needed. 
**what are the shortcomings of this approach, if any**
- This mechanism will not cover inactivity in running app channels, as they are not to be governed by objectives. We may or may not want a similar solution for detecting that. 
- This `progressLastMadeAt` soluition requires us to think harder and write more code around when "progress has been made", compared to alternative approaches.  See #3264. Alternatives I considered were:

### `updatedAt` 
This could be managed directly by the database, and would update automatically when a row was patched. The downside is that the objective does not really contain any information (aside from a `status`) that tells us about it's lifecycle. The information is actually stored in other DB tables such as `channels`. (i.e. the channels in the scope of the objective).

### `lastCrankedAt`
This seems much better as a column on the  `objectives` table, and would also be quite easy to decide when to update it (we already have code that finds relevant objectives when the runloop is triggered, and cranks them). The downside here, in my opinion, is that we can crank and make no progress. For example: my counterparty is relentlessly sending me old states, and my objective keeps getting cranked, but no progress is made. While this is not indicative of inactivity, it likely still requires some action to be taken.

## How Has This Been Tested?
Test coverage was increased to cover timestamps. I have done this in a fuzzy way, expecting timestamps to lie in a 2 second window either side of the database query. This window is very large and the test is probably over permissive as a result. I could add a few-second delay to the test to ensure the `createdAt` and `progressLastMadeAt` stamps are further apart.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/models/__test__/objectives.test.ts
+++ b/packages/server-wallet/src/models/__test__/objectives.test.ts
@@ -61,11 +61,11 @@ describe('Objective > insert', () => {
     const {objectiveId} = await ObjectiveModel.insert({...objective, status: 'pending'}, knex);
 
     const before = Date.now() - 1000; // scroll back 1000 ms to allow for finite precision / rounding
-    const {updatedAt} = await ObjectiveModel.succeed(objectiveId, knex);
+    const {progressLastMadeAt} = await ObjectiveModel.succeed(objectiveId, knex);
     const after = Date.now() + 1000; // scroll forward 1000 ms to allow for finite precision / rounding
 
-    expect(Date.parse(updatedAt) > before).toBe(true);
-    expect(Date.parse(updatedAt) < after).toBe(true);
+    expect(Date.parse(progressLastMadeAt) > before).toBe(true);
+    expect(Date.parse(progressLastMadeAt) < after).toBe(true);
   });
 });
 
--- a/packages/server-wallet/src/models/__test__/objectives.test.ts
+++ b/packages/server-wallet/src/models/__test__/objectives.test.ts
@@ -18,7 +18,7 @@ const objective: OpenChannel = {
   },
 };
 beforeEach(async () => {
-  await seedAlicesSigningWallet(knex);
+  await seedAlicesSigningWallet(knex); // this also truncates
 });
 
 describe('Objective > insert', () => {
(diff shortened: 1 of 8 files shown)


packages/server-wallet/src/models/objective.ts
```diff
@@ -31,43 +30,32 @@ function extractReferencedChannels(objective: Objective): string[] {
...
 };
 
-export type DBSubmitChallengeObjective = {
-  data: {targetChannelId: string};
-  participants: Participant[];
-  objectiveId: string;
-  status: ObjectiveStatus;
-  type: 'SubmitChallenge';
-};
+export type DBOpenChannelObjective = WalletObjective<OpenChannel>;
```
andrewgordstewart, line 40: Nit: this change should have probably been broken up. Otherwise, the functional change (the addition of the timestamp) is mixed in with the refactor, which shouldn't change behaviour.

packages/server-wallet/src/models/__test__/objectives.test.ts
```diff
@@ -44,6 +44,29 @@ describe('Objective > insert', () => {
...
+    expect(Date.parse(createdAt) > before).toBe(true);
+    expect(Date.parse(createdAt) < after).toBe(true);
+  });
+
+  it('updates the timestamp on an objective when it succeeds', async () => {
+    await Channel.query(knex).withGraphFetched('signingWallet').insert(c);
+    const {objectiveId} = await ObjectiveModel.insert({...objective, status: 'pending'}, knex);
+
+    const before = Date.now() - 1000; // scroll back 1000 ms to allow for finite precision / rounding
+    const {updatedAt} = await ObjectiveModel.succeed(objectiveId, knex);
```
andrewgordstewart: I'm confused here, since I didn't think `updatedAt` was a property on objectives?

andrewgordstewart (approved):

packages/server-wallet/src/models/__test__/objectives.test.ts
geoknee: This was a mistake I made, but I think I fixed it before you posted your comment (maybe you started your review a while ago?)

andrewgordstewart: I often review commits atomically, so I came across `updatedAt` and it hadn't been defined yet in the git history.

PR #3264, opened 2021-02-08
update progress (conditionally) during objective cranking

geoknee:
Examine each cranking function and manually progress each objective's timestamp where necessary. 

Follows on from #3263

### DIFF
--- a/packages/server-wallet/server-wallet.api.md
+++ b/packages/server-wallet/server-wallet.api.md
@@ -293,6 +293,7 @@ export class SingleThreadedWallet extends EventEmitter<EventEmitterType> impleme
     destroy(): Promise<void>;
     getChannels(): Promise<MultipleChannelOutput>;
     getLedgerChannels(assetHolderAddress: string, participants: Participant_2[]): Promise<MultipleChannelOutput>;
+    getObjective(objectiveId: string): Promise<DBObjective>;
     getSigningAddress(): Promise<Address>;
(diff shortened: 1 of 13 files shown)


geoknee:
Following a discussion today with @tomclose:

- this approach doesn't do justice to the fact that progress can be made when my store was updated _before_ cranking, i.e. simply by my having observed a blockchain event or receiving a state from a counterparty.

A proposal to fix that problem:

- add a state `string` to objectives that _summarises_ the state of the store, without replicating any of the details. It is purposefully _not_ a rich object. It should be considered "write only" meta-data about the store, and not used as a second source of truth. It is only going to be read by the code that updates `progressLastMadeAt`.**

As an example, the `DBOpenChannel` objective, the state could be : 

`''`
`'preFundSigned'`
`'preFundSupported'`
`'partiallyFunded'`
`'funded'`
`'postFundSigned'`
`'success'`

When the cranking function updates this state, it will have access to the previous state and be able to infer if some progress has been made. If progress has been made, update `progressLastMadeAt`.

This proposal should also aid in debugging; and could aid in creating a UI (e.g. for a future browser wallet). It preserves the property of not being "fool-able" by any influx of state events/states. It *might* be the compromise that gives us the benefits of explicit state machines without some of the downsides?

** we could rename to `objectiveStatusLastChangeAt`.

geoknee:
In favour of #3286

PR #3266, opened 2021-02-09
enhance:createdb - don't error out if db already exists

isoosiss7:
# Description
Make "createdb" more robust by avoiding throwing fatal error if db already exists

- [ ] what is the problem being solved
UT repeatedly create "TEST_A" and "TEST_B", if one fails to clean up somehow, it will fail all future test case
- [ ] why this is a good approach
If db already exists then no good to recreate it regardless
- [ ] what are the shortcomings of this approach, if any
We yet don't know how DB destroy failed but it can be due to process killed while running.


geoknee (approved):
Thank you @isoosiss7 ! I assume that you tested this new approach?

isoosiss7:
> Thank you @isoosiss7 ! I assume that you tested this new approach?

Yep this change is tested.

PR #3272, opened 2021-02-09
refactor: ignore UniqueViolationError at the query level

geoknee:
This should be considered a true refactor and is designed to not affect behaviour. 

There are outstanding questions about whether `status` should be patched during the insert, if there is a conflict. Those should be tackled in future work. 

It is inelegant and needlessly verbose to catch and swallow `UniqueViolationErrors` in wallet code, especially when that code is replicated across several functions that wrap `ObjectiveModel.insert()`. (Again this replication is an issue that can be tackled in future work). 

This approach is good because it defines what should happen when we get a conflict on the id column, and because it happens at a lower level it saves us quite a few lines in the higher level. 

From the [knex docs](http://knexjs.org/#Builder-onConflict): 

> A conflict occurs when a table has a PRIMARY KEY or a UNIQUE index on a column (or a composite index on a set of columns) and a row being inserted has the same value as a row which already exists in the table in those column(s).

---
### Code quality
### Project management


packages/server-wallet/src/models/objective.ts
```diff
@@ -131,21 +131,28 @@ export class ObjectiveModel extends Model {
...
+        .insert({
+          objectiveId: id,
+          status: objectiveToBeStored.status,
+          type: objectiveToBeStored.type,
+          data: objectiveToBeStored.data,
+          createdAt: new Date().toISOString(),
+          progressLastMadeAt: new Date().toISOString(),
+        })
+        .onConflict('objectiveId')
+        .ignore(); // this makes it an upsert
```
andrewgordstewart, line 144: Should this return `*` so that we always get the latest row?

geoknee: Yes. We also need to decide what to do if say `OpenChannel-0x123` exists with status 'pending', and we try to insert `OpenChannel-0x123` with status 'approved'. This might be an argument for doing `onConflict('objectiveId').merge({status: objectiveToBeStored.status})`. Either way we need some unit tests for this so we can assert on the desired behaviour.

tomclose (approved):

PR #3274, opened 2021-02-09
Objective timestamps are now treated consistently as a Date type.

geoknee:
# Description
Objective timestamps are now treated consistently as a Date type.

Previously we treated timestamps on the objective as `strings` however `knex` would return a `Date` type that we would sometimes get converted to a `string` (depending on the code path)

To fix this  the timestamp fields have been changed to `Date`s on the `Objective`. This reflects the type `knex` gives us and will be easier to manipulate than a string.

 
## How Has This Been Tested?
Two new basic tests have been added that test the correct type is returned.

### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -131,14 +131,17 @@ export class ObjectiveModel extends Model {
     const id: string = objectiveId(objectiveToBeStored);
 
     return tx.transaction(async trx => {
-      const model = await ObjectiveModel.query(trx).insert({
-        objectiveId: id,
-        status: objectiveToBeStored.status,
-        type: objectiveToBeStored.type,
-        data: objectiveToBeStored.data,
-        createdAt: new Date(),
-        progressLastMadeAt: new Date(),
-      });
+      const model = await ObjectiveModel.query(trx)
+        .insert({
+          objectiveId: id,
+          status: objectiveToBeStored.status,
+          type: objectiveToBeStored.type,
+          data: objectiveToBeStored.data,
+          createdAt: new Date(),
+          progressLastMadeAt: new Date(),
+        })
+        .returning('*')
+        .first(); // This ensures that the returned object undergoes any type conversion performed during insert
 
       // Associate the objective with any channel that it references
       // By inserting an ObjectiveChannel row for each channel
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -136,8 +136,8 @@ export class ObjectiveModel extends Model {
         status: objectiveToBeStored.status,
         type: objectiveToBeStored.type,
         data: objectiveToBeStored.data,
-        createdAt: new Date(Date.now()),
-        progressLastMadeAt: new Date(Date.now()),
+        createdAt: new Date(),
+        progressLastMadeAt: new Date(),
       });
 
       // Associate the objective with any channel that it references
@@ -180,7 +180,7 @@ export class ObjectiveModel extends Model {
     return (
       await ObjectiveModel.query(tx)
         .findById(objectiveId)
-        .patch({progressLastMadeAt: new Date(Date.now())})
+        .patch({progressLastMadeAt: new Date()})
         .returning('*')
         .first()
     ).toObjective();
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -33,8 +33,8 @@ type ObjectiveStatus = 'pending' | 'approved' | 'rejected' | 'failed' | 'succeed
 type WalletObjective<O extends Objective> = O & {
   objectiveId: string;
   status: ObjectiveStatus;
-  createdAt: string;
-  progressLastMadeAt: string;
+  createdAt: Date;
+  progressLastMadeAt: Date;
 };
 
 export type DBOpenChannelObjective = WalletObjective<OpenChannel>;
@@ -91,8 +91,8 @@ export class ObjectiveModel extends Model {
   readonly status!: DBObjective['status'];
   readonly type!: DBObjective['type'];
   readonly data!: DBObjective['data'];
-  createdAt!: string;
-  progressLastMadeAt!: string;
+  createdAt!: Date;
+  progressLastMadeAt!: Date;
 
   static tableName = 'objectives';
   static get idColumn(): string[] {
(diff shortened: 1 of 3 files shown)


andrewgordstewart:
Is this PR ready for review?

lalexgap:
> Is this PR ready for review?

I can clean this up and get it ready for review.

andrewgordstewart:
> > Is this PR ready for review?
> 
> I can clean this up and get it ready for review.

Ok. I'm inferring from this that it's not ready for review, so I am converting it to a draft.

lalexgap:
> > > Is this PR ready for review?
> > 
> > 
> > I can clean this up and get it ready for review.
> 
> Ok. I'm inferring from this that it's not ready for review, so I am converting it to a draft.

The PR now has been updated and is ready for review.

andrewgordstewart:
> Objective timestamps are a mix of strings and `Date`s

It is the case that, as of this change, timestamps are _not_ a mix of strings and `Date`s, right? That is, timestamps are now `Date`s?

packages/server-wallet/src/models/objective.ts
```diff
@@ -91,8 +91,8 @@ export class ObjectiveModel extends Model {
   readonly status!: DBObjective['status'];
   readonly type!: DBObjective['type'];
   readonly data!: DBObjective['data'];
-  createdAt!: string;
-  progressLastMadeAt!: string;
+  createdAt!: Date;
```
andrewgordstewart, line 94: I'm a bit surprised about this change. The DB schema hasn't changed. This seems like it's only changing the _type_ of `createdAt`. This, combined with changing the initial value of `createdAt` from `new Date().toISOString()` to `new Date(Date.now())` _that gets inserted_, seems to have changed the type of `createdAt` _that is returned_ when a row is fetched.

Can you explain this?
- Does switching from `new Date().toISOString()` to `new Date(Date.now())` change the data that gets stored?
- Or, is the data that gets stored the same, but changing the type of `createdAt` to `Date` change the data that gets returned?

lalexgap: > > Objective timestamps are a mix of strings and `Date`s
> 
> It is the case that, as of this change, timestamps are _not_ a mix of strings and `Date`s, right? That is, timestamps are now `Date`s?

That's correct.

geoknee: @andrewgordstewart firstly, we don't need to pass `Date.now()` into the `Date` constructor, so I removed that in d192b2c.

You are correct to say the schema has not changed. This PR changes the data type passed into the query builder, and as a result the stored data has *not* changed, while the data returned *has* changed. This might seem surprising, but can be explained as follows:
1. During testing with the very same db schema, denoting the `createdAt` column as a timestamp, we can successfully insert either a `Date` *or* a `Date.toISOString()`. This implies that the type will be converted as necessary by knex before it gets to the DB driver.
2. Inside `ObjectiveModel.insert()` we return the result of the query *without* doing `.returning('*')`**. This means that you get out whatever you passed in. Previously this was a string but now it will be a `Date`, and this is bypassing any type conversion done by knex. 

So you pretty much had it right in your second bullet point. 

** we could do this, and maybe we should do this. It seems unnecessary because we construct the entire row before inserting it, so we "already have it". But, in future we might (say) add the timestamps in a `$beforeInsert` hook, and then we don't have the full row in scope at this point. There is also this question of datatypes that becomes harder to understand because the types can change when the return from the DB. So therefore 47ea69e.

tomclose (approved):

packages/server-wallet/src/models/objective.ts
andrewgordstewart: Thanks. A derivative of this explanation probably belongs in the repository, (eg. See [here](https://mtlynch.io/code-review-love/#4-answer-questions-with-the-code-itself)) since this is an important nuance that explains why the code behaves the way it does.

geoknee: The link you referenced suggests that the best solution in a case like this, is to refactor the code to eliminate confusion: which I think we effectively have already done.

In my opinion:

- the end state of the code works as you would expect
- we only needed an explanation here because the start state of the code was behaving in an unexpected way, which made the transition to the end state also quite confusing.
- the additions to the test reduce confusion by asserting on the return type
- 47ea69e adds an explanatory comment

I think this is sufficient, and I am not sure what else we could add? Happy to take suggestions, though.

packages/server-wallet/src/models/objective.ts
```diff
@@ -131,14 +131,17 @@ export class ObjectiveModel extends Model {
...
+        .insert({
+          objectiveId: id,
+          status: objectiveToBeStored.status,
+          type: objectiveToBeStored.type,
+          data: objectiveToBeStored.data,
+          createdAt: new Date(),
+          progressLastMadeAt: new Date(),
+        })
+        .returning('*')
+        .first(); // This ensures that the returned object undergoes any type conversion performed during insert
```
andrewgordstewart, line 144: Besides the fact that the comment is on L144, which is misleading, since it pertains to L143, I think it deserves more explanation:

```
        // `Model.query(tx).insert(o)` returns `o` by default.
        // In this case, `o.createdAt` might therefore be returned as a `string` or as a `Date`
        // But, when fetched by the DB, the `pg` driver will automatically convert the returned `createdAt`
        // value to `Date`.
        // Therefore, `returning('*')` ensures we return a value with type `DBObjective`, where `createdAt`
        // has type `Date`.
        .returning('*')
        .first();
```

There is some question of whether this comment belongs above L143, since this is not an isolated problem -- we use `returning('*')` in several places, perhaps for this reason.

There is another question around whether this is an acceptable solution, since it incurs extra networking: the data stored has to be sent both directions with the addition of `returning('*')`.

Two alternatives are:
1. `ObjectiveModel.insert` accepts a flexible type, where `createdAt` can be a `string` or a `Date`, and converts it to the expected type before returning it.
2. `ObjectiveModel.insert` accepts a strict type, where `createdAt` is a `Date`. This ensures that the returned value is predictable.

Further context: [Stack overflow](https://stackoverflow.com/questions/50716120/can-we-always-fetch-date-column-as-string-varchar-with-knex-and-postgres/50717046#50717046) points the "blame" at `node-postgres`. The relevant documentation is [here](https://node-postgres.com/features/types#date-timestamp-timestamptz).
> Likewise, when reading a date, timestamp, or timestamptz column value back into JavaScript, node-postgres will parse the value into an instance of a JavaScript Date object.

Another notable point:
> Although PostgreSQL supports microseconds in dates, JavaScript only supports dates to the millisecond precision. Keep this in mind when you send dates to and from PostgreSQL from node: your microseconds will be truncated when converting to a JavaScript date object even if they exist in the database. If you need to preserve them, I recommend using a custom type parser.

geoknee: I have put a longer comment into the code here #3284 .

On reflection, `returning('*')` is not a good solution for this issue, as I try and explain on #3284. (We still want it there but for other reasons).

I'm pretty sure we don't need to worry about microsecond precision for our use case!

andrewgordstewart: We may care about microsecond precision loss when comparing two timestamps for equality!

PR #3275, opened 2021-02-09
Add SyncObjective method to Wallet

lalexgap:
# Description
Add a new SyncObjective method to the wallet API.

This PR introduces a new method on the wallet `syncObjectives`. It accepts a list of `objectiveId`s and returns a message that can be sent to other counterparties. 

The syncObjectives method will also call `syncChannel` for any relevant channels so the message payload will contain all relevant channels for the objective.

The main use case this is designed around is for resending objectives if a message gets dropped.

## How Has This Been Tested? 

Two new test files have been added to test the functionality. One is a unit test, while the other is more of an integration style test with two wallets.

It would be nice to test this is out on an e2e style test however I think the amount of effort to do that is outside the scope of this PR.

### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/__test__/sync-objectives.test.ts
+++ b/packages/server-wallet/src/__test__/sync-objectives.test.ts
@@ -57,7 +57,7 @@ test('Objectives can be synced if a message is lost', async () => {
   // We would then call sync after some time of waiting and not making progress
   const syncResult = await a.syncObjectives([objectiveId]);
 
-  // We should now see the objective
+  // We should not see the objective yet
   expect(await getObjective(b.knex, objectiveId)).toBeUndefined();
 
   // After sync funding should continue as normal
@@ -66,6 +66,82 @@ test('Objectives can be synced if a message is lost', async () => {
     status: 'proposed',
     turnNum: 0,
   });
+});
+
+test('handles the objective being synced even if no message is lost', async () => {
+  const createChannelParams: CreateChannelParams = createChannelArgs();
+
+  // We mimic not receiving a message containing objectives
+  const messageResponse = await a.createChannel(createChannelParams);
+
+  const channelId = messageResponse.channelResults[0].channelId;
+  const objectiveId = `OpenChannel-${channelId}`;
+  await b.pushMessage(getPayloadFor(bob().participantId, messageResponse.outbox));
+
+  // We expect both objectives to be there
+  expect(await getObjective(a.knex, objectiveId)).toBeDefined();
+  expect(await getObjective(b.knex, objectiveId)).toBeDefined();
+
+  const {outbox, newObjectives, channelResults} = await a.syncObjectives([objectiveId]);
+
+  // The only result will be a message containing the latest states
+  expect(outbox).toHaveLength(1);
+  expect(outbox[0]).toMatchObject({
+    method: 'MessageQueued',
+    params: {
+      recipient: 'bob',
+      sender: 'alice',
+      data: {
+        signedStates: [expect.objectContaining({turnNum: 0})],
+        requests: [],
+      },
+    },
+  });
+  expect(newObjectives).toHaveLength(0);
+  expect(channelResults).toHaveLength(0);
+});
+
+test('Can successfully push the sync objective message multiple times', async () => {
+  const createChannelParams: CreateChannelParams = createChannelArgs();
+
+  // We mimic not receiving a message containing objectives
+  const messageToLose = await a.createChannel(createChannelParams);
+
+  const channelId = messageToLose.channelResults[0].channelId;
+  const objectiveId = `OpenChannel-${channelId}`;
+
+  // Only A should have the objective since we "lost" the message
+  expect(await getObjective(a.knex, objectiveId)).toBeDefined();
+  expect(await getObjective(b.knex, objectiveId)).toBeUndefined();
+
+  // We would then call sync after some time of waiting and not making progress
+  const syncResult = await a.syncObjectives([objectiveId]);
+
+  // We should not see the objective yet
+  expect(await getObjective(b.knex, objectiveId)).toBeUndefined();
+
+  // After sync funding should continue as normal
+  await b.pushMessage(getPayloadFor(bob().participantId, syncResult.outbox));
+
+  const {outbox, newObjectives, channelResults} = await b.pushMessage(
+    getPayloadFor(bob().participantId, syncResult.outbox)
+  );
+
+  // The only result will be a message containing the latest states
+  expect(outbox).toHaveLength(1);
+  expect(outbox[0]).toMatchObject({
+    method: 'MessageQueued',
+    params: {
+      recipient: 'bob',
+      sender: 'alice',
+      data: {
+        signedStates: [expect.objectContaining({turnNum: 0})],
+        requests: [],
+      },
+    },
+  });
+  expect(newObjectives).toHaveLength(0);
+  expect(channelResults).toHaveLength(0);
 });
 
 async function getObjective(knex: Knex, objectiveId: string): Promise<DBObjective | undefined> {
--- a/packages/server-wallet/src/__test__/sync-objectives.test.ts
+++ b/packages/server-wallet/src/__test__/sync-objectives.test.ts
@@ -1,0 +1,79 @@
+import {CreateChannelParams} from '@statechannels/client-api-schema';
+import Knex from 'knex';
+
+import {defaultTestConfig, overwriteConfigWithDatabaseConnection} from '../config';
+import {DBAdmin} from '../db-admin/db-admin';
+import {seedAlicesSigningWallet, seedBobsSigningWallet} from '../db/seeds/1_signing_wallet_seeds';
+import {DBObjective, ObjectiveModel} from '../models/objective';
+import {Wallet} from '../wallet';
+import {createChannelArgs} from '../wallet/__test__/fixtures/create-channel';
+import {bob} from '../wallet/__test__/fixtures/participants';
+
+import {getChannelResultFor, getPayloadFor} from './test-helpers';
+
+jest.setTimeout(10_000);
+const aWalletConfig = overwriteConfigWithDatabaseConnection(defaultTestConfig(), {
+  database: 'TEST_A',
+});
+const bWalletConfig = overwriteConfigWithDatabaseConnection(defaultTestConfig(), {
+  database: 'TEST_B',
+});
+
+let a: Wallet;
+let b: Wallet;
+
+beforeAll(async () => {
+  await Promise.all([DBAdmin.dropDatabase(aWalletConfig), DBAdmin.dropDatabase(bWalletConfig)]);
+  await DBAdmin.createDatabase(aWalletConfig);
+  await DBAdmin.createDatabase(bWalletConfig);
+  await Promise.all([
+    DBAdmin.migrateDatabase(aWalletConfig),
+    DBAdmin.migrateDatabase(bWalletConfig),
+  ]);
+  a = await Wallet.create(aWalletConfig);
+  b = await Wallet.create(bWalletConfig);
+
+  await seedAlicesSigningWallet(a.knex);
+  await seedBobsSigningWallet(b.knex);
+});
+afterAll(async () => {
+  await Promise.all([a.destroy(), b.destroy()]);
+  await Promise.all([DBAdmin.dropDatabase(aWalletConfig), DBAdmin.dropDatabase(bWalletConfig)]);
+});
+
+test('Objectives can be synced if a message is lost', async () => {
+  const createChannelParams: CreateChannelParams = createChannelArgs();
+
+  // We mimic not receiving a message containing objectives
+  const messageToLose = await a.createChannel(createChannelParams);
+
+  const channelId = messageToLose.channelResults[0].channelId;
+  const objectiveId = `OpenChannel-${channelId}`;
+
+  // Only A should have the objective since we "lost" the message
+  expect(await getObjective(a.knex, objectiveId)).toBeDefined();
+  expect(await getObjective(b.knex, objectiveId)).toBeUndefined();
+
+  // We would then call sync after some time of waiting and not making progress
+  const syncResult = await a.syncObjectives([objectiveId]);
+
+  // We should now see the objective
+  expect(await getObjective(b.knex, objectiveId)).toBeUndefined();
+
+  // After sync funding should continue as normal
+  const resultB0 = await b.pushMessage(getPayloadFor(bob().participantId, syncResult.outbox));
+  expect(getChannelResultFor(channelId, resultB0.channelResults)).toMatchObject({
+    status: 'proposed',
+    turnNum: 0,
+  });
+});
+
+async function getObjective(knex: Knex, objectiveId: string): Promise<DBObjective | undefined> {
+  const model = await ObjectiveModel.query(knex).findById(objectiveId);
+
+  if (model) {
+    return model.toObjective();
+  } else {
+    return undefined;
+  }
+}
--- a/packages/server-wallet/src/__test__/sync-objectives.test.ts
+++ b/packages/server-wallet/src/__test__/sync-objectives.test.ts
@@ -12,6 +12,7 @@ import {bob} from '../wallet/__test__/fixtures/participants';
 import {getChannelResultFor, getPayloadFor} from './test-helpers';
 
 jest.setTimeout(10_000);
+
 const aWalletConfig = overwriteConfigWithDatabaseConnection(defaultTestConfig(), {
   database: 'TEST_A',
 });
@@ -24,8 +25,8 @@ let b: Wallet;
 
 beforeAll(async () => {
   await Promise.all([DBAdmin.dropDatabase(aWalletConfig), DBAdmin.dropDatabase(bWalletConfig)]);
-  await DBAdmin.createDatabase(aWalletConfig);
-  await DBAdmin.createDatabase(bWalletConfig);
+  await Promise.all([DBAdmin.createDatabase(aWalletConfig), DBAdmin.createDatabase(bWalletConfig)]);
+
   await Promise.all([
     DBAdmin.migrateDatabase(aWalletConfig),
     DBAdmin.migrateDatabase(bWalletConfig),
@@ -36,6 +37,7 @@ beforeAll(async () => {
   await seedAlicesSigningWallet(a.knex);
   await seedBobsSigningWallet(b.knex);
 });
+
 afterAll(async () => {
   await Promise.all([a.destroy(), b.destroy()]);
   await Promise.all([DBAdmin.dropDatabase(aWalletConfig), DBAdmin.dropDatabase(bWalletConfig)]);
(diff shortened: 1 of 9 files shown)


packages/server-wallet/src/wallet/wallet-response.ts
```diff
@@ -86,7 +86,9 @@ export class WalletResponse {
     myIndex: number,
     participants: Participant[]
   ): void {
-    this.createdObjectives.push(objective);
+    // TODO: The objective model currently always sets the participants to an empty array when creating the DBObjective
```
lalexgap, line 89: Currently `toObjective` always sets the `participants` to an [empty array](https://github.com/statechannels/statechannels/blob/517ed467a5dbc542a89aee837669817a7ccafd3d/packages/server-wallet/src/models/objective.ts#L204).

andrewgordstewart:
> This PR introduces a new method on the wallet syncObjectives. It accepts a list of objectiveIds and returns a message that can be sent to other counterparties.
> 
> The main use case this is designed around is for resending objectives if a message gets dropped.

I think the scope of this function should be larger. If my channel manager (CM) expects an objective to make progress, and I call `syncObjective`, I should expect it to make progress.

I think the change here is a good start, pending a review.

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  database: 'TEST_B',
+});
+
+let a: Wallet;
+let b: Wallet;
+
+beforeAll(async () => {
+  await Promise.all([DBAdmin.dropDatabase(aWalletConfig), DBAdmin.dropDatabase(bWalletConfig)]);
+  await DBAdmin.createDatabase(aWalletConfig);
+  await DBAdmin.createDatabase(bWalletConfig);
```
andrewgordstewart, line 28: Is this copied from some other file? If so, seems like the wrong pattern to _create & drop_ some databases within multiple files.

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,79 @@
...
+afterAll(async () => {
+  await Promise.all([a.destroy(), b.destroy()]);
+  await Promise.all([DBAdmin.dropDatabase(aWalletConfig), DBAdmin.dropDatabase(bWalletConfig)]);
+});
+
+test('Objectives can be synced if a message is lost', async () => {
+  const createChannelParams: CreateChannelParams = createChannelArgs();
+
+  // We mimic not receiving a message containing objectives
+  const messageToLose = await a.createChannel(createChannelParams);
```
andrewgordstewart, line 48: [required] What's missing in this test suite is what happens when the message here is not lost, but `syncObjectives` is called.

It's possible that this is already resolved by some other test case. Is there a test that checks what happens when an objective is pushed multiple times?

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,79 @@
...
+});
+
+async function getObjective(knex: Knex, objectiveId: string): Promise<DBObjective | undefined> {
+  const model = await ObjectiveModel.query(knex).findById(objectiveId);
+
+  if (model) {
+    return model.toObjective();
+  } else {
+    return undefined;
+  }
```
andrewgordstewart, line 78: I think this works?
```suggestion
  return model?.toObjective();
```

packages/server-wallet/src/wallet/store.ts
```diff
@@ -280,6 +280,14 @@ export class Store {
     return {states: vars.map(ss => _.merge(ss, channelConstants)), channelState};
   }
 
+  async getObjectivesByIds(objectiveIds: string[]): Promise<DBObjective[]> {
+    return ObjectiveModel.forIds(objectiveIds, this.knex);
+  }
+
+  async getChannelsByIds(channelIds: string[]): Promise<ChannelState[]> {
```
andrewgordstewart, line 287: This function is added in this PR, but it seems unrelated to the rest of the PR, and it's not used elsewhere. Is that intentional?

packages/server-wallet/src/wallet/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      .withGraphFetched('funding');
+    const objective = await ObjectiveModel.insert(openChannelObjective(), testKnex);
+
+    const syncResult = await wallet.syncObjectives([objective.objectiveId]);
+    // TODO: Currently the objective model differs slightly from what we see on the message
+    const massagedObjective = {
+      ...objective,
+      participants: [aliceP(), bob()],
+    };
+    expect(syncResult.newObjectives).toEqual([massagedObjective]);
```
andrewgordstewart, line 52: Why should the objective be a `newObjective` after calling `syncObjective`? It's not a new objective, right?

packages/server-wallet/src/wallet/wallet-response.ts
andrewgordstewart: I think we need some consult to @geoknee here. I do not know the reason why `participants` is an empty array. What is the effect of adding `participants` to it?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -287,6 +287,38 @@ export class SingleThreadedWallet
...
+        {objectiveIds: missingObjectives},
+        'The following objectives cannot be found'
+      );
+      throw new Error('Could not find all objectives');
+    }
+
+    for (const o of objectives) {
+      const channel = await this.store.getChannelState(o.data.targetChannelId);
+      // This will make sure any relevant channel information is synced
+      await this._syncChannel(channel.channelId, response);
```
andrewgordstewart, line 313: 👍

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -287,6 +287,38 @@ export class SingleThreadedWallet
...
+
+      const {participants} = channel;
+      response.queueCreatedObjective(o, channel.myIndex, participants);
```
andrewgordstewart, line 316: I find this confusing, since the objective wasn't just created. It might even be _incorrect_, since an app might read `newObjectives`, which is read from `response.createdObjectives` and assume that they are new objectives.

packages/server-wallet/src/wallet/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    expect(syncResult.channelResults).toEqual([targetChannel.channelResult]);
+
+    expect((syncResult.outbox[0].params.data as any).objectives).toEqual([
+      expect.objectContaining(_.pick(objective, 'data')),
+    ]);
+  });
```
andrewgordstewart, line 58: There's no expectation in this test on the effect of `syncChannel`. I think there should be one, right?

andrewgordstewart (changes requested):
I'm not ready to approve this yet, because
1. I'm not yet sure what the effect of calling `syncObjective` is, when the message is not dropped
2. Marking objectives as `createdObjectives` after a `syncObjective` call seems incorrect.
3. The test should also check that channels get synced, rigth?

andrewgordstewart:
> > This PR introduces a new method on the wallet syncObjectives. It accepts a list of objectiveIds and returns a message that can be sent to other counterparties.
> > The main use case this is designed around is for resending objectives if a message gets dropped.
> 
> I think the scope of this function should be larger. If my channel manager (CM) expects an objective to make progress, and I call `syncObjective`, I should expect it to make progress.
> 
> I think the change here is a good start, pending a review.

I see now that in the implementation, `syncChannel` is also triggered. This could be listed in the description, as I believe your intent is to `squash & merge`, and that is probably important information to be in the commit message.

packages/server-wallet/src/wallet/__test__/sync-objectives.test.ts
lalexgap: I'm not sure I understand? The only responsibility of `syncChannel` is to generate a message that can be sent to the other participants. Which is why the assertions are on the `syncResult`

andrewgordstewart: But the assertions do not check for the messages that are generated by `syncChannel`. Or, am I misreading the tests?

packages/server-wallet/src/wallet/wallet.ts
lalexgap: I think this is due to my misunderstanding of the `newObjectives`.  I thought that this was communicated to other counterparties instead of the `app`. Oops!

packages/server-wallet/src/wallet/wallet-response.ts
lalexgap: While I think we should figure that out, this change is no longer needed for this PR and will be removed.

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: Yup, this was copied from [here](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/__test__/create-and-close-channel/direct-funding.test.ts)

packages/server-wallet/src/wallet/__test__/sync-objectives.test.ts
lalexgap: It does by checking the `(syncResult.outbox[0].params.data as any).objectives`. However, this could be more clear/thorough so I will clean this up!

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: This probably should be a category of tests with its own `jest.config` that setups and tears down these databases.

packages/server-wallet/src/wallet/store.ts
lalexgap: Fixed in 7b17570

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: Resolved in 360b5df

packages/server-wallet/src/wallet/__test__/sync-objectives.test.ts
andrewgordstewart: `syncChannels` generates other messages that are not in the `objectives` field. There are no assertions specifying that these messages are generated.

lalexgap: Ah sorry I misread the initial method in your comment 🤦

packages/server-wallet/src/wallet/wallet-response.ts
```diff
@@ -110,6 +99,21 @@ export class WalletResponse {
...
+   * - notifying the app
+   */
+  queueCreatedObjective(
+    objective: DBObjective,
+    myIndex: number,
+    participants: Participant[]
+  ): void {
+    this.createdObjectives.push(objective);
+
+    this.queueSendObjective(objective, myIndex, participants);
```
andrewgordstewart, line 114: Clever!

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -68,6 +68,82 @@ test('Objectives can be synced if a message is lost', async () => {
...
+      recipient: 'bob',
+      sender: 'alice',
+      data: {
+        signedStates: [expect.objectContaining({turnNum: 0})],
+        requests: [],
+      },
+    },
+  });
+  expect(newObjectives).toHaveLength(0);
+  expect(channelResults).toHaveLength(0);
```
andrewgordstewart, line 101: These expectations should have the same behaviour no matter if you `b.pushMessage` or not, right?. The interesting thing to test is _what happens when `b` receives the objective again_, right?

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -68,6 +68,82 @@ test('Objectives can be synced if a message is lost', async () => {
...
+  // The only result will be a message containing the latest states
+  expect(outbox).toHaveLength(1);
+  expect(outbox[0]).toMatchObject({
+    method: 'MessageQueued',
+    params: {
+      recipient: 'bob',
+      sender: 'alice',
+      data: {
+        signedStates: [expect.objectContaining({turnNum: 0})],
+        requests: [],
```
andrewgordstewart, line 96: ## Real comment
Thinking through this, why should `b.pushMessage` change the behaviour of `a.syncObjective` in this case? Did somehow `a` learn that `b` has the objective in this test case?

## Meta comment
It's a bit sad that to understand the point of this test, I need to understand that
1. `requests: []` is the important thing to focus on
2. `expect(['foo']).toMatchObject([]);` fails

packages/server-wallet/src/wallet/store.ts
```diff
@@ -284,10 +284,6 @@ export class Store {
     return ObjectiveModel.forIds(objectiveIds, this.knex);
   }
 
-  async getChannelsByIds(channelIds: string[]): Promise<ChannelState[]> {
-    return (await Channel.forIds(channelIds, this.knex)).map(c => c.protocolState);
```
andrewgordstewart: `Channel.forIds` was also introduced in this PR, and is now no longer used.

packages/server-wallet/src/wallet/wallet.ts
lalexgap: Fixed in d39fc70

packages/server-wallet/src/wallet/__test__/sync-objectives.test.ts
lalexgap: Fixed in d39fc70

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: Oops, you're correct. Will correct.

packages/server-wallet/src/wallet/__test__/sync-objectives.test.ts
lalexgap: This has been updated in 103e7e0. We now more thoroughly check the `outbox`

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: Resolved in e6a1605

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: Resolved in e6a1605

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: Updated in 44611f7

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  expect(getChannelResultFor(channelId, resultB0.channelResults)).toMatchObject({
+    status: 'proposed',
+    turnNum: 0,
+  });
+});
+
+test('handles the objective being synced even if no message is lost', async () => {
+  const createChannelParams: CreateChannelParams = createChannelArgs();
+
+  // We mimic not receiving a message containing objectives
```
andrewgordstewart, line 74: This comment seems to be incorrect in this context.

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      data: {
+        signedStates: [expect.objectContaining({turnNum: 0})],
+      },
+    },
+  });
+  expect(newObjectives).toHaveLength(0);
+  expect(channelResults).toHaveLength(1);
+});
+
+test('Can successfully push the sync objective message multiple times', async () => {
```
andrewgordstewart, line 109: This test describes pushing the sync objective message multiple times, but it seems like it was only pushed once. Am I missing something?

packages/server-wallet/src/models/channel.ts
```diff
@@ -152,7 +152,14 @@ export class Channel extends Model implements ChannelColumns {
   };
 
   static jsonAttributes = ['vars', 'participants', 'initialSupport'];
-
+  static async forIds(channelIds: Bytes32[], txOrKnex: TransactionOrKnex): Promise<Channel[]> {
```
andrewgordstewart, line 155: This method is added in this PR, but it is not used elsewhere. Is there a reason it's included here?

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: We push on line 129 and line 131 but I could add comments making that clearer.

andrewgordstewart: yes, my apologies, I missed that.

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  // We would then call sync after some time of waiting and not making progress
+  const syncResult = await a.syncObjectives([objectiveId]);
+
+  // We should not see the objective yet
+  expect(await getObjective(b.knex, objectiveId)).toBeUndefined();
+
+  // We should be able to push
+  await b.pushMessage(getPayloadFor(bob().participantId, syncResult.outbox));
+
+  const {outbox, newObjectives, channelResults} = await b.pushMessage(
```
andrewgordstewart, line 131:
```suggestion
  // We assert on the resulting values if we push a 2nd time.
  const {outbox, newObjectives, channelResults} = await b.pushMessage(
```

packages/server-wallet/src/models/channel.ts
lalexgap: resolved in efa6f61

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: I've added a similar comment in 5f20aa2

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: resolved in 5f20aa2

packages/server-wallet/src/__test__/sync-objectives.test.ts
```diff
@@ -0,0 +1,152 @@
+import {CreateChannelParams} from '@statechannels/client-api-schema';
```
andrewgordstewart, line 1: nit: this file ought to be called something a bit different -- otherwise, it's not obvious why it's separate from the test file included below.

I think there's a set of tests which play two wallets against each other. Perhaps they should be grouped under `__test__/with-peers/...` or something like that.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -288,6 +288,38 @@ export class SingleThreadedWallet
...
+    if (missingObjectives.length > 0) {
+      this.logger.error(
+        {objectiveIds: missingObjectives},
+        'The following objectives cannot be found'
+      );
+      throw new Error('Could not find all objectives');
+    }
+
+    for (const o of objectives) {
+      const channel = await this.store.getChannelState(o.data.targetChannelId);
```
andrewgordstewart, line 312: Do we really need to fetch the `channel` twice? (`_syncChannel` also fetches the channel.)

When I look at the implementation of `_syncChannel`, it appears difficult to pull this off.

andrewgordstewart (approved):
I'm glad this was simpler than it was thought to be!

I worry it's going to get more complex once wallets communicate whether they've accepted or rejected the objectives, but that's ok for now.

packages/server-wallet/src/__test__/sync-objectives.test.ts
lalexgap: I agree! I think that could be done under this issue: https://github.com/statechannels/statechannels/issues/3279

packages/server-wallet/src/wallet/wallet.ts
lalexgap: I think we could do this with some refactoring but I'm not sure it's worth the effort right now?

andrewgordstewart: Agreed.

I think it's worth marking this, either with a comment or an issue. I'll leave that up to you.

lalexgap: Added a comment in https://github.com/statechannels/statechannels/pull/3275/commits/8158ea764c92be74378850d52591b4d28fb627df

PR #3278, opened 2021-02-10
objectives have an "approved' boolean and a wider type of "status" property

geoknee:
See https://github.com/statechannels/statechannels/pull/3264#issuecomment-776796176

Change the type of an objective to better reflect its lifecycle. 

The *only* magic string that is used in critical code for `status` currently is the `approved` string. On this branch, this has been pinched off into a boolean property of the objective. 

`status` is preserved and now has a new use. The type of `status` is changed to `string`. This is intentionally very permissive, and no critical code will depend on it. It is only there to help track progress and aid debugging. The strings that will end up being stored will be defined by the implementation of an objective cranker. If this implementation changes, and the strings that get stored changed, this won't present a data migration issue, because the only way these strings are used is to track progress. As an example: we make a change from `status: 'waitingForFunding` to `status: waiting-for-funding'. If the wallet reads the status from some stored data before the change, it will recompute the new string and store that. The only bad thing that happens is that the status will have been deemed to have _changed_ (i.e some progress has been made on that objective) when in fact it probably hasn't. And the data has now effectively been migrated.

I considered renaming `status` to `state` but neglected to do so in order to keep the changeset small. It also means we don't need to add a migration.

 

> Approval is more critical,
>   and so has been separated into its own property.
>   'approved' is currently the only value of
>   'status' that is ever checked.
> 
> # Description
> **Title** Clear, single line description of the pull request, written as though it was an order.
> 
> Then, following a single blank line, a summary of the change and which issue is fixed, including 
> - [ ] what is the problem being solved
> - [ ] why this is a good approach
> - [ ] what are the shortcomings of this approach, if any
> 
> It may include
> - [ ] background information, such as "Fixes #"
> - [ ] benchmark results
> - [ ] links to design documents
> - [ ] dependencies required for this change
> 
> The summary may omit some of these details if this PR fixes a single ticket that includes these details. It is the reviewer's discretion. 
> 
> ## Changes [Optional] 
> Multiple changes are not recommended, so this section should normally be omitted. Unfortunately, they are sometimes unavoidable. If there are multiple logical changes, list them separately.
> 
> 1. `'foo'` is replaced by `'bar'`
> 2. `'fizzbuzz'` is optimized for gas
> 
> ## How Has This Been Tested? [Optional] 
> 
> Did you need to run manual tests to verify this change? If so, please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
> 
> ## :warning: Does this require multiple approvals? [Optional]
> Please explain which reason, if any, why this requires more than one approval.
> - [ ] Is it security related?
> - [ ] Is it a significant process change?
> - [ ] Is it a significant change to architectural, design?
> 
> ---
> ## Checklist:
> 
> ### Code quality
> - [ ] I have written clear commit messages
> - [ ] I have performed a self-review of my own code
> - [ ] This change does not have an unduly wide scope
> - [ ] I have separated logic changes from refactor changes (formatting, renames, etc.)
> - [ ] I have commented my code wherever necessary (can be 0)
> - [ ] I have added tests that prove my fix is effective or that my feature works, if necessary
> ### Project management
> - [ ] I have applied the [appropriate labels](https://www.notion.so/Team-working-agreements-2a95c926bb5642e5a5c42e4b74a9dd24#b304e56734a74dfbb341b8b4b27b1c0c)
> - [ ] I have linked to all relevant issues (can be 0)
> - [ ] I have added all dependent tickets (can be 0)
> - [ ] I have assigned myself to this PR
> - [ ] I have chosen the appropriate pipeline on zenhub
>

### DIFF
--- a/packages/server-wallet/e2e-test/e2e.test.ts
+++ b/packages/server-wallet/e2e-test/e2e.test.ts
@@ -99,7 +99,7 @@ describe('e2e', () => {
     expect(events).toContainObject({
       event: 'objectiveStarted',
       type: 'OpenChannel',
-      status: 'pending',
+      status: expect.any(String),
     });
     expect(events).toContainObject({
       event: 'objectiveSucceeded',
--- a/packages/server-wallet/src/db/migrations/20210126122746_add_objective_approved_column.ts
+++ b/packages/server-wallet/src/db/migrations/20210126122746_add_objective_approved_column.ts
@@ -1,0 +1,14 @@
+import * as Knex from 'knex';
+
+const tableName = 'objectives';
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(tableName, table => {
+    table.boolean('approved').defaultTo(false).notNullable();
+  });
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(tableName, table => {
+    table.dropColumn('approved');
+  });
+}
--- a/packages/server-wallet/src/models/__test__/objectives.test.ts
+++ b/packages/server-wallet/src/models/__test__/objectives.test.ts
@@ -24,14 +24,14 @@ beforeEach(async () => {
 describe('Objective > insert', () => {
   it('returns an objective with Date types for timestamps', async () => {
     await Channel.query(knex).withGraphFetched('signingWallet').insert(c);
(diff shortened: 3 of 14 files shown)


geoknee:
In favour of #3286

PR #3281, opened 2021-02-11
MultiThreadedWallet emits events

### LINKED ISSUE #3228: Wallet emits "OperationStarted" & "OperationSucceeded"
# Why: Problem description :confused:

The ChannelManager currently has trouble managing channels -- it has trouble handling process crashes and network errors.

Much more motivation can be found [here](https://www.notion.so/RCF-1-First-class-objectives-2a26247ce3244b8aafa4f9397df8465f#187917c1a0d54dd3bf9715f8415f65f2).

# What: Describe the solution you'd like :sparkles:  
1. When the wallet begins a "top-level" objective for a channel, it should emit a "OperationStarted" event.
2. When the wallet successfully completes a "top-level" objective for a channel, it should emit a `"OperationSucceeded"` event.
3. In addition, wallet API calls that create new objectives should include a handle of the newly created objectives.

# Describe alternatives you've considered :thinking:
See [the section from the RFC](https://www.notion.so/RCF-1-First-class-objectives-2a26247ce3244b8aafa4f9397df8465f#8512ecba8e234575a852d0028b29fb58).

**Additional context**
[RFC 1](https://www.notion.so/e2a4c4b6c68d4d7693667687c54b416b?v=20a84ab82c284e5a92863fdb000f9b4a&p=2a26247ce3244b8aafa4f9397df8465f).

### LINKED ISSUE #3262: Events are not emitted by MultiThreadedWallet
**Describe the bug**
The `MultiThreadedWallet` extends the `SingleThreadedWallet`, which extends `EventEmitter`. However, events emitted by worker threads are not emitted by the main thread, meaning that consumers of a `MultiThreadedWallet` cannot listen to them.

**To Reproduce**
The [stress test](https://app.circleci.com/pipelines/github/statechannels/graph-payments/98/workflows/46d12aa4-472e-460a-adcc-3571f4f21a57/jobs/652) fails on https://github.com/statechannels/graph-payments/commit/a6e6f009957e2df46226683f0c342124409013d5.

**Expected behavior**
The `MultiThreadedWallet` should have the same API implementation as the `SingleThreadedWallet` API.

andrewgordstewart:
# Description
Workers forward events emitted from their wallet, on to the main thread. The main thread re-emits the event from their wallet.

Fixes #3262.
Fixes #3228.

To do this, I needed to make a significant change to the MultiThreadedWallet:
- before, the WorkerManager would send an operation to the worker, listen for one `'message'` event, and _assume that the event contains the result of API call_
- after, the WorkerManager will send an operation to the worker, listen to `'message'` events _until the event contains the result of the API call_

This allows the worker to also forward wallet events to the main thread.

With @lalexgap, I considered changing the type of the value that's sent back by the worker. This was ruled out, as it would have required a large change to the SingleThreadedWallet. It also would have been awkward to deal with the special case `'objectiveStarted'` event emitted by `_createChannel`.

I also considered adding a `messageId` to the operation sent to the worker, so that we can listen for messages with the correct `messageId`. I judged this as overkill, as we are still restricting workers to API call at a time. We may wish to allow workers to handle concurrent API calls, depending on where the bottleneck lies, so this strategy may be employed in the future. This possibility is another reason why I chose this implementation.

To test this effectively, I had to make some changes to the PayerClient, to be able to configure it to use the MultiThreadedWallet.

I snuck in a bug fix where events were not properly emitted during `pushMessage`. This arose naturally while doing this work, which is a bit of scope creep. ⚠️ This will result in multiple `'objectiveStarted'` events being emitted by the wallet, exacerbated by apps using `syncObjectives` from #3275. I'd prefer to tackle this shortcoming separately.

## Changes [Optional] 
In addition to the main change outlined above,

1. Against my own recommendation, I included two additional bugfixes
   1. `pushMessage` trigers wallet events
   2. `MultiThreadedWallet` throws an error when passed an invalid configuration
2. PayerClient can now be configured with a partial wallet config
3. tests are added for the MultiThreadedWallet in two places

1. The `e2e` test now runs against both a single- and multi-threaded wallet
2. A test was added that checks that events are emitted during `pushMessage`. This test runs against both single- and multi-threaded wallets

- [ ] Is it security related?
- [x] Is it a significant change to architectural, design?

This is a significant change to the MultiThreadedWallet. Considering there were previously no tests of the MultiThreadedWallet, it is probably worthwhile to get two reviewers.

---
### Code quality
       * explained above

### DIFF
--- a/packages/server-wallet/e2e-test/e2e.test.ts
+++ b/packages/server-wallet/e2e-test/e2e.test.ts
@@ -68,6 +68,7 @@ describe.each([0, 2])('e2e with %i threads', workerThreadAmount => {
     // Create actors
     payerClient = await PayerClient.create(alice().privateKey, `http://127.0.0.1:65535`, {
       workerThreadAmount,
+      loggingConfiguration: {logDestination: '/tmp/server-wallet.e2e-test.log', logLevel: 'trace'},
     });
 
     // Gets participant info for testing convenience
--- a/packages/server-wallet/e2e-test/e2e.test.ts
+++ b/packages/server-wallet/e2e-test/e2e.test.ts
@@ -58,7 +58,7 @@ afterAll(async () => {
   await knexReceiver.destroy();
 });
 
-describe('e2e', () => {
+describe.each([0, 2])('e2e with %i threads', workerThreadAmount => {
   let payerClient: PayerClient;
 
   let payer: Participant;
@@ -66,7 +66,9 @@ describe('e2e', () => {
 
   beforeAll(async () => {
     // Create actors
-    payerClient = await PayerClient.create(alice().privateKey, `http://127.0.0.1:65535`);
+    payerClient = await PayerClient.create(alice().privateKey, `http://127.0.0.1:65535`, {
+      workerThreadAmount,
+    });
 
     // Gets participant info for testing convenience
     payer = payerClient.me;
--- a/packages/server-wallet/e2e-test/e2e.test.ts
+++ b/packages/server-wallet/e2e-test/e2e.test.ts
@@ -58,7 +58,7 @@ afterAll(async () => {
   await knexReceiver.destroy();
 });
 
-describe.each([0, 2])('e2e with %i threads', workerThreadAmount => {
+describe.each([0, 2])('e2e with %i worker threads', workerThreadAmount => {
   let payerClient: PayerClient;
 
   let payer: Participant;
--- a/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
@@ -31,7 +31,7 @@ import {DBAdmin} from '../../../db-admin/db-admin';
 import {LedgerRequest} from '../../../models/ledger-request';
 import {WALLET_VERSION} from '../../../version';
 import {PushMessageError} from '../../../errors/wallet-error';
-import {Wallet} from '../..';
+import {MultiThreadedWallet, Wallet} from '../..';
 import {
   getChannelResultFor,
   getSignedStateFor,
@@ -44,14 +44,20 @@ const dropNonVariables = (s: SignedState): any =>
 jest.setTimeout(20_000);
 
 let wallet: Wallet;
+let multiThreadedWallet: MultiThreadedWallet;
 
 beforeAll(async () => {
   await DBAdmin.migrateDatabase(defaultTestConfig());
   wallet = await Wallet.create(defaultTestConfig());
+
+  multiThreadedWallet = await MultiThreadedWallet.create(
+    defaultTestConfig({workerThreadAmount: 2})
+  );
 });
 
 afterAll(async () => {
   await wallet.destroy();
+  await multiThreadedWallet.destroy();
 });
 beforeEach(async () => seedAlicesSigningWallet(wallet.knex));
 
@@ -318,23 +324,23 @@ describe('when the application protocol returns an action', () => {
     });
   });
 
-  it.each([0, 2])(
-    'emits objectiveStarted when a wallet with %i worker threads receives a new objective',
-    async workerThreadAmount => {
-      wallet = await Wallet.create(defaultTestConfig({workerThreadAmount}));
+  it.each(['with', 'without'] as const)(
+    'emits objectiveStarted when a wallet %s worker threads receives a new objective',
+    async withOrWithout => {
+      const _wallet: Wallet = withOrWithout === 'with' ? wallet : multiThreadedWallet;
       const turnNum = 6;
       const state = stateSignedBy()({outcome: simpleEthAllocation([]), turnNum});
 
       const c = channel({vars: [addHash(state)]});
-      await Channel.query(wallet.knex).insert(c);
+      await Channel.query(_wallet.knex).insert(c);
 
       const {channelId} = c;
       const finalState = {...state, isFinal: true, turnNum: turnNum + 1};
 
       const callback = jest.fn();
-      wallet.once('objectiveStarted', callback);
+      _wallet.once('objectiveStarted', callback);
 
-      const result = await wallet.pushMessage({
+      const result = await _wallet.pushMessage({
         walletVersion: WALLET_VERSION,
         signedStates: [serializeState(stateSignedBy([bob()])(finalState))],
         objectives: [
--- a/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
@@ -319,7 +319,7 @@ describe('when the application protocol returns an action', () => {
   });
 
   it.each([0, 2])(
-    'emits objectiveStarted when a %i-threaded wallet receives a new objective',
+    'emits objectiveStarted when a wallet with %i worker threads receives a new objective',
     async workerThreadAmount => {
       wallet = await Wallet.create(defaultTestConfig({workerThreadAmount}));
       const turnNum = 6;
--- a/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
@@ -318,6 +318,40 @@ describe('when the application protocol returns an action', () => {
     });
   });
 
+  it.each([0, 2])(
+    'emits objectiveStarted when a %i-threaded wallet receives a new objective',
+    async () => {
+      const turnNum = 6;
+      const state = stateSignedBy()({outcome: simpleEthAllocation([]), turnNum});
+
+      const c = channel({vars: [addHash(state)]});
+      await Channel.query(wallet.knex).insert(c);
+
+      const {channelId} = c;
+      const finalState = {...state, isFinal: true, turnNum: turnNum + 1};
+
+      const callback = jest.fn();
+      wallet.once('objectiveStarted', callback);
+
+      const p = wallet.pushMessage({
+        walletVersion: WALLET_VERSION,
+        signedStates: [serializeState(stateSignedBy([bob()])(finalState))],
+        objectives: [
+          {
+            type: 'CloseChannel',
+            participants: [],
+            data: {
+              targetChannelId: channelId,
+              fundingStrategy: 'Direct',
+            },
+          },
+        ],
+      });
+      await expect(p.then(r => r.newObjectives)).resolves.toHaveLength(1);
+      expect(callback).toHaveBeenCalledWith(expect.objectContaining({type: 'CloseChannel'}));
+    }
+  );
+
   it('forms a conclusion proof when the peer wishes to close the channel', async () => {
     const turnNum = 6;
     const state = stateSignedBy()({outcome: simpleEthAllocation([]), turnNum});
@@ -326,8 +360,8 @@ describe('when the application protocol returns an action', () => {
     await Channel.query(wallet.knex).insert(c);
 
     const {channelId} = c;
-
     const finalState = {...state, isFinal: true, turnNum: turnNum + 1};
+
     const p = wallet.pushMessage({
       walletVersion: WALLET_VERSION,
       signedStates: [serializeState(stateSignedBy([bob()])(finalState))],
--- a/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
+++ b/packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
@@ -333,7 +333,7 @@ describe('when the application protocol returns an action', () => {
       const callback = jest.fn();
       wallet.once('objectiveStarted', callback);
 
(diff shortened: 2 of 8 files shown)


packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
```diff
@@ -20,31 +20,34 @@ export class WorkerManager {
   private pool?: Pool<Worker>;
   private threadAmount: number;
   private logger: Logger;
-  constructor(walletConfig: ServerWalletConfig) {
+  constructor(walletConfig: ServerWalletConfig, onNewWorker: (worker: Worker) => void) {
```
lalexgap, line 23: nit but it might be nice to add some tsdocs here

packages/server-wallet/src/wallet/multi-threaded-wallet/worker.ts
```diff
@@ -25,7 +33,11 @@ async function startWorker() {
   const logger = createLogger(walletConfig);
 
   logger.debug(`Worker %o starting`, threadId);
-  const wallet = await Wallet.create(walletConfig);
+  const wallet = (await Wallet.create(walletConfig)) as SingleThreadedWallet;
```
lalexgap, line 36: Could we just call `SingleThreadedWallet.create` instead? That way we avoid the cast.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -334,6 +335,8 @@ export class Store {
       return {
         channelIds: stateChannelIds.concat(objectiveChannelIds),
         channelResults,
+        // HACK: This will possibly cause the wallet to re-emit `'objectiveStarted'` multiple times
```
lalexgap, line 338: This probably should be an issue? (If not one already)

packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
```diff
@@ -318,6 +318,40 @@ describe('when the application protocol returns an action', () => {
...
+            type: 'CloseChannel',
+            participants: [],
+            data: {
+              targetChannelId: channelId,
+              fundingStrategy: 'Direct',
+            },
+          },
+        ],
+      });
+      await expect(p.then(r => r.newObjectives)).resolves.toHaveLength(1);
```
lalexgap, line 350: Is there a reason we don't just `await` `wallet.pushMessage` in the line above? I think it would make the assertion cleaner

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
```diff
@@ -57,6 +60,14 @@ export class WorkerManager {
     });
   }
 
+  /**
+   *
+   * @param {operation, args}: operation & arguments to send to worker
+   *
+   * Sends {operation, args} to a worker thread.
+   * RESOLVES WHEN: the worker thread sends a "right" message containing a result
```
lalexgap, line 68: The resolve/reject info might be good to include in a `@returns` tag?

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
```diff
@@ -67,17 +78,23 @@ export class WorkerManager {
...
-      worker.once('message', (response: Either<Error, T>) => {
+    const resultPromise = new Promise<T>((resolve, reject) => {
+      const handler = (response: Either<Error, T>) => {
         this.pool?.release(worker);
 
         if (isLeft(response)) {
           reject(response.left);
-        } else {
+        } else if (isRight(response)) {
+          // WARNING: The correctness of this implementation relies on the fact that a worker
```
lalexgap, line 88: It might be worth putting a similar warning in `worker.ts` where we construct the `SingleThreadedWallet`. That way if someone tries changing it to a `MultiThreadedWallet` they see the warning as well.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -832,8 +838,12 @@ export class SingleThreadedWallet
     const {
       channelIds: channelIdsFromStates,
       channelResults: fromStoring,
+      storedObjectives,
     } = await this.store.pushMessage(wirePayload);
 
+    // HACK: This will possibly re-emit `'objectiveStarted'` multiple times
```
lalexgap, line 844: As above, it'd be nice to have an issue.

lalexgap (approved):
Looks good to me!

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
andrewgordstewart: Is that the expected use of `@returns`? I am unfamiliar with TSDoc.

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: I was waiting until this PR got accepted.

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
andrewgordstewart: I don't actually think there's a worry in `worker.ts`. A worker receives an operation and sends back the result. It's up to the manager to determine that the result message corresponded to the API call.

> That way if someone tries changing it to a MultiThreadedWallet they see the warning as well.

I don't quite understand: you mean if someone changed `SingleThreadedWallet` to `MultiThreadedWallet` in `worker.ts`? That would be a pretty flawed decision that the `worker_threads` API might not even allow -- can worker thread spawn worker threads?. I don't know that we want to protect against that with a comment?

packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
andrewgordstewart: 6a6a7c9da2d42b8e35d78333ec6d5fa2a9572316

packages/server-wallet/src/wallet/multi-threaded-wallet/worker.ts
andrewgordstewart: c56aade0bd3c9fe39640d5fb10e42d4ff5469f30

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
andrewgordstewart: 62efe6c54c7af52228fdde6ca801898f697af016

tomclose:
Slightly related: one of the proposals I wrote last week was about wallet architecture ([link](https://www.notion.so/Architecture-9eb315eee12b48a7a20adbff2f5f2cd7)), and would impact this work. The proposal is to split the majority of the wallet logic out into an engine, that is strictly request-response, and have a small "comms" wrapper that emits events:

Screen_Shot_2021-02-03_at_2 13 31_PM

I believe that if we did this it would make this change unnecessary, as we could set it up so that the comms code only runs in the main thread, and there would be no need for workers to submit events. I feel like this would probably be a less complex and more robust solution.

andrewgordstewart:
> Slightly related: one of the proposals I wrote last week was about wallet architecture ([link](https://www.notion.so/Architecture-9eb315eee12b48a7a20adbff2f5f2cd7)), and would impact this work. The proposal is to split the majority of the wallet logic out into an engine, that is strictly request-response, and have a small "comms" wrapper that emits events:
> 
> Screen_Shot_2021-02-03_at_2 13 31_PM
> 
> I believe that if we did this it would make this change unnecessary, as we could set it up so that the comms code only runs in the main thread, and there would be no need for workers to submit events. I feel like this would probably be a less complex and more robust solution.

While I'm not opposed to that, that work has not yet been planned, while this PR is resolving a high priority issue that blocks current sprint goals. I think the functional changes in this PR are easy to reverse, should we adopt your architecture.

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
lalexgap: `@returns` specifies what a function returns. 

So we could have something like `* @returns a promise that resolves when ....`

I don't think it really matters too much though! I'm happy that this information is in TSDoc in some manner.

andrewgordstewart: I'll switch it to `@returns`. I think TSDoc/we can do better, with a `@resolves` keyword, but I don't know how feasible that it.

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
lalexgap: My thinking was that it might be easy to miss the warning in `sendOperation`?  But that might be overkill.

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
lalexgap: There is some discussion about that here https://github.com/Microsoft/tsdoc/issues/63

packages/server-wallet/src/wallet/store.ts
geoknee: It's not obvious to me why this will cause the event to be emitted multiple times? Can you explain?

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
```diff
@@ -67,17 +84,23 @@ export class WorkerManager {
...
+          // only handles one concurrent operation.
+          // If a worker handles concurrent operations, then we should wait until we get a reply
+          // with the same message id.
+          worker.off('message', handler);
```
geoknee, line 98: I am trying to map the change here to 

> listen to 'message' events until the event contains the result of the API call

from the PR description. Is that the right mapping? If so, the only check is the `isRight()` check.

If left and right are mutually exclusive, I think this change wouldn't add any new functionality.

I think I would understand this,  if the response is being considered as left (error) , right (response), and neither left nor right (which you are using for relaying events). Am I in the right ballpark?

andrewgordstewart: > I think I would understand this, if the response is being considered as left (error) , right (response), and neither left nor right (which you are using for relaying events). Am I in the right ballpark?

Yes, that's correct. Perhaps the types could be "cleaned up", but the alternative I thought of was messy:
```
type WorkerWalletEvent = {
  type: 'WalletEvent';
  value: WalletEvent;
}

type APIResult = {
  type: 'APIResult';
  value: Either<Error, SingleOrMultipleChannelResult>
}

type WorkerMessage = WorkerWalletEvent | WorkerAPIResult
```

There is now an onion-like type.

Maybe an appealing alternative is
```
// snip
type WorkerMessage = WorkerWalletEvent | APIError | APIResult
```

which is, funny enough, the type in the codebase, if we have
```
type APIError = Left<Error>
type APIResult = Right<SingleOrMultipleChannelResult>
```

packages/server-wallet/e2e-test/e2e.test.ts
```diff
@@ -58,15 +58,18 @@ afterAll(async () => {
   await knexReceiver.destroy();
 });
 
-describe('e2e', () => {
+describe.each([0, 2])('e2e with %i threads', workerThreadAmount => {
```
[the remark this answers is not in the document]
andrewgordstewart: 7b64136479a81ef41b6bb5acc55c2dd3a3cb7c3c

packages/server-wallet/src/wallet/__test__/integration/push-message.test.ts
```diff
@@ -318,6 +318,40 @@ describe('when the application protocol returns an action', () => {
     });
   });
 
+  it.each([0, 2])(
+    'emits objectiveStarted when a %i-threaded wallet receives a new objective',
+    async () => {
```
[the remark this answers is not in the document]
andrewgordstewart: > Also nit: `'0-threaded wallet'` doesn't sound quite right?

7b64136479a81ef41b6bb5acc55c2dd3a3cb7c3c

> Unless I am missing something, you are not accepting a workerThreadAmount variable in the callback, so your test can't do anything differently each time it is run.

fcbd171fb8c4d85aaa909d46cb766860e07498d9

packages/server-wallet/src/wallet/multi-threaded-wallet/manager.ts
andrewgordstewart: b9fd23069f2cd6893d27b68a1aefe7225a457095

geoknee (approved):

packages/server-wallet/src/wallet/store.ts
andrewgordstewart: > It's not obvious to me why this will cause the event to be emitted multiple times? Can you explain?

48443417b610cf2a792510a91b3c78d450daf712

> This probably should be an issue? (If not one already)

https://github.com/statechannels/statechannels/issues/3289

PR #3285, opened 2021-02-11
refactor: rename getters and add comments

geoknee:
This is a genuine refactor and does not change any behaviour. I don't think we have any bugs around this at present, but I want to reduce the chance we get one in future. 

This PR aims to be a little more precise on terminology, and to reduce the chance of confusion between a complete setup round and a supported setup state. 

In the prefund stage, the two concepts are the same. But I have a preference for using the "complete" terminology, still, as it seems simpler to me to match the terminology for the postfund round, which I think should change.

In the postfund stage, participant `0` can make state `n` supported as soon as they sign it: they get the remaining `n-1` signatures from the prefund stage. I think it is important that wallets do not have their `OpenChannel` objective succeed, or to have the `Channel` deemed to be running, if they only have (say) 1 postfund signature. They should wait until all postfund signatures have been received, and not just until a postfund state has been supported.

[Relevant conversation on the forum. ](https://statechannels.discourse.group/t/do-we-really-need-the-post-fund-setup-round/76/5)

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -406,21 +406,22 @@ export class Channel extends Model implements ChannelColumns {
   }
 
   /**
-   * Have all participants signed a preFund state?
+   * Have all participants signed a consistent preFund state?
    */
   public get preFundComplete(): boolean {
     // The existence of any supported state implies a supported prefund state
-    // which implies a full set of signatures (since no state is earlier than a prefund state)
+    // which implies a full set (nParticipants) of consistent preFund signatures has been seen
+    // (since no state is earlier than a prefund state).
     return !!this.supported;
   }
 
   /**
-   * Have all participants signed a postFund state?
+   * Have all participants signed a consistent postFund state?
    */
   public get postFundComplete(): boolean {
     // The existence of a supported state with sufficiently high turnNum
     // (at least the highest postFund state)
-    // implies a full set of nParticipants postFund signatures have been seen.
+    // implies a full set (nParticipants) of consistent postFund signatures has been seen.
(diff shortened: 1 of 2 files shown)


packages/server-wallet/src/models/channel.ts
```diff
@@ -406,17 +406,21 @@ export class Channel extends Model implements ChannelColumns {
   }
 
   /**
-   * Is a prefund state (or later) supported
+   * Have all participants signed a preFund state?
```
tomclose, line 409: It's not exactly this - e.g. we don't want them to have signed conflicting preFunds. Maybe this should just read "is the preFund setup stage complete".

packages/server-wallet/src/models/channel.ts
```diff
@@ -406,17 +406,21 @@ export class Channel extends Model implements ChannelColumns {
...
-    // all states are later than the prefund, so we just check if have any supported state
+  public get preFundComplete(): boolean {
+    // The existence of any supported state implies a supported prefund state
+    // which implies a full set of signatures (since no state is earlier than a prefund state)
     return !!this.supported;
   }
 
   /**
-   * Is a postfund state (or later) supported
+   * Have all participants signed a postFund state?
```
tomclose, line 418: Same comment as above.

tomclose (approved):

packages/server-wallet/src/models/channel.ts
geoknee: Good point. I have inserted the word "consistent" in cd14064 to try and justice to that. 

I think it's nice to have the docstring say more than is already said by the name of the function.

Issue #3231, opened 2021-01-29
Scope out named "states" for objectives/operations, to aid with debugging (and more?)


andrewgordstewart:
Currently, it is difficult to gauge the progress of an objective. This makes it frustrating to debug flickering tests.

andrewgordstewart:
Related to https://github.com/statechannels/project-management/issues/44

andrewgordstewart:
This might be closed?

PR #3286, opened 2021-02-11
introduce waitingFor column on objectives table

geoknee:
⚠️ This PR is the lowest in a stack. Please do not squash- or rebase-merge it. The next PR up in the stack is #3301 .

We are building a mechanism to detect a stalled objective, and have the `progressLastMadeAt` timestamp to help with this. I have played around with a few different approaches for when to hook code into updating that timestamp #3264 #3278. In this PR, I have settled on something that makes good sense to me, with some input from @tomclose . 

To properly detect when progress has been made towards an objective when the run loop is triggered, we are going to need some sort of record or memory of state. In defining that state, we don't want to have to query and inspect all channels in scope of the objective; but we want to introduce the minimal amount of state replication, since the store should remain the source of truth. 

The idea of this PR is to concentrate on progress being *blocked* by events that my wallet has non control over. It doesn't make sense, for example, to report that we are blocked on signing our own pre fund state... we should just go ahead and do that, and anything else we are safe to do, until we reach a point where we have to wait for someone or something else. So these are the points where we should have state boundaries. 

The "states" of an objective are then just names for the next thing that blocks us from making progress (and that we have no direct control over). These are cached on the objective itself, and used to detect when progress is made. By storing this information in a column called `waitingFor` (rather than `state`, say), I am trying to discourage reading this name for the purposes of controlling protocol logic itself: it is just metadata for capturing progress. 

This PR introduces the required changes to the DB schema and ORM, along with a simple test and a single example in the `ChannelOpener.crank()` function. What's nice is that each time the cranker can infer that the objective is externally blocked, it does an early `return`. If we don't do the early `return`, immediately after we cross a state boundary and cache a new `waitingFor` string.  If, when we have finished cranking, the `waitingFor` string changed, we know that we made meaningful progress (and not just the boring progress we can make unilaterally). 

If we like this pattern, I think it should be relatively easy to roll out to the rest of the objective crankers.    

A conjecture:
_The progress of an objective can be linear, even if the events that drive that progress arrive in an unpredictable order__                                                   
                                                         

```                                                         
 first   second                                          
 crank   crank                    |                    | 
    |    |                        |                    | 
    |    |     waitingFor first   | waitingFor second  | 
    |    |     blocker            | blocker            | 
    |    |                        |                    | 
    |    |          +---------+   |                    | 
    +----|--------->| BLOCKED |   |                    | 
         |          +---------+   |                    | 
         |                |       |     +---------+    | 
         +------------------------|-----> BLOCKED |    |  etc
                          |       |     +---------+    | 
                          |       |         |          | 
                          v       |         v          | 
                        wait      |       wait         | 
                                  |                    |
```

## Changes 

1. new column in objectives table
2. new static method `updateWaitingFor` , which updates the `progressLastMadeAt` timestamp 
  - this could probably become an instance method
  - we could consider only updating the timestamp in the event of a nontrivial db patch
 3. `ChannelOpener.crank()` rewritten both for clarity and to adopt the new pattern of recording what is currently blocking the objective. 

## How Has This Been Tested? 

I manually verified (with console.log) that each blocker was logged out in turn during our `open-and-close-channel` tests.

---
### Code quality
### Project management
----
EDIT
Have asked @andrewgordstewart and @lalexgap both for a careful review of this changeset, and of the stacked changes #3301, and to pay particular attention to 

- any migrations that won't run in production
- any extra unnecessary work that might be generated by adopting this approach


lalexgap:
Another possible approach I was exploring was [ wrapping the crank in a transaction](https://github.com/statechannels/statechannels/pull/3131) and letting the objective manager determine if progress has been made by inspecting the transaction after crank and seeing if there are any writes.

Unfortunately, I didn't see any obvious way of checking `Knex.Transaction` for a write query :(

andrewgordstewart:
> Unfortunately, I didn't see any obvious way of checking `Knex.Transaction` for a write query :(

Is this a shortcoming of knex, or Postgres?

geoknee:
Relevant to #3231

geoknee:
> Another possible approach I was exploring was [ wrapping the crank in a transaction](https://github.com/statechannels/statechannels/pull/3131) and letting the objective manager determine if progress has been made by inspecting the transaction after crank and seeing if there are any writes.
> 
> Unfortunately, I didn't see any obvious way of checking `Knex.Transaction` for a write query :(

I think this is an interesting suggestion, but it has some drawbacks other than just "we don't quite know how to do it yet".

1. You'll be "making progress" when writing to the store inside the crank
2. You won't be "making progress" when writing to the store *before* the crank. 

So signing my own state would count as progress, but receiving my counterparty's state would not count as progress. I think this is the wrong way around. 

Also

3. You don't get any extra contextual information about what progress was made or what is now blocking.

packages/server-wallet/src/__test__/create-and-close-channel/direct-funding.test.ts
```diff
@@ -93,6 +93,10 @@ it('Create a directly funded channel between two wallets ', async () => {
     turnNum: 0,
   });
 
+  // Let's check what wallet a is waiting for. Expect it to be 'theirPreFundSetup' or ''
+  // (if the objective has never been cranked)
```
geoknee, line 97: This is a bit unsatisfactory, but I think easily fixable. When we open a channel via an API call, we do not trigger a crank, so the `waitingFor` doesn't get updated, and takes the current default of `''`. I think we should either 
- trigger the crank (it should be cheap and return early)
- manually patch the `waitingFor` or just do that when we insert the objective. 
- choose a better initial value for the `waitingFor` string.

andrewgordstewart: > manually patch the waitingFor or just do that when we insert the objective.

to me, this seems like the natural choice.
- the peer who creates the objective knows what they'll be waiting for
- other peers will often be waiting for `'toBeAccepted'` or something like that. (I don't know what the default should be for auto-accepted objectives.)

andrewgordstewart:
I don't really understand
> To properly detect when progress has been made towards an objective when the run loop is triggered, we are going to need some sort of record or memory of state. In defining that state, we don't want to have to query and inspect all channels in scope of the objective; but we want to introduce the minimal amount of state replication, since the store should remain the source of truth.

There are legitimate alternatives, even ones that have been discussed. Why is this solution preferred over those alternatives?

# Alternatives

These are three approaches that the app could take to make progress on stalled objectives.

## 1
A simple, alternative solution to the problem this change is solving is:
1. When my wallet takes an action for an objective, it records the current time.
2. My app periodically checks the last time my wallet took an action. If it is too far in the past, it calls `syncObjective`. After a few attempts, my app asks my wallet to abort the objective.

### How to handle restarts
When my app restarts, it triggers (2) for any objectives that are `'approved'`.

## 2
An even simpler alternative, 
1. When I accept the objective, my wallet records the current time.
2. My app periodically checks if the objective is complete. If not, it calls `syncObjective`. After a few attempts, my app asks my wallet to abort the objective.

### How to handle restarts
When my app restarts, it triggers (2) for any objectives that are `'approved'`.

## 3
This approach requires no additional timestamp.
1. When my app learns that an objective has started, eg. via the `'objectiveStarted'` event, my app periodically checks if the objective is complete. If not, it calls `syncObjective`. After a few attempts, my app asks my wallet to abort the objective.

### How to handle restarts
Factor
> [my app] periodically checks if the objective is complete. If not, it calls `syncObjective`. After a few attempts, my app asks my wallet to abort the objective.

into a generic bit of code. When my app restarts, it triggers this generic bit of code for any objectives that are `'approved'`.

I would probably choose Alternative 2, as it is the simplest approach that solves the problems outline in https://github.com/statechannels/statechannels/issues/3230. It seems as if alternative (2) is already implemented in the codebase, meaning we can potentially close #3230 and immediately being working on 2.2.

#Why shouldn't we use this approach?

The reason I am hesitant about the proposed approach here is:
- This change is a significant alteration to the current approach. Thus, it would have benefited from a discussion around a design doc.
- A fruitful discussion might be difficult, since the set of developers has just recently become familiar with the current wallet approach. I think it is important for a consistent approach to first be codified. That is the intention of https://github.com/statechannels/project-management/issues/28
- Thus, I am concerned that the "blessed design" is moving ahead yet again from the overall developer's understanding of the wallet. (It has certainly moved ahead of my understanding.)
- Moreover, this change feels like it can easily spur on a cascade of DB migrations in the near future. We do not have a well-formed plan for running migrations in production. So, we are adding short-term complexity that appears unnecessary.

As discussed 1-1, I am enthusiastic to adopt your approach. But, I am hesitant to say that this is the correct time to implement it.

packages/server-wallet/src/protocols/channel-opener.ts
```diff
@@ -38,34 +38,61 @@ export class ChannelOpener {
...
+      // otherwise, we might be waiting on funding
+      waitingFor = 'funding';
+
       // if we have a full pre fund setup, delegate to the funding process to (a) see if
       // the channel is funded, and (b) take action if not
       const funded = await this.crankChannelFunder(objective, channel, response, tx);
 
+      // if the channel is not funded, we are still waitingFor funding and cannot progress
+      if (!funded) {
+        await updateWaitingFor();
```
andrewgordstewart, line 67: According to my understanding, I don't think this pattern will work well in practice.

It seems like, when we hit L67, we will `await updateWaitingFor()` every time we call `crank`. That will update the timestamp to now. IE. we could repeatedly call `crank`, and update the timestamp to `now()`, without the state changing.

This means that the timestamp is not actually tracking the last-known-state-change.

geoknee: If you look at the implementation of `updateWaitingFor()`, you'll notice that it calls the function that will update the timestamp *only if* `waitingFor` has changed. 

The outer function should probably be named `updateWaitingForIfChanged()`.

geoknee:
I think your point about this being not-the-very-simplest solution is fair. However, I think on balance the complexity introduced is small, and achieves positive side effects such as “increasing the conceptual integrity” of objectives (a sprint goal) as well as tackling #3231 (in the funnel, admittedly) and possibly helping towards https://github.com/statechannels/project-management/issues/44 (in scope for this sprint). Another advantage of this PR is that the work is done. The PR stacked on top of this one  #3301 would close the associated issue. 

I accept that there are legitimate alternatives, and I wasn’t trying to claim otherwise when I said “we need some record of memory or state”.  

Your alternative 1 is definitely a simpler solution. With that in mind I prepared #3306.

One downside of the simpler solution (namely, to update the timestamp whenever we crank an objective = take an action), is that it makes objective cranking non-idempotent. Cranking a second time results in a different output state because the timestamps change.

My proposal would make objective cranking idempotent (the timestamp only gets updated if the crank progress past a point that it previously got blocked on, which would require an intervening operation such as a new state being added to the store).

packages/server-wallet/src/db/migrations/20210126122746_add_objective_waiting_for_column.ts
```diff
@@ -0,0 +1,14 @@
+import * as Knex from 'knex';
+
+const tableName = 'objectives';
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(tableName, table => {
+    table.string('waiting_for').notNullable();
```
andrewgordstewart, line 6: This migration will fail in production. A new column cannot be `notNullable` unless a default value is provided.

packages/server-wallet/src/models/objective.ts
```diff
@@ -33,6 +33,7 @@ type ObjectiveStatus = 'pending' | 'approved' | 'rejected' | 'failed' | 'succeed
 type WalletObjective<O extends Objective> = O & {
   objectiveId: string;
   status: ObjectiveStatus;
+  waitingFor: string;
```
andrewgordstewart, line 36: ~I don't like the idea of unconstrained strings ending up in a `text` column that stores finite state.~

e. I see it is updated to a constrained type in future work.

packages/server-wallet/src/protocols/channel-opener.ts
```diff
@@ -38,34 +38,61 @@ export class ChannelOpener {
...
 
-      // if we don't have a complete preFundSetup, we're done for now
+      // We might be waitingFor theirPreFundState
+      let waitingFor = 'theirPreFundState';
+
+      // if we don't have a complete preFundSetup, we are still waitingFor theirPreFundState and cannot progress
       if (!channel.preFundComplete) {
-        response.queueChannel(channel); // always queue the channel if we've potentially touched it
+        await updateWaitingFor();
         return;
```
andrewgordstewart, line 55: It seems as if `updateWaitingFor` is always followed by `return`. It also seems as if that's an important property. Why not `return updateWaitingFor();`?

packages/server-wallet/src/protocols/channel-opener.ts
```diff
@@ -38,34 +38,48 @@ export class ChannelOpener {
...
-        response.queueSucceededObjective(objective);
-      }
+      waitingFor = '';
+      objective = await this.store.markObjectiveStatus(objective, 'succeeded', tx);
+      response.queueSucceededObjective(objective);
 
-      response.queueChannel(channel); // always queue the channel if we've potentially touched it
+      if (objective.waitingFor !== waitingFor) {
+        await ObjectiveModel.updateWaitingFor(objective.objectiveId, waitingFor, tx);
+      }
```
andrewgordstewart: I think the intent of this commit has been lost? The commit definitely changes behaviour, and so I'm not sure what things were refactored.

packages/server-wallet/src/protocols/channel-opener.ts
```diff
@@ -38,34 +38,61 @@ export class ChannelOpener {
       if (!channel) {
         throw new Error(`ChannelOpener can't find channel with id ${channelToLock}`);
       }
+      // always queue the channel if we've potentially touched it
+      response.queueChannel(channel);
```
andrewgordstewart, line 42: What's the purpose of this addition? Is it necessary for this PR, or was it just noticed as an omission?

Oh, nevermind, I see it was moved up from a conditional block down below.

andrewgordstewart (approved):
Does this require some documentation? This is a new concept that the rest of developers are now expected to adopt. This needs to be communicated somehow.

... Considering that fact, does this need a second approval?

packages/server-wallet/src/db/migrations/20210126122746_add_objective_waiting_for_column.ts
geoknee: Good catch. Relevant to https://github.com/statechannels/project-management/issues/55.

packages/server-wallet/src/protocols/channel-opener.ts
geoknee: Yeah that's a good idea. The follow-on PR goes down a very similar route, in fact.

packages/server-wallet/src/db/migrations/20210126122746_add_objective_waiting_for_column.ts
geoknee: I made the column nullable in e4cfddd

packages/server-wallet/src/protocols/channel-opener.ts
geoknee: I can't tell which commit you mean. I think probably bbcebb8, which was claiming to be a refactor but which broke the tests. I'll confess I thought it was a refactor but I was mistaken. We can rewrite history to correct that, but we will need to be careful because of the PR stacked on top of this one.

andrewgordstewart: ah, I was commenting on the commit itself, and believed it would be associated with it.

That's the one. I am averse to re-writing history, and do not want to push that burden on you!

lalexgap:
> * Moreover, this change feels like it can easily spur on a cascade of DB migrations in the near future. We do not have a well-formed plan for running migrations in production. So, we are adding short-term complexity that appears unnecessary.

Don't we already have existing migrations that will need to run in production? I think this is a problem that we will need to deal with anyways.

packages/server-wallet/src/db/migrations/20210126122746_add_objective_waiting_for_column.ts
lalexgap: Outside the scope of this PR, but at some point we might want to automate testing of the schema migrations against existing data at some point.

andrewgordstewart:
> > * Moreover, this change feels like it can easily spur on a cascade of DB migrations in the near future. We do not have a well-formed plan for running migrations in production. So, we are adding short-term complexity that appears unnecessary.
> 
> Don't we already have existing migrations that will need to run in production? I think this is a problem that we will need to deal with anyways.

The best-known means of running "breaking" migrations in production is difficult to orchestrate. Therefore, we need to minimize the risk of introducing "breaking" migrations.

> What is a "breaking" migration?

We often write migrations with the mindset that the DB is empty, and we are just changing the schema. The original migration in this PR is an example of that.

The _current_ migration in this PR will run in production, as of https://github.com/statechannels/statechannels/pull/3286/commits/e4cfdddc0cacad2c6deb84eb4bf6b59940930f18. But, it might actually be incompatible with the changeset in this PR, since the changeset is written from the perspective of "`waitingFor` is a `string`". Thus, it might be the case that `waitingFor` should be not-nullable, and the migration has to fill in the correct value. **I now feel the need to re-review the code and ensure that this is the case.**

If we do need to fill in the correct value, that is an expensive operation to run in production, where there exist a million channels. That could potentially bring down the gateway for an extended period of time, which may be unacceptable. This is particular possible due to our use of `jsonb` columns. In that case, we'd either need to
1. write some (potentially slow) SQL queries using `jsonb` operators (which we're unfamiliar with)
2. write some (pretty much guaranteed to be slow) migration which fetches the data, manipulates it in JS, and inserts the data, all wrapped in a transaction

Both of these options are time consuming, and they both introduce risk, at least until we have a strong process for validating migrations before shipping to production.

There's another class of migrations that I'm worried about, when some protocol is fundamentally changed. For instance, if we change the shape of `Objective` in the wire format, we would potentially have a "breaking" migration, and then need to coordinate migrations + upgrades between all parties in a channel. This is difficult, since one participant (eg. the gateway) might need to coordinate with a large set of peers (many indexers).

**All that said**, I doubt it is the case that we are going to introduce a "breaking" migration in the future, as a result of this work. Due to the lack of understanding around "breaking" migrations, my aim here was to have a greater certainty.

packages/server-wallet/src/db/migrations/20210126122746_add_objective_waiting_for_column.ts
andrewgordstewart: Is that consistent with the changeset of this PR, which I think expects `waiting_for` to be `NOT NULL`? (checking for myself in a bit)

geoknee: You're right. The ORM asserts that `waitingFor` is a string, whereas as of that commit it can now be `undefined`. That should be resolved before merging this PR.

geoknee: I think it is best to make `waitingFor` an optional property of a DBObjective. I think that's a better default than `''` or any other string.

geoknee: 83bf115

lalexgap: Does that mean that in #3301 we'd be returning `undefined` instead of `WaitingFor.nothing`?

geoknee: yes. We would only be doing that when the objective is terminated, though.

lalexgap:
> ⚠️ This PR is the lowest in a stack. Please do not squash- or rebase-merge it. The next PR up in the stack is #3301 .

I wonder if it would be easier to merge these two PRs together, as they seem pretty intertwined.

geoknee:
> I wonder if it would be easier to merge these two PRs together, as they seem pretty intertwined.

I tend to agree. I can make a new branch with all the changes.

geoknee:
In favour of https://github.com/statechannels/statechannels/pull/3308, which (as per suggestion) combines the two stacked PRs.

PR #3287, opened 2021-02-11
feat: new ledger channel algorithm

tomclose:
Adds a new ledger channel algorithm

## Background

The ledger funding algorithm in current master (roughly) works as follows:

- When an objective wants a channel to added to / removed from a ledger channel it creates a `LedgerRequest`. The request stores the `channelId`, `ledgerId`, and whether it's a fund/defund.
- After all objectives have been cranked, the `LedgerManager` processes the ledger request queue.
- In order to decide what update to propose, the participants exchange *ledger proposals.* Roughly, both participants send each other the list of changes they're willing to make at this point, and the actual change is computed by taking the intersection.
- A ledger update is then proposed containing this intersection, which should be accepted by each party.

As far as I'm aware, there isn't high confidence that the current implementation is robust. ~~There have been some efforts to show that it is, which haven't yet been successful.~~

## Changes made / motivations

This work suggests a new ledger funding algorithm. It makes the following changes:

1. No ledger proposals are required, instead the ledger states themselves are used to communicate proposals. This allows us to remove an object from the store and wire protocol, and should improve robustness, as we can lean on the existing mechanisms for syncing states between participants. We still obtain agreement in a single round-trip. 
2. Adds an efficient and robust mechanism for handling the case where a protocol wants to cancel funding it has requested. In short, a protocol can cancel a funding request with a defund request. The ledger manager will cancel the initial fund request if it can, or defund if it is too late to cancel.
3. Ledger requests now hold the diff they want to apply to the ledger, instead of the just the channelId. This decouples the ledger manager from the state of the channel, which should make it easier to add more exotic ledger operations (e.g. partial withdrawals) in the future. It also slightly more efficient, as it means the ledger manager doesn't need to query the channel.
4. It adds a "missedOpportunityCount" to the ledger request, which tracks how many time the request could have been included in a new agreed state but wasn't. In the future the wallet can use this information to detect ledger updates that have likely stalled (e.g. because the opponent has abandoned/lost the objective), and take appropriate action.

A detailed description of the algorithm is included in the [comments in the code](https://github.com/statechannels/statechannels/blob/tc/new-ledger-algorithm/packages/server-wallet/src/protocols/ledger-manager.ts#L11-L68). 

## Alternative approaches considered

I spent some time considering the "ledger updates as objectives" proposal. My notes are [here](https://www.notion.so/Ledger-Channels-d5af6ed79b7e4d4eb3e3511b7b8f8b18). In short:

- In the current one-objective-per-channel architecture, ledger updates as objectives breaks down pretty quickly, unless you stick to the case where a ledger is used to fund a single set of channels that are funded / defunded at the same time.
- There are no alternative written proposals for architectures that don't use one-objective-per-channel, so I am unable to evaluate the "ledger updates as objectives" in these architectures.

In any case, my understanding is that a key motivation for "ledger updates as objectives" is that it is believed that it would be simple and would just work, and it's worth living with the downsides in the short-term for those reasons (as far as I'm aware there is no proposal outlining the approach and the motivations, so I could be wrong here). If the approach proposed here is simple and works, it dominates this alternative, as it doesn't share the limitations described above.
## How Has This Been Tested?

I have written a fairly comprehensive test suite to test the new ledger-manager.

---

### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/__test-with-peers__/create-and-close-channel/ledger-funding.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/create-and-close-channel/ledger-funding.test.ts
@@ -24,7 +24,13 @@ import {
 
 const ETH_ASSET_HOLDER_ADDRESS = makeAddress(ethers.constants.AddressZero);
 
-const tablesUsingLedgerChannels = ['channels', 'ledger_requests', 'ledger_proposals'];
+const tablesUsingLedgerChannels = [
+  'channels',
+  'ledger_requests',
+  'ledger_proposals',
+  'objectives_channels',
+  'objectives',
+];
 
 jest.setTimeout(10_000);
 
@@ -153,7 +159,7 @@ describe('Funding a single channel with 100% of available ledger funds', () => {
     await DBAdmin.truncateDatabase(bWalletConfig, tablesUsingLedgerChannels);
   });
 
-  it('can fund a channel by ledger between two wallets ', async () => {
+  it('can fund and close channel by ledger between two wallets ', async () => {
     ledgerChannelId = await createLedgerChannel(10, 10);
     const params = testCreateChannelParams(10, 10, ledgerChannelId);
 
@@ -189,33 +195,31 @@ describe('Funding a single channel with 100% of available ledger funds', () => {
     });
 
     appChannelId = channelId;
-  });
 
-  it('closing said channel', async () => {
     const {outbox: close} = await peerWallets.a.closeChannel({channelId: appChannelId});
 
     await exchangeMessagesBetweenAandB([], [close]);
 
     await interParticipantChannelResultsAreEqual(peerWallets.a, peerWallets.b);
-    const {channelResults} = await peerWallets.a.getChannels();
+    const {channelResults: channelResults2} = await peerWallets.a.getChannels();
 
-    const ledger = getChannelResultFor(ledgerChannelId, channelResults);
+    const ledger2 = getChannelResultFor(ledgerChannelId, channelResults2);
 
     const {
-      allocations: [{allocationItems}],
-    } = ledger;
+      allocations: [{allocationItems: allocationItems2}],
+    } = ledger2;
 
-    expect(getChannelResultFor(appChannelId, channelResults)).toMatchObject({
+    expect(getChannelResultFor(appChannelId, channelResults2)).toMatchObject({
       turnNum: 4,
       status: 'closed',
     });
 
-    expect(allocationItems).toContainAllocationItem({
+    expect(allocationItems2).toContainAllocationItem({
       destination: participantA.destination,
       amount: BN.from(10),
     });
 
-    expect(allocationItems).toContainAllocationItem({
+    expect(allocationItems2).toContainAllocationItem({
       destination: participantB.destination,
       amount: BN.from(10),
     });
@@ -370,7 +374,7 @@ describe('Funding multiple channels synchronously (in bulk)', () => {
       allocations: [{allocationItems}],
     } = ledger;
 
-    expect(allocationItems).toHaveLength(N);
+    expect(allocationItems).toHaveLength(N + 2); // we always keep participants in outcome, even if empty
 
     for (const channelId of channelIds) {
       expect(getChannelResultFor(channelId, channelResults)).toMatchObject({
@@ -468,7 +472,7 @@ describe('Funding multiple channels concurrently (in bulk)', () => {
 
     expect(ledger.status).toBe('running');
 
-    expect(allocationItems).toHaveLength(N * 2);
+    expect(allocationItems).toHaveLength(N * 2 + 2); // always include participants, even if 0
 
     for (const channelId of channelIds) {
       expect(getChannelResultFor(channelId, channelResults)).toMatchObject({
@@ -551,7 +555,7 @@ describe('Funding multiple channels syncronously without enough funds', () => {
       allocations: [{allocationItems}],
     } = getChannelResultFor(ledgerChannelId, channelResults);
 
-    expect(allocationItems).toHaveLength(2);
+    expect(allocationItems).toHaveLength(2 + 2); // always keep participants in outcome, even if 0
 
     const unfundedChannels = channelResults.filter(c => c.status === 'opening');
 
@@ -867,7 +871,7 @@ describe('Automatic channel syncing on successive API calls', () => {
     await DBAdmin.truncateDatabase(bWalletConfig, tablesUsingLedgerChannels);
   });
 
-  it('can fund two channels in a situation where first proposal is dropped', async () => {
+  it.skip('can fund two channels in a situation where first proposal is dropped', async () => {
     const ledgerChannelId = await createLedgerChannel(2, 2);
     const params = testCreateChannelParams(1, 1, ledgerChannelId);
 
@@ -880,7 +884,7 @@ describe('Automatic channel syncing on successive API calls', () => {
 
     const {outbox: joinAndProposeFirst} = await peerWallets.b.joinChannel({channelId: channelId1});
 
-    /* This message is ignored 👇
+    /* This message is ignored 👇 
     const {outbox: agreeToProposalAndSign} = */ await peerWallets.a.pushMessage(
       getPayloadFor(participantA.participantId, joinAndProposeFirst)
     );
--- a/packages/server-wallet/src/models/ledger-request.ts
+++ b/packages/server-wallet/src/models/ledger-request.ts
@@ -85,7 +85,7 @@ export class LedgerRequest extends Model implements LedgerRequestType {
       .orWhere({ledgerChannelId, status: 'queued'});
   }
 
-  static async ledgersWithNewReqeustsIds(tx: TransactionOrKnex): Promise<string[]> {
+  static async ledgersWithNewRequestsIds(tx: TransactionOrKnex): Promise<string[]> {
     return (
       await LedgerRequest.query(tx).column('ledgerChannelId').whereNull('lastSeenAgreedState')
     ).map(lr => lr.ledgerChannelId);
--- a/packages/server-wallet/src/models/ledger-request.ts
+++ b/packages/server-wallet/src/models/ledger-request.ts
@@ -1,16 +1,25 @@
+import {BN} from '@statechannels/wallet-core';
 import {Model, TransactionOrKnex} from 'objection';
 
-import {Bytes32} from '../type-aliases';
+import {Uint256, Destination, Bytes32} from '../type-aliases';
 
 export type LedgerRequestStatus =
+  | 'queued'
+  | 'cancelled'
   | 'pending' // Request added to DB to be handled by ProcessLedgerQueue
   | 'succeeded' // Ledger update became supported and thus request succeeded
-  | 'failed'; // Failed for any reason e.g., lack of available funds
+  | 'insufficient-funds'
+  | 'inconsistent' // channel already exists but has a different total
+  | 'failed'; // if the ledger closes, or there's a protocol error
 
 export interface LedgerRequestType {
-  ledgerChannelId: Bytes32;
-  channelToBeFunded: Bytes32;
+  ledgerChannelId: Destination;
+  channelToBeFunded: Destination;
   status: LedgerRequestStatus;
+  amountA: Uint256;
+  amountB: Uint256;
+  lastSeenAgreedState: number | null;
+  missedOpportunityCount: number;
   type: 'fund' | 'defund';
 }
 
@@ -19,6 +28,10 @@ export class LedgerRequest extends Model implements LedgerRequestType {
   ledgerChannelId!: LedgerRequestType['ledgerChannelId'];
   status!: LedgerRequestType['status'];
   type!: LedgerRequestType['type'];
+  amountA!: Uint256;
+  amountB!: Uint256;
+  missedOpportunityCount!: number;
+  lastSeenAgreedState!: number | null;
 
   static tableName = 'ledger_requests';
   static get idColumn(): string[] {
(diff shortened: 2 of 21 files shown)


andrewgordstewart:
Is [this](https://www.notion.so/Ledger-updates-caaac6d640c2448398264d3d1c1feef3) the (only?) document inspiring this PR?

packages/server-wallet/src/__test-with-peers__/create-and-close-channel/ledger-funding.test.ts
```diff
@@ -867,7 +871,7 @@ describe('Automatic channel syncing on successive API calls', () => {
     await DBAdmin.truncateDatabase(bWalletConfig, tablesUsingLedgerChannels);
   });
 
-  it('can fund two channels in a situation where first proposal is dropped', async () => {
+  it.skip('can fund two channels in a situation where first proposal is dropped', async () => {
```
kerzhner, line 874: Should we remove this test?

packages/server-wallet/src/protocols/ledger-funder.ts
```diff
@@ -53,8 +53,11 @@ export class LedgerFunder {
     // if we already requested funding return false
     if (await this.alreadyRequestedFunding(channel.channelId, tx)) return false;
 
+    // NOTE FOR TOMORROW:
```
kerzhner, line 56: Is this still relevant?

kerzhner (commented):
I completed the first pass of the `ledger-manager`, and it looks reasonable to me! I'll take another pass tomorrow to see if I understand the tests and other code that supports the ledger-manager.

packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
```diff
@@ -1,54 +1,24 @@
...
-import {LedgerProposal} from '../../models/ledger-proposal';
+import {LedgerRequest, LedgerRequestStatus} from '../../models/ledger-request';
 import {DBAdmin} from '../..';
-import {Channel} from '../../models/channel';
+import {State} from '../../models/channel/state';
+import {LedgerManager} from '../ledger-manager';
+import {WalletResponse} from '../../wallet/wallet-response';
+import {Destination} from '../../type-aliases';
 
 // TEST HELPERS
```
kerzhner, line 17:
```suggestion
```

packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
```diff
@@ -1,54 +1,24 @@
...
-    typeof destinationOrFixture === 'string'
-      ? makeDestination(destinationOrFixture)
-      : destinationOrFixture().destination;
-  return {destination, amount: BN.from(amount)};
-}
-
-function simpleEthAllocation(...items: PreAllocationItem[]): SimpleAllocation {
-  return coreSimpleEthAllocation(items.map(allocationItem));
-}
 // END TEST HELPERS
```
kerzhner, line 19:
```suggestion
```

packages/server-wallet/src/models/ledger-request.ts
```diff
@@ -1,16 +1,25 @@
+import {BN} from '@statechannels/wallet-core';
 import {Model, TransactionOrKnex} from 'objection';
 
-import {Bytes32} from '../type-aliases';
+import {Uint256, Destination, Bytes32} from '../type-aliases';
 
 export type LedgerRequestStatus =
+  | 'queued'
+  | 'cancelled'
   | 'pending' // Request added to DB to be handled by ProcessLedgerQueue
```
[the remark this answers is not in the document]
kerzhner: Maybe this is the reason we need both queued and pending statuses? https://github.com/statechannels/statechannels/blob/tc/new-ledger-algorithm/packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts#L173

packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
```diff
@@ -62,660 +32,735 @@ beforeEach(async () => {
...
-      channelToBeFunded: appChannel.channelId,
-      status: 'succeeded',
-    });
+describe('as leader', () => {
+  describe('in the accept state', () => {
+    it(
+      'will create a proposal from requests in the queue',
+      testLedgerCrank({
+        as: 'leader',
+        before: {
```
kerzhner, line 48: I am a big fan of this declarative test style. The tests are pleasant to read!

packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
```diff
@@ -62,660 +32,735 @@ beforeEach(async () => {
...
+        after: {
+          agreed: {turn: 5, bals: {a: 10, b: 10, c: 3}},
+          requests: [
+            ['fund', 'c', 2, 1, 'succeeded', {missedOps: 0, lastSeen: 5}],
+            ['defund', 'c', 2, 1, 'queued', {missedOps: 0, lastSeen: 5}],
+          ],
+        },
+      })
+    );
+    // TODO: it will mark a mismatched cancellation as inconsistent
```
kerzhner, line 403: Does this still apply?

packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
```diff
@@ -62,660 +32,735 @@ beforeEach(async () => {
...
+            ['defund', 'c', 2, 1, 'queued'], // included in proposal
+            ['fund', 'e', 1, 1, 'queued'],
+          ],
+        },
+        after: {
+          agreed: {turn: 6, bals: {a: 8, b: 10, d: 5}},
+          requests: [
+            ['fund', 'd', 4, 1, 'succeeded', {missedOps: 0, lastSeen: 6}],
+            ['defund', 'c', 2, 1, 'succeeded', {missedOps: 0, lastSeen: 6}],
+            ['fund', 'e', 1, 1, 'queued', {missedOps: 1, lastSeen: 6}],
```
kerzhner, line 424: Is it true that the follower has no way of relaying this request to the leader? If I understand correctly, the leader must send a state to the follower that includes this request.

packages/server-wallet/src/protocols/ledger-manager.ts
```diff
@@ -1,500 +1,552 @@
...
-}
-
+import {State} from '../models/channel/state';
+import {SimpleAllocationOutcome} from '../models/channel/outcome';
+import {LedgerRequest} from '../models/ledger-request';
+
+// Ledger Update algorithm:
+// ------------------------
+//
+// Define the "leader" of the channel to be the first participant.
```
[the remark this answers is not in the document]
tomclose: I think ledger channels will likely be 2-participants for the foreseeable future (even with virtual channels, you only need two party ledgers). 

The algorithm does naturally scale to multiple parties though: the first participant is still the "leader" but you then have multiple "followers", who each either agree with what's passed to them, or remove the parts they don't like.

packages/server-wallet/src/protocols/ledger-manager.ts
```diff
@@ -1,500 +1,552 @@
...
-  ): Promise<boolean> {
-    const {fundingChannel, channelsRequestingFunds, channelsReturningFunds} = protocolState;
-    const {supported, channelId: ledgerChannelId} = fundingChannel;
+    // for fundings check which are present, and then check that the totals match
+    const includedFundings = _.intersectionWith(
+      fundings,
+      agreedOutcome.destinations,
+      (fund, dest) => fund.channelToBeFunded === dest
+    );
+    includedFundings.forEach(f => {
```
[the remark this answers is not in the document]
tomclose: Yes. I believe that this is currently enforced by a db constraint. This isn't an essential constraint though, and we could modify the algorithm to allow multiple fund/defund attempts per channel. It's a bit hard to imagine how that would come about though.

packages/server-wallet/src/protocols/ledger-manager.ts
```diff
@@ -1,500 +1,552 @@
...
-  if (a.assetHolderAddress !== b.assetHolderAddress)
-    throw new Error('xorOutcome: assetHolderAddresses not equal');
-  return {
-    type: 'SimpleAllocation',
-    assetHolderAddress: b.assetHolderAddress,
-    allocationItems: _.xorWith(a.allocationItems, b.allocationItems, areAllocationItemsEqual),
-  };
-};
+    if (
+      !antepenultimateState.fullySigned &&
```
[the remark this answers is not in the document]
tomclose: I think the code here is wrong. Will fix.

packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
tomclose: Yes that's true. The ledger algorithm starts from a position where both participants have a queue of requests, that they want to apply to the channel. It's the objective's job to ensure that both participants have the same requests in their queue.

tomclose: If the followers wallet notices that a request is getting a lot of missedOps, it can use that as a trigger to sync/re-send that objective.

packages/server-wallet/src/models/ledger-request.ts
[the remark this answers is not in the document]
tomclose: I think it isn't actually important for the algorithm, as it's deducible from the ledger states, and I'm pretty sure I always do this.

It might be nice to be able to communicate this information to the objective and/or user though. It's effectively "we're just waiting for your opponent to sign now". Initially I thought that objectives might behave differently according to whether the request was pending, as once a request is pending it's out of the wallets control whether it gets accepted or not, so the request can't be withdrawn. However, that can be fragile (a pending request might revert to queued; a race condition might mean that a queued request becomes pending), and the just-send-a-cancelling-defund approach handles these issues optimally and robustly.

Overall, I think I'd be inclined to leave it at this point as it (a) works, (b) isn't doing any harm and (c) potentially provides a bit of extra debugging/UX info, even though it isn't strictly necessary.

packages/server-wallet/src/__test-with-peers__/create-and-close-channel/ledger-funding.test.ts
tomclose: Yeah, probably. I wasn't sure what to do with this - the old ledger algorithm did a lot of extra ("pessimistic") state communication, so it could handle cases like these. The new ledger algorithm doesn't do this - it just relies on sync states being called to handle the case where things have stalled. So the new ledger algorithm fails this test, but that's probably ok?

packages/server-wallet/src/protocols/__test__/ledger-manager.test.ts
tomclose: I think we probably don't need this behaviour at this point.

packages/server-wallet/src/__test-with-peers__/create-and-close-channel/ledger-funding.test.ts
kerzhner: Yep, I think it is totally fine to not handle the same edge cases with the new ledger funding algorithm.

packages/server-wallet/src/__test-with-peers__/create-and-close-channel/ledger-funding.test.ts
```diff
@@ -24,7 +24,13 @@ import {
 
 const ETH_ASSET_HOLDER_ADDRESS = makeAddress(ethers.constants.AddressZero);
 
-const tablesUsingLedgerChannels = ['channels', 'ledger_requests', 'ledger_proposals'];
+const tablesUsingLedgerChannels = [
+  'channels',
+  'ledger_requests',
+  'ledger_proposals',
```
kerzhner, line 30: Is the `LedgerProposals` table/model unused now? If so, should we remove code that references the model and write a db migration to remove the table?

kerzhner: We can take care of this after this PR is merged. Filed https://github.com/statechannels/statechannels/issues/3371, so that we don't forget.

packages/server-wallet/src/models/ledger-request.ts
```diff
@@ -46,56 +75,106 @@ export class LedgerRequest extends Model implements LedgerRequestType {
...
-  ): Promise<LedgerRequestType[]> {
-    return LedgerRequest.query(tx).select().where({ledgerChannelId, status: 'pending'});
+  ): Promise<LedgerRequest[]> {
+    return LedgerRequest.query(tx)
+      .select()
+      .where({ledgerChannelId, status: 'pending'})
+      .orWhere({ledgerChannelId, status: 'queued'});
+  }
+
+  static async ledgersWithNewReqeustsIds(tx: TransactionOrKnex): Promise<string[]> {
```
kerzhner, line 88:
```suggestion
  static async ledgersWithNewRequestsIds(tx: TransactionOrKnex): Promise<string[]> {
```

packages/server-wallet/src/models/ledger-request.ts
```diff
@@ -46,56 +75,106 @@ export class LedgerRequest extends Model implements LedgerRequestType {
...
+    return (
+      await LedgerRequest.query(tx).column('ledgerChannelId').whereNull('lastSeenAgreedState')
```
kerzhner, line 90: Would you be able to elaborate on how this query compares to a query that reads all ledger requests that have  `queued` or `pending` status?

tomclose: I'm using the fact that `lastSeenAgreedState` will be null if and only if the ledger request has never been involved in a crank before. These requests are therefore the ones that were added in the last round of objective cranks, and therefore these are the ledgers we need to crank.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -948,68 +944,32 @@ export class SingleThreadedWallet
...
+    while (channelIds.length > 0) {
+      const objectiveIds = await this.store.getApprovedObjectiveIds(channelIds);
+      for (const objectiveId of objectiveIds) {
+        await this.objectiveManager.crank(objectiveId, response);
+      }
 
-    response.createdObjectives.map(o => this.emit('objectiveStarted', o));
-    response.succeededObjectives.map(o => this.emit('objectiveSucceeded', o));
-  }
+      const ledgersFromRequests = await this.store.getLedgersWithNewRequestsIds();
```
kerzhner, line 959: This is a follow up from a previous question. These are the ledger requests with an empty `lastSeenAgreedState`. What would be the impact on this code if `getLedgersWithNewRequestsIds` returned `queued` requests?

kerzhner: I am trying to think through the following scenario. Let's say that A and B would like to ledger fund channels c and d using ledger channel L. L's initial outcome is [A, B]. Let's walk through the following sequence of events:
1. A calls `createChannel`  for c, then A calls `createChannel` for d.
2. B calls `joinChannel` for c, then B calls `joinChannel` for d.
3. B now has ledger requests for c and d.
4. A receives B's prefund signature for c. B's prefund signature on d gets delayed in transit.
5. A sends a ledger state Proposal that only includes c: [A, B, c].
6. B signs the Proposal [A, B, c] and sets the `lastSeenState` for d.
7. B's prefund signature on d finally arrives to A.
8. A receives B's signature on Proposal [A, B, c]. A now sends a second Proposal that includes d: [A, B, c, d].

I can't seem to convince myself that B's `takeActions` will process the second Proposal. I am probably missing something...

kerzhner: Maybe I am starting to understand. When B receives the Proposal of [A, B, c, d], `takeActions` should be called with `channelIds` parameter of `L`. So ` await this.store.filterChannelIdsByIsLedger(channelIds)` should result in the proposal getting processed.

tomclose: In the case above, the ledger would be included in `ledgersFromChannels` (the line below this one), and so would get processed by b. The general logic is that there are two cases where you need to crank a ledger:

1. When that ledger channel itself has received an update
2. When an objective has just added a new request to the queue

If the case where a ledger has (a) already seen all of its requests, and (b) hasn't changed, you don't need to crank again, as you'll just get the same result. So selecting `queued` requests here would lead to redundant work.

I guess there's an argument for adding a `new` status to the ledger request, to make the case of "not yet processed by a ledger crank" more explicit.

kerzhner: Thanks for the explanation!

kerzhner (approved):
Nicely done @tomclose!

PR #3292, opened 2021-02-12
Defunder protocol tests

kerzhner:
The bulk of this PR is the addition of Defunder protocol tests. A couple minor changes are also included. These are:
- Defunder protocol ensures that all relations needed for the protocol are joined when `crank` is invoked.
- Defunder for ledger funded channels checks that the channel has a conclusion proof.

This PR fixes https://github.com/statechannels/statechannels/issues/3124 as it is the last piece of work needed.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/protocols/__test__/defunder.test.ts
+++ b/packages/server-wallet/src/protocols/__test__/defunder.test.ts
@@ -1,0 +1,119 @@
+import {defaultTestConfig} from '../..';
+import {testKnex} from '../../../jest/knex-setup-teardown';
+import {MockChainService} from '../../chain-service';
+import {DBAdmin} from '../../db-admin/db-admin';
+import {createLogger} from '../../logger';
+import {AdjudicatorStatusModel} from '../../models/adjudicator-status';
+import {Channel} from '../../models/channel';
+import {Funding} from '../../models/funding';
+import {Store} from '../../wallet/store';
+import {TestChannel} from '../../wallet/__test__/fixtures/test-channel';
+import {Defunder} from '../defunder';
+
+const testChannelObj = TestChannel.create({
+  aBal: 5,
+  bBal: 3,
+  finalFrom: 4,
+});
+let store: Store;
+
+beforeEach(async () => {
+  await DBAdmin.truncateDataBaseFromKnex(testKnex);
+
+  store = new Store(
+    testKnex,
+    defaultTestConfig().metricsConfiguration.timingMetrics,
+    defaultTestConfig().skipEvmValidation,
+    '0'
+  );
+});
+
+describe('Direct channel defunding', () => {
+  it('Channel is defunded via submission of withdrawAndConlcude transaction', async () => {
+    // Channel has yet to be concluded
+    await testChannelObj.insertInto(store, {
+      participant: 0,
+      states: [3, 4],
+      funds: 8,
+    });
+    const chainService = new MockChainService();
+    const spy = jest.spyOn(chainService, 'concludeAndWithdraw');
+    const logger = createLogger(defaultTestConfig());
+
+    let channel = await Channel.forId(testChannelObj.channelId, testKnex);
+    const defunder = new Defunder(store, chainService, logger);
+
+    testKnex.transaction(async tx => {
+      // Defunder should take no actions as there is no conclusion proof.
+      expect(await defunder.crank(channel, tx)).toEqual({
+        didSubmitTransaction: false,
+        isChannelDefunded: false,
+      });
+      expect(spy).not.toHaveBeenCalled();
+
+      // Defunder should submit a transaction since there is a conclusion proof.
+      const state = testChannelObj.wireState(5);
+      channel = await store.addSignedState(channel.channelId, state, tx);
+      expect(await defunder.crank(channel, tx)).toEqual({
+        didSubmitTransaction: true,
+        isChannelDefunded: false,
+      });
+      expect(spy).toHaveBeenCalledWith(channel.support);
+
+      // The channel has NOT been fully defunded.
+      // Defunder does not complete
+      await Funding.updateFunding(tx, channel.channelId, '0x03', testChannelObj.assetHolderAddress);
+      expect(await defunder.crank(channel, tx)).toEqual({
+        didSubmitTransaction: false,
+        isChannelDefunded: false,
+      });
+
+      // The channel has been fully defunded
+      // Defunder completes
+      await Funding.updateFunding(tx, channel.channelId, '0x00', testChannelObj.assetHolderAddress);
+      expect(await defunder.crank(channel, tx)).toEqual({
+        didSubmitTransaction: false,
+        isChannelDefunded: true,
+      });
+    });
+  });
+
+  it('Channel is defunded via pushOutcome transaction', async () => {
+    // Channel has yet to be concluded
+    await testChannelObj.insertInto(store, {
+      participant: 0,
+      states: [3, 4],
+      funds: 8,
+    });
+    const state3 = testChannelObj.signedStateWithHash(3);
+    const state4 = testChannelObj.signedStateWithHash(4);
+
+    const chainService = new MockChainService();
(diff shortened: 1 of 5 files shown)


packages/server-wallet/src/protocols/__test__/defunder.test.ts
```diff
@@ -0,0 +1,119 @@
...
+import {DBAdmin} from '../../db-admin/db-admin';
+import {createLogger} from '../../logger';
+import {AdjudicatorStatusModel} from '../../models/adjudicator-status';
+import {Channel} from '../../models/channel';
+import {Funding} from '../../models/funding';
+import {Store} from '../../wallet/store';
+import {TestChannel} from '../../wallet/__test__/fixtures/test-channel';
+import {Defunder} from '../defunder';
+
+const testChannelObj = TestChannel.create({
```
andrewgordstewart, line 14: Why is this `testChannelObj`?

Does `Obj` mean `Object`? I originally read it to mean `Objective`. I think both are confusing to me, but maybe `testChannelObject` is not confusing?

packages/server-wallet/src/protocols/__test__/defunder.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      // Defunder should take no actions as there is no conclusion proof.
+      expect(await defunder.crank(channel, tx)).toEqual({
+        didSubmitTransaction: false,
+        isChannelDefunded: false,
+      });
+      expect(spy).not.toHaveBeenCalled();
+
+      // Defunder should submit a transaction since there is a conclusion proof.
+      const state = testChannelObj.wireState(5);
+      channel = await store.addSignedState(channel.channelId, state, tx);
```
andrewgordstewart, line 65: (out-of-PR-scope?): This smells a bit off to me, as we're using some store feature to set up the DB state for this test. I can't exactly say what's wrong with this.

`TestChannel` instances have a `insertInto` method, which seems useful. Should there be a `testChannelObj.setVars` method? I don't know if that's more maintainable or less maintainable.

packages/server-wallet/src/protocols/__test__/defunder.test.ts
```diff
@@ -0,0 +1,119 @@
...
+        didSubmitTransaction: true,
+        isChannelDefunded: false,
+      });
+      expect(spy).toHaveBeenCalledWith(state4, channel.myAddress);
+
+      // The channel has been fully defunded
+      // Defunder completes
+      await Funding.updateFunding(tx, channel.channelId, '0x00', testChannelObj.assetHolderAddress);
+      expect(await defunder.crank(channel, tx)).toEqual({
+        didSubmitTransaction: false,
```
andrewgordstewart, line 123: Why should this be `false`? It was just `true`. It now sounds a bit like the transaction was un-submitted.

packages/server-wallet/src/protocols/defunder.ts
```diff
@@ -77,6 +79,8 @@ export class Defunder {
       await ChainServiceRequest.insertOrUpdate(channel.channelId, 'pushOutcome', tx);
       await this.chainService.pushOutcomeAndWithdraw(
         adjudicatorStatus.states[0],
+        // todo: we are assuming that we submitted the challenge.
```
andrewgordstewart, line 82: It's nice to be consistent. `TODO` is highlighted by vs-code (or at least by a common plugin?). `todo` is not.

```suggestion
        // TODO: we are assuming that we submitted the challenge.
```

packages/server-wallet/src/protocols/defunder.ts
```diff
@@ -90,14 +90,18 @@ export class Defunder {
   }
 
   private async ledgerDefunder(channel: Channel, tx: Transaction): Promise<DefunderResult> {
+    const didSubmitTransaction = false;
```
andrewgordstewart, line 93: What's the purpose of `didSubmitTransaction` if it is a constant with value `false`? That makes it seem like it doesn't belong on `DefunderResult`, as its value appears to contain no information.

e. Ok, I now realize that `ledgerDefunder` should never return `didSubmitTransaction: true`. I wonder if this should be clarified via a code comment.

andrewgordstewart (approved):
LGTM, some small comments.

I also reviewed the codebase as of `defunder-test`, to verify that https://github.com/statechannels/statechannels/issues/3124 is "done". 

While reviewing the code, I wrote https://github.com/statechannels/statechannels/pull/3299 (⚠️ which is branched off of `defunder-test` ⚠️ ). Unfortunately, I see the tests are failing. This exposed an inconsistency between how a protocol is expected to handle errors. This is somewhat relevant to #3124, as the `channel-defunder` seems to be the odd one out.

packages/server-wallet/src/protocols/defunder.ts
kerzhner: Consistency sounds good to me! `todo` is not highlighted in my editor regardless of case.

packages/server-wallet/src/protocols/__test__/defunder.test.ts
kerzhner: I agree that is confusing. Changed in 7f30e0e86

packages/server-wallet/src/protocols/__test__/defunder.test.ts
kerzhner: `await testChannel.insertInto(store, {states: [5]})` would actually work in this case except that the transaction object must be used for all database operations that `insertInto` invokes.

packages/server-wallet/src/protocols/__test__/defunder.test.ts
kerzhner: I think `didSumbitTransaction` is a bit confusing. Added a comment in 3dddf847f

packages/server-wallet/src/protocols/defunder.ts
kerzhner: Curious if 3dddf847f helps

andrewgordstewart: It's very clarifying! But it does cause me to ask: Why doesn't `Defunder.crank` just return a `boolean`? I'm going to read the surrounding context to try to answer that.

andrewgordstewart: Actually, it seems like
- `didSubmitTransaction` is used purely for testing within this PR.
- On `master`, [`didSubmitTransaction`](https://github.com/statechannels/statechannels/blob/3d6270b0c49914ff0f8d367fce340189f4924225/packages/server-wallet/src/protocols/channel-closer.ts#L60) is not used

So, it seems to me like `Defunder.crank` should just return a `boolean`?

kerzhner: What about https://github.com/statechannels/statechannels/blob/3d6270b0c49914ff0f8d367fce340189f4924225/packages/server-wallet/src/protocols/defund-channel.ts#L56?

andrewgordstewart: Now I'm confused.
1. Why does a `ChannelDefunder` use a `Defunder`? What is the 2nd defunder for?
2. Why doesn't the `ChannelDefunder` just ask if `Defunder.crank` returned `true`?

kerzhner: `ChannelDefunder` was written before much thought went into first-class objectives. I don't think we will be using that protocol going forward. To eliminate `ChannelDefunder`, we need to think through how challenges work with first class objectives. It probably doesn't make sense to refactor `ChannelDefunder` and its test. My proposal going forward is (which is what has been implemented):
1. Enable `ChannelDefunder`'s current behavior via `didSubmitTransaction` return value.
2. When `ChannelDefunder` is removed, remove `didSubmitTransactoin`.

andrewgordstewart: Sounds good. Should we record that proposal somewhere?

kerzhner: That's a good suggestion! https://github.com/statechannels/statechannels/issues/3309

Issue #3269, opened 2021-02-09
isDirectFunded parameter is unclear


kerzhner:
Nit: With `this.isDirectFunded(true)` it's not immediately clear what the boolean means.

If we switched it to accept a `"Fully"|"Partially"` type we'd have `this.isDirectFunded('Fully')` which I'd argue is more straightforward.

_Originally posted by @lalexgap in https://github.com/statechannels/statechannels/pull/3256#discussion_r572501952_

PR #3293, opened 2021-02-12
refactor: change isDirectFunded  parameter to be more verbose

kerzhner:
Fixes https://github.com/statechannels/statechannels/issues/3269

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -13,6 +13,7 @@ import {
   checkThat,
   isSimpleAllocation,
   BN,
+  unreachable,
 } from '@statechannels/wallet-core';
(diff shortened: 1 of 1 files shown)


geoknee (approved):
Looks like a good improvement to me!

Nit/idea: Given that `isDirectFunded` is a private method, would it not be simpler to move each `case` block into the corresponding public function and not have the private method?

kerzhner:
> Nit/idea: Given that `isDirectFunded` is a private method, would it not be simpler to move each `case` block into the corresponding public function and not have the private method?

We could definitely do that! What would you do with logic shared between `isFullyDirectFunded` and `isPartlyDirectFunded`? Just copy/paste it? Or factor it out?

geoknee:
Not sure, probably just copy/paste it?
How about we merge this, and I have a play around with it and raise a new PR if I think it is worth it?

Issue #2939, opened 2020-11-20
[server-wallet] Protect against chain reorgs for contract holdings reads


kerzhner:
When the wallet registers a channel with the chain service, the chain service emits the current holdings of the channel. At the moment, the chain service:
1. Queries the asset holder storage for channel holdings.
1. Calls `holdingUpdated` method on the wallet.

We should protect against chain reorgs using something like the following algorithm:
1. Read channel holdings from asset holder storage. Let's say the chain service reads the value X for channel C in asset holder A.
1. Look back to see for how many blocks C has had X in A. Let's say that X last changed in block N.
1. If current block number minus N is less that required confirmation time, wait for the number of blocks required for confirmation.
1. If the holdings are updated while waiting on the previous step, restart the algorithm.

PR #3300, opened 2021-02-15
Get only confirmed blocks when calculating initial holdings

isoosiss7:
# Description
1. Get only confirmed blocks when calculating initial holdings.
2. Re-play events in the unconfirmed blocks (when confirmed) - to make sure they aren't missed.
3. Handle the racing condition that could be introduced by such.
https://github.com/statechannels/statechannels/issues/2927
https://github.com/statechannels/statechannels/issues/2939

- [ ] what is the problem being solved
To avoid including holdings that's not confirmed yet.
- [ ] why this is a good approach
BlockTag give us data as of 6 blocks ago
- [ ] what are the shortcomings of this approach, if any
Using BlockTag need provider support.


isoosiss7:
/********************************   
* These cases run good when alone, but unstable when run together with others, likely due to
* there're still threads created by others advancing the block chain.
* Removed
********************************/
  it('Initial holding right when deposit not confirmed', async () => {
    mineBlocks();
    const channelId = randomChannelId();
    await waitForChannelFunding(0, 2, channelId);
    await mineBlocks();
    await waitForChannelFunding(2, 3, channelId);

    // When we query holdings we only get confirmed amount (6 blocks ago),
    // at this time we have 5 latest but only 2 is confirmed
    await new Promise(resolve =>
      chainService.registerChannel(channelId, [ethAssetHolderAddress], {
        ...defaultNoopListeners,
        holdingUpdated: arg => {
          expect(arg).toMatchObject({
            channelId,
            assetHolderAddress: ethAssetHolderAddress,
            amount: BN.from(2),
          });
          resolve(true);
        },
      })
    );
  });

  it('Initial holding 0 when channel not confirmed', async () => {
    await delay(5000);
    const channelId = randomChannelId();
    await waitForChannelFunding(0, 5, channelId);

    // When we query holdings we only get confirmed amount (6 blocks ago),
    // if channel creation is within 6 blocks then we should get 0
    await new Promise(resolve =>
      chainService.registerChannel(channelId, [ethAssetHolderAddress], {
        ...defaultNoopListeners,
        holdingUpdated: arg => {
          expect(arg).toMatchObject({
            channelId,
            assetHolderAddress: ethAssetHolderAddress,
            amount: BN.from(0),
          });
          resolve(true);
        },
      })
    );
  });

andrewgordstewart (commented):
Without diving into this too deeply, how does the chain service behave when:
1. Holdings are 0 in block 1
2. Holdings are updated to 10 in block 2
3. Wallet restarts. The latest block is block 3.

As of this change, the chain service would return the confirmed holdings. Will it later inform the wallet that the holdings are updated to 10? (I think this should happen when block 7 occurs.)

isoosiss7:
> Without diving into this too deeply, how does the chain service behave when:
> 
> 1. Holdings are 0 in block 1
> 2. Holdings are updated to 10 in block 2
> 3. Wallet restarts. The latest block is block 3.
> 
> As of this change, the chain service would return the confirmed holdings. Will it later inform the wallet that the holdings are updated to 10? (I think this should happen when block 7 is mined, but only if it includes the block 2 that I noticed from before.)

No. 

It seems when we subscribe to events we only  get the latest, there is no blockTag passed into this interface. https://docs.ethers.io/v5/api/contract/example/#erc20-events

So sounds like, when deposit/allocation/wallet restart happens, we should create an objective, and come back to check again after 6 blocks?

isoosiss7:
> Without diving into this too deeply, how does the chain service behave when:
> 
> 1. Holdings are 0 in block 1
> 2. Holdings are updated to 10 in block 2
> 3. Wallet restarts. The latest block is block 3.
> 
> As of this change, the chain service would return the confirmed holdings. Will it later inform the wallet that the holdings are updated to 10? (I think this should happen when block 7 is mined, but only if it includes the block 2 that I noticed from before.)

I took another

isoosiss7:
> > Without diving into this too deeply, how does the chain service behave when:
> > 
> > 1. Holdings are 0 in block 1
> > 2. Holdings are updated to 10 in block 2
> > 3. Wallet restarts. The latest block is block 3.
> > 
> > As of this change, the chain service would return the confirmed holdings. Will it later inform the wallet that the holdings are updated to 10? (I think this should happen when block 7 is mined, but only if it includes the block 2 that I noticed from before.)
> 
> No.
> 
> It seems when we subscribe to events we only get the latest, there is no blockTag passed into this interface. https://docs.ethers.io/v5/api/contract/example/#erc20-events
> 
> So sounds like, when deposit/allocation/wallet restart happens, we should create an objective, and come back to check again after 6 blocks?

I took another look at Andrew's question, we seem already have some logic to handle confirmation on events.

[registerChannel](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L326) -> [setUpAssetHolderListener](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L339) -> [addAssetHolderObservable](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L437) -> [waitForConfirmations](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L489)

However, due to racing conditions there is still a slight chance that things can go wrong:
If the chain advances between[setUpAssetHolderListener](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L339) and [getInitialHoldings](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L343), a deposit event can be double counted. The mitigate would be, always re-read and overwrite the holdings amount in [event](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/chain-service/chain-service.ts#L493).

I'll think about this for a bit.

@kerzhner any suggestions?

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -360,7 +361,7 @@ describe('concludeAndWithdraw', () => {
     if (!transactionResponse) throw 'Expected transaction response';
     await transactionResponse.wait();
 
-    mineBlocks();
+    await mineBlocks();
```
kerzhner, line 364: I could be wrong, but I don't think this `await` is needed since we are blocking on channel finalization promise `p` in the line below.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -401,7 +402,7 @@ describe('concludeAndWithdraw', () => {
 
     expect(await provider.getBalance(aAddress)).toEqual(BigNumber.from(1));
     expect(await provider.getBalance(bAddress)).toEqual(BigNumber.from(3));
-    mineBlocks();
+    await mineBlocks();
```
kerzhner, line 405: Same comment as above.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -448,7 +449,7 @@ describe('concludeAndWithdraw', () => {
     expect(await erc20Contract.balanceOf(aAddress)).toEqual(BigNumber.from(1));
     expect(await erc20Contract.balanceOf(bAddress)).toEqual(BigNumber.from(3));
 
-    mineBlocks();
+    await mineBlocks();
```
kerzhner, line 452: Same comment as above.

isoosiss7:
Updated to wait on unconfirmed events instead.

isoosiss7:
My bad, cause getInitialHoldings is async already so it won't block registerChannel. registerChannel exist immediately, but by the time subscriber.holdingUpdated is called, it is 5 blocks away and all deposits included in the initial holdings are confirmed.

So the current version of the code should work.

@kerzhner pls help to do another round of review to see if this solution is good. Thank you!

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -412,7 +412,25 @@ export class ChainService implements ChainServiceInterface {
     contract: Contract,
     channelId: string
   ): Promise<HoldingUpdatedArg> {
+    const blockNumber = await this.provider.getBlockNumber();
     const holding = BN.from(await contract.holdings(channelId));
+    const depositEvents = await contract.queryFilter(
+      {
+        address: contract.address,
```
kerzhner, line 419: I am surprised that this is needed. Isn't `queryFilter` invoked on a contract object that is aware of its own address?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -412,7 +412,25 @@ export class ChainService implements ChainServiceInterface {
...
+      },
+      blockNumber - this.blockConfirmations - 1
```
kerzhner, line 421:
```suggestion
      blockNumber - this.blockConfirmations + 1
```

Number of confirmations is the number of blocks that have been mined (including the initial block) since this transaction was mined. If the current block is 10 and we require 3 confirmation blocks, we should inspect blocks 8, 9, and 10.

kerzhner (commented):
Thanks for putting this together! In my thinking, there are 2 components to initial holdings work:
1. The chain service reads holdings from the chain. The holdings are considered valid only if they holdings have been constant for a confirmation-number of blocks.
2. There might be recent updates to holdings. In this case, the chain service should notify subscribers when the updates become valid (which is after confirmation-number of blocks).

If I am interpreting this work correctly, the PR tackles point 2 but not point 1.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -308,6 +320,26 @@ describe('registerChannel', () => {
...
+        ...defaultNoopListeners,
+        holdingUpdated: arg => {
+          expect(arg.amount).toEqual(BN.from(expected[0]));
+          expected.shift();
+          if (!expected.length) resolve(true);
+        },
+      });
+    });
+    fundChannel(8, 2, channelId, ethAssetHolderAddress);
+    mineBlocksSlow(20);
```
kerzhner, line 339: Does this add 10 seconds to the testsuite run?

kerzhner (commented):
Thanks for thinking this through and putting this work together. This is definitely a very tricky part of the chain service.

My interpretation of the work is that we are now doing the following:
1. When an event E is observed from the blockchain, we wait the number of confirmation blocks before considering E as valid.
2. Once E is valid, we inspect whether another more recent and valid event E' has been observed. If yes, E is ignored.
3. If not, subscribers are notified of E.

It seems to me that this is the right approach to think about events. One thing that would be nice to try is to the following:
1. A blockchain event can be observed from [a contract listener](https://github.com/statechannels/statechannels/blob/06d5721af2b8ec95cb0b5629dd3947611440f235/packages/server-wallet/src/chain-service/chain-service.ts#L535) or from [looking at past blocks](https://github.com/statechannels/statechannels/blob/06d5721af2b8ec95cb0b5629dd3947611440f235/packages/server-wallet/src/chain-service/chain-service.ts#L458).
2. It seems that once an event is observed, the same 3 steps need to be taken (as outlined above). At the moment this logic is not shared. Is there a reasonable way to factor out/ share this logic?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -411,15 +411,103 @@ export class ChainService implements ChainServiceInterface {
...
+      confirmedBlock
+    );
+
+    // We're unsure if the same events are also played by contract observer callback,
+    // but need to play it regardless, so subscribers won't miss anything.
+    // See EventTracker documentation
+    for (const ethersEvent of (await contract.queryFilter({}, confirmedBlock)).sort(
+      e => e.blockNumber
+    )) {
+      await this.waitForConfirmations(ethersEvent);
```
kerzhner, line 461: It seems like this is the same logic that is applied for events that are generated by contract listeners. Can it be factored out?

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -411,15 +411,103 @@ export class ChainService implements ChainServiceInterface {
...
+    try {
+      // https://docs.ethers.io/v5/api/contract/contract/#Contract--metaclass
+      // provider can lose track of the latest block, force it to reload
+      const overrides = {
+        blockTag: confirmedBlock,
+      };
+
+      confirmedHolding = BN.from(await contract.holdings(channelId, overrides));
+      this.logger.debug(
+        `Successfully red confirmedHoldings (${confirmedHolding}), from block ${confirmedBlock}.`
```
kerzhner, line 430:
```suggestion
        `Successfully read confirmedHoldings (${confirmedHolding}), from block ${confirmedBlock}.`
```

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
isoosiss7: No - we don't wait for this, test case will finish without waiting for all async operations to finish.

isoosiss7:
> Thanks for thinking this through and putting this work together. This is definitely a very tricky part of the chain service.
> 
> My interpretation of the work is that we are now doing the following:
> 
> 1. When an event E is observed from the blockchain, we wait the number of confirmation blocks before considering E as valid.
> 2. Once E is valid, we inspect whether another more recent and valid event E' has been observed. If yes, E is ignored.
> 3. If not, subscribers are notified of E.
> 
> It seems to me that this is the right approach to think about events. One thing that would be nice to try is to the following:
> 
> 1. A blockchain event can be observed from [a contract listener](https://github.com/statechannels/statechannels/blob/06d5721af2b8ec95cb0b5629dd3947611440f235/packages/server-wallet/src/chain-service/chain-service.ts#L535) or from [looking at past blocks](https://github.com/statechannels/statechannels/blob/06d5721af2b8ec95cb0b5629dd3947611440f235/packages/server-wallet/src/chain-service/chain-service.ts#L458).
> 2. It seems that once an event is observed, the same 3 steps need to be taken (as outlined above). At the moment this logic is not shared. Is there a reasonable way to factor out/ share this logic?

Yes you understanding is correct.

For 1, yes, contract listener will play all events from "provider.blockNumber" to the latest block.  "provider.blockNumber" can be out of date at the time when we register channel. So we don't know which events are captured by listener and which are not. In theory we can move provider.blockNumber back and forth to replay events, but that's dangerous as it affects all contracts.

For 2, good suggestion. I'll do it and ping you for another review once ready.

isoosiss7:
> > Thanks for thinking this through and putting this work together. This is definitely a very tricky part of the chain service.
> > My interpretation of the work is that we are now doing the following:
> > 
> > 1. When an event E is observed from the blockchain, we wait the number of confirmation blocks before considering E as valid.
> > 2. Once E is valid, we inspect whether another more recent and valid event E' has been observed. If yes, E is ignored.
> > 3. If not, subscribers are notified of E.
> > 
> > It seems to me that this is the right approach to think about events. One thing that would be nice to try is to the following:
> > 
> > 1. A blockchain event can be observed from [a contract listener](https://github.com/statechannels/statechannels/blob/06d5721af2b8ec95cb0b5629dd3947611440f235/packages/server-wallet/src/chain-service/chain-service.ts#L535) or from [looking at past blocks](https://github.com/statechannels/statechannels/blob/06d5721af2b8ec95cb0b5629dd3947611440f235/packages/server-wallet/src/chain-service/chain-service.ts#L458).
> > 2. It seems that once an event is observed, the same 3 steps need to be taken (as outlined above). At the moment this logic is not shared. Is there a reasonable way to factor out/ share this logic?
> 
> Yes you understanding is correct.
> 
> For 1, yes, contract listener will play all events from "provider.blockNumber" to the latest block. "provider.blockNumber" can be out of date at the time when we register channel. So we don't know which events are captured by listener and which are not. In theory we can move provider.blockNumber back and forth to replay events, but that's dangerous as it affects all contracts.
> 
> For 2, good suggestion. I'll do it and ping you for another review once ready.

1. I added subscribeToConfirmedEvents and queryConfirmedEvents to generalize the logic to wait for confirmation. Now all events will need to be confirmed before passed to customer.
2. De-dupe logic cannot be further generalized - whether an asset holder event is duplicate, depends on the destination address it is passed to, for example:
    1. A deposit to Ether assetholder happened at block #100, picked up at #120
    2. A depost to ERC20 assetholder happened at block #95, picked up at #130
In this case, the later shouldn't be considered duplicate, as it has different destination.
It's better to leave this logic in EventTracker.

kerzhner:
@isoosiss7 thanks! I think we are homing in on the right approach to this problem!

I took a stab at simplifying and factoring out ethers asset holder event processing in 5d27a6600911d73f19809148830fe7c27897adfe..ff01ba2dc7b8f3a5550a1ba0d3e51d4ec038420c. I would love your thoughts on whether you think this is a good approach. If so, we can add those commits to this PR.

Also, a question. When I factor out [the event replay logic](https://github.com/statechannels/statechannels/blob/b0483886fc21ef9302b72ccce9764e0799c96877/packages/server-wallet/src/chain-service/chain-service.ts#L473-L503), chain unit tests pass. Do you think it would be possible to add unit tests to this logic?

isoosiss7:
> the event replay logic

Hi @kerzhner ,

1. Thanks for refactoring logic, things are further simplified. Looks good to me, thanks!
2. Yes we can add a UT case. Existing UT cases covered duplicate message got removed. We should have a case to check when event listener didn't play event, the query filter can catch it. I'll add such a case.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
kerzhner: Got it! Our approach in the past has been to mine the minimum number of blocks for a test to pass. This has the benefit of tests not interfering with each other. An example is the following. Let's say we have a test to verify that a callback is invoked 5 blocks after a transaction. We would set up the test in the following way:
1. Mine the transaction.
2. Mine 2 blocks.
3. Verify that the callback has not been invoked.
4. Mine 3 more blocks.
5. Verify that the callback has been invoked.

Mining extra blocks in a test has the unfortunate side-effect of not being able to write tests using the approach above.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -293,9 +304,10 @@ describe('registerChannel', () => {
     await p;
   });
 
-  it('Receives correct initial holding when holdings are not 0', async () => {
+  it('Receives correct initial holding when holdings are confirmed', async () => {
     const channelId = randomChannelId();
     await waitForChannelFunding(0, 5, channelId);
+    await mineBlocks(); // wait for deposit to be confirmed thenr egister channel
```
kerzhner, line 310:
```suggestion
    await mineBlocks(); // wait for deposit to be confirmed then register channel
```

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -108,13 +117,13 @@ export class ChainService implements ChainServiceInterface {
       ContractArtifacts.NitroAdjudicatorArtifact.abi
     );
 
-    this.nitroAdjudicator.on(ChallengeRegistered, (...args) => {
+    this.subscribeToConfirmedEvents(this.nitroAdjudicator, ChallengeRegistered, (...args) => {
```
kerzhner, line 120: In this case, the chain service should probably notify subscribers as soon as the chain service observes the challenge event. Challenges are time sensitive. We should give subscribers the maximum amount of time to process a challenge.

packages/server-wallet/src/chain-service/event-tracker.ts
```diff
@@ -0,0 +1,105 @@
...
+        return true;
+      } else {
+        return false;
+      }
+    } else {
+      return false;
+    }
+  }
+
+  constructor(managedSubscriber: ChainEventSubscriberInterface, logger: Logger) {
```
kerzhner, line 49: Nit, it would be nice to place the constructor as the first function in the class (in this case, on line 25)

packages/server-wallet/src/chain-service/event-tracker.ts
```diff
@@ -0,0 +1,105 @@
...
+        this.assetHolderMap.get(assetHolderAddress) || {}
+      )}`
+    );
+    if (!this.assetHolderMap.has(assetHolderAddress)) {
+      this.assetHolderMap.set(assetHolderAddress, {blockNumber: blockNumber, logIndex: logIndex});
+      return true;
+    } else if (blockNumber > (this.assetHolderMap.get(assetHolderAddress)?.blockNumber || 0)) {
+      this.assetHolderMap.set(assetHolderAddress, {blockNumber: blockNumber, logIndex: logIndex});
+      return true;
+    } else if (blockNumber == (this.assetHolderMap.get(assetHolderAddress)?.blockNumber || 0)) {
```
kerzhner, line 37:
```suggestion
    } else if (blockNumber === (this.assetHolderMap.get(assetHolderAddress)?.blockNumber || 0)) {
```

packages/server-wallet/src/chain-service/event-tracker.ts
isoosiss7: Will do

packages/server-wallet/src/chain-service/chain-service.ts
isoosiss7: will do
nitroAdjudicator should not use confirmed events then

packages/server-wallet/src/chain-service/event-tracker.ts
```diff
@@ -0,0 +1,105 @@
...
+  private managedSubscriber: ChainEventSubscriberInterface;
+
+  private isNewEvent(assetHolderAddress: Address, blockNumber: number, logIndex: number): boolean {
+    this.logger.debug(
+      `return.isNewEvent: ${assetHolderAddress}, ${blockNumber}, ${logIndex}, ${JSON.stringify(
+        this.assetHolderMap.get(assetHolderAddress) || {}
+      )}`
+    );
+    if (!this.assetHolderMap.has(assetHolderAddress)) {
+      this.assetHolderMap.set(assetHolderAddress, {blockNumber: blockNumber, logIndex: logIndex});
```
kerzhner, line 32: Nit: it might be nice to factor out this setter logic. One way is to create a small helper function. Another way is to not use early returns and use a mutable variable to determine whether a new value needs to be set.

isoosiss7: will do

packages/server-wallet/src/chain-service/event-tracker.ts
```diff
@@ -0,0 +1,105 @@
...
+ */
+export class EventTracker {
+  private logger: Logger;
+  private assetHolderMap: Map<Address, {blockNumber: number; logIndex: number}>;
+  private managedSubscriber: ChainEventSubscriberInterface;
+
+  private isNewEvent(assetHolderAddress: Address, blockNumber: number, logIndex: number): boolean {
+    this.logger.debug(
+      `return.isNewEvent: ${assetHolderAddress}, ${blockNumber}, ${logIndex}, ${JSON.stringify(
+        this.assetHolderMap.get(assetHolderAddress) || {}
```
kerzhner, line 28: Nit, this function invokes `this.assetHolderMap.get` 4 times. We might want to factor that out to the top of the function.

packages/server-wallet/src/chain-service/event-tracker.ts
```diff
@@ -0,0 +1,105 @@
...
+          JSON.stringify(arg)
+      );
+    }
+  }
+
+  // Pass event to managed subscriber only if new
+  assetOutcomeUpdated(
+    arg: AssetOutcomeUpdatedArg,
+    blockNumber: number,
+    logIndex: number = Number.MAX_SAFE_INTEGER
```
kerzhner, line 80: I am curious on why this default value is needed.

packages/server-wallet/src/chain-service/event-tracker.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  // Pass event to managed subscriber only if new
+  holdingUpdated(
+    arg: HoldingUpdatedArg,
+    blockNumber: number,
+    logIndex: number = Number.MAX_SAFE_INTEGER // if no logIndex passed in then this is from getInitialHoldings
+  ): void {
+    if (this.isNewEvent(arg.assetHolderAddress, blockNumber, logIndex)) {
+      this.managedSubscriber.holdingUpdated(arg);
+      this.logger.debug(
```
kerzhner, line 64: These logs can probably be added to `isNewEvent` to clean up this function and the one below

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
isoosiss7: I'll  modify it to await the mining process to finish before moving to the next case. The exact amount of time to wait will be hard to find. I'll find a value that is safe.

packages/server-wallet/src/chain-service/event-tracker.ts
isoosiss7: Yep will do

packages/server-wallet/src/chain-service/event-tracker.ts
isoosiss7: getInitialHoldings will read a block instead of an event. In that case, it doesn't have logIndex. However, it should be considered the last event in this block, because the balance it red is final for that block.

kerzhner:
> 1. Thanks for refactoring logic, things are further simplified. Looks good to me, thanks!

Awesome! If those commits look good to you, feel free to cherry-pick those into the branch for this PR.

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
kerzhner: By "the exact amount of time" do you mean relying on timeouts for testing? If so, would mind describing a scenario you are thinking of?

packages/server-wallet/src/chain-service/chain-service.ts
kerzhner: I think the intuition to use unconfirmed events for nitro adjudicator is probably correct. Chain service waits for a confirmation of a funding event to protect against the following scenario:
1. Users A and B hold prefund states 0 and 1.
2. User A deposits on block 0.
3. User B deposits on block 1.
4. User A signs postfund 0.
5. User B signs postfund1 and starts the state channel application.
6. User B pays 1 eth to user A in the state channel.
5. There is a chain reorg, and block 1 is no longer a part of the chain. Note that now User B has not deposited their share of the funds. So the payment above cannot be redeemed.

packages/server-wallet/src/chain-service/event-tracker.ts
kerzhner: That makes sense. Would you mind leaving a comment explaining this? Also, do we need this default value for `assetOutcomeUpdated` function? Or should `logIndex` be a required parameter for this function?

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
isoosiss7: that's from your original question, we don't want to wait for 10s unless necessary.

I've added await to make sure mineBlockSlow finishes before test case exists, but I suspect that is not the only reason that test cases are affecting each others - chain seems to move faster when multiple test cases run together.

packages/server-wallet/src/chain-service/event-tracker.ts
isoosiss7: done

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
kerzhner: If we can assume that each test case waits for all of the blocks the test case mines, the hope is a test case does not affect other test cases. The chain state is not reset between test cases, but each test uses random channel ids.

That being said, perhaps there are other ways that tests interfere with each other. If you notice behavior like that, it would be great to hear about it!

kerzhner: FYI: we use the `--runInBand` jest flag which forces tests to execute sequentially.

kerzhner:
@isoosiss7 I would be curious to hear your feedback on 3b055411306721a9643870bd8963354a7ca82682...dcf9cfa1f3f7ab7c2130eed908a83b9818fcda23. The commit messages contain my reasoning for that work.

Specifically, it would be great to hear if dcf9cfa1f3f7ab7c2130eed908a83b9818fcda23 test refactor eliminates any test conditions that are needed to test this PR. In other words, does dcf9cfa1f3f7ab7c2130eed908a83b9818fcda23 make our tests less robust?

If you are happy with these commits, let's include them in this PR.

isoosiss7:
> @isoosiss7 I would be curious to hear your feedback on [3b05541...3746e9b](https://github.com/statechannels/statechannels/compare/3b055411306721a9643870bd8963354a7ca82682...3746e9b1cdcb1add44f1c951f854608e2e106ce2). The commit messages contain my reasoning for that work.
> 
> Specifically, it would be great to hear if [dcf9cfa](https://github.com/statechannels/statechannels/commit/dcf9cfa1f3f7ab7c2130eed908a83b9818fcda23) test refactor eliminates any test conditions that are needed to test this PR. In other words, does [dcf9cfa](https://github.com/statechannels/statechannels/commit/dcf9cfa1f3f7ab7c2130eed908a83b9818fcda23) make our tests less robust?
> 
> If you are happy with these commits, let's include them in this PR.

Merged, thanks!

This test case is to make sure the unconfirmed deposits (while getInitialHoldings) are played. Due to racing condition between ts code and the chain, it's hard to predict which deposit is unconfirmed when the test runs. So I had to run deposit 10 times to make sure one of them is in unconfirmed state while it runs.

Thanks for your optimization. I tried it a few times and it is stable.

kerzhner (approved):
Good job @isoosiss7!

PR #3301, opened 2021-02-15
Roll out waitingFor to all objective crankers

geoknee:
⚠️ For the reviewer: stacked PR (base not master).  This PR will be marked ready for review after #3286 is merged, but reviewers may wish to refer to it or even review it before then. 

- moves the transaction instigation up one level into the objective manager. This is pretty much the same as #3131 , but without introducing any objective locking. It could be considered preparatory work for that.
- individual crankers *return* a typed `waitingFor` string *each and every time they return*. Otherwise, they continue to make progress or throw an error
- the manager updates the objective's `waitingFor` property (and associated timestamp) as the final step in the transaction only if that property has changed
- ⚠️  the changeset is relatively large, mostly because many test files were changed and there were several whitespace changes. I recommend i) hiding whitespace changes when reviewing, and ii) to ignore any changes to `.test.ts` on your first pass.

### DIFF
--- a/packages/server-wallet/src/protocols/channel-closer.ts
+++ b/packages/server-wallet/src/protocols/channel-closer.ts
@@ -11,6 +11,10 @@ import {Channel} from '../models/channel';
 
 import {Defunder} from './defunder';
 
+export const enum WaitingFor {
+  TODO = 'TODO',
+}
+
 export class ChannelCloser {
   constructor(
     private store: Store,
@@ -32,7 +36,7 @@ export class ChannelCloser {
     objective: DBCloseChannelObjective,
     response: WalletResponse,
     tx: Transaction
-  ): Promise<void> {
+  ): Promise<WaitingFor> {
     const channelToLock = objective.data.targetChannelId;
     const channel = await this.store.getAndLockChannel(channelToLock, tx);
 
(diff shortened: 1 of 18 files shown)


packages/server-wallet/src/protocols/channel-closer.ts
```diff
@@ -11,6 +11,13 @@ import {Channel} from '../models/channel';
 
 import {Defunder} from './defunder';
 
+export const enum WaitingFor {
+  allAllocationItemsToBeExternalDestination = 'ChannelCloser.allAllocationItemsToBeExternalDestination',
+  theirFinalState = 'ChannelCloser.theirFinalState', // i.e. other participants' final states
+  defunding = 'ChannelCloser.defunding',
+  nothing = '',
```
lalexgap, line 18: Perhaps `nothing` could probably be defined at a higher level and shared between crankers?

geoknee: Yeah that's a good suggestion. The PR below this one in the stack #3286 has recently been updated to make  `waitingFor` optional. So we can set that field to `undefined` instead of `Enum.nothing` and not have the `nothing` option in the enums at all.

geoknee:
In favour of #3308

PR #3303, opened 2021-02-15
docs: replace dapp dev section with server dev

geoknee:
The current docs website has a section for browser based dapp developers. This PR:

-  adds a similar section for server app devs (copy/paste)
- removes any browser wallet specific content from the new pages
- unlinks the current section (docs still exist and can be reused when we resuscitate the browser wallet)

why? 

-  we can have richer documentation such as sequence diagrams. @isoosiss7 reported these were useful during onboarding, -  the current diagram (and surrounding text) references the out-of-date xstate wallet

TODO
- [ ] ~try and make the sequence diagram easier to read (it is a bit small)~ gave up, tried [this](https://mermaid-js.github.io/mermaid/#/theming?id=theme-variables-reference-table), mermaid js is buggy (I think(
- [x] firm up sequence: events vs api responses, name of various objects, etc.


geoknee:
https://602ab050983cfd439dc89983--nitro-protocol.netlify.app/docs/server-app-devs/make-api-calls

packages/docs-website/docs/docs/server-app-devs/make-api-calls.md
```diff
@@ -0,0 +1,82 @@
...
+B->>WalletB: pushMessage(msg2)
+WalletB-->>B: 0xabc: running
+WalletB-->>B: MessageQueued(msg3)
+B->>A: msg3;
+A->>WalletA: pushMessage(msg3);
+WalletA-->>A: 0xabc: running
+end
+loop 0...m
+note left of WalletB: Running a channel
+A->>WalletA: updateChannel(state-4+2m);
```
andrewgordstewart, line 53: I don't really understand the math here.
- Is it meant to be `4+2i`? When I view the diagram, it suggests that we update the channel `m` times, so there should probably be a loop counter variable `i` or something.
- If ^ this is true, maybe the math is worth clarifying, something like
```suggestion
A->>WalletA: updateChannel(state-(2i+3)+1);
```

andrewgordstewart (approved):
There should probably be a warning that these instructions are incomplete for "production use", but we are happy to work directly with people who want to use it today?

PR #3304, opened 2021-02-15
perf: save an unecessary query during joinChannel

geoknee:
Brings into line with `joinChannels` (plural). 

API calls should probably be "Objective-centric": what matters is whether you have the objective or not (you couldn't have the objective without already having the channel, anyway, due to DB constraints added during insert). 

The channel will be re-fetched, later, (in any case) when the objective is cranked.

### DIFF
--- a/packages/server-wallet/src/wallet/wallet.ts
+++ b/packages/server-wallet/src/wallet/wallet.ts
@@ -33,7 +33,6 @@ import {ValidationErrorItem} from 'joi';
 import {Bytes32} from '../type-aliases';
 import {createLogger} from '../logger';
 import * as UpdateChannel from '../handlers/update-channel';
-import * as JoinChannel from '../handlers/join-channel';
 import * as ChannelState from '../protocols/state';
 import {PushMessageError} from '../errors/wallet-error';
 import {timerFactory, recordFunctionMetrics, setupMetrics} from '../metrics';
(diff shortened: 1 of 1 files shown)


andrewgordstewart:
> API calls should probably be "Objective-centric": what matters is whether you have the objective or not (**you couldn't have the objective without already having the channel, anyway, due to DB constraints added during insert**).

Is this true? There is no FK-relationship that I could see that enforces this.

lalexgap:
> Is this true? There is no FK-relationship that I could see that enforces this.

I believe our ORM enforces this? I remember attempting to create a test where an objective refers to a nonexistent channel but it failed on insert.

andrewgordstewart:
> > Is this true? There is no FK-relationship that I could see that enforces this.
> 
> I believe our ORM enforces this? I remember attempting to create a test where an objective refers to a nonexistent channel but it failed on insert.

There is a table `objectives_channels`, which will not allow a row to be inserted if the channel does not exist. But the DB will not prevent you from inserting a row into `objectives`, as there is no FK constraint in the `objectives` table.

Our [_application_ code](https://github.com/statechannels/statechannels/blob/2e444b1b00be01b0c94427f6a5f7a1163df0971a/packages/server-wallet/src/models/objective.ts#L173-L181) does prevent `Objective.insert` from passing if the channel does not exist.

**Pedantic summary**:
A. Neither the DB nor the ORM prevents us from inserting objectives when the channels are not known.
B. `ObjectiveModel.insert` will not succeed if the "target channel" of the four `DBObjectives` is not stored.

lalexgap:
> **Pedantic summary**:

image

geoknee:
I think it is probably not worth pursuing this now, as the relationship between objectives and channels is probably going to change soon anyway, and we will hopefully know better then, when we need to do checks like this. For instance: maybe we should do the check when approving the objective. 

The discrepancy between `joinChannel` and `joinChannels` remains, for now.

PR #3308, opened 2021-02-16
introduce waitingFor column for objectives

geoknee:
This PR contains #3286 plus post-review revisions, as well as #3301 with conflicts resolved.

**3286 Description**
⚠️ This PR is the lowest in a stack. Please do not squash- or rebase-merge it. The next PR up in the stack is #3301 .

We are building a mechanism to detect a stalled objective, and have the `progressLastMadeAt` timestamp to help with this. I have played around with a few different approaches for when to hook code into updating that timestamp #3264 #3278. In this PR, I have settled on something that makes good sense to me, with some input from @tomclose . 

To properly detect when progress has been made towards an objective when the run loop is triggered, we are going to need some sort of record or memory of state. In defining that state, we don't want to have to query and inspect all channels in scope of the objective; but we want to introduce the minimal amount of state replication, since the store should remain the source of truth. 

The idea of this PR is to concentrate on progress being *blocked* by events that my wallet has non control over. It doesn't make sense, for example, to report that we are blocked on signing our own pre fund state... we should just go ahead and do that, and anything else we are safe to do, until we reach a point where we have to wait for someone or something else. So these are the points where we should have state boundaries. 

The "states" of an objective are then just names for the next thing that blocks us from making progress (and that we have no direct control over). These are cached on the objective itself, and used to detect when progress is made. By storing this information in a column called `waitingFor` (rather than `state`, say), I am trying to discourage reading this name for the purposes of controlling protocol logic itself: it is just metadata for capturing progress. 

This PR introduces the required changes to the DB schema and ORM, along with a simple test and a single example in the `ChannelOpener.crank()` function. What's nice is that each time the cranker can infer that the objective is externally blocked, it does an early `return`. If we don't do the early `return`, immediately after we cross a state boundary and cache a new `waitingFor` string.  If, when we have finished cranking, the `waitingFor` string changed, we know that we made meaningful progress (and not just the boring progress we can make unilaterally). 

If we like this pattern, I think it should be relatively easy to roll out to the rest of the objective crankers.    

A conjecture:
_The progress of an objective can be linear, even if the events that drive that progress arrive in an unpredictable order__                                                   
                                                         

```                                                         
 first   second                                          
 crank   crank                    |                    | 
    |    |                        |                    | 
    |    |     waitingFor first   | waitingFor second  | 
    |    |     blocker            | blocker            | 
    |    |                        |                    | 
    |    |          +---------+   |                    | 
    +----|--------->| BLOCKED |   |                    | 
         |          +---------+   |                    | 
         |                |       |     +---------+    | 
         +------------------------|-----> BLOCKED |    |  etc
                          |       |     +---------+    | 
                          |       |         |          | 
                          v       |         v          | 
                        wait      |       wait         | 
                                  |                    |
```

## Changes 

1. new column in objectives table
2. new static method `updateWaitingFor` , which updates the `progressLastMadeAt` timestamp 
  - this could probably become an instance method
  - we could consider only updating the timestamp in the event of a nontrivial db patch
 3. `ChannelOpener.crank()` rewritten both for clarity and to adopt the new pattern of recording what is currently blocking the objective. 

## How Has This Been Tested? 

I manually verified (with console.log) that each blocker was logged out in turn during our `open-and-close-channel` tests.

**#3301 description** 

- moves the transaction instigation up one level into the objective manager. This is pretty much the same as #3131 , but without introducing any objective locking. It could be considered preparatory work for that.
- individual crankers *return* a typed `waitingFor` string *each and every time they return*. Otherwise, they continue to make progress or throw an error
- the manager updates the objective's `waitingFor` property (and associated timestamp) as the final step in the transaction only if that property has changed

---
### Code quality
### Project management
----
EDIT
Have asked @andrewgordstewart and @lalexgap both for a careful review of this changeset, and of the stacked changes #3301, and to pay particular attention to 

- any migrations that won't run in production
- any extra unnecessary work that might be generated by adopting this approach


packages/server-wallet/src/protocols/challenge-submitter.ts
```diff
@@ -27,55 +34,56 @@ export class ChallengeSubmitter {
...
-        await this.store.markObjectiveStatus(objective, 'failed', tx);
-        return;
-      }
+    if (channel.initialSupport.length === 0) {
+      this.logger.error(
+        {channelId: channel.channelId},
+        'There is no initial support stored for the channel'
+      );
+      await this.store.markObjectiveStatus(objective, 'failed', tx);
+      return null;
```
lalexgap, line 78: Is there a reason we're using `null` not `undefined`?

packages/server-wallet/src/protocols/channel-opener.ts
```diff
@@ -29,44 +34,48 @@ export class ChannelOpener {
...
+    if (!channel.postFundComplete) return WaitingFor.theirPostFundState;
 
-        response.queueSucceededObjective(objective);
-      }
+    objective = await this.store.markObjectiveStatus(objective, 'succeeded', tx);
+    response.queueSucceededObjective(objective);
 
-      response.queueChannel(channel); // always queue the channel if we've potentially touched it
-    });
+    return null;
```
lalexgap, line 78: I personally prefer returning something like `WaitingFor.nothing`.

packages/server-wallet/src/__test__/create-and-close-channel/direct-funding.test.ts
```diff
@@ -93,6 +94,11 @@ it('Create a directly funded channel between two wallets ', async () => {
     turnNum: 0,
   });
 
+  // Let's check what wallet a is waiting for. Expect it to be 'theirPreFundSetup' or ''
+  // (if the objective has never been cranked)
+  const waitingForBefore = (await a.getObjective(`OpenChannel-${channelId}`)).waitingFor;
+  expect([WaitingFor.theirPreFundSetup, '']).toContain(waitingForBefore);
```
lalexgap, line 100: Is `''''` a valid value anymore?

packages/server-wallet/src/models/objective.ts
```diff
@@ -139,6 +144,7 @@ export class ObjectiveModel extends Model {
           data: objectiveToBeStored.data,
           createdAt: new Date(),
           progressLastMadeAt: new Date(),
+          waitingFor: objectiveToBeStored.waitingFor, // the return value is null
```
lalexgap, line 147: I'm not sure I understand this comment?

packages/server-wallet/src/protocols/challenge-submitter.ts
geoknee: Yes: querying the database may return `null`. If we want to use `undefined`, I believe we would need to handle conversion from `null` to undefined. I felt that it was safer and easier to stick with `null`. Happy to discuss though.

packages/server-wallet/src/models/objective.ts
geoknee: Whoops, looks like I never finished writing this comment. Probably I should explain here the choice of `null` over `undefined`.

packages/server-wallet/src/__test__/create-and-close-channel/direct-funding.test.ts
geoknee: Good catch, this should be changed to `null`.

packages/server-wallet/src/protocols/challenge-submitter.ts
lalexgap: I think I'd prefer to keep `null` types restricted to the data layer only. We generally don't use the `null` type throughout our codebase and I could see a bug easily being introduced if someone adds a check like `o.waitingFor === undefined`.

lalexgap (approved):
I like this pattern! While this does add some overhead to writing a protocol I think that is outweighed by the clarity it provides!

PR #3311, opened 2021-02-17
increase clarity around ObjectiveModel type

geoknee:
Make our current type conversion more explicit and transparent, to aid developer understanding. 

Currently, there are some useful things to know which are not self evident from the code. These changes attempt to remedy that, and I think it will help us judge whether we actually want the current behaviour or not. 

These useful things are:
- we have one DB table that stores all objectives in the *union type* `Objective`. We also have a single ORM class called `ObjectiveModel`. It is not possible to make a union of classes. This implies that the correlations in the union type are washed out when we deal with `ObjectiveModels`. As an example, we could have an `ObjectiveModel` with `type: 'OpenChannel'` but the incompatible `data: submitChallengeData`. The static `insert` method is typed in such a way as to prevent that, but otherwise it is not impossible because no such typing is enforced by the db itself. For example, we can just write a query that inserts such a "bad" objective. 
- The instance method `toObjective()` (which is heavily used) does some typecasting to convert back to a proper union type with the correlations, but this isn't as obvious as it might be. 
- we drop the `participants` property when storing an objective, and add it back again (using `[]`) when reading an objective

Other changes:
I added a generic type to the `toObjective()` function, so that it will be possible to easily narrow the type even further if we want to. I anticipate this will be useful for future refactors. I can split that change onto a different PR if the reviewer prefers. 

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/models/objective.ts
+++ b/packages/server-wallet/src/models/objective.ts
@@ -89,7 +89,10 @@ export class ObjectiveChannelModel extends Model {
   }
 }
 
-export class ObjectiveModel extends Model {
+type FlattenedDBObjective = {[P in keyof DBObjective]: DBObjective[P]}; // this destroys the correlations in the union type
+type Columns = Omit<FlattenedDBObjective, 'participants'>;
+
+export class ObjectiveModel extends Model implements Columns {
   readonly objectiveId!: DBObjective['objectiveId'];
   readonly status!: DBObjective['status'];
   readonly type!: DBObjective['type'];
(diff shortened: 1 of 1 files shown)


andrewgordstewart:
>     * we have one DB table that stores all objectives in the _union type_ `Objective`. We also have a single ORM class called `ObjectiveModel`. It is not possible to make a union of classes. This implies that the correlations in the union type are washed out when we deal with `ObjectiveModels`. As an example, we could have an `ObjectiveModel` with `type: 'OpenChannel'` but the incompatible `data: submitChallengeData`. 

Classes can be generic over a type, which means we Can prevent this in the type system. It will still be a bit of a lie, because the data in the DB can be arbitrary

geoknee:
Yeah you are right about the generic class thing. Perhaps we should give that a go?

packages/server-wallet/src/models/objective.ts
```diff
@@ -247,11 +250,25 @@ export class ObjectiveModel extends Model {
...
     };
+    switch (this.type) {
+      case 'CloseChannel':
+      case 'DefundChannel':
+      case 'OpenChannel':
+      case 'SubmitChallenge':
+        return withParticipants as O;
+      default:
+        throw Error('unimplemented');
+    }
```
geoknee, line 272: In particular, I think this sort of thing can obsolete `ensureObjectives` in the `store`, which would save much code.

geoknee:
This was mostly a discussion piece, and will be superseded by #3315  and #3320 .

PR #3312, opened 2021-02-17
chore: change node version to 12.18.0

NiloCK:
version 2.18.0 does not exist - was a typo

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -50,7 +50,7 @@ This repository is a monorepo, and contains the following packages maintained wi
 
 ### Installing dependencies
   * **yarn** - version 1.22.4, for easy management of specific Yarn versions, we recommend using [Yarn Version Manager (YVM)](https://github.com/tophat/yvm).
-  * **node** - version 2.18.0, for easy management of specific versions, we recommend using [n](https://github.com/tj/n).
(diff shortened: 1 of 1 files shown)


geoknee (approved):
Good spot!

readme.md
```diff
@@ -50,7 +50,7 @@ This repository is a monorepo, and contains the following packages maintained wi
 
 ### Installing dependencies
   * **yarn** - version 1.22.4, for easy management of specific Yarn versions, we recommend using [Yarn Version Manager (YVM)](https://github.com/tophat/yvm).
-  * **node** - version 2.18.0, for easy management of specific versions, we recommend using [n](https://github.com/tj/n).
+  * **node** - version 12.18.0, for easy management of specific versions, we recommend using [n](https://github.com/tj/n).
```
andrewgordstewart, line 53:
```suggestion
  * **node** - version 12.16.3, for easy management of specific versions, we recommend using [n](https://github.com/tj/n).
```

We've got `12.16.3` in our `.nvmrc`, and specify it to be used by some github actions. So, I think we should stick to that in the readme!

andrewgordstewart (approved):
Suggested a change to the "blessed" version, but thanks!

PR #3313, opened 2021-02-17
Only return newObjectives if the objective is actually new

lalexgap:
# Description
Only return newObjectives if the objective is actually new.

`ensureObjective` now returns `isNew` to indicate whether the objective existed before attempting to store it.  The `wallet` only adds to `newObjectives` if `isNew===true`

- The commented out checks in sync-objective have been re-enabled
- New tests have been added for the new `doesExist` method on the `ObjectiveModel`

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/wallet/store.ts
+++ b/packages/server-wallet/src/wallet/store.ts
@@ -23,6 +23,7 @@ import {
   SubmitChallenge,
   isSubmitChallenge,
   DefundChannel,
+  objectiveId as calculateObjectiveId,
 } from '@statechannels/wallet-core';
 import {Payload as WirePayload, SignedState as WireSignedState} from '@statechannels/wire-format';
 import _ from 'lodash';
@@ -42,14 +43,7 @@ import {timerFactory, recordFunctionMetrics, setupDBMetrics, timerFactorySync} f
 import {isReverseSorted, pick} from '../utilities/helpers';
 import {Funding, TransferredOutEntry} from '../models/funding';
 import {Nonce} from '../models/nonce';
-import {
-  ObjectiveModel,
-  DBObjective,
-  DBCloseChannelObjective,
-  DBOpenChannelObjective,
-  DBSubmitChallengeObjective,
-  DBDefundChannelObjective,
-} from '../models/objective';
+import {ObjectiveModel, DBObjective, ObjectiveToStore} from '../models/objective';
 import {AppBytecode} from '../models/app-bytecode';
 import {LedgerRequest, LedgerRequestType} from '../models/ledger-request';
 import {shouldValidateTransition, validateTransition} from '../utilities/validate-transition';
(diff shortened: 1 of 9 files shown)


andrewgordstewart (changes requested):
Blocking this temporarily, while I try to answer this question
> If I call `closeChannel` twice, will `newObjectives` be empty the second time?

Some comments: 
1. I think it the objective should be returned the 2nd time. Hence, `newObjectives` is probably a bad name.
2. I _don't_ think `'objectiveStarted'` should be emitted the second time.

packages/server-wallet/src/wallet/store.ts
```diff
@@ -504,13 +500,23 @@ export class Store {
...
+    return this.saveObjective(objectiveToBeStored, tx);
+  }
+
+  private async saveObjective(
+    objective: ObjectiveToStore,
+    tx: Transaction
+  ): Promise<EnsureObjectiveResult> {
+    const objectiveId = calculateObjectiveId(objective);
+    const isNew = !(await ObjectiveModel.doesExist(objectiveId, tx));
+    const storedObjective = await ObjectiveModel.insert(objective, tx);
```
andrewgordstewart, line 512: I don't really think this is the best pattern.
1. It's two queries, which is inefficient.
2. It's not obviously correct.

though I think (2) might be wrong, it might obviously be correct?


Did you consider turning `insert` into an `upsert`? It can return a value that indicates that the row was already present, something like
```
INSERT INTO blah
VALUES (foo)
ON CONFLICT (...) DO NOTHING
RETURNING *, trueWhenRowDidExist
```

andrewgordstewart: There is a relatively succinct hack [here](https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql) whose tradeoff might not apply:
```
INSERT INTO blah
VALUES (foo)
ON CONFLICT (...) DO UPDATE SET id = excluded.id
RETURNING *, trueWhenRowDidExist
```

I don't know how to compute `trueWhenRowDidExist` yet. You could compare `existing.created_at` with `now()`, which might result in a rare false positive.

lalexgap: > I don't really think this is the best pattern.
> 
> 1. It's two queries, which is inefficient.
> 2. It's not obviously correct.
> 
> though I think (2) might be wrong: it might obviously be correct within a transaction, but it might throw a surprising error (which this PR does not seem to account for)

I think it's good enough to be correct in the transaction? Shouldn't any error thrown in a transaction cause us to rollback? Is there a particular error case you're thinking about?

> Did you consider turning `insert` into an `upsert`? It can return a value that indicates that the row was already present, something like
> 
> ```
> INSERT INTO blah
> VALUES (foo)
> ON CONFLICT (...) DO NOTHING
> RETURNING *, trueWhenRowDidExist
> ```

I went with a separate query as it seemd more straightforward than using an `upsert`. (I would argue all the caveats mentioned [here](https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql) are a point in that favor)

That said we might want to go with `upsert` for performance reasons.

lalexgap:
Switching to a `draft`  as this may no longer be needed with #3315. The plan is to wait and see where this fits in once #3315 is complete.

lalexgap:
Closing this as it's been superseded by @geoknee's work

Issue #3205, opened 2021-01-26
Add concludeTransactionSubmitter to the Defund objective


kerzhner:
Add `concludeTransactionSubmitter` to the Defund objective.
- For now, this objective is always set to the indexer.
- In the long term, the gateway and the indexer may choose to negotiate this field.

Issue #3206, opened 2021-01-26
Defund protocol only submits a transaction for concludeTransaction Submitter


kerzhner:
Modify the Defund protocol to submit a concludeAndWithdraw transaction only when my participant id matches concludeTransactionSubmitter

Issue #3217, opened 2021-01-26
Collaborative conclude with fair gas fee contributions by Gateway and Indexer


kerzhner:
https://www.notion.so/Conclude-gas-fees-be891231bc8542aeaf5e97dd3ece2cd0#fe4de4d4af51491ba6d6dfaf8dd00074

PR #3314, opened 2021-02-17
Introduce deterministic transaction submission

kerzhner:
Assume a 2-party, directly funded channel, though the following easily generalizes to `n` parties.

Once participants A and B hold a conclusion proof, the following sequence should occur:
1. The channel should be concluded on-chain.
1. The channel outcome should be pushed to all AssetHolders.
1. In each AssetHolder, funds should be transferred to each participant.

The above can be done in 1 or many transactions. Regardless of how many transactions are mined, at least 1 transaction is a "collaborative" transaction: a transaction that benefits both parties, but only one party pays gas fees.

This PR introduces a transaction submitter order for collaborative conclusions. The order is agreed upon when the collaborative conclusion objective is created. At the moment, the order is hardcoded as `[1, 0]`, which means that the second participant is the primary transaction submitter and the second participant is the secondary transaction submitter.

If a participant has no funds in a channel, the participant is not incentivized to submit collaborative transactions. If a participant is the primary transaction submitter but has no funds in the channel, the secondary transaction submitter is expected to submit a transaction. Same logic applies to the tertiary transaction submitter, etc.

Fixes https://github.com/statechannels/statechannels/issues/3205, https://github.com/statechannels/statechannels/issues/3206.

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -510,6 +510,13 @@ export class Channel extends Model implements ChannelColumns {
     return this.participants.length;
   }
 
+  nthParticipant(n: number): Participant {
+    if (n >= this.nParticipants) {
+      throw new Error(`Cannot get participant ${n} from ${this.participants}`);
+    }
+    return this.participants[n];
+  }
+
   private get _support(): Array<SignedStateWithHash> {
     // TODO: activate these fields for proper application checks (may be resource hungry)
     const logger = undefined;
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -440,7 +440,7 @@ export class Channel extends Model implements ChannelColumns {
   private isDirectFunded(threshold: 'FullyFunded' | 'PartlyFunded'): boolean {
     const outcome = this.supported?.outcome;
     if (!outcome) {
-      throw new Error(`Channel passed to isDirectFunded has no supported state`);
+      throw new ChannelError(ChannelError.reasons.noSupportedState);
     }
 
     const {assetHolderAddress, allocationItems} = checkThat(outcome, isSimpleAllocation);
@@ -476,17 +476,15 @@ export class Channel extends Model implements ChannelColumns {
   } {
     const supported = this.supported;
     if (!supported) {
-      throw new Error(`Channel has no supported state`);
+      throw new ChannelError(ChannelError.reasons.noSupportedState);
     }
     const {allocationItems} = checkThat(supported.outcome, isSimpleAllocation);
 
     const myAllocationItem = this.allocationItemForParticipantIndex(this.myIndex);
     if (!myAllocationItem) {
-      throw new Error(
-        `My destination ${
-          this.participants[this.myIndex].destination
-        } is not in allocations ${allocationItems}`
-      );
+      throw new ChannelError(ChannelError.reasons.destinationNotInAllocations, {
+        destination: this.participants[this.myIndex].destination,
+      });
     }
 
     const myDestination = this.participants[this.myIndex].destination;
@@ -509,13 +507,14 @@ export class Channel extends Model implements ChannelColumns {
    */
   allocationItemForParticipantIndex(index: number): AllocationItem | undefined {
     if (index >= this.participants.length) {
-      throw new Error(
-        `${index} is greater than number of participants ${this.participants.length}`
-      );
+      throw new ChannelError(ChannelError.reasons.indexNotInChannel, {
+        index,
+        numChannelParticipant: this.participants.length,
+      });
     }
     const supported = this.supported;
     if (!supported) {
-      throw new Error(`Channel has no supported state`);
+      throw new ChannelError(ChannelError.reasons.noSupportedState);
     }
 
     const myDestination = this.participants[index].destination;
@@ -603,7 +602,11 @@ export class ChannelError extends WalletError {
   static readonly reasons = {
     invalidChannelId: 'Invalid channel id',
     incorrectHash: 'Incorrect hash',
-    channelMissing: 'No channel found with id.',
+    channelMissing: 'No channel found with id',
+    noSupportedState: 'The channel is expected to contain supported state',
+    indexNotInChannel:
+      'The participant index is greater than the number of participants in the channel',
+    destinationNotInAllocations: 'The destination is not in the channel allocatons',
   } as const;
 
   constructor(reason: Values<typeof ChannelError.reasons>, public readonly data: any = undefined) {
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -502,7 +502,7 @@ export class Channel extends Model implements ChannelColumns {
   }
 
   /**
-   * The below logic assumes:
+   * WARNING: The below logic assumes
    *  1. Each destination occurs at most once.
    *  2. We only care about a single destination.
    * One reason to drop (2), for instance, is to support ledger top-ups with as few state updates as possible.
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -14,6 +14,7 @@ import {
   isSimpleAllocation,
   BN,
   unreachable,
+  AllocationItem,
 } from '@statechannels/wallet-core';
 import {JSONSchema, Model, Pojo, QueryContext, ModelOptions, TransactionOrKnex} from 'objection';
 import {ChannelResult, FundingStrategy} from '@statechannels/client-api-schema';
@@ -473,24 +474,22 @@ export class Channel extends Model implements ChannelColumns {
     targetAfter: Uint256;
     targetTotal: Uint256;
   } {
-    /**
-     * The below logic assumes:
-     *  1. Each destination occurs at most once.
-     *  2. We only care about a single destination.
-     * One reason to drop (2), for instance, is to support ledger top-ups with as few state updates as possible.
-     */
     const supported = this.supported;
     if (!supported) {
-      throw new Error(`Channel passed to DriectFunder has no supported state`);
+      throw new Error(`Channel has no supported state`);
+    }
+    const {allocationItems} = checkThat(supported.outcome, isSimpleAllocation);
+
+    const myAllocationItem = this.allocationItemForParticipantIndex(this.myIndex);
+    if (!myAllocationItem) {
+      throw new Error(
+        `My destination ${
+          this.participants[this.myIndex].destination
+        } is not in allocations ${allocationItems}`
+      );
     }
 
     const myDestination = this.participants[this.myIndex].destination;
-    const {allocationItems} = checkThat(supported.outcome, isSimpleAllocation);
-
-    const myAllocationItem = _.find(allocationItems, ai => ai.destination === myDestination);
-    if (!myAllocationItem) {
-      throw new Error(`My destination ${myDestination} is not in allocations ${allocationItems}`);
-    }
 
     const allocationsBefore = _.takeWhile(allocationItems, a => a.destination !== myDestination);
     const targetBefore = allocationsBefore.map(a => a.amount).reduce(BN.add, BN.from(0));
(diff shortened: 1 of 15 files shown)


packages/wallet-core/src/types.ts
```diff
@@ -107,6 +107,23 @@ export type CloseChannel = _Objective<
   {
     targetChannelId: string;
     fundingStrategy: FundingStrategy;
+    /**
+     * Collaboratively concluding a channel on-chain involves the following:
+     * 1. Submit a conclusion proof to the NitroAdjudicator.
+     * 2. Push the outcome to all AssetHolders.
+     * 3. Transfer out of the channel for all asset holders.
```
andrewgordstewart, line 114: Do you expect `tsSubmitterOrder[0]` will be responsible for all of step (3)? Or do you imagine an iterative approach, ie.
```
for (const idx of tsSubmitterOrder) {
  // participant[idx] withdraws their assets until they have no funds in the channel
}
```

packages/server-wallet/src/protocols/defunder.ts
```diff
@@ -126,4 +153,30 @@ export class Defunder {
       tx
     );
   }
+
+  /**
+   * Assume that we are going to submit collaborative transaction(s) unless:
+   * 1. There is an agreed upon order of transaction submitters.
+   * 2. There is a transaction submitter before us who has funds in the channel.
+   *
+   * This method is public for unit testing only
```
andrewgordstewart, line 162: One way to accomplish this, while maintaining privacy, is making it a `protected` function that can be overridden as public by a `TestDefunder` class.

But, why is it a `static` method, instead of just a function?

packages/server-wallet/src/models/channel.ts
```diff
@@ -502,6 +501,29 @@ export class Channel extends Model implements ChannelColumns {
...
+      throw new Error(
+        `${index} is greater than number of participants ${this.participants.length}`
+      );
+    }
+    const supported = this.supported;
+    if (!supported) {
+      throw new Error(`Channel has no supported state`);
+    }
+
+    const myDestination = this.participants[index].destination;
```
andrewgordstewart, line 521: I'll highlight that this code makes assumptions that we have multiple reasons to break:
1. In a unidirectional relationship, if the payer's `destination` is last in the list, then we can do special-case ledger top-ups for free.
2. In The Graph Network, we would like for a payment channel to serve multiple allocations. This will imply that a ledger channel will have multiple destinations for participant `1`.

Of course, a significant number of changes in the wallet are required to support either of these features.

packages/server-wallet/src/objectives/close-channel.ts
```diff
@@ -16,7 +16,11 @@ export class CloseChannelObjective {
           {
             type: 'CloseChannel',
             participants: [],
-            data: {targetChannelId: channelId, fundingStrategy: channel.fundingStrategy},
+            data: {
+              targetChannelId: channelId,
+              fundingStrategy: channel.fundingStrategy,
+              txSubmitterOder: [1, 0],
```
andrewgordstewart, line 22: I suppose you are asserting (paraphrasing from #3205)
> For now, the `txSubmitterOrder` is always set to `[1, 0]`

⚠️ It is spelled wrong throughout this PR.

```suggestion
              txSubmitterOrder: [1, 0],
```

packages/server-wallet/src/protocols/__test__/defunder.test.ts
```diff
@@ -26,6 +27,30 @@ const testLedgerChannel = TestChannel.create({
...
+      channel.closeChannelObjective([participantIndex, Math.abs(participantIndex - 1)]),
+      tx
+    );
+    await store.approveObjective(o.objectiveId, tx);
+    return o as DBCloseChannelObjective;
+  });
+}
+
+function ensureDefundObjective(channel: TestChannel): Promise<DBDefundChannelObjective> {
+  // add the closeChannel objective and approve
```
andrewgordstewart, line 46:
```suggestion
  // add the defundChannel objective and approve
```

packages/server-wallet/src/protocols/__test__/defunder.test.ts
```diff
@@ -26,6 +27,30 @@ const testLedgerChannel = TestChannel.create({
...
 let store: Store;
 
+function ensureCloseObjective(
+  channel: TestChannel,
+  participantIndex = 0
+): Promise<DBCloseChannelObjective> {
+  // add the closeChannel objective and approve
+  return store.transaction(async tx => {
+    const o = await store.ensureObjective(
+      channel.closeChannelObjective([participantIndex, Math.abs(participantIndex - 1)]),
```
andrewgordstewart, line 37: This is always equivalent to this, right?
```suggestion
      channel.closeChannelObjective([1, 0]),
```

packages/server-wallet/src/models/channel.ts
```diff
@@ -473,25 +474,23 @@ export class Channel extends Model implements ChannelColumns {
...
 
-    const myAllocationItem = _.find(allocationItems, ai => ai.destination === myDestination);
+    const myAllocationItem = this.allocationItemForParticipantIndex(this.myIndex);
     if (!myAllocationItem) {
-      throw new Error(`My destination ${myDestination} is not in allocations ${allocationItems}`);
+      throw new Error(
+        `My destination ${
+          this.participants[this.myIndex].destination
+        } is not in allocations ${allocationItems}`
+      );
```
andrewgordstewart, line 489: I don't think it's a best practice to throw an error with an unbounded data structure in it like this?

There are some options:
```
// 1. 
const err = new Error('My destination is missing from allocationItems');
logger.error({err, myDestination: this.participants[this.myIndex].destination, allocationItems});
throw err;
```

2. Extend the `Error` class with an error that can store properties.

(2) is nice, because
A. The error message is consistent. This has many benefits! (easy grepping, error analytics, etc.)
B. pino's StdErrSerializer, which invoked when you log an object with keyword `err`, will still log all the properties + the stack trace

It's also the case that the error message you're throwing is making assumptions about the implementation of `allocationItemForParticipantIndex`. So, if `allocationItemForParticipantIndex` changes, your error message might now be lying.

packages/server-wallet/src/protocols/__test__/defunder.test.ts
```diff
@@ -37,6 +62,59 @@ beforeEach(async () => {
...
+      type: 'CloseChannel',
+      data: {txSubmitterOder: [1, 0]},
+    } as DBCloseChannelObjective;
+    expect(Defunder._shouldSubmitCollaborativeTx(channel, objective2)).toEqual(false);
+
+    const objective3 = ({
+      type: 'CloseChannel',
+      data: {txSubmitterOder: []},
+    } as unknown) as DBCloseChannelObjective;
+    expect(Defunder._shouldSubmitCollaborativeTx(channel, objective3)).toEqual(true);
```
andrewgordstewart, line 89: There's a lot of repeated code here. Is it worth defining a simple helper?
```
function closeObjective(txSubmitterOrder: number[]): DBCloseChannelObjective {
  return {
      type: 'CloseChannel',
      data: {txSubmitterOder},
    } as DBCloseChannelObjective;
}
```

I see the pattern repeated below as well.

packages/server-wallet/src/protocols/defunder.ts
```diff
@@ -79,8 +89,21 @@ export class Defunder {
       tx,
       channel.channelId
     );
+
     let didSubmitTransaction = false;
-    if (adjudicatorStatus.channelMode !== 'Finalized' && channel.hasConclusionProof) {
+    const shouldSubmitCollaborativeTx = Defunder._shouldSubmitCollaborativeTx(channel, objective);
+
+    /**
+     * The below if/else does not account for the following scenario:
```
andrewgordstewart, line 97: Am I right to interpret that this is a TODO? 

```suggestion
     * TODO: The below if/else does not account for the following scenario:
```

packages/server-wallet/src/protocols/__test__/defunder.test.ts
```diff
@@ -37,6 +62,59 @@ beforeEach(async () => {
   );
 });
 
+describe('Collaborative transaction submitter', () => {
```
andrewgordstewart, line 65: I don't see a test case for:
- `txSubmitterOrder === [0, 1]`
- `participants[0]` has no stake in the outcome.

andrewgordstewart (approved):
This looks good! Though I think there are still some things to ponder over.

packages/server-wallet/src/models/channel.ts
```diff
@@ -502,6 +501,29 @@ export class Channel extends Model implements ChannelColumns {
     return {targetBefore, targetAfter, targetTotal};
   }
 
+  /**
+   * The below logic assumes:
+   *  1. Each destination occurs at most once.
+   *  2. We only care about a single destination.
```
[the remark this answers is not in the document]
kerzhner: 3a3405524

packages/wallet-core/src/types.ts
kerzhner: In the medium term, it likely makes sense for the primary transaction submitter to only withdraw their own assets. In the short term, the primary transaction submitter is responsible for all 3 steps.

andrewgordstewart: I'm not sure if we need to write down the medium term goal. In fact, I think we [already have](https://github.com/statechannels/statechannels/issues/3217)!

packages/server-wallet/src/protocols/defunder.ts
kerzhner: Good point! b3878c317

packages/server-wallet/src/objectives/close-channel.ts
kerzhner: Good catch! c39a36102

packages/server-wallet/src/protocols/__test__/defunder.test.ts
kerzhner: What if `participantIndex` is `0`. Then `txSubmitterOrder` is `[0, 1]`.

andrewgordstewart: ... right

it's always equivalent to 
```suggestion
      channel.closeChannelObjective([participantIndex, 1 - participantIndex]),
```

which was going to be my suggestion until I confused myself!

packages/server-wallet/src/protocols/__test__/defunder.test.ts
kerzhner: Good call 03dfd2d6c

packages/server-wallet/src/protocols/__test__/defunder.test.ts
kerzhner: Good point c951cef7f

packages/server-wallet/src/models/channel.ts
kerzhner: Just to be clear, your suggestion 2 does not get around the fact that if `allocationItems` is included in the error, an unbonded data structure is thrown. Is that correct?

I am not sure I understand the last part of your comment. The helper method `allocationItemForParticipantIndex` fetches an allocation item (maybe multiple in the future) for a participant. How would a helper implementation change make this error invalid?

kerzhner: 6416aec61 uses ChannelError instead of the base Error class.

packages/server-wallet/src/protocols/defunder.ts
kerzhner: It can be a TODO. Though a wallet that correctly treats the scenario described in the comment requires more changes than to just the code below the comment.

packages/server-wallet/src/protocols/__test__/defunder.test.ts
kerzhner: Good point! 042749835

kerzhner:
@andrewgordstewart thank you for the excellent feedback! I believe I addressed all of the comments on the PR.

packages/server-wallet/src/models/channel.ts
andrewgordstewart: > Just to be clear, your suggestion 2 does not get around the fact that if `allocationItems` is included in the error, an unbonded data structure is thrown. Is that correct?

That's a good point. However,
1. I think it's bad to use an _unbounded_ data structure in the error message, because humans read error messages. For instance, [graph-pino](https://www.npmjs.com/package/@graphprotocol/graph-pino) attempts to make pino logs human readable. When logging an error, it would displays `err.message`, and including `allocationItems` in `err.message` will probably make it unreadable.
2. Another negative of using _any_ data structure is that _computers_ read error messages. For instance, Sentry groups errors by message, and including a data structure in the message makes it difficult to track the grouped errors.

kerzhner: Both are good points!

PR #3321, opened 2021-02-19
introduce getAndLockObjective

geoknee:
and use in ObjectiveManager.

See #3131


lalexgap (approved):
This seems reasonable to me. Are there any deadlock issues we should be worried about?

geoknee:
I think as long as each transaction locks only a single row for update, there shouldn't be any deadlocks. 

When we come to locking multiple channels for update, we are going to need to think more carefully. For objectives, however, it should just be one objective at a time. 

 Since @tomclose raised this question on the original PR, I've requested a review from him, too.

tomclose (approved):

PR #3322, opened 2021-02-19
fix: do not use system user for postgres

kerzhner:
The below issue was flagged when attempting to get @NiloCK's Ubuntu machine to pass all wallet tests.

Database manipulations occur either via:
1. Postgres db commands such as `createdb`/`dropdb`
1. `knex`

Different postgres users are used throughout the codebase:
1. In some cases `postgres` user is hardcoded.
2. In some cases, server wallet database configuration is respected. This happens when `knex` is used for database manipulations. The default user for the server wallet database configuration is `postgres`. 
3. In some cases, the system user is used.

This PR eliminates case 3.

### DIFF
--- a/packages/server-wallet/package.json
+++ b/packages/server-wallet/package.json
@@ -79,8 +79,8 @@
   },
   "main": "lib/src/index.js",
   "scripts": {
-    "db:create": "createdb $SERVER_DB_NAME",
-    "db:drop": "dropdb $SERVER_DB_NAME",
+    "db:create": "createdb $SERVER_DB_NAME -U postgres",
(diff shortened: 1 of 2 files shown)


andrewgordstewart (dismissed):
I'm reserving the right to review this on Monday morning.

lalexgap:
> I'm reserving the right to review this on Monday morning.

@andrewgordstewart just flagging you to take a look.

NiloCK (approved):
This looks good and I believe it succeeds in removing a troubleshooting step when configuring a new linux machine.

PR #3324, opened 2021-02-22
Introduce check on chainId during store.addSignedState.

geoknee:
Currently, the first we learn of an incorrect chain id at `challenge` or `conclude` time. That’s far too late, as by that point we may have unwittingly locked up funds against the channel on “the wrong chain”, causing the twin problems of
- a running but unfunded channel (I might provide goods or services for “payments” that aren’t actually payments)
- funds deposited against a channel that hasn’t even been opened (no states and therefore no way to unlock the funds)

This approach is good because it performs the required check very early on. The check is against the wallet's passed-in configuration object. The single chain specified there might be a ganache server during testing, or ethereum mainnet in production. No check is necessary during channel creation, as the chainid is correct by construction in that case.

Downsides of this PR: a new error is thrown without any error-handling logic in place for it. I think this is OK for now.

Alternatives considered:
a) making the check mandatory further down the stack in nitro protocol or wallet core. I actually started off with this but decided that it was unnecessarily restrictive. The docs currently suggest that this is really a responsibility deferred to the wallet.
b) doing the check inside `validTransition` or elsewhere. I think this wouldn’t make the most sense: the on chain code already enforces that all states in a transition have the same `FixedPart` (and therefore the same `chainId`). So there’s no need to check it for every transition.
c) ignoring the states (and maybe log a warning) with a wrong chainid instead of throwing an error. I discarded this idea because it seems like quite a catastrophic error in our current "single chain" world / assumption. 

### Quirks
We use a `number` for the chain id in the wallet, but a `string` in the store. We sometimes use `chainId` and other times use `chainNetworkId`. 

The other place we should check the chain id is in the chain service. I think it wouldn't be too hard to get into a nasty state where the chain service is pointing at a different chain to the rest of the wallet. Ideally it gets the chain id from the wallet config, and that is considered the source of truth for the chain id. 

---
### Code quality
- [x] I have added tests that prove my fix is effective or that my feature works, if necessary // **new case for `pushMessage`**
### Project management


lalexgap (changes requested):
Does this work for cases where we're running against an actual chain? If we use `testChannel` we'll always get a `chainId` of `0` since `defaultTestConfig().networkConfiguration.chainNetworkID` === `0`.  I think we do need to read `process.env.CHAIN_NETWORK_ID` as that's how our `chain-setup` sets up the actual chain id.

geoknee:
Just to clarify: are you suggesting we read the environment variable from within the wallet code? I am very reluctant to do that since we have this information in the wallet configuration, and that's the right place to declare the configuration for library code (IMO). 

I'd be happier to consider changing the test code (only) to read from the environment variables: is that what you have in mind? It could be an easy fix to simply have the default test config point to the environment variable.

lalexgap:
> Just to clarify: are you suggesting we read the environment variable from within the wallet code? I am very reluctant to do that since we have this information in the wallet configuration, and that's the right place to declare the configuration for library code (IMO).
> 
> I'd be happier to consider changing the test code (only) to read from the environment variables: is that what you have in mind? It could be an easy fix to simply have the default test config point to the environment variable.

I think doing it in the `defaultTestConfig` makes sense. I agree that we don't want this in the actual wallet code, we shouldn't use that for configuration. But in this case, I think we're just using process.env.CHAIN_NETWORK_ID` to communicate the chain id between `chain-setup.ts` and the chain tests.

geoknee:
069d67e

lalexgap (approved):
👍 🙏

geoknee:
Thanks for the review @lalexgap. I went a bit further here #3328, you might be interested to take a look.

PR #3328, opened 2021-02-23
pass chainId to the chain service

geoknee:
So that we get an error early on if there is a mismatch. 

It is critical that the chain we deposit funds to is the same as the chain specified in the signed `States` that the rest of the wallet deals with.

This PR passes the `networkChainId` wallet config variable down to the `ChainService` constructor. There, it is passed down again to the ethers `JsonRpcProvider` constructor. This should mean an error is thrown (I think asynchronously) if a network mismatch is detected. 

Shortcomings: the constructor arguments aren't super elegant, being an object and a number (I tried putting the number in the object, but I wanted it to be required and that wasn't straightforward given the use of `Partial<>`.)

## Changes 

1.  when we start a ganache server, we now set `_rpcChainId`, so that the ganache reports its chain id back to the provider consistently.

## How Has This Been Tested? 
Without the `_rpcChainId`being set, we get
```
@statechannels/server-wallet:     underlying network changed (event="changed", network={"chainId":9002,"name":"unknown"}, detectedNetwork={"chainId":1337,"name":"unknown"}, code=NETWORK_ERROR, version=providers/5.0.9)
@statechannels/server-wallet:       at Logger.Object.<anonymous>.Logger.makeError (../../node_modules/@ethersproject/logger/src.ts/index.ts:205:28)
```
and something similar if we manually change the value passed in to `JsonRpcProvider()`. 

We could think about adding a test that does something like this. 

This change could be considered security related; I don't think it requires multiple approvals though because it is adding checks and shouldn't be undermining security. 
---
### Code quality
### Project management

### DIFF
--- a/packages/devtools/src/ganache/server.ts
+++ b/packages/devtools/src/ganache/server.ts
@@ -157,7 +157,7 @@ export class GanacheServer {
     this.server = ganache.server({
       ...serverOptions,
       _chainId: this.chainId,
-      _chainIdRpc: this.chainId
+      _chainIdRpc: this.chainId // see https://github.com/trufflesuite/ganache-core/tree/master#options
     } as any);
 
     // eslint-disable-next-line @typescript-eslint/no-empty-function
--- a/packages/devtools/src/ganache/server.ts
+++ b/packages/devtools/src/ganache/server.ts
@@ -154,7 +154,11 @@ export class GanacheServer {
       }
     };
 
-    this.server = ganache.server({...serverOptions, _chainId: this.chainId} as any);
+    this.server = ganache.server({
+      ...serverOptions,
+      _chainId: this.chainId,
+      _chainIdRpc: this.chainId
+    } as any);
 
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     this.server.listen(this.port, () => {});
--- a/packages/server-wallet/src/chain-service/chain-service.ts
+++ b/packages/server-wallet/src/chain-service/chain-service.ts
@@ -81,16 +81,19 @@ export class ChainService implements ChainServiceInterface {
 
   private finalizingChannels: {finalizesAtS: number; channelId: Bytes32}[] = [];
 
-  constructor({
-    provider,
-    pk,
-    pollingInterval,
-    logger,
-    blockConfirmations,
-    allowanceMode,
-  }: Partial<ChainServiceArgs>) {
+  constructor(
+    {
+      provider,
+      pk,
+      pollingInterval,
+      logger,
+      blockConfirmations,
+      allowanceMode,
+    }: Partial<ChainServiceArgs>,
+    chainNetworkId: number
+  ) {
     if (!pk) throw new Error('ChainService: Private key not provided');
-    this.provider = new providers.JsonRpcProvider(provider);
+    this.provider = new providers.JsonRpcProvider(provider, chainNetworkId);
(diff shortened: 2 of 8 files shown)


packages/devtools/src/ganache/server.ts
```diff
@@ -154,7 +154,11 @@ export class GanacheServer {
       }
     };
 
-    this.server = ganache.server({...serverOptions, _chainId: this.chainId} as any);
+    this.server = ganache.server({
+      ...serverOptions,
+      _chainId: this.chainId,
+      _chainIdRpc: this.chainId
```
geoknee, line 160: See https://github.com/trufflesuite/ganache-core/tree/master#options

lalexgap: This might be good to specify in a comment

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -81,16 +81,19 @@ export class ChainService implements ChainServiceInterface {
...
-    blockConfirmations,
-    allowanceMode,
-  }: Partial<ChainServiceArgs>) {
+  constructor(
+    {
+      provider,
+      pk,
+      pollingInterval,
+      logger,
+      blockConfirmations,
```
NiloCK, line 90: Wrt the shortcoming in your description, how about:

``` ts
constructor(
    {
      provider,
      pk,
      pollingInterval,
      logger,
      blockConfirmations,
      allowanceMode,
      chainNetworkId,
    }: Partial<ChainServiceArgs> & {chainNetworkId: number; pk: string} // chainNetworkId: number
  )
```
?

This makes the chainNetworkId and pk arguments required, avoiding the '2nd argument' problem and also the need for the `if (!pk) throw` check in the first line of the constructor (assuming this ctor doesn't have any remote invocation stuff that wouldn't know about the required pk arg).

NiloCK: (Remote invocations seem unlikely for this ctor given that the thing I'm talking about is a private key.....)

NiloCK (changes requested):
See my comment in https://github.com/statechannels/statechannels/pull/3328 for a potential fix to your described shortcoming.

Otherwise (lacking broader context on the why and what) this looks clean to me.

(I've selected 'request changes' here in order to avoid accidentally merging this in without giving you a chance to review the suggestion. Please **aggressively** correct my workflow errors for the next 6 months.)

packages/server-wallet/src/chain-service/chain-service.ts
geoknee: Thanks for this suggestion @NiloCK! I only went partway there, since the `pk` thing a) requires removing `undefined` from the type of `pk` in a few more places besides this (higher up in the stack) and b) is a slight mission creep. I did put a TODO in for it, though 73f48f6.

packages/devtools/src/ganache/server.ts
geoknee: 30e8aef

geoknee:
@NiloCK I don't think you made any workflow error here. I think we have a habit in cases like this to just leave an approval but not merge the PR: that way the author can choose to do the change now, in a separate PR, or not at all (at their discretion). I would personally reserve "changes requested" for cases where the reviewer is pretty sure they don't want this PR to be merged as is (for example it makes the codebase worse than it was).

NiloCK (approved):

lalexgap (approved):
👍

Issue #3279, opened 2021-02-10
Group test using peer wallets


lalexgap:
Wallet tests are grouped by the setup and teardown configuration required for the test. The groupings are:
- unit tests configured via `jest.config.js`
- tests that use the ganache chain configured via `jest.chain.config.js`
- e2e tests that spin up application servers configured via `jest.e2e.config.js`
- stress tests that test the performance of the wallet configured via `jest.stress.config.js`

There seems to be another class of tests that uses 2 wallets to test the interaction between the wallets. An example is https://github.com/statechannels/statechannels/blob/20e8101cd3d78ff808f699d497e6835444b0c62a/packages/server-wallet/src/__test__/sync-objectives.test.ts#L27

For this class of tests:
- [x] Common setup and teardown can be factored out into `jest.with-peers.config.js`
- [x] Tests that use this setup and teardown can be grouped under a directory like `__test__/with-peers/...`
- [x] A debug configuration can be added to `packages/server-wallet/.vscode/launch.json`

andrewgordstewart:
https://github.com/statechannels/statechannels/pull/3275#discussion_r574237745

> nit: this file ought to be called something a bit different -- otherwise, it's not obvious why it's separate from the test file included below.

> I think there's a set of tests which play two wallets against each other. Perhaps they should be grouped under __test__/with-peers/... or something like that.

NiloCK:
@lalexgap The test files that are affected by this refactor are already grouped by different folder hierarchies (eg, `/__test__/create-and-close-channel`, `/__test__/crash-tolerance`). Moving them all to `/__test__/with-peers/` would lose the previous logical grouping.

Any comment on whether it's better to leave them in their current physical location? I could leave them as-is and still create a `.config.js` and add a specific `test:with-peers' testing script.

andrewgordstewart:
I _think_ that every test in [`src/__test__`](https://github.com/statechannels/statechannels/tree/master/packages/server-wallet/src/__test__) is a "test with peer". So, it might be as easy as
1. creating `with-peers` as a subdirectory of `src/__test__`
2. Moving the relevant contents of `src/__test__` to `src/__test__/with-peers`

This leaves the option to have, eg,
- `src/__test__/integration/...`
- `src/__test__/stress/...`
to logically organize different subsets of "top-level" tests

PR #3333, opened 2021-02-24
test: DRY refactor of with-peers wallet tests

NiloCK:
See: #3279

Problem: server-wallet tests that use multiple wallets contained a lot of duplicated setup and teardown code. Refactoring to a single file improves maintainability and reduces 'read-time' burden for people investigating these tests.

## Changes

1. duplicated `beforeAll` and `afterAll` code is pulled to `jest/with-peers-setup-teardown.ts`
2. Affected tests grouped into `__test__/with-peers`
3. `jest/with-peers.config.js` added to enable `yarn test:with-peers`
4. debug launch configuration is added for with-peers

## How Has This Been Tested?

- `yarn test` produces same results pre and post refactor
- debugger attaches to target files correctly
- `yarn test:with-peers` produces expected result

## Shortcomings

`jest/with-peers-setup-teardown.ts` does not follow convention as a jest setup file - it does not contain `beforeAll` and `afterAll` functions. Rather, it provides functions and variables for the test files to import and pass to their own `beforeAll` / `afterAll`. This was done to allow for minor differences in the setup requirements of each test file in a readable way.

---
### Code quality
### Project management


packages/server-wallet/jest/with-peers-setup-teardown.ts
```diff
@@ -0,0 +1,69 @@
...
+    ]);
+
+    peerWallets = {
+      a: await Wallet.create(aWalletConfig),
+      b: await Wallet.create(bWalletConfig),
+    };
+
+    if (withWalletSeeding) {
+      await seedAlicesSigningWallet(peerWallets.a.knex);
+      await seedBobsSigningWallet(peerWallets.b.knex);
```
andrewgordstewart, line 46: Suppose someone calls `getPeersSetup(false)`. Is there a reason _not_ to seed these? (Do existing tests depend on these not being seeded??)

packages/server-wallet/src/__test__/with-peers/crash-tolerance/crash-tolerance.test.ts
```diff
@@ -85,25 +51,29 @@ it('Create a directly-funded channel between two wallets, of which one crashes m
...
+  );
 
   expect(getChannelResultFor(channelId, resultB0.channelResults)).toMatchObject({
     status: 'proposed',
     turnNum: 0,
   });
 
   // Destory Wallet b and restart
-  b = await crashAndRestart(b);
+  peerWallets.b = await crashAndRestart(peerWallets.b);
```
andrewgordstewart, line 64: This line is a bit strange: Since `peerWallets` is imported from another module, you are overwriting the value of an import. I think it's _ok_, given my understanding of how the jest test runner works, but it definitely smells a bit off.

andrewgordstewart: An alternative is
1. `path/to/with-peers-setup-teardown` exports a `peerWalletsFactory`
2. test files assign `let peerWallets = peerWalletsFactory()` at the top of their file

packages/server-wallet/package.json
```diff
@@ -101,7 +101,8 @@
     "test:chain": "jest -c ./jest/jest.chain.config.js --runInBand",
     "test:ci": "yarn test --bail && yarn test:chain --bail",
     "test:e2e": "./src/test_db.sh server_wallet_payer server_wallet_receiver; yarn tsc; yarn jest -c ./jest/jest.e2e.config.js --runInBand",
-    "test:stress": "./src/test_db.sh server_wallet_payer server_wallet_receiver; yarn tsc; NODE_ENV=test npx ts-node ./e2e-test/scripts/stress-test.ts"
+    "test:stress": "./src/test_db.sh server_wallet_payer server_wallet_receiver; yarn tsc; NODE_ENV=test npx ts-node ./e2e-test/scripts/stress-test.ts",
+    "test:with-peers": "yarn jest -c ./jest/jest.with-peers.config.js --runInBand"
```
andrewgordstewart, line 105: I see this script added here, but it's not reference elsewhere.

A natural extension of this work is
1. don't run the `with-peers` tests when running `yarn test`
2. Add a new step to circle to run `yarn test:with-peers`.

This would shorten the total length of the CircleCI run by maybe 2 minutes, at the cost of an additional 30-60s startup cost for the step introduced in (2).

andrewgordstewart (approved):
Awesome!

kerzhner:
Would you mind elaborating on the Shortcomings section in the description for why we are not able to follow the standard jest methodology for `beforeAll`/`afterAll`?

andrewgordstewart:
> Would you mind elaborating on the Shortcomings section in the description for why we are not able to follow the standard jest methodology for `beforeAll`/`afterAll`?

Related question here: https://github.com/statechannels/statechannels/pull/3333#discussion_r582329122

packages/server-wallet/package.json
kerzhner: The other thing to note is that if the `Shortcomings` section in the description is addressed and `beforeAll`/`afterAll` are included in the jest config, `yarn test` which maps to `./jest/jest.config.js` will no longer work with peer tests.

packages/server-wallet/jest/with-peers-setup-teardown.ts
NiloCK: (from memory) As is, four of the five tests do call `getPeersSetup(false)`.

Will investigate - intent was to faithfully reproduce initial conditions on existing tests, which I could do (without knowing anything about wallets or how the tests used them) by adding this flag.

packages/server-wallet/src/__test__/with-peers/crash-tolerance/crash-tolerance.test.ts
NiloCK: Agreed. My current understanding: Jest test framework uses (abuses?) global state as a tool to supply the minimal necessary context for unit tests to run. Global state is "generally bad", and abstracting global state for these tests to one file tends toward doing something that is "generally bad" but maybe an acceptable trade-off in context of Jest.

I'll give ~25 minutes to providing a cleaner export here (more if requested), but I know I had some issue that discouraged one prior approach.

caveat: scoping and hoisting in JS is sometimes surprising; I in no way claim complete knowledge of the assumptions Jest makes about available scope

packages/server-wallet/package.json
NiloCK: @andrewgordstewart My expectation of `yarn test` in a package would be that it runs all tests of that package - would the suggestion from (1) break that?

packages/server-wallet/src/__test__/with-peers/crash-tolerance/crash-tolerance.test.ts
andrewgordstewart: I know that if we do not set the `runInBand` option, each test suite (file) runs in a separate node process, meaning it's "ok" to modify objects in the global scope.

I also know that we have to run tests with the `runInBand` option, but don't know the implications of that here.

packages/server-wallet/jest/with-peers-setup-teardown.ts
andrewgordstewart: That approach is great, and makes this easy to review! It might have been worth a comment in the commit message or in this file.

 IMO doing it here is scope creep -- if you are planning on changing the beforeAll usage, it should be in a follow up PR.

kerzhner: I disagree with that. If there is a cost of changing a unit test file in order to create a more standardized way of using `beforeAll`/`afterAll`, I think it is totally fine to do so in this PR even if there is scope creep.

That being said, a valid reason to avoid changing the behavior of the unit test file is to merge the PR quicker.

packages/server-wallet/package.json
kerzhner: At the moment, `yarn test` does not run all tests in a package. It only runs tests that use `./jest/jest.config.js`

andrewgordstewart: I'll clarify a bit more.
1. As you know, a `jest.config.js` file can specify which tests should run with that config. Eg. `./jest/jest.config.js` uses [this `testMatch`](https://github.com/statechannels/statechannels/blob/2cb32c512e700c788224cec28896d2258904d1b0/packages/server-wallet/jest/jest.config.js#L8)
2. There's a few other configs [here](https://github.com/statechannels/statechannels/tree/master/packages/server-wallet/jest). You've added one in this PR.
3. CircleCI runs some of these tests in different steps. EG. the `test:e2e` script is run in the [`server-wallet-e2e-test`](https://app.circleci.com/pipelines/github/statechannels/statechannels/12046/workflows/30438ae2-7758-4e24-ab81-1cec41a09645/jobs/58746) job, while the `test` script is run in the [`test`](https://app.circleci.com/pipelines/github/statechannels/statechannels/12046/workflows/30438ae2-7758-4e24-ab81-1cec41a09645/jobs/58745) job.


Expanding on (3) a bit, the `test` job runs [`yarn test:ci`](https://app.circleci.com/pipelines/github/statechannels/statechannels/12046/workflows/30438ae2-7758-4e24-ab81-1cec41a09645/jobs/58745/parallel-runs/0/steps/0-106) _from the monorepo root_.

There is a lot of value in making the `test` job run faster, by running the `test:with-peers` test cases into a different circle step. Therefore, my suggestion is to
1. do not run the `with-peers` step in `test:ci`
2. Create a new circle step which runs `test:with-peers` OR run `test:with-peers` after `test:e2e` OR a sensible alternative.

This would lower the start-to-finish duration of the CI run by ~2 minutes, expecting that number to grow over time.

... Back to your comment:
> My expectation of `yarn test` in a package would be that it runs all tests of that package 

As it turns out, we can **and maybe should** change `yarn test` to run all the tests in the package, as a local way of quickly "running all the tests". It might do this by running `yarn run-s test:ci test:e2e test:with-peers test:stress ...`, or by running some shell script.

PR #3334, opened 2021-02-25
Add a chainService.checkChainId() and call during wallet creation

### LINKED ISSUE #3332: write a test for chain id mismatch
Following https://github.com/statechannels/statechannels/pull/3328, we will now get an error if the provider we pass in reports a chain id different to the one in the wallet configuration. 

We should add a test for this behaviour. 

The tricky thing: there is actually a cascade of errors thrown when we engineer a mismatch, I think thrown by a number of event listeners. It's difficult to catch all of those errors, or even to detach all the listeners in the test. This means the errors can cause other tests in the suite to (incorrectly) fail.

geoknee:
Closes #3332. Supercedes the work done in #3328.

Currently, we pass the chain id down to constructor for the ethereum `JsonRPCProvider`. This is good because errors will be thrown in the event that the chain reports a different id to the one passed to it's constructor. But it is bad because the constructor chain is all synchronous, and the errors are thrown asynchronously (they have to be, because they are triggered by RPCs). It's also bad that many errors are thrown. As a consequence, there is no test for this behaviour.

The approach of this PR is to allow the provider to setup without a named chain id (it will just use the one from the chain it talks to, eventually), but to expose an async method which can be used to ensure the chain has the expected id. This async method is called early on in the existing asynchronous `wallet.create()`. This way, a single error will be thrown and we will know precisely where to catch it. In the end, it's now easy to test.

## How Has This Been Tested? 
there is a new test.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/chain-service/mock-chain-service.ts
+++ b/packages/server-wallet/src/chain-service/mock-chain-service.ts
@@ -36,6 +36,10 @@ const mockTransactoinResponse: providers.TransactionResponse = {
 };
 
 export class MockChainService implements ChainServiceInterface {
+  async checkChainId(_networkChainId: number): Promise<void> {
+    // noop, a mock chain service will have the "correct" chain id
+  }
+
   fundChannel(_arg: FundChannelArg): Promise<providers.TransactionResponse> {
     return Promise.resolve(mockTransactoinResponse);
   }
--- a/packages/server-wallet/src/wallet/wallet.ts
+++ b/packages/server-wallet/src/wallet/wallet.ts
@@ -105,6 +105,8 @@ export class SingleThreadedWallet
   ): Promise<SingleThreadedWallet> {
     const wallet = new SingleThreadedWallet(walletConfig);
 
+    await wallet.chainService.checkChainId(walletConfig.networkConfiguration.chainNetworkID);
+
     await wallet.registerExistingChannelsWithChainService();
     return wallet;
   }
@@ -153,7 +155,6 @@ export class SingleThreadedWallet
       this.chainService = new ChainService({
         ...this.walletConfig.chainServiceConfiguration,
         logger: this.logger,
-        chainNetworkId: this.walletConfig.networkConfiguration.chainNetworkID,
       });
     } else {
       this.chainService = new MockChainService();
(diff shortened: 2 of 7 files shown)


packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -105,6 +105,8 @@ export class SingleThreadedWallet
   ): Promise<SingleThreadedWallet> {
     const wallet = new SingleThreadedWallet(walletConfig);
 
+    await wallet.chainService.checkChainId(walletConfig.networkConfiguration.chainNetworkID);
```
kerzhner, line 108: Have we considered performing this check in the `ChainService` constructor? It might be nice for a consumer of the `ChainService` to not concern itself with the internals of the `ChainService`

geoknee: The problem is that the check needs to be `async`, so we can't do it in the constructor.

geoknee: I suppose we could go down the route of `await ChainService.create()` , i.e. having an async method that calls into the constructor (as we do with the wallet).

packages/server-wallet/src/chain-service/mock-chain-service.ts
```diff
@@ -36,6 +36,10 @@ const mockTransactoinResponse: providers.TransactionResponse = {
 };
 
 export class MockChainService implements ChainServiceInterface {
+  async checkChainId(_networkChainId: number): Promise<void> {
+    // noop, a mock chain service will have the "correct" chain id
```
NiloCK, line 40: Noop here as in no operation (labelling blank function)? Not suggesting a change, just testing my assumption.

NiloCK (approved):
I'm still learning particulars about Jest, in particular here mocks, but all of this seems readable and reasonable to me.

I think your response to Mike's comment (burying the check in an async factory, hiding a private ctor) is worth doing and makes the ChainService class harder to use incorrectly.

packages/server-wallet/src/chain-service/mock-chain-service.ts
geoknee: Yeah that's right

geoknee:
Thanks @NiloCK. I decided against putting an async factory into the chain service now, although I'm open to the idea in general.

PR #3335, opened 2021-02-26
chore: clean up db names

lalexgap:
All the other DB names in the `server-wallet` are prefixed with `server_wallet`.

 This updates the `TEST_A/B` databases to be named `SERVER_WALLET_TEST_A/B` to follow the existing pattern.


geoknee (approved):
Do we care about having consistent casing for these db names? I think we have some lower case ones,  as well as these upper case ones?

lalexgap:
> Do we care about having consistent casing for these db names? I think we have some lower case ones, as well as these upper case ones?

Good point! I've switched these to lower case in cf9c2850082655ac6c917534ba9b2dd07832b144

PR #3356, opened 2021-03-03
docs: clarify the role of FixedPart

geoknee:
The question of "how programmable" the `challengeDuration` of a channel is has come up while thinking about HTLCs.

The answer I think is a bit subtle, and turns on the definitions of `FixedPart`, `channelId` and `support`.

### DIFF
--- a/packages/docs-website/docs/docs/protocol-tutorial/execute-state-transitions.md
+++ b/packages/docs-website/docs/docs/protocol-tutorial/execute-state-transitions.md
@@ -351,7 +351,7 @@ await(await tx).wait();
 
 ## Shortcutting `validTransition`
 
-Notice that a state may be supported simply by having a full set of signatures (on from each participant). In such a case, there will be no `validTransiiton` check made by the chain. 
+Notice that a state may be supported simply by having a full set of signatures (one from each participant). In such a case, there will be no `validTransiiton` check made by the chain. 
 
 State updates can always be made via unanimous consensus.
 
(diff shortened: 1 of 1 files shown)


packages/docs-website/docs/docs/protocol-tutorial/execute-state-transitions.md
```diff
@@ -348,3 +348,13 @@ const tx = NitroAdjudicator.checkpoint(
 );
 await(await tx).wait();
 ```
+
+## Shortcutting `validTransition`
+
+Notice that a state may be supported simply by having a full set of signatures (on from each participant). In such a case, there will be no `validTransiiton` check made by the chain.
```
NiloCK, line 354: > (on from each participant)

should read

> (one from each participant)

NiloCK (approved):
I can't offer comment on the truthiness of the edit, but this is very readable.

(typo noted below (above?))

packages/docs-website/docs/docs/protocol-tutorial/execute-state-transitions.md
geoknee, line 354:
```suggestion
Notice that a state may be supported simply by having a full set of signatures (one from each participant). In such a case, there will be no `validTransiiton` check made by the chain. 
```

geoknee:
Thanks @NiloCK , and good spot. Not sure if you are aware of this feature of github (when you comment on some changes in a PR): 
Screenshot 2021-03-03 at 14 28 29

which makes the suggestion in a way that can be immediately committed.

NiloCK (approved):

NiloCK:
> Not sure if you are aware of this feature of github (when you comment on some changes in a PR): which makes the suggestion in a way that can be immediately committed.

I was not, and I will be happy to make one-character contributions in the future. Thanks :)

PR #3358, opened 2021-03-05
nitro-protocol: Add a HashLock example ForceMoveApp

geoknee:
In this ForceMoveApp, there is only a single valid transition: from a two-alllocation outcome to another where the amounts are permuted, and where the **preimage** on the later state hashes to the **h** of the earlier state.

We can use this app as a primitive in a cross-chain swap application. This approach is good because it is about as simple as it could be. The drawback is that the channel is only good for a single conditional payment. The idea is that this application would be used in a setting where spinning up and tearing down channels is very cheap (i.e. ledger or virtual funding). Security  (via simplicity) is the top priority, rather than flexibility. 

A new test suite has been added (only 2 test cases so far).

---
### Code quality
### Project management

### DIFF
--- a/packages/nitro-protocol/contracts/examples/HashLock.sol
+++ b/packages/nitro-protocol/contracts/examples/HashLock.sol
@@ -17,9 +17,13 @@ contract HashLock is IForceMoveApp {
     function validTransition(
         VariablePart memory a,
         VariablePart memory b,
-        uint48,
-        uint256
+        uint48 turnNumB,
+        uint256 nParticipants,
     ) public override pure returns (bool) {
+
+        // is this the first and only swap?
+        require(turnNumB == 4);
+
         // Decode variables.
         // Assumptions:
         //  - single asset in this channel
@@ -40,6 +44,9 @@ contract HashLock is IForceMoveApp {
         // was the payment made?
         require(allocationA[0].amount == allocationB[1].amount);
         require(allocationA[1].amount == allocationB[0].amount);
+
+        // is this the first and only swap?
+        require(turnNumB == 4);
 
         return true;
     }
--- a/packages/nitro-protocol/contracts/examples/HashLock.sol
+++ b/packages/nitro-protocol/contracts/examples/HashLock.sol
@@ -50,13 +50,27 @@ contract HashLock is IForceMoveApp {
         returns (Outcome.AllocationItem[] memory allocation)
     {
         Outcome.OutcomeItem[] memory outcome = abi.decode(outcomeBytes, (Outcome.OutcomeItem[]));
+
+        // Throws if more than one asset
+        require(outcome.length == 1, 'outcome: Only one asset allowed');
+
         Outcome.AssetOutcome memory assetOutcome = abi.decode(
             outcome[0].assetOutcomeBytes,
             (Outcome.AssetOutcome)
         );
+
+        // Throws unless the assetoutcome is an allocation
+        require(
+            assetOutcome.assetOutcomeType == uint8(Outcome.AssetOutcomeType.Allocation),
+            'AssetOutcome must be Allocation'
+        );
+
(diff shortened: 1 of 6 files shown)


packages/nitro-protocol/contracts/examples/HashLock.sol
```diff
@@ -0,0 +1,85 @@
...
+        //  - single asset in this channel
+        //  - two parties in this channel
+        //  - not a "guarantee" channel (c.f. Nitro paper)
+        Outcome.AllocationItem[] memory allocationA = decode2PartyAllocation(a.outcome);
+        Outcome.AllocationItem[] memory allocationB = decode2PartyAllocation(b.outcome);
+        bytes memory preImage = abi.decode(b.appData, (AppData)).preImage;
+        bytes32 h = abi.decode(a.appData, (AppData)).h;
+
+        // is the preimage correct?
+        require(keccak256(preImage) == h, 'Incorrect preimage');
```
geoknee, line 37: We could use `sha256` hash function here, instead of `keccak256`. This would give better compatibility with chains such as bitcoin.

packages/nitro-protocol/contracts/examples/HashLock.sol
```diff
@@ -0,0 +1,85 @@
...
+            allocationA[0].destination == allocationB[0].destination &&
+                allocationA[1].destination == allocationB[1].destination,
+            'destinations may not change'
+        );
+
+        // was the payment made?
+        require(
+            allocationA[0].amount == allocationB[1].amount &&
+                allocationA[1].amount == allocationB[0].amount,
+            'amounts must be permuted'
```
geoknee, line 50: This key property should probably be reflected in the name of the contract: e.g. `HashLockedAllocationSwap`.

packages/nitro-protocol/contracts/examples/HashLock.sol
geoknee: 1dd945b

packages/nitro-protocol/contracts/examples/HashLock.sol
geoknee: 2260c14

packages/nitro-protocol/contracts/examples/HashLock.sol
andrewgordstewart: What's the gas cost of `sha256` compared to `keccak256`?

geoknee: It is significantly higher https://ethereum.stackexchange.com/questions/3184/what-is-the-cheapest-hash-function-available-in-solidity.

andrewgordstewart: IMO this is the kind of thing that should be recorded in either the code or the git history ([reference](https://mtlynch.io/code-review-love/#4-answer-questions-with-the-code-itself)).

Neither `sha256` nor `keccak256` is an obvious choice, and it's worthwhile indicating why one was chosen over the other.

As an 💡 : the contract could allow the application data to specify which hash function to use. This makes both the appData and the transition logic complex, at the benefit of specializing and saving gas in EVM chains, or generalizing at the expense of higher gas cost in more generic chains.

geoknee: Good point, I'll add some explanatory comments in the source code before merging this.

It's a good idea bout declaring the has function in the app data. Another one is to just to have two contracts deployed (one using each hash function) which would save on the branching logic inside the contract. I think we could tackle this in future work.

andrewgordstewart: Yep, I think it's best to commit to one hash right now!

geoknee: 957b86d

Issue #3362, opened 2021-03-09
implement ensureObjective


andrewgordstewart:
# Why: Problem description :confused:
Currently, the `Wallet` API relies on the App to forward messages from the `outbox`. With the "Core Channel Manager" RFC, the `ChannelManager` will handle this messaging directly.

# What: Describe the solution you'd like :sparkles:  
An `ensureObjective` function can
1. handle the "ping-pong" of exchanged messages
2. monitor the timestamps on the objective, reacting based on inactivity, re-send messages if need be, or cancelling the objective

It should look something like this:
```
export async function ensureObjective(
  o: WalletObjective,
  w: Wallet,
  ms: MessageServiceInterface,
  initialMessage: unknown
): Promise<'done'> {
  const done = new Promise<'done'>(resolve => {
    w.on('objectiveSucceeded', e => {
      if (e.objectiveId === o.objectiveId) resolve(e);
    });
  });

  await ms.send([initialMessage as Message]);

  // Do something with timeouts & timestamps

  return done;
}
```
It should resolve when the objective succeeds, or reject if
- a timeout is exceeded
- the objective fails

The initial implementation will only handle (1). It will not initially be designed to handle wallet restarts, as that may be handled by some [separate logic](https://www.notion.so/statechannels/RFC-4-Core-Channel-Manager-CM-d6c53ce6526e49af8df165435f805021#426a49a15aaf4afabf74daefd0c2b8e0). It might also probably accept an `initialMessage`, which is generated when creating the objective.

### Events
There is some benefit to making `wallet.ts` _not_ an event emitter.

It's difficult to get around listening for `wallet.on('objectiveSucceeded', ..)` right now. But, when we implement a `ChannelManager`, the `ChannelManager` can instead emit this event during it's [`messageHandler`](https://github.com/statechannels/statechannels/blob/e84ed353a3ef91b3025d93c049ca3eefb548b8e6/packages/server-wallet/src/message-service/test-message-service.ts#L11). (This might require some re-jigging of the way the `ChainService` interacts with `wallet.ts`.)

# Describe alternatives you've considered :thinking:
I did not consider alternatives, as this function is easy to write (we have a [complementary implementation](https://github.com/statechannels/graph-payments/blob/07d48450e5872cadf41aa1a6926806f89d4faa92/packages/payments/src/channel-manager.ts#L423)) as a reference, and it should be easy to refactor.

**Additional context**
[...]

PR #3369, opened 2021-03-09
Implement a basic Messaging Service

lalexgap:
# Description
Introduces a new message service responsible for sending messages between participants.

This introduces the concept of a "MessageService". The responsibility of this message service is to send messages to opponents and trigger the `incomingMessageHandler` when receiving a message.

Longer-term the `wallet` will accept a `MessageServiceFactory`  from the `app`. This allows the `wallet` to create a message service while supplying a callback that can push the message into the wallet.

The initial implementation is a `TestMessageService` used for the majority of the `with-peers` tests. (Ledger funding was skipped since we expect it to change drastically soon). 

## Changes 
1. New `MessageServiceInterface` and associated types.
2. New `TestMessageService`
3. Updated `with-peer` tests to use `TestMessageService`

## How Has This Been Tested? 

The `TestMessageService` is used in `with-peers` test.

Unit tests for the `TestMessageService` have been omitted since it's only used in tests.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/jest/with-peers-setup-teardown.ts
+++ b/packages/server-wallet/jest/with-peers-setup-teardown.ts
@@ -62,7 +62,7 @@ export async function crashAndRestart(
     {participantId: participantIdB, wallet: peerWallets.b},
   ]);
 
-  messageService =await  TestMessageService.createTestMessageService(handler);
+  messageService =await  TestMessageService.create(handler);
 
 }
 
@@ -112,7 +112,7 @@ export function getPeersSetup(withWalletSeeding = false): jest.Lifecycle {
 
      
   const handler =  createTestMessageHandler(participantWallets);
-  messageService =await  TestMessageService.createTestMessageService(handler);
+  messageService =await  TestMessageService.create(handler);
   };
 }
 
--- a/packages/server-wallet/jest/with-peers-setup-teardown.ts
+++ b/packages/server-wallet/jest/with-peers-setup-teardown.ts
@@ -8,7 +8,8 @@ import {
   seedBobsSigningWallet,
 } from '../src/db/seeds/1_signing_wallet_seeds';
 import {MessageServiceInterface} from '../src/message-service/types';
-import {setupTestMessagingService} from '../src/__test-with-peers__/utils';
+import { createTestMessageHandler, TestMessageService } from '../src/message-service/test-message-service';
+
 
 interface TestPeerWallets {
   a: Wallet;
@@ -56,10 +57,13 @@ export async function crashAndRestart(
     peerWallets.b = await Wallet.create(bWalletConfig);
   }
   
-  messageService = await setupTestMessagingService([
+  const handler = await createTestMessageHandler([
     {participantId: participantIdA, wallet: peerWallets.a},
     {participantId: participantIdB, wallet: peerWallets.b},
   ]);
+
+  messageService =await  TestMessageService.createTestMessageService(handler);
+
 }
 
 export function getPeersSetup(withWalletSeeding = false): jest.Lifecycle {
@@ -106,7 +110,9 @@ export function getPeersSetup(withWalletSeeding = false): jest.Lifecycle {
       {participantId: participantIdB, wallet: peerWallets.b},
     ];
 
-    messageService = await setupTestMessagingService(participantWallets);
+     
+  const handler =  createTestMessageHandler(participantWallets);
+  messageService =await  TestMessageService.createTestMessageService(handler);
   };
 }
 
--- a/packages/server-wallet/jest/with-peers-setup-teardown.ts
+++ b/packages/server-wallet/jest/with-peers-setup-teardown.ts
@@ -1,11 +1,14 @@
+import {Participant} from '@statechannels/client-api-schema';
+import {makeDestination} from '@statechannels/wallet-core';
+
 import {Wallet} from '../src/wallet';
 import {DBAdmin, defaultTestConfig, overwriteConfigWithDatabaseConnection} from '../src';
 import {
   seedAlicesSigningWallet,
   seedBobsSigningWallet,
 } from '../src/db/seeds/1_signing_wallet_seeds';
-import {Participant} from '@statechannels/client-api-schema';
-import {makeDestination} from '@statechannels/wallet-core';
+import {MessageServiceInterface} from '../src/message-service/types';
+import {setupTestMessagingService} from '../src/message-service/test-message-service';
 
 interface TestPeerWallets {
   a: Wallet;
@@ -21,8 +24,14 @@ export const bWalletConfig = overwriteConfigWithDatabaseConnection(defaultTestCo
 export let participantA: Participant;
 export let participantB: Participant;
 export let peerWallets: TestPeerWallets;
+/**
+ * A messaging service that links the two peer wallets.
+ * If this is not used it will have no effect.
+ * It can be used to automatically transport messages between the two participants.
+ */
+export let messagingService: MessageServiceInterface;
 
-export function getPeersSetup(withWalletSeeding: boolean = false): jest.Lifecycle {
+export function getPeersSetup(withWalletSeeding = false): jest.Lifecycle {
   return async () => {
     await Promise.all([DBAdmin.dropDatabase(aWalletConfig), DBAdmin.dropDatabase(bWalletConfig)]);
 
@@ -60,6 +69,11 @@ export function getPeersSetup(withWalletSeeding: boolean = false): jest.Lifecycl
         '0x00000000000000000000000000000000000000000000000000000000000bbbb2'
       ),
     };
+
+    messagingService = await setupTestMessagingService([
+      {participantId: participantA.participantId, wallet: peerWallets.a},
+      {participantId: participantB.participantId, wallet: peerWallets.b},
+    ]);
   };
 }
 
--- a/packages/server-wallet/jest/with-peers-setup-teardown.ts
+++ b/packages/server-wallet/jest/with-peers-setup-teardown.ts
@@ -8,9 +8,9 @@ import {
   seedBobsSigningWallet,
 } from '../src/db/seeds/1_signing_wallet_seeds';
 import {MessageServiceInterface} from '../src/message-service/types';
-import { setupTestMessagingService } from '../src/__test-with-peers__/utils';
+import {setupTestMessagingService} from '../src/__test-with-peers__/utils';
 
- interface TestPeerWallets {
+interface TestPeerWallets {
   a: Wallet;
   b: Wallet;
 }
@@ -35,26 +35,33 @@ export const participantIdA = 'a';
 export const participantIdB = 'b';
 
 /**
- * 
+ *
  */
-export async function crashAndRestart(): Promise<void> {
+export async function crashAndRestart(
+  walletsToRestart: 'A' | 'B' | 'Both' = 'Both'
+): Promise<void> {
   // Get all the configs for the wallets
 
   const {walletConfig: aWalletConfig} = peerWallets.a;
   const {walletConfig: bWalletConfig} = peerWallets.b;
 
   await messageService.destroy();
- await  peerWallets.a.destroy();
-  await peerWallets.b.destroy();
 
-    peerWallets = {
-      a: await Wallet.create(aWalletConfig),
-      b: await Wallet.create(bWalletConfig),
-    };
-messageService = await setupTestMessagingService([{participantId: participantIdA, wallet: peerWallets.a},{participantId: participantIdB, wallet: peerWallets.b} ]) ;
- 
+  if (walletsToRestart === 'A' || walletsToRestart === 'Both') {
+    await peerWallets.a.destroy();
+    peerWallets.a = await Wallet.create(aWalletConfig);
   }
 
+  if (walletsToRestart === 'B' || walletsToRestart === 'Both') {
+    await peerWallets.b.destroy();
+    peerWallets.b = await Wallet.create(bWalletConfig);
+  }
+  
+  messageService = await setupTestMessagingService([
+    {participantId: participantIdA, wallet: peerWallets.a},
+    {participantId: participantIdB, wallet: peerWallets.b},
+  ]);
+}
 
 export function getPeersSetup(withWalletSeeding = false): jest.Lifecycle {
   return async () => {
@@ -74,7 +81,6 @@ export function getPeersSetup(withWalletSeeding = false): jest.Lifecycle {
       a: await Wallet.create(aWalletConfig),
       b: await Wallet.create(bWalletConfig),
     };
-
 
     if (withWalletSeeding) {
       await seedAlicesSigningWallet(peerWallets.a.knex);
@@ -96,17 +102,13 @@ export function getPeersSetup(withWalletSeeding = false): jest.Lifecycle {
       ),
     };
 
-    
-  const participantWallets =  [
-     {participantId: participantIdA, wallet: peerWallets.a},
-      {participantId: participantIdB, wallet: peerWallets.b}];
+    const participantWallets = [
+      {participantId: participantIdA, wallet: peerWallets.a},
+      {participantId: participantIdB, wallet: peerWallets.b},
+    ];
 
     messageService = await setupTestMessagingService(participantWallets);
-
-  
   };
-
-
 }
 
 export const peersTeardown: jest.Lifecycle = async () => {
--- a/packages/server-wallet/jest/with-peers-setup-teardown.ts
+++ b/packages/server-wallet/jest/with-peers-setup-teardown.ts
@@ -24,6 +24,7 @@ export const bWalletConfig = overwriteConfigWithDatabaseConnection(defaultTestCo
 export let participantA: Participant;
 export let participantB: Participant;
 export let peerWallets: TestPeerWallets;
+
 /**
  * A messaging service that links the two peer wallets.
  * If this is not used it will have no effect.
@@ -35,15 +36,13 @@ export const participantIdA = 'a';
 export const participantIdB = 'b';
 
 /**
- *
+ * This destroys the peerWallet(s) and the message service and then re-instantiates them.
+ * This mimics a crash and restart
+ * @param walletsToRestart Specifies the peerWallet that will be restarted
  */
 export async function crashAndRestart(
   walletsToRestart: 'A' | 'B' | 'Both' = 'Both'
 ): Promise<void> {
-  // Get all the configs for the wallets
-
-  const {walletConfig: aWalletConfig} = peerWallets.a;
-  const {walletConfig: bWalletConfig} = peerWallets.b;
 
   await messageService.destroy();
 
--- a/packages/server-wallet/jest/with-peers-setup-teardown.ts
+++ b/packages/server-wallet/jest/with-peers-setup-teardown.ts
@@ -8,9 +8,9 @@ import {
   seedBobsSigningWallet,
 } from '../src/db/seeds/1_signing_wallet_seeds';
 import {MessageServiceInterface} from '../src/message-service/types';
-import {setupTestMessagingService} from '../src/message-service/test-message-service';
+import { setupTestMessagingService } from '../src/__test-with-peers__/utils';
 
-interface TestPeerWallets {
+ interface TestPeerWallets {
   a: Wallet;
   b: Wallet;
 }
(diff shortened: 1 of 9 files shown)


lalexgap:
@andrewgordstewart I'm hoping this is enough to unblock you on #3362?

packages/server-wallet/src/message-service/types.ts
```diff
@@ -0,0 +1,30 @@
...
+) => Promise<void>;
+
+/**
+ * A MessageServiceFactory is responsible for generating a MessageService.
+ * Eventually the wallet will require a MessageServiceFactory.
+ * The wallet will use it to construct the messagingService that it can use to send messages.
+ * messageHandler should be triggered whenever the messaging service receives a message.
+ */
+export type MessageServiceFactory = (
+  messageHandler: MessageHandler
```
andrewgordstewart, line 21: nit:
```suggestion
  incomingMessageHandler: MessageHandler
```

lalexgap: Done in 858df91

packages/server-wallet/jest/with-peers-setup-teardown.ts
```diff
@@ -21,8 +24,46 @@ export const bWalletConfig = overwriteConfigWithDatabaseConnection(defaultTestCo
...
+ * If this is not used it will have no effect.
+ * It can be used to automatically transport messages between the two participants.
+ */
+export let messageService: MessageServiceInterface;
+
+export const participantIdA = 'a';
+export const participantIdB = 'b';
+
+/**
+ *
```
lalexgap, line 38: TODO:  missing description

packages/server-wallet/src/__test-with-peers__/crash-tolerance/crash-tolerance.test.ts
```diff
@@ -117,6 +97,9 @@ it('Create a directly-funded channel between two wallets, of which one crashes m
     turnNum: 0,
   });
 
+  await messageService.send(resultA2.outbox.map(o => o.params));
```
lalexgap, line 100: TODO: Lots of `resultA2.outbox.map(o => o.params))`. Could be a helper or `send` could accept `o`

packages/server-wallet/src/__test-with-peers__/create-and-close-channel/ledger-funding.test.ts
```diff
@@ -21,7 +20,11 @@ import {
   aWalletConfig,
   bWalletConfig,
 } from '../../../jest/with-peers-setup-teardown';
-
+async function crashAndRestart(wallet: Wallet): Promise<Wallet> {
```
lalexgap, line 23: TODO: Leave comment specifying this is temporary until lf is refactored

packages/server-wallet/src/__test-with-peers__/utils.ts
```diff
@@ -0,0 +1,46 @@
+import {ChannelResult} from '@statechannels/client-api-schema';
+
+import {Wallet} from '..';
+import {TestMessageService} from '../message-service/test-message-service';
+import {MessageHandler, MessageServiceInterface} from '../message-service/types';
+
+// TODO: Is there a cleaner way of writing this?
+export async function expectLatestStateToMatch(
```
lalexgap, line 8: @andrewgordstewart you've written a lot of test helpers, is there a nicer way of doing this?

packages/server-wallet/src/__test-with-peers__/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * // This will send all the messages from the result of the create channel call
+ * // and will handle any responses to those messages and so on...
+ * await ms.handleMessages(result.outbox);
+ */
+export function setupTestMessagingService(
+  wallets: {participantId: string; wallet: Wallet}[]
+): Promise<MessageServiceInterface> {
+  const messageHandler: MessageHandler = async (to, message, me) => {
+    // TODO: This assumes only 1 matching wallet.
```
lalexgap, line 35: TODO: check for duplicates when creating

packages/server-wallet/src/__test-with-peers__/utils.ts
andrewgordstewart: What don't you find nice about this?

lalexgap: The main issue is that you have to `await` `expectLatestStateToMatch` which could be easy to miss when writing a test. Probably not worth worrying too much about it.

packages/server-wallet/src/__test-with-peers__/utils.ts
lalexgap: c8d8eba

packages/server-wallet/src/__test-with-peers__/create-and-close-channel/ledger-funding.test.ts
lalexgap: 64ea12c

packages/server-wallet/jest/with-peers-setup-teardown.ts
lalexgap: fece21e

packages/server-wallet/src/__test-with-peers__/utils.ts
andrewgordstewart: I don't think there's any way around that, right?

packages/server-wallet/src/message-service/types.ts
```diff
@@ -0,0 +1,30 @@
...
+) => Promise<MessagingServiceInterface>;
+
+export interface MessagingServiceInterface {
```
NiloCK: Do we have stated standards for declaring interfaces vs. types?

packages/server-wallet/src/message-service/types.ts
```diff
@@ -0,0 +1,35 @@
...
+ * The wallet will use it to construct the messagingService that it can use to send messages.
+ * messageHandler should be triggered whenever the messaging service receives a message.
+ */
+export type MessageServiceFactory = (
+  incomingMessageHandler: MessageHandler
+) => Promise<MessageServiceInterface>;
+
+export interface MessageServiceInterface {
+  /**
+   * Sends out the messages to the specified by participants.
```
NiloCK, line 26:
```suggestion
   * Sends out the messages to the specified recipients.
```

packages/server-wallet/src/__test-with-peers__/crash-tolerance/crash-tolerance.test.ts
```diff
@@ -117,6 +97,9 @@ it('Create a directly-funded channel between two wallets, of which one crashes m
     turnNum: 0,
   });
 
+  await messageService.send(getMessages(resultA2));
+  await messageService.send(getMessages(resultB2));
```
NiloCK, line 101: Clarifying intent: the various `messageService.send` calls are being inserted into existing tests in order to ensure that sending / receiving messages doesn't impact previous behavior?

lalexgap: Yup.  The intent here was to create a new messaging service and use the `with-peers` tests to ensure that using the new messaging service didn't break anything.

packages/server-wallet/src/message-service/types.ts
lalexgap: I don't think we do!

Generally, I tend to favour an interface when defining something like an API but that's mostly based on habit. 

From the [typescript handbook](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)
```
For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until you need to use features from type.
```
so it seems like we should prefer interfaces until they no longer suffice?

packages/server-wallet/src/__test-with-peers__/utils.ts
lalexgap: I thought there might be a more elegant way using a custom jest matcher? Either way, I don't think it's worth the time.

packages/server-wallet/src/message-service/types.ts
andrewgordstewart: Can a `Class` implement a `type`?

lalexgap: Beat me to it (I also tried it [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyglgWzAGwgFXNAvFA3gQwC4A7AVwQCMIAnAXwCh6IAPMAe2uCgGNl8BnfrEQoIAYT6CoI1AgjFgQ+ElQZIeelC1QwpCsjjco+KDgCM9WkA))

NiloCK: The differences between interfaces and types are small. The features from type that might apply in the above heuristic are in creating and extending from union types. Interfaces support declaration merging which can sometimes matter for user-augmentation of a library's exported types. Anyway, the usage here is not a problem at all, I just wanted to ask because I saw both types and an interface in the same file.

Thanks, Effective Typescript!

packages/server-wallet/src/message-service/types.ts
```diff
@@ -0,0 +1,35 @@
+import {Message} from '@statechannels/client-api-schema';
+
+/**
+ * This is the handler that any messaging service implementation should call when receiving a message
+ * The handler is responsible for pushing the message into the appropriate wallet
+ */
+export type MessageHandler = (
+  // Including the to makes it easier for a messaging service to handle multiple wallets if it chooses to do so
+  to: string,
+  message: unknown,
```
NiloCK, line 10:
```suggestion
  message: Message
```
Probably an improvement with respect to code re-use and type consistency.

packages/server-wallet/src/__test-with-peers__/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  if (!hasUniqueWallets) {
+    throw new Error('Duplicate wallets');
+  }
+  const messageHandler: MessageHandler = async (to, message, me) => {
+    const matching = wallets.find(w => w.participantId === to);
+
+    if (!matching) {
+      throw new Error(`Invalid to value ${to}`);
```
NiloCK, line 50:
```suggestion
  const messageHandler: MessageHandler = async (message, me) => {
    const matching = wallets.find(w => w.participantId === message.recipient);

    if (!matching) {
      throw new Error(`Invalid to value ${message.recipient}`);
```
Refactor (to match other suggestion)

packages/server-wallet/src/message-service/test-message-service.ts
```diff
@@ -0,0 +1,26 @@
...
+  static async createTestMessageService(
+    incomingMessageHandler: MessageHandler
+  ): Promise<MessageServiceInterface> {
+    const service = new TestMessageService(incomingMessageHandler);
+    return service;
+  }
+
+  async send(messages: Message[]): Promise<void> {
+    for (const message of messages) {
+      await this._receive(message.recipient, message.data, this);
```
NiloCK, line 18:
```suggestion
      await this._receive(message, this);
```

Refactor (to match other suggestion)

NiloCK (approved):
Comments aside, I understand and approve this work. I'm still pretty green, and the change is not tiny, so it may be a good idea for another set of eyes to have a look.

packages/server-wallet/src/message-service/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  messageService: MessageServiceInterface
```
andrewgordstewart, line 11: Why does `MessageHandler` require a message service? Do you plan to remove this in a future PR?

A `MessageService` should be able to:
1. allow the "owner" of the service to send messages to others
2. allow the "owner" to decide how to handle messages received

In production, each wallet has their own `MessageService` instance. Internally, it will listen for incoming messages. (eg. if it is on an HTTP server, it may listen to the `/inbox` endpoint. If it is an HTTP client, it may set up server sent events against a `/outbox` endpoint.) The `MessageHandler` allows the wallet to say "MessageService, when you receive a message, push it into my wallet."

Understandably, in our `with-peers` test suite, there needs to be some kind of a "broker" that has a handle on all of the `TestMessageService` implementations. In this PR, `TestMessageService` acts both as a broker _and_ the `MessageService` used by all of the peers. That has some undesirable side effects, like:
- if you `destroy()` the MessageService, it is destroyed for both peers

lalexgap: > Why does `MessageHandler` require a message service? Do you plan to remove this in a future PR?

This is done to allow the `MessageHandler` to easily call `mesageService.send` if it wants to. Since the `MessageHandler` callback is constructed before the `MessageService` is constructed we don't have a reference to the `MessageService` in the `MessageHandler`. By passing it in it allow for easily constructing a `MessageHandler` that pushes messages into the wallet and calles `messageService.send` to respond.

> Understandably, in our with-peers test suite, there needs to be some kind of a "broker" that has a handle on all of the TestMessageService implementations. In this PR, TestMessageService acts both as a broker and the MessageService used by all of the peers. That has some undesirable side effects, like:

I think that's a fair point that is acting as both a broker and message service. It seemed the most straightforward to have the `TestMessageService` act as the broker instead of having a more complicated setup of two message services communicating.

> if you destroy() the MessageService, it is destroyed for both peers

I think that's not too big of a drawback? I think it's pretty apparent that destroying the message service would destroy it for both peers.

andrewgordstewart: You're declaring a `MessageHandler` type here that is not a test type. But, the type is suited to fit the `TestMessageService`, and not the production use case.

In production, a `MessageHandler`
1. will not have a handle on the peer's `MessageService`
2. does not need a handle on the "local `MessageService`

Right?

andrewgordstewart: > I think that's not too big of a drawback? I think it's pretty apparent that destroying the message service would destroy it for both peers.

Since the test setup shares a `MessageService`, yes, I agree it's apparent. But it's also not representational of the production use case, where each wallet has a distinct `MessageService` instance.

lalexgap: > You're declaring a `MessageHandler` type here that is not a test type. But, the type is suited to fit the `TestMessageService`, and not the production use case.
> 
> In production, a `MessageHandler`
> 
> 1. will not have a handle on the peer's `MessageService`
> 2. does not need a handle on the "local `MessageService`
> 
> Right?

I think 2. is not true?

I'm under the assumption that a production message service might be setup roughly like this:

```
function handleMessage(message:Message){
     const response = this.wallet.pushMessage(message);
    await messageService.send(response);
}
const messageService = createMessageService(handleMessage);
```

However, to accomplish this the `handleMessage` needs a reference to the `messageService`. Since the function is being constructed before the message service is constructed, I've passed it in as an argument.

andrewgordstewart: I have a different (albeit fuzzy) model
```
// channel manager

function handleMessage(message:Message){
     return this.wallet.pushMessage(message);
}
const messageService = createMessageService(handleMessage);
wallet.on('outbox', message => messageService.send(message));
```

I think your approach is fine. Perhaps better, since the `wallet` does not have to emit events with your approach. What are the drawbacks of your approach?

In any case, the presence of the circular dependency indeed makes it confusing. Can you clarify that in the code?

lalexgap: > I think your approach is fine. Perhaps better, since the `wallet` does not have to emit events with your approach. What are the drawbacks of your approach?

I think the main drawback is that it's kinda confusing?

  In our [discussion](https://www.notion.so/statechannels/State-Channels-a19d0ee8714e49e0b7b07e889e97cedd?p=ff9740bfa8f04d4981cf74774d944f33) we came to the conclusion that we wanted to avoid `EventEmitter` so I tried exploring solution without it. I think I still prefer avoiding the `EventEmitter` even with the confusion of the current solution.

> In any case, the presence of the circular dependency indeed makes it confusing. Can you clarify that in the code?

Yup! It also might be worthwhile writing a `readme.md` for the message service (especially since longer-term other people besides us might write a message service)

andrewgordstewart: An alternative is that the `MessageService` can be an event emitter that emits `'messageReceived'`. This allows the `ChannelManager` to attach a callback after the `MessageService` is instantiated.

packages/server-wallet/jest/with-peers-setup-teardown.ts
```diff
@@ -22,7 +25,44 @@ export let participantA: Participant;
...
+
+  if (walletsToRestart === 'A' || walletsToRestart === 'Both') {
+    await peerWallets.a.destroy();
+    peerWallets.a = await Wallet.create(aWalletConfig);
+  }
+
+  if (walletsToRestart === 'B' || walletsToRestart === 'Both') {
+    await peerWallets.b.destroy();
+    peerWallets.b = await Wallet.create(bWalletConfig);
+  }
```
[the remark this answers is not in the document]
lalexgap: That's more concise but slightly easier to shoot yourself in the foot. IE: One of these fails but it's not immediately clear which one: `crashAndRestart(['A'])`  or `crashAndRestart(['a'])`

packages/server-wallet/src/__test-with-peers__/create-and-close-channel/direct-funding.test.ts
```diff
@@ -51,59 +52,12 @@ it('Create a directly funded channel between two wallets ', async () => {
...
-   * In this case, there is no auto-advancing to the running stage. Instead we have
-   * an intermediate 'opening' stage where each party must fund their channel. A funds
-   * first, and then B funds. A and B both signs turnNum 3 on the call to updateFundingForChannels
-   * and then sends the newly signed state to each other at the same time.
-   */
-
-  expect(getChannelResultFor(channelId, resultA1.channelResults)).toMatchObject({
-    status: 'opening',
-    turnNum: 0,
-  });
```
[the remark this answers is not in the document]
lalexgap: One drawback of the messaging services is that it becomes slightly more difficult to check the results at every step (since the messages are automatically sent). Perhaps a `pause` method could make these checks easier? Although I'm not certain it's worth the effort?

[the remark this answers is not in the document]
andrewgordstewart: What's the goal of the test? Is it to assert that a specific sequence of messages happens in a specific order?

If that's the goal, then the `TestMessageService` can be used to that effect:
- It can _record_ which messages are exchanged, in what order.
- We can then _expect_ a statement about the messages that are exchanged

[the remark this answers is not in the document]
lalexgap: I think the previous checks weren't looking at the `messages` but at the `channelResult` to check if the channel is in the correct state.

If we want to inspect the `channelResult` (or `message` payloads) at each step does it make more sense to just have the test manually call `pushMessage` like before? It seems like adding a `TestMessageService` then having to scan through the history is more work then just manually handing the pushMessage results.

packages/server-wallet/src/message-service/test-message-service.ts
lalexgap: Resolved in 0accce4

packages/server-wallet/src/__test-with-peers__/utils.ts
lalexgap: Resolved in 0accce4

packages/server-wallet/src/message-service/types.ts
lalexgap: Resolved in 0accce4

packages/server-wallet/jest/with-peers-setup-teardown.ts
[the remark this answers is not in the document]
andrewgordstewart: If you wanted it to be safer, you could accept `('a' | 'b')[]`?

packages/server-wallet/src/__test-with-peers__/create-and-close-channel/direct-funding.test.ts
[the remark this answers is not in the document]
andrewgordstewart: Let me restate: What _should_ the goal be of the `with-peers` tests?

- Should the goal be to ensure that the correct messages are passed, when one party calls `createChannels`?
- Should the goal be to ensure that when one party calls `createChannels`, that the promise resolves? and at the end, the party has the right number of funded channels?

> the previous checks weren't looking at the messages but at the channelResult to check if the channel is in the correct state.

I don't think it's necessarily wrong to have this duplicated, but there are already "unit tests" of these crankers that assert on the `channelResult`. So, maybe this test doesn't also need to do that.

> It seems like adding a TestMessageService then having to scan through the history is more work then just manually handing the pushMessage results.

IMO it's not a question about how much work it is, it's about what the test is testing. If it's meant to be testing what messages get passed back and forth, then I think writing the `TestMessageService` _once_ to record messages that are sent is worth it. You can then use the `TestMessageService` to assert that:
- in _this_ situation, these messages are exchanged
- in _that_ situation, those messages are exchanged

Again, I do not know if that's what we're trying to test here.

PR #3370, opened 2021-03-10
chore: use === instead of == for comparison

kerzhner:
Note that there are a couple places in the monorepo where a variable is compared with `undefined` via `== undefined`. These comparisons are left as is since maybe `null` is an acceptable value for these comparisons.

```
> null == undefined
true
> null === undefined
false
```


kerzhner:
> Shouldn't there be an eslint rule preventing this?

That would be nice.

Issue #3371, opened 2021-03-11
Clean up LedgerProposals


kerzhner:
Ledger proposals are deprecated after https://github.com/statechannels/statechannels/pull/3287. The cleanup involves:
- Remove code references.
- Add a DB migration to delete the table.

Issue #3372, opened 2021-03-12
Clean up skipped test case


kerzhner:
Yeah, probably. I wasn't sure what to do with this - the old ledger algorithm did a lot of extra ("pessimistic") state communication, so it could handle cases like these. The new ledger algorithm doesn't do this - it just relies on sync states being called to handle the case where things have stalled. So the new ledger algorithm fails this test, but that's probably ok?

_Originally posted by @tomclose in https://github.com/statechannels/statechannels/pull/3287#r592221482_

PR #3374, opened 2021-03-12
Remove ledger proposal table and model

kerzhner:
The [latest ledger funding ](https://github.com/statechannels/statechannels/pull/3287) does not use ledger proposals.

Fixes https://github.com/statechannels/statechannels/issues/3371, https://github.com/statechannels/statechannels/issues/3372


packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -879,7 +862,7 @@ export class SingleThreadedWallet
     response.createdObjectives = storedObjectives;
 
     const channelIdsFromRequests: Bytes32[] = [];
-    const requests = (wirePayload.requests || []).map(deserializeRequest);
+    const requests = wirePayload.requests || [];
```
geoknee, line 865: Curious about this change? It doesn't seem related to the main thrust of the PR?

packages/server-wallet/src/db/migrations/20210312082700_remove_ledger_proposals.ts
```diff
@@ -0,0 +1,7 @@
+import {
+  up as createLedgerProposalsTable,
+  down as dropLedgerProposalsTable,
+} from './20201201095501_ledger_proposals';
+
+export const up = dropLedgerProposalsTable;
+export const down = createLedgerProposalsTable;
```
geoknee, line 7: It occurs to me here that there is an opportunity to rewrite history so that this ledger proposals table never existed. What are your thoughts on doing that?

geoknee (approved):
Looks good to me!


> Fixes #3287, #3372

#3287 is a PR, not an Issue. It's obviously relevant, but I wonder if you meant to link to an issue?

packages/server-wallet/src/wallet/wallet.ts
geoknee: Nevermind, I see that the `deserializeRequest` function is very specific to ledger proposals, and otherwise is essentially an identity operation. Good call on removing it.

packages/server-wallet/src/db/migrations/20210312082700_remove_ledger_proposals.ts
kerzhner: I think that it would be nice to collapse all migrations into a single, initial migration since we are no longer on the hook for maintaining/migrating databases that are in use. Filed https://github.com/statechannels/statechannels/issues/3379

PR #3376, opened 2021-03-15
EnsureObjective

lalexgap:
# Description
Implements the basic functionality of `EnsureObjectives`.

To have a robust wallet we need to make sure that objectives still get completed even in the face of dropped messages.

This PR implements a basic EnsureObjective message that resends messages until the objectives get completed.

At the moment it uses a simple back-off strategy.

A very basic `ChannelManager` has been introduced that has `EnsureObjectives` and a fairly simple `CreateChannels` (for testing).

## Changes 

1. `TestMessageService` now has a function to set `LatencyOptions`. These can be used to simulate poor network conditions.  Currently a `dropRate` and `meanDelayTime` can be set .
2.  A new class called `ChannelManager` has been introduce with `EnsureObjectives/CreateChannels`.

## How Has This Been Tested? 

A `ensure-objective` test has been added to the `two-peers` tests.

This test uses the `TestMessageHandler` to simulate dropped and delayed messages to verify that `EnsureObjective` can handle those scenarios.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/package.json
+++ b/packages/server-wallet/package.json
@@ -57,6 +57,7 @@
     "eslint-plugin-jest": "23.6.0",
     "eslint-plugin-prettier": "3.3.1",
     "eslint-plugin-unicorn": "21.0.0",
+    "exponential-backoff": "^3.1.0",
     "express": "4.17.1",
     "express-pino-logger": "5.0.0",
     "jest": "26.6.3",
--- a/packages/server-wallet/package.json
+++ b/packages/server-wallet/package.json
@@ -57,7 +57,6 @@
     "eslint-plugin-jest": "23.6.0",
     "eslint-plugin-prettier": "3.3.1",
     "eslint-plugin-unicorn": "21.0.0",
-    "exponential-backoff": "^3.1.0",
     "express": "4.17.1",
     "express-pino-logger": "5.0.0",
     "jest": "26.6.3",
--- a/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
@@ -10,10 +10,12 @@ import {
   peerWallets,
 } from '../../jest/with-peers-setup-teardown';
 import {ChannelManager} from '../channel-manager';
+import {
+  createTestMessageHandler,
+  TestMessageService,
+} from '../message-service/test-message-service';
 import {WalletObjective} from '../models/objective';
 import {createChannelArgs} from '../wallet/__test__/fixtures/create-channel';
-
-import {setupTestMessagingService} from './utils';
 
 jest.setTimeout(30_000);
 
@@ -39,7 +41,11 @@ describe('EnsureObjectives', () => {
 
     const dropRate = messageDropPercentage === 0 ? 0 : messageDropPercentage / 100;
 
-    const messageService = await setupTestMessagingService(participantWallets, dropRate);
+    const messageService = await TestMessageService.create(
+      createTestMessageHandler(participantWallets),
+      peerWallets.a.logger,
+      dropRate
+    );
     const channelManager = await ChannelManager.create(peerWallets.a, messageService);
 
     peerWallets.b.on('objectiveStarted', async (o: WalletObjective) => {
@@ -57,7 +63,11 @@ describe('EnsureObjectives', () => {
       {participantId: participantIdB, wallet: peerWallets.b},
     ];
 
-    const messageService = await setupTestMessagingService(participantWallets, 1);
+    const messageService = await TestMessageService.create(
+      createTestMessageHandler(participantWallets),
+      peerWallets.a.logger,
+      1
+    );
     // We use a short backoff to avoid burning times in the tests
     const channelManager = await ChannelManager.create(peerWallets.a, messageService, [1]);
 
--- a/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
@@ -13,7 +13,7 @@ import {LatencyOptions} from '../message-service/test-message-service';
 import {WalletObjective} from '../models/objective';
 import {createChannelArgs} from '../wallet/__test__/fixtures/create-channel';
 
-jest.setTimeout(36_000);
+jest.setTimeout(60_000);
 
 beforeAll(getPeersSetup());
 afterAll(peersTeardown);
@@ -34,11 +34,11 @@ describe('EnsureObjectives', () => {
       meanDelay: undefined,
     },
     // Lots of messages dropping but no delay
-    {dropRate: 0.2, meanDelay: undefined},
+    {dropRate: 0.5, meanDelay: undefined},
     // delay but no dropping
     {dropRate: 0, meanDelay: 150},
-    // 10% message drop and 50 ms mean delay
-    {dropRate: 0.1, meanDelay: 50},
+    // Delay and drop
+    {dropRate: 0.3, meanDelay: 100},
   ];
   test.each(testCases)(
     'can successfully create a channel with the latency options: %o',
@@ -57,8 +57,10 @@ describe('EnsureObjectives', () => {
   //  This is a nice sanity check to ensure that messages do get dropped
   test('fails when all messages are dropped', async () => {
     messageService.setLatencyOptions({dropRate: 1});
-    // We use a short backoff to avoid burning times in the tests
-    const channelManager = await ChannelManager.create(peerWallets.a, messageService, [1]);
+    // We limit the attempts to avoid wasting times in the test
+    const channelManager = await ChannelManager.create(peerWallets.a, messageService, {
+      numOfAttempts: 1,
+    });
 
     peerWallets.b.on('objectiveStarted', async (o: WalletObjective) => {
       const {targetChannelId: channelId} = o.data;
--- a/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
@@ -27,7 +27,7 @@ function getCreateChannelsArgs(): CreateChannelParams {
   });
 }
 
-describe('EnsureObjectives', () => {
+describe('Ensure-Objectives', () => {
   const cases = [0, 0.1, 0.2, 0.3];
 
   test.each(cases)('creates a channel with a drop rate of %f ', async dropRate => {
--- a/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
@@ -50,7 +50,9 @@ describe('EnsureObjectives', () => {
         await peerWallets.b.joinChannels([o.data.targetChannelId]);
       });
 
-      await channelManager.createChannels(getCreateChannelsArgs(), 50);
+      await expect(
+        channelManager.createChannels(getCreateChannelsArgs(), 50)
+      ).resolves.not.toThrow();
     }
   );
 
--- a/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
@@ -36,6 +36,7 @@ describe('EnsureObjectives', () => {
       {participantId: participantIdA, wallet: peerWallets.a},
       {participantId: participantIdB, wallet: peerWallets.b},
     ];
+
     const dropRate = messageDropPercentage === 0 ? 0 : messageDropPercentage / 100;
 
     const messageService = await setupTestMessagingService(participantWallets, dropRate);
@@ -55,6 +56,7 @@ describe('EnsureObjectives', () => {
       {participantId: participantIdA, wallet: peerWallets.a},
       {participantId: participantIdB, wallet: peerWallets.b},
     ];
+
     const messageService = await setupTestMessagingService(participantWallets, 1);
     // We use a short backoff to avoid burning times in the tests
     const channelManager = await ChannelManager.create(peerWallets.a, messageService, [1]);
--- a/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
@@ -49,8 +49,7 @@ describe('EnsureObjectives', () => {
     const channelManager = await ChannelManager.create(peerWallets.a, messageService);
 
     peerWallets.b.on('objectiveStarted', async (o: WalletObjective) => {
-      const {targetChannelId: channelId} = o.data;
-      await peerWallets.b.joinChannel({channelId});
+      await peerWallets.b.joinChannels([o.data.targetChannelId]);
     });
 
     await channelManager.createChannels(getCreateChannelsArgs(), 50);
@@ -73,7 +72,7 @@ describe('EnsureObjectives', () => {
 
     peerWallets.b.on('objectiveStarted', async (o: WalletObjective) => {
       const {targetChannelId: channelId} = o.data;
-      await peerWallets.b.joinChannel({channelId});
+      await peerWallets.b.joinChannels([channelId]);
     });
 
     await expect(channelManager.createChannels(getCreateChannelsArgs(), 1)).rejects.toThrow(
--- a/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
@@ -27,14 +27,17 @@ function getCreateChannelsArgs(): CreateChannelParams {
   });
 }
 
-describe('Ensure-Objectives', () => {
-  const cases = [0, 0.1, 0.2, 0.3];
+describe('EnsureObjectives', () => {
+  // This is the percentages of messages that get dropped
+  const cases = [0, 10, 30];
 
-  test.each(cases)('creates a channel with a drop rate of %f ', async dropRate => {
+  test.each(cases)('creates a channel with a drop rate of %f%% ', async messageDropPercentage => {
     const participantWallets = [
       {participantId: participantIdA, wallet: peerWallets.a},
       {participantId: participantIdB, wallet: peerWallets.b},
     ];
+    const dropRate = messageDropPercentage === 0 ? 0 : messageDropPercentage / 100;
+
     const messageService = await setupTestMessagingService(participantWallets, dropRate);
     const channelManager = new ChannelManager(peerWallets.a, messageService);
 
@@ -43,7 +46,7 @@ describe('Ensure-Objectives', () => {
       await peerWallets.b.joinChannel({channelId});
     });
 
-    await channelManager.createChannels(getCreateChannelsArgs(), 5);
+    await channelManager.createChannels(getCreateChannelsArgs(), 50);
   });
 
   // This is a nice sanity check to ensure that messages do get dropped
@@ -60,7 +63,7 @@ describe('Ensure-Objectives', () => {
       await peerWallets.b.joinChannel({channelId});
     });
 
-    await expect(channelManager.createChannels(getCreateChannelsArgs(), 5)).rejects.toThrow(
+    await expect(channelManager.createChannels(getCreateChannelsArgs(), 1)).rejects.toThrow(
       'Unable to ensure objectives'
     );
   });
--- a/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
@@ -2,22 +2,18 @@ import {CreateChannelParams} from '@statechannels/client-api-schema';
 
 import {
   getPeersSetup,
+  messageService,
   participantA,
   participantB,
-  participantIdA,
-  participantIdB,
   peersTeardown,
   peerWallets,
 } from '../../jest/with-peers-setup-teardown';
 import {ChannelManager} from '../channel-manager';
-import {
-  createTestMessageHandler,
-  TestMessageService,
-} from '../message-service/test-message-service';
+import {LatencyOptions} from '../message-service/test-message-service';
 import {WalletObjective} from '../models/objective';
 import {createChannelArgs} from '../wallet/__test__/fixtures/create-channel';
 
-jest.setTimeout(30_000);
+jest.setTimeout(36_000);
 
 beforeAll(getPeersSetup());
 afterAll(peersTeardown);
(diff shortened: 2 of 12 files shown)


packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    fundingStrategy: 'Fake',
+  });
+}
+
+describe('EnsureObjectives', () => {
+  // This is the percentages of messages that get dropped
+  const cases = [0, 10, 30];
+
+  test.each(cases)('creates a channel with a drop rate of %f%% ', async messageDropPercentage => {
+    const participantWallets = [
```
lalexgap, line 35:
```suggestion

    const participantWallets = [
```

andrewgordstewart:
Sanity check: The PR description here suggests that our assumptions on a `MessageService` is relatively weak: that a message might never be delivered, but if it is, it's delivered correctly. Is that assumption currently explicit?

There is value to making this assumption, because we can (like you're doing here) bootstrap that assumption to
- objectives always finish, eventually
- if the `MessageService` is "typically available", objectives "typically succeed"

by re-trying in a function like `ensureObjective`. This code is re-usable across all `MessageService` implementations, lowering the requirements of a new `MessageService`.

lalexgap:
> Sanity check: The PR description here suggests that our assumptions on a `MessageService` is relatively weak: that a message might never be delivered, but if it is, it's delivered correctly. Is that assumption currently explicit?

I'm not sure if that's explicitly stated but I will add some comments to clarify that.

packages/server-wallet/package.json
```diff
@@ -57,6 +57,7 @@
     "eslint-plugin-jest": "23.6.0",
     "eslint-plugin-prettier": "3.3.1",
     "eslint-plugin-unicorn": "21.0.0",
+    "exponential-backoff": "^3.1.0",
```
andrewgordstewart, line 60: Did you consider `'backoff'`? It's already a (dev) dependency.

packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    // We limit the attempts to avoid wasting times in the test
+    const channelManager = await ChannelManager.create(peerWallets.a, messageService, {
+      numOfAttempts: 1,
+    });
+
+    peerWallets.b.on('objectiveStarted', async (o: WalletObjective) => {
+      const {targetChannelId: channelId} = o.data;
+      await peerWallets.b.joinChannels([channelId]);
+    });
+
```
andrewgordstewart, line 69: I think this is unnecessary: if every message is dropped, `b` will not find out about the objective!

```suggestion
```

packages/server-wallet/src/channel-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+import {backOff, IBackOffOptions} from 'exponential-backoff';
+
+import {MessageServiceInterface} from './message-service/types';
+import {getMessages} from './message-service/utils';
+import {WalletObjective} from './models/objective';
+import {Wallet} from './wallet';
+export const delay = async (ms = 10): Promise<void> =>
+  new Promise(resolve => setTimeout(resolve, ms));
+
+export class ChannelManager {
```
andrewgordstewart, line 11: I think we should discuss naming tomorrow in standup. Though I think it's fine to use `ChannelManager` here, we all want to call it `Wallet`. Tom suggested `Engine` or `WalletEngine` for what's currently in `wallet.ts`, which I think works well.

packages/server-wallet/src/channel-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+   * This will be cleaned up in https://github.com/statechannels/statechannels/issues/3365
+   * @param args
+   * @param numberOfChannels
+   * @returns
+   */
+  public async createChannels(args: CreateChannelParams, numberOfChannels: number): Promise<void> {
+    const createResult = await this._wallet.createChannels(args, numberOfChannels);
+
+    await this._messageService.send(getMessages(createResult));
+    await this.ensureObjectives(createResult.newObjectives);
```
andrewgordstewart, line 43: What if `this._messageService.send(getMessages(createResult))` results in the objectives succeeding?

packages/server-wallet/src/channel-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+      if (o.objectiveId === o.objectiveId) {
+        remaining.delete(o.objectiveId);
+      }
+
+      if (remaining.size === 0) {
+        this._wallet.removeListener('objectiveSucceeded', onObjectiveSucceeded);
+      }
+    };
+
+    this._wallet.on('objectiveSucceeded', onObjectiveSucceeded);
```
andrewgordstewart, line 65: In many wallet use cases, this will exceed the default max listener count of `10`. How much should we worry about that?

packages/server-wallet/src/channel-manager.ts
```diff
@@ -0,0 +1,96 @@
...
+            throw new Error('Still objectives to complete');
+          }
+        },
+
+        {...this._backOffOptions, retry: () => remaining.size !== 0}
+      );
+    } catch (error) {
+      wallet.removeListener('objectiveSucceeded', onObjectiveSucceeded);
+      wallet.logger.error('Unable to ensure objectives', {remaining: Array.from(remaining.keys())});
+      throw new Error('Unable to ensure objectives');
```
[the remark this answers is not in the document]
andrewgordstewart: Hmm. Are you making the assumption that `'Still objectives to complete'` is the only error that can be thrown here? If that's the case, we should perhaps make that explicit? ie.
- if `error.msg === 'Still objectives to complete'`, then we throw `'Unable to ensure objectives'`
- else, it's useful for us to know what the `error` is.

andrewgordstewart (approved):
Overall, LGTM

packages/server-wallet/src/channel-manager.ts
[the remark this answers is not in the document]
lalexgap: Good point. `Still objectives to complete` is the expected error but it could be very annoying if we end up swallowing the wrong error.

packages/server-wallet/src/channel-manager.ts
```diff
@@ -0,0 +1,96 @@
...
+            // Throwing an error indicates to the backoff library that the task is not complete
+            throw new Error('Still objectives to complete');
+          }
+        },
+
+        {...this._backOffOptions, retry: () => remaining.size !== 0}
+      );
+    } catch (error) {
+      wallet.removeListener('objectiveSucceeded', onObjectiveSucceeded);
+      wallet.logger.error('Unable to ensure objectives', {remaining: Array.from(remaining.keys())});
```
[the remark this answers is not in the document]
lalexgap: Yup I think so!

packages/server-wallet/src/channel-manager.ts
```diff
@@ -0,0 +1,96 @@
...
+    };
+
+    wallet.on('objectiveSucceeded', onObjectiveSucceeded);
+    try {
+      await backOff(
+        async () => {
+          const {outbox} = await wallet.syncObjectives(objectives.map(o => o.objectiveId));
+          await messageService.send(getMessages(outbox));
+
+          if (remaining.size !== 0) {
```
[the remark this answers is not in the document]
lalexgap: The `remaining.size` is actually checked in two places:
- The function we pass into `backOff`
- In the `retry` function a few lines below.

This means that in the case where `ObjectiveSuceeded` doesn't fire immediately, the `retry` function is called before the next retry attempt. If `remaining.size===0` then the function won't be retried.

packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
lalexgap: That's true! However I like to keep it in there as a sanity check. If messages aren't getting dropped properly then this would allow the test to fail. Otherwise, the test would always succeed whether or not messages are dropped.

packages/server-wallet/src/channel-manager.ts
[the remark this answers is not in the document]
andrewgordstewart: What happens if `remaining.size === 0` when `retry` is called?

packages/server-wallet/src/channel-manager.ts
lalexgap: My understanding is that the MaxListener count is just there to [flag possible memory leaks ](https://nodejs.org/api/events.html#events_emitter_setmaxlisteners_n) and is not an actual limit.

It's probably prudent to set the `MaxListenersCount` to a higher number. Once that is high enough that we don't normally see it but could be triggered if we have a memory leak.

andrewgordstewart:
I'm starting to wonder if we want to use `exponential-backoff`. [This](https://github.com/statechannels/statechannels/pull/3376/files/dcf1122433f0d97a180923de887db15d38febb29#r599874296) conversation showed me that
- I thought I did understand what was going on
- but now, I don't understand what's going on

Implementing an explicit exponential backoff is not that difficult, is it? The advantage of doing that is, we can understand what's going on by reading the code.

packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
andrewgordstewart: Good point!

packages/server-wallet/src/channel-manager.ts
lalexgap: That's a good point! We should probably filter out any successful objectives when `ensureObjectives` is called.

packages/server-wallet/src/channel-manager.ts
andrewgordstewart: If we set `MaxListenersCount` to a higher number, then we lose the ability to detect possible memory leaks. Should we be worried about that?

There is an alternative approach to implement `ensureObjectives`:
- currently, each `ensureObjectives` call creates an `wallet.on('objectiveSucceeded', listener)`
- instead, the wallet could manage a global set of "in progress" objectives, and set up one listener which does, roughly, `this.on('objectiveSucceeded', objective => this.removeInProgressObjective(objective.id))`

Is that, or some variation of that, preferable?

packages/server-wallet/src/channel-manager.ts
andrewgordstewart: The alternative to that is to pass the entire `createResult` to `ensureObjectives`, and send the initial messages from within that. I think that alternative is simpler?

packages/server-wallet/package.json
lalexgap: I did briefly look at backoff but I didn't realize it was something we were already using. Taking a look now I can't seem to find it as a dependency anywhere?

andrewgordstewart: It's a dependency of `ganache-core`, so it's already in `yarn.lock`.

I looked briefly at it, and maybe it doesn't fit our use case -- there's an outstanding issue requesting that the backoff function be called without delay, and I think we want that feature, for instance.

packages/server-wallet/src/channel-manager.ts
lalexgap: > If we set `MaxListenersCount` to a higher number, then we lose the ability to detect possible memory leaks. Should we be worried about that?

Setting `MaxListenerCount` to a higher value just means it would take longer to detect the memory leak. If we're adding listeners and never removing them then eventually we'll hit `MaxListenerCount` even if it's a bit higher.

> There is an alternative approach to implement `ensureObjectives`:
> 
> * currently, each `ensureObjectives` call creates an `wallet.on('objectiveSucceeded', listener)`
> * instead, the wallet could manage a global set of "in progress" objectives, and set up one listener which does, roughly, `this.on('objectiveSucceeded', objective => this.removeInProgressObjective(objective.id))`. then, `ensureObjectives` would check `wallet.anyInProgress?(objectivesOfInterest)`
> 
> Is that, or some variation of that, preferable?

Instead of calling a potential  `wallet.anyInProgress?(objectivesOfInterest)` couldn't we just query the objectives? That way we don't need the listeners at all.

packages/server-wallet/src/channel-manager.ts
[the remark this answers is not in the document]
lalexgap: The `retry` function determines whether the function is retried or not.

In our case we'd expect something like this:
- Function runs to sync objectives / send messages
- function throws an error due to objectives left to complete
- objectives succeed and `remaining.size` is 0
-  After a timeout the `retry` function is called
- `retry` returns false which causes `exponential backoff` to not retry anything
- the ensureObjective function completes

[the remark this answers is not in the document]
andrewgordstewart: If the `retry` function returns false, and `backoff` decides not to retry anything, what happens? does it move to the next line of execution? does it throw an error?

packages/server-wallet/src/channel-manager.ts
andrewgordstewart: > Setting MaxListenerCount to a higher value just means it would take longer to detect the memory leak. 

That's still a drawback! If it takes 12 hours under some load, to detect the memory leak, but we never leave a wallet running for 12 hours, we won't detect it.

> couldn't we just query the objectives?

Can you explain your suggestion in words or pseudocode? I think what you're suggesting introduces unnecessary DB calls, which can be a problem under load.

lalexgap:
> I'm starting to wonder if we want to use `exponential-backoff`. [This](https://github.com/statechannels/statechannels/pull/3376/files/dcf1122433f0d97a180923de887db15d38febb29#r599874296) conversation showed me that
> 
> * I thought I did understand what was going on
> * but now, I don't understand what's going on
> 
> Implementing an explicit exponential backoff is not that difficult, is it? The advantage of doing that is, we can understand what's going on by reading the code.



> I'm starting to wonder if we want to use `exponential-backoff`. [This](https://github.com/statechannels/statechannels/pull/3376/files/dcf1122433f0d97a180923de887db15d38febb29#r599874296) conversation showed me that
> 
> * I thought I did understand what was going on
> * but now, I don't understand what's going on
> 
> Implementing an explicit exponential backoff is not that difficult, is it? The advantage of doing that is, we can understand what's going on by reading the code.

I was debating this as well. Originally I was hoping to simplify the existing logic and get some extra features by using a library. However, I'm not sure if using the library simplified the logic. 

We do get a bunch of options for free by using `exponential-backoff` which is nice. I guess it depends on what options we care about.

Perhaps a good compromise is creating an exponential backoff code that uses the `exponential-backoff` under the hood.

andrewgordstewart:
> I was debating this as well. Originally I was hoping to simplify the existing logic and get some extra features by using a library. However, I'm not sure if using the library simplified the logic.

I think what you're saying is that we'd have equally simple, or simpler logic, without using the library. If this is the case, writing our own implementation that suits our current needs has the benefit of:
1. fewer dependencies (by just one, but still ...)
2. easier for teammates to understand the logic

(2) is the important thing to me. If we have to distill [this conversation](https://github.com/statechannels/statechannels/pull/3376#discussion_r599874296) into documentation explaining the implementation of `exponential-backoff`, we might be better suited to just write it ourselves.

What do you think?

packages/server-wallet/src/channel-manager.ts
lalexgap: > > Setting MaxListenerCount to a higher value just means it would take longer to detect the memory leak.
> 
> That's still a drawback! If it takes 12 hours under some load, to detect the memory leak, but we never leave a wallet running for 12 hours, we won't detect it.

It seems like a slightly convoluted constraint that we require the wallet have at most 10 listeners just for the purpose of a warning message?

I guess I feel that the benefits of the `maxListener` check are pretty slim. It seems like a really rough heuristic for a memory leak. 

Also by moving more logic out of listeners doesn't that mean we lessen our ability to detect a memory leak? IE: If the wallet fails to properly remove an objective and keeps eating up memory with it's collection of `InProgress` objectives there'd be nothing to catch that. 

> > couldn't we just query the objectives?
> 
> Can you explain your suggestion in words or pseudocode? I think what you're suggesting introduces unnecessary DB calls, which can be a problem under load.

This is what I sketched out but it probably requires too many DB calls:
```
  private async ensureObjectives(objectiveIds: string[]): Promise<void> {
    try {
      await backOff(
        async () => {
          const {outbox} = await this._wallet.syncObjectives(objectiveIds);
          await this._messageService.send(getMessages(outbox));

          if (await this.objectivesIncomplete(objectiveIds)) {
            // Throwing an error indicates to the backoff library that the task is not complete
            throw new Error('Still objectives to complete');
          }
        },

        {...this._backOffOptions, retry: async () => this.objectivesIncomplete(objectiveIds)}
      );
    } catch (error) {
      const remaining = await this.getIncompleteObjectives(objectiveIds);
      this._wallet.logger.error('Unable to ensure objectives', {
        remaining: Array.from(remaining.keys()),
      });
      throw new Error('Unable to ensure objectives');
    }
  }
```

packages/server-wallet/src/channel-manager.ts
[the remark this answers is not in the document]
lalexgap: From the [docs](https://www.npmjs.com/package/exponential-backoff):

```
retry?: (e: any, attemptNumber: number) => boolean | Promise<boolean>

The retry function can be used to run logic after every failed attempt (e.g. logging a message, assessing the last error, etc.). It is called with the last error and the upcoming attempt number. Returning true will retry the function as long as the numOfAttempts has not been exceeded. Returning false will end the execution.

Default value is a function that always returns true.
```

lalexgap:
> > I was debating this as well. Originally I was hoping to simplify the existing logic and get some extra features by using a library. However, I'm not sure if using the library simplified the logic.
> 
> I think what you're saying is that we'd have equally simple, or simpler logic, without using the library. If this is the case, writing our own implementation that suits our current needs has the benefit of:
> 
> 1. fewer dependencies (by just one, but still ...)
> 2. easier for teammates to understand the logic
> 
> (2) is the important thing to me. If we have to distill [this conversation](https://github.com/statechannels/statechannels/pull/3376#discussion_r599874296) into documentation explaining the implementation of `exponential-backoff`, we might be better suited to just write it ourselves.
> 
> What do you think?

I think `exponential-backoff` has all the features we want, it just works in a slightly different way than we're used to.  I think generally my preference is to rely on a library if it meets our needs since that way we get all the testing and maintenance for free. 

That said I don't feel too strongly about this particular case since a backoff is pretty trivial to write (and was previously implemented in ee2cd39)

packages/server-wallet/src/channel-manager.ts
andrewgordstewart: > > > Setting MaxListenerCount to a higher value just means it would take longer to detect the memory leak.
> > 
> > 
> > That's still a drawback! If it takes 12 hours under some load, to detect the memory leak, but we never leave a wallet running for 12 hours, we won't detect it.
> 
> It seems like a slightly convoluted constraint that we require the wallet have at most 10 listeners just for the purpose of a warning message?
> 
> I guess I feel that the benefits of the `maxListener` check are pretty slim. It seems like a really rough heuristic for a memory leak.

It is really easy to forget to do things like [this](https://github.com/statechannels/statechannels/commit/ee2cd39f84c672aed65303956e241d06b4941b3b#diff-d097197de4d5617abf526bc232c8ccd50cbebc3e695669e910b5cc77183ba6c3R67). The warning would point out that we haven't cleaned up in some code paths. To trigger the warning, we'd need an environment that hits all code paths "enough times".

To me, this is an indication that we should aim to _avoid using event emitters & listeners for managing internal logic_. I think it enables a lot of footguns, and makes things hard to debug. Since there's a general feeling that `wallet.ts` should not be an event emitter, I think it's important to consider what it would take to make that happen. 

> Also by moving more logic out of listeners doesn't that mean we lessen our ability to detect a memory leak? IE: If the wallet fails to properly remove an objective and keeps eating up memory with it's collection of `InProgress` objectives there'd be nothing to catch that.

A test can catch that, by asserting on the contents of `inProgress` objectives!

> This is what I sketched out but it probably requires too many DB calls:
> 
> (SNIP)
> 
> Both `this.objectivesIncomplete(objectiveIds)` and `this.getIncompleteObjectives(objectiveIds);` would be performing a query.

Yes, I think this is undesirable.

--- 

I think it's worth looking at this work from a higher level: we are _implicitly_ committed to a single wallet instance being responsible for managing channels. Why? Because two such wallet instances would both use the same private key to make blockchain transactions, and we'd result in nonce issues. We'd also have two chain services listening to, and trying to respond to, the same event.

Is it worth making this implicit assumption explicit? The benefit to doing that is, we can then manage some things in memory, such as a list of "in progress objectives". Keeping such a list in memory may allow us to make `wallet.ts` _stop being an event emitter_.

andrewgordstewart:
conclusion based on this discussion & ☠️IN PERSON CONVERSATION☠️:
- before merging, we'll remove the dependency on `exponential-backoff`, per [this](https://github.com/statechannels/statechannels/pull/3376#issuecomment-806092176)
- we'll flag the `wallet.on('objectiveSucceeded')` topic for a near-term design discussion, but we can merge this straight away

packages/server-wallet/src/channel-manager.ts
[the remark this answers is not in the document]
lalexgap: This no longer applies

packages/server-wallet/src/channel-manager.ts
lalexgap: Done in e389707

packages/server-wallet/src/channel-manager.ts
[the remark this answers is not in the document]
lalexgap: d470292cf

packages/server-wallet/src/__test-with-peers__/ensure-objectives.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    'can successfully create a channel with the latency options: %o',
+    async options => {
+      messageService.setLatencyOptions(options);
+      const channelManager = await ChannelManager.create(peerWallets.a, messageService);
+
+      peerWallets.b.on('objectiveStarted', async (o: WalletObjective) => {
+        await peerWallets.b.joinChannels([o.data.targetChannelId]);
+      });
+
+      await channelManager.createChannels(getCreateChannelsArgs(), 50);
```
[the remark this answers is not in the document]
lalexgap: b5e9f9f

packages/server-wallet/src/channel-manager.ts
lalexgap: As discussed in standup, we are planning on renaming the `ChannelManager` to `Wallet` and `Wallet` to `Engine`.

packages/server-wallet/src/channel-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+  ) {}
+
+  /**
+   * TODO: This is a basic implementation of createChannels.
+   * This will be cleaned up in https://github.com/statechannels/statechannels/issues/3365
+   * @param args
+   * @param numberOfChannels
+   * @returns
+   */
+  public async createChannels(args: CreateChannelParams, numberOfChannels: number): Promise<void> {
```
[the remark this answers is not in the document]
lalexgap: > * `createChannels` should enable the user to create channels that have different `CreateChannelParams`. so, could accept `CreateChannelParams[]`

Is that a more common case than creating a bunch of channels with the same `CreateChannelParams` ? My preference would be one of these:
- `CreateChannels(params:CreateChannelParams, numberOfChannels:number) `
- `CreateChannels(params:CreateChannelParams[])`

I don't think we should try to handle both cases with the same API call since that would make it more clunky to use.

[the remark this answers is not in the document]
andrewgordstewart: I think we should only handle `createChannels(params:CreateChannelParams[])`.
A user who wishes to call ```fancyCreateChannels(params:CreateChannelParams, numberOfChannels:number)``` can call ```createChannels(Array(n).fill(params))```

PR #3377, opened 2021-03-15
Use a full ETHAssetHolder during pushOutcomeAndTransferAll test

geoknee:
Currently, we use a test contract derived from the base AssetHolder contract. This means that funds are not actually transferred out of the contract during testing, resulting in an underestimate of gas consumed. 

This PR rectifies that by using a contract derived from the full ETHAssetHolder.

## Changes [Optional] 
Because ETH actually gets transferred now, the test script must also deposit ETH into the ETHAssetHolders. (There are two deployed for our test suite).

Also changed is the
- python script
- generated png (and svg) used in our blog post

We may use the updated image to publish a new post updating the community on (dollar) gas costs in our system. 

Note: the (gas) gas costs have changed over recent months due to many small changes to the contracts and solc compiler options. 

---
### Code quality
### Project management


packages/nitro-protocol/contracts/test/TestEthAssetHolder2.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.4;
+pragma experimental ABIEncoderV2;
+import '../ETHAssetHolder.sol';
+
+/**
+ * @dev This contract is a clone of the ETHAssetHolder contract. It will be initialized to point to the TestNitroAdjudicator.
+ */
+contract TestEthAssetHolder is ETHAssetHolder {
+    uint256 public dummy2; // this is simply to make the contract have distinct bytecode from the ETHAssetHolder (otherwise bytecode verification can fail)
```
andrewgordstewart, line 10: The only difference between this file and `TestEthAssetHolder` seems to be `dummy` vs `dummy2`. What's the purpose of this file?

andrewgordstewart: Ok, I figured it out by reading the tests. I think there are tests that use multiple EthAssetHolders, and you wanted to use different addresses for them? It's hard to tell for sure.

packages/nitro-protocol/deployment/deploy.ts
```diff
@@ -56,6 +57,11 @@ export async function deploy(): Promise<Record<string, string>> {
     {},
     TEST_NITRO_ADJUDICATOR_ADDRESS
   );
+  const TEST_ETH_ASSET_HOLDER2_ADDRESS = await deployer.deploy(
```
andrewgordstewart, line 60: Is there a risk of `deploy.js` being used in "production"? (to deploy to mainnet, lets say?)

The `TestEthAssetHolder` contracts now have `setHoldings`, which mean we only want to deploy it to a test environment. The reason for asking then, is to determine: how can we ensure that `TestEthAssetHolder` is only used in a test environment?

packages/nitro-protocol/contracts/test/TestEthAssetHolder.sol
```diff
@@ -2,12 +2,22 @@
...
     uint256 public dummy; // this is simply to make the contract have distinct bytecode from the ETHAssetHolder (otherwise bytecode verification can fail)
 
+    /**
+     * @dev Manually set the holdings mapping to a given amount for a given channelId.  Shortcuts the deposit workflow (ONLY USE IN A TESTING ENVIRONMENT)
+     * @param channelId Unique identifier for a state channel.
+     * @param amount The number of assets that should now be "escrowed: against channelId
+     */
+    function setHoldings(bytes32 channelId, uint256 amount) external {
+        holdings[channelId] = amount;
+    }
```
andrewgordstewart, line 19: I don't really understand why this change makes it a "full eth asset holder".

packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
```diff
@@ -173,6 +173,22 @@ describe('concludePushOutcomeAndTransferAll', () => {
...
+        ).wait();
+      }
+      if ('ETH2' in heldBefore) {
+        // Transfer some funds in to the other ETHAssetHolder. Use zero channel address for ease.
+        await (
+          await EthAssetHolder2.deposit(channelId, '0x00', heldBefore.ETH2.c, {
+            value: heldBefore.ETH2.c,
+          })
+        ).wait();
+      }
```
andrewgordstewart: Is this the change that makes means we are now "[using] a full eth asset holder in tests"?

packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
```diff
@@ -1,8 +1,8 @@
 import {expectRevert} from '@statechannels/devtools';
-import {Contract, Wallet, ethers, BigNumber} from 'ethers';
+import {Contract, Wallet, ethers, BigNumber, utils, constants} from 'ethers';
 
-import AssetHolderArtifact1 from '../../../artifacts/contracts/test/TESTAssetHolder.sol/TESTAssetHolder.json';
-import AssetHolderArtifact2 from '../../../artifacts/contracts/test/TESTAssetHolder2.sol/TESTAssetHolder2.json';
+import EthAssetHolderArtifact1 from '../../../artifacts/contracts/test/TestEthAssetHolder.sol/TestEthAssetHolder.json';
+import EthAssetHolderArtifact2 from '../../../artifacts/contracts/test/TestEthAssetHolder2.sol/TestEthAssetHolder.json';
```
andrewgordstewart, line 5: Ok, now I think I understand. The diff made me think there was a rename happening, combined with other changes. (There is a rename, as `AssetHolder1` is renamed to `EthAssetHolder1`.)

But this change is _not_ a rename, and you are importing

8b173676315f947ae8f9730b5365ea03526dcaa3 would have been more clear if it were written something like

```
use TestEthAssetHolder in pushOutcomeAndTransferAll test

The TESTAssetHolder that is currently used is not a real asset holder,
since it doesn't send funds to the destination. This is because it
extends the base AssetHolder contract.

We switch to TestEthAssetHolder & TestEthAssetHolder2 to
get a more accurate measure of gas consumption.
```

andrewgordstewart (approved):
Assuming that you're going to use squash-and-merge, I might suggest clarifying the resulting title & commit message with something like

```
use TestEthAssetHolder in pushOutcomeAndTransferAll test

The TESTAssetHolder that is currently used is not a real asset holder,
since it doesn't send funds to the destination. This is because it
extends the base AssetHolder contract.

We switch to TestEthAssetHolder & TestEthAssetHolder2, which
extend the EthAssetHolder contract. This gives a more accurate
measure of gas consumption.
```

packages/nitro-protocol/deployment/deploy.ts
geoknee: I'm going to address this in a separate PR.

packages/nitro-protocol/contracts/test/TestEthAssetHolder.sol
geoknee: This isn't. This is just a change that was necessary to use the full eth asset holder in the current test. I have another PR in the pipe that will actually do away with this contract altogether.

packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
geoknee: No, this is just another follow on change that is necessary, given that we now use the full eth asset holder in tests. You can't pay ETH out, unless you have paid some in :-)

PR #3393, opened 2021-03-22
change many forceMove -> challenge

geoknee:
Following #2963 , I found a few more instances where the old terminology survives. 

As the linked PR does not have a description, I'll record the reasoning here from memory.

It's not ideal to have the same name for a method and a contract. It is also not ideal to use a protocol-specific term where a more self-contained one would do. The change from `forceMove` to `challenge` reduces the overall size of our terminology.

### DIFF
--- a/packages/nitro-protocol/src/transactions.ts
+++ b/packages/nitro-protocol/src/transactions.ts
@@ -1,6 +1,5 @@
-import {Contract, providers, Signature} from 'ethers';
+import {providers, Signature} from 'ethers';
 
-import {Uint256} from './contract/types';
 import {State} from './contract/state';
 import * as forceMoveTrans from './contract/transaction-creators/force-move';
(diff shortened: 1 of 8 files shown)


andrewgordstewart (approved):

packages/nitro-protocol/src/transactions.ts
```diff
@@ -2,7 +2,7 @@ import {Contract, providers, Signature} from 'ethers';
 
 import {Uint256} from './contract/types';
 import {State} from './contract/state';
-import * as forceMoveTrans from './contract/transaction-creators/force-move';
+import * as challengeTrans from './contract/transaction-creators/force-move';
```
geoknee, line 5: The changes to this file don't look quite right, since forceMove here refers to the contract name (which we are keeping). Will fix.

PR #3394, opened 2021-03-22
Delete statusOf after a successful concludePushOutcomeAndTransferAll 

geoknee:
This change saves some gas by undoing the write to adjudicator storage that is currently part of `concludePushOutcomeAndTransferAll`.

Currently, gas costs are high and opportunities to reduce them are a priority. Opportunities should be treated with caution, however, since there is a high chance of introducing new attack vectors. 

This approach is good because it saves a substantial amount of gas, and while it does reduce some of the protection offered by our on-chain code, I believe that it does not introduce new attack vectors, or that the new vectors are easily blocked by some small changes to client behaviour. This can be realised by examining our current security assumptions.

A `finalised` status in the adjudicator is a terminal state and [**prevents**](https://docs.statechannels.org/docs/protocol-tutorial/understand-adjudicator-status#channel-modes) basically any change of state in the adjudicator, including:
- challenges
- concludes
- checkpoints
- respond

We need to keep that state in the machine because of the unhappy path to finalisation, a timed-out challenge. When that implicit state transition happens, the stored information **enables** the stored outcome to be "pushed" into asset holders, using `pushOutcome` and `pushOutcomeAndTransferAll`. In the happy path, however, (covered by `concludePushOutcomeAndTransferAll`), the outcome has already been pushed, and all assets have been transferred out of the channel, too). 

This change makes what was previously enabled, disabled, and what was previously prevented, possible. 

As we currently [hint at in the docs](https://docs.statechannels.org/docs/protocol-tutorial/precautions-and-limits#precautions), state channel client code should reject or ignore channels that are proposed with am "old" `channelNonce` (meaning one previously used by the same set of participants on the same chain). Correct clients (those that carefully uphold this principle) are therefore already prevented from depositing assets into an old channel if they only deposit into a channel during the setup stage.

Clients should therefore also be careful not to deposit assets into a channel that has been finalised. (This is almost the same principle, but not quite: a finalized channel may not necessarily be "old" if it is the most recent for a given `participants` array). A downside of the approach of this PR is that currently, clients can reach out to the chain for information on finalised channels. With this change, they will have to track that information themselves.  This does seem, however, very much in the spirit of state channels (why should we pay high costs to make this information public, when it only concerns a small number of parties?).

Another downside of this approach is that we end up doing some work that we potentially don't need to do:
1) computing and setting some storage that ends up being deleted anyway
2) emitting an event despite the lack of any change in state of the contract

I suggest tackling these in follow up work. 

- [x] Is it security related? Yes

---
### Code quality
- [ ] I have written clear commit messages
- [ ] I have performed a self-review of my own code
- [ ] I have separated logic changes from refactor changes (formatting, renames, etc.)
- [ ] I have commented my code wherever necessary (can be 0)
### Project management
- [ ] I have applied the [appropriate labels](https://www.notion.so/Team-working-agreements-2a95c926bb5642e5a5c42e4b74a9dd24#b304e56734a74dfbb341b8b4b27b1c0c)
- [ ] I have [linked to all dependent issues](https://help.zenhub.com/support/solutions/articles/43000010349-create-github-issue-dependencies) (can be 0)
- [ ] I have chosen the appropriate [pipeline](https://www.notion.so/Team-working-agreements-2a95c926bb5642e5a5c42e4b74a9dd24#d534e68e7edc46fe8a4cda61b2258c4e) on zenhub for the linked issue

### DIFF
--- a/packages/nitro-protocol/contracts/NitroAdjudicator.sol
+++ b/packages/nitro-protocol/contracts/NitroAdjudicator.sol
@@ -111,6 +111,10 @@ contract NitroAdjudicator is IAdjudicator, ForceMove {
             sigs
         );
         _transferAllFromAllAssetHolders(channelId, outcomeBytes);
+        // Now put the channel back into "Open" mode. This opens up the possibility
+        // of checkpointing, challenging, or concluding again. But the channel is now completely defunded:
+        // So clients have on reason to do that, nor to care if a counterparty does that.
+        delete statusOf[channelId]; // So that this method leaves no trace on chain whatsoever (for this channel)
     }
 
     /**
--- a/packages/nitro-protocol/contracts/NitroAdjudicator.sol
+++ b/packages/nitro-protocol/contracts/NitroAdjudicator.sol
@@ -113,7 +113,7 @@ contract NitroAdjudicator is IAdjudicator, ForceMove {
         _transferAllFromAllAssetHolders(channelId, outcomeBytes);
         // Now put the channel back into "Open" mode. This opens up the possibility
         // of checkpointing, challenging, or concluding again. But the channel is now completely defunded:
-        // So clients have on reason to do that, nor to care if a counterparty does that.
+        // So clients have no reason to do that, nor to care if a counterparty does that.
         delete statusOf[channelId]; // So that this method leaves no trace on chain whatsoever (for this channel)
     }
 
--- a/packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
+++ b/packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
@@ -238,16 +238,8 @@ describe('concludePushOutcomeAndTransferAll', () => {
           receipt.gasUsed
         );
 
-        // Compute expected ChannelDataHash
-        const blockTimestamp = (await provider.getBlock(receipt.blockNumber)).timestamp;
-        const expectedFingerprint = channelDataToStatus({
-          turnNumRecord: 0,
-          finalizesAt: blockTimestamp,
-          outcome,
-        });
-
-        // Check fingerprint against the expected value
-        expect(await NitroAdjudicator.statusOf(channelId)).toEqual(expectedFingerprint);
+        // Check fingerprint is reset (gas optimization)
+        expect(await NitroAdjudicator.statusOf(channelId)).toEqual(initialFingerprint);
 
         // Extract logs
         const {logs} = await (await tx).wait();
--- a/packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
+++ b/packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
@@ -238,6 +238,9 @@ describe('concludePushOutcomeAndTransferAll', () => {
           receipt.gasUsed
         );
 
+        // experiment:
+        const blockTimestamp = (await provider.getBlock(receipt.blockNumber)).timestamp;
+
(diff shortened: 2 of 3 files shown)


geoknee:
It looks like the chain service is getting thrown off by the channel not having a finalised status on chain. To get the tests to pass, we would need to resolve that somehow: under this approach we can't be calling out to the chain for this information.

geoknee:
Just underlining the fact that this change would need to be considered very carefully before merging. Even if we were 100% sure that it didn't introduce any new attacks, the fact that we have the TLA+ work and the audit on the existing code puts a lot of inertia in front of a change such as this.

geoknee:
Consider that any channel concluded/defunded off chain (such as a ledger or virtually funded channel), will not currently leave any imprint of its existence on the blockchain. So _we already rely on tracking channelNonces / channelIds off chain_ in order to protect against a replay attack. 

What happens if a state channel participant "loses their database"? This is fairly catastrophic, since that participant loses their channel signing key (and so cannot make any new updates to any of their channels), loses their states (which could be worth a lot of money), and loses their list of channel nonces (which protect them from unwittingly joining an old channel and suffering a replay attack). 

The current behaviour, where the chain stores *some* (but not all) ids for finalized channels, may offer some limited form of protection, even in the case of a lost database. If the participant had backed up their signing key, but had still lost their list of channel nonces, they could continue to open new channels with that signing key and not risk replay attacks from any channel finalized on chain. To me that seems like a pretty edge case and (probably) not worth supporting.

geoknee:
It seems to me that the right time for a wallet to consider a channel finalised is actually at the moment of signing an `isFinal=true` state. This is the moment where my wallet enables the channel to be finalised -- even if that has not happened yet, I cannot prevent it from happening as soon as I broadcast that signature. 

An attack could work like this: imagine Alice signs an `isFinal` state for a channel with Bob with `channelNonce = 1`, and with outcome `{Alice:1,  Bob: 9}`. Bob countersigns it (completing the conclusion proof), but holds on to that proof for now. Alice's wallet crashes, and when she recovers Bob proposes the same channel again, but with outcome `{Alice: 0, Bob: 10}`. Alice does a quick check on chain but sees no evidence this channel has been finalised. (Note that if Alice is very savvy, she may notice that the channel is already *funded* on chain, which might make her suspicious). Let's say her wallet is not programmed to check the `channelNonce` in Bob's proposal, and Alice proceeds to sign states in the channel. Now Bob proceeds to pay Alice in this channel, and Alice responds by supplying goods/services to Bob until the outcome is `{Alice:10, Bob:0}`. Bob submits the old conclusion proof and gets most of his money back.

So even for a directly funded channel, there may not yet be the imprint on chain to remind me not to reuse a `channelNonce`. This is another hint that wallets should take full responsibility for tracking `channelNonces`, and the current ability to query the chain is actually something of a red herring -- it only protects my wallet in certain circumstances. 

I would further argue that even having that partial protection in place (the on chain record of finalised channels) is damaging in the sense that it may have delayed us from tackling the problem in the generality it deserves.

geoknee:
I'd like to move this idea and discussion to [an RFC](https://www.notion.so/statechannels/RFC-11-Delete-statusOf-after-a-successful-concludePushOutcomeAndTransferAll-85d31bf223a54408a64439302cc4973c), so that we can return to it at a later date.

andrewgordstewart:
> I'd like to move this idea and discussion to [an RFC](https://www.notion.so/statechannels/RFC-11-Delete-statusOf-after-a-successful-concludePushOutcomeAndTransferAll-85d31bf223a54408a64439302cc4973c), so that we can return to it at a later date.

It's likely to be superseded by ninja nitro anyway, right?

geoknee:
Yes, but in case we end up sticking with the current implementation in some guise, we might still want to do something like this. And we may be able to move forward with some changes to the wallet code regardless.

geoknee:
(I plan to leave the RFC as a draft for now)

PR #3404, opened 2021-03-25
remove duplicated code

geoknee:


packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -297,9 +297,6 @@ contract ForceMove is IForceMove {
         require(largestTurnNum + 1 >= numStates, 'largestTurnNum too low');
         // ^^ SW-C101: prevent underflow
 
-        channelId = _getChannelId(fixedPart);
-        _requireChannelNotFinalized(channelId);
```
geoknee, line 301: These lines are identical to L286-287 above.

andrewgordstewart (commented):
Why is there a submodule added?

geoknee:
Whoops. I fixed it.

andrewgordstewart (approved):

PR #3402, opened 2021-03-24
server <-> browser interoperability test 

kerzhner:
This PR tackles all but the last bullet point of https://github.com/statechannels/statechannels/issues/3383.

The goal of this PR is to create a playground that allows for experimentation on how to share code between server and browser wallets. In this PR:
- `interop-tests` is a new package. The package contains a passing test that established a directly-funded channel between a server wallet and a browser wallet.
- `browser-wallet.ts` in the `xstate-wallet` package is a tiny wallet that re-uses the existing xstate store and chain watcher.
- `browser-wallet.ts` does **not** attempt to describe how reusable protocols should be written. This is to be done in follow up work.

This PR should be follow up with https://github.com/statechannels/statechannels/issues/3400 and https://github.com/statechannels/statechannels/issues/3401.

Fixes https://github.com/statechannels/statechannels/issues/3383 since other issues cover remaining work of https://github.com/statechannels/statechannels/issues/3383.

Issue #3401, opened 2021-03-24
BrowserWallet should function in the browser


kerzhner:
We currently have:
- [`BrowserWallet`](https://github.com/statechannels/statechannels/blob/390adf41660a26915fe5864473f14ebf2affe7db/packages/xstate-wallet/src/browser-wallet.ts#L34) class that was created to bootstrap work on the interoperable browser wallet.
- [`xstate-wallet`](https://github.com/statechannels/statechannels/blob/390adf41660a26915fe5864473f14ebf2affe7db/packages/xstate-wallet/src/index.ts#L23) that used to run (and work) in the browser.

These 2 "wallets" should converge into one, interoperable browser wallet that works in the browser. We should aim to reuse as much `xstate-wallet` logic as is reasonable.

Issue #3400, opened 2021-03-24
Reusable direct funding protocol logic


kerzhner:
The goal is to write the logic of the direct funding protocol in a way that can be reused between wallet implementations. Working on this card means figuring out what patterns to use for writing reusable protocols.

For reference, we already have the following attempts:
- From @andrewgordstewart: https://github.com/statechannels/statechannels/commit/b68da3563291fff10f8a5882e48d87e0e0dfdde5#diff-070b37813318b3195790b369a077a06cbc15c4fb1d254366b3148593798381b6R41. Andrew says: "You could shoehorn that kind of function into the xstate-wallet, somewhere like here: https://github.com/statechannels/legacy-packages/blob/1dd0181882e7719f104e31cc481d2247a137c8f9/packages/xstate-wallet/src/channel-wallet.ts#L192"
- From @kerzhner: https://github.com/statechannels/statechannels/blob/390adf41660a26915fe5864473f14ebf2affe7db/packages/xstate-wallet/src/browser-wallet.ts#L120

PR #3410, opened 2021-03-25
Add the latest browser wallet implementation to existing xstate wallet

kerzhner:
This work follows up on https://github.com/statechannels/statechannels/pull/3402.

The goal of this PR is to unblock work on https://github.com/statechannels/statechannels/issues/3400 and https://github.com/statechannels/statechannels/issues/3401.

This PR removes the new `BrowserWallet` introduced in https://github.com/statechannels/statechannels/pull/3402, and pulls in that functionality into the `ChannelWallet`. With this PR:
- The interop test instantiates the `ChannelWallet` (which is just the old xstate wallet).
- `pushMessage` of `ChannelWallet` implements the latest state channels protocol.

The overarching idea with this work is to gradually refactor/reimplement parts of the ChannelWallet to achieve protocol compatibility with the server wallet.


.circleci/config.yml
```diff
@@ -141,9 +141,14 @@ jobs:
       - run:
           name: wallet setup
           working_directory: packages/server-wallet
+          # Create databases for the server-wallet and interop-tests package tests
```
kerzhner, line 144: These changes are part of a preceding [PR](https://github.com/statechannels/statechannels/pull/3422)

packages/server-wallet/deployment/deploy.ts
```diff
@@ -15,9 +15,10 @@ export type TestNetworkContext = {
 
 export async function deploy(): Promise<TestNetworkContext> {
   const ethereumPrivateKey = ETHERLIME_ACCOUNTS[0].privateKey;
+  const ganachePort = Number(process.env.GANACHE_PORT);
```
kerzhner, line 18: These changes are part of a preceding [PR](https://github.com/statechannels/statechannels/pull/3422)

.circleci/config.yml
andrewgordstewart: Why are they still in this changeset?

packages/xstate-wallet/src/channel-wallet2.ts
```diff
@@ -0,0 +1,202 @@
+import {
```
andrewgordstewart: How much of this code is copy/paste?

It is difficult to review the diff, without opening up two files side by side. Should I instead simply review the content of `channel-wallet2.ts`, ignoring the content of `browser-wallet.ts` that was deleted?

.circleci/config.yml
kerzhner: Not sure...

packages/xstate-wallet/src/channel-wallet2.ts
kerzhner: I would recommend **not** reviewing `channel-wallet2.ts`. This file was removed in https://github.com/statechannels/statechannels/pull/3410/commits/d8a62b72ac6b0150917d967738e7667c369ecec4. The commit history could use a cleanup.

In this PR, the code added to `channel-wallet.ts` is a copy/paste from `browser-wallet.ts`.

.circleci/config.yml
andrewgordstewart: I think it's because #3422 was merged with a squash commit, meaning Github does not recognize that the diff in a5c67e7 is already included in `master`.

kerzhner:
Thanks for the feedback! Agreed that this is not an easy PR to review. For the `ChainWatcher` `chainAddress` constructor parameter, I added my reasoning to https://github.com/statechannels/statechannels/pull/3410/commits/ab280e12e5ae572454cae52550d1f650deb5e930.

kerzhner:
> This might beg the question: should the `ChainService` _always_ accept an address as a constructor param? And we read off of `window.ethereum` before constructing a `ChainService`?

We should definitely consider that.

Issue #3418, opened 2021-03-26
Implement a pure-function-based DirectFunder.


andrewgordstewart:
# Why: Problem description :confused:
Sharing code between the browser & the server is currently difficult because a lot of logic is delegated to the `Store`. Since the full `Store` interface needs to be implemented in each environment, this would to a lot of repeated code and difficulty maintaining parallel implementations.

A protocol implementation with a signature such as
```
function openChannelCranker(
  objective: OpenChannelObjective,
  event: OpenChannelEvent,
  myPrivateKey: string
): SomeResult
```
would
A. simplify the `Store` interface, making it easier to implement in each environment
B. simplify unit testing: no DB interaction is required
C. facilitate more rigorous testing strategies, such as property-based testing

# What: Describe the solution you'd like :sparkles:
Use something similar to the prototype code [here](https://github.com/statechannels/statechannels/commit/b68da3563291fff10f8a5882e48d87e0e0dfdde5#diff-070b37813318b3195790b369a077a06cbc15c4fb1d254366b3148593798381b6R41-R45) in the `server-wallet`. This probably ought to be a _new_ objective, so that the work can be done side-by-side with the current implementation. This allows us to compare the two implementations, to decide whether this is the correct approach.

- [x] #3426 Implement the logic in `wallet-core`, with tests.
- [x] #3437 and #3439 Use the logic in `server-wallet`.
- [ ] Commit to approach and clean up code.
- [ ] Update the `Store` or the DB schema to store directly funded objective state.

## Code cleanup
- [ ] [Move types to their own file](https://github.com/statechannels/statechannels/pull/3426#discussion_r605760179)

# Describe alternatives you've considered :thinking:
An alternative towards an interoperable browser wallet is to fully implement the existing `server-wallet` `Store` "interface" (it is currently a class, not a type) in a browser setting. This approach is ruled out due to stated benefits A,B,C.

Another reason identified both the `server-wallet` and `xstate-wallet` `Store` classes seem to have too much responsibility to share an interface in a maintainable fashion. For instance, the former implements things like `signAndAddState`, `pushMessage`, and `approveObjective`. The latter implements `pushMessage`, `signAndAddStateWithinTx`, `reserve/releaseFunds`. Wallets in these two settings may have different data storage responsibilities (such as managing budgets for sites, in the browser).

[...]

PR #3426, opened 2021-03-30
Implement pure-function DirectFunder in wallet-core

andrewgordstewart:
# Description
This implements one pure-function variation of a direct-funder "protocol", which is the first step towards #3418.

This introduces a "rich objective" data structure called, for lack of more creativity, an `OpenChannelObjective`. (I suggest we review terminology here.) This "rich objective" stores:
- the target channel id
- the goal starting state of the target channel
- my index in the channel

It then stores some additional data:
- signatures on the expected preFS state's hash
- on-chain funding status
- funding requests submitted
- signatures on the expected postFS state's hash

The cranker receives the current "objective state" as well as an event. This updates the "objective state" and generates some actions. (In other words, the cranker is a state chart!)

This implementation is a deviation from the current wallet behaviour. The current wallet
1. adds arbitrary states into the `Store`, or updates the funding on the channel
2. runs the objective cranker, asynchronously triggering actions as well as updating the store

IE. in my implementation here, 1 & 2 are both consumed by the logic of the cranker. This allows the cranker to **reject state updates that are not a part of the objective**. (I would suggest _aborting the protocol_ in this event, as it is a sign that your peer is not functioning according to the protocol.) In other words, states are _whitelisted_ while the objective is running.

### What are the benefits?
See [here](https://www.notion.so/statechannels/RFC-10-State-Channel-Protocol-SCP-fc7e468a160c45bcbf5b28653dcf7ff6#8d4e0c219c854b64b289c64dfeacbf29).

### How is the rich objective meant to be stored?
This change does not prescribe how the rich objective is meant to be stored in the DB.

The naive way is to directly store the rich objective in the DB. This is an _easy_ solution, and may be preferred. Indeed, it would mean that to crank an objective, we would only need to execute one, easy query.

However, there is no reason we cannot use the current `channels` table in either the Postgres store or the Dexie backend to store the pre/postFS states + funding data. We would then need to write "adapter code" to compute the rich data structure accepted by the cranker.

Storing the updated rich objective within the current `channels` table is more subtle, but still fine. Considering the recommendation [here](https://www.notion.so/statechannels/RFC-8-Securing-Private-StateChannel-Key-c4f2945172d84db2b84220d64502e38c#8dfd6d2275714c2399655a7c6738248d), a wallet will likely provide a `Keyring` that will be used to (asynchronously) sign states. In this case, we would probably change the cranker to _declare which hashes the keyring should sign_, rather than accept a private key by the cranker. In this case, when the cranker generates a `SendState` action, the imperative shell would
1. use the keyring to sign the state
2. attach the signature to the given state, and send the message
3. attach the signature to the given state, and store it in the `channels` table
This can be factored out to be re-usable by any protocol (or even an `AppManager`)

The naive approach is my preference, as it is easy to write, and also _limits what states you have to consider_ via the schema itself. On the contrary, if we fetch a record from the `channels` table, where an arbitrary bag of states can be stored, we have to ask these kinds of questions about code:
- do we have a supported state?
- if not, what's the turn number & outcome of the latest state?
- does the latest state match with the state we agreed to open the channel with?
- is the outcome even an allocation outcome? maybe it's a guarantor channel?
We can choose to ignore these questions via type assertions, but I think that's worse than creating a whitelist of the explicit states you're expecting, and comparing them against that!

It is a **shortcoming** of this approach that we either need to write an awkward adapter or change our storage to use this cranker.

I hand-wrote a specific sequence of events in `direct-funder.test.ts`. As indicated there, a more robust test of the protocol would consider all of the permutations of the order of events. I suggest this as future work. 

I think so? It is a significant change to wallet design.

I have not quite finished this [design doc](https://www.notion.so/statechannels/RFC-10-State-Channel-Protocol-SCP-fc7e468a160c45bcbf5b28653dcf7ff6), but it might still be worth a read.

---
### Code quality
- [ ] ~I have separated logic changes from refactor changes (formatting, renames, etc.)~
- [ ] ~I have written clear commit messages~
       My apologies. I am out of practice writing code, and impatiently put this together.
### Project management

### DIFF
--- a/packages/wallet-core/src/protocols/direct-funder.ts
+++ b/packages/wallet-core/src/protocols/direct-funder.ts
@@ -1,23 +1,23 @@
-import {ethers} from 'ethers';
 import * as _ from 'lodash';
 
+import {checkThat, isSimpleAllocation} from '../utils';
 import {Message} from '../wire-protocol';
 import {BN} from '../bignumber';
 import {addHash, hashState, signState} from '../state-utils';
 import {Address, SignatureEntry, State, StateWithHash, Uint256} from '../types';
-import {serializeMessage} from '../serde/wire-format/serialize';
 
-const WALLET_VERSION = 'pure-function-protocol'; // FIXME where does this come from?
-type Outgoing = any; // FIXME
+type AddressedMessage = {recipient: string; message: Message};
 
 // FIXME: For the purpose of prototyping, I am ignoring blockchain events.
-type OpenChannelEvent = Message;
+export type OpenChannelEvent = Message;
 
 type SignedStateHash = {hash: string; signatures: SignatureEntry[]};
 
 export type OpenChannelObjective = {
   channelId: string;
   openingState: State;
+  myIndex: number;
+
   preFS: SignedStateHash;
   // FIXME: The asset class is _ignored_ here.
   funding: {amount: Uint256; finalized: boolean};
@@ -25,8 +25,8 @@ export type OpenChannelObjective = {
   postFS: SignedStateHash;
 };
 
-type Action =
-  | {type: 'sendMessage'; message: Outgoing['params']}
+export type Action =
+  | {type: 'sendMessage'; message: AddressedMessage}
   // FIXME: What data is required here?
   | {type: 'deposit'; amount: Uint256};
 
@@ -58,30 +58,33 @@ export function openChannelCranker(
   event: OpenChannelEvent,
   myPrivateKey: string
 ): Result {
-  const {address} = new ethers.Wallet(myPrivateKey);
-  const me: Address = address as Address;
+  const {participants} = objective.openingState;
+  const me = participants[objective.myIndex];
 
   // First, receive the message
 
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const signedState = event.signedStates![0];
-  const hash = hashState(signedState);
-  const {signatures} = signedState;
+  // FIXME: Assume there's only one signed state
+  if (event.signedStates && event.signedStates[0]) {
+    const signedState = event.signedStates[0];
+    const hash = hashState(signedState);
+    const {signatures} = signedState;
 
-  if (hash === objective.preFS.hash) {
-    objective.preFS.signatures = mergeSignatures(objective.preFS.signatures, signatures);
-  } else if (hash === objective.postFS.hash) {
-    objective.postFS.signatures = mergeSignatures(objective.postFS.signatures, signatures);
-  } else {
-    throw new Error(
-      `Unexpected state hash ${hash}. Expecting ${objective.preFS.hash} or ${objective.postFS.hash}`
-    );
+    if (hash === objective.preFS.hash) {
+      objective.preFS.signatures = mergeSignatures(objective.preFS.signatures, signatures);
+    } else if (hash === objective.postFS.hash) {
+      objective.postFS.signatures = mergeSignatures(objective.postFS.signatures, signatures);
+    } else {
+      throw new Error(
+        `Unexpected state hash ${hash}. Expecting ${objective.preFS.hash} or ${objective.postFS.hash}`
+      );
+    }
   }
 
   // Then, react:
   const actions: Action[] = [];
 
-  if (!signedbyMe(objective, 'preFS', me)) {
+  if (!signedbyMe(objective, 'preFS', me.signingAddress)) {
     signStateAction('preFS', myPrivateKey, objective, actions);
   }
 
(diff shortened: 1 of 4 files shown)


andrewgordstewart:
For some inspiration on this approach, consider watching [this](https://www.destroyallsoftware.com/talks/boundaries) or [this](https://www.destroyallsoftware.com/screencasts/catalog/functional-core-imperative-shell).

packages/wallet-core/src/protocols/direct-funder.ts
```diff
@@ -0,0 +1,253 @@
...
+  theirPostFundState = 'DirectFunder.theirPostFundSetup'
+}
+
+export type OpenChannelObjective = {
+  status: WaitingFor | 'success' | 'error';
+  channelId: string;
+  openingState: State;
+  myIndex: number;
+
+  preFS: SignedStateHash;
```
lalexgap, line 30: nit: I prefer something `preFundSetupHash` instead of `preFS`

packages/wallet-core/src/protocols/direct-funder.ts
```diff
@@ -0,0 +1,253 @@
...
+type SignedStateHash = {hash: string; signatures: SignatureEntry[]};
+
+export enum WaitingFor {
+  theirPreFundSetup = 'DirectFunder.theirPreFundSetup',
+  safeToDeposit = 'DirectFunder.safeToDeposit',
+  channelFunded = 'DirectFunder.channelFunded',
+  theirPostFundState = 'DirectFunder.theirPostFundSetup'
+}
+
+export type OpenChannelObjective = {
```
lalexgap, line 24: nit: might be cleaner to move types to their own file.

packages/wallet-core/src/protocols/direct-funder.ts
```diff
@@ -0,0 +1,253 @@
...
+import {addHash, hashState, signState} from '../state-utils';
+import {Address, SignatureEntry, State, StateWithHash, Uint256} from '../types';
+
+type AddressedMessage = {recipient: string; message: Message};
+
+export type OpenChannelEvent =
+  | {type: 'MessageReceived'; message: Message}
+  | {type: 'FundingUpdated'; amount: Uint256; finalized: boolean};
+
+type SignedStateHash = {hash: string; signatures: SignatureEntry[]};
```
lalexgap, line 15: Is this not already a type we have defined in wallet-core?

packages/wallet-core/src/protocols/direct-funder.ts
```diff
@@ -0,0 +1,253 @@
...
+    return {objective, actions};
+  } else {
+    // otherwise, deposit
+    const amount = BN.sub(targetAfter, funding.amount); // previous checks imply this is >0
+    actions.push({type: 'deposit', amount});
+    objective.status = WaitingFor.channelFunded;
+    return {objective, actions};
+  }
+
+  // Now that
```
lalexgap, line 166: Truncated comment

packages/wallet-core/src/protocols/direct-funder.ts
andrewgordstewart: Agreed, I'd do that if we choose to adopt this approach!

packages/wallet-core/src/protocols/direct-funder.ts
andrewgordstewart: I don't think so! We typically pass around states. In this work, I am suggesting a slightly clever form of compression:
1. the preFS and postFS states are the same, except for their turn number.
2. we can therefore compute the hashes up front, and

One benefit of this is that it enables some aggressive network optimization, where we can send
```
{ objectiveId: string; hashes: SignedStateHash[] }
```

across the wire, instead of this, which contains a lot more data.
```
{ signedStates: SignedState[] }
```

The first wire protocol also has an advantage in terms of security: messages for an unauthorized objective can be ignored.

packages/wallet-core/src/protocols/direct-funder.ts
andrewgordstewart: https://github.com/statechannels/statechannels/pull/3426/commits/3015fe46a22cd385f8c8e7413ca8970d0c3aa499

packages/wallet-core/src/protocols/direct-funder.ts
```diff
@@ -0,0 +1,262 @@
...
+ * A wallet implementation can then use the result with this sequence of asynchronous operations
+ * 1. record the new objective state as well as the resulting actions
+ * 2. trigger the resulting actions asynchronously
+ * 3. mark the actions as being successful
+ *
+ * If the wallet crashes after 1 & before 3, the wallet can decide to re-trigger
+ * the actions on a case-by-case basis, based on whether the action is safe to
+ * re-trigger.
+ */
+export function openChannelCranker(
```
lalexgap, line 77: This function is slightly long so it might be nice to split it into  `processEvents` /`processActions` functions?

lalexgap:
I think this approach makes sense to me!

lalexgap (approved):

PR #3427, opened 2021-03-30
Create Channels API

lalexgap:
# Description
Introduces the new create channels API on the wallet.

The main focus of this PR was updating the create channels API method to return a result that would be informative and useful. A new `ObjectiveResult` has been introduced that can be returned by the `Wallet` for any objective API calls.

The `ObjectiveResult` contains some basic information: `ChannelId, ObjectiveId, currentStatus` that gives some information about the objective and the current status.

It also includes a `done` promise. This promise resolves when the objective is complete. In the case of `createChannels` it resolves when the channel is fully funded and ready to use. The idea is that there is a simple promise that API users can `await` to get back a channel that is useful.

The `done` promise should **always resolve**. If an error occurs the promise should return an `ObjectiveError`. That way consumers of the API can know exactly what errors to expect.

### Questions
- Is it an antipattern to avoid rejecting in a promise?
-  Right now the `done` promise resolves to `{type:'ObjectiveSuccess}` on success. Should we be returning something different?
- What level of testing do we want? Currently there's just the `ensure-objective` test that calls this new `createChannels`. Should we write unit-level tests or should we focus on a E2E flow?

Since this is specifying how our API looks it probably makes sense to get input from multiple people.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/server-wallet.api.md
+++ b/packages/server-wallet/server-wallet.api.md
@@ -189,6 +189,11 @@ export function getDatabaseConnectionConfig(config: EngineConfig): DatabaseConne
     port: number;
 };
 
+// @public (undocumented)
+export function hasNewObjective(response: SingleChannelOutput): response is SingleChannelOutput & {
+    newObjective: WalletObjective;
+};
+
 // @public
 export type IncomingEngineConfig = RequiredEngineConfig & Partial<OptionalEngineConfig>;
 
@@ -330,7 +335,9 @@ export class SingleThreadedEngine extends EventEmitter<EventEmitterType> impleme
     closeChannels(channelIds: Bytes32[]): Promise<MultipleChannelOutput>;
     // (undocumented)
     static create(engineConfig: IncomingEngineConfig): Promise<SingleThreadedEngine>;
-    createChannel(args: CreateChannelParams): Promise<MultipleChannelOutput>;
+    createChannel(args: CreateChannelParams): Promise<SingleChannelOutput & {
+        newObjective: WalletObjective;
+    }>;
     createChannels(args: CreateChannelParams, numberOfChannels: number): Promise<MultipleChannelOutput>;
     createLedgerChannel(args: Pick<CreateChannelParams, 'participants' | 'allocations' | 'challengeDuration'>, fundingStrategy?: 'Direct' | 'Fake'): Promise<SingleChannelOutput>;
     destroy(): Promise<void>;
@@ -383,7 +390,7 @@ export function validateEngineConfig(config: Record<string, any>): {
 
 // Warnings were encountered during analysis:
 //
-// src/engine/types.ts:23:3 - (ae-forgotten-export) The symbol "WalletObjective" needs to be exported by the entry point index.d.ts
+// src/engine/types.ts:72:39 - (ae-forgotten-export) The symbol "WalletObjective" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
--- a/packages/server-wallet/src/engine/engine.ts
+++ b/packages/server-wallet/src/engine/engine.ts
@@ -72,6 +72,7 @@ import {
   Output,
   EngineInterface,
   EngineEvent,
+  hasNewObjective,
 } from './types';
 import {EngineResponse} from './engine-response';
 
@@ -438,14 +439,21 @@ export class SingleThreadedEngine
    * @param args - Parameters to create the channel with.
    * @returns A promise that resolves to the channel output.
    */
-  async createChannel(args: CreateChannelParams): Promise<SingleChannelOutput> {
+  async createChannel(
+    args: CreateChannelParams
+  ): Promise<SingleChannelOutput & {newObjective: WalletObjective}> {
     const response = EngineResponse.initialize();
 
     await this._createChannel(response, args, 'app');
 
     // NB: We intentionally do not call this.takeActions, because there are no actions to take when creating a channel.
 
-    return response.singleChannelOutput();
+    const result = response.singleChannelOutput();
+    if (!hasNewObjective(result)) {
+      throw new Error('No new objective created for create channel');
+    } else {
+      return result;
+    }
   }
   /**
    * Creates multiple channels with the same parameters. See {@link SingleThreadedEngine.createChannel}.
--- a/packages/server-wallet/src/engine/engine.ts
+++ b/packages/server-wallet/src/engine/engine.ts
@@ -438,14 +438,14 @@ export class SingleThreadedEngine
    * @param args - Parameters to create the channel with.
    * @returns A promise that resolves to the channel output.
    */
-  async createChannel(args: CreateChannelParams): Promise<MultipleChannelOutput> {
+  async createChannel(args: CreateChannelParams): Promise<SingleChannelOutput> {
     const response = EngineResponse.initialize();
 
     await this._createChannel(response, args, 'app');
 
     // NB: We intentionally do not call this.takeActions, because there are no actions to take when creating a channel.
 
-    return response.multipleChannelOutput();
+    return response.singleChannelOutput();
   }
   /**
    * Creates multiple channels with the same parameters. See {@link SingleThreadedEngine.createChannel}.
--- a/packages/server-wallet/src/wallet/types.ts
+++ b/packages/server-wallet/src/wallet/types.ts
@@ -38,11 +38,6 @@ export type ObjectiveSuccess = {type: 'Success'};
 export type ObjectiveDoneResult = ObjectiveSuccess | ObjectiveError;
 
 /**
- * TODO: Can the channelId just be on the objective?
- * Or do we anticipate objectives that cover multiple channels?
- */
-
-/**
  * This is what is returned for any objective related API call
  *
  */
--- a/packages/server-wallet/src/wallet/types.ts
+++ b/packages/server-wallet/src/wallet/types.ts
@@ -33,7 +33,7 @@ export type InternalError = {
   error: Error;
 };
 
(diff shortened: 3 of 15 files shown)


andrewgordstewart:
Assuming that the client does not crash, the proposed API seems to handle both the happy path and the sad path well!

Have you thought about how the client should handle crashes, with respect to the new API?

lalexgap:
> Assuming that the client does not crash, the proposed API seems to handle both the happy path and the sad path well!
> 
> Have you thought about how the client should handle crashes, with respect to the new API?

My rough plan for this was to introduce a method on the API that would look something like this `jumpstartObjectives(objectiveIds:string[]):ObjectiveResult` .  This could be used after a crash to start up the objectives and return promises to `await`.

andrewgordstewart:
> Is it an antipattern to avoid rejecting in a promise?

My uninformed opinion:
+: Always resolving gives the benefit of a well-typed API. Typescript devs can be forced to handle errors.
-: Developers might assume that when the problem resolves, the objective succeeded. Particularly bad for javascript devs that ignore type annotations.

> Right now the done promise resolves to {type:'ObjectiveSuccess} on success. Should we be returning something different?

I think we can add data as we see fit. There's no current reason to freeze the API right now.

> What level of testing do we want? Currently there's just the ensure-objective test that calls this new createChannels. Should we write unit-level tests or should we focus on a E2E flow?

What do you mean by unit-level tests? Do you mean stubbing out the implementation of `Wallet.ensureObjective`?

packages/server-wallet/server-wallet.api.md
```diff
@@ -235,6 +253,28 @@ export type NetworkConfiguration = {
...
+
+// @public (undocumented)
+export type ObjectiveError = EnsureObjectiveFailed | InternalError;
+
+// @public
+export type ObjectiveResult = {
+    done: Promise<ObjectiveDoneResult>;
+    currentStatus: ObjectiveStatus;
+    objectiveId: string;
+    channelId: string;
```
andrewgordstewart, line 267: Have you considered providing the `channelId` in the `ObjectiveDoneResult`?

Currently, a developer could
```
const {done, channelId} = await wallet.createChannels(...);
await done;
await wallet.updateChannel({channelId, ...});
```

resulting in a runtime error in the case that `done` rejects.

We could potentially nudge them into more robust error handling, at least in Typescript, with a flow like

```
const {done, objectiveId} = await wallet.createChannels(...);
const result = await done;
switch (result.type) {
  case 'ObjectiveSuccess':
    await wallet.updateChannel({channelId: result.channelId, ...});
    break;
  case 'ObjectiveFailure':
    await handleError(result.error);
    break;
}
```

lalexgap: That's a good point!

I did consider this a little bit but I wasn't sure if it would be strange to include the `channelId` on the `objective` as well as the `ObjectiveDoneResult`.

However, it does make the flow a lot nicer so I think it makes sense!

packages/server-wallet/src/engine/engine.ts
```diff
@@ -438,14 +438,14 @@ export class SingleThreadedEngine
    * @param args - Parameters to create the channel with.
    * @returns A promise that resolves to the channel output.
    */
-  async createChannel(args: CreateChannelParams): Promise<MultipleChannelOutput> {
+  async createChannel(args: CreateChannelParams): Promise<SingleChannelOutput> {
```
andrewgordstewart: nit: it'd have been helpful to reviewers to split up b24e50c4a538d640489b92935ac355cc67db1adc, which does two things:
1. changes the `Engine.createChannel` API
2. changes the `Engine.createChannels` API

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -47,16 +34,27 @@ export class Wallet {
...
+      channelParameters.map(async p => {
+        const createResult = await this._engine.createChannel(p);
+
+        const {newObjective, channelResult} = createResult;
+
+        return {
+          channelId: channelResult.channelId,
+          currentStatus: newObjective.status,
+          objectiveId: newObjective.objectiveId,
+          done: this.ensureObjective(newObjective, getMessages(createResult)),
```
andrewgordstewart, line 54: If `done` is never meant to reject, should we 
```suggestion
          done: this.ensureObjective(newObjective, getMessages(createResult)).catch(parseErrorSomehow?),
```

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -90,23 +85,15 @@ export class Wallet {
...
     for (let i = 0; i < numberOfAttempts; i++) {
-      if (remaining.size === 0) return;
-
+      if (isComplete) return {type: 'Success'};
       const delayAmount = initialDelay * Math.pow(multiple, i);
       await delay(delayAmount);
 
-      const {outbox} = await this._engine.syncObjectives(objectives.map(o => o.objectiveId));
+      const {outbox} = await this._engine.syncObjectives([objective.objectiveId]);
       await this._messageService.send(getMessages(outbox));
```
andrewgordstewart, line 93: Am I right to believe that in order to remove the `EventEmitter` behaviour from the `Engine` class, that the `MessageService` would have to collect new objectives created, and return a result?

packages/server-wallet/src/wallet/types.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export type ObjectiveSuccess = {type: 'Success'};
+
+export type ObjectiveDoneResult = ObjectiveSuccess | ObjectiveError;
+
+/**
+ * This is what is returned for any objective related API call
+ *
+ */
+export type ObjectiveResult = {
```
andrewgordstewart, line 44: I like this API! It seems to make it pretty easy to use the wallet.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -66,18 +64,15 @@ export class Wallet {
    * @param objectiveMessages The collection of outgoing messages related to the objective.
    * @returns A promise that resolves when all the objectives are completed
    */
-  private async ensureObjectives(
-    objectives: WalletObjective[],
+  private async ensureObjective(
```
andrewgordstewart, line 67: I'm not sure about this change. It would trigger a flurry of network messaging when a client calls `createChannels(Array(10).fill(withParams))`.

You could instead change the return value of `ensureObjectives` to return `Promise<(ObjectiveSuccess | ObjectiveFailure)[]>`.

That has the downside of `createChannels` needing to do something like
```
const allDonePromise = this.ensureObjectives([...]);
const donePromiseForObjective5 = allResults.then(results => results[5])
```

to construct the corresponding promise for objective 5.

Is there a more elegant solution?

packages/server-wallet/server-wallet.api.md
andrewgordstewart: Just to be clear, I'm suggesting _not_ including the channelId in the result of `createChannels`. This forces them to `await done` in order to access the channelId.

(I think it's valid to leave the responsibility of checking the type of the return promise up to the developer -- the above question is more open-ended.)

andrewgordstewart (approved):
LGTM, but you might want to reconsider implementing `ensureObjectives` to avoid additional networking.

packages/server-wallet/src/wallet/wallet.ts
andrewgordstewart: I guess the `MessageService` could also be an event emitter?

packages/server-wallet/src/wallet/wallet.ts
lalexgap: That's a good point. I meant to wire that up to the `InternalError` type, so that any unhandled error gets caught and we just return an `InternalError` to the user.

packages/server-wallet/src/wallet/wallet.ts
lalexgap: Good point!

Let me take a look into this.

lalexgap: So this actually turned out to be tricker than I thought. When switching to `ensureObjectives` it becomes non-trivial to setup the promises for each individual objective.

I've done this in f64a972 but would appreciate a look to see if there is nicer way of doing it.

packages/server-wallet/src/wallet/wallet.ts
lalexgap: With the changes in 68a57ff I think we no longer need this, since we now are manually resolving the promises.

packages/server-wallet/src/wallet/wallet.ts
andrewgordstewart: Another option is to push the complexity into the message service. It can collect and batch outgoing messages, for instance.

lalexgap: > Another option is to push the complexity into the message service. It can collect and batch outgoing messages, for instance.

Hmm, that's a good point. I think I'd lean towards the `wallet` owning most of that complexity because
- The wallet better context of what should and should not be batched together than a message service
- The message service implementation should probably be as small/simple as possible so it's easy to implement for various communication protocols.

I could see the message service implementing batching on top of that though.

andrewgordstewart: Maybe we probably shouldn't rely on the user supplied message service to do batching. But, different channels may have differing capacity to batch messages, so it's not crazy. (Maybe my HHTP endpoint accepts up to 1Mb of data per request, but yours accepts only 256Kb.)

lalexgap: Thinking about this further I think I agree with you. It seems simpler for the MessageService to implement batching if needed.

lalexgap:
I think this is in good enough shape to merge in at the moment. Any further work can be tackled under a different PR.

PR #3428, opened 2021-03-31
Fix xstate yarn start

kerzhner:
With this PR:
- `yarn start` works when executed in `packages/xstate-wallet`
- `yarn test` continues to work when executed in `packages/interop-tests`

Fixes https://github.com/statechannels/statechannels/issues/3408

### DIFF
--- a/packages/xstate-wallet/scripts/start.js
+++ b/packages/xstate-wallet/scripts/start.js
@@ -23,22 +23,18 @@ configureEnvVariables();
 const fs = require('fs');
 const webpack = require('webpack');
 const WebpackDevServer = require('webpack-dev-server');
-const clearConsole = require('react-dev-utils/clearConsole');
-const checkRequiredFiles = require('react-dev-utils/checkRequiredFiles');
 const {
   choosePort,
   createCompiler,
   prepareProxy,
   prepareUrls
 } = require('react-dev-utils/WebpackDevServerUtils');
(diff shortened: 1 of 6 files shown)


packages/xstate-wallet/scripts/start.js
```diff
@@ -23,22 +23,18 @@ configureEnvVariables();
...
 } = require('react-dev-utils/WebpackDevServerUtils');
-const openBrowser = require('react-dev-utils/openBrowser');
 const paths = require('../config/paths');
 const configFactory = require('../config/webpack.config');
 const createDevServerConfig = require('../config/webpackDevServer.config');
 
-const config = require('../config/webpackDevServer.config');
 const {getNetworkName, setupGanache} = require('@statechannels/devtools');
-const {deploy} = require('../deployment/deploy');
+const {deploy} = require('../lib/deployment/deploy');
```
andrewgordstewart, line 37: Why do we want to use the compiled deploy script?

andrewgordstewart (approved):

packages/xstate-wallet/scripts/start.js
kerzhner: We currently [use `node` to run the start script](https://github.com/statechannels/statechannels/blob/master/packages/xstate-wallet/package.json#L165). We could probably switch to ts-node.

PR #3432, opened 2021-04-01
Ninja nitro

geoknee:
See RFC9


packages/nitro-protocol/contracts/ninja-nitro/AdjudicatorFactory.sol
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.7.4;
+pragma experimental ABIEncoderV2;
+
+import '@openzeppelin/contracts/utils/Create2.sol';
+import './SingleChannelAdjudicator.sol';
+
+contract AdjudicatorFactory {
```
geoknee, line 8: It might be nice to use the openzeppelin [`clones` library f](https://docs.openzeppelin.com/contracts/3.x/api/proxy#Clones)or this.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,94 @@
...
+- if the destination is not a padded Ethereum address, compute the create2 address of that channel and use that
+
+> A future optimization might be to do away with the 32 byte destination concept altogether. Destinations would then always be Ethereum addresses, and trying to allocate funds to a 32 byte destination would raise an exception.
+
+## Gas tradeoffs
+
+| Operation                                                      | Vanilla | Ninja |
+| -------------------------------------------------------------- | ------- | ----- |
+| infastructure deployment                                       | 4.3M    | 5.1M  |
+| directly-funded unidirectional payment channel (happy path)    | 212K    | 167K  |
```
andrewgordstewart, line 88: This row needs quite a bit of context to make sense of this:
- how many assets? what asset types?
- if this channel has an ERC20 asset, does Vanilla include the `approve` transaction in the 212k? Should it?
- for a fair comparison, should we take into account the gas savings in Vanilla from doing https://github.com/statechannels/statechannels/pull/3394?

packages/nitro-protocol/contracts/ninja-nitro/SingleChannelAdjudicator.sol
```diff
@@ -0,0 +1,476 @@
...
+                outcome[i] = Outcome.OutcomeItem(
+                    outcome[i].assetHolderAddress,
+                    abi.encode(assetOutcome)
+                );
+            } else {
+                revert('AssetOutcome not an allocation');
+            }
+        }
+
+        statusOf[fromChannelId] = _generateStatus(
```
andrewgordstewart, line 272: This unconditionally writes to storage, even when the resulting outcome is empty. At what point does the adjudicator self destruct?

packages/nitro-protocol/contracts/ninja-nitro/SingleChannelAdjudicator.sol
```diff
@@ -0,0 +1,476 @@
...
+     * @dev Verifies a conclusion proof, pays out all assets and selfdestructs
+     * @param largestTurnNum The largest turn number of the submitted states; will overwrite the stored value of `turnNumRecord`.
+     * @param fixedPart Data describing properties of the state channel that do not change with state updates.
+     * @param appPartHash The keccak256 of the abi.encode of `(challengeDuration, appDefinition, appData)`. Applies to all states in the finalization proof.
+     * @param outcomeBytes abi.encode of an array of Outcome.OutcomeItem structs.
+     * @param numStates The number of states in the finalization proof.
+     * @param whoSignedWhat An array denoting which participant has signed which state: `participant[i]` signed the state with index `whoSignedWhat[i]`.
+     * @param sigs An array of signatures that support the state with the `largestTurnNum`.
+     */
+    function concludeAndTransferAll(
```
andrewgordstewart, line 40: It seems to me like this should be `concludeAndTransfer`, and accept an array of indices for each asset outcome. (An empty indices object can indicate a `TransferAll` operation.)

This simplifies the exit algorithm:
1. Call `concludeAndTransfer`, transferring out as many outcome items as they're willing.
2. While the adjudicator exists, call `transfer` to iteratively transfer out remaining funds.

Currently,
- one needs to differentiate between two cases: "I am transferring all" vs "I am not transferring all"
- in the second case, one needs an extra transaction to conclude the channel, wasting a 21k base fee

packages/nitro-protocol/contracts/ninja-nitro/SingleChannelAdjudicator.sol
geoknee: The self destruction optimisation is  currently used only in the happy path, whereas the public function `transfer` is part of the unhappy path. The next round of improvements to ninja nitro should extend it to this function, too.

In vanilla nitro we track a `safeToDelete` boolean to cover this kind of thing. Here we could have a `safeToSelfDestruct` flag returned from one of the internal functions. 

Overall it would be nice if the adjudicator _always_ knew when it could self destruct, and did so eagerly if safe.

packages/nitro-protocol/contracts/ninja-nitro/SingleChannelAdjudicator.sol
geoknee: > It seems to me like this should be concludeAndTransfer, and accept an array of indices for each asset outcome. (An empty indices object can indicate a TransferAll operation.)

Agreed. 

> This simplifies the exit algorithm:

> Call concludeAndTransfer, transferring out as many outcome items as they're willing.
While the adjudicator exists, call transfer to iteratively transfer out remaining funds.

I'm not sure I follow this?

andrewgordstewart: Currently, the wallet needs to decide between:
- calling `concludePushOutcomeAndTransferAll`
- 1. calling `conclude`; 2. calling `pushOutcome`; 3. calling `transfer` iteratively until the channel is fully de-funded

With the proposed change, the wallet does not need to decide between these two strategies. The wallet can `concludeAndTransfer`, passing indices for which outcome items should be tackled. If the channel is not fully-defunded, the wallet continues to call `transfer` with more indices.

packages/nitro-protocol/contracts/ninja-nitro/SingleChannelAdjudicator.sol
andrewgordstewart: Is it easy to decide when it should self destruct? If either
- the outcome is empty
- the holdings is 0 for every asset

does the adjudicator serve a further purpose?

geoknee: It is easy to decide: it should self destruct only if the outcome is empty: which is to say if there is nothing stored in the adjudicator. It doesn't matter if there are some holdings, since these are not stored in the adjudicator. The adjudicator can always be redeployed (with blank storage).

packages/nitro-protocol/contracts/ninja-nitro/SingleChannelAdjudicator.sol
geoknee: I think what is confusing me is that in both of your strategies, you mention one wallet making calls to fully defund a channel: if that is their aim they could use the existing `concludeAndTransferAll` method.

What seems a more likely scenario to me: upon getting a conclusion proof, Wallet A wants to close the channel and pay out (only) their part of the outcome. Wallet B then needs to `transfer` out the remaining part of the outcome. We should support this scenario and your suggestion with indices ought to work well.

andrewgordstewart: >  think what is confusing me is that in both of your strategies, you mention one wallet making calls to fully defund a channel

I'm not sure I follow this. Are you referring to this step?

> 1. Call `concludeAndTransfer`, transferring out as many outcome items as they're willing. 

This step indicates calling a function that does not exist, called `concludeAndTransfer`. It does not indicate fully defunding a channel. It indicates transferring whatever outcome items are specified, which may have the effect of fully defunding the channel.

geoknee: Yes that step makes sense; I was referring to this step 

> 2. While the adjudicator exists, call transfer to iteratively transfer out remaining funds.

Is this the same wallet as in 1.?

andrewgordstewart: It can be either wallet! Since each transaction needs to supply the pre-image of the hashed state, the protocol should either decide on a standard sequence of transactions, or the peers should negotiate a sequence of transactions.

geoknee: I think we are on the same page here. I was just (mildly) confused because the way you wrote it, it seemed as if all the transactions came from the same wallet. Of course we would support that, but I think it doesn't quite expose the main benefit?

PR #3433, opened 2021-04-01
chore: resintate linting rule

geoknee:
We currently have a bunch of unused variables floating around in `nitro-protocol`, as well as a TODO to reinstate a linting rule that would warn us about them. 

This PR reinstates the rule, and quiets the resulting warnings by removing the unused variables.


andrewgordstewart (approved):

lalexgap (approved):
👍

PR #3437, opened 2021-04-02
Implement a minimally useful handling of pure direct funder logic

andrewgordstewart:
# Description
This PR aims to incorporate _some_ core DirectFunder logic into `server-wallet` with as little overhead as possible, in terms of code changes. The goal was to showcase something non-trivial. In this case, I am satisfied with the pure DirectFunder protocol getting to the point where it should trigger a deposit. The goal was _not_ to write permanent code.

The approach taken here is:
- when the engine receives a new objective, it instantiates a "rich objective", and stores it in memory
- when a pre-opening signed state is received, an "event" is constructed and sent to the cranker, updating the "rich objective"'s state and generating actions.

## Next steps
1. Handle deposit events
  - Continue to _ignore_ the `deposit` action generated by the pure DirectFunder
  - When the chain service triggers a "holdings updated" event, send it to the pure DirectFunder and trigger the resulting "send state" action
2. Decide whether we're comfortable committing to this approach, and use the `deposit` action to trigger deposits.
3. Persist the "rich objective" in the `Store`
4. Evaluate network optimizations for the wire protocol, sending something like `{objectiveId: string; signedHashes: SignedHash[]}` for accepted objectives.
5. Re-evaluate naming ...

## Changes [Optional] 
(Unfortunately, it was difficult to do this without a broad scope. Out of pragmatism I'm reluctant to split it up, but will consider it if someone feels strongly.)

Here are the supporting changes made towards the above goal:
1. The `DirectFunder` protocol exports an `initialize` function which constructs a "starting state" for the objective.
2.  A (temporary) `pushDirectFunderMessage` function is added to the engine. This
    1. initializes a "rich objective" in memory for new `OpenChannel` objectives with direct funding
    2. constructs an "event" from any signed states and calls `openChannelCranker`
    3. triggers the resulting actions (so far, only sending states)
3. the wallet's signing address is cached in memory. (it is used repeatedly in 2.1)
4. `createChannels` creates channels sequentially, allowing (3) to be effective. (without doing this, there is some unexplained resource contention as the engine makes `NUM_CHANNELS` attempts to fetch the signing address from the store, causing one test to take 20s to complete. with sequential `createChannels` and a cached signing address, this test takes 2s to complete.)

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/__test-with-peers__/sync-objectives.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/sync-objectives.test.ts
@@ -68,7 +68,11 @@ test('handles the objective being synced even if no message is lost', async () =
       recipient: 'alice',
       sender: 'bob',
       data: {
-        signedStates: [expect.objectContaining({turnNum: 0})],
+        signedStates: [
+          expect.objectContaining({turnNum: 0}),
+          // TODO (DirectFunder): Cranking the pure direct funder puts a second message in the outbox...
+          expect.objectContaining({turnNum: 0}),
+        ],
       },
     },
   });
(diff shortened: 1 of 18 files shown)


kerzhner (approved):
This looks reasonable to me.

andrewgordstewart:
I'm not totally happy with the code in this PR, but I'm merging as is to prevent a git nightmare.

PR #3438, opened 2021-04-03
Remove duplicate funding logic

andrewgordstewart:
# Description

There were two ways of "updating the channel funding" at the engine level. This removes one, and keeps the `ChainServiceSubscriberInterface` implementation, called `holdingUpdated`.

It is important to test `holdingUpdated`, because that is the function that gets called when the chain service notices that holdings are updated. Using the (removed) `updateChannelFunding` introduced the risk of using untested code.

The `ChainServiceSubscriberInterface` is more flexible as a generic interface. The `ChainService` doesn't need to know that `Engine.holdingUpdated` also accepts an optional `EngineResponse` and returns a `SingleChannelOutput`, but consumers of the `Engine` (mostly tests?) can take advantage of this to remove some indirection.

I intentionally stopped short of removing `updateChannelFundingFor`, as I was unsure of the original intent of this function. (I think that function can be removed entirely?)

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/chain-service/types.ts
+++ b/packages/server-wallet/src/chain-service/types.ts
@@ -40,11 +40,11 @@ export type ChallengeRegisteredArg = {
   challengeStates: SignedState[];
 };
 
-export interface ChainEventSubscriberInterface {
-  holdingUpdated(arg: HoldingUpdatedArg): void;
-  assetOutcomeUpdated(arg: AssetOutcomeUpdatedArg): void;
-  channelFinalized(arg: ChannelFinalizedArg): void;
-  challengeRegistered(arg: ChallengeRegisteredArg): void;
+export interface ChainEventSubscriberInterface<T extends void = void> {
+  holdingUpdated(arg: HoldingUpdatedArg): T;
+  assetOutcomeUpdated(arg: AssetOutcomeUpdatedArg): T;
+  channelFinalized(arg: ChannelFinalizedArg): T;
+  challengeRegistered(arg: ChallengeRegisteredArg): T;
 }
 
 interface ChainEventEmitterInterface {
--- a/packages/server-wallet/src/engine/engine.ts
+++ b/packages/server-wallet/src/engine/engine.ts
@@ -994,7 +994,7 @@ export class SingleThreadedEngine
   }
 
   /**
-   * Update the engine's knowledge about the funding for a channel.
+   * Update the engine's knowledge about the on-chain funding for a channel.
    *
    * @param args - An object specifying the channelId, asset holder address and amount.
    * @returns A promise that resolves to a channel output.
--- a/packages/server-wallet/src/engine/engine.ts
+++ b/packages/server-wallet/src/engine/engine.ts
@@ -13,7 +13,6 @@ import {
   validatePayload,
   Outcome,
   convertToParticipant,
-  BN,
   makeAddress,
   Address as CoreAddress,
   PrivateKey,
@@ -1005,7 +1004,7 @@ export class SingleThreadedEngine
     {channelId, amount, assetHolderAddress}: HoldingUpdatedArg,
     response = EngineResponse.initialize()
   ): Promise<void> {
-    await this.store.updateFunding(channelId, BN.from(amount), assetHolderAddress);
+    await this.store.updateFunding(channelId, amount, assetHolderAddress);
     await this.takeActions([channelId], response);
 
     response.channelUpdatedEvents().forEach(event => this.emit('channelUpdated', event.value));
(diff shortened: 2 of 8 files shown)


kerzhner:
IIRC, `updateFundingForChannels` was used as a crutch for a fake funded channels.

packages/server-wallet/src/chain-service/types.ts
```diff
@@ -40,11 +40,11 @@ export type ChallengeRegisteredArg = {
   challengeStates: SignedState[];
 };
 
-export interface ChainEventSubscriberInterface {
```
kerzhner, line 43: What is the reason for this type change?

packages/server-wallet/src/engine/engine.ts
```diff
@@ -1027,14 +993,25 @@ export class SingleThreadedEngine
     response.succeededObjectives.map(o => this.emit('objectiveSucceeded', o));
   }
 
+  /**
+   * Update the engine's knowledge about the funding for a channel.
```
kerzhner, line 997: This method applies to directly funded channels. Do we want to clarify that in the comment?

packages/server-wallet/src/engine/engine.ts
```diff
@@ -1027,14 +993,25 @@ export class SingleThreadedEngine
...
+  async holdingUpdated(
+    {channelId, amount, assetHolderAddress}: HoldingUpdatedArg,
+    response = EngineResponse.initialize()
+  ): Promise<SingleChannelOutput> {
+    await this.store.updateFunding(channelId, amount, assetHolderAddress);
     await this.takeActions([channelId], response);
 
     response.channelUpdatedEvents().forEach(event => this.emit('channelUpdated', event.value));
+
+    // holdingUpdated may be called by updateChannelsForFunding, which therefore includes
```
kerzhner, line 1012: I would be in favor of removing `updateChannelsForFunding` to reduce cognitive overhead of interpreting the wallet engine api.

packages/server-wallet/src/engine/types.ts
```diff
@@ -69,7 +63,8 @@ export interface EngineInterface {
 
   challenge(channelId: string): Promise<SingleChannelOutput>;
 
-  updateFundingForChannels(args: UpdateChannelFundingParams[]): Promise<MultipleChannelOutput>;
+  // TODO: Should this live on a TestEngineInterface?
```
kerzhner, line 66: Yes, or maybe it can be removed?

kerzhner (approved):

packages/server-wallet/src/engine/engine.ts
andrewgordstewart: I agree, but was reluctant to make that change in this PR!

packages/server-wallet/src/chain-service/types.ts
andrewgordstewart: Ah, I think I lost a commit in a rebase which explained this.

According to the current testing practices, it is useful eg. for `Engine.holdingUpdated` to _return a value_. For instance, the returned value is used [here](https://github.com/statechannels/statechannels/pull/3438/commits/456597cd73d6218e3bffa2cdd497cbaae3881228#diff-e9d644ce6622fc89b61b6ccb7a6897aa00664c937eeccbb470133b58c42c783dR89-R90). But, the `ChainService`, should not care about the value returned by `holdingUpdated`. Therefore, I changed the interface to be generic:
- the `ChainService` can expect a `ChainEventSubscriberInterface<void>`.
- the `Engine` can implement a more useful version of it, which returns some actions & a `ChannelResult`.

There is some danger due to this change, because the `ChainService` could change its mind and expect a `ChainEventSubscriberInterface<SingleChannelOutput>` and start relying on that.

kerzhner: Is the generic working as expected? `assetOutcomeUpdated` engine implementation still returns Promise<void>.

packages/server-wallet/src/engine/engine.ts
andrewgordstewart: 82b36e9524113bbc7094ec46da5477a34065e71f

As the implementation is agnostic to the `fundingStrategy` of the channel, I opted to avoid saying that it applies to directly funded channels.

(`holdingUpdated` may be called for a virtually-funded channel, for instance)

packages/server-wallet/src/engine/types.ts
andrewgordstewart: #3443

packages/server-wallet/src/engine/engine.ts
andrewgordstewart: #3443

packages/server-wallet/src/chain-service/types.ts
andrewgordstewart: I'm not sure it's working as _expected_, but it's working ...

Screen Shot 2021-04-05 at 9 56 17 AM

A quick check shoes that `Promise<Something>` might not extend `void`, so I don't know how a Promise is an allowed type for `T`.

PR #3448, opened 2021-04-06
Fix xstate-wallet storybook

kerzhner:
`yarn storybook` runs in `xstate` wallet package and displays all stories.

There is still an issue where some of the stories seem to display content that belongs to other stories.


geoknee:
I'm not sure how painful it was to get storybook working again this time: but I wanted to point out the [`--smoke-test` CLI arg](https://storybook.js.org/docs/react/api/cli-options) for storybook can be used to make it harder for us to break it.  We could slot the smoke test into our CI pipeline somewhere and get a failure if it won't start for some reason. 

I suppose it will extend the length of the pipeline (not sure by how much), but it might be worth it?

kerzhner:
Thanks for the suggestion @geoknee! I am the only person who is working on the xstate wallet at the moment, so we can probably delay implementing a storybook CI test until a bit later.

geoknee (approved):

PR #3449, opened 2021-04-06
chore: remove obsolete benchmarks

lalexgap:
Removes the obsolete benchmarks.

We don't compile the benchmarks in our build process so the benchmarks currently do not build.

This PR just removes them to help keep things cleaner.


andrewgordstewart (approved):
As long as it's easy to find this commit in `master`'s history, let's delete them!

(There's no point rewriting benchmarks when we can copy/paste and fix some compile issues.)

PR #3457, opened 2021-04-08
Add Jumpstart objectives method

lalexgap:
# Description
Add a new jumpstart objectives method that can be used to make progress on objectives.

The main use case for this is continuing progress after a wallet crashes or restarts.

The `jumpStart` works by querying for all `approved` objectives and calling `ensureObjective` on them.

## How Has This Been Tested? 
A new suite of tests have been added for the jumpstart method. This should also be tested by a more extensive E2E test.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/jest/custom-matchers.ts
+++ b/packages/server-wallet/jest/custom-matchers.ts
@@ -1,4 +1,5 @@
 import { AllocationItem, areAllocationItemsEqual } from "@statechannels/wallet-core";
+import { ObjectiveDoneResult, ObjectiveResult } from "../src/wallet";
 
 expect.extend({
     toContainAllocationItem(received: AllocationItem[], argument: AllocationItem) {
@@ -51,3 +52,43 @@ expect.extend({
     }
   },
 });
+
+expect.extend({
+  toBeObjectiveDoneType: async (
+    received: ObjectiveResult[],
+    expected: ObjectiveDoneResult['type']
+  ) => {
+    const pass = received.length > 0 && received.every(async o => (await o.done).type === expected);
+    if (pass) {
+      const matchingObjectives = received.filter(async o => (await o.done).type === expected);
+      return {
+        pass,
+        message: () =>
+          `expected every objective to not be ${expected}. Failing objectives: ${printObjectives(
+            matchingObjectives
+          )}`,
+      };
+    } else {
+      if (received.length === 0) {
+        return {message: () => 'expected at least one ObjectiveResult', pass};
+      } else {
+        const notMatchingObjectives = received.filter(async o => (await o.done).type === expected);
+        return {
+          pass,
+          message: () =>
+            `expected every objective to be ${expected}. Failing objectives: ${printObjectives(
+              notMatchingObjectives
+            )}`,
+        };
+      }
+    }
+  },
+});
+
+async function printObjectives(results: ObjectiveResult[]): Promise<string> {
+  return JSON.stringify(
+    results.map(async r => ({type: (await r.done).type, objectiveId: r.objectiveId})),
+    null,
+    1
+  );
+}
--- a/packages/server-wallet/src/__test-with-peers__/wallet/jumpstart.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/wallet/jumpstart.test.ts
@@ -5,7 +5,7 @@ import {
   peerEngines,
   peersTeardown,
 } from '../../../jest/with-peers-setup-teardown';
-import {WalletObjective} from '../../models/objective';
+import {ObjectiveModel, WalletObjective} from '../../models/objective';
 import {getWithPeersCreateChannelsArgs} from '../utils';
 
 beforeAll(getPeersSetup());
@@ -13,6 +13,27 @@ afterAll(peersTeardown);
 
 jest.setTimeout(60_000);
 describe('jumpstartObjectives', () => {
+  it('returns an empty array when there are no objectives', async () => {
+    const wallet = await Wallet.create(peerEngines.a, messageService, {numberOfAttempts: 1});
+
+    const jumpstartResponse = await wallet.jumpStartObjectives();
+    const jumpstartResults = await Promise.all(jumpstartResponse.map(r => r.done));
+
+    expect(jumpstartResults).toHaveLength(0);
+  });
+
+  it('ignores completed objectives', async () => {
+    const wallet = await Wallet.create(peerEngines.a, messageService, {numberOfAttempts: 1});
+
+    const createResponse = await wallet.createChannels([getWithPeersCreateChannelsArgs()]);
+
+    await ObjectiveModel.succeed(createResponse[0].objectiveId, peerEngines.a.knex);
+    const jumpstartResponse = await wallet.jumpStartObjectives();
+    const jumpstartResults = await Promise.all(jumpstartResponse.map(r => r.done));
+
+    expect(jumpstartResults).toHaveLength(0);
+  });
+
   it('can jumpstart objectives successfully', async () => {
     const wallet = await Wallet.create(peerEngines.a, messageService, {numberOfAttempts: 1});
 
--- a/packages/server-wallet/src/__test-with-peers__/wallet/jumpstart.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/wallet/jumpstart.test.ts
@@ -1,5 +1,6 @@
 import {Wallet} from '../../wallet';
 import {
+  crashAndRestart,
   getPeersSetup,
   messageService,
   peerEngines,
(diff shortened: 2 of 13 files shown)


packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -52,24 +52,50 @@ export class Wallet {
...
+    // Instead of getting messages per objective we just get them all at once
+    // This will prevent us from querying the database for each objective
+    const syncMessages = getMessages(await this._engine.syncObjectives(objectiveIds));
+    return Promise.all(
+      objectives.map(async o => {
+        return {
+          objectiveId: o.objectiveId,
+          currentStatus: o.status,
+          channelId: o.data.targetChannelId,
+          done: this.ensureObjective(o, syncMessages),
```
andrewgordstewart, line 89: `syncMessages` contains data for every approved objective. It's passed to `this.ensureObjective` once per active objective. What effect does this have?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -52,24 +52,50 @@ export class Wallet {
...
       })
     );
   }
 
+  /**
+   * Finds any approved objectives and attempts to make progress on them.
+   * This is useful for restarting progress after a restart or crash.
+   * @returns A collection of ObjectiveResults. There will be an ObjectiveResult for each approved objective found.
+   */
+  public async jumpStartObjectives(): Promise<ObjectiveResult[]> {
```
andrewgordstewart, line 77: Is it safe to call `jumpStartObjectives` during "normal operation"? Wouldn't it conflict with other ongoing operations?

packages/server-wallet/src/wallet/wallet.ts
lalexgap: This means that we're sending out the `syncMessages` payload for every objective (which means a lot of redundant network requests).

This might be a case of premature optimization as I was trying to avoid calling `syncObjective` per objectives to prevent excess db queries.

Longer-term I think the `engine` should return messages indexed by `objectiveId` so it is easy to only send the correct messages. (#3461)

packages/server-wallet/src/wallet/wallet.ts
lalexgap: That's a good question! I don't think there should be a conflict as jumpstart is just essentially triggering messages to be resent. I don't think that would cause a conflict?

Perhaps worth writing a test?

packages/server-wallet/src/wallet/wallet.ts
andrewgordstewart: Right. So, in the current system we need to choose between something like this list of options:
1. excessive # of DB calls
2. excessive network requests
3. modifying `syncObjectives` (#3461)
4. hacking the returned value of `syncObjectives`

If we're explicitly choosing the drawback (2), I think it's important to highlight this in the PR description.

packages/server-wallet/src/wallet/wallet.ts
andrewgordstewart: What kind of conflict could it create?

lalexgap: I've added a test in 55a6c88a8842778e494179e13ef9b4fafeec8c3a that does a simple test of this behavior. I think this is a good question to keep in mind when designing an end-to-end test. 

If we do encounter an error with this I think a pretty simple solution is to only call `jumpstart` internally when creating the wallet.

andrewgordstewart: > If we do encounter an error with this I think a pretty simple solution is to only call jumpstart internally when creating the wallet.

Good idea! We would still need to return a handle on the ongoing objectives to the wallet user somehow.

packages/server-wallet/jest/custom-matchers.ts
```diff
@@ -50,4 +51,44 @@ expect.extend({
       };
     }
   },
-});
\ No newline at end of file
+});
+
+expect.extend({
```
andrewgordstewart, line 56: 👏

packages/server-wallet/jest/custom-matchers.ts
```diff
@@ -50,4 +51,44 @@ expect.extend({
...
+  toBeObjectiveDoneType: async (
+    received: ObjectiveResult[],
+    expected: ObjectiveDoneResult['type']
+  ) => {
```
andrewgordstewart, line 60: Is there a different API that's more useful to the developer?

For instance, I might pass in an array `received: ObjectiveResult[]` and `expected: DeepPartial<ObjectiveDoneResult>[]`, enabling a developer to do

`await expect(wallet.createChannels(SNIP)).toBeObjectiveDoneType([{type: 'Succeed'}, {type: 'Fail', reason: 'Ledger is out of funds'}])` to test the behaviour of `createChannels`.

packages/server-wallet/src/__test-with-peers__/wallet/jumpstart.test.ts
```diff
@@ -0,0 +1,95 @@
...
+jest.setTimeout(60_000);
+
+describe('jumpstartObjectives', () => {
+  it('returns an empty array when there are no objectives', async () => {
+    const wallet = await Wallet.create(peerEngines.a, messageService, {numberOfAttempts: 1});
+
+    const jumpstartResponse = await wallet.jumpStartObjectives();
+    const jumpstartResults = await Promise.all(jumpstartResponse.map(r => r.done));
+
+    expect(jumpstartResults).toHaveLength(0);
```
andrewgordstewart, line 23: ~Why can't you use the `toBeObjectiveDoneType` matcher here?~

N/M, I understand now. Maybe you can improve the API of `toBeObjectiveDoneType` and use it here?

andrewgordstewart, line 23: Wait, isn't this equivalent to
```suggestion
    expect(await wallet.jumpStartObjectives()).toHaveLength(0)
```

packages/server-wallet/src/__test-with-peers__/wallet/jumpstart.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  it('ignores completed objectives', async () => {
+    const wallet = await Wallet.create(peerEngines.a, messageService, {numberOfAttempts: 1});
+
+    const createResponse = await wallet.createChannels([getWithPeersCreateChannelsArgs()]);
+
+    await ObjectiveModel.succeed(createResponse[0].objectiveId, peerEngines.a.knex);
+    const jumpstartResponse = await wallet.jumpStartObjectives();
+    const jumpstartResults = await Promise.all(jumpstartResponse.map(r => r.done));
+
+    expect(jumpstartResults).toHaveLength(0);
```
andrewgordstewart, line 35:
```suggestion
    expect(await wallet.jumpStartObjectives()).toHaveLength(0);
```

packages/server-wallet/src/__test-with-peers__/wallet/jumpstart.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    // await crashAndRestart('A');
+
+    wallet = await Wallet.create(peerEngines.a, messageService, {numberOfAttempts: 1});
+
+    const jumpstartResponse = await wallet.jumpStartObjectives();
+
+    await expect(jumpstartResponse).toBeObjectiveDoneType('Success');
+  });
+
+  it('can jumpstart multiple times', async () => {
```
andrewgordstewart, line 65: I really like this test, nice work.

packages/server-wallet/src/wallet/wallet.ts
andrewgordstewart: Actually, it's probably worth it to highlight this drawback in the code itself. What do you think?

andrewgordstewart (approved):

lalexgap:
> Actually, it's probably worth it to highlight this drawback in the code itself. What do you think?

I agree! Done in 100d2fd

packages/server-wallet/jest/custom-matchers.ts
lalexgap: That's a good point! 

I think my preference would be to wait until we need something like that and then modify/extend the custom matcher.

packages/server-wallet/src/wallet/wallet.ts
lalexgap: I agree! Done in 7758eb6

packages/server-wallet/jest/custom-matchers.ts
andrewgordstewart: How much work is it to modify it at this stage?

lalexgap: I don't think it would be a lot of work!

I do think it will make these tests slightly less succinct:
```
 await expect(response).toBeObjectiveDoneType(Array(10).fill({type: 'Success'}))
// vs
  await expect(response).toBeObjectiveDoneType('Success');
```

andrewgordstewart: Ahhh, the current API checks that they all resolve to success, good point.

PR #3467, opened 2021-04-12
[ninja-nitro] record infra deployment costs

geoknee:
```
Deploying AdjudicatorFactory... (cost estimated to be 688895)


Deploying MasterCopy... (cost estimated to be 3417269)
```


andrewgordstewart (approved):
Those costs seem somewhat bearable, even!

PR #3477, opened 2021-04-12
Add approved field to OpenChannelObjective

kerzhner:
When a wallet receives an open channel objective, the wallet should:
1. Store the objective.
2. Notify the application of the objective.
3. Not take any actions before the application approves the objective.

It is likely that the `approved` field belongs on all objective types. But, for now, we are only considering the open channel objective.

This PR adds objective approval to `OpenChannelObjective` while leaving xstate-wallet api unchanged.

### DIFF
--- a/packages/wallet-core/src/tests/direct-funder.test.ts
+++ b/packages/wallet-core/src/tests/direct-funder.test.ts
@@ -171,7 +171,7 @@ describe('cranking', () => {
         ]
       }
     });
-    const zeroOutcome = {...initialize(signStateHelper(zeroOutcomeState, 'A'), 1), approved: true};
+    const zeroOutcome = initialize(signStateHelper(zeroOutcomeState, 'A'), 1, true);
 
     const aliceSignedPre = objectiveFixture({
       status: WaitingFor.theirPreFundSetup,
--- a/packages/wallet-core/src/tests/direct-funder.test.ts
+++ b/packages/wallet-core/src/tests/direct-funder.test.ts
@@ -259,7 +259,8 @@ describe('cranking', () => {
       
 
       [ msg = 'Receiving a preFundSetup state',
-             approved,          sendState(richPreFS.stateSignedBy('B')), {preFundSetup: richPreFS.signedBy('A', 'B')}, [{type: 'sendStates'}, {type: 'deposit'}] ],
+             notApproved,    sendState(richPreFS.stateSignedBy('B')), {preFundSetup: richPreFS.signedBy('B')},      [] ],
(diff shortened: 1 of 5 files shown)


packages/wallet-core/src/tests/direct-funder.test.ts
```diff
@@ -167,7 +171,7 @@ describe('cranking', () => {
         ]
       }
     });
-    const zeroOutcome = initialize(signStateHelper(zeroOutcomeState, 'A'), 1);
+    const zeroOutcome = {...initialize(signStateHelper(zeroOutcomeState, 'A'), 1), approved: true};
```
andrewgordstewart, line 174: 💡 should `initialize` accept an `approved` flag?

kerzhner: I think that's a good idea 427adccc0

packages/wallet-core/src/tests/direct-funder.test.ts
```diff
@@ -388,17 +394,19 @@ test('pure objective cranker start to finish', () => {
...
+  4. Bob approves objective, signs preFS, trigers preFS action 2
+  5. Alice receives preFS event 2, triggers deposit action 1
+  6. Alice receives deposit event 1, does nothing
+  7. Bob receives deposit event 1, triggers deposit action 1
+  8. Bob receives deposit action 2 (UNFINALIZED)
+  9. Alice receives deposit action 2 (UNFINALIZED)
+  10. Bob receives deposit action 2 (FINALIZED), triggers postFS action 1
+  11. Alice receives deposit action 2 (FINALIZED), triggers postFS action 2
+  12. Alice receives postFS event 1, is finished
+  13. Bob receives postFS event 1, is finished
```
andrewgordstewart, line 409: This looks like a maintenance headache 😆

packages/xstate-wallet/src/channel-wallet.ts
```diff
@@ -280,6 +286,14 @@ export class ChannelWallet {
     }
   }
 
+  approveRichObjective(channelId: string) {
```
andrewgordstewart, line 289: Have you thought up a better term than "Rich objective"?

andrewgordstewart (approved):

packages/wallet-core/src/tests/direct-funder.test.ts
kerzhner: It is, but it is useful for now.

packages/xstate-wallet/src/channel-wallet.ts
kerzhner: I haven't put much thought into it. Rich objective is a bit of a cop out name though.

PR #3487, opened 2021-04-14
export NinjaNitro artifacts

geoknee:
... in their own namespace. This makes them usable by external projects.


NiloCK (approved):
Looks good.

As a random example, these exports are consumed by other projects as in [here](https://github.com/statechannels/statechannels/blob/4d50371b2d6c31bb27505dbe85008416eedcba25/packages/xstate-wallet/deployment/deploy.ts#L15)?

PR #3488, opened 2021-04-14
[ninja-nitro] add a readme.md

geoknee:
Work in progress

### DIFF
--- a/packages/nitro-protocol/contracts/ninja-nitro/readme.md
+++ b/packages/nitro-protocol/contracts/ninja-nitro/readme.md
@@ -1,10 +1,12 @@
 # Ninja Nitro
 
-Ninja-Nitro is an alternative implementation of Nitro protocol, with a different contract architecture to the main implementation (Mono-Nitro) in this package. The additional files should all have namespaced paths.
+Ninja-Nitro is an alternative implementation of Nitro protocol, with a different contract architecture to the main implementation (henceforth Vanilla-Nitro\*) in this package. The additional files should all have namespaced paths.
+
+\*this is just a temporary name.
 
 ## Overview
 
-### Recap of Mono-Nitro
+### Recap of Vanilla-Nitro
 
 The main implementation has 1 x monolithic `Adjudicator` contract and m x `AssetHolder` contracts (one for each of the m Assets in the state channel network). The `Adjudicator` API is
 
@@ -44,19 +46,19 @@ Note however, that the Proxy contracts each have their own storage. See https://
 
 ### Holdings
 
-The Mono-Nitro system has m x `holdings` mappings, one in each AssetHolder. This mapping is `public`, so has a public getter function.
+The Vanilla-Nitro system has m x `holdings` mappings, one in each AssetHolder. This mapping is `public`, so has a public getter function.
 Ninja-Nitro has no such mappings: but a public `holdings` function remains. This functions simply i) reads the eth balance of the current contract using the Ethereum accounting system or ii) reads the token balance from a given ERC20 Token contract
 
 ### Deposits
 
-In Mono-Nitro, depositing requires calling a method on each `AssetHolder`.
+In Vanilla-Nitro, depositing requires calling a method on each `AssetHolder`.
 In Ninja-Nitro, depositing can be achieved simply by sending funds or tokens to the address of the `Proxy`, and this can even be done before the `Proxy` is created/deployed. It can also be done after the `Proxy` is destroyed -- self destructing the adjudicator when the channel is finished is one way to save some gas.
 
 ### Adjudicator status
 
-Ninja-Nitro was implemented by modifying a copy of Mono-Nitro. This means that certain quirks remain in the implementation which would probably not exist had it been written from scratch. These quirks may or may not be removed in future.
+Ninja-Nitro was implemented by modifying a copy of Vanilla-Nitro. This means that certain quirks remain in the implementation which would probably not exist had it been written from scratch. These quirks may or may not be removed in future.
 
-One such quirk is as follows. In Mono-Nitro, the hash of the status of a channel (which includes the full outcome, the address of any challenger, the time when the channel will finalize, etc) is stored in a mapping, keyed by the `channelId`.
+One such quirk is as follows. In Vanilla-Nitro, the hash of the status of a channel (which includes the full outcome, the address of any challenger, the time when the channel will finalize, etc) is stored in a mapping, keyed by the `channelId`.
 
 In Ninja-Nitro, exactly the same is true: only each proxy has its own mapping. The reason that this is quirky is that each of these mappings is only ever going to have a single entry, because each proxy is in a 1-1 relation to a `channelId`. The reason that this quirk is not actually very troublesome are:
 
@@ -67,11 +69,11 @@ In Ninja-Nitro, exactly the same is true: only each proxy has its own mapping. T
 
 ### External destinations
 
-In Mono-Nitro, funds for channels are tracked _internally_, while funds for any Ethereum address (e.g. an externally owned account) are typically paid out of the system when defunding a channel. Padding an Ethereum address with enough zeros results in a 32 byte number which we call an external destination. Channel ids are known as internal destinations.
+In Vanilla-Nitro, funds for channels are tracked _internally_, while funds for any Ethereum address (e.g. an externally owned account) are typically paid out of the system when defunding a channel. Padding an Ethereum address with enough zeros results in a 32 byte number which we call an external destination. Channel ids are known as internal destinations.
 
 In Ninja-Nitro, funds are always tracked externally. This is possible since (up to hash collisions) each channel now has a unique Ethereum address. Funds are therefore always paid _out_ of a `Proxy` contract after a `transfer` or `claim`. There is therefore no such thing as an internal destination in Ninja-Nitro. Everything is external.
 
-To maintain compatibility with the off-chain code we already have for Mono-Nitro, the Ninja-Nitro implementation converts destinations to addresses in the following way:
+To maintain compatibility with the off-chain code we already have for Vanilla-Nitro, the Ninja-Nitro implementation converts destinations to addresses in the following way:
 
 - if the destination is a padded Ethereum address, slice off the Ethereum address and use that
 - if the destination is not a padded Ethereum address, compute the create2 address of that channel and use that
@@ -80,13 +82,13 @@ To maintain compatibility with the off-chain code we already have for Mono-Nitro
 
 ## Gas tradeoffs
 
-| Operation                                                      | Mono | Ninja |
-| -------------------------------------------------------------- | ---- | ----- |
-| infastructure deployment                                       | 4.3M | 5.1M  |
-| directly-funded unidirectional payment channel (happy path)    | 212K | 167K  |
-| directly-funded unidirectional payment channel (unhappy path)  | ?    | ?     |
-| ledger-funded unidirectional payment channel (unhappy path)    | ?    | ?     |
-| virtually-funded unidirectional payment channel (unhappy path) | ?    | ?     |
+| Operation                                                      | Vanilla | Ninja |
+| -------------------------------------------------------------- | ------- | ----- |
+| infastructure deployment                                       | 4.3M    | 5.1M  |
+| directly-funded unidirectional payment channel (happy path)    | 212K    | 167K  |
+| directly-funded unidirectional payment channel (unhappy path)  | ?       | ?     |
+| ledger-funded unidirectional payment channel (unhappy path)    | ?       | ?     |
+| virtually-funded unidirectional payment channel (unhappy path) | ?       | ?     |
 
 Unidirectional payment channel: only one deposit and one payout
 The unhappy path: all channels in the funding tree are finalized on chain via `challenge` and timeout, and the funds are then transferred and claimed until they are all liquidated.
--- a/packages/nitro-protocol/contracts/ninja-nitro/readme.md
+++ b/packages/nitro-protocol/contracts/ninja-nitro/readme.md
@@ -71,12 +71,10 @@ In Mono-Nitro, funds for channels are tracked _internally_, while funds for any 
 
 In Ninja-Nitro, funds are always tracked externally. This is possible since (up to hash collisions) each channel now has a unique Ethereum address. Funds are therefore always paid _out_ of a `Proxy` contract after a `transfer` or `claim`. There is therefore no such thing as an internal destination in Ninja-Nitro. Everything is external.
 
-To maintain compatibility with the off-chain code we already have for Mono-Nitro, the Ninja-Nitro implementation should convert destinations to addresses in the following way:
+To maintain compatibility with the off-chain code we already have for Mono-Nitro, the Ninja-Nitro implementation converts destinations to addresses in the following way:
 
 - if the destination is a padded Ethereum address, slice off the Ethereum address and use that
 - if the destination is not a padded Ethereum address, compute the create2 address of that channel and use that
-
-> TODO it does not currently do this and our test suite is not adequate enough to tell us!
 
 > A future optimization might be to do away with the 32 byte destination concept altogether. Destinations would then always be Ethereum addresses, and trying to allocate funds to a 32 byte destination would raise an exception.
 
--- a/packages/nitro-protocol/contracts/ninja-nitro/readme.md
+++ b/packages/nitro-protocol/contracts/ninja-nitro/readme.md
@@ -60,7 +60,7 @@ One such quirk is as follows. In Mono-Nitro, the hash of the status of a channel
 
 In Ninja-Nitro, exactly the same is true: only each proxy has its own mapping. The reason that this is quirky is that each of these mappings is only ever going to have a single entry, because each proxy is in a 1-1 relation to a `channelId`. The reason that this quirk is not actually very troublesome are:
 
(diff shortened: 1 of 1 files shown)


packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+In Mono-Nitro, funds for channels are tracked _internally_, while funds for any Ethereum address (e.g. an externally owned account) are typically paid out of the system when defunding a channel. Padding an Ethereum address with enough zeros results in a 32 byte number which we call an external destination. Channel ids are known as internal destinations.
+
+In Ninja-Nitro, funds are always tracked externally. This is possible since (up to hash collisions) each channel now has a unique Ethereum address. Funds are therefore always paid _out_ of a `Proxy` contract after a `transfer` or `claim`. There is therefore no such thing as an internal destination in Ninja-Nitro. Everything is external.
+
+To maintain compatibility with the off-chain code we already have for Mono-Nitro, the Ninja-Nitro implementation should convert destinations to addresses in the following way:
+
+- if the destination is a padded Ethereum address, slice of the Ethereum address and use that
+- if the destination is not a padded Ethereum address, compute the create2 address of that channel and use that
+
+> TODO it does not currently do this and our test suite is not adequate enough to tell us!
```
geoknee, line 79: I will create a New Issue for this.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+The essence of the architecture, then, is `pushOutcome`: a finalized outcome must be "pushed" from the `Adjudicator`, resulting in a piece of that outcome being stored in each `AssetHolder`. Nevertheless, `pushOutcome` is an implementation detail and not necessary part of the overall Nitro API.
+
+The `AssetHolder` API is:
+
+- `transfer`
+- `claim`
+- permissioned methods `transferAllAdjudicatorOnly` and `setAssetOutcomeHash` only callable from the `Adjudicator`.
+
+After a `pushOutcome` ssets may be released by targetting each `AssetHolder` and calling `claim` or `transfer` (which _are_ essential parts of the overall Nitro API). This is the unhappy path. The happy path will shortcut `pushOutcome` and use the permissioned methods.
```
NiloCK, line 23:
```suggestion
After a `pushOutcome` assets may be released by targetting each `AssetHolder` and calling `claim` or `transfer` (which _are_ essential parts of the overall Nitro API). This is the unhappy path. The happy path will shortcut `pushOutcome` and use the permissioned methods.
```

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Ninja-Nitro
+
+Ninja-Nitro, by contrast has 1 x `AdjudicatorFactory` contract, 1 x `SingleChannelAdjudicator` master contract, and k x `Proxy` contracts, one for each of the k channels in the network that resolves on chain. The factory and master contracts are "infastructure" and deployed once (most likely by the protocol team).
```
NiloCK, line 27: `most likely by the protocol team`

This refers to team X who are authoring some N-N based SCDapp?

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+One such quirk is as follows. In Mono-Nitro, the hash of the status of a channel (which includes the full outcome, the address of any challenger, the time when the channel will finalize, etc) is stored in a mapping, keyed by the `channelId`.
+
+In Ninja-Nitro, exactly the same is true: only each proxy has its own mapping. The reason that this is quirky is that each of these mappings is only ever going to have a single entry, because each proxy is in a 1-1 relation to a `channelId`. The reason that this quirk is not actually very troublesome are:
+
+- the gas cost are not dramatically higher than storing a single hash without using a mapping (it may in fact cost precisely the same amount). See https://docs.soliditylang.org/en/v0.8.3/types.html?highlight=mapping#mapping-types
+- we are able to reuse `ForceMove` (in the sense of `SingleChannelAdjudicator is ForceMove`) in its entirety without modifying it at all
+
+> A future optimization might be to replace the mapping with a regular, singleton storage variable. This would require a small but breaking change to `ForceMove.sol`.
```
NiloCK, line 66: Earmarked for discussion.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+### External destinations
+
+In Mono-Nitro, funds for channels are tracked _internally_, while funds for any Ethereum address (e.g. an externally owned account) are typically paid out of the system when defunding a channel. Padding an Ethereum address with enough zeros results in a 32 byte number which we call an external destination. Channel ids are known as internal destinations.
+
+In Ninja-Nitro, funds are always tracked externally. This is possible since (up to hash collisions) each channel now has a unique Ethereum address. Funds are therefore always paid _out_ of a `Proxy` contract after a `transfer` or `claim`. There is therefore no such thing as an internal destination in Ninja-Nitro. Everything is external.
+
+To maintain compatibility with the off-chain code we already have for Mono-Nitro, the Ninja-Nitro implementation should convert destinations to addresses in the following way:
+
+- if the destination is a padded Ethereum address, slice of the Ethereum address and use that
```
NiloCK, line 76:
```suggestion
- if the destination is a padded Ethereum address, slice off the Ethereum address and use that
```

NiloCK (approved):
This is very good. I read the overview summaries while jumping around our contract code. I now better understand things, and I expect this readme will be a pretty useful and compact summary for others on the project.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
+# Ninja Nitro
+
+Ninja-Nitro is an alternative implementation of Nitro protocol, with a different contract architecture to the main implementation (Mono-Nitro) in this package. The additional files should all have namespaced paths.
```
andrewgordstewart, line 3: This may seem pedantic, but I would prefer to use `Mono-Nitro` to refer to a monolithic nitro architecture, one with _one_ smart contract. The existing architecture decouples adjudication from funding across multiple contracts.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+- utility methods that batch up this API
+
+Note however, that the Proxy contracts each have their own storage. See https://eips.ethereum.org/EIPS/eip-1167.
+
+## Similarities and differences
+
+### Holdings
+
+The Mono-Nitro system has m x `holdings` mappings, one in each AssetHolder. This mapping is `public`, so has a public getter function.
+Ninja-Nitro has no such mappings: but a public `holdings` function remains. This functions simply reads the balance of the current contract (or another contract) from either i) the Ethereum accounting system or ii) a given Token contract.
```
andrewgordstewart, line 48:
```suggestion
Ninja-Nitro has no such mappings: but a public `holdings` function remains. This functions simply i) reads the eth balance of the current contract using the Ethereum accounting system or ii) reads the token balance from a given ERC20 Token contract
```

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Deposits
+
+In Mono-Nitro, depositing requires calling a method on each `AssetHolder`.
+In Ninja-Nitro, depositing can be achieved simply by sending funds or tokens to the address of the `Proxy`, and this can even be done before the `Proxy` is created/deployed.
```
andrewgordstewart, line 53: Can this be done after the `Proxy` is destroyed?

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Adjudicator status
+
+Ninja-Nitro was implemented by modifying a copy of Mono-Nitro. This means that certain quirks remain in the implementation which would probably not exist had it been written from scratch. These quirks may or may not be removed in future.
+
+One such quirk is as follows. In Mono-Nitro, the hash of the status of a channel (which includes the full outcome, the address of any challenger, the time when the channel will finalize, etc) is stored in a mapping, keyed by the `channelId`.
+
+In Ninja-Nitro, exactly the same is true: only each proxy has its own mapping. The reason that this is quirky is that each of these mappings is only ever going to have a single entry, because each proxy is in a 1-1 relation to a `channelId`. The reason that this quirk is not actually very troublesome are:
+
+- the gas cost are not dramatically higher than storing a single hash without using a mapping (it may in fact cost precisely the same amount). See https://docs.soliditylang.org/en/v0.8.3/types.html?highlight=mapping#mapping-types
```
andrewgordstewart, line 63: This does not feel possible to me. If a mapping contains a `uint` number , then 32 bytes would be required to store the number. If those 32 bytes are the only thing stored, how can the bookkeeping happen?

Are you sure that there's no potential gas savings from switching from a `mapping bytes32 -> bytes32` state variable to a `bytes32` state variable?

https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays suggests a mapping has 32 bytes of overhead in addition to the storage that their items contain. Thus, I would predict that a `mapping bytes32 -> bytes32` would require at least two storage slots to store one `bytes32` key/value pair.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
andrewgordstewart: Does our test suite also not detect this for the current nitro?

packages/nitro-protocol/contracts/ninja-nitro/readme.md
```diff
@@ -0,0 +1,97 @@
...
+In Ninja-Nitro, funds are always tracked externally. This is possible since (up to hash collisions) each channel now has a unique Ethereum address. Funds are therefore always paid _out_ of a `Proxy` contract after a `transfer` or `claim`. There is therefore no such thing as an internal destination in Ninja-Nitro. Everything is external.
+
+To maintain compatibility with the off-chain code we already have for Mono-Nitro, the Ninja-Nitro implementation should convert destinations to addresses in the following way:
+
+- if the destination is a padded Ethereum address, slice of the Ethereum address and use that
+- if the destination is not a padded Ethereum address, compute the create2 address of that channel and use that
+
+> TODO it does not currently do this and our test suite is not adequate enough to tell us!
+
+> A future optimization might be to do away with the 32 byte destination concept altogether. Destinations would then always be Ethereum addresses, and trying to allocate funds to a 32 byte destination would raise an exception.
```
andrewgordstewart, line 81: Destinations used to be addresses. They were changed to `bytes32`. Why?

andrewgordstewart (commented):
I left some feedback here, as I read this readme as a precursor to #3459. Most of the comments are not really related to this PR.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: Our current test is OK because it has cases that allocate to a channel as well as to an external destination. The ninja-nitro test suite flipped everything to being an external destination, and so was  bypassing an important code path. Fixing that over on #3492

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: No, I really mean "us". The team that maintains the state channel network. I'll try and make this clearer.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: Yes I believe so. I can add something to our test suite to show that, and an extra comment here.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: I don't think destinations were ever addresses?

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: 24c10f7

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: 75d5c94

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: > This does not feel possible to me. If a mapping contains a `uint` number , then 32 bytes would be required to store the number. If those 32 bytes are the only thing stored, how can the bookkeeping happen?
> 

Clearly some extra information must be stored for bookkeeping; but the question here is about whether there is more bookkeeping involved in storing something inside a mapping rather than as a simple state variable. In each case the location of the stored data in storage must be recorded. From what I can tell from the docs, determining the location of a mapping key involves computing the hash of that key concatenated with the mapping location itself. So I think there is going to be slightly elevated cost to using a mapping.

> Are you sure that there's no potential gas savings from switching from a `mapping bytes32 -> bytes32` state variable to a `bytes32` state variable?

I suspect there will not be significant savings, but I can't say I am sure there will be no savings at all. By significant I mean it won't be 2 storage slots versus 1 storage slot (20K difference in gas). 

> 
> https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays suggests a mapping has 32 bytes of overhead in addition to the storage that their items contain. Thus, I would predict that a `mapping bytes32 -> bytes32` would require at least two storage slots to store one `bytes32` key/value pair.

I think the extra slot you are referring to, which records the location of the mapping in storage, is reserved but not written to, so doesn't incur gas costs. 

I'll knock together a quick investigation to see if we can check this.

geoknee: Investigation [here](https://github.com/statechannels/statechannels/compare/mapping-investigation?expand=1) implies the costs are

packages/nitro-protocol/contracts/ninja-nitro/readme.md
andrewgordstewart: It's been a while!

https://github.com/statechannels/nitro-protocol/commit/3e095dd32f4da956c39fa2456f5c33d0148b9480#diff-52f9afa26d56174ce46794e68e317f0805d87599d37aa43dbcc24b7508c69abf

packages/nitro-protocol/contracts/ninja-nitro/readme.md
andrewgordstewart: > I think the extra slot you are referring to, which records the location of the mapping in storage, is reserved but not written to, so doesn't incur gas costs.

Ah, thanks for clarifying. I hadn't considered that you could reserve the storage slot without writing to it.

Thanks for investigating this!

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: I don't know why they were addresses in the first place, that was most likely a mistake.

packages/nitro-protocol/contracts/ninja-nitro/readme.md
geoknee: The existing architecture is monolithic in the sense of the total number of adjudicators (one), which is the important distinction with ninja-nitro. 

I agree that there is a potentially interesting architecture with one contract _total_. It's more monolithic, or monolithic on more ways than the existing architecture. 

I'd still like to use Mono-Nitro here, since it is immediately useful and will continue to be so in the short term. I would view the More-Mono-Nitro as being a replacement for the current architecture, rather than an alternative (in the way that Ninja-Nitro is an alternative). So it could still be called Mono-Nitro?

That said, I'd be willing to swap this term out for another one if you have a good suggestion? Or I suppose I could just use "the existing implementation" everywhere...

andrewgordstewart: > I would view the More-Mono-Nitro as being a replacement for the current architecture, rather than an alternative (in the way that Ninja-Nitro is an alternative). So it could still be called Mono-Nitro?

There is a third architecture for L1 channels which has:
1. a single AssetHolder
2. a single adjudicator

Off the top of my head, I might call this a "duplex architecture", "Duo Nitro", or something that emphasizes that adjudication and asset management are decoupled.

I think the difference between this and the current architecture is closer to an implementation detail, as I cannot think of many differences in what on-chain operations they'd enable. In particular, they both enable the following for L1 channels:
- AssetHolders live on L1
- the Adjudicator lives on an L2.
This allows peers in an L1 channel to dispute in L2 for much lower fees.

The "mono mono nitro" might be more suitable for satisfying some (not-yet-understood) hub feature features as well as minimizing the number of storage slots required (we need at least one, and I think we can do that with "mono mono nitro".

geoknee: I'll just use Vanilla Nitro for now 095a3b6 . It doesn't need to be the long term name for this.

Issue #3489, opened 2021-04-14
[interop-test] jsdom testEnvironment is slow


kerzhner:
If the `testEnvironment` is set to jsdom in the `interop-tests` package jest configuration, the testsuite takes seconds to start up. The test suite start times are especially slow if the test suite is started via VS Code debugger.

PR #3490, opened 2021-04-14
[xstate-wallet] Adding channel creation to browser wallet

kerzhner:
The main goal of this PR is to add the `**browser** wallet creates channel + cooperates with server wallet to fund channel` itnerop test and the browser wallet functionality for the test to pass. Before this PR, `**server** wallet creates channel + cooperates with browser wallet to fund channel` was already passing.

Browser wallet channel creation works as follows:
1. Existing `CreateChannel` JSON RPC API is used for channel creation.
2. On that API invocation, the browser wallet creates and stores a rich `OpenChannelObjective`. The objective is **not** approved.
3. In the test, the objective is approved via an API call to the browser wallet. When the wallet is run in the browser, the objective approval will come from the user clicking an approve button.
4. When the objective is approved, the rich objective is converted to a slim objective that is broadcast to other participants.

Some of the changes in this PR were made in order for the interop tests to pass. Hopefully these changes are somewhat minimal and do not add too much of an additional burden to the reviewer:
1. Interop tests are converted from `jsdom` tests to `node` tests. See https://github.com/statechannels/statechannels/issues/3489.
2. Interop tests refuse to exit cleanly due to ongoing ethers activity. I still do not understand the core root of the problem, but enough changes have been made for the jest process to at least exit with error code 0. We still see the following log lines at the end of interop tests
```
interop-tests: (node:1801) UnhandledPromiseRejectionWarning: Error: missing response (requestBody="{\"method\":\"eth_blockNumber\",\"params\":[],\"id\":113,\"jsonrpc\":\"2.0\"}", requestMethod="POST", serverError={"errno":"ECONNREFUSED","code":"ECONNREFUSED","syscall":"connect","address":"127.0.0.1","port":8546}, url="http://localhost:8546", code=SERVER_ERROR, version=web/5.0.7)
interop-tests: (node:1801) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 40)
```

### DIFF
--- a/packages/interop-tests/jest/jest.config.js
+++ b/packages/interop-tests/jest/jest.config.js
@@ -13,7 +13,7 @@ module.exports = {
   },
   globals: {
     'ts-jest': {tsconfig: './tsconfig.json'},
-    window: {localStorage: {}, addEventListener: (_p1, p2) => ({})}
+    window: {localStorage: {}, addEventListener: (_p1, _p2) => ({})}
   },
   globalSetup: '<rootDir>/jest/chain-setup.ts',
   globalTeardown: '<rootDir>/jest/test-teardown.ts'
--- a/packages/interop-tests/jest/jest.config.js
+++ b/packages/interop-tests/jest/jest.config.js
@@ -4,14 +4,17 @@ const root = resolve(__dirname, '../');
 module.exports = {
   rootDir: root,
   testMatch: ['<rootDir>/**/__test__/**/?(*.)test.ts'],
-  testEnvironment: 'jsdom',
+  testEnvironment: 'node',
   transform: {'^.+\\.ts$': 'ts-jest'},
   transformIgnorePatterns: ['[/\\\\]node_modules[/\\\\].+\\.(js|ts)$'],
   moduleFileExtensions: ['ts', 'js', 'json', 'node'],
   moduleNameMapper: {
     '.(scss|svg)$': 'identity-obj-proxy'
   },
-  globals: {'ts-jest': {tsconfig: './tsconfig.json'}},
+  globals: {
+    'ts-jest': {tsconfig: './tsconfig.json'},
+    window: {localStorage: {}, addEventListener: (_p1, p2) => ({})}
+  },
   globalSetup: '<rootDir>/jest/chain-setup.ts',
   globalTeardown: '<rootDir>/jest/test-teardown.ts'
 };
--- a/packages/interop-tests/src/__test__/interop.test.ts
+++ b/packages/interop-tests/src/__test__/interop.test.ts
@@ -172,7 +172,7 @@ it('server wallet creates channel + cooperates with browser wallet to fund chann
 
   await browserWallet.pushMessage(serverMessageToBrowserMessage(output1), 'dummyDomain');
 
-  // TODO: the proper way to do this is to wait for a ChannelProposed or some sort of new objective notification'
+  // TODO: the proper way to do this is to wait for a ChannelProposed or some sort of new objective notification
   await browserWallet.pushMessage(
     {
       jsonrpc: '2.0',
--- a/packages/interop-tests/src/__test__/interop.test.ts
+++ b/packages/interop-tests/src/__test__/interop.test.ts
@@ -8,7 +8,15 @@ import {
 import {ETHERLIME_ACCOUNTS} from '@statechannels/devtools';
 import {ChannelWallet} from '@statechannels/xstate-wallet';
 import {constants, Contract, providers, Wallet} from 'ethers';
-import {BN, formatAmount, makeAddress, makeDestination, Uint256} from '@statechannels/wallet-core';
+import {
+  Address,
+  BN,
+  Destination,
+  formatAmount,
+  makeAddress,
+  makeDestination,
+  Uint256
+} from '@statechannels/wallet-core';
 import {fromEvent} from 'rxjs';
 import {take} from 'rxjs/operators';
 import {ContractArtifacts} from '@statechannels/nitro-protocol';
@@ -52,8 +60,13 @@ const serverConfig = defaultTestConfig({
 let provider: providers.JsonRpcProvider;
 let serverWallet: SingleThreadedEngine;
 let browserWallet: ChannelWallet;
+let serverAddress: Address;
+let serverDestination: Destination;
+let browserAddress: Address;
+let browserDestination: Destination;
+let objectiveSuccededPromise: Promise<void>;
 
-beforeAll(async () => {
+beforeEach(async () => {
   await DBAdmin.truncateDatabase(serverConfig);
 
   provider = new providers.JsonRpcProvider(rpcEndpoint);
@@ -68,9 +81,27 @@ beforeAll(async () => {
   browserWallet = await ChannelWallet.create(
     makeAddress(new Wallet(ETHERLIME_ACCOUNTS[1].privateKey).address)
   );
+
+  serverAddress = await serverWallet.getSigningAddress();
+  serverDestination = makeDestination(serverAddress);
+
+  browserAddress = makeAddress(await browserWallet.getAddress());
+  browserDestination = makeDestination(browserAddress);
+
+  browserWallet.onSendMessage(message => {
+    if (isJsonRpcNotification(message)) {
+      serverWallet.pushMessage((message.params as Message).data);
+    }
+  });
+
+  objectiveSuccededPromise = new Promise<void>(r => {
+    serverWallet.on('objectiveSucceeded', (o: WalletObjective) => {
+      if (o.type === 'OpenChannel' && o.status === 'succeeded') r();
+    });
+  });
 });
 
-afterAll(async () => {
+afterEach(async () => {
   await serverWallet.destroy();
   await browserWallet.destroy();
   await provider.removeAllListeners();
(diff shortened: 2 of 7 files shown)


packages/interop-tests/src/__test__/interop.test.ts
```diff
@@ -148,25 +192,79 @@ it('server + browser wallet interoperability test', async () => {
...
-function generatePushMessage(messageParams: Message): PushMessageRequest {
-  return {
-    jsonrpc: '2.0',
-    id: 111111111,
-    method: 'PushMessage',
-    params: messageParams
-  };
-}
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const postFundA = await fromEvent<SingleChannelOutput>(serverWallet as any, 'channelUpdated')
```
andrewgordstewart, line 263: Can we can make the test less brittle by expecting something about `postFundA`?

packages/interop-tests/src/__test__/interop.test.ts
```diff
@@ -148,25 +192,79 @@ it('server + browser wallet interoperability test', async () => {
...
     .pipe(take(3))
     .toPromise();
 
+  await browserWallet.pushMessage(serverMessageToBrowserMessage(postFundA), 'dummyDomain');
+
+  await objectiveSuccededPromise;
+});
+
+it('browser wallet creates channel + cooperates with server wallet to fund channel', async () => {
+  const createChannelParams: CreateChannelParams = {
```
andrewgordstewart, line 205: Could (should?) we use the same parameters here?

In an ideal world, we could execute the same test with roles reversed. Even better, we could execute the same test for many combinations of actors:
- `[server, server]`
- `[server, browser]`
- `[browser, server]`
- `[browser, browser]`

We are not in that world right now, because the APIs differ, correct?

packages/interop-tests/jest/jest.config.js
```diff
@@ -4,14 +4,17 @@ const root = resolve(__dirname, '../');
 module.exports = {
   rootDir: root,
   testMatch: ['<rootDir>/**/__test__/**/?(*.)test.ts'],
-  testEnvironment: 'jsdom',
+  testEnvironment: 'node',
```
andrewgordstewart, line 7: Good catch!

packages/xstate-wallet/src/channel-wallet.ts
```diff
@@ -308,6 +313,10 @@ export class ChannelWallet {
     return this.store.richObjectives;
   }
 
+  public async destroy(): Promise<void> {
```
andrewgordstewart: I would prefer test-specific functions to be defined on a `TestChannelWallet` class that inherits from `ChannelWallet`.

I think it's fine to avoid that, but IMO we should clearly mark such code as test-specific, so it's easier to notice when we don't want to ship test code in production.

packages/xstate-wallet/src/channel-wallet.ts
```diff
@@ -273,9 +273,11 @@ export class ChannelWallet {
             break;
           case 'deposit':
             if (this.store.depositsSubmitted[channelId]) {
-              throw new Error(
-                `Attempting to submit a deposit for a channel with already submitted deposit ${this.store.depositsSubmitted[channelId]}`
+              logger.warn(
+                'Attempting to submit a deposit for a channel with already submitted deposit',
+                {depositSubmitted: this.store.depositsSubmitted[channelId]}
               );
+              continue;
```
andrewgordstewart, line 280: Why is this change in 69c942e9c61e06a950f973318ce77c0e49094aa4? Is it from prettier formatting on save?

If yes, this seems to indicate that `xstate-wallet` is not "up to code" WRT tooling -- we should lint on commit, as we do in other packages.

packages/xstate-wallet/src/channel-wallet.ts
```diff
@@ -286,12 +303,39 @@ export class ChannelWallet {
...
     if (!richObjective) {
       throw new Error(`Rich objective must exist for channelId ${channelId}`);
     }
-    this.crankRichObjectives({type: 'Approval'});
+
+    await this.crankRichObjectives({type: 'Approval'});
+
+    // TODO: need a better way to figure out when to broadcast an objective
+    //        since we don't know here if we are the creator of the objective or received the objective from elsewhere
+    if (!richObjective.myIndex) {
```
andrewgordstewart, line 316:
```suggestion
    const amJoiner = richObjective.myIndex === 1;
    if (amJoiner) {
```

packages/xstate-wallet/src/channel-wallet.ts
```diff
@@ -286,12 +303,39 @@ export class ChannelWallet {
...
+      // TODO: generalize to other objectives
+      const objectiveToSend: OpenChannel = {
+        type: 'OpenChannel',
+        participants: richObjective.openingState.participants,
+        data: {targetChannelId: channelId, fundingStrategy: 'Direct'}
+      };
+
+      // TODO: we might want to first crank the objective, gather all states that need to be sent, and then send
+      // both the objective and the new states
```
andrewgordstewart, line 325: What's the reason not to do that in this PR?

andrewgordstewart (approved):
> Interop tests refuse to exit cleanly due to ongoing ethers activity. I still do not understand the core root of the problem, but enough changes have been made for the jest process to at least exit with error code 0. We still see the following log lines at the end of interop tests

I've successfully used https://github.com/Raynos/leaked-handles in the past to debug this sort of thing.

I'm not sure if it would help here, as you had a handle on _what_ the open handle is already. IIRC it does log a stack trace, which may help.

kerzhner:
> I've successfully used https://github.com/Raynos/leaked-handles in the past to debug this sort of thing.
> 
> I'm not sure if it would help here, as you had a handle on _what_ the open handle is already. IIRC it does log a stack trace, which may help.

Filed https://github.com/statechannels/statechannels/issues/3494.

packages/interop-tests/src/__test__/interop.test.ts
kerzhner: That would be pretty nice https://github.com/statechannels/statechannels/issues/3495

packages/interop-tests/jest/jest.config.js
kerzhner: See PR description for the reason we had to do this.

packages/interop-tests/src/__test__/interop.test.ts
kerzhner: It would be nice to create a test suite that tests all of these combinations. At the moment:
- create channel API differs between server and browser wallet.
- message passing between server and browser wallets needs to be cleaned up in the interop test.

packages/xstate-wallet/src/channel-wallet.ts
kerzhner: In theory, `destroy` might be a functionality used for more than testing. We can imagine an mega-wallet that manages many wallet instances. Pretty far fetched at the moment...

packages/xstate-wallet/src/channel-wallet.ts
kerzhner: Good call, but your logic above is the inverse of what we need. e01de9914

packages/xstate-wallet/src/channel-wallet.ts
kerzhner: It is not clear to me how to batch together response data (akin to the `response` object in the server-wallet). I suspect this work will involve a somewhat significant refactor.

packages/xstate-wallet/src/channel-wallet.ts
kerzhner: That commit includes too much code.

This is a logic change to the behavior of the browser wallet. This is a better way to track submitted deposits: 33b2f3720

packages/xstate-wallet/src/channel-wallet.ts
andrewgordstewart: All the more reason 😆

packages/interop-tests/jest/jest.config.js
andrewgordstewart: yep, I appreciated the clarification in the commit message as well.

packages/interop-tests/src/__test__/interop.test.ts
andrewgordstewart: Is there a reason not to handle it in this PR?

packages/xstate-wallet/src/channel-wallet.ts
andrewgordstewart: hmm. I'm wary of adding "just in case" methods to the public API when they were included specifically for tests.

packages/xstate-wallet/src/channel-wallet.ts
andrewgordstewart: ah, I did not read the diff carefully, and assumed it was generated by prettier.

Are the arguments in the correct order? In a vanilla pino logger, the message would come second, right?

packages/interop-tests/src/__test__/interop.test.ts
andrewgordstewart: ... continued

it seems to me like you knew something about the value of `postFundA`, that it's currently deterministically the third event, containing some specific data. If you expect something of that specific data, then it detects regressions from that assumption.

kerzhner: We can. My preference is to merge this work and then address this as a follow up.

packages/xstate-wallet/src/channel-wallet.ts
kerzhner: In general, I agree. In this case, we would need a test channel wallet + a test chain watcher. Seems like a lot of wrappers for a small destroy method.

packages/xstate-wallet/src/channel-wallet.ts
kerzhner: Good catch on the arguments. That line has been ripped out in 33b2f37 though

packages/interop-tests/src/__test__/interop.test.ts
andrewgordstewart: Your call!

packages/xstate-wallet/src/channel-wallet.ts
andrewgordstewart: What about the alternative of clearly marking it as test code?

kerzhner: Good idea ce2cafc2f

PR #3503, opened 2021-04-16
Factor our UI functionality from browser-wallet ChannelWallet

kerzhner:
This PR separates UI from the state-channels-logic code

### DIFF
--- a/packages/browser-wallet/src/channel-wallet.ts
+++ b/packages/browser-wallet/src/channel-wallet.ts
@@ -218,16 +218,6 @@ export class ChannelWallet {
     return workflow;
   }
 
-  // TODO: factor UI out of channel wallet
-  // private renderUI(machine) {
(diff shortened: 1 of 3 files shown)


packages/browser-wallet/src/channel-wallet.ts
```diff
@@ -34,12 +31,11 @@ import {
 import {ADD_LOGS, WALLET_VERSION} from './config';
 import {logger} from './logger';
 import {ChainWatcher} from './chain';
-// TODO: factor UI out of channel-wallet
-// import {Wallet as WalletUi} from './ui/wallet';
 
+export type AnyInterpreter = Interpreter<any, any, any>;
```
andrewgordstewart, line 35: I suppose that when we care, we should leverage generic types to make `renderUI` easier to write.

andrewgordstewart (approved):
❤️

Issue #3497, opened 2021-04-14
[xstate-wallet] rich objective based UI


kerzhner:
The wallet UI for creating + funding a directly funded channel should be migrated from the old workflow-based logic to the new rich objective-based logic.

PR #3505, opened 2021-04-17
[browser-wallet] Add UI for Open Channel Objective

kerzhner:
Browser wallet UI can be:
- xstate machine based. This is the established UI methodology in the browser wallet.
- objective based. This is added in this PR.

For objective based UI, the following are the ways that the `ChannelWallet` communicates with the UI and the UI communicates with the `ChannelWallet`:
- On initialization, the `ChannelWallet` is supplied an `updateUI` function that the `ChannelWallet` must invoke whenever the `ChannelWallet` needs the UI to update.
- The UI is supplied a function that can be invoked with an Objective event argument. This function is wired up on `ChannelWallet` initialization.

This PR has been tested with:
- existing unit and interop tests.
- storybook to display new UI.
- manual calls to browser wallet APIs.

https://user-images.githubusercontent.com/1062379/115122417-1ce9b380-9f75-11eb-84bc-9510dd34082a.mov

Fixes https://github.com/statechannels/statechannels/issues/3497.

### DIFF
--- a/packages/browser-wallet/src/channel-wallet-types.ts
+++ b/packages/browser-wallet/src/channel-wallet-types.ts
@@ -8,7 +8,7 @@ export type TriggerObjectiveEvent = (event: DirectFunder.OpenChannelEvent) => vo
  * when UI should be updated.
  *
  * UI needs a way to communicate objective events to the channel wallet. UI invokes triggerObjectiveEvent
- * when UI triggers an objetive event
+ * when UI triggers an objective event
  */
 export type UpdateUI = (update: {
   service?: AnyInterpreter;
--- a/packages/browser-wallet/src/channel-wallet-types.ts
+++ b/packages/browser-wallet/src/channel-wallet-types.ts
@@ -3,9 +3,19 @@ import {Interpreter} from 'xstate';
 
 export type AnyInterpreter = Interpreter<any, any, any>;
 export type OnObjectiveEvent = (event: DirectFunder.OpenChannelEvent) => void;
-export type OnWorkflowStart = (service: AnyInterpreter, onObjectiveEvent: OnObjectiveEvent) => void;
-export type OnObjectiveUpdate = (
-  objective: DirectFunder.OpenChannelObjective,
+/**
+ * The channel wallet is not able to update UI. The channel wallet is supplied a callback to invoke
+ * when UI should be updated.
+ *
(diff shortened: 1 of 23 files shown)


packages/browser-wallet/src/channel-wallet-types.ts
```diff
@@ -0,0 +1,27 @@
+import {DirectFunder, SharedObjective, SignedState} from '@statechannels/wallet-core';
+import {Interpreter} from 'xstate';
+
+export type AnyInterpreter = Interpreter<any, any, any>;
+export type OnObjectiveEvent = (event: DirectFunder.OpenChannelEvent) => void;
```
lalexgap, line 5: The name`OnObjectiveEvent` seems to indicate an event handler but it's actually used by the UI to trigger an objective event. Could we name it something like `triggerObjectiveEvent` ?

packages/browser-wallet/src/channel-wallet-types.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * The channel wallet is not able to update UI. The channel wallet is supplied a callback to invoke
+ * when UI should be updated.
+ *
+ * UI needs a way to communicate objective events to the channel wallet. UI invokes onObjectiveEvent
+ * when UI triggers an objetive event
```
lalexgap, line 11:
```suggestion
 * when UI triggers an objective event
```

packages/browser-wallet/src/channel-wallet.ts
```diff
@@ -32,35 +31,38 @@ import {ADD_LOGS, WALLET_VERSION} from './config';
...
-  objectives: SharedObjective[];
-  signedStates: SignedState[];
-};
-
 export class ChannelWallet {
   public workflows: Workflow[];
+  /**
+   *
+   * @param chainAddress Ethereum address for the wallet
+   * @param updateUI Callback that will be invoked when the channel wallet wants to update UI
```
lalexgap, line 39: Could the docs be updated to describe the `setOnObjectiveEvent` param?

lalexgap (approved):

packages/browser-wallet/src/channel-wallet-types.ts
kerzhner: Good call. Addressed in 9141a23

PR #3508, opened 2021-04-19
deploy nitro contracts to arbitrum testnet

geoknee:
This is mostly just exploration of how easy it might be to deploy our stuff to arbitrum.

Answer: very easy.

See this transaction for example https://explorer.arbitrum.io/#/tx/0xc97de88cecc8b162a994577768cb07cf36496809f05cc2d00a1ead396748f96f 

I don't believe it is straightforward for us to run our test suite against something other than a local ganache instance... so I didn't do that yet.

### DIFF
--- a/packages/nitro-protocol/addresses.json
+++ b/packages/nitro-protocol/addresses.json
@@ -4,11 +4,11 @@
       "name": "mainnet",
       "chainId": "1",
       "contracts": {
-        "NitroAdjudicator": {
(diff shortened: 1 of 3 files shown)


andrewgordstewart:
> I don't believe it is straightforward for us to run our test suite against something other than a local ganache instance... so I didn't do that yet.

I know @kerzhner ran _some_ tests (may `test:chain` scoped tests?) against Goerli.

NiloCK (approved):
This looks good to me, and I have learned something about our deployment process and configuration.

PR #3511, opened 2021-04-19
[browser-wallet] Introduce RichObjective and RichObjectiveEvent types

kerzhner:
🥞 stacked on https://github.com/statechannels/statechannels/pull/3509

Browser wallet is the only consumer of rich objectives at the moment. Browser wallet hardcoded the following assumptions:
- Only one rich objective is expected in the store. As a result, the browser wallet assumes that an objective event is intended for the stored objective.
- Only the DirectFunder open channel objective is considered.

This PR introduces the `RichObjective` and `RichObjectiveEvent` abstractions to help the browser wallet remove the above assumptions. At the moment, a RichObjective is always DirectFunder.OpenChannelObjective and a RichObjectiveEvent is always a DirectFunder.OpenChannelEvent

### DIFF
--- a/packages/browser-wallet/src/channel-wallet.ts
+++ b/packages/browser-wallet/src/channel-wallet.ts
@@ -12,7 +12,9 @@ import {
   OpenChannel,
   Address,
   DirectFunder,
-  makeAddress
+  makeAddress,
+  RichObjectiveEvent,
+  RichObjective
 } from '@statechannels/wallet-core';
 import _ from 'lodash';
 
@@ -64,7 +66,7 @@ export class ChannelWallet {
     protected updateUI?: UpdateUI,
     setTriggerObjectiveEvent?: (callback: TriggerObjectiveEvent) => void
   ) {
-    setTriggerObjectiveEvent?.(_.bind(this.crankRichObjectives, this));
+    setTriggerObjectiveEvent?.(_.bind(this.crankRichObjective, this));
     this.workflows = [];
 
     // Whenever the store wants to send something call sendMessage
@@ -72,7 +74,7 @@ export class ChannelWallet {
       this.messagingService.sendMessageNotification(m);
     });
 
-    store.crankRichObjectivesFeed.subscribe(_.bind(this.crankRichObjectives, this));
+    store.crankRichObjectivesFeed.subscribe(_.bind(this.crankRichObjective, this));
 
     // Whenever an OpenChannel objective is received
     // we alert the user that there is a new channel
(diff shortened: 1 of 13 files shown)


packages/browser-wallet/src/channel-wallet.ts
```diff
@@ -242,58 +244,59 @@ export class ChannelWallet {
...
-              tx: txHash ?? '',
-              attempt: 0,
-              submittedAt: Date.now()
-            });
-
-            break;
-          default:
-            throw new Error('Not expected to reach here');
-        }
+  private async crankRichObjective(event: RichObjectiveEvent): Promise<void> {
```
lalexgap, line 247: It seems that there are some whitespace changes here. Could we run `prettier` against all files in `browser-wallet` in a separate PR?

kerzhner: A loop was removed from `crankRichObjective`. So the diff appears as a whitespace change.

That being said, it is not a bad idea to run prettier anyways!

andrewgordstewart: You can also hide whitespace changes.
- there's a UI way to do it
- or you can add the `&w=1` param. ([link](https://github.com/statechannels/statechannels/pull/3511/files?diff=split&w=1))

lalexgap: Good point! I think I noticed some whitespace changes somewhere else as well but it's quite possible I'm mistaken.

packages/browser-wallet/src/index.ts
```diff
@@ -79,12 +79,9 @@ function setTriggerObjectiveEvent(callback: TriggerObjectiveEvent): void {
...
 })();
 
-function renderUI(update: {
-  service?: AnyInterpreter;
-  objective?: DirectFunder.OpenChannelObjective;
-}) {
+function renderUI(update: {service?: AnyInterpreter; objective?: RichObjective}) {
   if (!triggerObjectiveEvent) {
-    throw new Error('triggerObjectiveEventCallback must be defined');
+    throw new Error('onObjectiveEventCallback must be defined');
```
lalexgap, line 84:
```suggestion
    throw new Error('triggerObjectiveEvent must be defined');
```

packages/browser-wallet/src/store/store.ts
```diff
@@ -582,11 +581,16 @@ export class Store {
         objectives.map(_.bind(this.createAndStoreRichObjectiveFromObjective, this))
       );
 
-    if (signedStates) {
-      this._eventEmitter.emit('crankRichObjectives', {
-        type: 'StatesReceived',
-        states: signedStates
-      });
+    if (signedStates?.length) {
```
lalexgap, line 584: nit but I found I had to think through this check for a second.

I'd prefer a more explicit but less concise check of `signedStates && signedStates.length > 0` but this might just be a matter of preference.

packages/browser-wallet/src/ui/wallet.tsx
```diff
@@ -16,8 +16,8 @@ import {ObjectiveContext} from './objective/objective-context';
 interface Props {
   workflow?: Interpreter<any, any, any>;
   // TODO: generalize to all rich objectives
```
lalexgap, line 18: Can this TODO go?

lalexgap (approved):

packages/browser-wallet/src/store/store.ts
kerzhner: I tend to read this as: is `signedStates` defined? If so, is `signedStates.length` truthy?

Admittedly I might be a bit too enamored with optional chaining.

packages/browser-wallet/src/ui/wallet.tsx
kerzhner, line 18:
```suggestion
```

kerzhner: Good catch

packages/browser-wallet/src/store/store.ts
andrewgordstewart: I don't see optional changing as the culprit here. `signedStates.length > 0` is a more explicit clause than `signStates.length`, reducing the mental burden for the reader (at least when the reader is me).

`signedStates?.length > 0` is unfortunately not allowed. There is another, higher level change that I believe would address this and be a slight simplification for developers: why isn't `signedStates` known to be a `SignedState[]` at this point? `pushMessage2` could require `signedStates` to be defined, for instance.

When sending messages across the wire, it makes sense to make properties like `signedStates` be optional, as you save on network overhead. But at the point of deserialization, we could choose to fill in the default values `[]`. This would simplify the lives of developers (across multiple codebases), who no longer need to check if `signedStates` is defined.

PR #3513, opened 2021-04-20
[browser-wallet] Guarantee type safety for rich objective feeds

kerzhner:
The browser wallet store defines feeds. The feeds are event emitters wrapped in an observable. This pattern does NOT provide type safety. The store can emit an event where the event name does not match the event data type.

With rxjs subjects, we guarantee type safety as well as reduce a level of indirection. Only rich objective feeds have been refactored. Legacy feeds from wallet
v1 remain as is.


kerzhner:
Closing in favor of https://github.com/statechannels/statechannels/pull/3520

PR #3516, opened 2021-04-21
chore: add no self compare lint rule

lalexgap:
Enables the no-self compare lint rule to prevent us from shooting ourselves in the feet.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,7 @@ module.exports = {
     'plugin:import/typescript'
   ],
   rules: {
+    'no-self-compare': 'error',
     'import/order': [
       1,
       {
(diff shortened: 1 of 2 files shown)


lalexgap:
During standup I mentioned enabling this rule showed a bunch of errors but I was incorrect. I was thinking of the [no shadow rule](https://eslint.org/docs/rules/no-shadow)

andrewgordstewart (approved):

kerzhner:
Should we add this rule to the monorepo `.eslintrc.js`?

lalexgap:
> Should we add this rule to the monorepo `.eslintrc.js`?

Yup! Done in f4b775f

kerzhner (approved):

PR #3520, opened 2021-04-22
[browser-wallet] Guarantee type safety for rich objective feeds

kerzhner:
🥞 stacked on https://github.com/statechannels/statechannels/pull/3511

The browser wallet store defines feeds. The feeds are event emitters wrapped in an observable. This pattern does NOT provide type safety. The store can emit an event where the event name does not match the event data type.

With rxjs subjects, we guarantee type safety as well as reduce a level of indirection. Only rich objective feeds have been refactored. Legacy feeds from wallet
v1 remain as is.

### DIFF
--- a/packages/browser-wallet/src/store/store.ts
+++ b/packages/browser-wallet/src/store/store.ts
@@ -74,7 +74,7 @@ export class Store {
   public richObjectives: Dictionary<RichObjective> = {};
 
   public richObjectiveFeed = new Subject<RichObjective>();
-  public crankRichObjectivesFeed = new Subject<RichObjectiveEvent>();
+  public crankRichObjectiveFeed = new Subject<RichObjectiveEvent>();
 
   /**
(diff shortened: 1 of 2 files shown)


packages/browser-wallet/src/store/store.ts
```diff
@@ -75,6 +73,9 @@ export class Store {
    */
   public richObjectives: Dictionary<RichObjective> = {};
 
+  public richObjectiveFeed = new Subject<RichObjective>();
+  public crankRichObjectiveFeed = new Subject<RichObjectiveEvent>();
```
lalexgap, line 77: How about a name like `richObjectiveEventFeed` ?

lalexgap (approved):

packages/browser-wallet/src/store/store.ts
kerzhner: Good call! 412095b0f

PR #3524, opened 2021-04-23
Fix pushMessage updating approved objective statuses to pending

### LINKED ISSUE #3523: PushMessage sets objective status to pending
**Describe the bug**

Currently the `PushMessage` function calls `ObjectiveModel.Insert` [on each objective it receives in a message](https://github.com/statechannels/statechannels/blob/3309bd2fc98e9026f90b156d7137f793aab01583/packages/server-wallet/src/engine/store.ts#L326:L326).

The `ObjectiveModel.Insert` has some upsert logic as seen [here](https://github.com/statechannels/statechannels/blob/ag/approve-o-api/packages/server-wallet/src/models/objective.ts#L196:L196):

```
  const model = await query
        .onConflict('objectiveId')
        .merge({status: preApproved ? 'approved' : 'pending'});
```

However, since we call `Objective.Insert` with `preApproved=false` (which is set [here](https://github.com/statechannels/statechannels/blob/3309bd2fc98e9026f90b156d7137f793aab01583/packages/server-wallet/src/engine/store.ts#L325)) that means the objective will be updated with a status of `pending` **even if it's already been approved**

I discovered this with the introduction of the `ApproveObjective` API. Now that participant B was calling `ensureObjectives` it meant that B was sending its objectives to A. When A received these objectives it would overwrite it's approved status with `pending`.

**To Reproduce**
This can be seen on the CI tests on https://github.com/statechannels/statechannels/pull/3485. If you disable database deletion after the tests run you can see that A's objectives are pending.

lalexgap:
# Description
Stops `PushMessage` from updating the status of approved objectives to pending.
Fixes #3523.

Before this PR the `pushMessage` on method on the `store` would call `ObjectiveModel.insert` on each objective it received in a message.

For every objective type besides `CloseChannel` `pushMessage` would call `ObjectiveModel.insert` with `preApprove` set to `false`.

This means that `ObjectiveModel.Insert` would update the status of approved objectives to pending.

The change in this PR is to switch from a `preApprove` boolean flag to an optional status argument: `approvalStatus?:'approved'|'pending'`. This allows us to only override the status if we intend to. Otherwise, it is left untouched.

It does seems like a code smell that we have a method called `insert` that's a partial `upsert` but I didn't want to increase the scope of this too much.

## How Has This Been Tested?
A test for the specific scenario was added in 4f9ae21

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
+++ b/packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
@@ -27,7 +27,7 @@ import {MultiThreadedEngine, Engine} from '../..';
 const dropNonVariables = (s: SignedState): any =>
   _.pick(s, 'appData', 'outcome', 'isFinal', 'turnNum', 'stateHash', 'signatures');
 
-jest.setTimeout(120_000);
+jest.setTimeout(20_000);
 
 let engine: Engine;
 let multiThreadedEngine: MultiThreadedEngine;
--- a/packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
+++ b/packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
@@ -3,6 +3,8 @@ import {
   simpleEthAllocation,
   serializeState,
   SignedState,
+  OpenChannel,
+  objectiveId,
 } from '@statechannels/wallet-core';
 import {ChannelResult} from '@statechannels/client-api-schema';
 import _ from 'lodash';
@@ -53,6 +55,35 @@ const zero = 0;
 const four = 4;
 const five = 5;
 const six = 6;
+
+it('does not set the status of an objective if it already exists', async () => {
+  const {channelId} = await Channel.query(engine.knex).insert(channel());
+
+  const objective: OpenChannel = {
+    type: 'OpenChannel',
+    participants: [],
+    data: {
+      targetChannelId: channelId,
+      fundingStrategy: 'Direct',
+      role: 'app',
+    },
+  };
+
+  const inserted = await ObjectiveModel.insert(objective, true, engine.knex);
+  //Sanity check: We expect the objective to be approved since we used the preApproved flag
+  expect(inserted.status).toBe('approved');
+
+  await engine.pushMessage({
+    walletVersion: WALLET_VERSION,
+    signedStates: [],
+    objectives: [objective],
+  });
+
+  const latest = await ObjectiveModel.forId(objectiveId(objective), engine.knex);
+
+  // The status should still be approved after we receive the message in a payload
+  expect(latest.status).toBe('approved');
+});
 
 it('stores states contained in the message, in a single channel model', async () => {
   const channelsBefore = await Channel.query(engine.knex).select();
--- a/packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
+++ b/packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
@@ -14,7 +14,7 @@ import {addHash} from '../../../state-utils';
 import {alice, bob, charlie} from '../fixtures/signing-wallets';
 import {alice as aliceP, bob as bobP, charlie as charlieP} from '../fixtures/participants';
 import {seedAlicesSigningWallet} from '../../../db/seeds/1_signing_wallet_seeds';
-import {stateSignedBy} from '../fixtures/states';
+import {stateSignedBy, stateWithHashSignedBy} from '../fixtures/states';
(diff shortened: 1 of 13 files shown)


packages/server-wallet/src/models/objective.ts
```diff
@@ -191,9 +191,12 @@ export class ObjectiveModel extends Model {
         .returning('*')
         .first();
 
-      const model = await query
-        .onConflict('objectiveId')
-        .merge({status: preApproved ? 'approved' : 'pending'});
+      // Knex does not like properties with undefined values so we strip them out
+      const override = _.pickBy({status: approvalStatus, waitingFor}, _.identity);
+      // This allows for for the insert method to override the existing status or waitingFor if desired
+      const model = _.isEmpty(override)
```
kerzhner, line 197: Curious, is the conditional necessary here? Does `const model = await query.onConflict('objectiveId').merge(override);` work with a empty `override` object?

packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
```diff
@@ -25,7 +27,7 @@ import {MultiThreadedEngine, Engine} from '../..';
 const dropNonVariables = (s: SignedState): any =>
   _.pick(s, 'appData', 'outcome', 'isFinal', 'turnNum', 'stateHash', 'signatures');
 
-jest.setTimeout(20_000);
```
kerzhner, line 28: Was this timeout increased while playing around with test? Or do we need the longer timeout?

packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
```diff
@@ -54,6 +56,36 @@ const four = 4;
...
+    participants: [],
+    data: {
+      targetChannelId: channelId,
+      fundingStrategy: 'Direct',
+      role: 'app',
+    },
+  };
+
+  const inserted = await ObjectiveModel.insert(objective, engine.knex, 'approved');
+  //Sanity check: We expect the objective to be approved since we used the preApproved flag
```
kerzhner, line 76: Nit
```suggestion
  //Sanity check: We expect the objective to be approved since we used the 'approved' parameter
```

packages/server-wallet/src/protocols/__test__/defunder.test.ts
```diff
@@ -40,17 +40,17 @@ async function ensureCloseObjective(
...
   );
 }
 
 async function ensureDefundObjective(
   channel: TestChannel,
   tx: Transaction
 ): Promise<WalletObjective<DefundChannel>> {
-  // add a preapproved defundChannel objective and approve
-  return ObjectiveModel.insert(channel.defundChannelObjective(), true, tx);
+  // add a approved defundChannel objective and approve
```
kerzhner, line 52:
```suggestion
  // add an approved defundChannel objective and approve
```

packages/server-wallet/src/objectives/close-channel.ts
```diff
@@ -23,8 +23,9 @@ export class CloseChannelObjective {
               txSubmitterOrder: [1, 0],
```
kerzhner, line 23: Side note: It seems like the `CloseChannelObjective` does not use the shared `insert` method? Do you know if there is a reason for that?

kerzhner (approved):

packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
lalexgap: Ah, good catch. This was changed when I was originally investigating the issue and I forgot to modify it back.

packages/server-wallet/src/models/objective.ts
lalexgap: Yup, unfortunately `knex` fails when passed in an empty objective.
 ``` 
Empty .update() call detected! Update data does not contain any values to update. This will result in a faulty query. Table: objectives. 
```

packages/server-wallet/src/objectives/close-channel.ts
lalexgap: Doesn't `CloseChannelObjective` use the same `ObjectiveModel.Insert` that other code uses?

packages/server-wallet/src/engine/__test__/integration/push-message.test.ts
lalexgap: b9cac13

packages/server-wallet/src/objectives/close-channel.ts
kerzhner: Yeah you are totally right. I misread this code.

PR #3526, opened 2021-04-24
add test for create2ing the channel

geoknee:
Record a reference of what deploying our proxies would be.

Pen and paper suggests 62K , the test suggests 64.7K, implying about 2.7K for constructing the proxy creation code (via [`abi.encodePacked`](https://ethereum.stackexchange.com/questions/729/how-to-concatenate-strings-in-solidity)). 

## Changes 
1. Code path stripped down to minimize unnecessary gas expenditure (no event emitted, for example)
2. test added
---
### Code quality
### Project management

### DIFF
--- a/packages/nitro-protocol/contracts/ninja-nitro/AdjudicatorFactory.sol
+++ b/packages/nitro-protocol/contracts/ninja-nitro/AdjudicatorFactory.sol
@@ -42,7 +42,6 @@ contract AdjudicatorFactory {
     /// @dev Allows us to create new channel contract and get it all set up in one transaction
     function createChannel(bytes32 channelId) public returns (address payable channel) {
         channel = payable(_deployChannelProxy(channelId));
-        emit ChannelCreation(channel);
     }
 
     /// @dev Allows us to create new channel contract, payout all of the funds, and destroy the contract
--- a/packages/nitro-protocol/contracts/ninja-nitro/AdjudicatorFactory.sol
+++ b/packages/nitro-protocol/contracts/ninja-nitro/AdjudicatorFactory.sol
@@ -40,8 +40,8 @@ contract AdjudicatorFactory {
     }
 
     /// @dev Allows us to create new channel contract and get it all set up in one transaction
-    function createChannel(bytes32 channelId) public returns (address payable channel) {
-        channel = payable(_deployChannelProxy(channelId));
+    function createChannel(bytes32 channelId) public {
+        _deployChannelProxy(channelId);
     }
 
     /// @dev Allows us to create new channel contract, payout all of the funds, and destroy the contract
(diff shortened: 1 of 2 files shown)


geoknee:
Bytes concatenation looks likely to be available in solidity in the near future https://github.com/ethereum/solidity/issues/10903 . It would be more readable and possibly more gas efficient than using `abi.encodePacked` (as the current implementation we use does) to do the concatenation required in[ EIP 1167](https://eips.ethereum.org/EIPS/eip-1167).

geoknee:
It turns out `bytes.concat` is already available in solidity 0.8.x . I upgraded and made the switch on [this branch](https://github.com/statechannels/statechannels/compare/create2-cost...bytes-concat?expand=1): upshot is, we can save 19 gas  this way 😄.

packages/nitro-protocol/contracts/ninja-nitro/AdjudicatorFactory.sol
```diff
@@ -76,7 +76,7 @@ contract AdjudicatorFactory {
 
     /// @dev Allows us to create new channel contact using CREATE2
     function _deployChannelProxy(bytes32 channelId) internal returns (address) {
-        bytes32 salt = channelId;
```
NiloCK, line 80: I am curious whether `solc` optimizes this away.

geoknee: I would have thought any compiler worth its salt would do that :-P (but I haven't checked)

packages/nitro-protocol/contracts/ninja-nitro/AdjudicatorFactory.sol
```diff
@@ -40,8 +40,8 @@ contract AdjudicatorFactory {
     }
 
     /// @dev Allows us to create new channel contract and get it all set up in one transaction
-    function createChannel(bytes32 channelId) public returns (address payable channel) {
-        channel = payable(_deployChannelProxy(channelId));
+    function createChannel(bytes32 channelId) public {
```
NiloCK, line 43: For my understanding: N-N doesn't require the return address because the contract address is computable.

NiloCK (approved):
This looks good. We have a number of jest tests that may be better labelled `benchmarks` than `unit tests`, or that blur the line between the two.

Is this a distinction worth configuring for?

geoknee:
> Is this a distinction worth configuring for?

I think in future, perhaps: Although that might involve some unnecessary duplication? It seems to me that testing the behaviour of some on-chain code and measuring the gas consumption of that execution are quite tightly bound (when you do one, chances are you want to also do the other)?

geoknee:
We do have something like this configured already https://github.com/statechannels/statechannels/blob/56e70cc97062f3a6cad809778ef410fa6fb719f4/.circleci/config.yml#L71-L78 although I've no doubt it could be tidied up quite a bit or enhanced. 

Overall I would like to take a more targeted approach to which gas metrics we are particularly interested in. Sketch of that [here](https://github.com/statechannels/statechannels/pull/3488/files#diff-1b85a372060ed7823e27745d6ff14f36c177694e0c3dddd9cae60f1beaf4b006R85-R91). When we get to that, we could think about the best way to collect, store and display that information.

geoknee:
(currently we just have a bucket of files, roughly one for each test file, with a line for each test case https://app.circleci.com/pipelines/github/statechannels/statechannels/12779/workflows/0498ae16-2ebd-4e0d-9b51-fd35390be922/jobs/63143/artifacts)

Issue #3531, opened 2021-04-29
Update Wallet to accept a Message Service Factory


lalexgap:
Currently, we are using the `TestMessageService` to deliver messages directly into the other participant's `engine`. However, this should be the responsibility of the `wallet`. That way the `wallet` will function with any `MessageService`.

This means that the `Wallet` should accept a `MessageServiceFactory`. It should pass in a handler that triggers `engine.pushMessage` and handle the result.

PR #3533, opened 2021-04-30
Add a Objective Monitor

### LINKED ISSUE #3530: Dropped messages can cause CloseChannel Objectives to stall out for a participant
If the right messages get dropped we can run into a scenario where one participant has a completed `CloseChannel` objective and one participant has a `pending` `CloseChannelObjective`.

My current theory is that this is due to messages containing final states being dropped. This could result in one participant having both signed final states and the other only having one.

I think the reason we only run into this with CloseChannel is that it is unique in being "pre-approved".  That means the other participant doesn't have to call `ApproveObjective` on the objective, meaning `ensureObjective` does not get called for one of the participants.

lalexgap:
# Description
Fixes #3530 

Introduces a simple ObjectiveMonitor that can be used to check if an objective is new or is complete. The `ObjectiveMonitor` is mostly used to work around the current behaviour of the `engine`. 

Since the `ObjectiveMonitor` is an in-memory cache it implies some constraints on concurrency (ie: two wallets using the same database).   When the `engine` API gets [cleaned up
](https://www.notion.so/statechannels/Remove-events-from-Engine-and-improve-Engine-API-c156edc960c34c8196e8c758d23c3939) the `ObjectiveMonitor` should be retired.

This `ObjectiveMonitor` is used to call `EnsureObjective` on `CloseChannel` objectives started by other participants, thus resolving #3540.

The `ObjectiveMonitor` is essentially an in-memory list of objectives we're watching and their status. This lets the `wallet` easily check:
- If an objective has been completed
- If there is a new objective to handle

The `wallet` can now check the `ObjectiveManager` to determine if the objective is complete or a new objective instead of listening for an ObjectiveCompleted event.

The `ObjectiveMonitor` is also an EventEmitter that emits one kind of event: `NewObjective`. This event is fired every time an objective is added to the monitor for the first time. We use this event to call `ensureObjective` on `CloseChannel` objectives to resolve #3530. 

Once the `wallet` is updated to take in a `MessageServiceFactory` for #3531 we can probably remove the `NewObjective` event and just rely on calling `monitorObjectives` after calling `pushMessage`.

## How Has This Been Tested? 

Currently, I'm relying on the existing `wallet` tests to inherently test the new functionality. The ObjectiveMonitor is pretty simple so I felt that it didn't justify a new test suite.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/__test-with-peers__/utils.ts
+++ b/packages/server-wallet/src/__test-with-peers__/utils.ts
@@ -35,7 +35,6 @@ export function waitForObjectiveEvent(
         handledObjectiveIds.add(o.objectiveId);
 
         if (handledObjectiveIds.size === _.uniq(objectiveIds).length) {
-          engine.removeListener(objectiveEventType);
           resolve();
         }
       }
--- a/packages/server-wallet/src/__test-with-peers__/wallet/close-channels.test.ts
+++ b/packages/server-wallet/src/__test-with-peers__/wallet/close-channels.test.ts
@@ -31,8 +31,8 @@ describe('CloseChannels', () => {
 
     // Delay and drop
     {dropRate: 0.1, meanDelay: 15, closer: 'A'},
-    // TODO: This test case can fail due to https://github.com/statechannels/statechannels/issues/3530
-    //{dropRate: 0.1, meanDelay: 15, closer: 'B'},
+
+    {dropRate: 0.1, meanDelay: 15, closer: 'B'},
   ];
   test.each(testCases)(
     'can successfully create and close channel with the latency options: %o',
--- a/packages/server-wallet/src/wallet/objective-monitor.ts
+++ b/packages/server-wallet/src/wallet/objective-monitor.ts
@@ -24,6 +24,7 @@ export class ObjectiveMonitor extends EventEmitter<ObjectiveMonitorEvents> {
   public constructor(engineEmitter: EngineEmitter, logger?: Logger) {
     super();
 
+    // Listen for new objectives and make sure we're monitoring them
     engineEmitter.on('objectiveStarted', async (o: WalletObjective) => {
       const {objectiveId} = o;
       if (!this.isMonitored(objectiveId)) {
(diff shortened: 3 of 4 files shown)


packages/server-wallet/src/__test-with-peers__/utils.ts
```diff
@@ -35,7 +35,6 @@ export function waitForObjectiveEvent(
         handledObjectiveIds.add(o.objectiveId);
 
         if (handledObjectiveIds.size === _.uniq(objectiveIds).length) {
-          engine.removeListener(objectiveEventType);
```
lalexgap, line 38: This was causing issues since it would remove **all** listeners for the event type. 

Since this is used only in tests and `engine.destroy` removes all listeners, I think it's fine to not remove it here.

andrewgordstewart (commented):
Here is a summary of a conversation from slack:
1. The main purpose of the `ObjectiveMonitor` is to answer the question "is this objective a new objective?"
2. The reason we need this component is, the `engine` emits `'ObjectiveStarted'` on new objectives as well as already known objectives.
3. We plan for the engine to _not_ emit events.
4. Once the engine does _not_ emit events, the `ObjectiveMonitor` should be removed.

@lalexgap do you agree with this summary?

If so,
- I'm going to review it, thinking of it as temporary code. (maybe I won't review it at all)
- We should clearly indicate that it is temporary code. (Currently, the description + the documentation on the code does not give the sense that it's temporary.)

lalexgap:
> Here is a summary of a conversation from slack:
> 
> 1. The main purpose of the `ObjectiveMonitor` is to answer the question "is this objective a new objective?"
> 2. The reason we need this component is, the `engine` emits `'ObjectiveStarted'` on new objectives as well as already known objectives.
> 3. We plan for the engine to _not_ emit events.
> 4. Once the engine does _not_ emit events, the `ObjectiveMonitor` should be removed.
> 
> @lalexgap do you agree with this summary?
> 
> If so,
> 
> * I'm going to review it, thinking of it as temporary code. (maybe I won't review it at all)
> * We should clearly indicate that it is temporary code. (Currently, the description + the documentation on the code does not give the sense that it's temporary.)

👍  I've updated the comments to reflect that the `ObjectiveMonitor` will not be long lived.

lalexgap:
Closing this as per this [doc](https://www.notion.so/statechannels/Engine-Refactor-Design-Question-fa5fad1c9e8740ca969ad0688e3fa374). We'll just query the store instead of using a Objective Manager

PR #3543, opened 2021-05-24
make interop-tests private

geoknee:
We should also unpublish https://www.npmjs.com/package/interop-tests if we can.


NiloCK (approved):
As I understand it, this flag prevents the package from being published via automated or semi-automated publishing pipelines like gh-actions or lerna-publish. ?

Sure :)

geoknee:
> We should also unpublish https://www.npmjs.com/package/interop-tests if we can.

done

PR #3544, opened 2021-05-24
Improved gas benchmarking 

geoknee:
https://www.notion.so/statechannels/Improved-gas-benchmarking-in-nitro-protocol-d14938f9671a48588e6974462e9ffb00

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/fixtures.ts
+++ b/packages/nitro-protocol/gas-benchmarks/fixtures.ts
@@ -8,7 +8,7 @@ export const channelNonce = 2;
 export const Alice = new Wallet(
   '0x277fb9e0ad81dc836c60294e385b10dfcc0a9586eeb0b1d31da92e384a0d2efa'
 );
-export const Bob = new Wallet('0x277fb9e0ad81dc836c60294e385b10dfcc0a9586eeb0b1d31da92e384a0d2efa');
+export const Bob = new Wallet('0xc8774aa98410b3e3281ff1ec40ea2637d2b9280328c4d1ff00d06cd95dd42cbd');
 export const participants = [Alice.address, Bob.address];
 
 const channel: Channel = {chainId, channelNonce, participants};
--- a/packages/nitro-protocol/gas-benchmarks/fixtures.ts
+++ b/packages/nitro-protocol/gas-benchmarks/fixtures.ts
@@ -1,0 +1,19 @@
+import {Wallet} from '@ethersproject/wallet';
+
+import {Channel, getChannelId} from '../src';
+
+export const chainId = '0x1';
+export const channelNonce = 2;
+
+export const Nigel = new Wallet(
+  '0x277fb9e0ad81dc836c60294e385b10dfcc0a9586eeb0b1d31da92e384a0d2efa'
+); // Pays gas for infastructure deployments.
+
+export const Alice = new Wallet(
+  '0x277fb9e0ad81dc836c60294e385b10dfcc0a9586eeb0b1d31da92e384a0d2efa'
+);
+export const Bob = new Wallet('0x277fb9e0ad81dc836c60294e385b10dfcc0a9586eeb0b1d31da92e384a0d2efa');
+export const participants = [Alice.address, Bob.address];
+
+const channel: Channel = {chainId, channelNonce, participants};
+export const channelId = getChannelId(channel);
--- a/packages/nitro-protocol/gas-benchmarks/fixtures.ts
+++ b/packages/nitro-protocol/gas-benchmarks/fixtures.ts
@@ -5,7 +5,7 @@ import {Channel, getChannelId} from '../src';
 export const chainId = '0x1';
 export const channelNonce = 2;
 
-export const TokenAddress = '0x6b175474e89094c44da98b954eedeac495271d0f'; // DAI stablecoin on mainnet
+export const tokenAddress = '0x6b175474e89094c44da98b954eedeac495271d0f'; // DAI stablecoin on mainnet
 
 export const Alice = new Wallet(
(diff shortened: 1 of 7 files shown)


andrewgordstewart (commented):
This seems useful!

I am not convinced that we should aim to keep vanilla nitro, mono nitro, and ninja nitro around in the same repo. (I can see having mono + ninja around.)

If the purpose of this is to assert that gas costs have not changed significantly, it belongs in this repository. If the purpose is to do a comparison, perhaps this code belongs in `cross-chain-swaps`? (I think the former is valuable!)

Another point: the gas consumption is a function of which version of the Ethereum blockchain it is tested against. Is the purpose of this test to record via test code how much gas an operation consumes? Should it be to notify us of a change in gas consumption due to hard forks? (That type of script probably ought to run on a schedule, not in a CI run?)

geoknee:
> I am not convinced that we should aim to keep vanilla nitro, mono nitro, and ninja nitro around in the same repo. (I can see having mono + ninja around.)

Ok -- I guess we don't need to decide right now. My plan is to "finish" the suite of benchmarks using vanilla nitro. We can add / remove / replace implementations later as we see fit. 

> 
> If the purpose of this is to assert that gas costs have not changed significantly, it belongs in this repository. If the purpose is to do a comparison, perhaps this code belongs in `cross-chain-swaps`? (I think the former is valuable!)

I think this code serves both purposes, and it certainly needs to cover cases beyond cross chain swaps. 

> 
> Another point: the gas consumption is a function of which version of the Ethereum blockchain it is tested against. Is the purpose of this test to record via test code how much gas an operation consumes? Should it be to notify us of a change in gas consumption due to hard forks? (That type of script probably ought to run on a schedule, not in a CI run?)

Yes the purpose of the tests is to record how much gas is consumed by a certain "path" (i.e. directly funding a channel with ERC20 tokens). That's a function of our solidity code, the current hardfork, the compiler optimization flags and so on. We can get notified any time the gas consumption changes, due to any change in those factors. I think it should run in CI, because that way we enforce that the gas values committed to the repository remain in sync with all of the aforementioned factors. 

On hardforks: ganache is actually some way behind the latest ethereum hardfork https://github.com/trufflesuite/ganache-cli/issues/821 . We might want to think about using an alternative toolchain so we can stay closer to the bleeding edge. 

When an update to ganache is available that allows a new hardfork to be selected, we can raise a PR that upgrades ganache. Tests will fail on the first commit, but we update the gas consumption numbers, tests pass and we get a very clear snapshot of what the hardfork did to our numbers.

geoknee:
> When an update to ganache is available that allows a new hardfork to be selected, we can raise a PR that upgrades ganache. Tests will fail on the first commit, but we update the gas consumption numbers, tests pass and we get a very clear snapshot of what the hardfork did to our numbers.

See #3558

packages/nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -0,0 +1,53 @@
+type GasRequiredTo = Record<
+  Path,
+  {
+    vanillaNitro: number | any;
```
NiloCK, line 4:
```suggestion
    vanillaNitro: number | Record<string, number>;
```
(nit)

packages/nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import {Channel, getChannelId} from '../src';
+
+export const chainId = '0x1';
+export const channelNonce = 2;
+
+export const Alice = new Wallet(
+  '0x277fb9e0ad81dc836c60294e385b10dfcc0a9586eeb0b1d31da92e384a0d2efa'
+);
+export const Bob = new Wallet('0x277fb9e0ad81dc836c60294e385b10dfcc0a9586eeb0b1d31da92e384a0d2efa');
```
NiloCK, line 11: Alice and Bob are constructed here from the same private key: 0x277...efa
- is this number significant? If not, why use it rather than 0x1?
- if I understand correctly, this will also mean that they have the same public key / address. Does that make sense?

NiloCK (approved):
This looks good to me.

packages/nitro-protocol/gas-benchmarks/gas.ts
geoknee: Yeah I tried this before but it didn't seem to work! Will have another go

packages/nitro-protocol/gas-benchmarks/fixtures.ts
geoknee: We should at least use distinct keys for Alice and Bob, I think. Will fix that. While it doesn't make much sense for an actual state channel that might get used, I don't think it makes any difference to the gas consumption to have the two the same (which is the main concern here). Still it's easy enough to have distinct ones, and less confusing. Also having two the same might confuse the gas accounting if we end up paying out money to Alice and Bob.

The number chosen for the private key is not significant: but I wanted to avoid magic values with many zeros in them, in case that were to lead to some unrealistic gas savings somewhere. I generated the private key using `ethers.Wallet.createRandom()`.

packages/nitro-protocol/gas-benchmarks/gas.ts
geoknee: 977ef96

packages/nitro-protocol/gas-benchmarks/fixtures.ts
geoknee: Interestingly, changing Bob's private key did have a small effect on gas consumption 0161c5b. I think this is because of the follow-on change to the `channelId` used in the benchmarking test -- I'm not entirely sure why writing to the `holdings` mapping with a different key would cost a different amount of gas...

PR #3545, opened 2021-05-25
POC of embedding an application channel in a three-party joint channel

geoknee:
See https://www.notion.so/statechannels/Virtual-funding-research-11388f704f84496ba8febe30ddb522d2

### DIFF
--- a/packages/nitro-protocol/contracts/examples/TwoOfThree.sol
+++ b/packages/nitro-protocol/contracts/examples/TwoOfThree.sol
@@ -52,10 +52,7 @@ contract TwoOfThree is IForceMoveApp2 {
             );
 
             // participant 2's slot may not change
-            require(
-                allocationA[2].amount == allocationB[2].amount,
-                'amount for participant 2 cannot change without their signature'
-            );
+            require(allocationA[2].amount == allocationB[2].amount, 'p2.amt: no sig');
 
             // slice off the two party outcome and data
             VariablePart memory a2 = VariablePart(
--- a/packages/nitro-protocol/contracts/examples/TwoOfThree.sol
+++ b/packages/nitro-protocol/contracts/examples/TwoOfThree.sol
@@ -52,12 +52,12 @@ contract TwoOfThree is IForceMoveApp2 {
             );
 
             // slice off the two party outcome and data
-            a.appData = aAppData.twoPartyAppData;
-            b.appData = bAppData.twoPartyAppData;
+            a.appData = abi.encode(aAppData.twoPartyAppData);
+            b.appData = abi.encode(bAppData.twoPartyAppData);
             a.outcome = encode2PartyAllocation(allocationA[0], allocationA[1]);
             b.outcome = encode2PartyAllocation(allocationB[0], allocationB[1]);
 
-            return bAppData.twoPartyApp.validTransition(a, b, turnNumB, nParticipants, signedBy);
+            return bAppData.twoPartyApp.validTransition(a, b, turnNumB, 2, signedBy);
         } else return false;
     }
 
--- a/packages/nitro-protocol/contracts/examples/TwoOfThree.sol
+++ b/packages/nitro-protocol/contracts/examples/TwoOfThree.sol
@@ -19,7 +19,7 @@ contract TwoOfThree is IForceMoveApp2 {
         VariablePart memory a, // TODO consider less confusing variable names
         VariablePart memory b,
         uint48 turnNumB,
-        uint256 nParticipants,
+        uint256,
         uint8 signedBy // Who has signed state b?
     ) public override pure returns (bool) {
         // If the new state is signed by all three participants, it is automatically considered a valid transition
--- a/packages/nitro-protocol/contracts/examples/TwoOfThree.sol
+++ b/packages/nitro-protocol/contracts/examples/TwoOfThree.sol
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.4;
+pragma experimental ABIEncoderV2;
+
+import '../interfaces/IForceMoveApp.sol';
+import '../Outcome.sol';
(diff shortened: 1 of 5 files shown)


packages/nitro-protocol/contracts/interfaces/IForceMoveApp.sol
```diff
@@ -27,3 +27,40 @@ interface IForceMoveApp {
...
+    ) external pure returns (bool);
+    // signedBy
+    // 0b000 = 0 : No-one
+    // 0b001 = 1 : participant 0
+    // 0b010 = 2 : participant 1
+    // 0b011 = 3 : participant 0 and participant 1
+    // 0b100 = 4 : participant 2
+    // 0b101 = 5 : participant 0 and participant 2
+    // 0b110 = 6 : participant 1 and participant 2
+    // 0b111 = 7 : everyone
```
geoknee, line 65: This interface is identical to the existing `IForceMoveApp`, save for the addition of the `signedBy` argument. It's a bit field which declares which participants have signed the state being transitioned to. I'm assuming here that we will verify this field against the supplied signatures at an outer level of the protocol.

packages/nitro-protocol/contracts/examples/TwoOfThree.sol
```diff
@@ -0,0 +1,122 @@
...
+        VariablePart memory a, // TODO consider less confusing variable names
+        VariablePart memory b,
+        uint48 turnNumB,
+        uint256 nParticipants,
+        uint8 signedBy // Who has signed state b?
+    ) public override pure returns (bool) {
+        // If the new state is signed by all three participants, it is automatically considered a valid transition
+        if (signedBy == 7) return true;
+        if (signedBy == 0) return false;
+        if (signedBy < 4) {
```
geoknee, line 28: The joint (3 party channel) behaves a bit like the null app: unanimous signing implies a supported state (here, it implies a valid transition).

When participant 2 (the intermediary) has not signed, we fall through to another 2 party `ForceMoveApp.validTransition()`. That "inner" valid transition function can define any desired turn taking rules, by using `signedBy` and `turnNumB`.

packages/nitro-protocol/contracts/interfaces/IForceMoveApp.sol
andrewgordstewart: The signedBy here seems to be specific to 3-party channels. Am I interpreting this correctly?

packages/nitro-protocol/contracts/interfaces/IForceMoveApp.sol
```diff
@@ -27,3 +27,40 @@ interface IForceMoveApp {
...
+    }
+
+    /**
+     * @notice Encodes application-specific rules for a particular ForceMove-compliant state channel.
+     * @dev Encodes application-specific rules for a particular ForceMove-compliant state channel.
+     * @param a State being transitioned from.
+     * @param b State being transitioned to.
+     * @param turnNumB Turn number being transitioned to.
+     * @param nParticipants Number of participants in this state channel.
+     * @param signedBy Bit field showing which participants signed state b.
```
andrewgordstewart, line 47: This `validTransition` function is awkward. It is hard (impossible?) to support things like:
- From an ABC state, Alice can move to an A state, and Bob can move to a B state
- From an A state, Bob can move to an AB state
- From a B state, Alice can move ton an AB state

packages/nitro-protocol/contracts/examples/TwoOfThree.sol
```diff
@@ -0,0 +1,119 @@
...
+            VariablePart memory a2 = VariablePart(
+                encode2PartyAllocation(allocationA[0], allocationA[1], aAssetHolderAddress),
+                aAppData.twoPartyAppData
+            );
+            VariablePart memory b2 = VariablePart(
+                encode2PartyAllocation(allocationB[0], allocationB[1], bAssetHolderAddress),
+                bAppData.twoPartyAppData
+            );
+
+            return bAppData.twoPartyApp.validTransition(a2, b2, turnNumB, 2, signedBy);
```
andrewgordstewart, line 67: This validates the updated app data. How does one force an update in the _outcome_ of `J`, based on some update in the embedded `twoPartyApp`?

andrewgordstewart: I'm a bit confused here, but this implementation seems to couple app updates & joint channel updates. Why is `turnNumB` passed to the `twoPartyApp`'s valid transition function? How do Alice & Bob update the app channel with larger and larger turn numbers, without signing updates in the joint channel?

packages/nitro-protocol/contracts/interfaces/IForceMoveApp.sol
geoknee: It's only specific by dint of being a `uint8`, which is easily relaxed. Then, it should be able to handle more parties naturally.

packages/nitro-protocol/contracts/interfaces/IForceMoveApp.sol
geoknee: I wasn't sure what you were driving at here until the extra context of #3552.

packages/nitro-protocol/contracts/examples/TwoOfThree.sol
geoknee: The idea was to have no distinction between the app channel and the joint channel: so that Alice and Bob are always signing updates to the joint channel.

This is a bad design, however, as it doesn't allow Irene to force the channel to conclude in all cases.

geoknee:
Closing in favour of an approach that follows #3552

packages/nitro-protocol/contracts/examples/TwoOfThree.sol
andrewgordstewart: Ahhhh. That makes sense!

PR #3546, opened 2021-05-25
Remove all events from engine

lalexgap:
# Description
Removes the `ChannelUpdated` event from the engine and thus stops the `engine` from being an `EventEmitter`.

This means the `engine` is a simple request/response model that is much easier to deal with.

This causes the directly funded chain test to fail, so it has been skipped for now and will be re-enabled after the chain service is removed from the engine. (See https://github.com/statechannels/statechannels/issues/3548, https://github.com/statechannels/statechannels/issues/3549 )

---
### Code quality
### Project management
- [ ] I have [linked to all dependent issues](https://help.zenhub.com/support/solutions/articles/43000010349-create-github-issue-dependencies) (can be 0)

### DIFF
--- a/packages/interop-tests/src/__test__/interop.test.ts
+++ b/packages/interop-tests/src/__test__/interop.test.ts
@@ -2,7 +2,6 @@ import {
   DBAdmin,
   defaultTestConfig,
   Output,
-  SingleChannelOutput,
   SingleThreadedEngine
 } from '@statechannels/server-wallet';
 import {ETHERLIME_ACCOUNTS} from '@statechannels/devtools';
@@ -11,25 +10,20 @@ import {constants, Contract, providers, Wallet} from 'ethers';
 import {
   Address,
   BN,
-  deserializeMessage,
   Destination,
   formatAmount,
   makeAddress,
   makeDestination,
   Uint256
 } from '@statechannels/wallet-core';
-import {fromEvent} from 'rxjs';
-import {first} from 'rxjs/operators';
 import {ContractArtifacts} from '@statechannels/nitro-protocol';
 import _ from 'lodash';
-
 import {
   CreateChannelParams,
   isJsonRpcNotification,
   Message,
   PushMessageRequest
 } from '@statechannels/client-api-schema';
-import {Message as WireMessage} from '@statechannels/wire-format';
 
 jest.setTimeout(60_000);
 
@@ -133,15 +127,6 @@ function generatePushMessage(messageParams: Message): PushMessageRequest {
   };
 }
 
-// In theory, we should be able to check the channelResult. In practice, the channelResult seems to have an incorrect turn number
-function containsPostfundState(singleChannelOutput: SingleChannelOutput): boolean {
-  if (!singleChannelOutput.outbox.length) return false;
-
-  const signedStates = deserializeMessage(singleChannelOutput.outbox[0].params as WireMessage)
-    .signedStates;
-  return signedStates ? signedStates?.some(ss => ss.turnNum === 3) : false;
-}
-
 it('server wallet creates channel + cooperates with browser wallet to fund channel', async () => {
   const output1 = await serverWallet.createChannel({
     appData: '0x',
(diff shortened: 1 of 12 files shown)


packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -104,7 +104,8 @@ afterAll(async () => {
   provider.polling = false;
 });
 
-it('Create a directly funded channel between two engines ', async () => {
+// TODO: This should be re-enabled once the chain service has been decoupled from the engine
```
lalexgap, line 107: This should be updated and re-enabled when https://github.com/statechannels/statechannels/issues/3548 is completed

packages/interop-tests/src/__test__/interop.test.ts
```diff
@@ -94,17 +89,18 @@ beforeEach(async () => {
 
   browserWallet.onSendMessage(message => {
     if (isJsonRpcNotification(message)) {
+      // TODO: Since we're not awaiting this it can execute while knex is being destroyed
```
andrewgordstewart, line 92: Why not await it?

packages/interop-tests/src/__test__/interop.test.ts
```diff
@@ -186,14 +171,6 @@ it('server wallet creates channel + cooperates with browser wallet to fund chann
...
     'dummyDomain'
   );
-
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const postFundA = await fromEvent<SingleChannelOutput>(serverWallet as any, 'channelUpdated')
-    .pipe(first(containsPostfundState))
-    .toPromise();
-
-  serverWallet.on('channelUpdated', e => console.log(JSON.stringify(e)));
-  await browserWallet.pushMessage(serverMessageToBrowserMessage(await postFundA), 'dummyDomain');
```
andrewgordstewart, line 196: Re. 4416ef139705671252c01782d15732dd6aac579b

This commit seems like it's changing the interop test to _no longer check that the direct funding protocol works_. Is that correct?

That definitely makes me feel uneasy!

packages/interop-tests/src/__test__/interop.test.ts
```diff
@@ -62,20 +62,21 @@ let serverDestination: Destination;
...
+beforeAll(async () => {
   provider = new providers.JsonRpcProvider(rpcEndpoint);
   assetHolderContract = new Contract(
     ethAssetHolderAddress,
     ContractArtifacts.EthAssetHolderArtifact.abi,
     provider
   );
   mineOnEvent(assetHolderContract);
+  // TODO: The onSendMessage listener can still be executing
+  // so we can't properly restart the engine
```
andrewgordstewart, line 74: What makes it so that `onSendMessage` can still be executing? Should it be, after a test case is finished? It seems to me like a test case is _not finished_, if the browser is still sending a message to the server.

andrewgordstewart (commented):
I feel pretty wary of intentionally breaking the _most important feature_ of a wallet in the `master` branch.

Does it make sense to _not_ merge this branch `ag/remove-all-events`, but instead leave it as the "development" branch for your pitch until directly funded channels are supported in that branch?

packages/interop-tests/src/__test__/interop.test.ts
lalexgap: The problem is that we're not `await`ing pushMessage so we're not waiting for the  `pushMessage` to complete.

So the browser wallet can try and send a message **before** the test completes (and the server wallet is destroyed) that continues executing **after** the test completes.

This is because the `engine.pushMessage` is triggered in the listener. This means the test can't be smart enough to `await` the completion of `pushMessage` before allowing things to be destroyed.

packages/interop-tests/src/__test__/interop.test.ts
lalexgap: This PR does break direct funding by removing the `ChannelUpdated` event, which should get restored in #3549.

I'll create an issue for this test as well.

lalexgap: > I feel pretty wary of intentionally breaking the _most important feature_ of a wallet in the `master` branch.
> 
> Does it make sense to _not_ merge this branch `ag/remove-all-events`, but instead leave it as the "development" branch for your pitch until directly funded channels are supported in that branch?

I agree that a feature branch would make sense here since I am breaking functionality here.

I'm going to create a branch `eventless-engine` that will serve as my feature branch. I'll still raise PRs for review into it but we can also do a review of all the changes at the end as well.

packages/interop-tests/src/__test__/interop.test.ts
lalexgap: I think this [comment](https://github.com/statechannels/statechannels/pull/3546#discussion_r641679623) covers it

packages/interop-tests/src/__test__/interop.test.ts
andrewgordstewart: Is there a test case where we want the test to finish before the messaging has finished for some protocol?

I'm sure there could be, but my gut says that all messages should be resolved before the test is finished.

lalexgap:
@andrewgordstewart I'm going to merge this into my feature branch. It's starting to get stale and I think I've addressed most of the concerns?

andrewgordstewart:
Sounds good!

PR #3555, opened 2021-05-28
PR1: Move chain-service to wallet

lalexgap:
Moves the `chain-service` from the `wallet` to the `engine`. This allows the engine to properly work without emitting events. 

The `engine` no longer has a `chain-service` and instead returns a collection of `ChainRequest`s. The `ChainRequest` is a new type that represents a request from the engine to interact with the chain. The `wallet` can then take the chain requests and call the chain service.

The `wallet` now subscribes to the `chain-service` events with its own handlers.
Each  handler follows the same basic pattern:
- Update the store with the results from the chain event
- Crank the engine to see if there are any new messages, chain requests, or objective status changes.
- Send any messages, process any transactions requested, and emit any events for objective status changed.

PR #3557, opened 2021-06-01
PR3: Direct funding test using chain and two wallets

lalexgap:
**based on top of #3564 and #3555** 

Adds a very basic test that ensures direct funding works between two wallets using a chain service running against a real ganache.

Currently, the test creates and then closes a channel. It ensures that the wallet promises resolve successfully and that the funds on ganache get updated correctly.

### DIFF
--- a/packages/server-wallet/jest/chain-setup.ts
+++ b/packages/server-wallet/jest/chain-setup.ts
@@ -1,10 +1,18 @@
 /* eslint-disable no-process-env */
+import * as fs from 'fs';
+
 import {ETHERLIME_ACCOUNTS, GanacheServer} from '@statechannels/devtools';
 import {utils} from 'ethers';
 
 import {deploy} from '../deployment/deploy';
-
+export const ARTIFACTS_DIR = '../../artifacts';
 export default async function setup(): Promise<void> {
+  try {
+    fs.mkdirSync(ARTIFACTS_DIR);
+  } catch (err) {
+    if (err.message !== "EEXIST: file already exists, mkdir '../../artifacts'") throw err;
+  }
+
   if (process.env.CHAIN_NETWORK_ID) {
     console.log(
       `CHAIN_NETWORK_ID defined as ${process.env.CHAIN_NETWORK_ID}. Assuming chain env vars are set by caller`
--- a/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
+++ b/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
@@ -15,6 +15,7 @@ import {TestMessageService} from '../message-service/test-message-service';
 import {SyncOptions, Wallet} from '../wallet';
 import {ONE_DAY} from '../__test__/test-helpers';
 import {waitForObjectiveProposals} from '../__test-with-peers__/utils';
+import {ARTIFACTS_DIR} from '../../jest/chain-setup';
 
 jest.setTimeout(30_000);
 
@@ -39,7 +40,7 @@ let a: Wallet;
 let b: Wallet;
 let aEngine: Engine;
 let bEngine: Engine;
-const ARTIFACTS_DIR = '../../artifacts';
+
 const bEngineConfig: EngineConfig = {
   ...overwriteConfigWithDatabaseConnection(config, {database: 'server_wallet_test_b'}),
   loggingConfiguration: {
--- a/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
+++ b/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
@@ -1,16 +1,20 @@
+import path from 'path';
+
 import {CreateChannelParams, Participant, Allocation} from '@statechannels/client-api-schema';
 import {ETHERLIME_ACCOUNTS} from '@statechannels/devtools';
 import {ContractArtifacts} from '@statechannels/nitro-protocol';
 import {BN, makeAddress, makeDestination} from '@statechannels/wallet-core';
 import {BigNumber, constants, Contract, ethers, providers} from 'ethers';
 import _ from 'lodash';
-import {fromEvent} from 'rxjs';
-import {take} from 'rxjs/operators';
 
+import {ChainService} from '../chain-service';
 import {defaultTestConfig, overwriteConfigWithDatabaseConnection, EngineConfig} from '../config';
 import {DBAdmin} from '../db-admin/db-admin';
-import {Engine, SingleChannelOutput} from '../engine';
-import {getChannelResultFor, getPayloadFor, ONE_DAY} from '../__test__/test-helpers';
+import {Engine} from '../engine';
+import {TestMessageService} from '../message-service/test-message-service';
+import {SyncOptions, Wallet} from '../wallet';
+import {ONE_DAY} from '../__test__/test-helpers';
+import {waitForObjectiveProposals} from '../__test-with-peers__/utils';
 
 // eslint-disable-next-line no-process-env, @typescript-eslint/no-non-null-assertion
 const ethAssetHolderAddress = makeAddress(process.env.ETH_ASSET_HOLDER_ADDRESS!);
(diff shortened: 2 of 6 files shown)


packages/server-wallet/jest/chain-setup.ts
```diff
@@ -1,10 +1,18 @@
...
+import * as fs from 'fs';
+
 import {ETHERLIME_ACCOUNTS, GanacheServer} from '@statechannels/devtools';
 import {utils} from 'ethers';
 
 import {deploy} from '../deployment/deploy';
-
+export const ARTIFACTS_DIR = '../../artifacts';
 export default async function setup(): Promise<void> {
+  try {
```
lalexgap, line 10: This allows chain tests to write out logs to the artifact folder for circleCI

packages/server-wallet/jest/chain-setup.ts
```diff
@@ -1,10 +1,18 @@
...
+    fs.mkdirSync(ARTIFACTS_DIR);
+  } catch (err) {
+    if (err.message !== "EEXIST: file already exists, mkdir '../../artifacts'") throw err;
```
kerzhner, line 13: Does this prevent multiple, local runs of the chain tests?

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -76,43 +93,64 @@ function mineOnEvent(contract: Contract) {
...
+
+    TestMessageService.linkMessageServices(a.messageService, b.messageService, aEngine.logger);
+    const assetHolder = new Contract(
+      ethAssetHolderAddress,
+      ContractArtifacts.EthAssetHolderArtifact.abi,
+      provider
+    );
+    mineOnEvent(assetHolder);
+  } catch (error) {
+    console.error('beforeAll failed');
```
kerzhner, line 134: Should we use a logger instead of `console`?

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -139,143 +177,31 @@ it.skip('Create a directly funded channel between two engines ', async () => {
...
-  const resultB0 = await b.pushMessage(getPayloadFor(participantB.participantId, preFundA.outbox));
+  const response = await a.createChannels([channelParams]);
+  await waitForObjectiveProposals([response[0].objectiveId], b);
+  const bResponse = await b.approveObjectives([response[0].objectiveId]);
 
-  expect(getChannelResultFor(channelId, resultB0.channelResults)).toMatchObject({
-    status: 'proposed',
-    turnNum: 0,
-  });
+  await expect(response).toBeObjectiveDoneType('Success');
```
kerzhner, line 188: Does this work consistently? If so, I am a bit surprised!

I could be misreading the code, but it seems to me that we are only waiting for the objective approval, not objective succeeding.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -139,143 +177,31 @@ it.skip('Create a directly funded channel between two engines ', async () => {
...
-
-  const channelClosedA = await channelClosedAPromise;
-
-  expect(channelClosedA.channelResult).toMatchObject({
-    status: 'closed',
-    turnNum: 4,
-    fundingStatus: 'Defunded',
-  });
+  const {channelId} = response[0];
+  const closeResponse = await b.closeChannels([channelId]);
```
kerzhner, line 195: Do we need to wait for the objective to succeed?

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -139,143 +177,31 @@ it.skip('Create a directly funded channel between two engines ', async () => {
...
+  await expect(closeResponse).toBeObjectiveDoneType('Success');
 
   const aBalanceFinal = await getBalance(aAddress);
   const bBalanceFinal = await getBalance(bAddress);
 
   expect(BN.sub(aBalanceFinal, aBalanceInit)).toEqual(aFunding);
```
kerzhner, line 201: It might be worth it to set `aFunding` and `bFunding` to something more interesting that 1 and 0. Maybe 2 and 3? This would force `b` to make a deposit.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -225,7 +224,8 @@ export class Wallet extends EventEmitter<WalletEvents> {
       channelIds.map(async channelId => {
         const closeResult = await this._engine.closeChannel({channelId});
         const {newObjective, channelResult} = closeResult;
-
+        // create the promise before we send anything out
```
kerzhner, line 227: Curious, why is it important create a promise here instead of below, in the return dictionary declaration?

kerzhner (approved):
Thanks for putting this together!

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
lalexgap: This should be waiting for the objective to complete fully.

In the result from both `approveObjectives` and `createChannels` we return a `done` promise that resolves when the objective is completed.

The idea is that you get a result right away with a channelId/ObjectiveId  and a `promise`  that will resolve when the objective is completed.

packages/server-wallet/src/wallet/wallet.ts
lalexgap: A few lines lower we call `await this.handleEngineOutput(closeResult)` which will emit events for objectives in `closeResult`.  If `this.handleEngineOutput(closeResult)` is called before we setup the promise then the objective event may be emitted **before** we construct the promise that resolves when the event is emitted.

That means we can end up with a promise that never resolves even though the objective is complete.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
lalexgap: Yup, this is just some debugging code that snuck in.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
lalexgap: I did this in f42a5a9 which broke the tests! Thanks for bringing this up, investigating now.

lalexgap: Ah it looks like it was due to the fact that we only `mine` blocks on the `Deposited` event. Since the `Deposited` event only fires when depositing funds, there's nothing to trigger mining a bunch of blocks when conclude is called.

   I'm guessing that with only one deposit the mining from the `Deposited` event was close enough to mine the conclude transaction as well so that's why we didn't see this with `aFunding=1, bFunding=0`.

This is fixed in https://github.com/statechannels/statechannels/pull/3557/commits/6f304b73de94b1ee13b763ae5d8095cad0e50218 by mining on the allocationUpdated event as well.

packages/server-wallet/jest/chain-setup.ts
lalexgap: Nope! It attempts to create the directory and if it fails due to it existing (which would happen when running again) it swallows the error.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
kerzhner: If there was a single deposited event, in theory, the test should have failed. If I recall correctly, the test is set up to mine the minimum number of blocks for the test to pass.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
kerzhner: Got it! Thanks for the explanation!

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
kerzhner: Resolved above.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
lalexgap: > If there was a single deposited event, in theory, the test should have failed. If I recall correctly, the test is set up to mine the minimum number of blocks for the test to pass.

Ah you're right!

I think the reason it was passing was because `b` was the one calling `closeChannel`. Since it had `0` funding it would succeed as soon as the conclude proof was signed.

lalexgap: I've added some test cases in https://github.com/statechannels/statechannels/pull/3557/commits/40c47f5a0e3374fc5fbda10668010e4b66bc6e47 so we now test both A and B closing. I also threw in some cases with messages being dropped / delayed.

packages/server-wallet/src/wallet/wallet.ts
kerzhner: Thanks for the explanation.

PR #3558, opened 2021-06-02
Use hardhat network instead of ganache for gas benchmarks (and upgrade to berlin hardfork)

geoknee:
~⚠️ base not master~

Ganache does not yet support the[ berlin hardfork](https://blog.ethereum.org/2021/03/08/ethereum-berlin-upgrade-announcement/), whereas hardhat does. We already have hardhat installed, so swapping ganache out and [hardhat network](https://hardhat.org/hardhat-network/) in, is relatively easy. 

Note the changes in gas consumption flagged by the upgrade.


lalexgap (approved):
👍

PR #3559, opened 2021-06-02
embed an X in a J (2nd attempt)

geoknee:
https://www.notion.so/statechannels/Virtual-funding-research-11388f704f84496ba8febe30ddb522d2


geoknee:
In favour of #3565 (cleaner history)

PR #3562, opened 2021-06-02
Feature: Eventless engine

lalexgap:
This stops the engine being a `EventEmitter`.  The `engine` now follows a basic request/response model. 

To accomplish this there were a couple of large changes:

### The chain service was moved from the engine to the wallet.
The `wallet` now handles events from the  chain service. When an event occurs the `wallet`
- updates the store
- cranks the engine
- handles the response (send messages, etc..)
- If the `engine` wants to send a transaction to chain it returns a `ChainRequest`. The `wallet` is responsible for handling those chain requests by using the chain service.

### The Wallet become an `EventEmitter`
The wallet now emits a couple of objective related events (`ObjectiveCompleted`, `ObjectiveProposed`, `ObjectiveTimedOut`). These are used internally to know when a objective has been completed and are also available externally.

### The Wallet retry/sync logic was refactored 
Once the wallet was emitting events for objectives it was much simpler to construct promises and implement the retry logic.

Now the retry logic follows a simple poll based pattern. Every so often it polls for objectives that have not made progress recently. If it finds any objectives like that it sends the sync messages for them. If the objectives don't make progress after a certain amount of time they are marked as failed.

### Engine now exposes a crank method
The engine now exposes a method for cranking objectives. It is used after the wallet updates the store due to a chain event, to make progress on objectives.

 ### Remove of some tests
Some previous tests were designed around the `engine` and `chain-service` working together. [These have been removed](https://github.com/statechannels/statechannels/pull/3562/commits/240607593a73fe0e9329b758b95301798fffe927). Some of them are obsolete but I think we will want to write a `challenge` and `ledger funding` test once we have that functionality in the `wallet`.


packages/interop-tests/src/__test__/interop.test.ts
```diff
@@ -67,22 +61,22 @@ let serverAddress: Address;
...
-beforeAll(() => {
+beforeAll(async () => {
   provider = new providers.JsonRpcProvider(rpcEndpoint);
   assetHolderContract = new Contract(
     ethAssetHolderAddress,
     ContractArtifacts.EthAssetHolderArtifact.abi,
     provider
   );
   mineOnEvent(assetHolderContract);
+  // TODO: The onSendMessage listener can still be executing
```
lalexgap, line 73: I think this `interop` test should probably be rewritten using the server `wallet`.

packages/server-wallet/e2e-test/challenge.test.ts
```diff
@@ -55,7 +55,9 @@ afterAll(async () => {
   await knexPayer.destroy();
 });
 
-test('the engine handles the basic challenging v0 behavior', async () => {
+// TODO: This test relies on the engine being tightly coupled to the chain service.
+// This should be refactored to use the wallet once the challenge method is exposed there.
+test.skip('the engine handles the basic challenging v0 behavior', async () => {
```
lalexgap, line 60: We should probably create an issue for this.

andrewgordstewart: Can you create issues for the lost test coverage? The skipped test here is about challenging. On this branch, we now have a great directly funded channel test, but challenging isn't tested e2e

andrewgordstewart: that is, _assuming_ we merge #3587

packages/interop-tests/src/__test__/interop.test.ts
```diff
@@ -193,16 +173,6 @@ it('server wallet creates channel + cooperates with browser wallet to fund chann
...
-
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const postFundA = await fromEvent<SingleChannelOutput>(serverWallet as any, 'channelUpdated')
-    .pipe(first(containsPostfundState))
-    .toPromise();
-
-  serverWallet.on('channelUpdated', e => console.log(JSON.stringify(e)));
-  await browserWallet.pushMessage(serverMessageToBrowserMessage(await postFundA), 'dummyDomain');
-
-  await objectiveSuccededPromise;
```
andrewgordstewart, line 205: What assertion should be made at the end of this test?

packages/server-wallet/server-wallet.api.md
```diff
@@ -167,10 +161,6 @@ export interface EngineInterface {
     // (undocumented)
     pushUpdate(m: unknown): Promise<SingleChannelOutput>;
     // (undocumented)
-    registerAppBytecode(appDefinition: string, bytecode: string): Promise<void>;
```
andrewgordstewart, line 170: This removal needs to be addressed.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -86,9 +104,26 @@ beforeAll(async () => {
...
+    logger: bEngine.logger,
+  });
+
+  const syncOptions: SyncOptions = {
+    pollInterval: 1_000,
+    timeOutThreshold: 60_000,
+    staleThreshold: 10_000,
+  };
+  a = await Wallet.create(aEngine, aChainService, TestMessageService.create, syncOptions);
+  b = await Wallet.create(bEngine, bChainService, TestMessageService.create, syncOptions);
```
andrewgordstewart, line 124: do you see this being extracted as "common test setup" code?

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -100,181 +135,88 @@ beforeAll(async () => {
...
+    const assetHolderBalanceUpdated = await getBalance(ethAssetHolderAddress);
+    expect(BN.sub(assetHolderBalanceUpdated, assetHolderBalanceInit)).toEqual(
+      BN.add(aFunding, bFunding)
+    );
+
+    const {channelId} = response[0];
+    const closeResponse =
+      options.closer === 'A'
+        ? await a.closeChannels([channelId])
+        : await b.closeChannels([channelId]);
```
andrewgordstewart, line 213: For future consideration, I like the pattern where we have something like a `peers: { A: Wallet, B: Wallet, all: [Wallet, Wallet]}` object, such that we can do
```suggestion
    const closeResponse = peers[options.closer].closeChannels([channelId]);
```

or
```
await Promise.all(peers.all.map(w => w.destroy()));
```

packages/server-wallet/src/__test-with-peers__/crash-tolerance/crash-tolerance.test.ts
```diff
@@ -1,127 +0,0 @@
-import {
```
andrewgordstewart, line 1: Just to clarify, this test case is now covered by the cases
```
  {dropRate: 0.1, meanDelay: 50, closer: 'A'},
  {dropRate: 0.1, meanDelay: 50, closer: 'B'},
```
in the direct funding test?

packages/server-wallet/src/engine/store.ts
```diff
@@ -357,6 +357,11 @@ export class Store {
     const results = await ObjectiveModel.query(this.knex).where({status: 'approved'});
     return results.map(o => o.toObjective());
   }
+  public async updateObjectiveProgressDate(objectiveIds: string[]): Promise<WalletObjective[]> {
```
andrewgordstewart, line 360: Is this method ever used?

packages/server-wallet/src/engine/store.ts
```diff
@@ -428,6 +433,11 @@ export class Store {
     return await ObjectiveModel.forId(objectiveId, tx);
   }
 
+  async isObjectiveComplete(objectiveId: string): Promise<boolean> {
```
andrewgordstewart, line 436: is this method ever used?

packages/server-wallet/server-wallet.api.md
lalexgap: Right now the wallet exposes a `registerAppDefinition` which can be used to register bytecode for an app.

packages/server-wallet/src/__test-with-peers__/crash-tolerance/crash-tolerance.test.ts
lalexgap: I think the `crash-tolerance`  test is slightly different since it is focusing on "restarting" the engine and make sure it can keep on executing. The test cases in this PR are just testing messages being dropped/delayed.

packages/server-wallet/src/engine/store.ts
lalexgap: [Yup](https://github.com/statechannels/statechannels/blob/eventless-engine/packages/server-wallet/src/wallet/wallet.ts#L254:L254).

packages/server-wallet/server-wallet.api.md
andrewgordstewart: ohh -- this is the API change for the _engine_. I thought that it was for the _wallet_, and therefore believed that `registerAppDefinition` was removed from the wallet API.

packages/server-wallet/src/engine/store.ts
lalexgap: Nope, I think this can be removed!

packages/server-wallet/src/engine/store.ts
andrewgordstewart: I thought my vs-code editor failed to find it. It must have been either user-error or a glitch.

packages/server-wallet/src/protocols/direct-funder.ts
```diff
@@ -59,12 +58,15 @@ export class DirectFunder {
     // otherwise, deposit
     const amountToDeposit = BN.sub(targetAfter, currentAmount); // previous checks imply this is >0
     await ChainServiceRequest.insertOrUpdate(channel.channelId, 'fund', tx);
-    await this.chainService.fundChannel({
-      channelId: channel.channelId,
-      assetHolderAddress: assetHolderAddress,
-      expectedHeld: BN.from(currentAmount),
-      amount: BN.from(amountToDeposit),
-    });
+    await response.queueChainRequest([
```
andrewgordstewart, line 61:
```suggestion
    response.queueChainRequest([
```

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
lalexgap: Yeah, I think that's pretty likely. Right now this setup is very similar to our "with peers" [tests](https://github.com/statechannels/statechannels/blob/eventless-engine/packages/server-wallet/jest/with-peers-setup-teardown.ts#L89:L89).

I think we should organize our tests a bit since right now we have:
- "with peers" tests that use **two** databases and a **mock** chain
- "direct-funding" tests that uses **two** databases and a **real** chain
- "chain" tests that use **one** database and a **real** chain
-  regular tests that use **one** and a **mock** chain

packages/interop-tests/src/__test__/interop.test.ts
lalexgap: Right now this test won't work since it's using `ganache` and just an `engine`.

 I'm going to see if I can quickly update it to work with a `wallet` . If not I'll revert my changes to this test and just skip it for now (and create an issue).

packages/server-wallet/src/protocols/direct-funder.ts
lalexgap: Yup! 

Unfortunately, I can't seem to find any lint rule to catch unecessary awaits :(

packages/server-wallet/src/engine/store.ts
```diff
@@ -357,6 +357,11 @@ export class Store {
     const results = await ObjectiveModel.query(this.knex).where({status: 'approved'});
     return results.map(o => o.toObjective());
   }
+  public async updateObjectiveProgressDate(objectiveIds: string[]): Promise<WalletObjective[]> {
+    return this.knex.transaction(async tx =>
+      Promise.all(objectiveIds.map(oId => ObjectiveModel.updateProgressDate(oId, tx)))
```
andrewgordstewart, line 362: This seems like it would be an issue for "busy wallets" -- we'd fire off many queries that each update a single row.

Referring to [this code](https://github.com/statechannels/statechannels/blob/eventless-engine/packages/server-wallet/src/wallet/wallet.ts#L252-L256), there are `2N + 2` queries required to `jumpStartObjectives` when there are `N` approved objectives. The `N` queries triggered on this line will likely fail for some modest number `N` (eg. `N = 1000`) with knex timeout issues.

It also feels like there's a concurrency issue here: `updateObjectiveProgressDate` may update the objective concurrent with a `pushMessage` grabbing the objective or something. It's hard to think through the implication of that.

With regards to this PR, it's not **too** hard to get it down to `N+1` queries, by replacing
```
    const objectives = await this._engine.getApprovedObjectives();
    const objectiveIds = objectives.map(o => o.objectiveId);
    await this._engine.store.updateObjectiveProgressDate(objectiveIds);
```
with something like
```
    const objectives = await this._engine.getApprovedObjectives(updateProgressData=true);
```
that returns the single `getApprovedObjectives` query and the `N` queries in
```
       Promise.all(objectiveIds.map(oId => ObjectiveModel.updateProgressDate(oId, tx)))
```

I don't think you need to make that change within this PR, because `2N+2` queries and `N+1` queries are just about as bad as one another.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -57,15 +138,24 @@ export class Wallet {
...
     );
+
+    const results = objectives.map(async o => ({
+      objectiveId: o.objectiveId,
+      currentStatus: o.status,
+      channelId: o.data.targetChannelId,
+      done: this.createObjectiveDoneResult(o),
+    }));
+
+    // TODO: ApproveObjective should probably just return a MultipleChannelOuput
```
andrewgordstewart, line 152: Why is there an `approveObjectives` method, instead of just a `approveObjective` method?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -26,26 +42,91 @@ export class Wallet {
...
 
     this._messageService = messageServiceFactory(handler);
+
+    this._chainListener = {
+      holdingUpdated: this.createChainEventlistener('holdingUpdated', e =>
+        this._engine.store.updateFunding(e.channelId, e.amount, e.assetHolderAddress)
+      ),
+      assetOutcomeUpdated: this.createChainEventlistener('assetOutcomeUpdated', async e => {
+        const transferredOut = e.externalPayouts.map(ai => ({
+          toAddress: makeDestination(ai.destination),
```
andrewgordstewart, line 90: Why is this keyed `toAddress` and not `destination`?

(and why is `ai.destination` not a `Destination`?)

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -136,112 +251,138 @@ export class Wallet {
...
+  ): Promise<ObjectiveDoneResult> {
+    return new Promise<ObjectiveDoneResult>(resolve => {
+      this.on('ObjectiveTimedOut', o => {
+        if (o.objectiveId === objective.objectiveId) {
+          this._engine.logger.trace({objective: o}, 'Objective Timed out');
+
+          resolve({
+            type: 'ObjectiveTimedOutError',
+            objectiveId: o.objectiveId,
+            lastProgressMadeAt: o.progressLastMadeAt,
```
andrewgordstewart, line 321: nit: this change in key is a bit awkward

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -136,112 +251,138 @@ export class Wallet {
...
+  /**
+   * Registers any channels existing in the database with the chain service.
+   *
+   * @remarks
+   * Enables the chain service to alert the engine of of any blockchain events for existing channels.
+   *
+   * @returns A promise that resolves when the channels have been successfully registered.
+   */
+  private async registerExistingChannelsWithChainService(): Promise<void> {
+    const channelsToRegister = (await this._engine.store.getNonFinalizedChannels()).map(
```
andrewgordstewart, line 343: unrelated to this PR, but `getNonFinalizedChannels` seems to only return `'Open'` channels. But:
- A channel in the `Challenge` mode is not finalized
- Should we also register channels that are _finalized_ but still have holdings on chain?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -136,112 +251,138 @@ export class Wallet {
...
+
+  private createChainEventlistener<
+    K extends keyof ChainEventSubscriberInterface,
+    EH extends ChainEventSubscriberInterface[K]
+  >(eventName: K, storeUpdater: EH) {
+    return async (
+      event: HoldingUpdatedArg &
+        AssetOutcomeUpdatedArg &
+        ChannelFinalizedArg &
+        ChallengeRegisteredArg
```
andrewgordstewart, line 368: This is a strange type -- I'd expect it to be
```suggestion
      event: HoldingUpdatedArg ||
        AssetOutcomeUpdatedArg ||
        ChannelFinalizedArg ||
        ChallengeRegisteredArg
```

What's the point of the intersection type?

andrewgordstewart (approved):
This seems like a pretty large simplification of the wallet!

packages/interop-tests/src/__test__/interop.test.ts
lalexgap: While I don't think it's going to be too much effort it seems like too much to do it in the scope of this PR. I've just skipped the test for now in a58319e

andrewgordstewart: 👍 Let's keep this PR as small as possible!

packages/server-wallet/src/wallet/wallet.ts
lalexgap: This was copied from the original behaviour in the `engine` so I'm not sure why those choices were made.

packages/server-wallet/src/wallet/wallet.ts
lalexgap: What's the advantage of having `approveObjective` instead of `approveObjectives` ?

packages/server-wallet/src/wallet/wallet.ts
andrewgordstewart: https://github.com/statechannels/statechannels/issues/3594

PR #3564, opened 2021-06-02
PR2: Simplify Message Retry Logic

lalexgap:
**based on top of #3555** 

Simplifies the message retry logic by switching to a simple timer-based approach.

Instead of relying on `ensureObjective`'s exponential delay we periodically check for objectives that have not made progress and send out their messages.

We set up an interval timer to check for any objectives that `lastProgressDate` are "stale" and send out a `sync` message.

If the objective `lastProgressDate` does not get updated after an even longer period of time we consider that objective "timed out". The `wallet` will resolve the objective promise with an `ObjectiveTimedOutError`.

This means we can remove `ensureObjective` and simplify the logic greatly. By emitting events on objective success (see 549a603) we can return a promise that resolves as soon as the objective succeeds. 

**This also means we remove all the queries that `ensureObjective` was doing and any unneeded delays before resolving.**

Note: Currently, we rely only on the`lastProgressDate`, however, we could add a `numberOfAttempts` field to the objective db and use that to determine when we give up.


packages/server-wallet/package.json
```diff
@@ -24,6 +24,7 @@
     "pg-connection-string": "^2.4.0",
     "pino": "6.2.0",
     "rxjs": "6.6.3",
+    "set-interval-async": "^2.0.3",
```
lalexgap, line 27: This just makes test cleanup much easier. We can `await` `clearIntervalAsync` and be sure that the sync method is done before destroying.

packages/server-wallet/src/wallet/types.ts
```diff
@@ -60,6 +61,8 @@ export type ObjectiveResult = {
   channelId: string;
 };
 
-export type ObjectiveProposed = {
-  ObjectiveProposed: WalletObjective;
-};
+export interface WalletEvents {
+  ObjectiveCompleted: (o: WalletObjective) => void;
+  ObjectiveProposed: (o: WalletObjective) => void;
+  ObjectiveTimedOut: (o: WalletObjective) => void;
```
lalexgap, line 67: The new `ObjectiveCompleted` and `ObjectiveTimedOut` are for internal use in the `wallet` but I think it's ok to expose them externally? (And easy to change if needed).

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -168,7 +168,12 @@ it.skip('Create a directly funded channel between two engines ', async () => {
     turnNum: 0,
   });
 
-  const resultB0 = await b.pushMessage(getPayloadFor(participantB.participantId, preFundA.outbox));
+  const resultB0 = await b.pushMessage(
+    getPayloadFor(
+      participantB.participantId,
+      preFundA.outbox.map(o => o.params)
+    )
```
andrewgordstewart, line 175: `getPayloadFor` is a test helper. Can it extract the `params` off of an object when needed? The changeset has a lot of noise because of every `o => o.params` callback.

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -321,6 +331,137 @@ export class Wallet extends EventEmitter<WalletEvents> {
...
+    await this.waitForTransactions(this._chainService.handleChainRequests(output.chainRequests));
+  }
+
+  /**
+   * Creates a listener that will be subscribed to the chain service.
+   * It creates a listener for a specific channel id.
+   * @param channelIdToListenFor The channel Id the listener is for
+   * @returns A chain event subscriber that can be passed into the chain service.
+   */
+  private createChainListener(channelIdToListenFor: string): ChainEventSubscriberInterface {
```
andrewgordstewart, line 308: nit: it would have been nice to move the chain listener construction to the wallet in its own commit, to facilitate reviewing. As is, it is difficult to see which functionality is new, and what is copy/pasted.

andrewgordstewart, line 308: this now creates a new chain listener for each channel. why's that?

it seems like that choice will cause the chain service to call `holdingUpdated` a quadratic number of times, as a function of the number of channels?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -321,6 +331,137 @@ export class Wallet extends EventEmitter<WalletEvents> {
...
+        try {
+          if (channelId !== channelIdToListenFor) return;
+
+          this._engine.logger.trace(
+            {channelId, assetHolderAddress, externalPayouts},
+            'assetOutcomeUpdated'
+          );
+          const transferredOut = externalPayouts.map(ai => ({
+            toAddress: makeDestination(ai.destination),
+            amount: ai.amount as Uint256,
```
andrewgordstewart, line 333: is this type assertion necessary?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -321,6 +331,137 @@ export class Wallet extends EventEmitter<WalletEvents> {
...
+  /**
+   * Emits events, sends messages and requests transactions based on the output of the engine.
+   * @param output
+   */
+  private async handleEngineOutput(
+    output: MultipleChannelOutput | SingleChannelOutput
+  ): Promise<void> {
+    this.emitObjectiveEvents(output);
+    await this._messageService.send(getMessages(output));
+    await this.waitForTransactions(this._chainService.handleChainRequests(output.chainRequests));
```
andrewgordstewart, line 299: can these be done concurrently?
```
await Promise.all([
  this._messageService.send(getMessages(output)),
  await this.waitForTransactions(this._chainService.handleChainRequests(output.chainRequests))
]);
```

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -321,6 +331,137 @@ export class Wallet extends EventEmitter<WalletEvents> {
...
+            'challengeRegistered'
+          );
+          await this._engine.store.insertAdjudicatorStatus(channelId, finalizedAt, challengeStates);
+          const result = await this._engine.crank([channelId]);
+          await this.handleEngineOutput(result);
+        } catch (err) {
+          this._engine.logger.error(err, 'challengeRegistered error');
+          throw err;
+        }
+      },
```
andrewgordstewart, line 362: You're right that these all follow the same pattern. Is there a way to encapsulate that pattern in code?

as an idea:
```
function createHandler(name, inserter) {
  return async (data) => {
    if (data.channelId !== channelIdToListenFor) return;
    this._engine.logger.trace(data, name);
    await inserter(data);
    await this._engine.crank([channelId]);
    await this.handleEngineOutput(result);
  }
}
```

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -240,94 +242,20 @@ export class Wallet extends EventEmitter<WalletEvents> {
...
+      const done = this.createObjectiveSuccessPromise(o);
+      return {
+        objectiveId: o.objectiveId,
+        currentStatus: o.status,
+        channelId: o.data.targetChannelId,
+        done,
+      };
     });
+
+    await this.messageService.send(getMessages(outbox));
```
andrewgordstewart: You may be sending a LOT of messages at this point. Is the message service expected to do batching?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -486,14 +414,25 @@ export class Wallet extends EventEmitter<WalletEvents> {
     objective: WalletObjective
   ): Promise<ObjectiveDoneResult> {
     // TODO: This should resolve to an error
-    return new Promise(resolve =>
+    return new Promise<ObjectiveDoneResult>(resolve => {
+      this.on('ObjectiveTimedOut', o => {
+        console.log(o);
```
andrewgordstewart: ✂️

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -239,8 +239,7 @@ export class Wallet extends EventEmitter<WalletEvents> {
   public async jumpStartObjectives(): Promise<ObjectiveResult[]> {
     const objectives = await this._engine.getApprovedObjectives();
     const objectiveIds = objectives.map(o => o.objectiveId);
-    // Instead of getting messages per objective we just get them all at once
-    // This will prevent us from querying the database for each objective
+    await this._engine.store.updateObjectiveProgressDate(objectiveIds);
```
andrewgordstewart, line 239: This could be a single query? ie. `this._engine.store.updateObjectiveProgressData()` can
- filter by the approved objectives
- update the progress data
- return the data that `syncObjectives` returns

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -247,7 +250,7 @@ export class Wallet extends EventEmitter<WalletEvents> {
       };
     });
 
-    await this.messageService.send(getMessages(outbox));
+    await this.messageService.send(syncMessages);
```
andrewgordstewart, line 253: 🎉

andrewgordstewart (commented):
I read through the commits one by one. The changeset is pretty big, so I think I need to let it digest and return for a review of the entire diff!

packages/server-wallet/src/wallet/wallet.ts
lalexgap: I think you're right. We need to pass in a listener every time we call `registerChannel` but we can re-use the same listener and just check the channel id

packages/server-wallet/src/wallet/wallet.ts
lalexgap: Yup. I was hoping to do something like that but never got around it

packages/server-wallet/src/wallet/wallet.ts
lalexgap: Updated in 19a3e2dbedcd9153b142f49d3f208ce009e6ceb1

packages/server-wallet/src/wallet/wallet.ts
andrewgordstewart: coming back to this, I don't think it's worth making this a single query unless it's easy.

packages/server-wallet/src/wallet/wallet.ts
lalexgap: I've updated this in https://github.com/statechannels/statechannels/pull/3564/commits/2c21498ab50446a9880fb66d48fe9db8df374d5b

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -242,55 +258,67 @@ export class Wallet extends EventEmitter<ObjectiveProposed> {
...
-
-        const messagesForObjective = this.getMessagesForObjective(
-          objective.objectiveId,
-          syncResult
-        );
+        let isComplete = false;
+        this.on('ObjectiveCompleted', (o: WalletObjective) => {
+          if (o.objectiveId == objective.objectiveId && o.status === 'succeeded') {
+            isComplete = true;
+            resolve({type: 'Success', channelId: objective.data.targetChannelId});
```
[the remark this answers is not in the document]
lalexgap: I think this is from an older commit?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -180,6 +184,24 @@ export class Wallet extends EventEmitter<ObjectiveProposed> {
...
+
+    for (const o of timedOutObjectives) {
+      this.emit('ObjectiveTimedOut', o);
+    }
+
+    const staleObjectives = objectives
+      .filter(o => o.progressLastMadeAt.getTime() < staleDate)
+      .map(o => o.objectiveId);
+
+    const syncMessages = await this._engine.syncObjectives(staleObjectives);
```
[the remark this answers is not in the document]
lalexgap: Good point! I've created https://github.com/statechannels/statechannels/issues/3579 for that, as I don't want to increase the scope of this PR.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
lalexgap: https://github.com/statechannels/statechannels/pull/3564/commits/3c1c5ae13281cf595bb1fbd90c60f222a0f356fe

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -117,6 +128,11 @@ export class Wallet extends EventEmitter<ObjectiveProposed> {
     const {objectives, messages, chainRequests} = await this._engine.approveObjectives(
       objectiveIds
     );
+
+    // Emit for any succeed objectives
+    const completedObjectives = objectives.filter(o => o.status === 'succeeded');
+    completedObjectives.forEach(o => this.emit('ObjectiveCompleted', o));
```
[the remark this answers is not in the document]
lalexgap: I ended up removing the `for (const o of completeObjectives) { ... } ` above since it actually wasn't needed!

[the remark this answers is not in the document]
lalexgap: https://github.com/statechannels/statechannels/pull/3564/commits/a1d13c2ba8ac1af8cfa9aba28ae59b2af1311679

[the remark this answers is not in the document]
andrewgordstewart: Even better!

packages/server-wallet/src/wallet/wallet.ts
andrewgordstewart: I can imagine some edge cases where that's still an issue (a hub closes shop, closes all virtual channels, wallet crashes, restarts, and jump starts the objectives). But that doesn't seem worth worrying about

andrewgordstewart (approved):

Issue #3556, opened 2021-06-01
Explore a `refundGuarantee` operation


andrewgordstewart:
# Why: Problem description :grey_question:
Referring to the virtual funding protocol outlined in the Nitro paper, it is possible for Alice & Bob to open a directly funded channel with three rounds of network messages (from Alice's perspective):
1a. Alice signs preFS in `J` & `X`.
1b. Bob signs preFS in `J` & `X`.
1c. Irene signs preFS in `J`.
2a. Alice signs presFS in `G`
2b. Irene signs preFS in `G`, proposes ledger update in `L`
3. Alice co-signs ledger update in `L`.

By removing the guarantor `G`, as described in [Streamlining Virtual Channels](https://www.notion.so/statechannels/Streamlining-Virtual-Channels-8a8650ba849d4221b7e93c125a794ecf), we can reduce this to 2 rounds of network messages:
1a. Alice signs preFS in `J` & `X`.
1b. Bob signs preFS in `J` & `X`.
1c. Irene signs preFS in `J`.
2a. Alice signs a ledger update in `L` to include a guarantee.
2b. Irene co-signs the ledger update from 2a.

Can we reduce this to 1 round of network messages?

# What: Describe what a solution should resolve :grey_exclamation:
Explore ways in which Alice can safely create a funded guarantee (eg. by signing an update in a directly funded ledger channel to include a guarantee) _before holding a supported 

Define the 'exit criteria' for the spike, e.g. whether a required deliverable is to be created (i.e. issue cards, prototypes, POCs) or an implementation decision should be documented (i.e. notion or refining issue cards).

# Describe any ideas you've considered :bulb:
❌  Embedding the app channel `X` in `J`, as per #3552, doesn't help. You would still need to record a supported outcome for `J` on-chain in order to call `claim`.

💡 By including some "refund data" in the guarantee, we could implement a `refundGuarantee` operation that replaces the guarantee outcome item with some outcome items specified by the "refund data" _when the adjudicator knows nothing about the target `J` of the guarantee_. When Alice has a supported outcome for `J`, she can `challenge` in `L` & `J`, followed by a `claim`. If she doesn't, she can `challenge` in `L`, followed by a `refundGuarantee`. (If someone successfully challenges in `J`, then she can fall back to claiming the guarantee.)

💡 Another idea is to enable challenges with a partially supported preFS as a special case. This way, Alice could challenge in `J` with some outcome that allows her to claim her guarantee.

[...]

PR #3565, opened 2021-06-03
Introduce an "Embedded application" application

### LINKED ISSUE #3552: Implement a virtual funding "joint channel" definition with an embedded app channel.
# Why: Problem description :confused:
Post-hackathon, the virtual funding protocol (VFP) contains 4 channels:

- one ledger channel per "arm"
- the joint channel J
- the app channel X

By removing turn-taking from the core nitro protocol, we are fairly certain that we can consolidate J & X into a single channel. This *does* decrease the gas cost & complexity of challenging, and *may* decrease the network overhead in opening a channel.

# How could an embedded app channel work?

Let's talk about the joint channel `J`. Suppose we "embed" the app channel `X` in the channel `J`. What does this mean?

1. `J`'s outcome starts out as `{A: a, B: b, I: a + b}`.
2. No matter what, `a+b` should be allocated to `I`.
3. There should be some (dynamic) rules that enable Alice to update the outcome to her latest "supported state of `X`", and likewise for Bob.
4. Any one of Alice, Bob & Irene should be able to force the channel to finalize through some valid transition function.

(3) is open-ended, as I haven't said anything about what `X` can be. Some options are:

A. `X` is a full-blown Nitro state channel. In this case, the app data in J would contain things like the channel definition for X, the latest supported state for X, etc.
B. `X` is a "light channel". It could be a `TransferDefinition`, for instance — something that resolves to an `(outcome, nonce)`, given appropriate signatures. In this case, the `nonce` of the "app channel" would be stored in the app data of `J`. The `outcome` of the "app channel" does not need to be stored in the app data of `J`. Instead, the `outcome` would help determine what valid transitions there are in `J`.

(4) is the important detail to consider, when removing turn-taking . One possible design is:
i) `J` has 4 types of states: `ABC`, `A`, `B`, and `AB`.
ii) In the happy path, peers always triple-sign `ABC` states
iii) At any time, Alice can move from an `ABC` state to an `A` state, or from a `B` state to an `AB` state
iv) At any time, Bob can more from an `ABC` state to a `B` state, or from an `A` state to an `AB` state

Every transition has a few rules:
- There's always an outcome item with destination `I`, and its amount doesn't change. It is the last outcome item.
- The total amount allocated in the outcome is invariant.
- When moving from an `ABC` state to an `A` or `B` state, the outcome of `J` changes according to the outcome of `X` that's revealed in the transition. Also, the app turn number revealed in `X` is recorded in `J`'s app data.
- When moving from an `A` state or a `B` state to an `AB` state, `J`'s outcome similarly changes according to the outcome of `X` that's revealed. In addition, the app turn number revealed in `X` needs to exceed the app turn number recorded in `J`'s app data.

This gives Alice & Bob exactly one opportunity each to reveal the most recent outcome agreed to in `X`, without giving them an opportunity to grief Irene out of not accessing her funds. They are allowed to reveal stale states (just as they are in a separate channel). The incentives may be different from the current system, as they only get _one opportunity_ to reveal a state, but I think it's still "fair".

**Additional context**
See https://www.notion.so/statechannels/Virtual-funding-research-11388f704f84496ba8febe30ddb522d2 for more discussion

geoknee:
See this [pitch](https://www.notion.so/statechannels/Virtual-funding-research-11388f704f84496ba8febe30ddb522d2).

Fixes #3552 
 

TODO:
- [x] tidy up utilities
- [x] document meaning of EmbeddedApplication
- [x] complete test coverage for an embedded null app
- [ ] add test coverage for an embedded non-null app
- [x] think about whether or when this PR gets merged


kerzhner (commented):
This work makes sense to me! Nicely done!

There is a lot of nuance in this code. A part of me wonders if introducing this nuance is worth it to avoid a few network roundtrips (discussed [here](https://www.notion.so/statechannels/Virtual-channel-construction-trade-offs-540462846a3d400d8651e4958b500447?d=a16704c1-6e25-43ef-b33a-a836df9eeaca#3b6733104bb84dc6b5af60764578fc0e).

But perhaps embedding the application channel has other benefits like enabling use cases that are not possible with Nitro-as-is?

andrewgordstewart:
> This work makes sense to me! Nicely done!
> 
> There is a lot of nuance in this code. A part of me wonders if introducing this nuance is worth it to avoid a few network roundtrips (discussed [here](https://www.notion.so/statechannels/Virtual-channel-construction-trade-offs-540462846a3d400d8651e4958b500447?d=a16704c1-6e25-43ef-b33a-a836df9eeaca#3b6733104bb84dc6b5af60764578fc0e).)
> 
> But perhaps embedding the application channel has other benefits like enabling use cases that are not possible with Nitro-as-is?

A few network round trips is a large cost in some situations. For instance, if Alice wants to pay Bob for an on demand service in a virtual channel, then additional round trips can greatly effect the quality of service

geoknee:
@kerzhner thanks for the feedback. I think you're right to weigh the increased nuance against the potential benefits, and your document is very helpful for that. I think in any case we should copy some of that insight into some comments or readme next to this code.

In particular, to write down the virtual funding protocol that would accompany the `EmbeddedApplication`. It's only with that protocol in front of us that we can fully judge the safety, pros / cons of this approach.

geoknee:
@kerzhner I did a little bit more tidying up (only changes are the addition of the `EmbeddedApplication.sol`, a TS helper, a test file, and an entry into our test deployment script).

I'm now proposing that we merge this PR and consider the relevant half of the linked pitch as complete.

If we return to it in a future pitch, we should add a test where the embedded application is not the null app.

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
```diff
@@ -0,0 +1,438 @@
...
+    struct SupportProof {
+        IForceMove.FixedPart fixedPart;
+        VariablePart[] variableParts; // either one or two states
+        uint48 turnNumTo;
+        IForceMove.Signature[2] sigs; // one for each participant
+        uint8[2] whoSignedWhat; // whoSignedWhat = [0,0] or [0,1] or [1,0]
+    }
+
+    // Finite states for J
+    enum AlreadyMoved {A, B, AB, ABI}
```
andrewgordstewart, line 139: This is probably a helpful change.

```suggestion
    enum AlreadyMoved {None, A, B, AB}
```

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
```diff
@@ -0,0 +1,438 @@
...
+                fromAllocation[2].destination == toAllocation[2].destination,
+            'destinations may not change'
+        );
+
+        require(fromAllocation[2].amount == toAllocation[2].amount, 'p2.amt !constant');
+        require(
+            fromAllocation[0].amount + fromAllocation[1].amount + fromAllocation[2].amount ==
+                toAllocation[0].amount + toAllocation[1].amount + toAllocation[2].amount,
+            'total allocation changed'
+        );
```
andrewgordstewart, line 176: I prefer to use constants, rather than "magic numbers", for indexes of importance.

As an example benefit, if we wish to change the outcome ordering, we don't need to find-and-replace all uses of the number "1".

andrewgordstewart: I think we could get away with an enum like
```
enum indices {
  A,
  B,
  I
}
```

While this might technically work, it is a bit "clever", since it depends on the enum's values being sequential integers starting at 0.

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
```diff
@@ -0,0 +1,438 @@
...
+                    toAppData.supportProofForX.turnNumTo,
+                    false, // Assume isFinal is false
+                    toAppData.channelIdForX,
+                    appPartHash,
+                    keccak256(
+                        toAppData.supportProofForX.variableParts[toAppData
+                            .supportProofForX
+                            .variableParts
+                            .length - 1]
+                            .outcome
```
andrewgordstewart, line 281: Readability here might benefit from defining a `greaterStateOutcome` variable to avoid these line breaks

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
```diff
@@ -0,0 +1,438 @@
...
+                    )
+                )
+            )
+        );
+
+        if (
+            (toAppData.supportProofForX.whoSignedWhat[0] == 0) &&
+            (toAppData.supportProofForX.whoSignedWhat[1] == 0)
```
andrewgordstewart, line 289: `whoSignedWhat` seems like a confusing name -- I've been confused by it before. 
```
        uint8[] memory whoSignedWhat // whoSignedWhat[i] is the index of the state in stateHashes that was signed by participants[i]
```

I think it's confusing because `whoSignedWhat` is actually a list of "_what_ is signed, indexed by _whom_".

Brainstorming: Is something more along the lines of `didSignState` more clear? `didSignState[2] == 3` means "participant 2 did sign state 3". (I don't like `didSignState` either, just throwing an idea out there.)

andrewgordstewart: maybe `stateIdxSignedBy`? `stateIdxSignedBy[2] == 1` => "participant 2 signed the state with index 1"

geoknee: or `stateIndexSignedByParticipant`?

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
```diff
@@ -0,0 +1,438 @@
...
+        if (fromAppData.alreadyMoved == AlreadyMoved.ABI) {
+            require(
+                (toAppData.alreadyMoved == AlreadyMoved.A && signedByA) ||
+                    (toAppData.alreadyMoved == AlreadyMoved.B && signedByB),
+                'incorrect move from ABI'
+            );
+        } else {
+            // If a support proof has already been supplied, the current support proof must be greater
+            require(
+                toAppData.supportProofForX.turnNumTo > fromAppData.supportProofForX.turnNumTo,
```
andrewgordstewart, line 196: [pre-emptive edit]: The below comment probably requires #3556 to enable the proposed optimization.

--- 

There's probably an opportunity for optimization here: in the _beginning_ of the VFP, Alice & Bob don't need to sign preFS for `X`. They can sign preFS for `J`. Alice can make a state update in the channel once the VFP is done.

This is unsafe for Alice if Alice can _take money from Bob_ in her first move -- she could lose money that's rightfully hers, by not being allowed to to progress the channel. But it's fine for unidirectional Alice -> Bob channels.

💡 can we address that by _not storing a support proof_ in the app data of `J`? Or, not including `sigs` in `J.preFS`? (Alice & Bob will both sign `J.preFS`, so they're implicitly agreeing to the "preFS" that's embedded in `J.preFS.appData`.)

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
```diff
@@ -0,0 +1,438 @@
...
+        require(
+            fromAppData.supportProofForX.fixedPart.appDefinition ==
+                toAppData.supportProofForX.fixedPart.appDefinition,
+            'X.appDefinition changed'
+        );
+        require(
+            fromAppData.supportProofForX.fixedPart.challengeDuration ==
+                toAppData.supportProofForX.fixedPart.challengeDuration,
+            'X.challengeDuration changed'
+        ); // this is actually never used so could be 0
```
andrewgordstewart, line 248: Ideally, this would be refactored to use the same code as the adjudicator.

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
andrewgordstewart: maybe define
```
uint256 greaterStateTurnNum = toAppData.supportProofForX.variableParts.length - 1
```
?

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
geoknee: We should make a new issue for this and tackle it separately to this PR.

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
geoknee: 1a69e8c

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
geoknee: 6019e82

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
andrewgordstewart: #3605 3605

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
```diff
@@ -0,0 +1,446 @@
...
+
+        uint256 finalIndex = toAppData.supportProofForX.variableParts.length - 1;
+
+        bytes memory greaterStateOutcome = toAppData.supportProofForX.variableParts[finalIndex]
+            .outcome;
+
+        bytes32 greaterStateHash = keccak256(
+            abi.encode(
+                IForceMove.State(
+                    toAppData.supportProofForX.turnNumTo,
```
andrewgordstewart, line 286: `toAppData.supportProofForX` is referenced many times. Should we name this as a variable? Maybe something like `toSupportProof`?

geoknee: We could, but I'm a bit skeptical about the benefit. Is `toSupportProofForX.turnNumTo` plus an extra variable declaration really any better than `toAppData.supportProofForX.turnNumTo` ? If it is, would we stop there or also have `toSupportProofVariableParts[0]` and `toSupportProofFixedPart.participant[0]` or even `toSupportProofParticipants[0]` plus many more variables?

Personally I like seeing where each function argument comes from explicitly. I think naming new variables increases the chances of bugs.

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
geoknee: This feels out-of-scope, to me. Perhaps this could go into a pitch for a future cycle?

andrewgordstewart: it is definitely out of scope for this PR! But, it plays into the analysis [here](https://www.notion.so/statechannels/Virtual-funding-optimizations-16998f6d3baa41a49dd69b9948721b66#db9744e2dfff4973b0b23d3de85d1f55), so I was hoping you had some thoughts around it already 😆

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
andrewgordstewart: https://github.com/statechannels/statechannels/issues/3610

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
andrewgordstewart: > Personally I like seeing where each function argument comes from explicitly. I think naming new variables increases the chances of bugs.

In this case, we want to validate many things about `toAppData.supportProofForX`. Assigning its value to a variable, _in theory_, prevents mistakes like
```
            (toAppData.supportProofForX.whoSignedWhat[0] == 0) &
            (fromAppData.supportProofForX.whoSignedWhat[1] == 0)
```
though I admit that this mistake might be easy to catch?

My suggestion is mostly motivated by improving the readability of all of the places where we are validating things about `toAppData.supportProofForX`. If we call it `nextSupportProof`, it makes it easier for a reader to "keep everything in ones head" -- instead of having to check:
- which state's app data does this come from?
- what are we checking about the app data?

you are indicating: "I'm going to focus on the `nextSupportProof`, which is `toAppData.supportProofForX`. I am going to check these properties of the next support proof."

I'm going to try to understand the properties verified below as is, and then come back to this suggestion!

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
```diff
@@ -0,0 +1,446 @@
...
+            Xallocation[AIndex].amount == toAllocation[AIndex].amount &&
+                Xallocation[BIndex].amount == toAllocation[BIndex].amount &&
+                Xallocation[AIndex].destination == toAllocation[AIndex].destination &&
+                Xallocation[BIndex].destination == toAllocation[BIndex].destination,
+            'X / J outcome mismatch'
+        );
+        return true;
+    }
+
+    function validateSupportProofForX(AppData memory fromAppData, AppData memory toAppData)
```
andrewgordstewart, line 240: On discussion with @kerzhner I decided _not_ to dig into the implementation of this. We should merge this, and treat it as "prototype/reference code", and _not_ as code that we believe to be safe enough to use in the virtual funding protocol.

andrewgordstewart (approved):
Let's merge this in!

Let's be clear though -- this should be treated as _prototype code_. I did not read through this carefully from a security point of view.

geoknee:
> Let's be clear though -- this should be treated as prototype code. I did not read through this carefully from a security point of view.

Should we add some comment to that effect next to the code?

packages/nitro-protocol/contracts/examples/EmbeddedApplication.sol
geoknee: Definitely sounds like an interesting idea: although ideally we don't want to rely on any particular property of the application rules of `X` (such as unidirectionality). 

I need to think more about refunding of guarantees. Also interesting, of course, but adding to the core nitro API throws up a lot of new things that can happen.

PR #3568, opened 2021-06-04
Increase clarity around the maximum number of participants

geoknee:
Our docs state that the maximum length of the participants array is 255, but we use the `uint256` type for the `nParticipants` (derived from the length of the participants array).

We already have a runtime check that the length of the participants array is less than 255. It should check that the length is less than or equal to 255.

Given that check, we can cast `nParticipants` as a `uint8` in internal functions. 

This PR fixes the check, fixes the docs, and makes `nParticipants` a `uint8` everywhere.


packages/nitro-protocol/contracts/ForceMove.sol
```diff
@@ -935,7 +935,7 @@ contract ForceMove is IForceMove {
             (numSigs == numParticipants) && (numWhoSignedWhats == numParticipants),
             'Bad |signatures|v|whoSignedWhat|'
         );
-        require(numParticipants < type(uint8).max, 'Too many participants!');
+        require(numParticipants <= type(uint8).max, 'Too many participants!');
```
geoknee, line 938: It's ok to have 255 participants.

geoknee:
Closing as using `uint8` instead of `uint256` is actually[ more gas expensive](https://ethereum.stackexchange.com/questions/3067/why-does-uint8-cost-more-gas-than-uint256).

PR #3569, opened 2021-06-04
Fix check on maximum number of participants

geoknee:
Our docs state that the maximum length of the participants array is 255, but our  runtime check is actually that the length of the participants array is less than 255. It should check that the length is less than or equal to 255.

This PR fixes the check.

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/gas.ts
+++ b/packages/nitro-protocol/gas-benchmarks/gas.ts
@@ -15,7 +15,7 @@ type Path =
 export const gasRequiredTo: GasRequiredTo = {
   deployInfrastructureContracts: {
     vanillaNitro: {
-      NitroAdjudicator: 2421626, // Singleton
+      NitroAdjudicator: 2421830, // Singleton
       ETHAssetHolder: 1634011, // Singleton (could be more in principle)
       ERC20AssetHolder: 1657410, // Per Token (could be more in principle)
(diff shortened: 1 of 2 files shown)


packages/nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -15,7 +15,7 @@ type Path =
 export const gasRequiredTo: GasRequiredTo = {
   deployInfrastructureContracts: {
     vanillaNitro: {
-      NitroAdjudicator: 2421626, // Singleton
+      NitroAdjudicator: 2421830, // Singleton
```
NiloCK, line 18: This is what you had just referenced? This is pretty surprising.

If the reported differences are true, and not the result of weird gas estimation approximations leaving the exact reporting (rather than +/- 1%) will present some learning opportunities. How much value there is in these opportunities is another question.

NiloCK (approved):
LGTM

packages/nitro-protocol/gas-benchmarks/gas.ts
geoknee: I don't think this is quite what I just mentioned in standup. This is an expected increase, I think, since we may be doing more work by checking a non-strict inequality (rather than a strict one). 

I suppose we could save 4 gas by converting it back into a strict inequality.

PR #3570, opened 2021-06-04
increase timeout for concludePushOutcomeAndTransferAll

geoknee:
I'm seeing this flickering a bit, citing timeout. Possibly as side effect of #3560


packages/nitro-protocol/test/contracts/NitroAdjudicator/concludePushOutcomeAndTransferAll.test.ts
```diff
@@ -28,6 +28,8 @@ import {
 import {signStates} from '../../../src';
 import {NITRO_MAX_GAS} from '../../../src/transactions';
 
+jest.setTimeout(15_000);
```
NiloCK, line 31: After a small wild goose chase I have learned that the default timeout for jest tests is 5 seconds.

NiloCK (approved):
LGTM

PR #3571, opened 2021-06-04
require non-zero holdings in _transfer

geoknee:
* Currently `concludePushOutcomeAndTransferAll` will happily succeed if the channel is not funded at all. While this is not necessarily *wrong*, it is pretty unexpected.

### DIFF
--- a/packages/nitro-protocol/contracts/AssetHolder.sol
+++ b/packages/nitro-protocol/contracts/AssetHolder.sol
@@ -235,6 +235,8 @@ contract AssetHolder is IAssetHolder {
         );
         uint256 initialHoldings = holdings[fromChannelId];
 
+        require(initialHoldings > 0, 'no holdings for this channel');
+
         (
             Outcome.AllocationItem[] memory newAllocation,
             bool safeToDelete,
--- a/packages/nitro-protocol/gas-benchmarks/gas.ts
+++ b/packages/nitro-protocol/gas-benchmarks/gas.ts
@@ -16,8 +16,8 @@ export const gasRequiredTo: GasRequiredTo = {
   deployInfrastructureContracts: {
     vanillaNitro: {
       NitroAdjudicator: 2421626, // Singleton
-      ETHAssetHolder: 1632711, // Singleton (could be more in principle)
-      ERC20AssetHolder: 1654810, // Per Token (could be more in principle)
+      ETHAssetHolder: 1651445, // Singleton (could be more in principle)
+      ERC20AssetHolder: 1673544, // Per Token (could be more in principle)
     },
   },
   directlyFundAChannelWithETHFirst: {


andrewgordstewart (commented):
What about deleting the stored data to make the ongoing transaction cheaper?

geoknee:
> What about deleting the stored data to make the ongoing transaction cheaper?

I'm not sure I follow?

I'm suggesting that the transaction ought to revert when there are no holdings. 

There's already logic that causes slots in the Asset Holder to be deleted or zero-ed out if appropriate: so if the transaction does not revert then it will be made cheaper via those mechanisms (gas refund).

andrewgordstewart:
> > What about deleting the stored data to make the ongoing transaction cheaper?
> 
> I'm not sure I follow?
> 
> I'm suggesting that the transaction ought to revert when there are no holdings.
> 
> There's already logic that causes slots in the Asset Holder to be deleted or zero-ed out if appropriate: so if the transaction does not revert then it will be made cheaper via those mechanisms (gas refund).

If the transaction reverts, the user pays all of the gas. But if there's a non zero outcome slot, then the transaction could instead zero it out (there's no holdings, so why not zero it out?) and not change the holdings. This would provide a gas refund.

But, having said that, there's a big downside to this: you would then be almost forced into challenging a ledger channel, then transferring funds, then challenging in a ledger funded channel. Why? If I challenge in a ledger funded channel, and the challenge times out, then my suggestion above would wipe the asset holder clean. So Eve could annoy Alice, and Alice would have to re challenge the ledger funded channel.

So, I think reverting is the best change to make, if you're going to make a change. We shouldn't expect to hit this path anyway.

packages/nitro-protocol/contracts/AssetHolder.sol
```diff
@@ -235,6 +235,8 @@ contract AssetHolder is IAssetHolder {
         );
         uint256 initialHoldings = holdings[fromChannelId];
 
+        require(initialHoldings > 0, 'no holdings for this channel');
```
andrewgordstewart, line 238: What's the gas cost of the additional statement?

We don't expect to hit this revert. So, i could argue that you're requiring every user to pay a small amount of gas for a line of code that in theory never triggers a revert, and also doesn't provide any additional safety.

andrewgordstewart: (meta comment: this highlights the benefit of the gas accounting scripts)

geoknee: The additional statement costs 17 gas, according to my investigations (plus a bigger increase to deployment costs). 

I agree that it doesn't provide any additional safety, and that we don't expect to hit this code path. I hit the code path while writing a gas benchmarking script, which this change helped me debug.

I think on balance maybe we just close this, or replace the change with a comment.

PR #3572, opened 2021-06-04
More gas benchmarking

geoknee:
adds a case for a happy exit with ETH, and one for ERC20 tokens. Follow on work from #3544.

Non-essential Changes:
~Includes #3571~

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -2,7 +2,7 @@ import {channelId} from './fixtures';
 import {gasRequiredTo} from './gas';
 import {erc20AssetHolder, ethAssetHolder, nitroAdjudicator, token} from './vanillaSetup';
 
-describe('Consumes the expected gas', () => {
+describe('Consumes the expected gas for deployments', () => {
   it(`when deploying the NitroAdjudicator >>>>>  ${gasRequiredTo.deployInfrastructureContracts.vanillaNitro.NitroAdjudicator} gas`, async () => {
     await expect(await nitroAdjudicator.deployTransaction).toConsumeGas(
       gasRequiredTo.deployInfrastructureContracts.vanillaNitro.NitroAdjudicator
@@ -18,6 +18,8 @@ describe('Consumes the expected gas', () => {
       gasRequiredTo.deployInfrastructureContracts.vanillaNitro.ERC20AssetHolder
     );
   });
+});
+describe('Consumes the expected gas for deposits', () => {
   it(`when directly funding a channel with ETH (first deposit) >>>>>  ${gasRequiredTo.directlyFundAChannelWithETHFirst.vanillaNitro} gas`, async () => {
     await expect(await ethAssetHolder.deposit(channelId, 0, 5, {value: 5})).toConsumeGas(
       gasRequiredTo.directlyFundAChannelWithETHFirst.vanillaNitro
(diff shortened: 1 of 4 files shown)


packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -1,31 +1,31 @@
-import {channelId} from './fixtures';
+import {channelId, finalizationProof, finalState} from './fixtures';
 import {gasRequiredTo} from './gas';
 import {erc20AssetHolder, ethAssetHolder, nitroAdjudicator, token} from './vanillaSetup';
 
 describe('Consumes the expected gas for deployments', () => {
-  it(`when deploying the NitroAdjudicator >>>>>  ${gasRequiredTo.deployInfrastructureContracts.vanillaNitro.NitroAdjudicator} gas`, async () => {
```
NiloCK, line 6: This change (removing the interpolated gas value) is in the spirit of maintaining a single "look here to understand the gas costs" point in `./gas.ts`?

geoknee: Partly, yes. It was also just getting annoying putting these here, especially when a test case covered multiple gas consumption values.

packages/nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -47,7 +49,16 @@ export const gasRequiredTo: GasRequiredTo = {
...
   },
+  ETHexit: {
+    // e completely liquidate the channel (paying out both parties)
+    vanillaNitro: 146793,
+  },
+  ERC20exit: {
+    // We completely liquidate the channel (paying out both parties)
+    // This is cheaper than the ETHexit because storage slots are zero-ed out.
+    vanillaNitro: 124148,
+  },
```
NiloCK, line 63: Understanding check:

`ETHexit` consumes more gas than `ERC20exit`, despite ETH transfers being privileged by the EVM, because of differences in the setup:

- `ETHAssetHolder.deposit` writes a new storage slot in the assetholder (`holdings[destination]` is being upped from 0 to 10), while
- ... (20 minutes later) actually I don't seem to understand this. Comment?

NiloCK (commented):
This looks good to me.

NiloCK (approved):

packages/nitro-protocol/gas-benchmarks/gas.ts
geoknee: I think you are broadly on the correct track. 

ETH deposits and payouts require no writes to storage other than the `AssetHolder.holdings` mapping.

ERC20 deposits and payouts require writes to storage in the Token contract itself. The first deposit is very expensive (20K), but if we pair it with the refund from zeroing the slot when paying out (-15K) it's only really 5K more expensive. 

What we are seeing here is the ERC20 exit benefitting from such a refund. If we think about that in isolation, it seems cheaper than an ETH exit. But when we look above, we see that the ERC20 deposit carried a gas premium that outweighs that refund. 

EIP-3529 is going to greatly reduce the refund. 

However, you have made me realise that this benchmark is probably not a realistic one. We only get to zero the balance of the AssetHolder (in the Token contract) when there are no other funded channels in the universe. That is very unlikely. We should alter the benchmark to keep some ERC20 tokens in the AssetHolder (for another channel, say).

geoknee: I made the change in cb05d14; for the ETH case too just for completeness (it makes no difference, as expected)

geoknee: It results in exactly a 15K increase :-)

PR #3573, opened 2021-06-04
Add a toConsumeGas jest matcher

geoknee:
The current gas benchmark test file:
1. is quite repetitive with converting to/from BigNumbers and awaiting transactions to be mined
2. doesn't give you a hint about what to do in the case of a test failure (you should fix the expectation, and not the test, in many cases)
3. won't be that easy to change to a "fuzzy" check which doesn't require strict equality

A custom matcher can:
*  DRY out the code, and provide a single place to fuzzify the expectation if we want to later on
* provide a custom message, including context-specific hints

## Changes [Optional] 
1.  adds a custom matcher, following [the documentation for jest.](https://jestjs.io/docs/expect#async-matchers)
2. uses that matcher in the benchmark test file

Existing test continue to pass.  A manual mutation confirms the new message pops up as expected.

---
### Code quality
### Project management
- [ ] I have chosen the appropriate [pipeline](https://www.notion.so/statechannels/Team-working-agreements-3cbcd6e85a7e481db4fe572ddf50cbd6#a96b6b02704d46afbcff147cf5a85566) on zenhub for the linked issue

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/vanillaSetup.ts
+++ b/packages/nitro-protocol/gas-benchmarks/vanillaSetup.ts
@@ -5,6 +5,17 @@ import nitroAdjudicatorArtifact from '../artifacts/contracts/NitroAdjudicator.so
 import ethAssetHolderArtifact from '../artifacts/contracts/ETHAssetHolder.sol/ETHAssetHolder.json';
 import erc20AssetHolderArtifact from '../artifacts/contracts/ERC20AssetHolder.sol/ERC20AssetHolder.json';
 import tokenArtifact from '../artifacts/contracts/Token.sol/Token.json';
+import {BigNumber, BigNumberish} from '@ethersproject/bignumber';
+import {Transaction} from 'ethers';
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface Matchers<R> {
+      toConsumeGas(benchmark: number): R;
+    }
+  }
+}
 
 export let ethAssetHolder: Contract;
 export let erc20AssetHolder: Contract;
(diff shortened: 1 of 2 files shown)


packages/nitro-protocol/gas-benchmarks/vanillaSetup.ts
```diff
@@ -49,3 +60,25 @@ beforeEach(async () => {
 afterAll(async () => {
   await ganacheServer.close();
 });
+
+expect.extend({
+  async toConsumeGas(
+    received: any, // TransactionResponse
```
geoknee, line 66: sadly [this type](https://docs.ethers.io/v5/api/providers/types/#providers-TransactionResponse) does not seem to be exported by ethers.

packages/nitro-protocol/gas-benchmarks/vanillaSetup.ts
```diff
@@ -49,3 +60,25 @@ beforeEach(async () => {
...
+    benchmark: number
+  ) {
+    const {gasUsed} = await received.wait();
+    const pass = (gasUsed as BigNumber).eq(benchmark); // This could get replaced with a looser check with upper/lower bounds
+    if (pass) {
+      return {
+        message: () => `expected to NOT consume ${benchmark} gas, but did`,
```
NiloCK, line 73:
```suggestion
        message: () => `expected to consume ${benchmark} gas, and did`,
```
I've read this about 4 times and one of us has something backwards in our head.

packages/nitro-protocol/gas-benchmarks/vanillaSetup.ts
```diff
@@ -49,3 +60,25 @@ beforeEach(async () => {
...
+        pass: true,
+      };
+    } else {
+      return {
+        message: () =>
+          `expected to consume ${benchmark} gas, but actually consumed ${(gasUsed as BigNumber).toNumber()} gas. Consider updating the appropriate number in gas.ts!`,
```
NiloCK, line 79:
```suggestion
    const gasUsed: BigNumber = (await received.wait()).gasUsed as BigNumber;
    const pass = gasUsed.eq(benchmark); // This could get replaced with a looser check with upper/lower bounds
    if (pass) {
      return {
        message: () => `expected to NOT consume ${benchmark} gas, but did`,
        pass: true,
      };
    } else {
      return {
        message: () =>
          `expected to consume ${benchmark} gas, but actually consumed ${gasUsed.toNumber()} gas. Consider updating the appropriate number in gas.ts!`,
```
This might be a little easier to read, with all of the type assertion on one line.

NiloCK (approved):
Looks good.

packages/nitro-protocol/gas-benchmarks/vanillaSetup.ts
geoknee: Yeah it's a bit confusing but this is the slot where you put the message to be printed in the case `expect().not.toConsumeGas()` fails. 

See this article https://everyday.codes/javascript/create-custom-jest-matchers-to-test-like-a-pro/ .

geoknee: (No message is printed when `pass==true`)

PR #3574, opened 2021-06-04
fail earlier by running gas benchmarks first

geoknee:
They are the shortest tests, and the easiest to break.

### DIFF
--- a/packages/nitro-protocol/package.json
+++ b/packages/nitro-protocol/package.json
@@ -97,7 +97,7 @@
     "prettier:write": "prettier --write './contracts/**/*.sol'",
     "test": "yarn test:contracts && yarn test:app --all",
     "test:app": "jest -c ./config/jest/jest.config.js",
-    "test:ci": "yarn test:ci:app && yarn test:ci:contracts && yarn test:ci:gas",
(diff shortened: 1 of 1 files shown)


packages/nitro-protocol/package.json
```diff
@@ -97,7 +97,7 @@
     "prettier:write": "prettier --write './contracts/**/*.sol'",
     "test": "yarn test:contracts && yarn test:app --all",
     "test:app": "jest -c ./config/jest/jest.config.js",
-    "test:ci": "yarn test:ci:app && yarn test:ci:contracts && yarn test:ci:gas",
+    "test:ci": "yarn test:ci:gas && yarn test:ci:app && yarn test:ci:contracts ",
```
NiloCK, line 100: A true CircleCI gas optimization.

Any gut feeling about the [fragility / runtime] ratio on `test:ci:app` and `test:ci:contracts`?

NiloCK (approved):
Seems sensible to me.

packages/nitro-protocol/package.json
geoknee: Hmm I think the contracts probably take longer, but I would need to check.

PR #3575, opened 2021-06-07
add typechain support to nitro-protocol

geoknee:
https://www.notion.so/statechannels/Add-TypeChain-tooling-f17342226cb44244be4820b9937c9fae

This didn't go completely smoothly, unfortunately.

The first issue I hit was that the typings exported by typechain did not inherit from the ethers `Contract` class properly: meaning that I got errors when trying to access e.g. `nitroAdjudicator.address`. I tracked this down to a mismatch in versions between ethers and typechain. Newer versions of ethers have a `BaseContract` class, which is imported by typechain. Older versions (still v5) do not export such a class. See https://github.com/ethers-io/ethers.js/issues/1384 for more info.

The next issue was with the typechain [contract factories](https://github.com/ethereum-ts/TypeChain/tree/master/packages/target-ethers-v5#contract-factories). I got some typescript errors when trying to use these, so I had to fall back to regular ethers contract factories plus type asserting the return value as a typechain class.

Nevertheless, this PR brings us to a state where we can do this: 

Screenshot 2021-06-07 at 13 11 49

I didn't spread the types around the codebase much, yet. For now I just used them in the gas benchmarking test. We can do more of that kind of thing in future PRs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,6 +120,7 @@ jobs:
             - packages/*/dist
             - packages/*/build
             - packages/nitro-protocol/artifacts
+            - packages/nitro-protocol/typechain
             - packages/*/lib
             - packages/*/node_modules
             - packages/client-api-schema/types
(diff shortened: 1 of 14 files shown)


geoknee:
In favour of #3604

PR #3580, opened 2021-06-08
Update architecture.md

andrewgordstewart:
This is overdue, after @lalexgap's changes to the wallet API.

- add message service
- remove `scope` from objective description


packages/server-wallet/docs/architecture.md
```diff
@@ -34,39 +34,56 @@ A state channel wallet has the following _runtime dependencies_:
...
+          +------^-----+                                               
+                 |                                                     
+                 |                                                     
+                 |                                                     
+                 |                                                     
+                 |                                                     
+       +---------v--------+                                            
+       |      ENGINE      |                                            
+       |                  |                                            
+       |   +-----------+  |
```
lalexgap, line 76: Currently both the `wallet` and `engine` access the `store`

lalexgap (approved):

packages/server-wallet/docs/architecture.md
andrewgordstewart: What does the `wallet` use the store for?

lalexgap: The most common use case is updating chain data in the `store` from `ChainEvent`s from the chain service before cranking the engine.

andrewgordstewart: is that the _only_ use case?

In my head, I expected that the wallet would just push a chain event into the engine, just like it pushes a message into the engine. In that world, a wallet would not need a store.

andrewgordstewart: da76ba062b78d4f8c91fcd897507f0a1a43b19c0

Issue #3576, opened 2021-06-07
EVM Validation does not work 


lalexgap:
Since we always [use a](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/models/channel.ts#L609:L609) `undefined` bytecode we can never validate with the eve.

geoknee:
@lalexgap can you clarify what the issue is? Is it that we should not be using `undefined` bytecode? 

To my mind it should be ok to use an `undefined` bytecode, and it is correct behaviour in that case that no transitions can be validated -- but we can still check to see if a state is supported by looking at the signatures.

lalexgap:
> @lalexgap can you clarify what the issue is? Is it that we should not be using `undefined` bytecode?
> 
> To my mind it should be ok to use an `undefined` bytecode, and it is correct behaviour in that case that no transitions can be validated -- but we can still check to see if a state is supported by looking at the signatures.

The issue that we **always** use an undefined `bytecode` (https://github.com/statechannels/statechannels/blob/effe360f6d6aff4e5c1e9644e3fe6b062b7283df/packages/server-wallet/src/models/channel.ts#L594). 

This means that we always return false for app transitions if a channel has an `AppDefinition`.

geoknee:
right! thanks for clarification

PR #3582, opened 2021-06-09
Add "Update Channel" to wallet API

### LINKED ISSUE #3504: Add updateChannel method to Wallet API
We need to expose an API for updating a channel once it is open.

lalexgap:
Fixes #3504 

Adds an update channel method to the `wallet`. The directly funded channel test has been updated to call `updateChannel`.

Due to #3600 validate transition fails when there is no `bytecode` provided.  The directly funded test has been updated to use the `COUNTING_APP` to get around this. This also means once #3576 is resolved we'll be validating the transition with the evm.

### DIFF
--- a/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
+++ b/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
@@ -169,14 +169,18 @@ test.each(testCases)(
   async options => {
     TestMessageService.setLatencyOptions({a, b}, options);
 
+    const startBalance = {aAmount: 3, bAmount: 2};
+    const updatedBalance = {aAmount: 1, bAmount: 4};
+
     const channelParams: CreateChannelParams = {
       participants: [participantA, participantB],
-      allocations: [createAllocation(3, 2)],
+      allocations: [createAllocation(startBalance)],
       appDefinition: COUNTING_APP_DEFINITION,
       appData: utils.defaultAbiCoder.encode(['uint256'], [1]),
       fundingStrategy: 'Direct',
       challengeDuration: ONE_DAY,
     };
+
     const aBalanceInit = await getBalance(aAddress);
     const bBalanceInit = await getBalance(bAddress);
     const assetHolderBalanceInit = await getBalance(ethAssetHolderAddress);
@@ -189,12 +193,14 @@ test.each(testCases)(
     await expect(bResponse).toBeObjectiveDoneType('Success');
 
     const assetHolderBalanceUpdated = await getBalance(ethAssetHolderAddress);
-    expect(BN.sub(assetHolderBalanceUpdated, assetHolderBalanceInit)).toEqual(BN.add(2, 3));
+
+    const totalDepositAmount = BN.add(startBalance.aAmount, startBalance.bAmount);
+    expect(BN.sub(assetHolderBalanceUpdated, assetHolderBalanceInit)).toEqual(totalDepositAmount);
 
     const {channelId} = response[0];
     const updated = await a.updateChannel(
       channelId,
-      [createAllocation(1, 4)],
+      [createAllocation(updatedBalance)],
       utils.defaultAbiCoder.encode(['uint256'], [2])
     );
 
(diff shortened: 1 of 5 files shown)


packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -216,7 +215,21 @@ test.each(testCases)(
     const aBalanceFinal = await getBalance(aAddress);
     const bBalanceFinal = await getBalance(bAddress);
 
-    expect(BN.sub(aBalanceFinal, aBalanceInit)).toEqual(aFunding);
-    expect(BN.sub(bBalanceFinal, bBalanceInit)).toEqual(bFunding);
+    expect(BN.sub(aBalanceFinal, aBalanceInit)).toEqual(BN.from(1));
+    expect(BN.sub(bBalanceFinal, bBalanceInit)).toEqual(BN.from(4));
```
NiloCK, line 219: The test values `(1,4)` are used a few times here and their meaning wasn't very obvious to me on reading their individual lines. The block in which they are used is not large, but it left me wondering whether the same values are inserted elsewhere as well.

Something like:
```
const finalOutcome = {
 aBalance: 1,
 bBalance: 4,
}
```
can make it easier to reason about the scope of those values. Similar for `initBalance`.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -216,7 +215,21 @@ test.each(testCases)(
...
+      destination: participantA.destination,
+      amount: hexZeroPad(BigNumber.from(aAmount).toHexString(), 32),
+    },
+    {
+      destination: participantB.destination,
+      amount: hexZeroPad(BigNumber.from(bAmount).toHexString(), 32),
+    },
+  ],
+  assetHolderAddress: ethAssetHolderAddress,
+});
```
NiloCK, line 235: This could instead receive the participants as input as well, which would save the refactor that pulled `participantA` and `B` out to the global scope.

Global scope is "generally bad" but it's not clear to me that it reduces readability etc in this file. Is there much potential future where this testing suite is expanded and the reuse, rather than re-instantiation of `A` and `B` here presents problems?

For that matter, how immediately confident are we that reusing A and B in this `test.each` loop produces all of the same behaviour as previously?

packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -224,13 +256,16 @@ export class Wallet extends EventEmitter<WalletEvents> {
   public async closeChannels(channelIds: string[]): Promise<ObjectiveResult[]> {
     return Promise.all(
       channelIds.map(async channelId => {
+        const objectiveId = ['CloseChannel', channelId].join('-');
```
NiloCK, line 259:
```suggestion
        const objectiveId = `CloseChannel-${channelId}`;
```
I personally have to think less reading this than the join. I dug around a bit here to see what would be involved in type-checking the `'CloseChannel'` string, and it seemed surprisingly complicated! The string literal union type of `_Objective.type` is inferred from the literals supplied in implementing generic types.

I have no comment on whether this is an anti-pattern, but it took a few minutes to wrap my head around it.

NiloCK (approved):
Looks good to me!

packages/server-wallet/src/wallet/wallet.ts
lalexgap: I ended up [updating](https://github.com/statechannels/statechannels/pull/3582/commits/9fa21b60972dedd68423a476293c7cdfa1e048fd) `wallet-core` to expose a `calculateObjectiveId` which can generate an objective id with just the type and channel Id.

That way the `wallet` doesn't have to worry at all about how an objective id is calculated.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
lalexgap: Good idea! I've done something like that in https://github.com/statechannels/statechannels/pull/3582/commits/60029bdad9fa02b25ea298be1f3b952fcf921ff2

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
lalexgap: Good point, I've updated the method to take in the participants in https://github.com/statechannels/statechannels/pull/3582/commits/c7d3908b8d7056ef6af1f2964c0c59e0c1a3f34e

PR #3586, opened 2021-06-10
fix some eslint settings and warnings

geoknee:
At the top level of the monorepo, we thread the `--max-warnings=0` flag down to the `yarn lint:check` script.

This usually forces us to tidy up lint warnings (unused variables, etc). However, if `yarn lint:check = eslint && something else`, the flag never gets to the eslint bit where it is needed. 

I think we probably ought to be writing the flag into each package. Here's a step in that direction with nitro-protocol.


NiloCK (commented):
Looks good.

NiloCK (approved):

PR #3587, opened 2021-06-10
Remove E2E test

lalexgap:
With the upcoming changes in  https://github.com/statechannels/statechannels/pull/3562 most of the existing e2e test will be disabled. Since the existing e2e tests just use `engine`s, they will no longer work after the chain service is removed from the engine.

By removing e2e tests we no longer have to worry about maintaining them / understanding them. They will be replaced by the work for this [pitch](https://www.notion.so/statechannels/Pitches-1c3c0a682552482a89ccc241e9e335df?p=88ed836137a74f4c8ce1ad56dc3ef3ff). In the meantime the [directly funded test](https://github.com/statechannels/statechannels/pull/3557) at least tests some of the same behaviour.

It also simplifies our circleCI config a bit and lets us get rid of a bunch of dependencies  🎉


andrewgordstewart:
Screen Shot 2021-06-10 at 11 15 28 AM

😄 🚀 🌋

andrewgordstewart (approved):
> > is it better to "reclaim" this work, or write the "realistic testing" infrastructure from scratch? (if the latter, keep in mind that nothing stops us from copy-pasting from these deleted files.)
> 
> I suspect that it will be better to start from scratch but I think that should be figured out once we hammer out the pitch a bit more. The current e2e tests have a lot of logic to handle sending and receiving messages. The new "realistic" e2e tests will be using a `MessageService` for this so the current e2e framework might not be that helpful.

If you turn out to be wrong, it's easy to revert this PR.

(I think it's easiest to revert if you use a merge commit, instead of squash & merge?)

andrewgordstewart:
> > what test coverage do we lose by merging this, considering the new directly funded integration test?
> > We lose:
> 
>     * "stress" test
> 
>     * `clinic` profiling
> 
>     * the challenge test (but this is disabled for #3562 anyways)

I added this to a **Downsides** section in the PR description. It's nice to be clear about what changes are being made!

PR #3590, opened 2021-06-11
Remove deployment code from devtools and remove etherlime dependency

geoknee:
Our toolchain for smart contract development and integration testing has been evolving. Time was, we used a package called [etherlime](https://github.com/LimeChain/etherlime), which is now considered obsolete. The aim of this PR is to remove that dependency from our monorepo. 

We still depend on ganache (for a local blockchain for most of our tests) and hardhat (for contract compilation, production deploys, and also for local blockchain for some tests), and of course ethers (for pretty much everything else, including JSON-RPC providers so we can interact with ganache). 

The approach to this PR is to rely on ethers directly in our deployment scripts, which are the only files that depend on etherlime. They do this indirectly, via our own devtools package. To remove the dependency from there, I ripped pretty much all of the contract-deployment related code -- the majority of which is unused. 

✅ Pros:
Code deletion, removal of deprecated package. 

👎 Cons: 
The major functionality that we will lose with this PR (but we don't currently make use of) is the "shared ganache" and "ganache n cache" code. Historically we used that for developing apps such as web3torrent. My suggestion is that we can look into reimplementing that (without using etherlime, of course) if we need it in future

## Changes
* etherlime dependency removed from devtools.
* deploy scripts in server-wallet, nitro-protocol, browser-wallet and interop-tests modified to not depend on devtools.

## :warning: Does this require multiple approvals? 
I'll say yes, since some may feel we want to preserve the shared ganache deployment functionality. We could try and preserve that now, but it would be more work.

---
### Code quality
### Project management


andrewgordstewart (approved):
I don't see any benefit of keeping code around that depends on an obsolete package that we don't need!

andrewgordstewart:
> I'll say yes, since some may feel we want to preserve the shared ganache deployment functionality. We could try and preserve that now, but it would be more work.

Let's not do the work to preserve the functionality right now. When we knwo we need that functionality, we can copy/paste from the removed code as needed, or write it from scratch.

PR #3593, opened 2021-06-11
Fix Linting Errors

lalexgap:
Not sure how they got introduced to `master` but this fixes the listing errors causing it to fail.


andrewgordstewart:
It was probably introduced by https://github.com/statechannels/statechannels/pull/3586

lalexgap:
> Do you know if there are vscode settings (checked into git) or pre-commit hooks that might prevent this down the line?
> It was probably introduced by #3586

I bet if we enabled `Require branches to be up to date before merging` we would have caught this.

PR #3596, opened 2021-06-14
Add a gas benchmark for the sad path defunding of a directly funded channel

geoknee:
Follow on work for #3544 .

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -7,7 +7,7 @@ import {
   someState,
 } from './fixtures';
 import {gasRequiredTo} from './gas';
-import {erc20AssetHolder, ethAssetHolder, nitroAdjudicator, token} from './vanillaSetup';
+import {erc20AssetHolder, ethAssetHolder, nitroAdjudicator, token, provider} from './vanillaSetup';
 
 describe('Consumes the expected gas for deployments', () => {
   it(`when deploying the NitroAdjudicator`, async () => {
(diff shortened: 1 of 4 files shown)


packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -98,3 +105,50 @@ describe('Consumes the expected gas for happy-path exits', () => {
     ).toConsumeGas(gasRequiredTo.ERC20exit.vanillaNitro);
   });
 });
+describe('Consumes the expected gas for sad-path exits', () => {
+  it(`when exiting a directly funded (with ETH) channel`, async () => {
+    // begin setup
+    await (await ethAssetHolder.deposit(someOtherChannelId, 0, 10, {value: 10})).wait(); // other channels are funded by this asset holder
+    await (await ethAssetHolder.deposit(channelId, 0, 10, {value: 10})).wait();
```
NiloCK, line 112:
```suggestion
    // begin setup by funding the channel
    await (await ethAssetHolder.deposit(channelId, 0, 10, {value: 10})).wait();
    // fund a second channel in order to avoid the rare-in-practice gas refund for writing
    // a storage slot to zero
    await (await ethAssetHolder.deposit(someOtherChannelId, 0, 10, {value: 10})).wait();
```

This change set looks good to me. My only suggestion is to add the "why" to this setup step because it might not be obvious.

NiloCK (approved):
Hooray for benchmarks!

packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
geoknee: Good call. Made a new PR for it #3608

PR #3597, opened 2021-06-14
Re-enable interop test

### LINKED ISSUE #3592: Re-enable interop test
While I don't think it's going to be too much effort it seems like too much to do it in the scope of this PR. I've just skipped the test for now in a58319e

_Originally posted by @lalexgap in https://github.com/statechannels/statechannels/pull/3562#discussion_r650239202_

lalexgap:
Fixes #3592.

Update the interop tests to use a server wallet and message service. This allows us to re-enable these tests.

### DIFF
--- a/packages/interop-tests/src/__test__/interop.test.ts
+++ b/packages/interop-tests/src/__test__/interop.test.ts
@@ -1,9 +1,9 @@
 import {
   DBAdmin,
   defaultTestConfig,
-  Output,
-  SingleChannelOutput,
-  SingleThreadedEngine
+  SingleThreadedEngine,
+  SyncOptions,
+  Wallet as ServerWallet
 } from '@statechannels/server-wallet';
 import {TEST_ACCOUNTS} from '@statechannels/devtools';
 import {ChannelWallet} from '@statechannels/browser-wallet';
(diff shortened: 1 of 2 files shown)


packages/interop-tests/src/__test__/interop.test.ts
```diff
@@ -79,39 +72,28 @@ beforeAll(async () => {
...
     makeAddress(new Wallet(TEST_ACCOUNTS[1].privateKey).address)
   );
-
-  serverAddress = await serverWallet.getSigningAddress();
+  const engine = await SingleThreadedEngine.create(serverConfig);
+  const chainService = new ChainService(serverConfig.chainServiceConfiguration);
+  const factory = BrowserServerMessageService.createFactory(browserWallet);
+  // Currently the browser wallet crashes if it receives the same objective again
+  // To avoid this we prevent the wallet from retrying objectives by using
+  // a really large poll value (one hour)
```
andrewgordstewart, line 84: This is a nice bug to file!

andrewgordstewart (approved):
Nice!!!

packages/interop-tests/src/__test__/interop.test.ts
andrewgordstewart: (Does the server wallet spam peers with the same objective, with the default sync options?)

lalexgap: > (Does the server wallet spam peers with the same objective, with the default sync options?)

The [defaults](https://github.com/statechannels/statechannels/blob/6368ddca63fa2f67ba9f90770902755ccf67ce59/packages/server-wallet/src/wallet/wallet.ts#L34) only wait 1 second before resending messages which is probably much too short outside of test settings.

PR #3601, opened 2021-06-15
Benchmark sad defunding of a ledger-funded channel

geoknee:
Follows the established benchmarking patterns. introduces a ledger channel funding the channel in question. Both are challenged and the money is pushed down from L to X and then out. (The scenario being that the counter-party goes offline).

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -21,11 +21,13 @@ describe('Consumes the expected gas for deployments', () => {
       gasRequiredTo.deployInfrastructureContracts.vanillaNitro.NitroAdjudicator
     );
   });
+
   it(`when deploying the ETHAssetHolder`, async () => {
     await expect(await ethAssetHolder.deployTransaction).toConsumeGas(
       gasRequiredTo.deployInfrastructureContracts.vanillaNitro.ETHAssetHolder
     );
   });
+
   it(`when deploying the ERC20AssetHolder`, async () => {
     await expect(await erc20AssetHolder.deployTransaction).toConsumeGas(
       gasRequiredTo.deployInfrastructureContracts.vanillaNitro.ERC20AssetHolder
@@ -38,6 +40,7 @@ describe('Consumes the expected gas for deposits', () => {
       gasRequiredTo.directlyFundAChannelWithETHFirst.vanillaNitro
     );
   });
+
   it(`when directly funding a channel with ETH (second deposit)`, async () => {
     // begin setup
     const setupTX = ethAssetHolder.deposit(channelId, 0, 5, {value: 5});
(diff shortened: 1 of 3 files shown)


packages/nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -62,11 +65,23 @@ export const gasRequiredTo: GasRequiredTo = {
...
     },
   },
+  ETHexitSadLedgerFunded: {
+    // Scenario: counterparty goes offline
+    vanillaNitro: {
+      challengeX: 93404,
+      challengeL: 92122,
+      pushOutcomeAndTransferAllL: 58640,
+      pushOutcomeAndTransferAllX: 107742,
+      total: 351908, // An alternative would be to 1. pushOutcome then 2. transfer(indices)
```
geoknee, line 83: TLDR it costs 352K gas to exit a ledger-funded channel in the worst case.

packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -105,36 +111,27 @@ describe('Consumes the expected gas for happy-path exits', () => {
...
     await (await ethAssetHolder.deposit(channelId, 0, 10, {value: 10})).wait();
-    const finalizesAtPromise = new Promise(resolve =>
-      nitroAdjudicator.once('ChallengeRegistered', (...event) => {
-        resolve(event[2]); // 2 is the position of finalizesAt
-      })
-    );
-    async function waitForChallengeToTimeOut(finalizesAt) {
-      await provider.send('evm_setNextBlockTimestamp', [finalizesAt + 1]);
-      await provider.send('evm_mine', []);
-    }
```
NiloCK, line 121: Glad to see this moved out. I should have commented on it in the first place.

packages/nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -17,7 +17,12 @@ type Path =
   | 'ETHexitSadLedgerFunded';
 
 // The channel being benchmarked is a 2 party null app funded with 5 wei / tokens each.
-
+// KEY
```
NiloCK, line 20: These diagrams are straightforward but quite useful.

Should each step be included inside the relevant implementing test at the appropriate spots?

NiloCK (approved):
This looks good to me, although I confess it took me a few minutes to parse the ledger funded sad exit.

I pushed a trivial commit on top of these (separating each test case with a blank line), which seemed like an appropriate thing to instead of opening 5-10 suggested edits.

packages/nitro-protocol/gas-benchmarks/gas.ts
geoknee: I've got some improvements coming to these diagrams in a stacked PR. I think putting them in the implementation is also a good idea -- I'll take a look at that in a follow-up PR.

PR #3602, opened 2021-06-15
benchmark sad path for exiting a virtually-funded channel

geoknee:
Follows on from #3601.

* adds a benchmark for a virtually funded channel
* improves the inline diagrams a bit and replicates them in actual test script
* introduces a `TestChannel` class which helps DRY out the fixtures somewhat

When this PR is merged, we can consider this [associated pitch](https://www.notion.so/statechannels/Improved-gas-benchmarking-in-nitro-protocol-d14938f9671a48588e6974462e9ffb00) as completed.

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -1,4 +1,4 @@
-import {Transaction} from '@ethersproject/transactions';
+import {encodeAllocation, encodeGuarantee} from '../src';
 
 import {
   channelId,
@@ -10,7 +10,14 @@ import {
   someState,
   waitForChallengesToTimeOut,
   getFinalizesAtFromTransactionHash,
-  someLedgerState,
+  someLedgerStateFundingX,
+  someLedgerStateFundingG,
+  someGuarantorState,
+  someJointChannelState,
+  guarantorChannelId,
+  jointChannelId,
+  guarantee,
+  jointChannelAllocation,
 } from './fixtures';
 import {gasRequiredTo} from './gas';
 import {erc20AssetHolder, ethAssetHolder, nitroAdjudicator, token} from './vanillaSetup';
@@ -95,6 +102,7 @@ describe('Consumes the expected gas for deposits', () => {
 describe('Consumes the expected gas for happy-path exits', () => {
   it(`when exiting a directly funded (with ETH) channel`, async () => {
     // begin setup
+    await (await ethAssetHolder.deposit(someOtherChannelId, 0, 10, {value: 10})).wait(); // other channels are funded by this asset holder
     await (await ethAssetHolder.deposit(channelId, 0, 10, {value: 10})).wait();
     // end setup
     const fP = finalizationProof(finalState(ethAssetHolder.address));
@@ -173,7 +181,7 @@ describe('Consumes the expected gas for sad-path exits', () => {
     // begin setup
     await (await ethAssetHolder.deposit(ledgerChannelId, 0, 10, {value: 10})).wait();
     // end setup
-    const ledgerProof = counterSignedSupportProof(someLedgerState(ethAssetHolder.address));
+    const ledgerProof = counterSignedSupportProof(someLedgerStateFundingX(ethAssetHolder.address));
     const challengeLedgerTx = await nitroAdjudicator.challenge(
       ledgerProof.fixedPart,
       ledgerProof.largestTurnNum,
@@ -200,7 +208,7 @@ describe('Consumes the expected gas for sad-path exits', () => {
     await expect(challengeTx).toConsumeGas(
       gasRequiredTo.ETHexitSadLedgerFunded.vanillaNitro.challengeX
     );
-    const finalizesAt = await getFinalizesAtFromTransactionHash((challengeTx as Transaction).hash);
+    const finalizesAt = await getFinalizesAtFromTransactionHash(challengeTx.hash);
     // begin wait
     await waitForChallengesToTimeOut([ledgerFinalizesAt, finalizesAt]); // just go to the max one
     // end wait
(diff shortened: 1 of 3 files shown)


packages/nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -71,28 +73,47 @@ export const gasRequiredTo: GasRequiredTo = {
...
+      challengeL: 92350,
+      challengeG: 94645,
+      challengeJ: 101748,
+      challengeX: 93404,
+      pushOutcomeAndTransferAllL: 58652,
+      pushOutcomeG: 61422,
+      pushOutcomeJ: 60558,
+      claimG: 58856,
+      pushOutcomeAndTransferAllX: 107742,
+      total: 729377,
```
geoknee, line 116: TLDR it costs at least 729K gas to liquidate a virtually funded channel, in the worst case.

NiloCK: This is quite a bit and potentially makes ZK verification costs (~400-600k) look less insane. IF a proof could batch the channel closures somehow (rather than proof-per closure) there could be a substantial saving.

andrewgordstewart: Most of these operations would disappear by optimizing:
- [before](https://www.notion.so/statechannels/Virtual-funding-optimizations-16998f6d3baa41a49dd69b9948721b66#380ea45d9b2c4c93abda7dc6734691c8) 7 operations
- [after](https://www.notion.so/statechannels/Virtual-funding-optimizations-16998f6d3baa41a49dd69b9948721b66#4199ddc61f1d495fb6374171381b72cb): 3 operations

**Note**: These numbers do not account for `pushOutcome` operations. `pushOutcome` can be removed by changing the asset holder/adjudicator architecture.

andrewgordstewart: This isn't to discount the possibility that batching channel closures can't provide a large improvement!

Batching the closure of many joint channels would indeed help. But, that's a can of worms: if challenges are batched, then is the channel status still visible? Are there new griefing attacks?

packages/nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -21,129 +23,208 @@ import {Bytes} from '../src/contract/types';
...
+// Virtual funding
+export const J = new TestChannel(
+  5,
+  [Alice, Bob, Ingrid],
+  [
+    {destination: X.channelId, amount: '0xa'},
+    {destination: convertAddressToBytes32(Ingrid.address), amount: '0xa'},
+  ]
+);
+export const G = new TestChannel(6, [Alice, Ingrid], {
```
NiloCK, line 183:
```suggestion

/** Guarantor channel between Alice and Ingid, funding application channel X */
export const G = new TestChannel(6, [Alice, Ingrid], {
```

packages/nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -21,129 +23,208 @@ import {Bytes} from '../src/contract/types';
...
+  targetChannelId: J.channelId,
+  destinations: [
+    convertAddressToBytes32(Alice.address),
+    convertAddressToBytes32(Bob.address),
```
NiloCK, line 187: Probably immaterial with respect to the gas cost tests, but shouldn't this guarantor channel be funding [alice, ingrid, X] rather than [alice, bob, X]?

packages/nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -21,129 +23,208 @@ import {Bytes} from '../src/contract/types';
...
   await provider.send('evm_mine', []);
 }
 
-export const ledgerChannel: Channel = {chainId, channelNonce: channelNonce + 10000, participants};
-export const ledgerChannelId = getChannelId(ledgerChannel);
+/**
+ * Constructs a support proof from the supplied state; calls challenge, asserts the expected gas and returns the proof and finalizesAt
+ * @param state
+ * @param expectedGas
+ * @returns
```
NiloCK, line 208:
```suggestion
 * Constructs a support proof for the supplied channel, calls challenge,
 * and asserts the expected gas
 * @returns The proof and finalizesAt
```

packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -171,64 +146,154 @@ describe('Consumes the expected gas for sad-path exits', () => {
...
+        X.channelId,
+        proof.largestTurnNum,
+        finalizesAt, // finalizesAt
+        proof.stateHash, // stateHash
+        proof.challengerAddress, // challengerAddress
+        proof.outcomeBytes // outcomeBytes
+      )
+    ).toConsumeGas(gasRequiredTo.ETHexitSadVirtualFunded.vanillaNitro.pushOutcomeAndTransferAllX);
+    // pushOutcomeAndTransferAllX  ⬛ -----------------------------> 👩
+    expect(
```
NiloCK, line 293:
```suggestion

    // meta-test here to confirm the total recorded in gas.ts is up to date
    // with the recorded costs of each step
    expect(
```

NiloCK (approved):
Looks great.

packages/nitro-protocol/gas-benchmarks/fixtures.ts
geoknee: Good catch. Fixed 91b0f75 (did cause some small changes to gas).

packages/nitro-protocol/gas-benchmarks/gas.ts
geoknee: It looks like we can save a lot of gas with some minor changes, even without the more adventurous VFP optimizations.

 I'm not sure I see a problem with having a `batchChallenge` method? This would simply accept an array of `challenge` calls. That should save at least 3x21K = 63K gas (the base fees).

And it looks like we would save at least another 120K by [deprecating `pushOutcome`](https://www.notion.so/statechannels/Deprecate-discrete-AssetHolders-884d5728716d41ef9fc2d13c5d5d3281). 

Together, a >~25% saving without breaking a sweat.

PR #3603, opened 2021-06-15
upgrade ethers to 5.3.1

geoknee:
Leaving @ethersproject/experimental where it is (I had trouble when upgrading that package).


geoknee:
In favour of #3604 , which is based off this branch anyway.

PR #3604, opened 2021-06-15
Add typechain to nitro-protocol

geoknee:
https://www.notion.so/statechannels/Add-TypeChain-tooling-f17342226cb44244be4820b9937c9fae

This PR brings us to a state where we can do this: 

Screenshot 2021-06-07 at 13 11 49

I didn't spread the types around the codebase much, yet. For now I just used them in the gas benchmarking test. We can do more of that kind of thing in future PRs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,7 @@ jobs:
             - packages/*/dist
             - packages/*/build
             - packages/nitro-protocol/artifacts
+            - packages/nitro-protocol/typechain
             - packages/*/lib
             - packages/*/node_modules
             - packages/client-api-schema/types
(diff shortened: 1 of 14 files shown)


packages/nitro-protocol/gas-benchmarks/vanillaSetup.ts
```diff
@@ -21,10 +24,10 @@ declare global {
   }
 }
 
-export let ethAssetHolder: Contract;
-export let erc20AssetHolder: Contract;
-export let nitroAdjudicator: Contract;
-export let token: Contract;
+export let ethAssetHolder: ETHAssetHolder & Contract;
```
geoknee, line 27: I don't think this intersection type is the canonical way to use typechain. But it works well!

.circleci/config.yml
```diff
@@ -112,6 +112,7 @@ jobs:
             - packages/*/dist
             - packages/*/build
             - packages/nitro-protocol/artifacts
+            - packages/nitro-protocol/typechain
```
NiloCK, line 115: Clarifying circle config:

This is writing these directories into the filesystem of circleCI's vm, and not persisting in the 'artifact' sense? It writes these to make them available for the later test job.

geoknee: Exactly!

NiloCK (approved):
This is great! The gas benchmarks, along with typechain, serve as a much improved walk-through of nitro workflows with great discoverability.

I ~am having~ had some trouble running this locally. For reference, things were cleared up after some `yarn clean`, `yarn`, and closing out *all* VSCode windows, which seemed to be hanging onto some out of date cached typings.

PR #3608, opened 2021-06-16
introduce addResidualTokenBalance fn

geoknee:
It simply funds someOtherChannel with tokens.

Also remove that step from tests that work with ETH only.


NiloCK (approved):
Great!

PR #3612, opened 2021-06-24
[docs-website] remove mention of "thumbprint"

geoknee:
This is a hangover from an old terminology we no longer use,
and is inaccurate / confusing.


NiloCK (approved):
I've never heard of this! Out, damn thumbprint, out!

PR #3613, opened 2021-06-24
[nitro-protocol] Do not cast enum type to uint8

geoknee:
This is unnecessary.

The change also saves a tiny bit of gas.


NiloCK (approved):
LGTM. I'm a little curious about the gas savings. As far as I can tell Solidity's enum implementation is backed by a uint8, so the differences here seems like something that could / should be erased during compilation.

:man_shrugging:

geoknee:
Yeah I don't really understand it either.

PR #3625, opened 2021-06-29
Replace AssetHolder contracts with MultiAssetHolder contract

geoknee:
See [this pitch](https://www.notion.so/statechannels/Deprecate-discrete-AssetHolders-884d5728716d41ef9fc2d13c5d5d3281). This PR contains changes to `.sol` files only. I think it makes sense to get reviews on the contract API before moving forward with the supporting typescript code. 

The new architecture features **only** a singleton `NitroAdjudicator` contract, instead of that singleton plus n `AssetHolder` contracts. 

The n `AssetHolders` have been combined into a single `MultiAssetHolder` base contract which the `NitroAdjudicator` now extends. The `deposit`, `transfer`, and `claim` methods have been preserved, but now take an `assetIndex` which is used to slice into the outcome stored on chain. The slice gives an `AssetOutcome`, which is what the old `AssetHolders` stored. The contract methods manipulate these `AssetOutcomes` as before, but updated the **entire** outcome.

The `NitroAdjudicator` method `pushOutcome` has been removed, including from the name and implementation of compound methods such as `concludePushOutcomeAndTransferAll`.

We will have a linear inheritance relationship, since the `MultiAssetHolder` needs to make changes to the `statusOf` mapping defined in `ForceMove` contract. The inheritance graph is `ForceMove => MultiAssetHolder => NitroAdjudicator` (`base => derived`). 

The new set of contracts compiles and deploys, but many existing tests and source files are broken. These will be fixed in future PRs (the work has been done already). 

The contracts are also not gas optimised. This will also be addressed in future PRs.

⚠️ failing tests should not block this PR being merged. The base branch is not master.


packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -2,104 +2,220 @@
...
+        );
+
+        Outcome.OutcomeItem[] memory outcome = abi.decode(outcomeBytes, (Outcome.OutcomeItem[]));
+        Outcome.AssetOutcome memory assetOutcome = abi.decode(
+            outcome[assetIndex].assetOutcomeBytes,
+            (Outcome.AssetOutcome)
+        );
+        require(
+            assetOutcome.assetOutcomeType == Outcome.AssetOutcomeType.Allocation,
+            '!allocation'
```
kerzhner, line 110: This is probably copy/pasted, but might be nice to have a friendlier message.
```suggestion
            'Outcome type must be an allocation'
```

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -2,104 +2,220 @@
...
+        );
+        Outcome.AllocationItem[] memory allocation = abi.decode(
+            assetOutcome.allocationOrGuaranteeBytes,
+            (Outcome.AllocationItem[])
+        );
+        address asset = outcome[assetIndex].assetHolderAddress;
+
         // effects and interactions
-        _transfer(fromChannelId, allocationBytes, indices);
+        allocation = _transfer(asset, fromChannelId, allocation, indices); // update in place to newAllocation
```
kerzhner, line 119: Do you know what happens in the case where the transfer leaves the allocation empty (in other words, after the transfer, the `allocationItem` array contains no more entries)?

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -2,104 +2,220 @@
...
-    // Permissioned methods
-    // **************
-
-    modifier AdjudicatorOnly {
-        require(msg.sender == AdjudicatorAddress, 'Only NitroAdjudicator authorized');
-        _;
-    }
+        Outcome.Guarantee memory guarantee;
+        address asset;
+        {
```
kerzhner, line 147: Curious, what is the reason there are separate blocks in this function?

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -2,104 +2,220 @@
...
-        _setAssetOutcomeHash(channelId, assetOutcomeHash);
+            // effects and interactions
+            allocation = _claim(asset, guarantorChannelId, guarantee, allocation, indices);
+            outcome[assetIndex].assetOutcomeBytes = abi.encode(
+                Outcome.AssetOutcome(Outcome.AssetOutcomeType.Allocation, abi.encode(allocation))
+            );
+            bytes32 outcomeHash = keccak256(abi.encode(outcome));
+            _updateFingerprint(
+                guarantee.targetChannelId,
+                targetStateHash,
```
kerzhner, line 213: Is it an issue that we update the outcome but do not update the state hash?

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -221,23 +337,20 @@ contract AssetHolder is IAssetHolder {
...
-            (Outcome.AllocationItem[])
-        );
-        uint256 initialHoldings = holdings[fromChannelId];
+    ) internal returns (Outcome.AllocationItem[] memory) {
+        uint256 initialHoldings = holdings[asset][fromChannelId];
 
         (
-            Outcome.AllocationItem[] memory newAllocation,
-            bool safeToDelete,
+            Outcome.AllocationItem[] memory newAllocation, // safeToDelete
```
kerzhner, line 352: Do you mean the next field is `safeToDelete`?

kerzhner (approved):
Looks good in general! Are we planning on replacing `pushOutcome` with any other methods? Or do we think that `pushOutcome` functionality is no longer needed.

packages/nitro-protocol/contracts/NitroAdjudicator.sol
```diff
@@ -2,87 +2,17 @@
...
 pragma experimental ABIEncoderV2;
 
-import './interfaces/IAdjudicator.sol';
 import './ForceMove.sol';
 import './Outcome.sol';
-import './AssetHolder.sol';
+import './MultiAssetHolder.sol';
 
 /**
  * @dev The NitroAdjudicator contract extends ForceMove and hence inherits all ForceMove methods, and also extends and implements the Adjudicator interface, allowing for a finalized outcome to be pushed to an asset holder.
```
kerzhner, line 10: This comment seems stale now

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: I think this would trip our linting rules, since we are aiming to have reasons strings shorter than 32 characters (for gas reasons). There is probably something shorter that is still friendly, though. In future we could move to having error codes instead of messages.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -2,104 +2,220 @@
...
     /**
      * @notice Transfers as many funds escrowed against `channelId` as can be afforded for a specific destination. Assumes no repeated entries.
      * @dev Transfers as many funds escrowed against `channelId` as can be afforded for a specific destination. Assumes no repeated entries.
      * @param fromChannelId Unique identifier for state channel to transfer funds *from*.
-     * @param allocationBytes The abi.encode of AssetOutcome.Allocation
+     * @param outcomeBytes The encoded Outcome of this state channel
      * @param indices Array with each entry denoting the index of a destination to transfer funds to. An empty array indicates "all".
      */
     function transfer(
+        uint256 assetIndex, // TODO consider a uint48?
```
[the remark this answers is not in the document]
geoknee: That's an interesting suggestion, and is relevant over in the exit-format repo. 

I think a dictionary might be easier to work with, but I don't think it is possible to implement, sadly. 

In[ typescript](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype), I think your idea would look something like this:

```ts
type Outcome = Record<Address, AssetOutcome>
````
In solidity we can use a [mapping](https://docs.soliditylang.org/en/v0.8.6/types.html#mapping-types) to get something similar, but that only works for `storage` variable. We would then be using a separate storage slot for each `AssetOutcome`, which is actually the very thing we are trying to avoid doing here. It's also not clear how we can encode a mapping, which is something we need (in order to store the hash, etc). 

The only other thing we could try to do is have:

```solidity
struct Outcome {
    AssetOutcome 0x000;
    AssetOutcome 0xabc;
}
```
but this doesn't work either, for a couple of reasons. The keys are invalid (addresses are not valid identifiers), and they have to be specified at compile-time (we need them to be specified at runtime).

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: I've found it much easier to operate "in place" on the allocation, or to build up a copy of it with the same length. This means that "zero allocations" (e.g. `Alice: 0`) are included in the outcome and not trimmed out. 

Some "clean up" is still possible, by tracking some "safe to delete" flags. When we know the only remaining allocations are for zero, we can just delete that `AssetOutcome` entirely. That's what happens on `master`. On this branch, we don't store `AssetOutcomes` separately so we can't do that cleanup.

I think we should circle back to this question once we get the project's goals finished, and look to optimise the behaviour. For example, if all `AssetOutcomes` are safe to delete, we can delete the `statusOf` entry for the channel (see https://github.com/statechannels/statechannels/pull/3394).

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: They are there to prevent "stack too deep" problems. See point 2 of this page https://soliditydeveloper.com/stacktoodeep .

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: I don't think so. As long as it is clear that the `stateHash` is a record of the latest state the chain is aware of, and it can only change (on chain) *before* the channel finalizes. After finalization, the `outcome` can change (on chain) without the `stateHash` changing. 

Of course the `stateHash` refers to an outcome, and one might think that implies some redundancy -- why not simply update the outcome in the `stateHash`? I think what we have now is more efficient, because we only have to supply the `stateHash` and the preimage of the `outcomeHash` in the calldata for `transfer`. If we only stored that `stateHash` and applied transformations to that directly, we would need to supply the entire preimage of the `stateHash` every time.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: Yes. I tried to put the comment on the next line, but prettier keeps moving it!

I will just remove the comment for now.

geoknee:
Thanks for the review @kerzhner !

packages/nitro-protocol/contracts/MultiAssetHolder.sol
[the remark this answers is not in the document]
kerzhner: That makes sense. Thanks for the enlightening explanation!

NiloCK (approved):
after-the-fact comments!

packages/nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * @dev Indicates that `amountDeposited` has been deposited into `destination`.
+     * @param destination The channel being deposited into.
+     * @param amountDeposited The amount being deposited.
+     * @param destinationHoldings The new holdings for `destination`.
+     */
+    event Deposited(
+        bytes32 indexed destination,
+        address asset,
```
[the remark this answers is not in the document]
geoknee: Interesting one. I'll hold off on this now as it won't be hard to put in later if we think there is a need.

PR #3628, opened 2021-07-01
update gas benchmarking code

geoknee:
...to reflect removal of pushOutcome. 

This PR shows off the new flows enabled by the new architecture. 

It also shows off the fact that the new implementation is not gas optimised; although we still get some gas savings by removing `pushOutcome` I expect there is quite a lot of room for improvement in future PRs. 

⚠️ base not master. 

Expected to pass:
* `yarn contract:compile` in nitro-protocol
* `yarn test:gas` in nitro-protocol

Expected to fail
* everything else

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/gas.ts
+++ b/packages/nitro-protocol/gas-benchmarks/gas.ts
@@ -28,17 +28,15 @@ type Path =
 export const gasRequiredTo: GasRequiredTo = {
   deployInfrastructureContracts: {
     vanillaNitro: {
-      NitroAdjudicator: 2416171, // Singleton
-      ETHAssetHolder: 1636603, // Singleton (could be more in principle)
-      ERC20AssetHolder: 1659990, // Per Token (could be more in principle)
+      NitroAdjudicator: 3954624, // Singleton
     },
   },
   directlyFundAChannelWithETHFirst: {
-    vanillaNitro: 47608,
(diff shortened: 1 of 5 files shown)


packages/nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -28,17 +28,15 @@ type Path =
 export const gasRequiredTo: GasRequiredTo = {
   deployInfrastructureContracts: {
     vanillaNitro: {
-      NitroAdjudicator: 2416171, // Singleton
-      ETHAssetHolder: 1636603, // Singleton (could be more in principle)
-      ERC20AssetHolder: 1659990, // Per Token (could be more in principle)
+      NitroAdjudicator: 3954624, // Singleton
```
kerzhner, line 31: Curious, what is the source of these numbers?

kerzhner (approved):

packages/nitro-protocol/gas-benchmarks/gas.ts
geoknee: All the numbers in this file are checked in the neighbouring `benchmark.test.ts` file:

https://github.com/statechannels/statechannels/blob/a755ed16e2b96550feec13ea94040e20f6d08bb7/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts#L33-L37

PR #3629, opened 2021-07-01
update transaction-creators and other TS code

geoknee:
Updates the typescript part of nitro-protocol package to reflect the new architecture without asset holders. 

The overall API is changed a bit: 

Events
---
Notably, the event emitted during a `transfer` or `claim` has changed. Previously, this was an `AllocationUpdated` event, but now it is a `FingerprintUpdated`. The purpose of the event is the same: to allow clients to update their off-chain records such that they can supply the requisite outcome preimage(s) for subsequent `transfers` or `claims`. The difference is that now the singleton `NitroAdjudicator` stores a hash of the entire `Outcome`, rather than individual `AssetHolders` storing a hash of "their slice" (i.e. their `AssetOutcome`). The TS helpers that help clients consume these events and update their records has been updated. 

Transactions
---
The transaction creators have been updated to follow the new contract API (no more `pushOutcome`). 

⚠️ base not master. 

Expected to pass:
* `yarn contract:compile` in nitro-protocol
* `yarn test:gas` in nitro-protocol
* `yarn prepare` in nitro-protocol 🆕 

Expected to fail
* everything else


kerzhner (approved):
Looks good to me!

PR #3630, opened 2021-07-01
refactor deposit tests

geoknee:
Combines the two existing deposit tests (each targeting their own `AssetHolder` contract) into a single test targeting an `is MultiAssetHolder` contract (in this case the `NitroAdjudicator`).


packages/nitro-protocol/test/contracts/MultiAssetHolder/deposit.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  `('$description', async ({asset, held, expectedHeld, amount, reasonString, heldAfter}) => {
+    held = BigNumber.from(held);
+    expectedHeld = BigNumber.from(expectedHeld);
+    amount = BigNumber.from(amount);
+    heldAfter = BigNumber.from(heldAfter);
+
+    const destinationChannel: Channel = {chainId, channelNonce, participants};
+    const destination = getChannelId(destinationChannel);
+
+    if ((asset = ERC20)) {
```
NiloCK, line 78:
```suggestion
    if (asset === ERC20) {
```

NiloCK (approved):
:+1:

packages/nitro-protocol/test/contracts/MultiAssetHolder/deposit.test.ts
geoknee: good spot!

PR #3631, opened 2021-07-02
point AssetHolder tests at MultiAssetHolder

geoknee:
This is just a simple change so that the existing test suite can run against the new architecture.


NiloCK (commented):
Looks good.

PR #3635, opened 2021-07-02
Update browser wallet to singleton contract architecture

geoknee:
Gets the browser wallet building and its (unskipped) tests passing.

⚠️ base not master. Towards #3624

### DIFF
--- a/packages/browser-wallet/src/chain.ts
+++ b/packages/browser-wallet/src/chain.ts
@@ -287,14 +287,6 @@ export class ChainWatcher implements Chain {
   private configureContracts() {
     if (!this.ethereumIsEnabled) return;
 
-    this._assetHolders = [
-      new Contract(
-        ETH_ASSET_HOLDER_ADDRESS,
-        ContractArtifacts.EthAssetHolderArtifact.abi,
-        this.signer
-      )
-    ];
-
     // Log all contract events (for now)
     this._assetHolders[0].on('*', event => chainLogger.trace({event}, 'assetHolder[0] event'));
 
@@ -355,7 +347,7 @@ export class ChainWatcher implements Chain {
 
   public async finalizeAndWithdraw(finalizationProof: SignedState[]): Promise<string | undefined> {
     const transactionRequest = {
-      ...Transactions.createConcludePushOutcomeAndTransferAllTransaction(
+      ...Transactions.createConcludeAndTransferAllAssetsTransaction(
         finalizationProof.flatMap(toNitroSignedState)
       ),
       to: NITRO_ADJUDICATOR_ADDRESS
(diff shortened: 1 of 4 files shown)


packages/browser-wallet/src/chain.ts
```diff
@@ -333,6 +318,7 @@ export class ChainWatcher implements Chain {
       } catch (error) {
         // TODO: Handle error. Likely the user rejected the login
         chainLogger.error(error);
+        console.log(error);
```
NiloCK, line 321:
```suggestion
```
I'm making a guess that this was a temporary convenience, and that the existing logger statement is OK on its own.

packages/browser-wallet/src/tests-with-chain/chain.test.ts
```diff
@@ -12,32 +12,18 @@ import {
...
-  chain.ethereumEnable();
-
-  const signer = await provider.getSigner('0x28bF45680cA598708E5cDACc1414FCAc04a3F1ed');
-  ETHAssetHolder = new Contract(
-    ETH_ASSET_HOLDER_ADDRESS,
-    ContractArtifacts.EthAssetHolderArtifact.abi,
-    signer
-  );
+  await chain.ethereumEnable();
 });
```
NiloCK, line 27: Why is it that `NitriAdjudicator` doesn't need to be initialized here in the same way that `ETHAssetHolder` was?

packages/browser-wallet/src/utils/contract-utils.ts
```diff
@@ -30,12 +27,3 @@ export function getProvider(): providers.Web3Provider | providers.JsonRpcProvide
...
 }
-
-export function getEthAssetHolderContract() {
-  const provider = getProvider();
-  return new Contract(
-    ETH_ASSET_HOLDER_ADDRESS || '0x00',
-    ContractArtifacts.EthAssetHolderArtifact['abi'],
-    provider
-  );
-}
```
NiloCK, line 41: As far as I can tell this function wasn't used anywhere anyway... ?

NiloCK (approved):
I'm enjoying this whirlwind tour!

packages/browser-wallet/src/chain.ts
geoknee: Your guess is spot on.

packages/browser-wallet/src/tests-with-chain/chain.test.ts
geoknee: Well the truth is that it does... only the test in question is currently being skipped so it "doesn't matter". I presume the test is being skipped because it fails. But I will double check; and it might be nice to not make it more broken with this changeset 😛

geoknee: 0a31d6b

PR #3639, opened 2021-07-05
gas optimise `transferAllAssets`

geoknee:
The existing implementation of this function calls into an "outer" `transfer` function in a loop. This is inefficient because for every asset
* indices are checked for validity (monotonically increasing) unnecessarily
* events are emitted
* storage is checked and written to

This PR has the function in question call in to the inner `_transfer` function in the loop, and then has the requisite checks and writes to storage performed once before/after the loop.

~The commit history will show that~ I flirted with deleting the `statusOf` storage variable as a further optimisation, but decided against doing that now (see #3394).

### DIFF
--- a/packages/nitro-protocol/contracts/MultiAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
@@ -116,7 +116,7 @@ contract MultiAssetHolder is IMultiAssetHolder, ForceMove {
         address asset = outcome[assetIndex].asset;
 
         // effects and interactions
-        allocation = _transfer(asset, fromChannelId, allocation, indices); // update in place to newAllocation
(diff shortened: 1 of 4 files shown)


NiloCK (approved):
This LGTM. I spent a bit of time fiddling with particulars seeking a person gain of intuition on gas costs, and also looking at refactoring `_updateFingerprint` to include the event emit. (As is, that function is always preceded and followed by the same code - a smell, but maybe necessary according to the laws of ethereum gassiness).

geoknee:
I think there is definitely scope for further improvements to this and surrounding code. I think of it being a three stage process:

1. refactor to preserve behaviour and get tests passing (done)
2. refactor to reduce gas consumption (mostly done in this PR, but always the potential of further reductions)
3. refactor to reduce repetition, readability, security.... (not done yet, and the most open ended)

PR #3642, opened 2021-07-06
swap `assetHolderAddress` column for `isLedgerChannel` column in channels table

geoknee:
Our server wallet database schema has a `channels` table with an `assetHolderAddress` column.  This is currently only used for flagging when a channel is a ledger channel.  This is needlessly indirect and potentially confusing, as well as making refactoring the contract architecture difficult (what if there aren't any asset holders). 

This approach is good because the new column name and data type more accurately reflect the semantic meaning of the information being stored. It also makes #3624 easier to achieve. 

There's one point of contention, which is the way I have updated the db migrations here. I have just edited an existing migration file rather than adding a new migration to rename the column name. I figured this was preferable because we have no DBs in production right now, and we already have many meanders in the list of migrations that make it more difficult to infer what the actual db schema is. 

### Changes 
DB schema modified
Channel model modified
isLedger, setLedger etc changed to not require an asset holder address.

---
Aside: My guess for the original motivation for using asset holder address to stand for a ledger channel is as follows: 
* ledger channels must be directly funded (incorrect); 
* ledger channels can only have one asset holding contract (incorrect)
* application channels will not be directly funded (incorrect); 
* being directly funded means that there is an contract holding assets for the channel (correct); 
* being indirectly funded means there is no contract holding assets for the channel (mostly incorrect, because if I have to liquidate ledger or virtual channels via the sad path they become directly funded). 

So generally based on (at best) a long list of implicit assumptions or (at worst) things that are simply untrue.

---
### Code quality
### Project management

### DIFF
--- a/packages/server-wallet/src/models/channel.ts
+++ b/packages/server-wallet/src/models/channel.ts
@@ -48,8 +48,9 @@ export const REQUIRED_COLUMNS = [
   'participants',
   'vars',
   'fundingStrategy',
+  'isLedgerChannel',
 ] as const;
-export const OPTIONAL_COLUMNS = ['assetHolderAddress', 'fundingLedgerChannelId'] as const;
+export const OPTIONAL_COLUMNS = ['fundingLedgerChannelId'] as const;
 export const COMPUTED_COLUMNS = ['channelId', 'signingAddress'] as const;
 
 export const CHANNEL_COLUMNS = [...REQUIRED_COLUMNS, ...COMPUTED_COLUMNS, ...OPTIONAL_COLUMNS];
@@ -63,6 +64,7 @@ export interface RequiredColumns {
   readonly vars: SignedStateVarsWithHash[];
   readonly signingAddress: Address;
   readonly fundingStrategy: FundingStrategy;
+  readonly isLedgerChannel: boolean;
 }
 
 export type ComputedColumns = {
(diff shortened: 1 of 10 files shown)


packages/server-wallet/src/models/channel.ts
```diff
@@ -394,7 +389,7 @@ export class Channel extends Model implements ChannelColumns {
   }
 
   public get isLedger(): boolean {
-    return !!this.assetHolderAddress;
+    return this.isLedgerChannel;
```
geoknee, line 392: This is the motivation for this PR, in a nutshell.

packages/server-wallet/src/models/channel.ts
```diff
@@ -173,28 +175,21 @@ export class Channel extends Model implements ChannelColumns {
...
   }
 
-  static async setLedger(
-    channelId: Bytes32,
-    assetHolderAddress: Address,
-    txOrKnex: TransactionOrKnex
-  ): Promise<void> {
-    await Channel.query(txOrKnex).findOne({channelId}).patch({assetHolderAddress});
+  static async setLedger(channelId: Bytes32, txOrKnex: TransactionOrKnex): Promise<void> {
+    await Channel.query(txOrKnex).findOne({channelId}).patch({isLedgerChannel: true});
```
andrewgordstewart, line 179: aside: this function is _only ever used during testing_, and seems like it doesn't belong in production code

andrewgordstewart (approved):
This seems like an excellent change to make regardless of its original motivation!

PR #3643, opened 2021-07-06
Do not return outcome from transfer method

geoknee:
Just cleaning up this old bit of wiring that we no longer need.


NiloCK (approved):
Good call on relabelling as external as well.

PR #3644, opened 2021-07-06
Remove duplicate call to abi.encode

NiloCK:
... a small refactor to save a little gas.

`yarn test:gas`

### DIFF
--- a/packages/nitro-protocol/contracts/MultiAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
@@ -207,14 +207,15 @@ contract MultiAssetHolder is IMultiAssetHolder, ForceMove {
             outcome[assetIndex].assetOutcomeBytes = abi.encode(
                 Outcome.AssetOutcome(Outcome.AssetOutcomeType.Allocation, abi.encode(allocation))
             );
-            bytes32 outcomeHash = keccak256(abi.encode(outcome));
+            bytes memory outcomeBytes = abi.encode(outcome);
+            bytes32 outcomeHash = keccak256(outcomeBytes);
             _updateFingerprint(
                 guarantee.targetChannelId,
                 targetStateHash,
                 targetChallengerAddress,
                 outcomeHash
             );
-            emit FingerprintUpdated(guarantee.targetChannelId, abi.encode(outcome));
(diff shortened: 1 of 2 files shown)


geoknee (approved):
👏

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -207,14 +207,15 @@ contract MultiAssetHolder is IMultiAssetHolder, ForceMove {
...
+            bytes memory outcomeBytes = abi.encode(outcome);
+            bytes32 outcomeHash = keccak256(outcomeBytes);
             _updateFingerprint(
                 guarantee.targetChannelId,
                 targetStateHash,
                 targetChallengerAddress,
                 outcomeHash
             );
-            emit FingerprintUpdated(guarantee.targetChannelId, abi.encode(outcome));
+            emit FingerprintUpdated(guarantee.targetChannelId, outcomeBytes);
```
andrewgordstewart, line 218: Are `FingerprintUpdated` events new? These events are rather expensive to emit, costing 8 gas per byte.

If things are factored a certain way, where the resulting transformation is computed by a pure function, then a wallet can use the input to `claim` to compute the new outcome. This is the general pattern used in current-nitro, except [here](https://github.com/statechannels/statechannels/blob/546d164eeeb08efc877916b3886431001f88e993/packages/nitro-protocol/contracts/ForceMove.sol#L92-L103) -- this log statement in particular costs ~20k gas, vs the ~500 gas for most other log statements currently triggered.

geoknee: Yes they are new, and replace the old `AllocationUpdated` events (see #3625).  You're right that we can gas optimise the event emission, and I think we should do that. But it might save a heap of work to wait until we have moved over to the exit format, since the pure functions you talk about already exist 

https://github.com/statechannels/exit-format/blob/6c00a7d9e34ee03271984824f3cdf5a7096ec641/contracts/Nitro.sol#L161-L172

andrewgordstewart: > But it might save a heap of work to wait until we have moved over to the exit format, since the pure functions you talk about already exist

Sounds like a plan!

PR #3645, opened 2021-07-06
Cleanup deploy script in server-wallet

geoknee:
and also take out a few "assetHolderAddress" low hangers.


packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
```diff
@@ -111,12 +112,12 @@ beforeAll(async () => {
   const logger = createLogger(defaultTestWalletConfig());
 
   TestMessageService.linkMessageServices(a.messageService, b.messageService, logger);
-  const assetHolder = new Contract(
-    ethAssetHolderAddress,
-    ContractArtifacts.EthAssetHolderArtifact.abi,
+  const nitroAdjudicator = new Contract(
+    nitroAdjudicatorAddress,
```
NiloCK, line 116: Is ethers OK with this empty string?

Confirming side note: branch as-is contains conflicts and leftover references to assetHolderAddress and the eth/erc20 assetholder contracts, correct?

NiloCK (approved):
Looks good.

packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts
geoknee: We only have an empty string in the *production* environment (since we didn't deploy the new contract to a testnet or mainnet yet). Tests such as this one  still pass because our test environment provides an actual address (from ganache)

geoknee: Yes I think the branch as is is based on some slightly older commit, so it may not have all the changes we did recently. I'm not seeing any conflicts, though? We should get all the changes we want when merging this.

PR #3647, opened 2021-07-07
Remove (another) duplicate abi.encode call

NiloCK:
This was spotted by Andrew and is similar to the previous PR.

Notable:
- this change **only** had effects on the deploy gas cost. It looks like the benchmarking suite takes the `transferAllAssets` route for each case. In the happy-path it gets there via [concludeAndTransferAllAssets](https://github.com/statechannels/statechannels/blob/cfba1f6975cbdd85958254ff8abb54c9722fc351/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts#L91), and [directly](https://github.com/statechannels/statechannels/blob/cfba1f6975cbdd85958254ff8abb54c9722fc351/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts#L126) in the sad-path. Do we consider this a gap in the benchmarking?
- unlike the previous PR, this one re-uses the `outcomeBytes memory` function parameter, which I think in context is OK, but I thought worth pointing out in case it isn't

`yarn prepare && yarn test:gas`

### DIFF
--- a/packages/nitro-protocol/contracts/MultiAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
@@ -120,10 +120,11 @@ contract MultiAssetHolder is IMultiAssetHolder, ForceMove {
         outcome[assetIndex].assetOutcomeBytes = abi.encode(
             Outcome.AssetOutcome(Outcome.AssetOutcomeType.Allocation, abi.encode(allocation))
         );
-        bytes32 outcomeHash = keccak256(abi.encode(outcome));
+        outcomeBytes = abi.encode(outcome);
+        bytes32 outcomeHash = keccak256(outcomeBytes);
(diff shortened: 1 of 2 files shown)


geoknee (approved):

geoknee:
>  Do we consider this a gap in the benchmarking?
I suppose yes, but there's a question about how "comprehensive" we ought to or need to be with that. Perhaps the rule of thumb might be to record the lowest possible gas cost given the assumptions about counterparty behaviour (i.e. happy vs sad). If Alice can spend less gas to just get *her* funds back (using `transfer` instead of `transferAllAssets`), then maybe that is the number we ought to record?

> * unlike the previous PR, this one re-uses the `outcomeBytes memory` function parameter, which I think in context is OK, but I thought worth pointing out in case it isn't
I also believe this is OK.

PR #3648, opened 2021-07-07
update chain service to not use asset holders or pushoutcome

geoknee:
Tests that pass on this branch:
`chain-service.test.ts` ( but only if `yarn link @statechannels/native-utils` against https://github.com/statechannels/native-utils/pull/26 )
`stress.test.ts` (confirmed manually, it doesn't run by default)

### DIFF
--- a/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
+++ b/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
@@ -57,7 +57,7 @@ export function createConcludeAndTransferAllAssetsTransaction(
   };
 }
 
-export function transferAllAssetsArgs(
+function transferAllAssetsArgs(
   state: State,
   challengerAddress: string,
   overrideStateHash = false // set to true if channel concluded happily
--- a/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
+++ b/packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
@@ -1,6 +1,7 @@
-import {utils, providers, Signature} from 'ethers';
+import {utils, providers, Signature, constants} from 'ethers';
 
 import NitroAdjudicatorArtifact from '../../../artifacts/contracts/NitroAdjudicator.sol/NitroAdjudicator.json';
+import {getChannelId, hashState} from '../../';
 import {encodeOutcome} from '../outcome';
 import {getFixedPart, hashAppPart, State} from '../state';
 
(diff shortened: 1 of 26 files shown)


packages/server-wallet/src/chain-service/event-tracker.ts
```diff
@@ -66,15 +66,21 @@ export class EventTracker {
...
   }
 
   // Pass event to managed subscriber only if new
-  assetOutcomeUpdated(arg: AssetOutcomeUpdatedArg, blockNumber: number, logIndex: number): void {
-    if (this.isNewEvent(arg.assetHolderAddress, blockNumber, logIndex)) {
-      this.managedSubscriber.assetOutcomeUpdated(arg);
+  fingerprintUpdated(arg: FingerprintUpdatedArg, blockNumber: number, logIndex: number): void {
+    if (
+      this.isNewEvent(
+        arg.channelId, // TODO does this make sense to put here?
```
geoknee, line 78: Flagging for discussion.

NiloCK: This file, and the chain service in general, are new to me. My shallow understanding right now looks something like:

- the `assetMap`, formerly `assetHolderMap` (line 22) is used to store the most-recent-event.
- the flagged line proposes to use the channelId as the key to index these events, since they all now originate from the same MAH/Adjudicator


Comments:
- as far as I saw, these EventTrackers [belong to individual channels](https://github.com/statechannels/statechannels/blob/f8779927900fa14838e46e11008b29b9e24da398/packages/server-wallet/src/chain-service/chain-service.ts#L62). Is this indexing redundant?
- `assetMap` probably isn't a useful name for this storage mechanism anymore?

Aside:
- the doccomments for `EventTracker` are not up-to-date

NiloCK (commented):
Abandoning review part-way. Need to drop my dog to the border. Will pick up tonight.

packages/server-wallet/src/chain-service/event-tracker.ts
NiloCK: Having come back to this without a clock ticking at me:

In the chain service each channel has its own collection of EventTrackers. The `assetMap` keys here are meant to keep separate event queues for the different potential emitters. The `channelId` serves as something distinct from the asset address keys, but so would any constant string that wasn't an asset address.

If the key here were `"fingerprint"` it might be more obvious that these fingerPrintUpdated events aren't being sorted into separate queues.

(The above assumes that this class never receives events that belong to other channels)

packages/server-wallet/src/chain-service/chain-service.ts
```diff
@@ -545,56 +495,42 @@ export class ChainService implements ChainServiceInterface {
...
 
-  private async onAssetHolderEvent(
-    assetHolderContract: Contract,
-    ethersEvent: Event
-  ): Promise<void> {
+  private async onAssetEvent(ethersEvent: Event): Promise<void> {
     switch (ethersEvent.event) {
       case Deposited:
         {
-          const depostedEvent = await this.getDepositedEvent(assetHolderContract, ethersEvent);
```
NiloCK, line 575: It took me a really long time to find the type in `deposted`.

NiloCK (approved):
Nice work!

packages/nitro-protocol/src/contract/transaction-creators/nitro-adjudicator.ts
```diff
@@ -55,3 +56,27 @@ export function createConcludeAndTransferAllAssetsTransaction(
     ),
   };
 }
+
+export function transferAllAssetsArgs(
```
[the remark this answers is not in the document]
geoknee: Fixed the import. I take your point about the  function name -- but I will keep it for now as it matches other code in this file. We can clear it up in a future PR.

packages/server-wallet/src/chain-service/event-tracker.ts
geoknee: This is a helpful summary. Through some manual testing I have done, it seems important that there is a separate event queue for each asset (otherwise tests fail). I renamed `assetMap` to `eventRecordFor`, and went with your suggestion for using a constant string for the fingerPrintUpdated queue (rather than the channelId).

PR #3649, opened 2021-07-07
Wallet tweaks for testing infrastructure

lalexgap:
These are a collection of changes that I've made to the wallet and message service while working on testing infrastructure.

# Changes
- Wallet now emits a ChannelUpdated event
- Wallet accepts a private key to use to create its address (useful for testing)
- call updateFunding in the AssetOutcomeUpdated event handler so the holdings are correctly updated when closing a channel.

Happy to break these into separate PRs if needed/wanted.

### DIFF
--- a/packages/server-wallet/src/wallet/wallet.ts
+++ b/packages/server-wallet/src/wallet/wallet.ts
@@ -130,9 +130,6 @@ export class Wallet extends EventEmitter<WalletEvents> {
 
     const handler: MessageHandler = async message => {
       const result = await this._engine.pushMessage(message.data);
-      const {channelResults} = result;
-
-      await this.registerChannels(channelResults);
 
       await this.handleEngineOutput(result);
     };
@@ -198,10 +195,14 @@ export class Wallet extends EventEmitter<WalletEvents> {
    * @returns A promise that resolves to a collection of ObjectiveResult.
    */
   public async approveObjectives(objectiveIds: string[]): Promise<ObjectiveResult[]> {
-    const {objectives, messages, chainRequests} = await this._engine.approveObjectives(
-      objectiveIds
-    );
+    const {
+      objectives,
+      messages,
+      chainRequests,
+      channelResults,
+    } = await this._engine.approveObjectives(objectiveIds);
 
+    await this.registerChannels(channelResults);
     const results = objectives.map(async o => ({
       objectiveId: o.objectiveId,
       currentStatus: o.status,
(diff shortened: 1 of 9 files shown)


packages/server-wallet/src/wallet/wallet.ts
```diff
@@ -142,7 +148,9 @@ export class Wallet extends EventEmitter<WalletEvents> {
           toAddress: makeDestination(ai.destination),
           amount: ai.amount as Uint256,
         }));
-
+        // holdingUpdated only fires for the deposit event
+        // so we also need to update the funding in this event
+        await this._engine.store.updateFunding(e.channelId, e.newHoldings, e.assetHolderAddress);
```
andrewgordstewart, line 153: There should be a test failing due to this addition. Since there isn't, that suggests there's missing test coverage?

Also, the way it's written, the wallet can enter an inconsistent state, if it crashes after L153's query is committed, but before L154's query is committed. How should you recover from that state? Should the two queries execute within a transaction?

(It seems to me as though updating transferred out should decrease the channel's funding.)

lalexgap: > There should be a test failing due to this addition. Since there isn't, that suggests there's missing test coverage?

It actually took me a sec to figure this out. I would of expected the [directly funded channel test](https://github.com/statechannels/statechannels/blob/ag/wallet-tweaks/packages/server-wallet/src/__chain-test__/directly-funded-channel.test.ts#L9:L9) to catch this.

The reason we didn't is that every time pushMessage is called we call[ registerChannel on the chain service](https://github.com/statechannels/statechannels/blob/fe2c4fec27a060aa2433d2bc3953c7c880e06d59/packages/server-wallet/src/wallet/wallet.ts#L129).

registerChannel  is used to register listeners for the chain service events. Once of the first things it does it query the initial holdings and dispatches a `holdingUpdated` event.

This means that every time a wallet received a message it would be creating another listener for chain events. Every time we created a listener the chain service would query the chain and dispatch a holdingUpdated event. 

Obviously this is not very performant! I've updated the wallet to only call registerChannel on createChannels and approveObjectives in 5bc5af7. This allows the directly funded channel test to fail as expected.

> Also, the way it's written, the wallet can enter an inconsistent state, if it crashes after L153's query is committed, but before L154's query is committed. How should you recover from that state? Should the two queries execute within a transaction?
> 
> (It seems to me as though updating transferred out should decrease the channel's funding.)

That makes sense. I've just updated `updateTransferredOut` to update the holdings as well (within the same tx).

andrewgordstewart (approved):

PR #3650, opened 2021-07-07
Chain service tweaks for testing infrastructure

lalexgap:
These are a collection of changes that I've made to the chain service while working on testing infrastructure.

# Changes
- deploy now be passed an rpc endpoint (nice for scripting)
- chain service now reads env vars when constructed (nice for scripting) 
- remove some obsolete methods from the chain service interface

### DIFF
--- a/packages/server-wallet/deployment/deploy.ts
+++ b/packages/server-wallet/deployment/deploy.ts
@@ -15,10 +15,10 @@ export type TestNetworkContext = {
 
 // NOTE: deploying contracts like this allows the onchain service package to
 // be easily extracted
-
-export async function deploy(): Promise<TestNetworkContext> {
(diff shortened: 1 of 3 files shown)


andrewgordstewart (approved):
I thought the wallet no longer read from env variables?

lalexgap:
> I thought the wallet no longer read from env variables?

I don't think that's true for contract deployment. We're still reading the env vars for the contract addresses.

andrewgordstewart:
> > I thought the wallet no longer read from env variables?
> 
> I don't think that's true for contract deployment. We're still reading the env vars for the contract addresses.

Should we read contract env vars from the config instead? It seems bad to mix configuration strategies.

PR #3651, opened 2021-07-09
Revert to old event system

### LINKED ISSUE #3646: gas optimize `FingerprintUpdated` events
Are `FingerprintUpdated` events new? These events are rather expensive to emit, costing 8 gas per byte.

If things are factored a certain way, where the resulting transformation is computed by a pure function, then a wallet can use the input to `claim` to compute the new outcome.

This is the general pattern used in current-nitro, except [here](https://github.com/statechannels/statechannels/blob/546d164eeeb08efc877916b3886431001f88e993/packages/nitro-protocol/contracts/ForceMove.sol#L92-L103) -- this log statement in particular costs ~20k gas, vs the ~500 gas for most other log statements currently triggered.

_Originally posted by @andrewgordstewart in https://github.com/statechannels/statechannels/pull/3644#discussion_r664736489_

(this issue only makes sense after we get rid of pushOutcome)

geoknee:
In the process of removing asset holders / pushOutcome, I started off replacing n x `AllocationUpdated` events with a single `FingerprintUpdated` event. With hindsight this wasn't a good idea as it ended up costing more gas and making the existing client code difficult or impossible to update. This PR moves us back to the existing event system. Instead of emitting the entire outcome, we emit just enough information such that a client can reconstruct that outcome themselves, using the event plus the calldata of the transaction that triggered it. The `AllocationUpdated` events now include an `assetIndex` parameter to make this simple. It's possible that we may be able to drop that in future optimizations.

See this comment thread: https://github.com/statechannels/statechannels/pull/3644#discussion_r664768233 

This may not be the optimal way of organising events, but it's an important step because it gets the chain service tests passing. I anticipate this will be the last significant PR into our trunk branch before that becomes ready for a final review. 

The upshot of these changes compared to the base branch: Saves about 5K gas per transfer or claim. Allows the current server wallet to continue to track money that has been transferred out of channels on chain. 

Fixes #3646


packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -189,33 +190,43 @@ contract MultiAssetHolder is IMultiAssetHolder, ForceMove {
...
+                );
+                require(
+                    assetOutcome.assetOutcomeType == Outcome.AssetOutcomeType.Allocation,
+                    '!allocation'
+                );
+                allocation = abi.decode(
+                    assetOutcome.allocationOrGuaranteeBytes,
+                    (Outcome.AllocationItem[])
+                );
+            }
```
NiloCK, line 206:
```suggestion
            Outcome.AssetOutcome memory assetOutcome = abi.decode(
                outcome[assetIndex].assetOutcomeBytes,
                (Outcome.AssetOutcome)
            );
            require(
                assetOutcome.assetOutcomeType == Outcome.AssetOutcomeType.Allocation,
                '!allocation'
            );
            allocation = abi.decode(
                assetOutcome.allocationOrGuaranteeBytes,
                (Outcome.AllocationItem[])
            );
```

As far as I can tell this extra indent + [block scope](revert-to-old-event-system) is unneeded and probably a mistake.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -189,33 +190,43 @@ contract MultiAssetHolder is IMultiAssetHolder, ForceMove {
...
-            emit FingerprintUpdated(guarantee.targetChannelId, outcomeBytes);
+            {
+                bytes32 outcomeHash = keccak256(abi.encode(outcome));
+                _updateFingerprint(
+                    guarantee.targetChannelId,
+                    targetStateHash,
+                    targetChallengerAddress,
+                    outcomeHash
+                );
+            }
```
NiloCK, line 228:
```suggestion
            bytes32 outcomeHash = keccak256(abi.encode(outcome));
            _updateFingerprint(
                guarantee.targetChannelId,
                targetStateHash,
                targetChallengerAddress,
                outcomeHash
            );
```

packages/server-wallet/src/chain-service/__chain-test__/chain-service.test.ts
```diff
@@ -415,18 +412,38 @@ describe('concludeAndWithdraw', () => {
...
+          amount: BN.from(1),
+          destination: makeDestination(aAddress).toLowerCase(),
+        },
+        {
+          amount: BN.from(3),
+          destination: makeDestination(bAddress).toLowerCase(),
+        },
+      ],
+      internalPayouts: [],
+      newAssetOutcome,
```
NiloCK, line 439: Might be an improvement to de-duplicate this chunk of code. I'll have a look at it myself now.

NiloCK (approved):
Looks good! Besides reading here, I did a quick scan through the remaining instances of the string `fingerprint` in the repository.

andrewgordstewart (approved):
good idea -- `remove-pushOutcome-trunk` has a lot of changes, so it's important to limit its scope as much as possible!

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: I took your word for it and committed this suggestion along with the other one. However, it caused a "stack too deep" compiler error. I don't think the block scoping here is terribly neat or optimal, but it does seem to avoid that problem. If you want to play around with it some more, maybe have a go on a new branch and check each change compiles successfully.

PR #3665, opened 2021-07-16
Clean up load-generator timing options

### LINKED ISSUE #3661: Switch time arguments for load generator

lalexgap:
Fixes #3661 

Implements some saner options for the load generator.  Update steps have been removed to keep things simpler.

The new options:
- duration:  Roughly how long test will run for
- createRate: how many channels should be created (per second)
- closeRate: how many channels should be closed (per second). If set to 0 no channels are closed.
- closeDelay: The miminum amount of time to wait after a channel has been created before scheduling a close step.

### DIFF
--- a/packages/server-wallet/e2e-testing/scripts/load-generator.ts
+++ b/packages/server-wallet/e2e-testing/scripts/load-generator.ts
@@ -69,13 +69,12 @@ async function createLoad() {
 
   const roles = (await jsonfile.readFile(roleFile)) as Record<string, RoleConfig>;
 
-  const jobIds = new Array<string>();
-  const steps: Step[] = [];
-
   console.log(chalk.whiteBright(`Generating a test load file to  ${outputFile}`));
   console.log(
     chalk.whiteBright(
       `Using the following options ${util.inspect({
+        outputFile,
+        roleFile,
         duration,
         createRate,
         closeRate,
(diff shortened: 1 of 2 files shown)


packages/server-wallet/e2e-testing/scripts/load-generator.ts
```diff
@@ -24,91 +24,116 @@ const ETH_ASSET_HOLDER_ADDRESS = '0x9eD274314f0fB37837346C425D3cF28d89ca9599';
...
+    })
+    .option('createRate', {
+      alias: 'cr',
+      default: 5,
+      describe: 'The number of channels that should be created per a second.',
+    })
+    .option('createWait', {
+      default: 5,
+      describe: `The minumum amount of time (in seconds) to wait for a channel be fully open, before another step is scheduled.`,
+    })
```
andrewgordstewart, line 64: What happens when this is `0`? Thinking ahead to ledger channels, we'd want `createWait` to be very close to 0 to simulate a heavily burdened virtual channel hub, right?

packages/server-wallet/e2e-testing/scripts/load-generator.ts
```diff
@@ -24,91 +24,116 @@ const ETH_ASSET_HOLDER_ADDRESS = '0x9eD274314f0fB37837346C425D3cF28d89ca9599';
...
+    .option('closeRate', {
+      default: 5,
+      describe:
+        'The amount of channels to be closed per a second. If this is larger than the createRate then all channels will eventually get closed.',
```
andrewgordstewart, line 68: This option seems nice!
```suggestion
        'The amount of channels to be closed per a second. If this is larger than the createRate then all channels will eventually get closed. Otherwise, some channels will remain open.',
```

packages/server-wallet/e2e-testing/scripts/load-generator.ts
lalexgap: If `createWait ` is `0` it's essentially ignored, and steps could have a timestamp that happens right after a create step.
```
        const timestamp = Math.max(
          generateRandomNumber(createStep.timestamp, toMilliseconds(duration)),
          createStep.timestamp + toMilliseconds(createWait) // 
        );

```

With ledger funding we expect channels to be created quite quickly so a `createWait` of 0 or near 0 makes sense!

lalexgap: FYI: `createWait` has been renamed to `closeDelay`

packages/server-wallet/e2e-testing/test-data/load-data.json
```diff
@@ -1,9 +1,9 @@
 [
  {
   "type": "CreateChannel",
-  "jobId": "quaint-ancient-calm-crowd",
+  "jobId": "cuddly-raspy-whining-pillow",
```
andrewgordstewart, line 4: This file should be ignored by git

andrewgordstewart (approved):

packages/server-wallet/e2e-testing/test-data/load-data.json
lalexgap: This is more to show a sample of load data and provide a default for the start all script. The load generator generates files in `temp` by default.

andrewgordstewart: I see. It’s a pretty large diff on a decently large file, but if it’s not going to change much, that seems fine

PR #3667, opened 2021-07-20
get CI tests passing

geoknee:
We have been allowing tests to fail on the base branch to aid quicker development and review. The time has come to put that right in preparation for a final review of #3624.

This is a mix of changes that would ordinarily have been "caught" during the previous development work if we hadn't allowed tests to fail. There aren't too many of them, thankfully. 

Two changes require a bit more explanation: 
1. upgrading wasm-utils. I published a custom version of that package to npm (see the commit message for more detail). This step is essentially mirroring the change to the `Outcome` in nitro protocol that are consumed by the server-wallet, but via a package compiled from rust source code in a different repo.
2. regnerating the api report for the server wallet. Annoyingly, there are a large number of formatting changes. I do not know why, and I can't undo them without CI failing. Viewing with "hide whitespace changes" helps (a bit).


packages/server-wallet/server-wallet.api.md
```diff
@@ -3,101 +3,101 @@
 > Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
```
geoknee, line 3: The changes here are only formatting changes. But we still need them. It's an auto-generated file.

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
```diff
@@ -220,7 +220,7 @@ describe('claim', () => {
             args: {
               channelId: targetId,
               assetIndex: BigNumber.from(0),
-              initialHoldings: heldBefore[addresses.g],
+              initialHoldings: heldBefore[guarantorId],
```
NiloCK, line 223: From my reading of the file, I can't figure out why (if?) this was necessary.

packages/nitro-protocol/config/jest/jest.contracts.config.js
```diff
@@ -3,4 +3,5 @@ config.testMatch = ['<rootDir>/test/contracts/**/*.test.ts'];
 config.reporters = ['default'];
 config.globalSetup = '<rootDir>/jest/contract-test-setup.ts';
 config.globalTeardown = '<rootDir>/jest/contract-test-teardown.ts';
+config.testTimeout = 90_000;
```
NiloCK, line 6: I can't think of a reason that these changes should require, on average, substantially more time in the test suites. Any comment?

NiloCK (approved):
This looks good to me! I had some trouble with `yarn prepare` in `nitro-protocol` but it was PEBCAC. I'm running tests locally now and will return if problems present.

packages/nitro-protocol/config/jest/jest.contracts.config.js
geoknee: I also cannot think of a compelling reason 😞 . I think we may have been quite close to the existing timeout already, and we have disturbed the number of tests that run (concurrently), which could lead to hard-to-predict changes in test runtime. 

I didn't yet try and reduce this timeout again, there is probably quite a bit of slack in it now.

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
geoknee: I think it is unnecessary in fact. Probably a debugging artifact.

PR #3674, opened 2021-07-23
remove timeout configuration

NiloCK:
This test consistently fails due to timeout on at least one dev machine (mine!) with tests run in parallel (`yarn test`), but succeeds as is when run independently (eg: `yarn test _validSignatures`).

With this timeout config removed, the test is still constrained by the global timeout config of 90 seconds.

Incremental adjustments to the timeout (20 seconds, 30 seconds, 40 seconds) resulted in fewer and less predictable errors but failing tests nonetheless.

### Shortcoming

Relaxing the timeout constraint makes the feedback from a successful test run less informative.

### Alternative approach

Alternatively, the test suite could be adjusted to run tests sequentially by default. This slows the test suite runs substantially.

---
### Code quality
### Project management

### DIFF
--- a/packages/nitro-protocol/test/contracts/ForceMove/_validSignatures.test.ts
+++ b/packages/nitro-protocol/test/contracts/ForceMove/_validSignatures.test.ts
@@ -7,8 +7,6 @@ import {sign} from '../../../src/signatures';
 const provider = getTestProvider();
 let ForceMove: Contract;
 
-jest.setTimeout(10_000);
-
 beforeAll(async () => {
   ForceMove = setupContract(provider, ForceMoveArtifact, process.env.TEST_FORCE_MOVE_ADDRESS);
 });


geoknee:
> ### Shortcoming
> Relaxing the timeout constraint makes the feedback from a successful test run less informative.

What feedback do you mean? 

I would have thought the downside would be that we are not failing as quickly as we could be, in the event that ganache has ground to a halt (or something). 

The more I think about it, the more pretty long timeouts make sense to me for these chain tests. We have a local blockchain that mines blocks "instantly", but is being bombarded with dozens (hundreds?) of transactions concurrently.

geoknee (approved):

NiloCK:
I guess I mean: larger timeout allowances give less information overall. "Completes sometime" is a different result than "completes quickly", and is especially different in context of performance-sensitive operations.

Difficulty here is that the performance measure is already fuzzy because many operations are simultaneously using resources. A slow test run doesn't make the bottlenecks obvious.

geoknee:
Right, I see. Although I think we still get to learn how quickly each test suite takes?
Screenshot 2021-07-23 at 16 33 52

PR #3680, opened 2021-07-26
Code suggestions

andrewgordstewart:
When reviewing the security aspects of this, it is important to understand what changes are made to storage variables. These were some code modifications I made along the way while understanding storage related changes. Some of them come with minor gas savings.

In addition, I re-ordered `_claim` & `_transfer` according to the order in which they are called in `MultiAssetHolder`.


geoknee (approved):
These are some nice changes! I particularly like the use of a new storage mapping to slice into a single asset's holdings -- that seemed counterintuitive to me at first since "making a new storage variable" seems like it should be expensive.

PR #3687, opened 2021-07-27
More performant random channel/destination generation

andrewgordstewart:
The affected test helpers do not require knowing the private key corresponding to the signing address that generates:
- the destination; or
- the channel id

When this is the case, it's much faster to generate a random hex string. Creating an `ethers` wallet is already slow; doing so in a jest environment is _very slow_ (I believe it takes on the order of 100ms.)


geoknee:
Because the base branch was merged to master with a "squash", the history and diff here are messed up. I cleaned it up on  #3691

PR #3691, opened 2021-07-27
More performant random channel/destination generation

geoknee:
(Rewrite of the commit history of #3687 , originally authored by @andrewgordstewart )

The affected test helpers do not require knowing the private key corresponding to the signing address that generates:

the destination; or
the channel id
When this is the case, it's much faster to generate a random hex string. Creating an ethers wallet is already slow; doing so in a jest environment is very slow (I believe it takes on the order of 100ms.)


geoknee:
@andrewgordstewart I can't approve this. But I approve this.

andrewgordstewart (approved):

PR #3693, opened 2021-07-27
Sync Ledger Channels for OpenChannel Objective

### LINKED ISSUE #3583: Flickering server-wallet-with-peers test
[This run](https://app.circleci.com/pipelines/github/statechannels/statechannels/13103/workflows/522869dc-bce8-408a-8103-56a1f24b0b39/jobs/64964) was performed on `master`. It included a documentation change only, which should not affect the `with-peers` test.

lalexgap:
Fixes #3583  (maybe?)

When syncing an OpenChannel Objective we now sync the funding ledger channel as well.

Previously syncObjective method would only send the **app** channel to peers and not **ledger** channels. This meant that if the initial message containing the ledger channel was dropped no amount of syncing would complete the objective.

I believe this explains the flickering tests. We have a test scenario using `{dropRate:0.1}` which means 10% of messages should be dropped. If the first message happens to be dropped then the test was doomed to fail, since syncing wouldn't include the ledger info.

### DIFF
--- a/packages/server-wallet/src/engine/engine.ts
+++ b/packages/server-wallet/src/engine/engine.ts
@@ -200,13 +200,9 @@ export class SingleThreadedEngine implements EngineInterface {
       const {participants} = channel;
       response.queueSendObjective(o, channel.myIndex, participants);
 
-      // Sync any ledger channels used for an open channel objective
-      if (
-        o.type === 'OpenChannel' &&
-        o.data.fundingStrategy === 'Ledger' &&
-        !!o.data.fundingLedgerChannelId
-      ) {
-        await this._syncChannel(o.data.fundingLedgerChannelId, response);
+      // Sync any ledger channels used for the objective
+      if (channel.fundingStrategy === 'Ledger' && channel.fundingLedgerChannelId) {
+        await this._syncChannel(channel.fundingLedgerChannelId, response);
       }
     }
 
--- a/packages/server-wallet/src/engine/engine.ts
+++ b/packages/server-wallet/src/engine/engine.ts
@@ -199,6 +199,15 @@ export class SingleThreadedEngine implements EngineInterface {
 
       const {participants} = channel;
       response.queueSendObjective(o, channel.myIndex, participants);
+
+      // Sync any ledger channels used for an open channel objective
+      if (
+        o.type === 'OpenChannel' &&
+        o.data.fundingStrategy === 'Ledger' &&
+        !!o.data.fundingLedgerChannelId
+      ) {
+        await this._syncChannel(o.data.fundingLedgerChannelId, response);
+      }
     }
 
     return response.allMessages;
--- a/packages/server-wallet/src/engine/__test__/sync-objectives.test.ts
+++ b/packages/server-wallet/src/engine/__test__/sync-objectives.test.ts
@@ -68,6 +68,8 @@ describe('SyncObjective', () => {
   });
 
   it('includes the ledger channel when syncing a ledger funded open channel objective', async () => {
+    // We up the nonce to avoid collisions with the other tests not using TestChannel
+    TestChannel.maximumNonce = 100;
     const ledger = TestLedgerChannel.create({});
     const app = TestChannel.create({
       fundingStrategy: 'Ledger',
@@ -98,13 +100,13 @@ describe('SyncObjective', () => {
       data: {
         // We expect states for the app channel as well as the ledger channel that funds it
         signedStates: [
-          expect.objectContaining({channelId: app.channelId}),
           expect.objectContaining({channelId: ledger.channelId}),
+          expect.objectContaining({channelId: app.channelId}),
         ],
         // We expect a getChannel request (from syncChannel) for both channels
         requests: [
-          expect.objectContaining({channelId: app.channelId, type: 'GetChannel'}),
           expect.objectContaining({channelId: ledger.channelId, type: 'GetChannel'}),
+          expect.objectContaining({channelId: app.channelId, type: 'GetChannel'}),
         ],
       },
     };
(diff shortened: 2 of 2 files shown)


packages/server-wallet/src/engine/engine.ts
```diff
@@ -199,6 +199,15 @@ export class SingleThreadedEngine implements EngineInterface {
...
       response.queueSendObjective(o, channel.myIndex, participants);
+
+      // Sync any ledger channels used for an open channel objective
+      if (
+        o.type === 'OpenChannel' &&
+        o.data.fundingStrategy === 'Ledger' &&
+        !!o.data.fundingLedgerChannelId
+      ) {
+        await this._syncChannel(o.data.fundingLedgerChannelId, response);
+      }
```
andrewgordstewart, line 210: This sort of feels like the wrong way to go about it:
- if one channel is out of sync, is there a possibility that many channels are out of sync?
- if there are many channels out of sync, then we would be syncing the ledger channel many times, right?

In my mind, it seems like you'd be better off syncing from the ledger channel "downwards", rather than from the funded channel "upwards".

All in all, I find it hard to keep everything together in my head about what could go wrong, how to detect things going wrong, and how to fix things when things go wrong. This seems like a good fix for the short term, and we should schedule ourselves to come back to the problem during cooldown?

lalexgap: > This sort of feels like the wrong way to go about it:
> 
> * if one channel is out of sync, is there a possibility that many channels are out of sync?
> * if there are many channels out of sync, then we would be syncing the ledger channel many times, right?


> In my mind, it seems like you'd be better off syncing from the ledger channel "downwards", rather than from the funded channel "upwards".

Is there a reason to do this besides performance? Right now our model is based on syncing objectives. After an objective does not make progress for a certain amount of time we resend a message containing the objective and channel states. So it feels natural to include the ledger information in the "sync objective" payload, as it's needed to make progress on the objective.

To sync from a ledger channel "downwards" I think we'd need to change our approach to poll ledger channels and check when they were last updated.

If performance is an issue it's pretty trivial to fetch all the ledger channels at once for a batch of objectives we want to sync.
  
> All in all, I find it hard to keep everything together in my head about what could go wrong, how to detect things going wrong, and how to fix things when things go wrong. This seems like a good fix for the short term, and we should schedule ourselves to come back to the problem during cooldown?

Could you elaborate on what you mean by "go wrong"?  It seems like in the worst case we may send messages including ledger information already known by another participant, but that doesn't seem too bad?

andrewgordstewart: > > This sort of feels like the wrong way to go about it:
> > 
> > * if one channel is out of sync, is there a possibility that many channels are out of sync?
> > * if there are many channels out of sync, then we would be syncing the ledger channel many times, right?
> 
> > In my mind, it seems like you'd be better off syncing from the ledger channel "downwards", rather than from the funded channel "upwards".
> 
> Is there a reason to do this besides performance? Right now our model is based on syncing objectives. After an objective does not make progress for a certain amount of time we resend a message containing the objective and channel states. So it feels natural to include the ledger information in the "sync objective" payload, as it's needed to make progress on the objective.

At the same time, the ledger information is relevant to _many_ objectives, and the ledger protocol is quite complicated, so I (for one) can't really understand what would happen if we synced many objectives at once, which were all open channel objectives funded by the same ledger channel. This change is introducing some potentially overwhelming behaviour into the wallet, and that is what I'm expressing concern about.

> Does syncing from a ledger channel "downwards" mean polling the ledger database and making sure those ledger channels make progress?

There is a ledger funding protocol that I do not believe anyone on the team really understands. Therefore, I don't think that anybody can really say for sure. As a general idea, though, since the ledger protocol is not objective-driven, and is designed to handle concurrent ledger requests, it feel like it should not be synced when syncing an objective for a child channel.

> > All in all, I find it hard to keep everything together in my head about what could go wrong, how to detect things going wrong, and how to fix things when things go wrong. This seems like a good fix for the short term, and we should schedule ourselves to come back to the problem during cooldown?
> 
> Could you elaborate on what you mean by "go wrong"?

"go wrong" is pretty vague, but there are a few behaviours that we _didn't think would happen_ that we then _observed did happen_, either in some production environment or as a result of running the new test infrastructure (or even the existing e2e tests, as shown by the flickering test that we ignored!)

In my opinion, we'd be in better shape if we started with a good argument of why things _shouldn't_ "go wrong"; as this PR points out, we don't have such an argument (things "went wrong" in a very simple situation) and I am advocating that we think about it and come up with one.

> It seems like in the worst case we may send messages including ledger information already known by another participant, but that doesn't seem too bad?

- it depends on how many messages we send. If we're sending the same message 100 times, and it takes some real amount of time or resources to process that message (as I can imagine being true for a well-used ledger channel) that could be a problem. Is it possible to "go critical" and trigger more failures?
- is syncing a ledger channel enough to ensure that ledger funding proceeds? why? what about ledger requests that never made it into a state update? does syncing a ledger channel also sync the ledger requests?
- what about closing ledger funded channels?
- what about opening/closing virtually funded channels?

What I'm trying to get at is, I don't know a good argument why the proposed solution in this PR is a good solution long term. At a minimum, it does not cover things that can go wrong today (messages dropped while closing a ledger-funded channel), and at worst, it could lead to unpredictable overloading of a wallet due to redundant messages.

lalexgap: @andrewgordstewart I've created an issue for this, would you be ok merging this as-is for now? (And if so could you approve it?)

andrewgordstewart (approved):

packages/server-wallet/src/engine/engine.ts
andrewgordstewart: done!

PR #3696, opened 2021-07-28
🏃‍♂️ Integrate the exit format into nitro protocol  

geoknee:
This is a trunk branch that we can merge smaller PRs into.

Work in progress.

~Note: `exit-format` has not been published recently, but development can continue against the latest code by using `yarn link`.~

The aim is to have all tests passing in `nitro-protocol`. 

~But otherwise the monorepo will remain "broken" (for now). We obviously won't merge to master until that is fixed -- but the strategy for that is TBD.~

UPDATE: I have "pinched off" nitro-protocol from the rest of the monorepo so that the whole monorepo can compile and test successfully. That'll make it easier to review and merge this work and return to the full integration later. 

**only** in nitro protocol:
- [x] exit format imported to solidity and compiling
- [x] exit format imported to typescript and compiling
- [x] transfer implemented
- [x] transfer test(s) passing
- [x] claim implemented
- [x] claim test(s) passing
- [x] the rest of the tests passing
- [x] latest version of exit format imported from npm


geoknee:
Superseded by #3758

PR #3697, opened 2021-07-28
remove challengerAddress from ChannelData

geoknee:
`challengerAddress` is a variable that is computed on chain when raising a challenge: and is used to ensure only participants can raise challenges. It is also stored (inside a hash with some other data), and is a required to be resubmitted in the calldata of `transfer`, `claim`, `respond`, `checkpoint`... etc. However, the `challengerAddress` stored on-chain is never referenced.

It therefore costs a small amount of gas, bloats our API significantly, and heightens the chance of a "stack too deep" error (and the gymnastics that is often required to get around it).

This PR removes the `challengerAddress` from the storage and the calldata of the mentioned methods. It is still computed and checked on chain. 

Shortcomings / Open questions: 
- **what was the original motivation to store this variable?** My guess for this is that it would allow a mechanism for preventing multiple consecutive challenges from the same participant and/or a mechanism to **slash** the participant if they challenged with a state later proven to be "stale". We have no such mechanism in place at present.
- **is it important to restrict challenges to participants?** If we wanted that slashing mechanism, we might need to store the challenger address -- although the funds in the channel aren't always assignable to one participant or another (destinations are not always 1-1 with participants) so we would probably need to introduce separate "slashing deposits" for each participant, then. So we might argue that the check on the challenger being a participant could be removed (it is **not removed**, yet, on this PR). 

The summary may omit some of these details if this PR fixes a single ticket that includes these details. It is the reviewer's discretion. 

### Changes
Ih particular, don't emit `challenger` in the `ChallengeRegistered` event. This saves 300 gas or so from each challenge. 

## :warning: Does this require multiple approvals? 
Yes

---
### Code quality
### Project management

### DIFF
--- a/packages/nitro-protocol/contracts/NitroAdjudicator.sol
+++ b/packages/nitro-protocol/contracts/NitroAdjudicator.sol
@@ -58,11 +58,7 @@ contract NitroAdjudicator is MultiAssetHolder {
     ) public {
         // checks
         _requireChannelFinalized(channelId);
-        _requireMatchingFingerprint(
-            stateHash,
-            keccak256(outcomeBytes),
-            channelId
-        );
+        _requireMatchingFingerprint(stateHash, keccak256(outcomeBytes), channelId);
         Outcome.OutcomeItem[] memory outcome = abi.decode(outcomeBytes, (Outcome.OutcomeItem[]));
 
         for (uint256 assetIndex = 0; assetIndex < outcome.length; assetIndex++) {
--- a/packages/nitro-protocol/contracts/NitroAdjudicator.sol
+++ b/packages/nitro-protocol/contracts/NitroAdjudicator.sol
@@ -41,7 +41,7 @@ contract NitroAdjudicator is MultiAssetHolder {
             sigs
         );
 
-        transferAllAssets(channelId, outcomeBytes, bytes32(0), address(0));
+        transferAllAssets(channelId, outcomeBytes, bytes32(0));
     }
 
     /**
@@ -50,19 +50,16 @@ contract NitroAdjudicator is MultiAssetHolder {
      * @param channelId Unique identifier for a state channel
      * @param outcomeBytes abi.encode of an array of Outcome.OutcomeItem structs.
      * @param stateHash stored state hash for the channel
-     * @param challengerAddress stored challenger address for the channel
      */
     function transferAllAssets(
         bytes32 channelId,
         bytes memory outcomeBytes,
-        bytes32 stateHash,
-        address challengerAddress
+        bytes32 stateHash
     ) public {
         // checks
         _requireChannelFinalized(channelId);
         _requireMatchingFingerprint(
             stateHash,
-            challengerAddress,
             keccak256(outcomeBytes),
             channelId
         );
@@ -97,7 +94,7 @@ contract NitroAdjudicator is MultiAssetHolder {
         }
         outcomeBytes = abi.encode(outcome);
         bytes32 outcomeHash = keccak256(outcomeBytes);
-        _updateFingerprint(channelId, stateHash, challengerAddress, outcomeHash);
+        _updateFingerprint(channelId, stateHash, outcomeHash);
     }
 
     /**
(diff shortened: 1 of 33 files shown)


packages/nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -140,10 +139,8 @@ interface IForceMove {
      */
     event ChallengeRegistered(
         bytes32 indexed channelId,
-        // everything needed to respond or checkpoint
         uint48 turnNumRecord,
         uint48 finalizesAt,
-        address challenger,
```
andrewgordstewart, line 146: If the goal is to save gas, the vast majority of the data on this event can be removed. I think only the `channelId` needs to be emitted; we can then lean on the "get data from `calldata`" technique to fill out the rest.

andrewgordstewart: https://github.com/statechannels/statechannels/issues/3699

andrewgordstewart (approved):

andrewgordstewart:
>     * **what was the original motivation to store this variable?** My guess for this is that it would allow a mechanism for preventing multiple consecutive challenges from the same participant and/or a mechanism to **slash** the participant if they challenged with a state later proven to be "stale". We have no such mechanism in place at present.

The first instance that I can find of it is [here](https://github.com/statechannels/statechannels/blob/5bdd5b4587635878283d557e39c13e0f71bcd1ef/new-packages/nitro-protocol/notes/force-move-spec.md), which unfortunately doesn't make any reference to the purpose. The challenger address did not seem to be stored on-chain prior to the gas optimization project.

No matter, I think it's best to remove the challenger address until there is a purpose for storing it.

>     * **is it important to restrict challenges to participants?** If we wanted that slashing mechanism, we might need to store the challenger address -- although the funds in the channel aren't always assignable to one participant or another (destinations are not always 1-1 with participants) so we would probably need to introduce separate "slashing deposits" for each participant, then. So we might argue that the check on the challenger being a participant could be removed (it is **not removed**, yet, on this PR).

Requiring that the challenger be a participant seems to add some "practical" security, since the challenger must know a participant's private key. (It doesn't seem to add any theoretical security, since Alice can't control what Bob does with his private key. But, it increases the difficulty of a non-participant launching a griefing challenge.)

packages/nitro-protocol/contracts/NitroAdjudicator.sol
```diff
@@ -41,7 +41,7 @@ contract NitroAdjudicator is MultiAssetHolder {
             sigs
         );
 
-        transferAllAssets(channelId, outcomeBytes, bytes32(0), address(0));
+        transferAllAssets(channelId, outcomeBytes, bytes32(0));
```
NiloCK, line 44: Unrelared to current PR, but stuck out to me as something I didn't know:

`stateHash = 0` is a magic value for concluded channels?

NiloCK (approved):
I tacked a few commits onto this branch which I think should be uncontroversial.

I came across [this commit by Mike](https://github.com/statechannels/statechannels/commit/4eb1114b9a88320153107a5acf009e929af951e1) which introduces challengerAddress to some pushOutcome workflow under the title _pushOutcome transaction creator should not assume challenger address_, but I did very little digging to understand broader context of the commit.

packages/nitro-protocol/contracts/NitroAdjudicator.sol
geoknee: Yes. This is buried in the docs https://docs.statechannels.org/docs/protocol-tutorial/release-assets but should be more prominent.

geoknee:
> I came across [this commit by Mike](https://github.com/statechannels/statechannels/commit/4eb1114b9a88320153107a5acf009e929af951e1) which introduces challengerAddress to some pushOutcome workflow under the title _pushOutcome transaction creator should not assume challenger address_, but I did very little digging to understand broader context of the commit.

I think that was just correcting the mistaken and implicit assumption that the participant *whose turn it is* is the only one who can challenge.

packages/nitro-protocol/contracts/NitroAdjudicator.sol
andrewgordstewart: I don't think there's any real reason to require `stateHash == 0`, it's just convenient. (Note that setting `stateHash = 0` during the execution of exits is [one idea](https://github.com/statechannels/statechannels/pull/3624#discussion_r677590732) for how to clearly separate adjudication and asset management.)

PR #3702, opened 2021-07-29
`import {ExitFormat as Outcome}` into solidity files

### LINKED ISSUE #3683: Refactor claim and transfer (on the remove-pushOutcome-trunk branch)
We would like to:

- fix the API for these functions
- specify their behaviour
- refactor for readability (why is there a `claim` and a `_claim` function?)
- stick to ["checks, effects, interactions" strictly](https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html)

This could be done as a part of the "integrate the exit format" project.

geoknee:
Should close #3683 .
⚠️ base not master

TODO: 

- [x] rewrite `transfer`
- [ ] ~rewrite `claim` (inserted [existing version](https://github.com/statechannels/exit-format/blob/a0a54ed99b04dba25dca33c9ec1b0459a2e185db/contracts/Nitro.sol#L27) from exit-format repo)~
- [x] reinstate `EmbeddedApplication.sol` Note this has required turning the solidity optimizer off (selectively, just for this file) 🐷 

Acceptance criteria for this PR: 
- [x] all contracts compile and deploy successfully when linked against `master` in the exit-format repo

Future PRs will deal with things like tests.


packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -93,359 +93,153 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+    )
+        internal
+        view
+        returns (
+            Outcome.SingleAssetExit[] memory outcome,
+            address asset,
+            uint256 initialAssetHoldings
+        )
+    {
+        _requireIncreasingIndices(indices); // This assumption relied on by compute_transfer_effects_and_interactions
```
NiloCK, line 143:
```suggestion
        _requireIncreasingIndices(indices); // This assumption is relied on by compute_transfer_effects_and_interactions
```

NiloCK (approved):
This looks good to me, but I have 2 cents on the new `verb_compound_noun` functions. Solidity convention (according to leading zombie authorities) is for `_underscore` prefacing on internal and private functions. Probably these utility functions should follow this convention.

I'm also thinking (again, still) about tightening up our accessors in general, in service of reducing API size on all contracts and hardening the separation of concerns between different contracts. It's definite scope creep to introduce this here, so I'll make a separate issue of it.

NiloCK:
This branch passes the compile-and-deploy threshold according to the time honoured standard of works-on-my-machine.

Compilation on Circle is expected to fail here because it is pulling the published version of exit-format, as opposed to the in-progress work linked locally (see #3696).

packages/nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
```diff
@@ -2,50 +2,12 @@
...
-     */
-    function claim(
-        uint256 assetIndex,
-        bytes32 guarantorChannelId,
-        bytes memory guarantorOutcomeBytes,
-        bytes32 guarantorStateHash,
-        bytes memory targetOutcomeBytes,
-        bytes32 targetStateHash,
-        uint256[] memory indices
-    ) external;
```
geoknee, line 47: @NiloCK I think we want something like the above API for `claim`. 

Firstly, about the `pure` declaration. We do want a pure function at the "heart" of claim -- but fundamentally the top level claim method needs to read from and mutate the state of the chain. So what you have now should become `_compute_claim_effects_and_interactions`, and we should then wrap it with inside a `claim` that has the above API. We will also then need a `_apply_claim_checks` and a `_apply_claim_effects` as well as `_apply_claim_interactions` so that we mirror the factorization of `transfer` that we have on this branch.

Secondly, about the inclusion of `assetIndex` in the API. I think we want it! We do not want to loop over all assets  (see https://github.com/statechannels/exit-format/issues/48). Although this is what we currently do in the exit format repo, looping over all assets doesn't really make sense with only a single `targetChannelIndex` -- it means all allocations in the source (aka guaranteeing) channel need to be present in every asset slice of its outcome. What I would like to do in the short term is match the current API of nitro protocol -- that is to say, slice into a single asset with `assetIndex`, and then only provide a 1d array of `indices` with which to select allocations from the target to pay out. I think you should find something pretty much exactly what I have in mind on this PR https://github.com/statechannels/exit-format/pull/52 under the guise of `_computeNewAllocationWithGuarantee`. I think that should be pasted here and become the `_compute_claim_effects_and_interactions`.

geoknee:
I think getting `claim` right is sufficiently involved that we should work on that on a different branch. To keep things neat, I'll revert the work towards `claim` on this branch for now.

packages/nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
kerzhner: I have been thinking about the second point for a bit. When `claim` is invoked, the guarantor and the channels it targets are finalized. At that point, it is in the interest of the guarantor participants to move all funds from the guarantor channel to the target channel(s). So `claim` would get invoked with every asset that contains funds for the guarantor channel.

If the above point make sense, then it seems like we are including the `assetIndex` as a consequence of how asset-to-channel data is stored. I wonder if there is a data structure we could use to enable the functionality to claim all assets without the restriction that every asset must contain every guarantor channel.

kerzhner: One alternative that comes to mind is specifying the `targetChannelId` in the `claim` API. With this API, we need to traverse every allocation for every asset. We would then check whether the the allocation destination matches the `targetChannelId`. This is probably too inefficient.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -94,356 +94,172 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+            address asset,
+            uint256 initialAssetHoldings
+        ) = _apply_transfer_checks(assetIndex, indices, fromChannelId, stateHash, outcomeBytes); // view
 
-        // Update the fingerprint
-        outcome[assetIndex].assetOutcomeBytes = abi.encode(
-            Outcome.AssetOutcome(Outcome.AssetOutcomeType.Allocation, abi.encode(allocation))
+        (
+            Outcome.Allocation[] memory newAllocations,
+            ,
```
kerzhner, line 105: Any particular reason to keep this parameter?

packages/nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
geoknee: I think the overall data structures and claim API are all fair game for improvement. In the short term however, the goal is to bring in the current exit format. This is so we can benefit from it allowing heterogenous outcomes (and also several guarantees in one outcome) and move away from having dedicated guarantor channels. 

I think a good strategy for now is to keep the API for `claim` as close to the current API as possible. Seeing as the current API targets a single guarantee and a single asset, we will need to add `targetChannelIndex` or `targetChannelId` to the API -- given that a ledger channel can contain many guarantees. As you say, `targetChannelId` would force us to do more "searching" through the outcomes -- and give us even more nested loops. So think sticking with an index makes sense for now. 


My attempt at `claim` is over at #3714 .

packages/nitro-protocol/contracts/MultiAssetHolder.sol
kerzhner: Gotcha! Is there a case where a single asset is being transferred, but it is the last asset to be transferred? Then would the entire outcome allocate zero amounts?

geoknee: In principle yes. But since a) we don't remove any parts of the outcome during a partial exit and b) we don't tend to inspect other parts of the outcome besides the assets targeted by the blockchain transaction, it's still a bit awkward to detect that it's safe to clear out the storage slot. 

We might want to factor this logic out in to a `deleteStatusIfOutcomeIsEmpty` helper. But I expect that may be somewhat expensive gas wise.

andrewgordstewart: Caveat: I have not read the code, only this conversation. (On my phone, making code review difficult.)

 Why don’t we remove any parts of the outcome during a partial exit? 

For transfer, it is necessary to remove parts of the outcome to prevent the first payee from being able to drain the funds.

For claim, it still seems like the right thing to do, rather than depend on a correct implementation of the virtual funding protocol to ensure that subsequent claims have no effect, right?

geoknee: We don't remove parts of the outcome in `transfer` either. We simply set the relevant `amount` to zero. 

Actually changing the shape of the outcome is tricky because of the way arrays work in solidity. Furthermore, things like `assetIndex` can have a dynamic meaning if the shape of the outcome can change. 

What is relatively straightforward is to check when all the amounts are zero, and then delete the entire outcome if safe to do so.

andrewgordstewart: Got it. Setting the amount to 0 makes sense!

PR #3706, opened 2021-07-30
Stress test sanity checker 

### LINKED ISSUE #3627: Add script/utility to assert things on the database
After the load test has run we want to see make sure everything worked as expected.

To help with this we can create a script that checks the database for the status of various channels. This can be used to make some rough assertions that the wallet behaviour is correct.

lalexgap:
Fixes #3627 

Introduces a "sanity check" script that performs some simple checks on the database and the chain to make sure the correct number of channels were created/funded/closed/etc...

The sanity checker accepts a load file to determine the number of channels that should have been created and closed. It then queries the database and asserts that the expected number of channels are there.

The ganache script has been updated to write out the balances of the adjudicator and the accounts to a file.  The sanity checker checks this file to ensure that the correct amount is in the participant's accounts / deposited against the adjudicator. 

The sanity checker is now included in the `start-all` script and runs after the test is complete.

image


packages/server-wallet/e2e-testing/test-data/roles.json
```diff
@@ -17,6 +18,7 @@
     "chainId": 9001,
     "artifactFile": "temp/contract_artifacts.json",
     "privateKey": "0xb3ab7b031311fe1764b657a6ae7133f19bac97acd1d7edca9409daa35892e727",
-    "chainPrivateKey": "0x2030b463177db2da82908ef90fa55ddfcef56e8183caf60db464bc398e736e6f"
+    "chainPrivateKey": "0x2030b463177db2da82908ef90fa55ddfcef56e8183caf60db464bc398e736e6f",
+    "destination":"0xBBDDFb5DADcf0E710D898DB8da7861aA8F635B9b"
```
lalexgap, line 22: Using a unique destination (which starts with 0 ETH) makes performing the sanity checks much easier.

.circleci/config.yml
```diff
@@ -171,6 +171,20 @@ jobs:
...
+    steps:
+      - checkout
+      - attach_workspace:
+          at: /home/circleci/project
+      - install_postgresql_client
+      - run:
+          name: server-wallet-simple-stress-test
+          working_directory: packages/server-wallet
+          command: npx ts-node ./e2e-testing/scripts/start-all.ts -l ./e2e-testing/test-data/simple-ci-run.json
+          no_output_timeout: 30m
```
andrewgordstewart, line 185: This is a really large value. What's the purpose of a 30m timeout here?

.circleci/config.yml
```diff
@@ -229,6 +243,9 @@ workflows:
       - server-wallet-with-peers-test:
           requires:
             - prepare
+      - server-wallet-simple-stress-test:
```
andrewgordstewart, line 246: It's nice that this step is snappy, taking <2m to finish!

packages/server-wallet/e2e-testing/scripts/sanity-checker.ts
```diff
@@ -0,0 +1,165 @@
...
+import {defaultTestWalletConfig, overwriteConfigWithDatabaseConnection} from '../../src';
+import {getKnexFromConfig} from '../../src/db-admin/db-admin';
+import {ObjectiveModel} from '../../src/models/objective';
+import {Channel} from '../../src/models/channel';
+import {setupUnhandledErrorListeners} from '../utils';
+
+setupUnhandledErrorListeners();
+assertSanity();
+
+async function assertSanity() {
```
andrewgordstewart, line 21: enhancement: I think checking the overall state of the user's balance is a good thing to check here, but can be added to the sanity checker on an as-needed basis.

For instance, if we end the simulation believing there are 34 active ledger-funded channels and 21 active virtually-funded channels, then we should somehow check that's reflected in the DB. This would be pretty close to taking a "snapshot" of the system mid-use, and making sure that everything is as we expect.

(I think there are a lot of ways of solving this general problem of "how do we know it's working correctlY?" and don't mean to express a strong opinion on _how_ to answer this question, just on its importance!)

packages/server-wallet/e2e-testing/scripts/start-ganache.ts
```diff
@@ -1,21 +1,31 @@
 import yargs from 'yargs/yargs';
 import {hideBin} from 'yargs/helpers';
 import {TEST_ACCOUNTS} from '@statechannels/devtools';
-import {writeFile} from 'jsonfile';
+import jsonfile, {writeFile} from 'jsonfile';
```
andrewgordstewart, line 4: nit: Can we make the imports consistent?

```suggestion
import jsonfile from 'jsonfile';
```

packages/server-wallet/e2e-testing/scripts/start-ganache.ts
```diff
@@ -83,14 +97,67 @@ async function setupGanache() {
...
+  process.removeAllListeners('SIGTERM');
+  process.removeAllListeners('beforeExit');
+
   console.log(chalk.green(`Contract artifacts written to ${commandArguments.artifactFile}`));
+
+  const roles = await getRoles(commandArguments.roleFile);
+  const destinations = Object.keys(roles).map(rId => roles[rId].destination);
+
+  // exitHook will run when the process gets terminated for whatever reason
+  exitHook(done =>
```
andrewgordstewart, line 122: Why did we need to add a new package for this? What's wrong with `process.on('exit', ...)`?

packages/server-wallet/e2e-testing/test-data/sample-direct-load-data.json
```diff
@@ -1,39 +1,3688 @@
 [
  {
   "type": "CreateChannel",
-  "jobId": "better-massive-shaggy-battery",
```
andrewgordstewart, line 4: I think it's pretty nasty to check in these (large) files that randomly change. I don't really see the benefit of having example load files checked in.

Consider:
- I change the load generator/runner in some way
- I forget to update the sample files
Then, when you checkout my changes and try to run the sample files, they won't work!

packages/server-wallet/e2e-testing/scripts/start-ganache.ts
lalexgap: `process.on('exit', ...)` is not `async` so we would have to introduce some mechanism to handle waiting for the async operation to complete. `async-exit-hook` does that for us.

andrewgordstewart: which async operation? writing to a file? You could avoid that with `jsonFile.writeFileSync`, right?

.circleci/config.yml
lalexgap: This was copied from the existing `server-wallet-with-peers-test` job. It looks like all jobs currently have `     no_output_timeout: 30m` ?

packages/server-wallet/e2e-testing/scripts/start-ganache.ts
lalexgap: We are also getting the on-chain balances (ie:  `provider.getBalance`)

packages/server-wallet/e2e-testing/scripts/start-ganache.ts
lalexgap: ad3ec9a

.circleci/config.yml
andrewgordstewart: I see! Might be worth using the default of 10m, unless there's a good reason for that?

andrewgordstewart (approved):
🎉 nice work!

PR #3710, opened 2021-07-30
remove defunct test helper

geoknee:
some cleanup following #3624


packages/nitro-protocol/test/test-helpers.ts
```diff
@@ -3,6 +3,7 @@ import {Contract, ethers, BigNumberish, BigNumber, constants, providers, Event}
 import {ChallengeClearedEvent, ChallengeRegisteredStruct} from '../src/contract/challenge';
 import {Bytes} from '../src/contract/types';
 import {channelDataToStatus} from '../src/contract/channel-storage';
+import {Outcome} from '../src/contract/outcome';
```
NiloCK, line 6:
```suggestion
```
Duplicate identifier from the existing import on the next line.

NiloCK (approved):
Looks good. Side note is that it seems feasible that some static analysis could mark unused functions like these.

PR #3711, opened 2021-07-30
Integrate exit format ts

### LINKED ISSUE #3707: `import {ExitFormat as Outcome}` into nitro-protocol typescript files

geoknee:
Should close #3707 

TODO: 

- [x] `yarn tsc` works  in `nitro-protocol`. 
- [ ] ~nitro protocol needs to be isolated from the other packages in the monorepo, or adapters need to be written to allow those dependent packages to compile (punt to another PR)~ we will punt this due to time constraints


packages/nitro-protocol/src/contract/outcome.ts
```diff
@@ -1,181 +1,67 @@
...
+      },
+    ],
+  },
+];
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const exampleGuaranteeOutcome2: ExitFormat.Exit = exampleGuaranteeOutcome1; // GuaranteeOutcome is assignable to Exit
+
+export function encodeGuaranteeData(destinations: string[]): BytesLike {
+  return defaultAbiCoder.encode(['bytes32[]'], [destinations]);
```
NiloCK, line 62:
```suggestion
  return defaultAbiCoder.encode(['bytes32[]'], destinations);
```
:question: 

`encode` here wants an array as its second parameter, but `destinations` as passed in is already an array. This double-wrapped array seems intentional, based on `decodeGuaranteeData` function, which returns the first element of the decoded data.

Just pointing at it because it pops out as a potential typo style error.

NiloCK (approved):
Looks great.

packages/nitro-protocol/src/contract/multi-asset-holder.ts
```diff
@@ -1,15 +1,15 @@
-import {utils, BigNumber, ethers} from 'ethers';
+import {utils, BigNumber, ethers, constants} from 'ethers';
+import ExitFormat from '@statechannels/exit-format';
```
andrewgordstewart, line 2: Is this a long-term decision?

packages/nitro-protocol/src/contract/outcome.ts
geoknee: I think this is just the [ethers API](https://docs.ethers.io/v5/api/utils/abi/coder/#AbiCoder-encode) which calls for an array of types and an array of values. 

The `decode()[0]` returning the first element is (I think) also just a quirk of the ethers API, which returns the decoded data inside of an array.

geoknee: On https://playground.ethers.org/: 

```
utils.defaultAbiCoder.encode(['bytes32[]'], [[constants.HashZero]]);
"0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000"
utils.defaultAbiCoder.encode(['bytes32[]'], [constants.HashZero]);
expected array value (argument=null, value="0x0000000000000000000000000000000000000000000000000000000000000000", code=INVALID_ARGUMENT, version=abi/5.4.0)
```

packages/nitro-protocol/src/contract/multi-asset-holder.ts
geoknee: I would say it is a medium term decision. 

It shouldn't be difficult to absorb that package into `nitro-protocol` if we want to in the future. Keeping it separate makes it easier for other projects to use.

PR #3714, opened 2021-08-03
add claim implementation

geoknee:
I have had to use some tricks to avoid[ `stack too deep`. ](https://soliditydeveloper.com/stacktoodeep)

Notably, block scoping and the use of a new struct `ClaimArgs` to reduce the number of variables on the stack. 

See #3683

### DIFF
--- a/packages/nitro-protocol/contracts/MultiAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
@@ -122,7 +122,7 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
             totalPayouts
         );
 
-        _apply_transfer_interactions(outcome[assetIndex], exitAllocations); // in future, we may pass in the top-level metadata field too
+        _apply_transfer_interactions(outcome[assetIndex], exitAllocations);
     }
 
     function _apply_transfer_checks(
@@ -164,7 +164,7 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         )
     {
         // `indices == []` means "pay out to all"
-        // Note: by initializing payouts to be an array of fixed length, its entries are initialized to be `0`
+        // Note: by initializing exitAllocations to be an array of fixed length, its entries are initialized to be `0`
         exitAllocations = new Outcome.Allocation[](
             indices.length > 0 ? indices.length : allocations.length
         );
@@ -188,7 +188,7 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
                 // found a match
                 // reduce the current allocationItem.amount
                 newAllocations[i].amount = allocations[i].amount - affordsForDestination;
-                // increase the relevant payout
+                // increase the relevant exit allocation
                 exitAllocations[k] = Outcome.Allocation(
                     allocations[i].destination,
                     affordsForDestination,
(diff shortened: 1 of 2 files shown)


packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -242,6 +242,255 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+            sourceStateHash,
+            keccak256(sourceOutcomeBytes),
+            sourceChannelId
+        );
+
+        sourceOutcome = Outcome.decodeExit(sourceOutcomeBytes);
+        targetOutcome = Outcome.decodeExit(targetOutcomeBytes);
+        address asset = sourceOutcome[assetIndex].asset;
+        require(targetOutcome[assetIndex].asset == asset, 'asset mismatch');
+        initialAssetHoldings = holdings[asset][sourceChannelId];
```
kerzhner, line 339: I believe this check implies that the funding channel outcome should have the same assets (in the same order) as the virtually-funded (aka target) channel outcome. I believe that a single channel can fund many virtually-funded channels. In that case, every virtually-funded channel must contain every asset used by any other virtually-funded channel. 

I am not really recommending a change here, but just want to make sure my understanding is correct.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -242,6 +242,255 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+     */
+    function claim(uint256 assetIndex, ClaimArgs memory claimArgs) external override {
+        Outcome.SingleAssetExit[] memory sourceOutcome;
+        Outcome.SingleAssetExit[] memory targetOutcome;
+        uint256 initialAssetHoldings;
+        {
+            bytes memory sourceOutcomeBytes = claimArgs.sourceOutcomeBytes;
+            bytes memory targetOutcomeBytes = claimArgs.targetOutcomeBytes;
+            (sourceOutcome, targetOutcome, initialAssetHoldings) = _apply_claim_checks(
+                assetIndex,
```
kerzhner, line 259: Would it be cleaner to just pass in `ClaimArgs`? But perhaps that uses more gas.

kerzhner (commented):
I am sure there are plans for this, but I didn't see any testing for the claim implementation.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: I think your understanding is correct. I think it is worth avoiding that requirement if we can, so I'll take a stab at having both a `sourceAssetIndex` and `targetAssetIndex`.

geoknee:
> I am sure there are plans for this, but I didn't see any testing for the claim implementation.

Yep #3709

packages/nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
```diff
@@ -40,6 +40,37 @@ interface IMultiAssetHolder {
...
+     * @param assetIndex Will be used to slice the outcome into a single asset outcome.
+     * @param sourceChannelId Unique identifier for a guarantor state channel.
+     * @param sourceStateHash Hash of the state stored when the guarantor channel finalized.
+     * @param sourceOutcomeBytes The abi.encode of guarantor channel outcome
+     * @param indexOfTargetInSource The index of the guarantee allocation to the target channel in the source outcome.
+     * @param targetStateHash Hash of the state stored when the target channel finalized.
+     * @param targetOutcomeBytes The abi.encode of target channel outcome
+     * @param indices Array with each entry denoting the index of a destination (in the target channel) to transfer funds to. Should be in increasing order. An empty array indicates "all"
+     */
+    struct ClaimArgs {
```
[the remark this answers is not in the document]
geoknee: I think this one in particular was useful to have in the scope of the entire function. Putting it inside `ClaimArgs` meant that I was just immediately unpacking it and using up a stack slot anyway.

That said, I have now put it inside :-) It seems less surprising that way, and neater.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: 42dd3b6

packages/nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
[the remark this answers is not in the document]
geoknee: 42dd3b6

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: It would be cleaner, I think -- I went ahead and did that 42dd3b6.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -243,10 +243,280 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+        _requireChannelFinalized(targetChannelId);
+        _requireMatchingFingerprint(
+            claimArgs.targetStateHash,
+            keccak256(targetOutcomeBytes),
+            targetChannelId
+        );
+    }
+
+    /**
+     * @dev Computes side effects for the claim function. First, computes the amount the source channel can afford for the target. Then, computes and returns an updated allocations for the source and for the target, as well as exit allocations (to be paid out). It does this by walking the target allocations, testing against the guarantee in the source, and conditionally siphoning money out. See the Nitro paper.
```
NiloCK, line 355:
```suggestion
     * @dev Computes side effects for the claim function. First, computes the amount the source channel can afford for the target. Then, computes and returns updated allocations for the source and for the target, as well as exit allocations (to be paid out). It does this by walking the target allocations, testing against the guarantee in the source, and conditionally siphoning money out. See the Nitro paper.
```

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -243,10 +243,280 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+            if (targetSurplus == 0) break;
+            for (uint256 i = 0; i < newTargetAllocations.length; i++) {
+                if (targetSurplus == 0) break;
+                // search for it in the allocation
+                if (guaranteeDestinations[j] == newTargetAllocations[i].destination) {
+                    // if we find it, compute new amount
+                    uint256 affordsForDestination = min(targetAllocations[i].amount, targetSurplus);
+                    // decrease surplus by the current amount regardless of hitting a specified index
+                    targetSurplus -= affordsForDestination;
+                    if ((indices.length == 0) || ((k < indices.length) && (indices[k] == i))) {
```
NiloCK, line 431: Earmarking for #3728. Can the intent / logic of this check hide behind some nicely named internal pure function?

packages/nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
```diff
@@ -40,6 +40,37 @@ interface IMultiAssetHolder {
...
+        bytes32 sourceChannelId;
+        bytes32 sourceStateHash;
+        bytes sourceOutcomeBytes;
+        uint256 sourceAssetIndex;
+        uint256 indexOfTargetInSource;
+        bytes32 targetStateHash;
+        bytes targetOutcomeBytes;
+        uint256 targetAssetIndex;
+        uint256[] indices;
```
NiloCK, line 64: Earmark for #3728. Lots of arrays exist in implementation of `claim`, so what's being indexed isn't necessarily obvious.

Maybe a rename to `payeeIndices`?

NiloCK (approved):
Looks great. Checks-Effects-Interactions refactor provides a substantial readability boost.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: good idea!

packages/nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
```diff
@@ -40,6 +40,37 @@ interface IMultiAssetHolder {
         uint256[] memory indices
     ) external;
 
+    /**
+     * @param assetIndex Will be used to slice the outcome into a single asset outcome.
```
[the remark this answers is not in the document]
geoknee: Good catch. Unfortunately these comments are a bit of an abuse of the natspec system -- it would be nice if the compiler picked up mismatches as it would with function parameter docstrings.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: Let's circle back to this, for sure. It's also something that should be applied to master, regardless of the fate of this branch.

PR #3715, opened 2021-08-03
add missing dependency

geoknee:
Things like this are difficult to spot in a `yarn` monorepo -- that is, until someone tries to use one of our packages by installing from npm.

I discovered this while experimenting with pinning dependants of nitro-protocol to a version published to npm.


lalexgap (approved):

andrewgordstewart:
Is this related to https://github.com/statechannels/statechannels/issues/3717?

geoknee:
I think not. It's the same kind of issue, though.

PR #3716, opened 2021-08-03
Add basic drop/delay to message service and load testing

### LINKED ISSUE #3619: Add drop/delay to Message Service
Add the ability to drop and delay messages. We do this in the `TestMessageService` so we could recopy that logic.

Maybe we could find some tool designed for this that could sit in the middle?

lalexgap:
Takes the basic drop and delay logic from the `TestMessageService` and adds it to the `SocketIoMessageService`.

The `start-load-node` and `start-all` both take two new arguments:
- `meanDelay`: The average delay that we should wait for before sending a message. 
- `dropRatePercentage` The percentage of messages that should be dropped.

I've also added a simple timer for job processing to the load node.

Fixes #3619


packages/server-wallet/src/message-service/socket-io-message-service.ts
```diff
@@ -22,13 +30,32 @@ export class SocketIOMessageService implements MessageServiceInterface {
...
+
   public async send(messages: Message[]): Promise<void> {
-    if (messages.length > 0) {
-      this.server.emit('message', messages);
+    if (this._frozen) {
+      this._messageQueue.push(...messages);
+    } else {
+      const shouldDrop = Math.random() > 1 - this._latencyOptions.dropRate;
+
+      if (!shouldDrop) {
```
lalexgap, line 43: This is based on the logic in the [TestMessageService](https://github.com/statechannels/statechannels/blob/master/packages/server-wallet/src/message-service/test-message-service.ts)

packages/server-wallet/src/message-service/socket-io-message-service.ts
```diff
@@ -1,12 +1,20 @@
...
+import delay from 'delay';
+import {AbortController} from 'abort-controller';
 
 import {MessageHandler, MessageServiceFactory, MessageServiceInterface} from './types';
+import {LatencyOptions} from './test-message-service';
 
 export class SocketIOMessageService implements MessageServiceInterface {
   private server: Server;
   private peers: ClientSocket[] = [];
+  private _frozen = false;
```
kerzhner, line 13: Can this variable ever get toggled? Have we tested this option?

packages/server-wallet/src/message-service/socket-io-message-service.ts
```diff
@@ -22,13 +30,32 @@ export class SocketIOMessageService implements MessageServiceInterface {
...
+        }
+        if (!this._destroyed) {
+          this.server.emit('message', messages);
+        }
+      }
     }
   }
 
   public async destroy(): Promise<void> {
+    this._destroyed = true;
```
kerzhner, line 58: Do we need to call `this._abortController.abort();`?

packages/server-wallet/e2e-testing/scripts/start-load-node.ts
```diff
@@ -129,6 +147,8 @@ async function setupNode(): Promise<WalletLoadNode> {
   for (const {messagePort} of peers) {
     await serverNode.registerMessagePeer(messagePort);
   }
+  const dropRate = dropRatePercentage === 0 ? 0 : dropRatePercentage / 100;
```
kerzhner, line 150: Why do we need to check if the dropRatePercentage is `0`?

kerzhner (approved):

packages/server-wallet/src/message-service/socket-io-message-service.ts
```diff
@@ -22,13 +30,32 @@ export class SocketIOMessageService implements MessageServiceInterface {
...
+    this._latencyOptions = options;
+  }
+
   public async send(messages: Message[]): Promise<void> {
-    if (messages.length > 0) {
-      this.server.emit('message', messages);
+    if (this._frozen) {
+      this._messageQueue.push(...messages);
+    } else {
+      const shouldDrop = Math.random() > 1 - this._latencyOptions.dropRate;
```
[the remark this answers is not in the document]
lalexgap: Interestingly enough `prettier` won't let me do that.
image


Why do you think the statement would always be `false` ?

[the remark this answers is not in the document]
lalexgap: > Are we planning on using the `SocketIOMessageService` in production? If so, how do we feel about test logic in a production service?

I was originally thinking about creating a separate `TestSocketIOMessageService` but I think it might make more sense to keep it simple and then create a production-ready message service when we need one.

Perhaps I should rename the existing `SocketIOMessageService` to `TestSocketIOMessageService` or some variation of that to indicate that it's not production-ready.

packages/server-wallet/e2e-testing/scripts/start-load-node.ts
lalexgap: Oops, we definitely don't need to do that.

packages/server-wallet/src/message-service/socket-io-message-service.ts
lalexgap: Oops, this was copy and pasted by accident.

packages/server-wallet/e2e-testing/scripts/start-load-node.ts
lalexgap: Fixed in 4d37d37

packages/server-wallet/src/message-service/socket-io-message-service.ts
lalexgap: Done in 4d37d37

packages/server-wallet/src/message-service/socket-io-message-service.ts
lalexgap: Removed in 4d37d37

packages/server-wallet/src/message-service/socket-io-message-service.ts
```diff
@@ -22,13 +30,32 @@ export class SocketIOMessageService implements MessageServiceInterface {
...
 
+  public setLatencyOptions(options: LatencyOptions): void {
+    this._latencyOptions = options;
+  }
+
   public async send(messages: Message[]): Promise<void> {
-    if (messages.length > 0) {
-      this.server.emit('message', messages);
+    if (this._frozen) {
+      this._messageQueue.push(...messages);
```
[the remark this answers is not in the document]
lalexgap: Removed in 4d37d37

packages/server-wallet/src/message-service/socket-io-message-service.ts
```diff
@@ -22,13 +30,32 @@ export class SocketIOMessageService implements MessageServiceInterface {
...
+      const shouldDrop = Math.random() > 1 - this._latencyOptions.dropRate;
+
+      if (!shouldDrop) {
+        const {meanDelay} = this._latencyOptions;
+        if (meanDelay && meanDelay > 0) {
+          const delayAmount = meanDelay / 2 + Math.random() * meanDelay;
+
+          await delay(delayAmount, {signal: this._abortController.signal});
+        }
+        if (!this._destroyed) {
```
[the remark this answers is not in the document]
lalexgap: Removed in 4d37d37

PR #3723, opened 2021-08-04
Simplify Load Generator arguments

lalexgap:
Instead of using `ledgerRate` and `createLedgerDuration` arguments, the load-generator takes a `amountOfLedgerChannels`. 

CreateLedgerChannel steps will be scheduled in the first 5 seconds regardless of the duration. 

This makes the load-generator slightly more straightforward to use. I've also tried to clean up the argument descriptions so they're more understandable.


NiloCK (approved):
This looks good to me! Argument description updates are an improvement.

PR #3724, opened 2021-08-04
Fix Funding Typo

### LINKED ISSUE #3722: Node B said it was using direct funding when it should be using ledger funding

lalexgap:
Fixes #3722 
If the node wasn't assigned any create ledger steps it would log that direct funding was being used. We now check for the presence of a create ledger step for any serverId to avoid this.


packages/server-wallet/e2e-testing/wallet-load-node.ts
```diff
@@ -127,10 +127,11 @@ export class WalletLoadNode {
     const ourSteps = this.steps.filter(s => s.serverId === this.loadNodeConfig.serverId);
     const createdAmount = ourSteps.filter(s => s.type === 'CreateChannel').length;
     const ledgerAmount = ourSteps.filter(s => s.type === 'CreateLedgerChannel').length;
+    const isLedgerFunding = this.steps.some(s => s.type === 'CreateLedgerChannel');
```
NiloCK, line 130:
```suggestion
    const isLedgerFunding = this.steps.some(s => s.type === 'CreateLedgerChannel');
    
```

NiloCK (approved):
:+1:

PR #3725, opened 2021-08-04
Chain Service reads process.env.NITRO_ADJUDICATOR_ADDRESS when constructed

lalexgap:
Read the `process.env.NITRO_ADJUDICATOR_ADDRESS` var when constructing the `chainService` instead of when the chain servic file is loaded.

This is to make testing more convenient. Previously you could inadvertently load the file before you had a chance to set `process.env.NITRO_ADJUDICATOR_ADDRESS` resulting in the `chain-service` using a `0x0000...`.

I ran into this with the stress test as the LoadNodes have references to the `ServerWallet` which has a reference to `ChainService`. This meant that `nitroAdjudicatorAddress` got set as soon as the LoadNode file was loaded before we had a chance to set the `process.env` vars.

### DIFF
--- a/packages/server-wallet/src/chain-service/chain-service.ts
+++ b/packages/server-wallet/src/chain-service/chain-service.ts
@@ -53,10 +53,6 @@ type AllocationUpdatedEvent = {
   ethersEvent: Event;
 } & AllocationUpdatedArg;
 
-/* eslint-disable no-process-env, */
-const nitroAdjudicatorAddress = makeAddress(
-  process.env.NITRO_ADJUDICATOR_ADDRESS || constants.AddressZero
-);
 export class ChainService implements ChainServiceInterface {
   private logger: Logger;
   private readonly ethWallet: NonceManager;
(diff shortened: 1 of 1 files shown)


lalexgap:
Without this change, this [PR](https://github.com/statechannels/statechannels/pull/3706) would fail on CI

andrewgordstewart:
Why doesn’t the chain service accept it as a constructor argument?

(this question feels like déjà vue — I thought the wallet did not need to read from process.env)

NiloCK:
This seems fishy to me! I think it makes sense for the address to be part of the the chain service's constructor, but for the purpose of this PR that may only be kicking the problem somewhere else.

What I'm more confused about is:

> This meant that `nitroAdjudicatorAddress` got set as soon as the LoadNode file was loaded **before we had a chance to set the process.env vars.**

aren't process.env values generally set before run-time?

lalexgap:
> Why doesn’t the chain service accept it as a constructor argument?

Would this mean that the `ServerWallet` would also accept the address as a constructor argument? Now that we only have 1 contract it might not be too painful. (Although it's going to touch a lot of files!).

I think ideally the contract addresses would get baked in at compile time (maybe per chain id). So if you use the wallet you just need to tell it what chain to use and not worry about contract addresses.

> (this question feels like déjà vue — I thought the wallet did not need to read from process.env)

I think that we stopped the wallet from being **configured** by process.env vars but we didn't consider contract addresses "configuration"?

lalexgap:
> This seems fishy to me! I think it makes sense for the address to be part of the the chain service's constructor, but for the purpose of this PR that may only be kicking the problem somewhere else.
> 
> What I'm more confused about is:
> 
> > This meant that `nitroAdjudicatorAddress` got set as soon as the LoadNode file was loaded **before we had a chance to set the process.env vars.**
> 
> aren't process.env values generally set before run-time?

We often set process env variables in a [test setup script](https://github.com/statechannels/statechannels/blob/ag/assert-db/packages/server-wallet/jest/chain-setup.ts#L50:L50) to set the contract addresses.

PR #3730, opened 2021-08-05
Remove sample load files

lalexgap:
😢


andrewgordstewart (approved):
✂️

(I'm happy to add these back in if there's clear demand!)

PR #3736, opened 2021-08-11
Compute claim effects and interactions test

geoknee:
https://ethereum.stackexchange.com/questions/26279/why-memory-arrays-are-unresizable


packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -416,6 +416,12 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
 
         uint256 targetSurplus = min(sourceSurplus, sourceAllocations[indexOfTargetInSource].amount);
 
+        require(
+            sourceAllocations[indexOfTargetInSource].allocationType ==
+                uint8(Outcome.AllocationType.guarantee),
+            'not a guarante allocation'
```
geoknee, line 422:
```suggestion
            'not a guarantee allocation'
```

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -416,6 +416,12 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+        );
```
NiloCK, line 423: This should probably live inside of `_apply_claim_checks`. I'm not sure about applying a 'suggestion' in the review that isn't in-line, but inserting:
```
require(
            sourceOutcome[sourceAssetIndex].allocations[targetAssetIndex].allocationType ==
                uint8(Outcome.AllocationType.guarantee),
            'not a guarantee allocation'
        );
```
in or around line 339 seems to work.

NiloCK (approved):
Looks great. Sorry for the delayed review.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: good shout! I had thought that at that early point in the flow we didn't yet have all the variables we needed unpacked. But... we do.

PR #3737, opened 2021-08-12
tweak comment

geoknee:


NiloCK (approved):
Would it be petty to complain about a typo in the branch name?

Issue #3672, opened 2021-07-22
check assets match in target and guarantor during claim


geoknee:
(on the pushOutcome-trunk branch #3624).

We should write a test that fails, first.

geoknee:
We also need to consider whether this fix needs to be applied to the exit-format repo, too!

NiloCK:
A little more context. Currently, the external `MultiAssetHolder.claim(...)` [function](https://github.com/statechannels/statechannels/blob/ea8f896b24658212f680e3a1664729065e556057/packages/nitro-protocol/contracts/MultiAssetHolder.sol#L134) receives an `assetIndex` uint from the caller, which it uses to:

- [pull](https://github.com/statechannels/statechannels/blob/ea8f896b24658212f680e3a1664729065e556057/packages/nitro-protocol/contracts/MultiAssetHolder.sol#L168) an individual `AssetOutcome`, the guarantee, out of the guarantor channel's outcome and decode it
- [determine](https://github.com/statechannels/statechannels/blob/ea8f896b24658212f680e3a1664729065e556057/packages/nitro-protocol/contracts/MultiAssetHolder.sol#L176) the asset in question
- [pull](https://github.com/statechannels/statechannels/blob/ea8f896b24658212f680e3a1664729065e556057/packages/nitro-protocol/contracts/MultiAssetHolder.sol#L195) an individual `AssetOutcome` out of the target channel's outcome, decode it, and then decode its list of `AllocationItems`

The determined asset, guarantee, and target Allocation are passed to the internal `_claim` function.

Potential issue is that there are no checks in place to confirm that the target channel's `outcome[assetIndex]` refers to the same determined asset as the guarantee, which opens up a potential attack vector where, say, a guarantee of USDT tokens is used to update the ETH allocation on a target channel.

PR #3740, opened 2021-08-13
Fix serious bug in `claim` and enhance sanity checks in gas benchmarking suite. 

geoknee:
The `claim` function uses an array called `payouts` (internally) to build up payouts for each of the destinations in the target channel at the supplied `indices`. We are not currently updating this array correctly in general: particularly in the case where the `guarantee` has an entry that is absent from the target outcome. Essentially, a different party gets the funds that should go to their rightful owner.

The change here decouples indexing of `payouts` from indexing of `indices`. `payouts` will always contain an entry for every allocation in the target, but many of them may be zero. 

### Changes
I discovered this bug through our gas benchmarking suite, but only in a very roundabout way, by running it against the exit format branch which has a non-buggy claim implementation and looking carefully at the gas consumption. On the base branch here, the guarantee in the benchmark is incorrectly ordered, but the bug in `claim` meant that Alice still recovered her funds. So the two problems cancelled each other out, making the problems hard to spot.

This is a good approach because it increases our test "depth" and "breadth" and fixes a critical security bug.

The new test infra will be useful for future refactor of `claim` (including the aforementioned exit format version). 

## How Has This Been Tested? 

 I have added some sanity check assertions into the gas benchmarking suite that would have caught this. I have also added some appropriate test cases to our unit testing suite.

- [x] Is it security related?

---
### Code quality
### Project management

### DIFF
--- a/packages/nitro-protocol/contracts/MultiAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
@@ -368,7 +368,7 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
 
         for (uint256 j = 0; j < payouts.length; j++) {
             if (payouts[j] > 0) {
-                bytes32 destination = allocation[indices.length > 0 ? indices[j] : j].destination;
+                bytes32 destination = allocation[j].destination;
                 // storage updated BEFORE external contracts called (prevent reentrancy attacks)
                 if (_isExternalDestination(destination)) {
                     _transferAsset(asset, _bytes32ToAddress(destination), payouts[j]);
@@ -398,7 +398,7 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
     {
         // `indices == []` means "pay out to all"
         // Note: by initializing payouts to be an array of fixed length, its entries are initialized to be `0`
-        payouts = new uint256[](indices.length > 0 ? indices.length : allocation.length);
+        payouts = new uint256[](allocation.length);
         totalPayouts = 0;
         allocatesOnlyZeros = true; // switched to false if there is an item remaining with amount > 0
         uint256 surplus = initialHoldings; // tracks funds available during calculation
@@ -430,14 +430,14 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
                         // reduce the new allocationItem.amount
                         newAllocation[i].amount -= affordsForDestination;
                         // increase the relevant payout
-                        payouts[k] += affordsForDestination;
+                        payouts[i] += affordsForDestination;
                         totalPayouts += affordsForDestination;
+                        // move on to the next supplied index
+                        ++k;
                     }
                     break; // start again with the next guarantee destination
                 }
             }
-            // move on to the next supplied index
-            ++k;
         }
 
         for (uint256 i = 0; i < allocation.length; i++) {
--- a/packages/nitro-protocol/contracts/MultiAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
@@ -432,12 +432,12 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
                         // increase the relevant payout
                         payouts[k] += affordsForDestination;
                         totalPayouts += affordsForDestination;
-                        // move on to the next supplied index
-                        ++k;
                     }
                     break; // start again with the next guarantee destination
                 }
             }
+            // move on to the next supplied index
+            ++k;
         }
 
         for (uint256 i = 0; i < allocation.length; i++) {
--- a/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -1,3 +1,4 @@
+import {constants} from 'ethers';
 import {encodeOutcome} from '../src';
 import {MAGIC_ADDRESS_INDICATING_ETH} from '../src/transactions';
 
@@ -10,9 +11,12 @@ import {
   LforG,
   G,
   J,
+  Alice,
+  Ingrid,
+  Bob,
 } from './fixtures';
 import {gasRequiredTo} from './gas';
-import {nitroAdjudicator, token} from './vanillaSetup';
+import {nitroAdjudicator, provider, token} from './vanillaSetup';
 
 /**
  * Ensures the supplied asset holder always has a nonzero token balance.
@@ -186,7 +190,7 @@ describe('Consumes the expected gas for sad-path exits', () => {
     ).toEqual(gasRequiredTo.ETHexitSadLedgerFunded.vanillaNitro.total);
   });
 
-  it(`when exiting a virtual funded (with ETH) channel`, async () => {
+  it.only(`when exiting a virtual funded (with ETH) channel`, async () => {
     // begin setup
     await (
       await nitroAdjudicator.deposit(MAGIC_ADDRESS_INDICATING_ETH, LforG.channelId, 0, 10, {
@@ -234,6 +238,7 @@ describe('Consumes the expected gas for sad-path exits', () => {
     ]);
     // end wait
     // challenge L,G,J,X + timeout ⬛ -> (L) -> (G) -> (J) -> (X) -> 👩
+    await logEthHoldingsAndBalances();
     await expect(
       await nitroAdjudicator.transferAllAssets(
         LforG.channelId,
@@ -242,6 +247,7 @@ describe('Consumes the expected gas for sad-path exits', () => {
       )
     ).toConsumeGas(gasRequiredTo.ETHexitSadVirtualFunded.vanillaNitro.transferAllAssetsL);
     // transferAllAssetsL  ⬛ --------> (G) -> (J) -> (X) -> 👩
+    await logEthHoldingsAndBalances();
     await expect(
       await nitroAdjudicator.claim(
         0,
@@ -253,6 +259,7 @@ describe('Consumes the expected gas for sad-path exits', () => {
         [] // meaning "all"
       )
     ).toConsumeGas(gasRequiredTo.ETHexitSadVirtualFunded.vanillaNitro.claimG);
+    await logEthHoldingsAndBalances();
     // claimG                      ⬛ ----------------------> (X) -> 👩
     await expect(
       await nitroAdjudicator.transferAllAssets(
@@ -270,5 +277,37 @@ describe('Consumes the expected gas for sad-path exits', () => {
         (a, b) => a + b
       ) - gasRequiredTo.ETHexitSadVirtualFunded.vanillaNitro.total
     ).toEqual(gasRequiredTo.ETHexitSadVirtualFunded.vanillaNitro.total);
+
+    await logEthHoldingsAndBalances();
   });
 });
+
+async function logEthHoldingsAndBalances() {
+  const internalDestinations = {
+    LforG: LforG.channelId,
+    G: G.channelId,
+    J: J.channelId,
+    X: X.channelId,
+  };
+  const externalDestinations = {
+    Alice: Alice.address,
+    Bob: Bob.address,
+    Ingrid: Ingrid.address,
+  };
+
+  const results = {};
+  await Promise.all([
+    ...Object.keys(internalDestinations).map(
+      async key =>
+        (results[key] = await nitroAdjudicator.holdings(
+          constants.AddressZero,
+          internalDestinations[key]
+        ))
+    ),
+    ...Object.keys(externalDestinations).map(
+      async key => (results[key] = await provider.getBalance(externalDestinations[key]))
+    ),
+  ]);
+  Object.keys(results).map(key => (results[key] = results[key].toNumber()));
+  console.log(results);
+}
--- a/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -1,4 +1,5 @@
 import {constants} from 'ethers';
+
 import {encodeOutcome} from '../src';
 import {MAGIC_ADDRESS_INDICATING_ETH} from '../src/transactions';
 
@@ -190,7 +191,7 @@ describe('Consumes the expected gas for sad-path exits', () => {
     ).toEqual(gasRequiredTo.ETHexitSadLedgerFunded.vanillaNitro.total);
   });
 
-  it.only(`when exiting a virtual funded (with ETH) channel`, async () => {
+  it(`when exiting a virtual funded (with ETH) channel`, async () => {
     // begin setup
     await (
       await nitroAdjudicator.deposit(MAGIC_ADDRESS_INDICATING_ETH, LforG.channelId, 0, 10, {
--- a/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -1,5 +1,3 @@
-import {constants} from 'ethers';
-
 import {encodeOutcome} from '../src';
 import {MAGIC_ADDRESS_INDICATING_ETH} from '../src/transactions';
 
@@ -12,12 +10,10 @@ import {
   LforG,
   G,
   J,
-  Alice,
-  Ingrid,
-  Bob,
+  assertETHSanityChecks,
 } from './fixtures';
 import {gasRequiredTo} from './gas';
-import {nitroAdjudicator, provider, token} from './vanillaSetup';
+import {nitroAdjudicator, token} from './vanillaSetup';
 
 /**
  * Ensures the supplied asset holder always has a nonzero token balance.
@@ -282,48 +278,3 @@ describe('Consumes the expected gas for sad-path exits', () => {
     await assertETHSanityChecks({Alice: 5, Bob: 5, Ingrid: 0}, {LforG: 0, G: 0, J: 0, X: 0});
   });
 });
-
-interface ETHBalances {
-  Alice: number;
-  Bob: number;
-  Ingrid: number;
-}
-
-interface ETHHoldings {
-  LforG: number;
-  G: number;
-  J: number;
-  X: number;
-}
-
-async function assertETHSanityChecks(
-  ethBalances: Partial<ETHBalances>,
-  ethHoldings: Partial<ETHHoldings>
-) {
-  const internalDestinations: {[Property in keyof ETHHoldings]: string} = {
-    // todo type this
-    LforG: LforG.channelId,
-    G: G.channelId,
-    J: J.channelId,
-    X: X.channelId,
-  };
-  const externalDestinations: {[Property in keyof ETHBalances]: string} = {
-    Alice: Alice.address,
-    Bob: Bob.address,
-    Ingrid: Ingrid.address,
-  };
-  await Promise.all([
-    ...Object.keys(ethHoldings).map(async key => {
-      expect(
-        (
-          await nitroAdjudicator.holdings(constants.AddressZero, internalDestinations[key])
-        ).toNumber()
-      ).toEqual(ethHoldings[key]);
-    }),
-    ...Object.keys(ethBalances).map(async key => {
-      expect((await provider.getBalance(externalDestinations[key])).toNumber()).toEqual(
-        ethBalances[key]
-      );
-    }),
-  ]);
-}
--- a/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -11,6 +11,8 @@ import {
   G,
   J,
   assertETHSanityChecks,
+  amountForAlice,
+  amountForBob,
 } from './fixtures';
 import {gasRequiredTo} from './gas';
 import {nitroAdjudicator, token} from './vanillaSetup';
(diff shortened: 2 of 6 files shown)


geoknee:
It would be good to know if this bug was introduced by #3624.

NiloCK:
Stray thought: this is the `destination` equivalent to the `asset-mismatch` issue (#3672) which was correctly anticipated and dealt with during #3624.

packages/nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -230,3 +230,50 @@ export async function challengeChannelAndExpectGas(
...
+          await nitroAdjudicator.holdings(constants.AddressZero, internalDestinations[key])
+        ).toNumber()
+      ).toEqual(ethHoldings[key]);
+    }),
+    ...Object.keys(ethBalances).map(async key => {
+      expect((await provider.getBalance(externalDestinations[key])).toNumber()).toEqual(
+        ethBalances[key]
+      );
+    }),
+  ]);
```
NiloCK, line 278:
```suggestion
  await Promise.all([
    ...Object.keys(ethHoldings).map(async channelId => {
      expect(
        (
          await nitroAdjudicator.holdings(constants.AddressZero, internalDestinations[channelId])
        ).toNumber()
      ).toEqual(ethHoldings[channelId]);
    }),
    ...Object.keys(ethBalances).map(async address => {
      expect((await provider.getBalance(externalDestinations[address])).toNumber()).toEqual(
        ethBalances[address]
      );
    }),
  ]);
```

packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -270,5 +274,7 @@ describe('Consumes the expected gas for sad-path exits', () => {
         (a, b) => a + b
       ) - gasRequiredTo.ETHexitSadVirtualFunded.vanillaNitro.total
     ).toEqual(gasRequiredTo.ETHexitSadVirtualFunded.vanillaNitro.total);
+
+    await assertETHSanityChecks({Alice: 5, Bob: 5, Ingrid: 0}, {LforG: 0, G: 0, J: 0, X: 0});
```
NiloCK, line 278: I see the 5,5 outcome specified in the creation of channel `X` in fixtues.ts.

A future work could be to add a utility function to `TestChannel` which outputs the inputs for these sanity assertions, avoiding the duplicated input of this sort of information.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -430,7 +430,7 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
                         // reduce the new allocationItem.amount
                         newAllocation[i].amount -= affordsForDestination;
                         // increase the relevant payout
-                        payouts[k] += affordsForDestination;
+                        payouts[i] += affordsForDestination;
```
NiloCK, line 433: I've been looking at this for a good few minutes and maybe I've confused myself, so I'll put the question forward rather than keep trying to see through to the answer.

Given that `payouts` is now the same length as the allocations array, can't some of this logic (ie, 428 check, k-indexing) be smoothed out?

NiloCK (approved):
Great catch.

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: I think there's some unavoidable complexity here in having to track an index for i) the guarantee, ii) the target and iii) the `indices` themselves.

There was some avoidable complexity in having the payouts be two different lengths based on `indices`. We've removed that here. 

That's not to say we can't keep improving this!

packages/nitro-protocol/gas-benchmarks/benchmark.test.ts
geoknee: 5357799 and 902889f

packages/nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -107,9 +107,9 @@ export const gasRequiredTo: GasRequiredTo = {
       challengeJ: 101068,
       challengeX: 92757,
       transferAllAssetsL: 65462,
-      claimG: 82369,
-      transferAllAssetsX: 114930,
-      total: 642227,
+      claimG: 94758,
+      transferAllAssetsX: 44478,
```
andrewgordstewart: Does this large change in gas reflect the fact that assets are being transferred to different addresses?.

geoknee: The large change here reflects the fact that X is actually not funded at all so cannot afford any payouts. 

At this point the benchmarking script is still wrong (in the sense that the channels have been setup incorrectly), whereas the on chain code has been (partly) fixed.

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
```diff
@@ -40,24 +41,26 @@ const reason6 = 'targetAsset != guaranteeAsset';
...
+    ${'15. (all) swap guarantee, overfunded, 2 destinations'} | ${{g: 12}} | ${['B', 'A']}         | ${{A: 5, B: 5}}       | ${[]}   | ${{A: 0, B: 0}}       | ${{g: 2}}        | ${{A: 5, B: 5}} | ${undefined}
+    ${'16. (all) underspecified guarantee, overfunded      '} | ${{g: 12}} | ${['B']}              | ${{A: 5, B: 5}}       | ${[]}   | ${{A: 5, B: 0}}       | ${{g: 7}}        | ${{A: 5, B: 5}} | ${undefined}
+    ${'17. guarantee and target assets do not match'}         | ${{g: 1}}  | ${['A', 'B']}         | ${{A: 5, B: 5}}       | ${[1]}  | ${{A: 4, B: 5}}       | ${{g: 0}}        | ${{A: 1}}       | ${reason6}
+    ${'18. absent guarantee entry, indices=[]'}               | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[]}   | ${{x: 10, I: 0}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
+    ${'19. absent guarantee entry, indices=[1]'}              | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[1]}  | ${{x: 10, I: 0}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
```
andrewgordstewart: This diff is really hard to understand. I don't have much of a chance of figuring out which tests are changed without a lot of manual effort.

In this case, I believe that the only thing changed is two rows were added, correct?

---

We should adopt one of two principles:
1. we use the `test.each(Array<TestCase>)` api, instead of the magic `test.each\`${magicTable}\`` API.
2. we commit formatting changes separately from "logic" changes, or in this case, changes to the test cases

In fact, we already claim to have adopted (2):
>  I have separated logic changes from refactor changes (formatting, renames, etc.)

but in this case, formatting-on-save encourages the developer to violate this principle.

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
```diff
@@ -40,24 +41,28 @@ const reason6 = 'targetAsset != guaranteeAsset';
...
+    ${'17. guarantee and target assets do not match'}                 | ${{g: 1}}  | ${['A', 'B']}         | ${{A: 5, B: 5}}       | ${[1]}  | ${{A: 4, B: 5}}       | ${{g: 0}}        | ${{A: 1}}       | ${reason6}
+    ${'18. absent guarantee entry, indices=[], bad guarantee order'}  | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[]}   | ${{x: 10, I: 0}}      | ${{g: 0}}        | ${{I: 10}}      | ${undefined}
+    ${'19. absent guarantee entry, indices=[1], bad guarantee order'} | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[1]}  | ${{x: 10, I: 0}}      | ${{g: 0}}        | ${{I: 10}}      | ${undefined}
+    ${'20. absent guarantee entry, indices=[]'}                       | ${{g: 10}} | ${['A', 'x', 'I']}    | ${{x: 10, I: 10}}     | ${[]}   | ${{x: 0, I: 10}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
+    ${'21. absent guarantee entry, indices=[1]'}                      | ${{g: 10}} | ${['A', 'x', 'I']}    | ${{x: 10, I: 10}}     | ${[0]}  | ${{x: 0, I: 10}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
```
andrewgordstewart, line 65: Am I right to understand that this test case would have failed before fixing the bug?

packages/nitro-protocol/contracts/MultiAssetHolder.sol
andrewgordstewart: @geoknee I'm not sure I understand your answer here.

After making `payouts.length` equal to `allocation.length`, we are:
1. iterating through each allocation item at index `i`
2. making sure that `k == i`
3. setting `payouts[k]` depending on whether something should be payed out to `allocations[i].destination`

I believe Colin is suggesting that we remove the point where we ensure `k == i`, resulting in this diff:
```
diff --git a/packages/nitro-protocol/contracts/MultiAssetHolder.sol b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
index 9cf170d51..4a35c43fb 100644
--- a/packages/nitro-protocol/contracts/MultiAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
@@ -402,7 +402,6 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         totalPayouts = 0;
         allocatesOnlyZeros = true; // switched to false if there is an item remaining with amount > 0
         uint256 surplus = initialHoldings; // tracks funds available during calculation
-        uint256 k = 0; // indexes the `indices` array
         //  We rely on the assumption that the indices are strictly increasing.
         //  This allows us to iterate over the destinations in order once, continuing until we hit the first index, then the second etc.
         //  If the indices were to decrease, we would have to start from the beginning: doing a full search for each index.
@@ -425,16 +424,14 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
                     uint256 affordsForDestination = min(allocation[i].amount, surplus);
                     // decrease surplus by the current amount regardless of hitting a specified index
                     surplus -= affordsForDestination;
-                    if ((indices.length == 0) || ((k < indices.length) && (indices[k] == i))) {
-                        // only if specified in supplied indices, or we if we are doing "all"
-                        // reduce the new allocationItem.amount
-                        newAllocation[i].amount -= affordsForDestination;
-                        // increase the relevant payout
-                        payouts[i] += affordsForDestination;
-                        totalPayouts += affordsForDestination;
-                        // move on to the next supplied index
-                        ++k;
-                    }
+                    // only if specified in supplied indices, or we if we are doing "all"
+                    // reduce the new allocationItem.amount
+                    newAllocation[i].amount -= affordsForDestination;
+                    // increase the relevant payout
+                    payouts[i] += affordsForDestination;
+                    totalPayouts += affordsForDestination;
+                    // move on to the next supplied index
+
                     break; // start again with the next guarantee destination
                 }
             }
```

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
geoknee: yes

packages/nitro-protocol/contracts/MultiAssetHolder.sol
andrewgordstewart: https://github.com/statechannels/statechannels/pull/3742

andrewgordstewart: ok, so I understand a bit more now:
`(indices.length == 0) || ((k < indices.length) && (indices[k] == i))` does not always evaluate to `true`, as I previously believed it to.

Instead, it evaluates to `true` when the index `i` is "the next" index supplied by the user to execute.

andrewgordstewart: @NiloCK do you think [this](https://github.com/statechannels/statechannels/pull/3743) clarifies things?

andrewgordstewart (commented):
I was trying to understand why `payouts` was changed to be equal in length to the allocation. It seems as though things can be simplified a bit.

The issue seems to be with fact that `payouts` is an array of amounts. This creates an implicit dependency: the _destination_ for the amount is not found in the `payouts` array.

One way to address this is to include the `destinations` in the `payouts` array.

Your solution instead uses `allocations` to (correctly) look up the destination by making sure that `payouts.length == allocations.length`. The solution you've given now duplicates some data: in `_claim`, you now have:
1. `allocation`
2. `newAllocation`
3. `payouts`, which is equal to `allocation[i].amount - newAllocation[i].amount` for `i = 0,...,numAllocations`

Then, iterating through `i`, `_claim` does:
- get `destination = allocation[i]`
- get `amount = payouts[i]`
- transfer `amount` to `destination`

This can be simplified by removing `payouts` entirely. `_claim` can instead
- `get amount = allocation[i] - newAllocation[i]`

Is [this](https://github.com/statechannels/statechannels/pull/3744) a better way to close the hole?

andrewgordstewart:
I hesitate to approve this PR, since it combines two changes:
1. sanity checks are added to gas benchmarks (I did not look at this at all)
2. a bug is fixed in `claim` (I looked at this, and would prefer a simpler solution)

geoknee:
> I hesitate to approve this PR, since it combines two changes:
> 
> 1. sanity checks are added to gas benchmarks (I did not look at this at all)
> 2. a bug is fixed in `claim` (I looked at this, and would prefer a simpler solution)

For 1: I think we have an established pattern of committing extra test coverage along with a fix. The "sanity checks" do that - they would cause a test failure when run against the unpatched code.  I could spend some time to remove those changes and make a new PR for them if you want?

For 2: We could merge the changes that fix the bug and then work on a follow up PR with subsequent improvements?

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
geoknee: If you view the changes with "hide whitespace changes" enabled, you will get a nice diff showing the two additional lines.

packages/nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -230,3 +243,50 @@ export async function challengeChannelAndExpectGas(
...
+  LforG: BigNumberish;
+  G: BigNumberish;
+  J: BigNumberish;
+  X: BigNumberish;
+}
+
+/**
+ * Asserts the ETH balance of the supplied ethereum account addresses and the ETH holdings in the statechannels asset holding contract for the supplied channelIds.
+ */
+export async function assertETHSanityChecks(
```
andrewgordstewart, line 263: Given that this is meant to be an "actually specifying test helper", and not just a sanity check, a name like `assertEthBalances` would probably be more appropriate?

andrewgordstewart:
> > I hesitate to approve this PR, since it combines two changes:
> > 
> > 1. sanity checks are added to gas benchmarks (I did not look at this at all)
> > 2. a bug is fixed in `claim` (I looked at this, and would prefer a simpler solution)
> 
> For 1: I think we have an established pattern of committing extra test coverage along with a fix. The "sanity checks" do that - they would cause a test failure when run against the unpatched code. I could spend some time to remove those changes and make a new PR for them if you want?

It wasn't really clear to me that adding these assertions to the gas benchmark suite was the right choice. That choice seemed to then create two different unit tests for asset management logic. It is then more difficult to understand how `nitro-protocol` is tested, and terminology is now muddled: the `benchmark` test is not just benchmarking, but is also testing the logic in the system itself. But I did not aim to comment on that choice, as **it is not security related**.

Now that I think about it, I believe your intention with the sanity checkers was to _test the benchmark suite itself_:
- we aimed to benchmark `claim`, intending to take a certain action, eg. "claim some of alice's funds"
- we were not doing that, because the benchmark was not performing the intended action. as a result, our benchmark was not benchmarking the correct behaviour
- the sanity checks that are now in place check that we're benchmarking the intended actions, by checking "after claiming some of alice's funds, alice's account balance has increased"

Is that understanding correct? If so, then I would have preferred separate PRs for those two separate pieces of work, because they really are separate!

> For 2: We could merge the changes that fix the bug and then work on a follow up PR with subsequent improvements?

Yes, let's do that.

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
andrewgordstewart: ah, neat trick!

andrewgordstewart: I'm not sure I understand these test cases.

In isolation, I'm not sure I can figure out what "absent guarantee entry" is supposed to mean. In the context of this PR, I think it means that the destination A in `[A, I, X]` is missing from the outcome of the target channel. Is that correct?

If that's the case, maybe this is more clear?
```suggestion
    ${'18. guarantee destination is not in target outcome, indices=[]'}               | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[]}   | ${{x: 10, I: 0}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
    ${'19. guarantee destination is not in target outcome, indices=[1]'}              | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[1]}  | ${{x: 10, I: 0}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
```

andrewgordstewart: In addition, I am not sure that the tests are finding the bug. When I run these test cases on `master`, they pass!

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
andrewgordstewart: I checked that [here](https://github.com/statechannels/statechannels/pull/3745), and found that the test cases that were added did not fail on master. Are you sure there's a bug in the current implementation?

After reviewing the implementation of `computeNewAllocationFromGuarantee` as well as the current implementation of `_claim`, I am not convinced that there's a bug:
- `payouts[k]` is set [here](https://github.com/statechannels/statechannels/blob/f0cc53fdc3ecf64c12535f9f83e86bf7a0f37f0d/packages/nitro-protocol/contracts/MultiAssetHolder.sol#L325). Note that `payouts[k]` is marking the amount that should get paid to `allocation[indices[k]].destination`
- [later](https://github.com/statechannels/statechannels/blob/f0cc53fdc3ecf64c12535f9f83e86bf7a0f37f0d/packages/nitro-protocol/contracts/MultiAssetHolder.sol#L371), when deciding where to send `payouts[j]`, we set `destination = allocation[indices[j]]`

This is more complicated than it needs to be, but it seems to be correct?

andrewgordstewart (changes requested):
I got _pretty_ confused trying to review this with my security hat on. I wanted to understand:
- how the bug was not caught by our current test suite
- what parts of the code lead to the bug
- how we're going to prevent this kind of bug in the future

etc.

To answer these questions, I reviewed the whole implementation of `claim`.

I am not sure that I am satisfied with where things would be left after merging these changes. My concerns are:

1. The new test helpers added seem to be sanity checkers for the benchmark test, and not tests of the logic of `claim`. (IMO they should not test the logic of claim) Ideally, they would be introduced in a separate PR.
2. The new test cases added to `claim` [don't appear to catch the bug](https://github.com/statechannels/statechannels/pull/3745).
3. There is a [simpler solution](https://github.com/statechannels/statechannels/pull/3744) to the apparent bug in `claim` which should probably be employed.
4. There [doesn't actually appear](https://github.com/statechannels/statechannels/pull/3740/files#r689869498) to be a bug in `claim`.

Given 2 & 4, I don't think it makes sense to merge this to fix a security hole:
- either there is a bug, and our test coverage is not finding it
- or there is no bug, and we are making an unnecessary change to `claim`. (if we're going to make a change, let's make [the one that reduces complexity](https://github.com/statechannels/statechannels/pull/3744))

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
geoknee: Those tests should fail, but don't fail. I'll work on fixing that.

geoknee:
This thread has gotten very noisy now, so I am going to close this.

#3746 the sanity check stuff has been moved here
#3747 extending unit tests to catch the bug in claim + fixing the bug has been moved here

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
andrewgordstewart: Why should the tests fail? Is my understanding outlined above incorrect?

geoknee: Let's tale one test case in particular [(18.)](https://github.com/statechannels/statechannels/pull/3745/files?diff=unified&w=1).

We are claiming a guarantee of `[A,I,x]` (funded with 10) targeting an outcome of `{x:10,I:10}` with `indices=[]` signalling our wish to claim all of the target allocations. `I` should be paid 10 and `x` should not be paid anything. Instead, `x` is paid 10 and `I` is not paid anything. 

The problem with the implementation is that `k` remains at its initial value of `0`; so the payout calculated for `I` is stored in the slot for `x`. The outer function then pays the coins to `x`.

Unfortunately the tests are not actually checking the payouts, hence the tests pass. I have remedied that on #3747.

PR #3742, opened 2021-08-16
refactor: remove unnecessary index k

andrewgordstewart:
note that previously, the expression
`(indices.length == 0) || ((k < indices.length) && (indices[k] == i))`
would always evaluate to `true`. This meant that `i == k` was always true.

This refactor removes the unnecessary index k.


geoknee:
This change causes 2 of the test cases in `claim.test.ts` to fail.

andrewgordstewart:
closing, as this is a mistake!

PR #3744, opened 2021-08-16
refactor: remove payouts from _computeNewAllocationWithGuarantee

andrewgordstewart:
note that previously, information was duplicated, since
newAllocation[i].amount + payouts[i] == allocation[i].amount
always held true


geoknee:
superceded by #3747

PR #3746, opened 2021-08-17
add an assertEthBalancesAndHoldings fn

geoknee:
to gas benchmarks.

 Buy calling this function at judicious moments during the benchmarking script,
 we can be sure the script achieves the defined goal of moving money through a network of channels to Alice

### DIFF
--- a/packages/nitro-protocol/gas-benchmarks/fixtures.ts
+++ b/packages/nitro-protocol/gas-benchmarks/fixtures.ts
@@ -1,6 +1,6 @@
 import {Signature} from '@ethersproject/bytes';
 import {Wallet} from '@ethersproject/wallet';
-import {ContractReceipt, ethers} from 'ethers';
+import {BigNumber, BigNumberish, constants, ContractReceipt, ethers} from 'ethers';
 
 import {
   Allocation,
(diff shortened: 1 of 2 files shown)


packages/nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -230,3 +243,50 @@ export async function challengeChannelAndExpectGas(
...
+    Bob: Bob.address,
+    Ingrid: Ingrid.address,
+  };
+  await Promise.all([
+    ...Object.keys(ethHoldings).map(async key => {
+      expect(
+        (await nitroAdjudicator.holdings(constants.AddressZero, internalDestinations[key])).eq(
+          BigNumber.from(ethHoldings[key])
+        )
+      ).toBe(true);
```
NiloCK, line 284:
```suggestion
  await Promise.all([
    ...Object.keys(ethHoldings).map(async channel => {
      expect(
        (await nitroAdjudicator.holdings(constants.AddressZero, internalDestinations[channel])).eq(
          BigNumber.from(ethHoldings[channel])
        )
      ).toBe(true);
```

packages/nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -230,3 +243,50 @@ export async function challengeChannelAndExpectGas(
...
+    }),
+    ...Object.keys(ethBalances).map(async key => {
+      expect(
+        (await provider.getBalance(externalDestinations[key])).eq(BigNumber.from(ethBalances[key]))
+      ).toBe(true);
```
NiloCK, line 289:
```suggestion
    ...Object.keys(ethBalances).map(async participant) => {
      expect(
        (await provider.getBalance(externalDestinations[participant])).eq(BigNumber.from(ethBalances[participant]))
      ).toBe(true);
```

NiloCK (approved):
Looks good. Nice change to add `amountForX`.

Suggestions were made in-browser with no editor support, so they may be wrong!

PR #3747, opened 2021-08-17
fix claim.test.ts and remove payouts from computeNewAllocationWithGuarantee in order to fix critical bug

geoknee:
The `claim` function uses an array called `payouts` (internally) to build up payouts for each of the destinations in the target channel at the supplied `indices`. We are not currently updating this array correctly in general: particularly in the case where the `guarantee` has an entry that is absent from the target outcome. Essentially, a different party gets the funds that should go to their rightful owner.

Example: We are claiming a guarantee of `[A,I,x]` (funded with `10`) targeting an outcome of `{x:10,I:10}` with `indices=[]` signalling our wish to claim all of the target allocations. `I` should be paid 10 and `x` should not be paid anything. Instead, `x` is paid `10` and `I` is not paid anything. The problem with the current implementation is that inside `_computeNewAllocationWithGuarantee`, `k` remains at its initial value of `0`; so the payout calculated for `I` is stored in the slot for `x`. The outer function then pays the coins to `x`.

The fix I have opted for here is to remove `payouts` entirely. Seeing as it contains redundant information, we can simply recompute the payouts from the changes to `allocations` in the outer function. There is now implicitly a payout for every allocation item (even if some of those are zero).

If we like that pattern, we can use it inside `transfer`, too. 
 

### Changes
Strengthen and extend test coverage to catch this bug. 
Fix the bug. 

- [x] Is it security related?

---
### Code quality
### Project management

Note to reviewer: recommend viewing with whitespace changes hidden: 

Screenshot 2021-08-17 at 16 43 32

### DIFF
--- a/packages/nitro-protocol/contracts/MultiAssetHolder.sol
+++ b/packages/nitro-protocol/contracts/MultiAssetHolder.sol
@@ -352,7 +352,6 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         (
             Outcome.AllocationItem[] memory newAllocation,
             ,
-            uint256[] memory payouts,
             uint256 totalPayouts
         ) = _computeNewAllocationWithGuarantee(initialHoldings, allocation, indices, guarantee);
 
@@ -366,14 +365,15 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         // INTERACTIONS
         // *******
 
-        for (uint256 j = 0; j < payouts.length; j++) {
-            if (payouts[j] > 0) {
-                bytes32 destination = allocation[indices.length > 0 ? indices[j] : j].destination;
+        for (uint256 j = 0; j < newAllocation.length; j++) {
+            uint256 payout = allocation[j].amount - newAllocation[j].amount;
+            if (payout > 0) {
+                bytes32 destination = newAllocation[j].destination;
                 // storage updated BEFORE external contracts called (prevent reentrancy attacks)
                 if (_isExternalDestination(destination)) {
-                    _transferAsset(asset, _bytes32ToAddress(destination), payouts[j]);
+                    _transferAsset(asset, _bytes32ToAddress(destination), payout);
                 } else {
-                    assetHoldings[destination] += payouts[j];
+                    assetHoldings[destination] += payout;
                 }
             }
         }
@@ -392,13 +392,10 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         returns (
             Outcome.AllocationItem[] memory newAllocation,
             bool allocatesOnlyZeros,
-            uint256[] memory payouts,
             uint256 totalPayouts
         )
     {
         // `indices == []` means "pay out to all"
-        // Note: by initializing payouts to be an array of fixed length, its entries are initialized to be `0`
-        payouts = new uint256[](indices.length > 0 ? indices.length : allocation.length);
         totalPayouts = 0;
         allocatesOnlyZeros = true; // switched to false if there is an item remaining with amount > 0
         uint256 surplus = initialHoldings; // tracks funds available during calculation
@@ -429,8 +426,8 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
                         // only if specified in supplied indices, or we if we are doing "all"
                         // reduce the new allocationItem.amount
                         newAllocation[i].amount -= affordsForDestination;
-                        // increase the relevant payout
-                        payouts[k] += affordsForDestination;
+                        // the amount subtracted should be paid out to this destination
+                        // increase the total payouts
                         totalPayouts += affordsForDestination;
                         // move on to the next supplied index
                         ++k;
--- a/packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
+++ b/packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
@@ -40,36 +40,55 @@ const addresses = {
 const reason5 = 'Channel not finalized';
 const reason6 = 'targetAsset != guaranteeAsset';
 
-// 1. claim G1 (step 1 of figure 23 of nitro paper)
-// 2. claim G2 (step 2 of figure 23 of nitro paper)
-// 3. claim G1 (step 1 of alternative in figure 23 of nitro paper)
-// 4. claim G2 (step 2 of alternative of figure 23 of nitro paper)
+const names = {
+  1: ' 1. straight-through guarantee, 3 destinations', // 1. claim G1 (step 1 of figure 23 of nitro paper)
+  2: ' 2. swap guarantee,             2 destinations', // 2. claim G2 (step 2 of figure 23 of nitro paper)
+  3: ' 3. swap guarantee,             3 destinations', // 3. claim G1 (step 1 of alternative in figure 23 of nitro paper)
+  4: ' 4. straight-through guarantee, 2 destinations', // 4. claim G2 (step 2 of alternative of figure 23 of nitro paper)
+  5: ' 5. allocation not on chain',
+  6: ' 6. guarantee not on chain',
+  7: ' 7. swap guarantee, overfunded, 2 destinations',
+  8: ' 8. underspecified guarantee, overfunded      ',
+  9: ' 9. (all) straight-through guarantee, 3 destinations',
+  10: '10. (all) swap guarantee,             2 destinations',
+  11: '11. (all) swap guarantee,             3 destinations',
+  12: '12. (all) straight-through guarantee, 2 destinations',
+  13: '13. (all) allocation not on chain',
+  14: '14. (all) guarantee not on chain',
+  15: '15. (all) swap guarantee, overfunded, 2 destinations',
+  16: '16. (all) underspecified guarantee, overfunded      ',
+  17: '17. guarantee and target assets do not match',
+  18: '18. guarantee includes destination missing in target outcome, indices=[], lowest priority is a channel',
+  19: '19. guarantee includes destination missing in target outcome, indices=[1], lowest priority is a channel',
+  20: '20. guarantee includes destination missing in target outcome, indices=[], lowest priority is an external destination', // this guarantee is typical of the virtual funding construction
+  21: '21. guarantee includes destination missing in target outcome, indices=[1], lowest priority is an external destination', // this guarantee is typical of the virtual funding construction
+};
 
 // Amounts are valueString representations of wei
 describe('claim', () => {
   it.each`
-    name                                                              | heldBefore | guaranteeDestinations | tOutcomeBefore        | indices | tOutcomeAfter         | heldAfter        | payouts         | reason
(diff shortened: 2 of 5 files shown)


packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -366,14 +365,15 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         // INTERACTIONS
         // *******
 
-        for (uint256 j = 0; j < payouts.length; j++) {
-            if (payouts[j] > 0) {
-                bytes32 destination = allocation[indices.length > 0 ? indices[j] : j].destination;
+        for (uint256 j = 0; j < newAllocation.length; j++) {
```
andrewgordstewart, line 368: note that you can also iterate over `indices[j]` to save some gas when `indices.length` is small and `allocation.length` is large

packages/nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -429,8 +426,8 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
                         // only if specified in supplied indices, or we if we are doing "all"
                         // reduce the new allocationItem.amount
                         newAllocation[i].amount -= affordsForDestination;
-                        // increase the relevant payout
-                        payouts[k] += affordsForDestination;
+                        // the amount subtracted should be paid out to this destination
+                        // increase the total payouts
                         totalPayouts += affordsForDestination;
                         // move on to the next supplied index
                         ++k;
```
andrewgordstewart, line 433: > The problem with the current implementation is that inside _computeNewAllocationWithGuarantee, k remains at its initial value of 0

I still don't understand how this could be true, as `k` is incremented here!

As such, I have a really hard time understanding the bug. If `k` is not incremented, will the new allocation be correctly computed?

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
```diff
@@ -40,24 +48,28 @@ const reason6 = 'targetAsset != guaranteeAsset';
...
+    ${'17. guarantee and target assets do not match'}                 | ${{g: 1}}  | ${['A', 'B']}         | ${{A: 5, B: 5}}       | ${[1]}  | ${{A: 4, B: 5}}       | ${{g: 0}}        | ${{A: 1}}       | ${reason6}
+    ${'18. absent guarantee entry, indices=[], bad guarantee order'}  | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[]}   | ${{x: 10, I: 0}}      | ${{g: 0}}        | ${{I: 10}}      | ${undefined}
+    ${'19. absent guarantee entry, indices=[1], bad guarantee order'} | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[1]}  | ${{x: 10, I: 0}}      | ${{g: 0}}        | ${{I: 10}}      | ${undefined}
+    ${'20. absent guarantee entry, indices=[]'}                       | ${{g: 10}} | ${['A', 'x', 'I']}    | ${{x: 10, I: 10}}     | ${[]}   | ${{x: 0, I: 10}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
+    ${'21. absent guarantee entry, indices=[1]'}                      | ${{g: 10}} | ${['A', 'x', 'I']}    | ${{x: 10, I: 10}}     | ${[0]}  | ${{x: 0, I: 10}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
```
andrewgordstewart, line 72: nit: One nasty thing about the table is it's hard to explain what the case is -- "absent guarantee entry" is kind of cryptic!

I'd probably appreciate something like this, if I were reading the test case as is.
```suggestion
    ${'18. guarantee includes destination missing in target's outcome, indices=[], bad guarantee order'}  | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[]}   | ${{x: 10, I: 0}}      | ${{g: 0}}        | ${{I: 10}}      | ${undefined}
    ${'19. absent guarantee entry, indices=[1], bad guarantee order'} | ${{g: 10}} | ${['A', 'I', 'x']}    | ${{x: 10, I: 10}}     | ${[1]}  | ${{x: 10, I: 0}}      | ${{g: 0}}        | ${{I: 10}}      | ${undefined}
    ${'20. absent guarantee entry, indices=[]'}                       | ${{g: 10}} | ${['A', 'x', 'I']}    | ${{x: 10, I: 10}}     | ${[]}   | ${{x: 0, I: 10}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
    ${'21. absent guarantee entry, indices=[1]'}                      | ${{g: 10}} | ${['A', 'x', 'I']}    | ${{x: 10, I: 10}}     | ${[0]}  | ${{x: 0, I: 10}}      | ${{g: 0, x: 10}} | ${{}}           | ${undefined}
```

andrewgordstewart: Also, the meaning "bad guarantee order" kind of depends on understanding the virtual funding protocol (VFP), which gets to a deeper point, touching on our discussion during stand-up:

- the behaviour of claim is very general
- yet, we only care about the behaviour of claim in specific cases
- notably, the cases we care most about were missing from this test file (and therefore did not catch the bug this PR fixes)

I think that's something worth reflecting on, if our goal is to prove security of the VFP.

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
```diff
@@ -221,6 +243,18 @@ describe('claim', () => {
...
+        await Promise.all(
+          Object.keys(payouts).map(async destination => {
+            const address = convertBytes32ToAddress(destination);
+            expect(
+              (await provider.getBalance(address)).eq(
+                ethBalancesBefore[address].add(payouts[destination])
+              )
+            ).toBe(true);
+          })
+        );
```
andrewgordstewart, line 257: This is a bit tidier.
```suggestion
        await Promise.all(
          Object.keys(payouts).map(async destination => {
            const address = convertBytes32ToAddress(destination);
            const balanceAfter = await provider.getBalance(address);
            expect(balanceAfter.toString()).toEqual(ethBalancesBefore[address].add(payouts[destination]);
          })
        );
```

andrewgordstewart: If you're willing, test feedback is even better if you:
1. construct the expected and actual `ethBalancesAfter` for each address
2. assert that the expected and actual are equal

```suggestion
       const expectedAfter = {};
       const actualAfter = {};
        await Promise.all(
          Object.keys(payouts).map(async destination => {
            const address = convertBytes32ToAddress(destination);
            const actualAfter[address]  = await provider.getBalance(address);
            const expectedAfter[address] = ethBalancesBefore[address].add(payouts[destination]
          })
        );
        expect(actualAfter).toBeExactlyTheSameAs(expectedAfter);
```
Then, on a test failure you get more than `Expected x to be True, but got False`.

andrewgordstewart (commented):
Left some comments, but awaiting feedback on this point:

> The problem with the current implementation is that inside _computeNewAllocationWithGuarantee, k remains at its initial value of 0

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
geoknee: I agree

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: `k` is incremented here, *but only when the condition on L420 is true*. While `j=0`, the condition is never true in the example I gave in the PR description, so `k` will *not be incremented*. When `j=1, i=1` the condition is true (we found `I` from the guarantee in the target allocation), and `payouts[k]` is increased  -- but  `k` still has its initial value of `0`.

`k` is then incremented afterwards, but we have now run out of funds so will break out of the loops without any more payouts being changed.

When `indices=[]`, `payouts[k]` will be interpreted as the payout for `allocation[k].destination`. So in this case we have `payouts[0] = 10` coins for `allocation[0].destination = x`. Which is wrong. It is in this sense that "not incrementing `k` when we don't find a guarantee destination in the target" is a problem. Of course if we no longer have that interpretation for `payouts`, then it may no longer be a problem -- which is why the fixed implementation having the same logic for incrementing `k` is OK.

There is no such problem when computing the new allocation, since that doesn't use the index `k`; it uses the index `i`. 

If you are having a hard time understanding the bug, I think it is worth remembering that that is a secondary consideration. The more important things to understand are:
1. That the current implementation has a bug. You can see this from the new test cases, which fail before the fix is applied.
2. That the fixed implementation is correct. 

It might be that it is difficult to understand the bug because it is difficult to understand the current implementation. Happily, the fixed implementation is much simpler anyway -- which is probably reason enough to accept if there were no bug in the current implementation.

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
geoknee: 8e7f89c

packages/nitro-protocol/test/contracts/MultiAssetHolder/claim.test.ts
geoknee: d53a69f

packages/nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: I tried this out: but because of needing to switch on the `indices = []` case, the code got significantly messier and ended up costing more gas in at least some scenarios.

I would like to punt this for now, but happy to circle back to it in future.

andrewgordstewart (approved):
Re. https://github.com/statechannels/statechannels/pull/3747#discussion_r691020980

> If you are having a hard time understanding the bug, I think it is worth remembering that that is a secondary consideration. The more important things to understand are:
> 1. That the current implementation has a bug. You can see this from the new test cases, which fail before the fix is applied.
> 2. That the fixed implementation is correct.
> It might be that it is difficult to understand the bug because it is difficult to understand the current implementation. Happily, the fixed implementation is much simpler anyway -- which is probably reason enough to accept if there were no bug in the current implementation.

Yes, this is unquestionably an improvement!

My only hesitance in approving this is:
- previously there was a bug that we believed was due to `k` not incrementing
- but, I don't see any logic that implies that `k` will now increment
- 
which leads me to wonder if there is still a bug.

_That being said_, this PR does provide a failing test case which should pass, then fixes it, and is an overall improvement in the codebase. So, we might as well merge it?

geoknee:
Yeah I think the issue is that the bug is not solely "due to k not incrementing" but due to that plus some other facts about our implementation. The fix ended up altering those other facts.

PR #3750, opened 2021-08-19
Update wire-format to comply with exit-format

geoknee:
The wire format consists of typescript types, a json schema and validator functions for messages sent between nitro wallets. 

It is necessary to update the package if we want wallets to work with the exit format, since they need to pass messages (for signing by counterparties) that can eventually be submitted to chain.

The approach taken here is:
*  ensure the new `metadata` and `allocationType` fields are present and correct, as well as
*  renaming `allocationItems` to `allocations` (as per the exit format). 
* I have removed the `Guarantee` outcome type, since this is now just an allocation with certain metadata. 
* I have also removed the `Allocation(s)` type as this collides with new types coming out of exit format. It is also unnecessary given the removal of `Guarantee` type.
* regenerate the schema
* update the sample messages so the tests pass

I considered whether these changes were really necessary -- in other packages we can probably get away with reusing different types and writing adapter code. But here it seems to me that it really is necessary to have things compatible here, since we do runtime checks (on inter-wallet messages) against a schema generated from the types in _this_ package, and go on to use those messages to construct transactions that need to work against types defined in the exit-format package. 

I also considered whether we should be duplicating types here, or in fact just re-importing them from the exit format (seeing as they seem to need to be tightly coupled). I didn't reach a conclusion on that, and it felt like mission creep in any case. 

⚠️ reminder: CI is *not* expected or required to pass (yet)

---
### Code quality
### Project management

### DIFF
--- a/packages/wire-format/src/__tests__/good_sample_messages.ts
+++ b/packages/wire-format/src/__tests__/good_sample_messages.ts
@@ -27,17 +27,22 @@ export const goodMessage = {
         isFinal: false,
         outcome: [
           {
-            allocationItems: [
+            allocations: [
               {
                 amount: '0x00000000000000000000000000000000000000000000000006f05b59d3b20000',
-                destination: '0x00000000000000000000000063e3fb11830c01ac7c9c64091c14bb6cbaac9ac7'
+                destination: '0x00000000000000000000000063e3fb11830c01ac7c9c64091c14bb6cbaac9ac7',
+                allocationType: 0,
+                metadata: '0x'
               },
               {
                 amount: '0x00000000000000000000000000000000000000000000000006f05b59d3b20000',
-                destination: '0x00000000000000000000000063e3fb11830c01ac7c9c64091c14bb6cbaac9ac7'
+                destination: '0x00000000000000000000000063e3fb11830c01ac7c9c64091c14bb6cbaac9ac7',
+                allocationType: 0,
+                metadata: '0x'
               }
             ],
-            asset: '0x4ad3F07BEFDC54511449A1f553E36A653c82eA57'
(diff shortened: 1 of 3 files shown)


packages/wire-format/src/__tests__/good_sample_messages.ts
```diff
@@ -27,17 +27,22 @@ export const goodMessage = {
         isFinal: false,
         outcome: [
           {
-            allocationItems: [
+            allocations: [
               {
                 amount: '0x00000000000000000000000000000000000000000000000006f05b59d3b20000',
-                destination: '0x00000000000000000000000063e3fb11830c01ac7c9c64091c14bb6cbaac9ac7'
+                destination: '0x00000000000000000000000063e3fb11830c01ac7c9c64091c14bb6cbaac9ac7',
+                allocationType: 0,
```
andrewgordstewart, line 34: Should this be the number literal `0`? It seems as though `0` has special meaning defined by `nitro-protocol`.

andrewgordstewart: fair point!

NiloCK:
Thanks for the prod on reviewing. My general impression is that a package like this is a good thing to have in protocol development, and that it makes things like future interoperability between different versions feasible.

> I also considered whether we should be duplicating types here, or in fact just re-importing them from the exit format (seeing as they seem to need to be tightly coupled). I didn't reach a conclusion on that, and it felt like mission creep in any case.

This is an interesting question, but some part of me says that the current solution is correct. Schema is defined in one place, implementation in another. This guards against making breaking changes in the exit-format without noticing.

NiloCK (approved):

PR #3754, opened 2021-08-25
Update wallet core and client-api-schema to work with exit format

geoknee:
`yarn prepare` and `yarn test` should work in `wallet-core` but *not* in `browser-wallet` or `server-wallet`. I couldn't see any easy way to preserve the types exported by wallet core, so unfortunately this change is breaking and more work will be needed to convert those packages too.

We cannot easily maintain the current `SimpleGuarantee` type because it does not have an `amount` field. In the exit format, guarantee allocations require an `amount`. 

~Early feedback appreciated on the following design decision: the internal types in wallet core have optional fields. These fields are given default values when converting to (serializing) types for the wire or for the app. Although this seems kind of nasty, it leads to a smaller change set here and can be fixed later on.~ Turns out this wasn't too hard to fix now, in fact.


geoknee:
In favour of #3762

